Issue #16483, opened 2021-03-23
SDK managed workload installer


sfoslund:
Top level: Implement the [.NET SDK managed workload installer](https://github.com/dotnet/designs/blob/4cd0cf814d875301cc291b4ae8f1068363b68c15/accepted/2021/workloads/workload-installation.md#net-sdk-managed-installation), which will be used to local/non-admin scenarios.

- [x] Basic nuget package download API: https://github.com/dotnet/sdk/pull/16443. This will be expanded later, see https://github.com/dotnet/sdk/issues/16470
- [x] Implement workload install process (layer above IAL): [spec](https://github.com/dotnet/designs/blob/4cd0cf814d875301cc291b4ae8f1068363b68c15/accepted/2021/workloads/workload-installation.md#workload-installation-process). This is necessary since this will be the first workload installer implemented. https://github.com/dotnet/sdk/pull/16530
- [x] Implement installer abstraction layer: [spec](https://github.com/dotnet/designs/blob/4cd0cf814d875301cc291b4ae8f1068363b68c15/accepted/2021/workloads/workload-installation.md#installer-abstraction-api) https://github.com/dotnet/sdk/pull/16530
- [x] Write first implementation of the installer abstraction layer, the SDK managed installer: [spec](https://github.com/dotnet/designs/blob/4cd0cf814d875301cc291b4ae8f1068363b68c15/accepted/2021/workloads/workload-installation.md#net-sdk-managed-ial-implementation). 
  - [x] Write basic install: https://github.com/dotnet/sdk/pull/16530
  - [x] Write garbage collection: https://github.com/dotnet/sdk/pull/16530
  - [x] Per https://github.com/dotnet/sdk/pull/16530#discussion_r608237834, refactor workload installation into a separate project https://github.com/dotnet/sdk/pull/16823
  - [x] Write offline cache logic: https://github.com/dotnet/sdk/pull/17324
  - [x] Add `--print-download-link-only` implementation (required for VS4Mac): https://github.com/dotnet/sdk/pull/17324
  - [x] Write manifest updates logic: https://github.com/dotnet/sdk/pull/16891
- [x] Preserve executable permission on unix https://github.com/dotnet/sdk/issues/16566
- [x] A command to detect if a workload is installed already (required by VSMac installer) #17324 
- [x] A command To install previously cached download packages #17324 
- [x] Add manifest caching to offline cache install (more detail: https://github.com/dotnet/sdk/pull/17324#discussion_r627964123)
[...]

PR #17324, opened 2021-04-30
Adding offline cache for workload install

sfoslund:
- Adding offline caching for workload packs (not manifests)
- Implementing `--print-download-link-only` option


sfoslund:
Open questions: 

- How should offline caches work with manifest updates? If we download a manifest package to an offline cache but not install it then we'll be caching the packages associated with the old manifest, which will be out of date when we update to the new manifest.

wli3:
@sfoslund we should ignore advertise manifest, and use only the packages as the input, a manifest should be in the cache too. So once the packages are install, the manifest should be updated too. Unless it is an downgrade compared to existing install manifest (there is an update happened later).

The scenario we have is mostly low bandwidth computer. They might spend a week to download all the packages, if a new advertise manifest is update the following week, it is not worth to invalid the existing packages.

wli3:
sorry I am slow to review this

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -69,32 +52,48 @@ public NuGetPackageDownloader(DirectoryPath packageInstallDir, ILogger logger =
...
             {
-                throw new NuGetPackageInstallerException(
+                throw new NuGetPackageDownloaderException(
                     string.Format(LocalizableStrings.FailedToLoadNuGetSource, source.Source));
             }
 
-            string nupkgPath = Path.Combine(_packageInstallDir.Value, packageId.ToString(),
-                packageVersion.ToNormalizedString(),
-                $"{packageId}.{packageVersion.ToNormalizedString()}.nupkg");
+            string nupkgPath = downloadFolder == null ?
```
joeloff, line 59: Should this be IsNullOrWhiteSpace in case empty string is passed in?

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -87,8 +81,21 @@ public void InstallWorkloadPack(PackInfo packInfo, SdkFeatureBand sdkFeatureBand
...
-                            tempFilesToDelete.Add(packagePath);
+                            string packagePath;
+                            if (string.IsNullOrWhiteSpace(offlineCache))
+                            {
+                                packagePath = _nugetPackageDownloader.DownloadPackageAsync(new PackageId(packInfo.ResolvedPackageId), new NuGetVersion(packInfo.Version)).Result;
+                                tempFilesToDelete.Add(packagePath);
+                            }
+                            else
+                            {
+                                _reporter.WriteLine(string.Format(LocalizableStrings.UsingCacheForPackInstall, packInfo.Id, packInfo.Version, offlineCache));
```
wli3, line 92: Offline cache should have a matching manifest file and we should install according to that. And once the packs are installed, the manfiest file should be installed too

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -69,32 +52,48 @@ public NuGetPackageDownloader(DirectoryPath packageInstallDir, ILogger logger =
...
                 destinationStream,
                 _cacheSettings,
                 _logger,
                 cancellationToken);
 
             if (!success)
             {
-                throw new NuGetPackageInstallerException(
-                    $"Downloading {packageId} version {packageVersion.ToNormalizedString()} failed");
+                throw new NuGetPackageDownloaderException(
```
joeloff, line 74: Do the NuGet APIs provide more detail that we potentially hide, e.g. let's say the service returns HTTP 503 which may contain retry indicators, etc. Or there's an authentication issue. Do we risk hiding the root cause of the failure from the user?

src/Cli/dotnet/NugetPackageDownloader/INuGetPackageDownloader.cs
```diff
@@ -10,7 +10,14 @@ namespace Microsoft.DotNet.Cli.NuGetPackageDownloader
 {
     internal interface INuGetPackageDownloader
     {
-        Task<string> DownloadPackageAsync(PackageId packageId, NuGetVersion packageVersion = null,
+        Task<string> DownloadPackageAsync(PackageId packageId,
+            NuGetVersion packageVersion = null,
+            PackageSourceLocation packageSourceLocation = null,
+            bool includePreview = false,
+            string downloadFolder = null);
```
wli3, line 17: nit: use DirectoryPath type instead of string

src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadInstaller.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 {
     internal interface IWorkloadInstaller : IInstaller
     {
-        void InstallWorkload(WorkloadId workloadId, bool useOfflineCache = false);
+        void InstallWorkload(WorkloadId workloadId, string offlineCache = null);
```
wli3, line 11: nit: use DirectoryPath type instead of string

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -116,16 +116,63 @@ public void GivenWorkloadInstallItCanUpdateInstalledManifests()
...
+            var mockWorkloadIds = new WorkloadId[] { new WorkloadId("xamarin-android") };
+            var cachePath = "mockCachePath";
+            var parseResult = Parser.GetWorkloadsInstance.Parse(new string[] { "dotnet", "workload", "install", "xamarin-android", "--download-to-cache", cachePath });
+            (_, var installManager, var installer, _, _, _) = GetTestInstallers(parseResult);
+
+            installManager.Execute();
+
+            // 8 android pack packages
+            installer.CachedPacks.Count.Should().Be(8);
+            installer.CachedPacks.Where(pack => pack.Id.Contains("Android")).Count().Should().Be(8);
```
wli3, line 138: should also have a manifest here. The older manifest file thing is annoying but we do need that.

src/Tests/dotnet-workload-install.Tests/GivenNetSdkManagedWorkloadInstall.cs
```diff
@@ -300,7 +300,69 @@ public void GivenManagedInstallItCanInstallManifestVersion()
...
+            // Otherwise install should be normal
+            mockNugetInstaller.ExtractCallParams.Count.Should().Be(1);
+            mockNugetInstaller.ExtractCallParams[0].ShouldBeEquivalentTo((nupkgPath, Path.Combine(dotnetRoot, "metadata", "temp", $"{packInfo.Id}-{packInfo.Version}-extracted")));
+            var installationRecordPath = Path.Combine(dotnetRoot, "metadata", "workloads", "InstalledPacks", "v1", packInfo.Id, packInfo.Version, version);
+            File.Exists(installationRecordPath).Should().BeTrue();
+            Directory.Exists(packInfo.Path).Should().BeTrue();
+        }
+
+        [Fact]
+        public void GivenManagedInstallItCanErrorsWhenMissingOfflineCache()
```
wli3, line 355: Could you also add an test for print URLs?

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
wli3: I don't think so so far. I haven't seen info exposed from http layer

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
sfoslund: Where should that manifest come from? Should we add logic to download the latest manifest package, extract it and determine which packs to download and cache based on that manifest? cc @dsplaisted

wli3: > Should we add logic to download the latest manifest package, extract it and determine which packs to download and cache based on that manifest

I think so. For a normal install, we download the latest manifest first and use it to determine the installation too right?

sfoslund: Yes, although with the normal installation we install the new manifests to the manifest directory before we start using them. You're saying we should read the manifests from the temp directory, not fully install them, right? Or are we okay to fully install them before determining the packs to download?

src/Tests/dotnet-workload-install.Tests/GivenNetSdkManagedWorkloadInstall.cs
sfoslund: I wrote one, it's in the other test file: https://github.com/dotnet/sdk/pull/17324/files#diff-c8bc89094b206b506f35f0cb9ac6e55591ae2468130efe0a19dab1810c0f0c1cR161

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
wli3: > Yes, although with the normal installation we install the new manifests to the manifest directory before we start using them

Hmm, I don't think that's spec's behavior. However, if we do that first, and the install failed in the middle, we would invalid all installed workload since they mismatch with the manifest and should be garbage collected.

The temp folder is secure so we can do our work base on the temp manifest.

> we'll have to create a new manifest resolver to determine which packs to install

We cannot just point the manifest resolver  to the temp manifests? 

I think we kind of have too. Manifest should be always in sync with the package as part of the installation.

wli3: But if it is too hard we can still let offline cache to have a manifest package. And the first action of offline install is to install this offline manifest, and the do the rest. If the manifest is older than what's install, stop

For "if we do that first, and the install failed in the middle, we would invalid all installed workload since they mismatch with the manifest and should be garbage collected." This could be not ideal user experience. We could fix that after we ship

sfoslund: > However, if we do that first, and the install failed in the middle, we would invalid all installed workload since they mismatch with the manifest and should be garbage collected.

Yes, that's why any manifest updates are part of the rollback, so if install fails we roll back the manifests as well as the packs.

> We cannot just point the manifest resolver to the temp manifests?

I haven't looked at it closely, but right now the manifest resolver takes in the dotnet root directory and appears to have some logic that involves more than just the manifests directory. I'll take a closer look and work on finding a way to install and extract to a temp directory, although it might be big enough that I'll want to add downloading manifests to an offline cache in a separate PR

wli3: So just update manifest from the offline cache instead of from the internet might work.

src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
```diff
@@ -180,6 +183,9 @@
   <data name="PrintDownloadLinkOnlyDescription" xml:space="preserve">
     <value>Only print the list of links to download without downloading it.</value>
   </data>
+  <data name="DownloadToCacheOptionDescription" xml:space="preserve">
+    <value>Download packs to the offline cache.</value>
```
dsplaisted, line 187:
```suggestion
    <value>Download packages needed to install a workload to a folder which can be used for offline installation.</value>
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
```diff
@@ -42,8 +42,9 @@ internal static class WorkloadInstallCommandParser
         public static readonly Option IncludePreviewOption =
             new Option<bool>("--include-previews", LocalizableStrings.IncludePreviewOptionDescription);
 
-        public static readonly Option FromCacheOption =
-            new Option<string>("--from-cache", LocalizableStrings.FromCacheOptionDescription) { };
+        public static readonly Option FromCacheOption = new Option<string>("--from-cache", LocalizableStrings.FromCacheOptionDescription);
+
+        public static readonly Option DownloadToCacheOption = new Option<string>("--download-to-cache", LocalizableStrings.DownloadToCacheOptionDescription);
```
dsplaisted, line 47: @KathleenDollard, were you going to work on a spec for what the CLI commands for workloads would be?

In this PR, it looks like the following would be how to create and use a cache:

- `dotnet workload install ios android --download-to-cache "c:\netsdkofflinecache"
- `dotnet workload install ios android --from-cache "c:\netsdkofflinecache"

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
dsplaisted: > So does that mean we'll need a full set of manifests in a temp directory, then we'll have to create a new manifest resolver to determine which packs to install? It seems like a lot of duplicated work and since the manifest reader/ resolver works off of the dotnet root directory right now, we'll need to make significant changes there to enable working from a temp directory.

Yes, this is what I think we should basically do.

It doesn't seem like the code changes will be too bad to me.  We'll want a new implementation of `IWorkloadManifestProvider` instead of `SdkDirectoryWorkloadManifestProvider`, which we can point at a temporary folder where we put the updated manifests.  I think we can use the existing `WorkloadResolver` class.  It only uses the `dotnetRootPath` to locate the runtime identifier chain (which is still what we want to do in the cache creation case), and to determine the path where workload packs should be installed (which we shouldn't need to do if we're just downloading them to a folder).

If the `--skip-manifest-update` option is specified, than rather then getting the latest manifest packages, we should get the versions matching what is currently installed.

When we install workloads using an offline cache, we should first update the advertising manifests using the manifests from the offline cache.  If the latest version of a manifest in the offline cache is older than what is currently installed then it should error out.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -87,8 +81,21 @@ public void InstallWorkloadPack(PackInfo packInfo, SdkFeatureBand sdkFeatureBand
...
+                                packagePath = Path.Combine(offlineCache, $"{packInfo.ResolvedPackageId}.{packInfo.Version}.nupkg");
+                                if (!File.Exists(packagePath))
+                                {
+                                    throw new Exception(string.Format(LocalizableStrings.CacheMissingPackage, packInfo.ResolvedPackageId, packInfo.Version, offlineCache));
```
dsplaisted, line 96: If possible, it would probably be a lot better user experience if we provided the full list of packages that were missing from the offline cache, rather than just the first one we hit.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -266,5 +215,58 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
...
+            }
+            else
+            {
+                throw new NotImplementedException();
+            }
+
+            return packageUrls;
+        }
+		
+		private void DownloadToOfflineCache(IEnumerable<WorkloadId> workloadIds, string offlineCache)
```
dsplaisted, line 246: Nit: Tabs instead of spaces

src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
```diff
@@ -11,6 +11,8 @@ internal interface IWorkloadManifestUpdater
     {
         Task UpdateAdvertisingManifestsAsync(SdkFeatureBand featureBand, bool includePreviews);
 
+        IEnumerable<string> GetManifestsUrls(SdkFeatureBand featureBand, bool includePreviews);
```
dsplaisted, line 14: Consider renaming this to `GetManifestPackageUrls`, I think that would be more clear.

How would this work with the standalone IAL?  Would this be returning the Urls to the NuGet packages that contain the MSIs?  But IIRC this interface isn't part of the IAL.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
sfoslund: Sounds good, thanks for the clarification. I'm going to go ahead and add manifest caching as a separate task in https://github.com/dotnet/sdk/issues/16483 and try to get this PR in without it, since it will help unblock VS for mac.

src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
sfoslund: I'm going to back this out for now since I'll be addressing manifest caching in a separate PR.

wli3 (approved):

PR #17306, opened 2021-04-30
Disable workload resolver sentinel for workload imports

dsplaisted:
The resolver sentinel was previously disabled (or more precisely, switched from opt-in to opt-out) for the workload resolver.  However, it is also used in the .NET SDK targets to determine whether the workload .props and .targets should be imported, so that logic needs to be updated as well in order to enable workloads without the opt-in.

PR #17381, opened 2021-05-03
Forward port #17306 Disable workload resolver sentinel for workload imports to main

lewing:
Hit this when testing p5.  If you'd rather do it some other way feel free to close this.


dsplaisted:
Normally we'd just do a merge from preview 4 to main, is that what happened?

lewing:
yeah we just haven't merged p4 -> main in a while it looks like.  I realized there was more missing and closed this.

PR #17384, opened 2021-05-03
Use ProjectCapability to determine hot reloadibility

pranavkm:
dotnet-watch currently relies on the presence of a value in the launchSettings.json to
determine if a project supports hot reload. This is problematic for workloads other than
ASP.NET Core that do not contain a launchSettings.

This change introduces a ProjectCapability in the SDK that is configured for projects targeting 6.0 or newer
and updates dotnet-watch to start using this capability. In addition, for the "esoteric" cases of BlazorWasm
and BlazorWasm-hosted, we'll rely on the presence of additional capabilities rather than the profile.

### DIFF
--- a/src/BuiltInTools/dotnet-watch/HotReload/CompilationHandler.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/CompilationHandler.cs
@@ -11,7 +11,6 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Build.Locator;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.ExternalAccess.Watch.Api;
@@ -35,26 +34,22 @@ namespace Microsoft.DotNet.Watcher.Tools
 
         public async ValueTask InitializeAsync(DotNetWatchContext context, CancellationToken cancellationToken)
         {
+            Debug.Assert(context.ProjectGraph is not null);
+
             if (_deltaApplier is null)
             {
-                _deltaApplier = context.DefaultLaunchSettingsProfile.HotReloadProfile switch
+                var hotReloadProfile = HotReloadProfileReader.InferHotReloadProfile(context.ProjectGraph, _reporter);
+                _deltaApplier = hotReloadProfile switch
                 {
-                    "blazorwasm" => new BlazorWebAssemblyDeltaApplier(_reporter),
-                    "blazorwasmhosted" => new BlazorWebAssemblyHostedDeltaApplier(_reporter),
-                    _ => new AspNetCoreDeltaApplier(_reporter),
+                    HotReloadProfile.BlazorWebAssembly => new BlazorWebAssemblyDeltaApplier(_reporter),
+                    HotReloadProfile.BlazorHosted => new BlazorWebAssemblyHostedDeltaApplier(_reporter),
+                    _ => new DefaultDeltaApplier(_reporter),
                 };
             }
 
             await _deltaApplier.InitializeAsync(context, cancellationToken);
 
-            if (context.Iteration == 0)
-            {
-                var instance = MSBuildLocator.QueryVisualStudioInstances().First();
-
-                _reporter.Verbose($"Using MSBuild at '{instance.MSBuildPath}' to load projects.");
-                MSBuildLocator.RegisterInstance(instance);
-            }
-            else if (_currentSolution is not null)
+            if (_currentSolution is not null)
             {
                 _currentSolution.Workspace.Dispose();
                 _currentSolution = null;
@@ -243,5 +238,7 @@ namespace Microsoft.DotNet.Watcher.Tools
                 _currentSolution.Workspace.Dispose();
             }
         }
+
+        
     }
 }
--- a/src/BuiltInTools/dotnet-watch/HotReload/HotReloadProfileReader.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/HotReloadProfileReader.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#nullable enable
+
 using System.Collections.Generic;
+using Microsoft.Build.Execution;
 using Microsoft.Build.Graph;
 using Microsoft.Extensions.Tools.Internal;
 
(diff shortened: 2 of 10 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1000,6 +1000,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Remove="ReferenceManagerAssemblies" />
   </ItemGroup>
 
+  <!-- Enable hot reload in 6.0 and newer apps by default -->
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0'))">
+    <ProjectCapability Include="SupportsHotReload" />
+  </ItemGroup>
```
pranavkm, line 1006: @dsplaisted since this change touches the NET.SDK.


@BillHiebert this change adds a new project capability dotnet-watch would use to determine if a project can do hot reload.  My hope was that we could also use this in VS rather than rely on TFMs. How does that sound?

src/BuiltInTools/dotnet-watch/HotReload/CompilationHandler.cs
```diff
@@ -243,5 +238,7 @@ public void Dispose()
                 _currentSolution.Workspace.Dispose();
             }
         }
+
+
```
pranavkm, line 242:
```suggestion
```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -59,6 +61,13 @@ public class Program : IDisposable
 
         public Program(IConsole console, string workingDirectory)
         {
+            // We can register the MSBuild that is bundled with the SDK to perform MSBuild things. dotnet-watch is in
+            // a nested folder of the SDK's root, we'll back up to it.
+            // AppContext.BaseDirectory = $sdkRoot\$sdkVersion\DotnetTools\dotnet-watch\$version\tools\net6.0\any\
+            // MSBuild.dll is located at $sdkRoot\$sdkVersion\MSBuild.dll
+            var sdkRootDirectory = Path.Combine(AppContext.BaseDirectory, "..", "..", "..", "..", "..", "..");
```
captainsafia, line 68: Is there no way to get `sdkRoot` and `sdkVersion` directly without having to travel up the directory tree like this?

src/BuiltInTools/dotnet-watch/HotReload/HotReloadProfileReader.cs
```diff
@@ -0,0 +1,52 @@
...
+
+                foreach (var item in projectCapability)
+                {
+                    if (item.EvaluatedInclude == "AspNetCore")
+                    {
+                        seenAspNetCoreProject = true;
+                    }
+                    else if (item.EvaluatedInclude == "WebAssembly")
+                    {
+                        if (seenAspNetCoreProject)
```
captainsafia, line 31: Is the `AspNetCore` capability guaranteed to be defined before the WASM one?

src/BuiltInTools/dotnet-watch/HotReload/CompilationHandler.cs
```diff
@@ -35,26 +34,22 @@ public CompilationHandler(IReporter reporter)
...
 
-            if (context.Iteration == 0)
-            {
-                var instance = MSBuildLocator.QueryVisualStudioInstances().First();
-
-                _reporter.Verbose($"Using MSBuild at '{instance.MSBuildPath}' to load projects.");
-                MSBuildLocator.RegisterInstance(instance);
-            }
-            else if (_currentSolution is not null)
+            if (_currentSolution is not null)
```
captainsafia, line 52: Why was this check removed?

captainsafia (approved):

dsplaisted:
@dotnet/project-system for use of `ProjectCapability` items outside of VS.  I took a quick glance at this and it seems reasonable to me.

tmeschter:
> @dotnet/project-system for use of `ProjectCapability` items outside of VS. I took a quick glance at this and it seems reasonable to me.

This looks reasonable to me.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
tmeschter: The implication here is that we expect hot reload to work by default on any and every app targeting .NETCoreApp 6.0 and above, correct?

BillHiebert: @pranavkm using a capability rather than checking the tfm is fine by me.

pranavkm: /cc @tommcdon / @DamianEdwards in case you think we should narrow this down. But we generally think all .NET 6 projects would be eligible to be hot reloaded.

tmeschter: For example, will hot reload work on a Xamarin app running on an Android simulator so long as it targets 6.0?

pranavkm: The current plan is for that to work (and for dotnet-watch to support this) by of 6.0. We could narrow it down to the project types that are currently supported (WinForms, WPF, Web, and Blazor) if that's not true closer to 6.0.

src/BuiltInTools/dotnet-watch/HotReload/HotReloadProfileReader.cs
pranavkm: It was kind of incidental because of import ordering. Made a bit more robust with the update.

src/BuiltInTools/dotnet-watch/Program.cs
pranavkm: I couldn't find anything obvious. You can get to the path of the dotnet.exe and the shared framework, but nothing obvious about this thing. Note that the opposite of this code is already baked in to the CLI today: https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/CommandFactory/CommandResolution/DotnetToolsCommandResolver.cs#L39-L43

src/BuiltInTools/dotnet-watch/HotReload/CompilationHandler.cs
pranavkm: We register MSBuild super early in Program.Main and consequently no longer need to do it here.

drewnoakes:
Looks fine to me, from the project system side.

A more general question: would we use this capability to light up hot reload properties in the new project properties UI? Is there anything the customer would want to customise related to hot reload?

BillHiebert:
@drewnoakes No. Hot reload is going to be a central VS setting and there aren't any per-project settings planned, at least for now

PR #17387, opened 2021-05-03
Implement workload list command

sfoslund:


src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -1,59 +1,74 @@
...
+            IReporter reporter = null,
+            IWorkloadInstallationRecordRepository workloadRecordRepo = null,
+            string version = null) : base(result)
         {
+            _reporter = reporter ?? Reporter.Output;
             _machineReadableOption = result.ValueForOption<bool>(WorkloadListCommandParser.MachineReadableOption);
+            _verbosity = result.ValueForOption<VerbosityOptions>(WorkloadListCommandParser.VerbosityOption);
+
+            var sdkVersion = new ReleaseVersion(version ?? Product.Version);
+            var dotnetPath = Path.GetDirectoryName(Environment.ProcessPath);
```
wli3, line 39: nit: later we might want to have an abstraction to control all install paths. So far we have the same calculation for each command. dotnet tools have a similar problem. We later created IToolPackageStore.

wli3 (approved):

PR #16891, opened 2021-04-13
Workload manifest updates

sfoslund:

PR #17455, opened 2021-05-05
Workload manifest updates PR feedback

sfoslund:
Responding to after the fact feedback on https://github.com/dotnet/sdk/pull/16891

### DIFF
--- a/src/Cli/dotnet/NugetPackageInstaller/NuGetPackageDownloader.cs
+++ b/src/Cli/dotnet/NugetPackageInstaller/NuGetPackageDownloader.cs
@@ -72,7 +72,7 @@ namespace Microsoft.DotNet.Cli.NuGetPackageDownloader
 
             if (resource == null)
             {
-                throw new NuGetPackageInstallerException(
+                throw new NuGetPackageNotFoundException(
                     string.Format(LocalizableStrings.FailedToLoadNuGetSource, source.Source));
             }
 
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -83,47 +83,56 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             string packagePath = null;
             string extractionPath = null;
 
-            var adManifestPath = GetAdvertisingManifestPath(_sdkFeatureBand, manifestId);
             try
             {
-                packagePath = await _nugetPackageDownloader.DownloadPackageAsync(GetManifestPackageId(_sdkFeatureBand, manifestId), includePreview: includePreviews);
-            }
-            catch
-            {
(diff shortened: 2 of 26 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -76,59 +78,59 @@ private IEnumerable<ManifestId> GetInstalledManifestIds()
...
-
-                if (Directory.Exists(adManifestPath))
-                {
-                    Directory.Delete(adManifestPath, true);
-                }
-                Directory.CreateDirectory(Path.GetDirectoryName(adManifestPath));
-                FileAccessRetrier.RetryOnMoveAccessFailure(() => Directory.Move(Path.Combine(extractionPath, "data"), adManifestPath));
-
-                _reporter.WriteLine(string.Format(LocalizableStrings.AdManifestUpdated, manifestId));
+                packagePath = await _nugetPackageDownloader.DownloadPackageAsync(GetManifestPackageId(_sdkFeatureBand, manifestId), includePreview: includePreviews);
```
dsplaisted, line 89: Ideally this method would either throw a specific exception or have another way to indicate that no package was found, rather than having to catch any Exception that the method throws.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -76,59 +78,59 @@ private IEnumerable<ManifestId> GetInstalledManifestIds()
...
+            FileAccessRetrier.RetryOnMoveAccessFailure(() => Directory.Move(Path.Combine(extractionPath, "data"), adManifestPath));
 
-                if (!string.IsNullOrEmpty(packagePath) && File.Exists(packagePath))
-                {
-                    File.Delete(packagePath);
-                }
+            _reporter.WriteLine(string.Format(LocalizableStrings.AdManifestUpdated, manifestId));
+
+            // Clean up temp dirs
+            if (!string.IsNullOrEmpty(extractionPath) && Directory.Exists(extractionPath))
```
dsplaisted, line 109: It looks like a bunch of this cleanup logic was moved out of a finally block, should we be cleaning up in the case of unexpected errors?  Does this fit into the transactions at all?

It's probably fine as is, I'm just wondering.

dsplaisted (approved):

src/Cli/dotnet/NugetPackageInstaller/NuGetPackageDownloader.cs
```diff
@@ -72,7 +72,7 @@ public NuGetPackageDownloader(DirectoryPath packageInstallDir, IFilePermissionSe
 
             if (resource == null)
```
dsplaisted, line 73: Do the calls to `GetLatestVersionInternalAsync` and `GetPackageMetadataAsync` succeed if there is no package found?  I would have expected to need to check for an error there.

PR #17460, opened 2021-05-05
Update the conventions for discovering and invoking metadata update handlers

pranavkm:
* Introduce ClearCache and UpdateApplication action names
* Invoke actions in topologically sorted order.

Contributes to https://github.com/dotnet/runtime/issues/51545

### DIFF
--- a/src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
+++ b/src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
@@ -6,8 +6,6 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Reflection.Metadata;
-using System.Runtime.InteropServices;
 using Microsoft.DotNet.Watcher.Tools;
 
 namespace Microsoft.Extensions.HotReload
@@ -54,37 +52,37 @@ namespace Microsoft.Extensions.HotReload
 
         private UpdateHandlerActions GetMetadataUpdateHandlerActions()
         {
-            // In a typical app, a handful of assemblies will have metadata handler attributes. We need to discover these assemblies
-            // and ensure that they are topologically sorted so that handlers in a dependency are executed before the dependent (e.g.
-            // the reflection cache action in System.Private.CoreLib is executed before System.Text.Json clears it's own cache.)
-            // This would ensure that there is a well-defined order and that caches and updates more lower in the application stack
-            // are up to date before ones higher in the stack are recomputed.
-            var assemblies = GetAssembliesWithMetadataUpdateHandlerAttributes();
-            var sortedAssemblies = TopologicalSort(CollectionsMarshal.AsSpan(assemblies));
+            // We need to execute MetadataUpdateHandlers in a well-defined order. For v1, the strategy that is used is to topologically
+            // sort assemblies so that handlers in a dependency are executed before the dependent (e.g. the reflection cache action
+            // in System.Private.CoreLib is executed before System.Text.Json clears it's own cache.)
+            // This would ensure that caches and updates more lower in the application stack are up to date
+            // before ones higher in the stack are recomputed.
+            var sortedAssemblies = TopologicalSort(AppDomain.CurrentDomain.GetAssemblies());
             var handlerActions = new UpdateHandlerActions();
-            foreach (var assembly in CollectionsMarshal.AsSpan(sortedAssemblies))
+            foreach (var assembly in sortedAssemblies)
             {
-                foreach (var attribute in assembly.GetCustomAttributes<MetadataUpdateHandlerAttribute>())
+                foreach (var attr in assembly.GetCustomAttributesData())
                 {
-                    GetHandlerActions(handlerActions, attribute.HandlerType);
+                    // Look up the attribute by name rather than by type. This would allow netstandard targeting libraries to
+                    // define their own copy without having to cross-compile.
+                    if (attr.AttributeType.FullName != "System.Reflection.Metadata.MetadataUpdateHandlerAttribute")
+                    {
+                        continue;
+                    }
+
+                    IList<CustomAttributeTypedArgument> ctorArgs = attr.ConstructorArguments;
+                    if (ctorArgs.Count != 1 ||
+                        ctorArgs[0].Value is not Type handlerType)
+                    {
+                        _log($"'{attr}' found with invalid arguments.");
+                        continue;
+                    }
+
+                    GetHandlerActions(handlerActions, handlerType);
                 }
             }
 
             return handlerActions;
-
-            static List<Assembly> GetAssembliesWithMetadataUpdateHandlerAttributes()
-            {
-                var assemblies = new List<Assembly>();
-                foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
-                {
-                    if (assembly.IsDefined(typeof(MetadataUpdateHandlerAttribute)))
-                    {
-                        assemblies.Add(assembly);
-                    }
-                }
-
-                return assemblies;
-            }
         }
 
         internal void GetHandlerActions(UpdateHandlerActions handlerActions, Type handlerType)
@@ -160,7 +158,7 @@ namespace Microsoft.Extensions.HotReload
             }
         }
 
-        internal static List<Assembly> TopologicalSort(ReadOnlySpan<Assembly> assemblies)
+        internal static List<Assembly> TopologicalSort(Assembly[] assemblies)
         {
             var sortedAssemblies = new List<Assembly>(assemblies.Length);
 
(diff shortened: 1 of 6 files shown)


src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -45,116 +44,211 @@ private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs eventArgs)
...
+        private UpdateHandlerActions GetMetadataUpdateHandlerActions()
         {
-            public UpdateHandlerActions(List<Action<Type[]?>> before, List<Action<Type[]?>> after)
+            // In a typical app, a handful of assemblies will have metadata handler attributes. We need to discover these assemblies
+            // and ensure that they are topologically sorted so that handlers in a dependency are executed before the dependent (e.g.
+            // the reflection cache action in System.Private.CoreLib is executed before System.Text.Json clears it's own cache.)
+            // This would ensure that there is a well-defined order and that caches and updates more lower in the application stack
+            // are up to date before ones higher in the stack are recomputed.
+            var assemblies = GetAssembliesWithMetadataUpdateHandlerAttributes();
+            var sortedAssemblies = TopologicalSort(CollectionsMarshal.AsSpan(assemblies));
```
jkotas, line 63: There is no point in using `CollectionsMarshal.AsSpan` here. Change it to use a regular enumerator instead. 

The potentially dangerous APIs like ``CollectionsMarshal.AsSpan` are meant to be only used for core that needs to be micro-optimized to the last cycle. It is not the case here. You are calling GetCustomAttributes inside the loop that is like 1000x times more expensive than the foreach.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -45,116 +44,211 @@ private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs eventArgs)
...
+            var handlerActions = new UpdateHandlerActions();
+            foreach (var assembly in CollectionsMarshal.AsSpan(sortedAssemblies))
             {
-                Before = before;
-                After = after;
+                foreach (var attribute in assembly.GetCustomAttributes<MetadataUpdateHandlerAttribute>())
```
jkotas, line 67: We have said that we are going to match the attribute by name, so that netstandard targeting libraries can define their own copy, without cross-compilation: https://github.com/dotnet/runtime/issues/49361#issuecomment-795992329 . Is it still the plan?

This is matching the attribute by type identity, not by name.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -45,116 +44,211 @@ private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs eventArgs)
...
+            var visited = new HashSet<string>(StringComparer.Ordinal);
+
+            foreach (var assembly in assemblies)
+            {
+                Visit(assemblies, assembly, sortedAssemblies, visited);
+            }
+
+            static void Visit(ReadOnlySpan<Assembly> assemblies, Assembly assembly, List<Assembly> sortedAssemblies, HashSet<string> visited)
+            {
+                var assemblyIdentifier = assembly.FullName ?? assembly.ToString();
```
jkotas, line 176:
```suggestion
                var assemblyIdentifier = assembly.FullName!;
```
Runtime loaded assemblies always have non-empty full names. This is just nullability annotations not being helpful.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -45,116 +44,211 @@ private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs eventArgs)
...
+
+                sortedAssemblies.Add(assembly);
+            }
 
-                    MethodInfo? GetUpdateMethod(Type handlerType, string name)
+            static Assembly? FindDependency(ReadOnlySpan<Assembly> assemblies, AssemblyName dependencyName)
+            {
+                foreach (var assembly in assemblies)
+                {
+                    if (assembly.FullName == dependencyName.FullName)
```
jkotas, line 198: The dependencies should be matched using simple name. FullName includes version that will often not match.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -45,116 +44,211 @@ private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs eventArgs)
...
+
+            static void Visit(ReadOnlySpan<Assembly> assemblies, Assembly assembly, List<Assembly> sortedAssemblies, HashSet<string> visited)
+            {
+                var assemblyIdentifier = assembly.FullName ?? assembly.ToString();
+                if (!visited.Add(assemblyIdentifier))
+                {
+                    return;
+                }
+
+                foreach (var dependencyName in assembly.GetReferencedAssemblies())
```
jkotas, line 182: Is this going to handle situations like `A -> B -> C -> D`, A and D having the attribute, B and C do not have the attribute?

If I am reading the code correctly, FindDependency will return null for B and we end up with A and D not getting sorted.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -45,116 +44,211 @@ private void OnAssemblyLoad(object? _, AssemblyLoadEventArgs eventArgs)
...
+                _handlerActions ??= GetMetadataUpdateHandlerActions();
+                var handlerActions = _handlerActions;
+
+                // TODO: Get types to pass in
+                Type[]? updatedTypes = null;
 
-                beforeAfterUpdates.Before.ForEach(b => b(null)); // TODO: Get types to pass in
+                InvokeActions(handlerActions.Before, updatedTypes);
 
-                foreach (var item in deltas)
```
jkotas, line 138: What was wrong with `foreach` ?

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
pranavkm: Ahh, I might have missed that conversation. I'll switch it back - that's what this code was initially doing before I recently changed it .

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
pranavkm: I started with that, but I was worried because the nullability annotations suggested that the name could be null. Does it follow that if FullName is non-null for runtime loaded assemblies, `GetName().Name` is also non-null?

jkotas: Yes, `GetName().Name` is always going to be non-null for runtime loaded assemblies.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
pranavkm: Doesn't look like it. It incidentally worked for the few scenarios that we have now but it's clearly buggy. I'll fix it up.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
pranavkm: Do you have suggestions on how we would approach a graph like this:

```
[
    "System.Private.CoreLib": /* depends on */ [] ,
    "System.Text.Json": ["System.Text.PrivateCoreLib"],
    "Microsoft.AspNetCore.Components, Version=3.1.0.0": ["System.Text.PrivateCoreLib"],
    "Microsoft.AspNetCore.Components, Version=6.0.0.0": ["System.Text.PrivateCoreLib", "System.Text.Json"],
    "Microsoft.AspNetCore.Components.Web, Version=6.0.0.0": ["Microsoft.AspNetCore.Components, Version=6.0.0.0"],
]
```

Relying on simple name might result in `Microsoft.AspNetCore.Components.Web` to be sorted alongside `System.Text.Json`.

jkotas: You should not see both `Microsoft.AspNetCore.Components, Version=6.0.0.0` and `Microsoft.AspNetCore.Components, Version=3.1.0.0` at runtime. The 3.1 version will be unified with the 6.0 version. The runtime will only load  `Microsoft.AspNetCore.Components, Version=6.0.0.0` and it will be used to satisfy both 3.1 and 6.0 references. 

This is assuming default build and loader behavior. If somebody has a custom solutions using AssemblyLoadContexts and multiple versions of the same thing loaded SxS, it gets really complex. I do not think we need to worry about cases like that for hot reload.

src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
pranavkm: I had it updated to use a local and I was worried I would get complaints about capturing delegates. I reverted it - the original version had less code 👍🏽

pranavkm (commented):
This is updated. 

Also sent https://github.com/dotnet/runtime/pull/52409 to update the runtime to use the new names

jkotas (approved):
LGTM

Issue #3917, opened 2019-11-15
Respect `DO_NOT_TRACK` as standardized alternative to `DOTNET_CLI_TELEMETRY_OPTOUT`


khellang:
`DO_NOT_TRACK` is a proposed unified standard for opting out of telemetry for TUI/console apps.
It would be nice if the dotnet CLI followed the same standard.

See https://consoledonottrack.com/

fowl2:
It seems like this has been denied. https://github.com/dotnet/sdk/pull/17472#issuecomment-855087362

Extarys:
A year and a half later, any update on using this? I feel like it's a standard now on many OSS.

marcpopMSFT:
Coming back to this, we reviewed the original effort and it doesn't appear to have been picked up broadly so we're going to continue to hold off for now.

coopbri:
I respect the concern for the lack of widespread usage. With that being said, adoption by an entity such as Microsoft would likely be a monumental step towards such public usage, and it seems to me like this would not only be a straightforward implementation in the .NET project, but a well-received publicity and marketing measure for .NET and Microsoft as a whole. Considering there are 125 upvotes at the time of writing on the OP in this issue, and the lack of other well-known console opt-out standards (please correct me if I'm wrong), I am curious what is desired by the .NET team to move this budding community effort into the SDK. I believe as a general technology community (beyond .NET), we should combine efforts to hone in on standards such as what was done in hardware with e.g. USB. Despite the claim that there is no broad adoption, there certainly is a meaningful amount of it. Perhaps there is simply a lack of marketing and awareness; I argue that the idea itself has merit.

I would be curious to hear concrete adoption hesitations. An idea is supporting both the existing opt-out strategy (bespoke env var), as well as the standard one and simply checking for either.

khellang:
> Considering there are 125 upvotes at the time of writing on the OP in this issue

It's the #2 most upvoted issue in this repo 😅

cos4ni2s:
I would be very, very surprised if a company such as Microsoft aligns with consumer initiatives—even as *light* as this one in their open-source projects.

coopbri:
> I would be very, very surprised if a company such as Microsoft aligns with consumer initiatives—even as _light_ as this one in their open-source projects.

Interesting and imo misaligned with reality, considering they have been extremely responsive to consumers in FOSS projects such as VS Code and TypeScript (I don't use VS Code anymore btw, just credit where credit is due)

[...]

PR #17472, opened 2021-05-06
Honor DO_NOT_TRACK as well as DOTNET_CLI_TELEMETRY_OPTOUT

fowl2:
An attempt at #3917

### DIFF
--- a/src/Cli/Microsoft.DotNet.Configurer/LocalizableStrings.resx
+++ b/src/Cli/Microsoft.DotNet.Configurer/LocalizableStrings.resx
@@ -128,7 +128,7 @@ SDK Version: {1}</value>
   <data name="TelemetryMessage" xml:space="preserve">
     <value>Telemetry
 ---------
-The .NET tools collect usage data in order to help us improve your experience. It is collected by Microsoft and shared with the community. You can opt-out of telemetry by setting the DOTNET_CLI_TELEMETRY_OPTOUT environment variable to '1' or 'true' using your favorite shell.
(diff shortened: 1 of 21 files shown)


Diti:
Let that be a lesson to the open-source communities of Lobsters, Hacker News, Reddit, and StackExchange. Microsoft does not care about standardization attempts started outside of their company.

delanym:
[Not invented here](https://en.wikipedia.org/wiki/Not_invented_here)

Diti:
[xkcd: Standards](https://xkcd.com/927/)

delanym:
The variable DOTNET_CLI_TELEMETRY_OPTOUT is not a standard

kasperk81:
DO_NOT_TRACK is not the standard either

summing up https://github.com/dotnet/sdk/issues/6145 to reflect what community at large want:

1. the best thing would be if tracking is made opt-in.
2. the second best option is if user is _asked_ on the first run whether to opt-in to this data collection program.
3. the third best option is to provide a way to opt-out, for which there is existing support for DOTNET_CLI_TELEMETRY_OPTOUT flag.

DO_NOT_TRACK is just the variant of third best thing, and it is not that interesting as first two options, which will _really_ inspire confidence transparency.

delanym:
Tracking will never be made opt-in because who in their right mind would go out of their way to turn it on? Its unrealistic.
Your second best option is not friendly for unattended installations. It also means that if I remove the application and install it again I have to answer the same question again.
I make the decision to allow telemetry on a system level, not on an application level. If I say yes to telemetry for dotnet, chances are I'm happy to do it for others and I don't want to have to answer the same dumb question over again for every machine, for every app, for every install. It also means if I change my mind, I can change it once! DO_NOT_TRACK is my system level telemetry policy and it costs nothing for dotnet to respect it - probably not even another line of code.

khellang:
Hey @marcpopMSFT! 👋🏻 

Would it hurt anyone to honor a standard flag *in addition* to the existing flag? I'd love to hear some reasoning for this 😅

PR #17489, opened 2021-05-06
Better error strings, null check, Add some more test, use sdk default logger

Anipik:
- better error strings
- null check
- display color 
- avoid double checks if package has no refs.


src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
```diff
@@ -136,21 +136,27 @@
...
     <value>There is no compatible runtime asset for target framework {0} in the package.</value>
   </data>
   <data name="BaselineVersionValidatorHeader" xml:space="preserve">
-    <value>There should not be any breaking changes between the packages.</value>
+    <value>There are breaking changes between the versions. Please add or modify the apis in the recent version or suppress the intentional breaking changes.</value>
   </data>
-  <data name="TargetFramework" xml:space="preserve">
-    <value>TargetFramework {0}</value>
+  <data name="ApiCompatibilityBaselineHeader" xml:space="preserve">
+    <value>Api compatibility errors in between {0} (left) and {1} (right) for versions {2} and {3} respectively are as follows:</value>
```
safern, line 151:
```suggestion
    <value>Found compatibility errors in between {0} (left) and {1} (right) for versions {2} and {3} respectively:</value>
```

safern: Or `There are`?

src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
```diff
@@ -118,16 +118,16 @@
...
   <data name="CompatibleTfmValidatorHeader" xml:space="preserve">
-    <value>The runtime assets should be compatible with the compile time assets for all target frameworks and rids.</value>
+    <value>The compile time assemblies public api surface area should be compatible with the runtime assemblies for all target frameworks and rids. Please add the following apis to the runtime assemblies.</value>
   </data>
   <data name="InvalidPackagePath" xml:space="preserve">
     <value>Package Path {0} is invalid. Please Check the path .</value>
   </data>
   <data name="ApiCompatibilityHeader" xml:space="preserve">
-    <value>There are API Compatibility errors in between {0} (left) and {1} (right).</value>
+    <value>Api Compatibility errors  between {0} (left) and {1} (right) are as follows:</value>
```
safern, line 130:
```suggestion
    <value>API Compatibility errors between {0} (left) and {1} (right):</value>
```

src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
```diff
@@ -136,21 +136,27 @@
...
-    <value>TargetFramework {0} Rid {1}</value>
+  <data name="BaselinePackageNotFound" xml:space="preserve">
+    <value>No baseline package with id {0} found in feeds:  {1}.</value>
   </data>
-</root>
+  <data name="EmptyPackagePath" xml:space="preserve">
+    <value>{0} is empty or null.  Please check the package path.</value>
+  </data>
+  <data name="NonExistentPackagePath" xml:space="preserve">
+    <value>{0} doesnt exist. Please check the package path.</value>
```
safern, line 160:
```suggestion
    <value>{0} does not exist. Please provide a valid package path.</value>
```

src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
```diff
@@ -136,21 +136,27 @@
...
+    <value>Api compatibility errors in between {0} (left) and {1} (right) for versions {2} and {3} respectively are as follows:</value>
   </data>
-  <data name="TargetFrameworkRid" xml:space="preserve">
-    <value>TargetFramework {0} Rid {1}</value>
+  <data name="BaselinePackageNotFound" xml:space="preserve">
+    <value>No baseline package with id {0} found in feeds:  {1}.</value>
   </data>
-</root>
+  <data name="EmptyPackagePath" xml:space="preserve">
+    <value>{0} is empty or null.  Please check the package path.</value>
```
safern, line 157:
```suggestion
    <value>{0} is empty or null.  Please provide a valid package path.</value>
```

src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
```diff
@@ -136,21 +136,27 @@
...
   </data>
-  <data name="TargetFramework" xml:space="preserve">
-    <value>TargetFramework {0}</value>
+  <data name="ApiCompatibilityBaselineHeader" xml:space="preserve">
+    <value>Api compatibility errors in between {0} (left) and {1} (right) for versions {2} and {3} respectively are as follows:</value>
   </data>
-  <data name="TargetFrameworkRid" xml:space="preserve">
-    <value>TargetFramework {0} Rid {1}</value>
+  <data name="BaselinePackageNotFound" xml:space="preserve">
+    <value>No baseline package with id {0} found in feeds:  {1}.</value>
```
safern, line 154:
```suggestion
    <value>No baseline package with id {0} found in feeds: {1}.</value>
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -35,9 +35,7 @@ public void RunApiCompat()
         {
             foreach (var apicompatTuples in _queue.Distinct())
             {
-                // TODO: Add Assembly version check.
```
safern, line 38: I don't see the assembly check, are you removing this cause you opened the issues I guess?

src/Compatibility/Microsoft.DotNet.PackageValidation/PackageValidationLogger.cs
```diff
@@ -9,81 +9,21 @@ namespace Microsoft.DotNet.PackageValidation
...
+        public void Log(ILogMessage message) => throw new NotImplementedException();
+        public Task LogAsync(LogLevel level, string data) => throw new NotImplementedException();
+        public Task LogAsync(ILogMessage message) => throw new NotImplementedException();
+        public void LogDebug(string data) => throw new NotImplementedException();
+        public void LogInformation(string data) => throw new NotImplementedException();
+        public void LogInformationSummary(string data) => throw new NotImplementedException();
+        public void LogMinimal(string data) => throw new NotImplementedException();
+        public void LogVerbose(string data) => throw new NotImplementedException();
+        public void LogWarning(string data) => throw new NotImplementedException();
         public void LogError(string data)
```
safern, line 22: Curious question, why not use the MSBuild logger that already handles coloring and all that?

src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
ericstj: I don't think we should be exposing the terms "left" and "right" to end users.  Instead we should be giving them descriptive identifiers, such as:
compile asset, runtime asset
framework X asset, compatible framework Y asset
etc

safern: Yes I totally agree with you. This is temporary while I add support for API Compat to have custom naming for `left` and `right`. However, since at the moment it is hardcoded to `left` and `right` in the API Compat diagnostic messages, it made sense to add it here so that it is clearer what's `left` and what's `right`.

ericstj: I see, makes sense.

safern: I created: https://github.com/dotnet/sdk/issues/17495 to track this.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
Anipik: yep https://github.com/dotnet/sdk/issues/17507

src/Compatibility/Microsoft.DotNet.PackageValidation/PackageValidationLogger.cs
Anipik: its an internal class that lives in the arcade. All the loggers in the sdk repo are not that appropiate. eg the errors need to start with ".net sdk error"

safern: As noted offline we should use an MSBuild logger rather than logging to the console ourselves.

src/Compatibility/Microsoft.DotNet.PackageValidation/GetLastStablePackage.cs
```diff
@@ -37,7 +37,7 @@ protected override void ExecuteCore()
                 SourceRepository repository = Repository.Factory.GetCoreV3(nugetFeed);
                 FindPackageByIdResource resource = repository.GetResource<FindPackageByIdResource>();
                 SourceCacheContext cache = new SourceCacheContext();
-                IEnumerable<NuGetVersion> versions = resource.GetAllVersionsAsync(PackageId, cache, NullLogger.Instance, CancellationToken.None).Result;
+                IEnumerable<NuGetVersion> versions = resource?.GetAllVersionsAsync(PackageId, cache, NullLogger.Instance, CancellationToken.None).Result;
```
ViktorHofer, line 40: If resource is null, will any of logic below actually do anything? If not, should you just stop execution at this point instead?

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
```diff
@@ -16,17 +16,16 @@ namespace Microsoft.DotNet.PackageValidation
     /// Validates that no target framework / rid support is dropped in the latest package.
     /// Reports all the breaking changes in the latest package.
     /// </summary>
-    public class BaselinePackageValidator
+    internal class BaselinePackageValidator
```
safern, line 19: If we wanted these object models to be used to build an App or a tool on top of them we would need them to be public.

What if we leave the object models to receive an `ILogger` and then you could add a class that is internal and implements `ILogger` and then `Log.LogError` redirects to `Log.LogNonSdkError`? 

Then if we use these object models for a dotnet tool or another app on top of them to validate packages, those apps can implement their own `ILogger` classes and log to the console or somewhere else.

src/Compatibility/Microsoft.DotNet.PackageValidation/CompatibleTFMValidator.cs
```diff
@@ -22,9 +21,9 @@ public class CompatibleTfmValidator
         private static Dictionary<NuGetFramework, HashSet<NuGetFramework>> s_packageTfmMapping = InitializeTfmMappings();
 
         private readonly bool _runApiCompat;
-        private readonly ApiCompatRunner _apiCompatRunner;
-        private readonly ILogger _log;
         private readonly DiagnosticBag<IDiagnostic> _diagnosticBag;
+        private ApiCompatRunner _apiCompatRunner;
+        private ILogger _log;
```
safern, line 26: why can't these be readonly anymore?

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -6,7 +6,7 @@
     <IsShipping>true</IsShipping>
     <StrongNameKeyId>Open</StrongNameKeyId>
     <PackageType>MSBuildSdk</PackageType>
-    <VersionPrefix>0.1.0</VersionPrefix>
+    <VersionPrefix>1.0.0</VersionPrefix>
```
safern, line 9: why are we changing to 1.0.0?

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -32,17 +31,31 @@ protected override void ExecuteCore()
...
-                new BaselinePackageValidator(baselinePackage, NoWarn, null, RunApiCompat, log).Validate(package);
+                new BaselinePackageValidator(baselinePackage, NoWarn, null, RunApiCompat, logger).Validate(package);
             }
-            
-            new CompatibleFrameworkInPackageValidator(NoWarn, null, log).Validate(package);
+        }
+    }
+
+
+    internal class PackageValidationLogger : ILogger
```
safern, line 47: Move to it's own file?

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
Anipik: dont we want to ship the first package as 1.0.0-preview4 ?

safern: Hmm... yeah I don't have an strong opinion.

src/Compatibility/Microsoft.DotNet.PackageValidation/GetLastStablePackage.cs
Anipik: resource is not null, i will remove it.

Anipik:
@safern @ericstj @ViktorHofer can you take a look on this one ?

src/Compatibility/Microsoft.DotNet.PackageValidation/ILogger.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.PackageValidation
+{
+    public interface ILogger
```
safern, line 6: NIT: How do you feel about using the same name as the abstraction MSBuild provides for this? I think it is not a big deal as this is in a different namespace, but I don't know if it might cause any problems in the future?

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -3,10 +3,10 @@
   <PropertyGroup>
     <TargetFrameworks>$(SdkTargetFramework);net472</TargetFrameworks>
     <IsPackable>true</IsPackable>
-    <IsShipping>true</IsShipping>
+    <IsShippingPackage>true</IsShippingPackage>
```
safern, line 6: why changed it to use `IsShippingPackage` instead?

src/Compatibility/Microsoft.DotNet.PackageValidation/PackageValidationLogger.cs
```diff
@@ -1,89 +1,23 @@
...
-        public void LogInformationSummary(string data)
-        {
-            Console.WriteLine($"[Info Summary] {data}");
-        }
-
-        public void LogMinimal(string data)
-        {
-            Console.WriteLine($"{data}");
-        }
+        private Logger _log;
```
safern, line 11:
```suggestion
        private readonly Logger _log;
```

src/Compatibility/Microsoft.DotNet.PackageValidation/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.DotNet.PackageValidation.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
safern, line 6: Do you still need this?

src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx
```diff
@@ -118,16 +118,16 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="CompatibleFrameworkInPackageValidatorHeader" xml:space="preserve">
-    <value>Api surface area should be compatible among compatible frameworks.</value>
+    <value>{0} assembly api surface area should be compatible with {1} assembly surface area so we can compile against {0} and run on {1} .framework.</value>
   </data>
   <data name="CompatibleTfmValidatorHeader" xml:space="preserve">
-    <value>The runtime assets should be compatible with the compile time assets for all target frameworks and rids.</value>
+    <value>The compile time assemblies public api surface area should be compatible with the runtime assemblies for all target frameworks and rids. Please add the following apis to the runtime assemblies.</value>
```
safern, line 124:
```suggestion
    <value>The compile time assemblies public api surface area should be compatible with the runtime assemblies for all target frameworks and RIDs. Please add the following apis to the runtime assemblies.</value>
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -32,17 +31,15 @@ protected override void ExecuteCore()
             }
 
             Package package = NupkgParser.CreatePackage(PackageTargetPath, runtimeGraph);
-            PackageValidationLogger log = new();
-
-            new CompatibleTfmValidator(NoWarn, null, RunApiCompat, log).Validate(package);
+            PackageValidationLogger logger = new(Log);
 
+            new CompatibleTfmValidator(NoWarn, null, RunApiCompat, logger).Validate(package);
+            new CompatibleFrameworkInPackageValidator(NoWarn, null, logger).Validate(package);
```
safern, line 37: It seems like a weird design to have objects that are instantiated but not cached and not used after it's creation. Should these instead be static methods?

src/Tests/Microsoft.DotNet.PackageValidation.Tests/BaseLineVersionValidatorTests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.DotNet.PackageValidation.Tests
+{
+    public class BaseLineVersionValidatorTests : SdkTest
+    {
+        private TestLogger _log;
+
+        public BaseLineVersionValidatorTests(ITestOutputHelper log) : base(log)
+        {
+            _log = new();
```
safern, line 16: use `new TestLogger()` instead as the type declaration is on another line.

src/Tests/Microsoft.DotNet.PackageValidation.Tests/CompatibleFrameworkValidatorTests.cs
```diff
@@ -0,0 +1,190 @@
...
+
+namespace Microsoft.DotNet.PackageValidation.Tests
+{
+    public class CompatibleFrameworkValidatorTests : SdkTest
+    {
+        private TestLogger _log;
+
+        public CompatibleFrameworkValidatorTests(ITestOutputHelper log) : base(log)
+        {
+            _log = new();
```
safern, line 16: Ditto.

ericstj (commented):
Just a couple minor things.

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
Anipik: IsShipping is not working as IsShippingPackage is set to false for all the repo. we need to set this true in order to ship it as a shippping asset.

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -46,6 +46,7 @@
 
   <PropertyGroup>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);_AddBuildOutputToPackageCore;_AddBuildOutputToPackageDesktop</TargetsForTfmSpecificContentInPackage>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
```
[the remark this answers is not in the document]
Anipik: i added the properties folder to add the assemblyInfo file to expose the internalsVisbleTo with the public key.
But its no longer required as we made the apis public

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
safern: Hmm, then I need to add it to the Compatibility project as well.

src/Compatibility/Microsoft.DotNet.PackageValidation/PackageValidationLogger.cs
```diff
@@ -1,89 +1,23 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
safern, line 3: Super nit. 
```suggestion
```

safern (approved):

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
Anipik: yes

PR #17537, opened 2021-05-10
Reuse binaries from compiler directory to reduce SDK size

captainsafia:
Part of https://github.com/dotnet/aspnetcore/issues/31212.

Size of `~/github.com/dotnet/sdk/artifacts/bin/redist/Debug/dotnet/sdk/6.0.100-dev ` on this branch: 176MB
Size of `~/github.com/dotnet/sdk/artifacts/bin/redist/Debug/dotnet/sdk/6.0.100-dev ` on this main: 184MB

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Program.cs
+++ b/src/BuiltInTools/dotnet-watch/Program.cs
@@ -79,7 +79,7 @@ Examples:
             _reporter = CreateReporter(verbose: true, quiet: false, console: _console);
 
             // Register listeners that load Roslyn-related assemblies from the `Rosyln/bincore` directory.
-            RegisterAssemblyResolutionEvents();
+            RegisterAssemblyResolutionEvents(sdkRootDirectory);
         }
 
         public static async Task<int> Main(string[] args)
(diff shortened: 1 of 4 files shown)


src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -378,5 +382,20 @@ public void Dispose()
...
         }
+
+        private static void RegisterAssemblyResolutionEvents()
+        {
+            var roslynPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "..", "..", "..", "..", "..", "..", "Roslyn", "bincore");
+
+            AssemblyLoadContext.Default.Resolving += (context, assembly) =>
+            {
+                var assemblyFilename = $"{assembly.Name}.dll";
+                if (assemblyFilename == "Microsoft.CodeAnalysis.CSharp.dll" || assemblyFilename == "Microsoft.CodeAnalysis.dll")
```
pranavkm, line 393: You could do the concat after you've checked the assembly name. Something like:

```C#
if (assembly.Name is "Microsoft.CodeAnalysis" or "Microsoft.CodeAnalysis.CSharp")
{
   return context.LoadFromAssemblyPath(Path.Combine(roslynPath, assemblyFilename + ".dll");
}
```

Avoids checking strings that you know would match.

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -378,5 +382,20 @@ public void Dispose()
             _console.CancelKeyPress -= OnCancelKeyPress;
             _cts.Dispose();
         }
+
+        private static void RegisterAssemblyResolutionEvents()
+        {
+            var roslynPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "..", "..", "..", "..", "..", "..", "Roslyn", "bincore");
```
pranavkm, line 388: Could we re-use the SDK root that we calculated earlier?

src/RazorSdk/Tool/Program.cs
```diff
@@ -49,5 +60,20 @@ public static int Main(string[] args)
 
             return result;
         }
+
+        private static void RegisterAssemblyResolutionEvents()
+        {
+            var roslynPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "..", "..", "..", "Roslyn", "bincore");
```
pranavkm, line 66: `AppContext.BaseDirectory` instead of `AppDomain.CurrentDomain.BaseDirectory`

pranavkm (commented):
Cool beans. Could you make sure hot reload works after this change (super manual testing)

src/RazorSdk/Tool/Program.cs
captainsafia: Is there a particular reason to prefer one over the other?

pranavkm: AppDomain is considered the legacy feature and is mostly shimmed in .NET Core. AppContext is the preferred replacement.

pranavkm (approved):

PR #17540, opened 2021-05-11
Refactor how composite image creation works

davidwrighton:
- Move it to allow mixed composite and non-composite r2r in the same build, controlled by the PublishReadyToRunCompositeExclusions list
- Enhance the accuracy of how msbuild tracks which files are input vs reference for composite images
- Implement the R2R exclusion list for composite images
- Remove the always on usage of --inputbubble when compiling a composite image

Refactor crossgen2 targets so that all targets specifically associated with Crossgen2 are located in Microsoft.NET.CrossGen.targets
- Add a Crossgen2TasksOverriden flag to indicate that the tasks are overriden and the UsingTask feature should not bring the normal ones into msbuild
- This is all done to better facilitate carrying a completely independent replacement copy in the runtime repo for use during feature development and dogfooding

### DIFF
--- a/src/Tasks/Common/MetadataKeys.cs
+++ b/src/Tasks/Common/MetadataKeys.cs
@@ -1,5 +1,5 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
 
 namespace Microsoft.NET.Build.Tasks
 {
@@ -25,7 +25,7 @@ namespace Microsoft.NET.Build.Tasks
         public const string TargetFramework = "TargetFramework";
         public const string FrameworkName = "FrameworkName";
         public const string FrameworkVersion = "FrameworkVersion";
-        public const string IsTrimmable = "IsTrimmable"; 
+        public const string IsTrimmable = "IsTrimmable";
         public const string RuntimeFrameworkName = "RuntimeFrameworkName";
         public const string RuntimePackRuntimeIdentifiers = "RuntimePackRuntimeIdentifiers";
 
@@ -123,5 +123,6 @@ namespace Microsoft.NET.Build.Tasks
         public const string OutputPDBImage = "OutputPDBImage";
         public const string EmitSymbols = "EmitSymbols";
         public const string IsVersion5 = "IsVersion5";
+        public const string CreateCompositeImage = "CreateCompositeImage";
     }
 }
--- a/src/Tasks/Common/MetadataKeys.cs
+++ b/src/Tasks/Common/MetadataKeys.cs
@@ -1,5 +1,5 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 namespace Microsoft.NET.Build.Tasks
 {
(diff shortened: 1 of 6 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
```diff
@@ -248,16 +214,131 @@ protected override void ExecuteCore()
...
+                if (hideFromCompositeBuilds)
+                    return new Eligibility(EligibilityEnum.Reference | EligibilityEnum.HideReferenceFromComposite);
+                else
+                    return new Eligibility(EligibilityEnum.Reference);
+            }
+
+            public static Eligibility CreateCompileEligibility(bool doNotBuildIntoComposite)
+            {
+                if (doNotBuildIntoComposite)
+                    return new Eligibility(EligibilityEnum.Reference | EligibilityEnum.HideReferenceFromComposite | EligibilityEnum.CompileSeparately);
```
trylek, line 316: All of this looks great, I'm just somewhat confused what is the difference between <code>EligibilityEnum.Reference | EligibilityEnum.HideReferenceFromComposite</code> and <code>EligibilityEnum.Reference | EligibilityEnum.HideReferenceFromComposite | EligibilityEnum.CompileSeparately</code>, can you please elaborate?

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
```diff
@@ -248,16 +214,131 @@ protected override void ExecuteCore()
...
+        {
+            if (peReader.PEHeaders == null)
+                return false;
+
+            if (peReader.PEHeaders.CorHeader == null)
+                return false;
+
+            if ((peReader.PEHeaders.CorHeader.Flags & CorFlags.ILLibrary) == 0)
+            {
+                // This is likely a composite image, but those can't be re-r2r'd
```
trylek, line 332: I don't see how we could arrive here for a composite image as it has no CorHeader, can you please clarify?

trylek (approved):
LGTM overall. I am a bit confused about some of the details and I must admit I haven't spent a ton of time eye-debugging the SDK msbuild script changes as that's just non-productive, there's no way to see the scripts work until we see them actually work.

src/Tasks/Common/MetadataKeys.cs
```diff
@@ -1,5 +1,5 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
AntonLapounov, line 2: Do we have to change the copyright text for these files?

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
davidwrighton: The first will represent a dll which is referenced only by separate R2R compilation, and the second is one which is actually compiled separately into a unique dll via crossgen2 or crossgen. Effectively, no code will be included from the assembly into a compiled R2R image unless one of the `Compile` flags is set.

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
davidwrighton: Actually, it turns out that composite images DO have a CorHeader. In fact, this code is a copy from r2rdump, but with the special handling to find the `RTR_HEADER` export removed.

AntonLapounov (approved):
Thank you!

AntonLapounov:
We may want to undo copyright text changes.

PR #17586, opened 2021-05-12
Revert "Reuse binaries from compiler directory to reduce SDK size"

captainsafia:
Reverts dotnet/sdk#17537.

This is causing the insertion into the installer to break due to incompatible assembly versions. The `rzc` tool has an assembly reference to 3.8.0 but the built in compiler contains a binary with a Version=4.0.0.

This causes the Crossgen targets in the installer to break because they resolve what assemblies to find based on the references in the `rzc.dll`.

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Program.cs
+++ b/src/BuiltInTools/dotnet-watch/Program.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Runtime.Loader;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Invocation;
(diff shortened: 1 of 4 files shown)


src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -382,19 +378,5 @@ public void Dispose()
             _console.CancelKeyPress -= OnCancelKeyPress;
             _cts.Dispose();
         }
-
-        private static void RegisterAssemblyResolutionEvents(string sdkRootDirectory)
```
pranavkm, line 386: Could we leave the dotnet-watch changes as-is. AFAIK, it's not being cross-gened today.

pranavkm (approved):

src/BuiltInTools/dotnet-watch/Program.cs
captainsafia: Sure -- done.

PR #15134, opened 2020-12-31
Generate errors when referencing an Executable project with different SelfContained value

dsplaisted:
Fixes #15117

Requires changes to MSBuild in dotnet/msbuild#5994 before this will work.

PR #17594, opened 2021-05-13
Fix test projects which reference self-contained Exes

### LINKED ISSUE #17579: "A self-contained executable cannot be referenced by a non self-contained executable" error on unit tests
I'm getting the error:
> The referenced project '{project}' is a self-contained executable.  A self-contained executable cannot be referenced by a non self-contained executable.

In my scenario, there's a project which is indeed self-contained, and a unit test project which depends on that project and tests it.

This error seems to imply that self-contained projects should not have unit tests, which can't be the case.

An easy workaround is to set `<ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>`, but it seems like the validation is incorrect in this scenario.

For reference, the self-contained project looks something like:

```xml
  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <DeployOnBuild>true</DeployOnBuild>
    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
    <PublishSingleFile>true</PublishSingleFile>
  </PropertyGroup>
</Project>
```

and the unit test project:

```xml
  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="..\TheSelfContainedProject\TheSelfContainedProject.csproj" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.3" />
    <PackageReference Include="MSTest.TestAdapter" Version="2.1.2" />
    <PackageReference Include="MSTest.TestFramework" Version="2.1.2" />
  </ItemGroup>
</Project>
```

dsplaisted:
Fix #17579

## Description
In #15134, we added an error when a self-contained app references a non self-contained app, or vice versa.  This is because if you are trying to get copies of both apps in the output folder, they won't both work.

However, I didn't consider that test projects are also use the `Exe` OutputType.  So with that change, test projects which reference Exe projects which were self-contained would fail to build.

## Customer Impact
Test projects referencing Exe projects (fairly common, especially for Web projects) which were self-contained (not as common, but enough that at least two projects hit this bug in the preview) would fail to build.  This could be worked around by setting the `ValidateExecutableReferencesMatchSelfContained` property to false, but developers would have to look through our MSBuild code or search for help on the internet to figure out they needed to do that.

## Regression?
- [x] Yes
- [ ] No

Regressed from 5.0.2xx and prior behavior in #15134.

## Risk
- [ ] High
- [ ] Medium
- [x] Low

[Justify the selection above]

The change is to disable the error check for unit test projects, and is a simple code change.

## Verification
- [ ] Manual (required)
- [x] Automated

Added test cases covering failing scenario

- [ ] Yes
- [ ] No
- [x] N/A

Addresses [issue number]
https://github.com/dotnet/sdk/issues/17579

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1060,6 +1060,13 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ShouldBeValidatedAsExecutableReference>false</ShouldBeValidatedAsExecutableReference>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsTestProject)' == 'true' And '$(ValidateExecutableReferencesMatchSelfContained)' == ''">
+    <!-- Don't generate an error if a test project references a self-contained Exe.  Test projects
+         use an OutputType of Exe but will usually call APIs in a referenced Exe rather than try
+         to run it. -->
+    <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
+  </PropertyGroup>
+
   <UsingTask TaskName="ValidateExecutableReferences" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
   <Target Name="ValidateExecutableReferences"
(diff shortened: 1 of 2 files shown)


sfoslund:
@dsplaisted I was going to try to prep a build for insertion today, do we need to wait for this PR to do that?

sfoslund (approved):

dsplaisted:
@sfoslund We're going to take this to tactics today, so yes hold off until after that.  Thanks!

PR #17607, opened 2021-05-13
Add IsTrimmable property

### LINKED ISSUE #17201: Add `IsTrimmable` property for libraries
Two things we want to get out of this:

- [ ] Enable feature switch defaults
- [ ] Define `IsTrimmable` attribute in generated sources

sbomer:
This property will add IsTrimmable assembly metadata and enable the trim
analyzer. This doesn't set feature switches since libraries should not
in general be built with the assumption that feature settings have known
values.

Also adds a test that PublishTrimmed has the desired effect during build,
and doesn't block building without a RID. PublishTrimmed sets feature
switches and enables trim analysis but doesn't inject the assembly
attribute.

Fixes https://github.com/dotnet/sdk/issues/17201

### DIFF
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
@@ -683,16 +683,16 @@ namespace Microsoft.NET.Publish.Tests
                 JObject runtimeConfig = JObject.Parse(runtimeConfigContents);
                 runtimeConfig["runtimeOptions"]["configProperties"]
                     ["System.Runtime.InteropServices.Marshal.IsBuiltInComSupported"].Value<bool>()
-                    .Should().Be(false);
+                    .Should().BeFalse();
                 runtimeConfig["runtimeOptions"]["configProperties"]
                     ["System.StartupHookProvider.IsSupported"].Value<bool>()
-                    .Should().Be(false);
+                    .Should().BeFalse();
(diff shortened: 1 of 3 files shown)


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -1251,6 +1251,65 @@ class Program
...
+
+            var buildCommand = new BuildCommand(testAsset);
+            // IsTrimmable enables analysis warnings during build
+            buildCommand.Execute()
+                .Should().Pass()
+                .And.HaveStdOutMatching("warning IL2026.*Program.IL_2026.*Testing analysis warning IL2026");
+
+            var outputDirectory = buildCommand.GetOutputDirectory(targetFramework).FullName;
+            var assemblyPath = Path.Combine(outputDirectory, $"{projectName}.dll");
+            var runtimeConfigPath = Path.Combine(outputDirectory, $"{projectName}.runtimeconfig.json");
```
vitek-karas, line 1272: This will build an "app" test project, right?
In that case we should add another test case which builds "classlib" - as that is the primary target for this feature. (We won't be able to rely on runtimeconfig though).

sbomer: I added a classlib test that just doesn't check the feature settings (since they don't affect libraries)

vitek-karas (approved):
LGTM

PR #17608, opened 2021-05-13
Fix a couple of issues with dotnet-watch

pranavkm:
* Print compilation errors in red and don't say hot reload succeeded when there's errors
* Fix an issue where the browser does not refresh after changes
* Fix an issue where deltas are not applied on assemblies loaded after the fact.
* Fix an issue where hot reload no longer works after an app restart.
* Fix a local dev issue to avoid using the local SDK when running dotnet-watch as an executable

### DIFF
--- a/src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
+++ b/src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Extensions.HotReload
     {
         private readonly Action<string> _log;
         private readonly AssemblyLoadEventHandler _assemblyLoad;
-        private readonly ConcurrentDictionary<Guid, IReadOnlyList<UpdateDelta>> _deltas = new();
+        private readonly ConcurrentDictionary<Guid, List<UpdateDelta>> _deltas = new();
         private readonly ConcurrentDictionary<Assembly, Assembly> _appliedAssemblies = new();
         private volatile UpdateHandlerActions? _handlerActions;
 
@@ -196,6 +196,10 @@ namespace Microsoft.Extensions.HotReload
                     {
                         System.Reflection.Metadata.AssemblyExtensions.ApplyUpdate(assembly, item.MetadataDelta, item.ILDelta, ReadOnlySpan<byte>.Empty);
                     }
+
+                    // Additionally stash the deltas away so it may be applied to assemblies loaded later.
+                    var cachedDeltas = _deltas.GetOrAdd(item.ModuleId, static _ => new());
+                    cachedDeltas.Add(item);
                 }
 
                 handlerActions.ClearCache.ForEach(a => a(updatedTypes));
--- a/src/BuiltInTools/DotNetDeltaApplier/StartupHook.cs
+++ b/src/BuiltInTools/DotNetDeltaApplier/StartupHook.cs
@@ -53,11 +53,6 @@ internal sealed class StartupHook
             Log("Attempting to apply deltas.");
 
             hotReloadAgent.ApplyDeltas(update.Deltas);
-
-            // We want to base this off of mvids, but we'll figure that out eventually.
-            var applyResult = update.ChangedFile is string changedFile && changedFile.EndsWith(".razor", StringComparison.Ordinal) ?
-                ApplyResult.Success :
-                ApplyResult.Success_RefreshBrowser;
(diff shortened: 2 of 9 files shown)


src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -196,6 +196,10 @@ public void ApplyDeltas(IReadOnlyList<UpdateDelta> deltas)
                     {
                         System.Reflection.Metadata.AssemblyExtensions.ApplyUpdate(assembly, item.MetadataDelta, item.ILDelta, ReadOnlySpan<byte>.Empty);
                     }
+
+                    // Additionally stash the deltas away so it may be applied to assemblies loaded later.
+                    var cachedDeltas = _deltas.GetOrAdd(item.ModuleId, static _ => new());
```
pranavkm, line 201: We were applying these deltas, but I missed storing them.

src/BuiltInTools/DotNetDeltaApplier/StartupHook.cs
```diff
@@ -53,11 +53,6 @@ public static async Task ReceiveDeltas(HotReloadAgent hotReloadAgent)
             Log("Attempting to apply deltas.");
 
             hotReloadAgent.ApplyDeltas(update.Deltas);
-
-            // We want to base this off of mvids, but we'll figure that out eventually.
-            var applyResult = update.ChangedFile is string changedFile && changedFile.EndsWith(".razor", StringComparison.Ordinal) ?
```
pranavkm, line 58: I changed the heuristic about when we refresh the browser in web apps - we won't it if it looks like a Blazor app (presence of any .razor file). Works better than relying on individual files.

src/BuiltInTools/dotnet-watch/HotReload/CompilationHandler.cs
```diff
@@ -171,7 +179,7 @@ private ImmutableArray<string> GetDiagnostics(Solution solution, CancellationTok
                     if (item.Severity == DiagnosticSeverity.Error)
                     {
                         var diagnostic = CSharpDiagnosticFormatter.Instance.Format(item);
-                        _reporter.Output(diagnostic);
+                        _reporter.Output("\x1B[40m\x1B[31m" + diagnostic);
```
pranavkm, line 182: Pretty red color

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -67,7 +67,13 @@ public Program(IConsole console, string workingDirectory)
             // AppContext.BaseDirectory = $sdkRoot\$sdkVersion\DotnetTools\dotnet-watch\$version\tools\net6.0\any\
             // MSBuild.dll is located at $sdkRoot\$sdkVersion\MSBuild.dll
             var sdkRootDirectory = Path.Combine(AppContext.BaseDirectory, "..", "..", "..", "..", "..", "..");
+#if DEBUG
+            // In the usual case, use the SDK that contains the dotnet-watch. However during local testing, it's
+            // much more common to run dotnet-watch from a different SDK. Use the ambient SDK in that case.
+            MSBuildLocator.RegisterDefaults();
```
pranavkm, line 73: Annoying local build issue where it would use the dogfood SDK. This was causing the annoying FailFast error that I poked you about.

src/BuiltInTools/dotnet-watch/HotReload/DefaultDeltaApplier.cs
```diff
@@ -29,14 +30,8 @@ public DefaultDeltaApplier(IReporter reporter)
 
         public bool SuppressBrowserRefreshAfterApply { get; init; }
 
-        public async ValueTask InitializeAsync(DotNetWatchContext context, CancellationToken cancellationToken)
+        public ValueTask InitializeAsync(DotNetWatchContext context, CancellationToken cancellationToken)
         {
-            if (_pipe is not null)
```
captainsafia, line 34: Why was this removed? It seems like good hygiene to have it?

captainsafia (approved):

src/BuiltInTools/dotnet-watch/HotReload/DefaultDeltaApplier.cs
pranavkm: Turns out we were newing up these instances in every hot reload loop. So this was never true. Also why the named pipe name is a static because we were only configuring it once.

PR #17612, opened 2021-05-13
Update Compatibility version and remove IsShipping

safern:
While chatting with @ericstj we found out this doesn't need to ship as a package just yet as the package validation ships it as part of it's tooling.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <IsPackable>true</IsPackable>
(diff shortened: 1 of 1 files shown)


ericstj (approved):

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -3,9 +3,8 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <IsPackable>true</IsPackable>
```
ericstj, line 5: Actually should we set IsPackable to false, or do you have a scenario for consuming the library today?

safern: I was going to remove it, but we might end up the need to ship it if we want to build more tooling on top of it that doesn't leave on the SDK? (Analyzer?). I don't mind stop producing a package on the meantime though.

Anipik (approved):

PR #17618, opened 2021-05-13
Add support for new files

pranavkm:


src/BuiltInTools/dotnet-watch/HotReloadDotNetWatcher.cs
```diff
@@ -122,8 +122,20 @@ public async Task WatchAsync(DotNetWatchContext context, CancellationToken cance
                         }
                         else
                         {
-                            _reporter.Output($"File changed: {fileItem.FilePath}.");
+                            if (fileItem.IsNewFile)
```
pranavkm, line 125: The FSW in dotnet-watch generally knows about newly added files. However, the current behavior is to ignore it if the file wasn't previously being watched (which is a list computed by msbuild evaluation one time and has to be explicitly re-computed). Most of the work in this PR is to plumb this information thru and the watcher to do a build if it sees a new file that could possibly need a rebuild (i.e. .cs, .razor, .cshtml).

I'm not super happy with how this ends up being, (it's more complicated code inside the watch loop) but it's functional. An alternative might to this would be to have a cancelation token that fires that exists this iteration of watch when a new file is added, but we still have to bake in knowledge of file extensions somewhere. For .e.g, we don't want to rebuild if you add a css or js file.

PR #17624, opened 2021-05-14
Add mock dotnet workload update

wli3:
Workload update flow. Currently in mock.

# display available update

```shell
dotnet workload list --machine-readable --target-sdk-version 6.0.200
```

--target-sdk-version should also be set by VSMac updater. Since VSMac updater could update .NET SDK from one version band to the other, say 6.0.100 to 6.0.200. And .NET SDK need to know the version band of the targeting SDK (in this case 6.0.200). VSMac updater should always pass the .NET SDK version that it installed.

Response:

```shell
==workloadListJsonOutputStart==
{"installed":[],"updateAvailable":[{"manifestVersion":"6.0.100","description":"android-workload-description: for testing you can delete the content of C:\\work\\sdk\\artifacts\\bin\\redist\\Debug\\dotnet\\DEV_mockworkloads\\update to revert the mock update","workloadId":"microsoft-android-sdk-full"},{"manifestVersion":"6.0.100","description":"ios-workload-description: for testing you can delete the content of C:\\work\\sdk\\artifacts\\bin\\redist\\Debug\\dotnet\\DEV_mockworkloads\\update to revert the mock update","workloadId":"microsoft-ios-sdk-full"}]}
==workloadListJsonOutputEnd==
```

VSMac updater should read the updateAvailable session. _workloadId_, _description_ should also be displayed to the user. The mock implementation does not take what is installed in to consideration. The real implementation would.

# download files

Similar to VSMac installer. VSMac update would download the package itself while asking for the URLS to download. Note, if the user decide the update, all workload would be updated. VSMac updater should reflect that.

```shell
 dotnet workload update --print-download-link-only --target-sdk-version 6.0.200
```

```shell
==allPackageLinksJsonOutputStart==
["https://www.myget.org/F/mockworkloadfeed/api/v3/flatcontainer/microsoft.ios.bundle/6.0.100/microsoft.ios.bundle.6.0.100.nupkg","https://www.myget.org/F/mockworkloadfeed/api/v3/flatcontainer/microsoft.net.workload.android/6.0.100/microsoft.net.workload.android.6.0.100.nupkg"]
==allPackageLinksJsonOutputEnd==
```

# finish update

```shell
dotnet workload update --from-cache /vsmacupdater/tempfolder --target-sdk-version 6.0.200
```

return code 0 means success. This command has no output.

/vsmacupdater/tempfolder is the temp location of VSMac updater. VSMac updater should be in charge of cleaning up when update is done.

Mock uses folder `DEV_mockworkloads\update` that is sitting in the same root folder as dotnet executable. You can delete the entire folder to set the environment back. And since all operation happens in a root access location, sudo are required. This is also true for the real implementation.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
@@ -14,6 +14,7 @@ using Microsoft.DotNet.Workloads.Workload.Install;
 using Microsoft.DotNet.Workloads.Workload.Install.InstallRecord;
 using Microsoft.NET.Sdk.WorkloadManifestReader;
 using System.Linq;
+using Microsoft.DotNet.Workloads.Workload.Update;
 
 namespace Microsoft.DotNet.Workloads.Workload.List
 {
@@ -24,6 +25,7 @@ namespace Microsoft.DotNet.Workloads.Workload.List
         private readonly bool _machineReadableOption;
         private readonly IWorkloadInstallationRecordRepository _workloadRecordRepo;
         private readonly SdkFeatureBand _sdkFeatureBand;
+
 
         public WorkloadListCommand(
             ParseResult result,
@@ -40,8 +42,11 @@ namespace Microsoft.DotNet.Workloads.Workload.List
             var workloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(dotnetPath, sdkVersion.ToString());
             var workloadResolver = WorkloadResolver.Create(workloadManifestProvider, dotnetPath, sdkVersion.ToString());
             _sdkFeatureBand = new SdkFeatureBand(sdkVersion);
-            _workloadRecordRepo = workloadRecordRepo ?? 
-                WorkloadInstallerFactory.GetWorkloadInstaller(_reporter, _sdkFeatureBand, workloadResolver, _verbosity).GetWorkloadInstallationRecordRepository();
+            _workloadRecordRepo = workloadRecordRepo ??
+                                  WorkloadInstallerFactory
+                                      .GetWorkloadInstaller(_reporter, _sdkFeatureBand, workloadResolver, _verbosity)
+                                      .GetWorkloadInstallationRecordRepository();
+            _sdkVersion = result.ValueForOption<string>(WorkloadUpdateCommandParser.SdkVersionOption);
         }
 
         public override int Execute()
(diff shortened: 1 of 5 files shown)


src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -40,23 +42,26 @@ internal class WorkloadListCommand : CommandBase
...
             var installedList = _workloadRecordRepo.GetInstalledWorkloads(_sdkFeatureBand);
             if (_machineReadableOption)
             {
-                var outputJson = new Dictionary<string, string[]>()
-                {
-                    ["installed"] = installedList.Select(id => id.ToString()).ToArray()
-                };
+                var updateAvailable = MockUpdateAvailable();
+                var listOutput = new ListOutput(Installed: installedList.Select(id => id.ToString()).ToArray(),
+                    UpdateAvailable: updateAvailable);
```
sfoslund, line 59: So we expect to display updates by default when the machine readable option is present? Or is there a switch for displaying updates?

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommandParser.cs
```diff
@@ -9,18 +9,15 @@ namespace Microsoft.DotNet.Cli
...
-
+        public static readonly Option MachineReadableOption = new Option<bool>("--machine-readable") {IsHidden = true};
         public static readonly Option VerbosityOption = CommonOptions.VerbosityOption();
 
         public static Command GetCommand()
         {
             var command = new Command("list", LocalizableStrings.CommandDescription);
             command.AddOption(MachineReadableOption);
             command.AddOption(VerbosityOption);
+            command.AddOption(WorkloadUpdateCommandParser.SdkVersionOption);
```
sfoslund, line 20: nit: if we want to have separate option descriptions for this option when applied to list vs update we'll have to have a separate copy here for list.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -1,25 +1,94 @@
...
         public override int Execute()
         {
-            // TODO stub
-            Reporter.Output.WriteLine("WIP workload update");
+            if (_printDownloadLinkOnly || !string.IsNullOrWhiteSpace(_fromCacheOption))
+            {
+                if (string.IsNullOrWhiteSpace(_sdkVersion))
+                {
+                    throw new ArgumentException(
+                        "SDK version is require when using --print-download-link-only option. Since the update may cross version band");
```
sfoslund, line 47: Is it also required for _fromCacheOption?

sfoslund, line 47: Why do we need to require this as opposed to defaulting to the current feature band?

sfoslund (approved):

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
wli3: Just one flag to display both. Since even for VSMac installer, if a workload has update available, it will be marked as "update available" instead.

wli3: that reminds me, what happens if I run `dotnet workload install android` when the android workload is installed already? Will it get updated, get an error or do nothing

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommandParser.cs
wli3: This is for internal use, so we don't need to have a too detailed description

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
wli3: yes. We should verify the cached manifest machines the version.

wli3: For VSMac, it is hard to tell what SDK is within version band what is not, might as well require that all the time

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
sfoslund: It should get updated

Issue #14955, opened 2020-12-12
Uninstalling and updating an optional workload


marcpopMSFT:

sfoslund:
Removing "Implicit check for update (check periodically or tie to a command. Check for update of project and check for SDK should be able to be distinguished by the user) below cutline." as it's tracked by https://github.com/dotnet/sdk/issues/17394

sfoslund:
> Check update with explicit command. Same logic for both non admin and VS/standalone installer since there is no permission issue. (add to dotnet --info too?)

@wli3 does the list PR you have out now cover this? Or does that only advertise when the print urls option is set?

wli3:
@sfoslund it is more for https://github.com/dotnet/sdk/issues/17394 if we decide not to a background check, but an explicit check instead. I will just remove the line here.

PR #17642, opened 2021-05-14
Workload update

sfoslund:
Tested manually with Android workload

One of https://github.com/dotnet/sdk/issues/14955


sfoslund:
@wli3 I'm planning on implementing the VS for mac options (offline caches, download links) in a separate PR after this.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -269,6 +270,13 @@ public void GarbageCollectInstalledWorkloadPacks()
                         unneededBandRecords = unneededBandRecords.Append(currentBandRecordPath);
                     }
 
+                    if (!unneededBandRecords.Any())
+                    {
+                        continue;
+                    }
+
+                    // Save the pack info in case we need to delete the pack
```
wli3, line 278: Any test coverage for adding pack in for to installation record? This is to cover manifest pack info change in between versions?

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -90,5 +121,100 @@ public override int Execute()
...
+                        {
+                            // Don't hide the original error if roll back fails
+                            _reporter.WriteLine(string.Format(LocalizableStrings.RollBackFailedMessage, e.Message));
+                        }
+                    });
+            }
+            else
+            {
+                var installer = _workloadInstaller.GetWorkloadInstaller();
+                foreach (var workloadId in workloadIds)
```
wli3, line 213: should have a test on when packs are not installed, this update is considered an install

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -90,5 +121,100 @@ public override int Execute()
...
+                            .Select(packId => _workloadResolver.TryGetPackInfo(packId));
+
+                        foreach (var packId in workloadPackToUpdate)
+                        {
+                            installer.InstallWorkloadPack(packId, sdkFeatureBand);
+                        }
+
+                        foreach (var workloadId in workloadIds)
+                        {
+                            _workloadInstaller.GetWorkloadInstallationRecordRepository()
```
wli3, line 178: there is no delete of the old record ? and we don't run a garbage collection after the update?

wli3 (approved):
Looks good. But I have some questions

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
sfoslund: Garbage collection (including deleting the old records) is called here: https://github.com/dotnet/sdk/pull/17642/files#diff-32dff2c2d8df112281b252add6af08593e98425431807590ed2422e93c8bcdb1R138

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
sfoslund: Yes, good idea, I'll add a test.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
sfoslund: I think [this existing test ](https://github.com/dotnet/sdk/pull/17642/files#diff-7e506f78957029dd1bd1c2929f3e64a542fe4c0c29bbbbc54f95e29b5dc99499R40)covers this case, since right now the top level update command doesn't know if the pack is installed or not and it just asks the managed installer to install them all, knowing it will be a no op if the pack is already installed.

src/Tests/dotnet-workload-update.Tests/GivenDotnetWorkloadUpdate.cs
```diff
@@ -0,0 +1,97 @@
...
+            _reporter = new BufferedReporter();
+            _manifestPath = Path.Combine(_testAssetsManager.GetAndValidateTestProjectDirectory("SampleManifest"), "Sample.json");
+            _parseResult = Parser.GetWorkloadsInstance.Parse(new string[] { "dotnet", "workload", "update" });
+        }
+
+        [Fact]
+        public void GivenWorkloadUpdateItUpdatesOutOfDatePacks()
+        {
+            var mockWorkloadIds = new WorkloadId[] { new WorkloadId("xamarin-android") };
+            (_, var command, var installer, _, _, _) = GetTestInstallers(_parseResult, installedWorkloads: mockWorkloadIds);
```
wli3, line 43: Is there an existing workload installed and then update? Looks like the test setup, there is no existing workload?

sfoslund:
@wli3 I added a more realistic update test, look okay to you?

wli3:
@sfoslund thanks! That looks good. Could you also find me the logic to remove the old installation record?

sfoslund:
> Could you also help me find the logic to remove the old installation record?

It's removed in garbage collection, this method: https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs#L250. It calls into this method (https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs#L292), which gets the "whitelist" of packs we actually need based on which workloads we have installed and any records from the current feature brand that isn't on that list is removed. In the context of an update, the list of expected, "whitelist" packs will contain the updated packs and pack versions but not the old ones, so the old records will be removed.

wli3 (approved):

PR #17644, opened 2021-05-15
[Api Compat] Add localized resources and update tests to not assert on those resources

### LINKED ISSUE #17371: [Api Compat] Add localized errors using resources
Currently all errors are hardcoded, we should move them to a resx file.

safern:
Fixes: #17371

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
@@ -9,7 +9,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// <summary>
     /// Class representing a difference of compatibility, containing detailed information about it.
     /// </summary>
-    public class CompatDifference : IDiagnostic, IEquatable<CompatDifference>
+    public class CompatDifference : IDiagnostic
     {
         /// <summary>
         /// The Diagnostic ID for this difference.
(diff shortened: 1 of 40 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// <summary>
     /// Class representing a difference of compatibility, containing detailed information about it.
     /// </summary>
-    public class CompatDifference : IDiagnostic, IEquatable<CompatDifference>
```
joperezr, line 12: What was the reasoning behind removing the IEquatable here? I see you are now moving it to the tests but wouldn't it still make sense to keep it in here in case there is ever a need to diff to CompatDifferences? In particular, I think that this may be useful when dealing with suppressions.

joperezr (approved):
Added one small comment but this looks good otherwise

Anipik (approved):

joperezr:
Merging this as is for now, I might address that comment I left about CompatDifference once we build the warning baseline story.

safern:
> I might address that comment I left about CompatDifference once we build the warning baseline story

You might want to look at how `DiagnosticBag` filters diagnostics. I don't think implementing IEquatable is the way to go as we will not baseline based on messages since those are localized. Also I don't think you should create an instance of CompatDifference on the filter side for each filter just to call .Equals, when we can compare and access the properties exposed in the interface/class.

safern:
By the way, thanks for the review and merging while I'm OOF.

PR #17655, opened 2021-05-16
Update property name in `TargetFrameworkEval` telemetry event

Nirmal4G:
#### CHANGES

Update the property name in `TargetFrameworkEval` telemetry event to use the correct terminology.
Here, the `TargetFramerworkMoniker` is being sent but the property name used is `TargetFrameworkVersion`.

#### NOTES

These are some of the missed refactoring, case changes and incorrect names
that might've been overlooked in this section of the .NET SDK's codebase.

_Since, this involves a change in how the data is sent, server-side update with similar changes may be necessary._

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
+++ b/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
@@ -14,16 +14,17 @@ namespace Microsoft.DotNet.Tools.MSBuild
 {
     public sealed class MSBuildLogger : INodeLogger
     {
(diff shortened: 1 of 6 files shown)


Nirmal4G:
@sfoslund @wli3 🔔

Nirmal4G:
@sfoslund @wli3 🔔

wli3:
Thank you for your contribution. However, the data process pipeline is hooked up with the existing property name. It is non trivial to migrate them over.

Nirmal4G:
It's fine. It thought it's a very trivial change but it's not, huh!? Just my luck! 😔

PR #17657, opened 2021-05-16
SDK support for runtime native host and COM changes

LakshanF:
SDK property support for the native host (IJW & Managed Component) feature switches. Also fixed name changes for the COM feature switch

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
@@ -52,15 +52,37 @@ Copyright (c) .NET Foundation. All rights reserved.
     <EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>false</EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>
   </PropertyGroup>
 
-  <!-- We disable in-built COM support for trimmed apps here so that the feature
-       switch can flow to the runtimeconfig.json. Built-in COM support is disabled
+  <!-- We disable in-built COM Interop support for trimmed apps here so that the feature
+       switch can flow to the runtimeconfig.json. Built-in COM Interop support is disabled
        by default since they may require assemblies, types or members that
        could be removed by the linker, causing a trimmed app to crash. -->
-  <PropertyGroup Condition="'$(BuiltInComSupport)' == '' And
+  <PropertyGroup Condition="'$(BuiltInComInteropSupport)' == '' And
                             '$(PublishTrimmed)' == 'true' And
                             $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))">
-    <BuiltInComSupport>false</BuiltInComSupport>
-  </PropertyGroup>  
+    <BuiltInComInteropSupport>false</BuiltInComInteropSupport>
+  </PropertyGroup>
+
+  <!-- We disable Native Host entry point support for IJW (C++/CLI Assemblies) for trimmed
+       apps here so that the feature switch can flow to the runtimeconfig.json. 
+       Native Host entry point support is disabled
+       by default since they may require assemblies, types or members that
+       could be removed by the linker, causing a trimmed app to crash. -->
+  <PropertyGroup Condition="'$(EnableCPlusPlusCLIHostActivation)' == '' And
+                            '$(PublishTrimmed)' == 'true' And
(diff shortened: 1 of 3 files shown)


vitek-karas (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
```diff
@@ -52,15 +52,37 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  </PropertyGroup>
+
+  <!-- We disable Native Host entry point support for IJW (C++/CLI Assemblies) for trimmed
+       apps here so that the feature switch can flow to the runtimeconfig.json. 
+       Native Host entry point support is disabled
+       by default since they may require assemblies, types or members that
+       could be removed by the linker, causing a trimmed app to crash. -->
+  <PropertyGroup Condition="'$(EnableCPlusPlusCLIHostActivation)' == '' And
+                            '$(PublishTrimmed)' == 'true' And
+                            $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))">
```
eerhardt, line 72: I wonder if it is time to extract 

```
'$(PublishTrimmed)' == 'true' And                            $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))
```

Into a `<PropertyGroup>` around all these settings. It is duplicated 6 times now with these changes. This can easily be re-written to:

```xml
  <PropertyGroup Condition="'$(PublishTrimmed)' == 'true' And
                            $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))">

    <StartupHookSupport Condition="'$(StartupHookSupport)' == ''">false</StartupHookSupport>
    <CustomResourceTypesSupport Condition="'$(CustomResourceTypesSupport)' == ''">false</CustomResourceTypesSupport>
    <EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization Condition="'$(EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization)' == ''">false</EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>
    <BuiltInComSupport Condition="'$(BuiltInComSupport)' == ''">false</BuiltInComSupport>
    <EnableCPlusPlusCLIHostActivation Condition="'$(EnableCPlusPlusCLIHostActivation)' == ''">false</EnableCPlusPlusCLIHostActivation>
    <_EnableCallingManagedFunctionFromNativeHosting Condition="'$(_EnableCallingManagedFunctionFromNativeHosting)' == ''">false</_EnableCallingManagedFunctionFromNativeHosting>

  </PropertyGroup>
```

sbomer (approved):

mateoatr (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -359,9 +359,14 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   <ItemGroup>
-    <RuntimeHostConfigurationOption Include=" System.Runtime.InteropServices.Marshal.IsBuiltInComSupported"
-                                    Condition="'$(BuiltInComSupport)' != ''"
-                                    Value="$(BuiltInComSupport)"
+    <RuntimeHostConfigurationOption Include="Internal.Runtime.InteropServices.ComponentActivator.IsSupported"
+                                    Condition="'$(_EnableCallingManagedFunctionFromNativeHosting)' != ''"
+                                    Value="$(_EnableCallingManagedFunctionFromNativeHosting)"
+                                    Trim="true" />
+
+    <RuntimeHostConfigurationOption Include="Internal.Runtime.InteropServices.InMemoryAssemblyLoader.IsSupported"
```
eerhardt, line 367: Why does this app context switch start with `Internal.`, but the MSBuild property doesn't start with an underscore?

If this is a public property we want to advertise and document, it shouldn't be named `Internal.`. We should name it something along the lines of `System.XXX.InMemoryAssemblyLoader.IsSupported`.

Even better, the naming between the app context switch and the MSBuild property should be aligned. One is called `InMemoryAssemblyLoader` while the other is `EnableCPlusPlusCLIHostActivation`. We should be using consistent names.

eerhardt (approved):
LGTM.

We should be careful about merging this in preview5. The runtime still needs to be updated with the new names. If this change went into preview5, but the corresponding runtime change didn't, we would have mis-matches.

PR #17699, opened 2021-05-18
[preview5] Remove linker warnings for native host and COM

sbomer:
This includes @LakshanF 's changes from https://github.com/dotnet/sdk/pull/17657, except for the renamed properties, so that the feature switches get set using the existing names. This gets rid of the following warnings which were not caught by the test infra:
```
ILLink : Trim analysis warning IL2026: Internal.Runtime.InteropServices.ComponentActivator.GetFunctionPointer(IntPtr,IntPtr,IntPtr,IntPtr,IntPtr,IntPtr
ILLink : Trim analysis warning IL2026: Internal.Runtime.InteropServices.InMemoryAssemblyLoader.LoadInMemoryAssembly(IntPtr,IntPtr
```

The warning checks in the tests have been made more strict. The issue was that the collapsed warnings weren't caught by the "ILLink :" check (they look like this):
```
System.Private.CoreLib.dll : warning IL2104: Assembly 'System.Private.CoreLib' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries
```

To fix this I'm checking for "warning IL" instead, and also not collapsing the warnings for the baseline test cases to produce detailed warnings if the test fails.


eerhardt (approved):
Thanks for taking care of this @sbomer!

LakshanF (approved):

PR #17720, opened 2021-05-19
Adding buildMultiTargeting package folder in order to add support for PackageReference import for multitargeting projects.

joperezr:
cc: @ericstj @Anipik @safern 

**Context:**
Currently the PackageValidation package is setup in a way that it runs Validation after the Pack target. Additionally, when using PackageReference this gets hooked by importing the props and targets located in the build\*props and build\*targets files. Finally, when a project multitargets, there are two types of builds that happen internally when build is invoked: the outer build (main build) and the inner builds (called  by outer, and will be individual builds which represent each supported target framework).

**Problem Definition:**
The problem is that when a consuming project is multitargeting, the Pack target gets called from the outer build but the build\*props and build\*targets files get imported only by the inner builds which means that today, using PackageReference for a project that is multitargeting is not supported.

**Proposed solution:**
This PR  is also copying the build\*props and targets into the buildMultiTargeting\ folder in the package, which will make it so that the targets also get imported by the outer build, so that validation will also run with projects that are multitargeting.

**Out of scope:**
One additional issue with using PackageReference when adding the validation infrastructure is that if the consuming project opted into validation against a baseline, we cannot add PackageDownload early enough in order to not require an additional project evaluation. This may be something that is fixed after this fix is in, but we will need to explore to make sure this can make that work.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
@@ -34,6 +34,15 @@
     <None Include="**/*.props;**/*.targets" Pack="true">
       <PackagePath>%(RecursiveDir)%(Filename)%(Extension)</PackagePath>
     </None>
+    
+    <!-- 
+      Additionally also copy the props and targets from the build folder into buildMultiTargeting so that
+      NuGet will also have these files automatically imported for consuming projects which multitarget
+      (use TargetFrameworks instead of TargetFramework) 
+    -->
+    <None Include="build/*.props;build/*.targets" Pack="true">
+      <PackagePath>buildMultiTargeting/%(Filename)%(Extension)</PackagePath>
+    </None>
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
@@ -34,15 +34,6 @@
     <None Include="**/*.props;**/*.targets" Pack="true">
       <PackagePath>%(RecursiveDir)%(Filename)%(Extension)</PackagePath>
     </None>
-    
-    <!-- 
-      Additionally also copy the props and targets from the build folder into buildMultiTargeting so that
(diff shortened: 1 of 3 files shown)


src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -34,6 +34,15 @@
     <None Include="**/*.props;**/*.targets" Pack="true">
       <PackagePath>%(RecursiveDir)%(Filename)%(Extension)</PackagePath>
     </None>
+    
+    <!-- 
+      Additionally also copy the props and targets from the build folder into buildMultiTargeting so that
+      NuGet will also have these files automatically imported for consuming projects which multitarget
+      (use TargetFrameworks instead of TargetFramework) 
+    -->
+    <None Include="build/*.props;build/*.targets" Pack="true">
```
ericstj, line 43: Instead of copying these, can we just have stubs that import the ones in the build folder?

buildMultiTargeting\Microsoft.DotNet.PackageValidation.props
```xml
<Project>
  <Import Project="..\build\$(MSBuildThisFile)" />
</Project>
```
buildMultiTargeting\Microsoft.DotNet.PackageValidation.targets
```xml
<Project>
  <Import Project="..\build\$(MSBuildThisFile)" />
</Project>
```

joperezr: sure, I can do that too.

ericstj (approved):
LGTM

Anipik (approved):

ViktorHofer:
This package isn't supposed to be consumed via PackageReference as it contributes to the project restore and hence it is equired to be referenced as an msbuild sdk via the `<Sdk />` element. If I'm not missing a detail here, I would just delete both buildTransitive and the build folder in favor of the sdk folder.

joperezr:
The only reason why the suggested path for us to consume this package is via SDK is due to the automatic download of the baseline in case you want to enable that type of validation. For the rest of the validation, consuming this via PackageReference works (now after this PR is merged). Even if you want to do baseline validation, PackageReference still works as long as you manually add the packagedownload to your csproj, so I think it is good to allow this flexibility on the consumption side for consumers.

One more point here is that we might change the design altogether for how to restore that baseline package and remove the requirement for this to be an SDK, so I don't think we should remove anything just yet.

Issue #16484, opened 2021-03-23
VS mac update/installer - .NET SDK integration


wli3:
- [x] Design [spec](https://microsoft-my.sharepoint.com/:w:/p/wul/EQq7eaqvdChBoN7ucln0H_UBRICzaZcC1IS5CroKOotUgA?e=ujCicM)
- [x] Dummy implementation so VS mac update/installer does not need to integration until the very end
  - [x] dotnet workload install
  - [x] dotnet workload update within a feature band
  - [x] dotnet workload update while crossing band update
  - [x] dotnet workload list (dotnet workload list --machine-readable --target-sdk-version 6.0.200)
- [x] Real implementation so VS mac update/installer does not need to integration until the very end
  - [x] dotnet workload install
  - [x] dotnet workload update within a feature band
  - [x] dotnet workload update while crossing band update
  - [x] dotnet workload list (dotnet workload list --machine-readable --target-sdk-version 6.0.200)
- [x] Collaborate with @GouriKumari to finish the implementation on VS mac updater and installer
- [x] Proper command implementation specific to VSMac install flow
  - [x] _machineReadableOption FromCacheOption validation
- [x] Testing

marcpopMSFT:
Testing is ongoing but closing out as all known functional work is complete.

PR #17766, opened 2021-05-20
Adding offline caching for workload update

sfoslund:


src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -26,6 +25,8 @@ internal class WorkloadListCommand : CommandBase
         private readonly IWorkloadInstallationRecordRepository _workloadRecordRepo;
         private readonly SdkFeatureBand _sdkFeatureBand;
 
+        public static readonly string MockUpdateDirectory = Path.Combine(Path.GetDirectoryName(Environment.ProcessPath),
```
sfoslund, line 28: @wli3 are you already tracking replacing this mock update advertisement with a real implementation or should I add it to a issue?

sfoslund:
One note here, we're seeing some errors due to the way workloads are versioning. For example, 14.4.100-preview.3.1326 and 14.4.100-ci.TJTestingLocaliztionSync.820 can't be compared. @jonathanpeppers is your team going to standardize your versioning scheme or is this an issue with the feeds that I'm testing with?

joeloff:
> One note here, we're seeing some errors due to the way workloads are versioning. For example, 14.4.100-preview.3.1326 and 14.4.100-ci.TJTestingLocaliztionSync.820 can't be compared. @jonathanpeppers is your team going to standardize your versioning scheme or is this an issue with the feeds that I'm testing with?

Can't you use the semver implementation from Releases API? We're already pulling that into the SDK. You can't do a straight string compare if the version string has semantic elements

jonathanpeppers:
At one point the iOS team was accidentally pushing packages from PRs, which should be fixed now.

Any of the following versions would be valid, I think:

* 14.4.100-ci.main.820
* 14.4.100-preview.4.820
* 14.4.100-rc.1.820

There might be other `ci.[branchname]` packages.

I don't think we plan to bump the `100` until GA ships. 14.4 matches the OS version, so it might change when Apple releases something.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
wli3: please add a new issue. I need to somehow make that work or just implement the real logic. Although that could be take some work considering we need to surface version and description information

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -218,5 +210,108 @@ public void UpdateWorkloads(bool includePreviews = false)
...
+
+        private async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
+        {
+            var manifestPackagePaths = await _workloadManifestUpdater.DownloadManifestPackagesAsync(includePreview, tempPath);
+            await _workloadManifestUpdater.ExtractManifestPackagesToTempDirAsync(manifestPackagePaths, tempPath);
+            _workloadManifestProvider = new TempDirectoryWorkloadManifestProvider(tempPath.Value, _sdkVersion.ToString());
+            _workloadResolver = WorkloadResolver.Create(_workloadManifestProvider, _dotnetPath, _sdkVersion.ToString());
+        }
+
+        private IEnumerable<PackInfo> GetUpdatablePacks(IWorkloadPackInstaller installer)
```
wli3, line 280: I guess list will use this method. And other than piping the description info, that's all

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
wli3: nvm, https://github.com/dotnet/sdk/issues/16484 has a separate list todo item

src/Tests/dotnet-workload-update.Tests/GivenDotnetWorkloadUpdate.cs
```diff
@@ -142,20 +143,80 @@ public void GivenWorkloadUpdateItRollsBackOnFailedUpdate()
...
+        public void GivenWorkloadUpdateItPrintsDownloadUrls()
+        {
+            var mockWorkloadIds = new WorkloadId[] { new WorkloadId("xamarin-android") };
+            var parseResult = Parser.GetWorkloadsInstance.Parse(new string[] { "dotnet", "workload", "update", "--print-download-link-only" });
+            (_, var command, _, _, _, _) = GetTestInstallers(parseResult, installedWorkloads: mockWorkloadIds, includeInstalledPacks: true);
+
+            command.Execute();
+
+            _reporter.Lines.Should().Contain("==allPackageLinksJsonOutputStart==");
+            string.Join(" ", _reporter.Lines).Should().Contain("mock-url-xamarin.android.templates"); // New packs should be included
```
wli3, line 193: nit: fluent assertion usually has a second parameter as reason. The reason will show up as part of the assertion failure

```c#
string.Join(" ", _reporter.Lines).Should().Contain("mock-url-xamarin.android.templates", "New` packs should be included");
```

wli3 (approved):

PR #17799, opened 2021-05-22
use display string instead of name for namespaces

Anipik:
Symbol.name was returning the same output for fxResources.Microsoft.Extensions.FileGlobbing and Microsoft.Extensions.FileGlobbing., due to which we were getting false positives here https://github.com/dotnet/runtime/pull/52741

Display string fixes this issue. 

The assembly name was being passed with .version.nupkg, passing the package id. (No known effect of this but correcting it as required by apicompat)

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
@@ -16,16 +16,6 @@ namespace Microsoft.DotNet.ApiCompatibility
         public int GetHashCode(ISymbol obj) =>
             GetKey(obj).GetHashCode();
 
-        private static string GetKey(ISymbol symbol) =>
-            symbol switch
-            {
-                IMethodSymbol => symbol.ToComparisonDisplayString(),
-                IFieldSymbol => symbol.ToComparisonDisplayString(),
-                IPropertySymbol => symbol.ToComparisonDisplayString(),
-                IEventSymbol => symbol.ToComparisonDisplayString(),
-                ITypeSymbol => symbol.ToComparisonDisplayString(),
-                INamespaceSymbol => symbol.ToComparisonDisplayString(),
-                _ => symbol.ToComparisonDisplayString(),
-            };
+        private static string GetKey(ISymbol symbol) => symbol.ToComparisonDisplayString();
     }
 }
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
@@ -24,7 +24,8 @@ namespace Microsoft.DotNet.ApiCompatibility
                 IPropertySymbol => symbol.ToComparisonDisplayString(),
                 IEventSymbol => symbol.ToComparisonDisplayString(),
                 ITypeSymbol => symbol.ToComparisonDisplayString(),
-                _ => symbol.Name,
+                INamespaceSymbol => symbol.ToComparisonDisplayString(),
+                _ => symbol.ToComparisonDisplayString(),
             };
     }
 }
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
@@ -56,7 +56,7 @@ namespace Microsoft.DotNet.PackageValidation
                         baselineCompileTimeAsset.Path,
                         package.PackagePath,
                         latestCompileTimeAsset.Path,
-                        Path.GetFileName(package.PackagePath),
+                        package.PackageId,
                         Resources.BaselineVersionValidatorHeader,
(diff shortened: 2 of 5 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
```diff
@@ -24,7 +24,8 @@ symbol switch
                 IPropertySymbol => symbol.ToComparisonDisplayString(),
                 IEventSymbol => symbol.ToComparisonDisplayString(),
                 ITypeSymbol => symbol.ToComparisonDisplayString(),
-                _ => symbol.Name,
+                INamespaceSymbol => symbol.ToComparisonDisplayString(),
+                _ => symbol.ToComparisonDisplayString(),
```
ericstj, line 28: Did you mean to change the default case?  Does this mean we just always call `ToComparisonDisplayString()` and the switch does nothing?

ericstj:
> Symbol.name was returning the same output for fxResources.Microsoft.Extensions.FileGlobbing and Microsoft.Extensions.FileGlobbing

What was it returning?  `FileGlobbing`?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
ericstj: Actually, if I'm understanding this correctly as an identifier we use as a key for comparison, shouldn't we use DocID?

ericstj:
Please add test case(s) to cover this and any variants.   

A good test that I can imagine would be changing root of namespace
Left: `namespace a.b { public class c { } }`
Right: `namespace b.b { public class c { } }`
Should fail.

Another thing that comes to mind is making sure we capture the difference between a namespace and a nested class.
Left: `namespace a { public class b { } }`
Right: `public class a { public class b { } }`
Should fail.

Depending on what this `ToComparisonDisplayString` returns there might be lots of other tests to try to make sure we don't have collisions between different types of symbols.

Anipik:
> What was it returning? FileGlobbing?

yes thats correct. name returns the last thing.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
Anipik: yes we can remove the switch here. I think DisplayString is pretty close to documentation ID as well. its returning the strings in same manner

ericstj: Does it include the prefix to indicate the symbol type (eg property/field/type/etc)?

ericstj (approved):
Thanks for adding tests

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
```diff
@@ -56,7 +56,7 @@ public void Validate(Package package)
                         baselineCompileTimeAsset.Path,
                         package.PackagePath,
                         latestCompileTimeAsset.Path,
-                        Path.GetFileName(package.PackagePath),
+                        package.PackageId,
```
joperezr, line 59: The good thing about PackagePath though was that as you mention it has the version in it, so that makes it so that in the diff we can get more detailed messages when comparing against a baseline since we will get from which package version is each asset being selected from. I'm sure we can grab it from somewhere else, but for instance, I was using the fact that we had packagepath for the initial prototype of the warning suppressions.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/TypeMustExistTests.cs
```diff
@@ -55,6 +55,66 @@ public class First { }
...
+namespace B.B
+{
+  public class C { }
+}
+";
+            ApiComparer differ = new();
+            bool enableNullable = false;
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax, enableNullable);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax, enableNullable);
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(new[] { left }, new[] { right });
```
safern, line 78: NIT: there is an API that doesn't need an array for left and right.

safern (approved):
Thanks for fixing!

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
Anipik: We are passing this for apicompat usage which requires just the assembly name. package.version contains the package version.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs
Anipik: No it doesnt include that yet.

ericstj: I see, I wonder what other sorts of ambiguities might exist if we are omitting those designations cc @safern

PR #17825, opened 2021-05-24
[Just For review] Content For Broadcasting the packageValidation Package

Anipik:


src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Another common issue in the .NET ecosystem is that binary breaking changes aren't well understood. Developers tend to think of "does the source still compile" as the bar to determine whether an API is compatible. However, at the API level certain changes that are fine in C# aren't compatible at the level of IL, for example, adding a defaulted parameter or changing the value of a constant.
+
+Package Validation tooling will allow library developers to validate that their packages are well-formed and they they didn't make unintentional breaking changes.
+
+# How To Add Package Validation To Your Projects
+
+Package Validation is currently being shipped as a msbuild sdk package which can be consumed in the csproj. It is set of tasks and targets that run after the dotnet pack
```
joperezr, line 13:
```suggestion
Package Validation is currently being shipped as a msbuild sdk package which can be consumed by a project. It is a set of tasks and targets that run after generating the package when calling dotnet pack.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+The csproj will look like
```
joperezr, line 14:
```suggestion

The project should look like the following:
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+```C#
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <Sdk Name="Microsoft.DotNet.PackageValidation" Version="1.0.0-preview4.21121.3" />
```
joperezr, line 23: is this the final preview version we are planning to go with? If not, should we not include Version and let MSBuild pick the latest one in NuGet? (Not sure if that works with preview versions)

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+</Project>
+```
+
+## Scenarios and User Experience
+
+The first preview version of the package validation covers the basic scenarios for package validation. They are as follows:-
```
joperezr, line 29:
```suggestion
The first preview version of package validation covers the following basic scenarios:
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Validation Among Compatible Frameworks
```
joperezr, line 31:
```suggestion
### Validation Among Compatible Frameworks
```
Should this ( and the rest of scenarios) have an extra indentation?

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Package containing compatible frameworks need to ensure that code compiled against one can run against other. Examples of compatible pairs will be .NET Standard2.0 and .NET 6.0, .NET 5.0 and .NET 6.0 . In both of these cases, the users can compile against .NET Standard 2.0 or NET 5.0 and try to run on .NET 6.0, if the binaries are not compatible the users might end up with runtime errors.
```
joperezr, line 33:
```suggestion
Packages containing compatible frameworks need to ensure that code compiled against one can run against another. Examples of compatible framework pairs are: .NET Standard2.0 and .NET 6.0, .NET 5.0 and .NET 6.0 . In both of these cases, the users can compile against .NET Standard 2.0 or NET 5.0 and try to run on .NET 6.0, if the binaries are not compatible, users could end up with runtime errors.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Package Validation will catch these errors at pack time. eg.
```
joperezr, line 35:
```suggestion
Package Validation will catch these errors at pack time. Here is an example scenario:
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Finley is working on a game which do a lot of string manipulation. They need to support both .NET Framework and .NET Core customers. They started with just targeting .NET Standard 2.0 but now they realize they want to take advantage of the spans in .NET Core 3.0 and above to avoid unnecessary string allocations. In order to do that, they are multi-targeting for NET Standard2.0 and .NET 6.0,.
```
joperezr, line 37:
```suggestion
Finley is working on a game which does a lot of string manipulation. They need to support both .NET Framework and .NET Core customers. They started with just targeting .NET Standard 2.0 but now they realize they want to take advantage of spans in .NET Core 3.0 and above to avoid unnecessary string allocations. In order to do that, they now want to multi-target for NET Standard2.0 and .NET 6.0.
```

joperezr: NIT: this paragraph kind of says that Finley should multitarget .NET Standard2.0 and .NET Core 3.0/3.1 and above, instead of .NET 6.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+    {
+        // Do some string operations.
+    }
+#endif
+```
+
+When Finley packs the project it fails with the following error:
+
+```c#
+error : .NETStandard,Version=v2.0 assembly api surface area should be compatible with net6.0 assembly surface area so we can compile against .NETStandard,Version=v2.0 and run on net6.0 .framework.
```
joperezr, line 57: NIT: might be useful to show the command that Finley wrote to the console (dotnet pack) so that users can follow along where are those errors coming from.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+    public void DoStringManipulation(ReadOnlySpan<char> input)
+    {
+        // use spans to do string operations.
+    }
+#endif
+    public void DoStringManipulation(string input)
+    {
+        // Do some string operations.
+    }
+```
```
joperezr, line 75: Perhaps in here include another example after the fix of how calling dotnet pack now shows a successful build.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Validation Against Baseline Package Version
```
joperezr, line 77:
```suggestion
### Validation Against Baseline Package Version
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Package Validation helps to validate the latest version of the package with a previous released stable version of the package. This is an opt in feature. The user will need to add the ```EnablePackageBaselineValidation``` to the csproj.
```
joperezr, line 79:
```suggestion
Package Validation can also help validating a users library project against a previous released stable version of the package. In order to use this feature, the user will need to add the ```EnablePackageBaselineValidation``` to their project.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Package validation will detect any breaking changes on any of the shipped target frameworks and will also detect if any target framework support has been dropped.
+
+eg Tom works on the AdventureWorks.Client NuGet package. They want to make sure that they don't accidentally make breaking changes so they configure their project to instruct the package validation tooling to run API compatibility on the previous version of the package.
```
joperezr, line 83:
```suggestion
For example consider the following scenario: Tom works on the AdventureWorks.Client NuGet package. They want to make sure that they don't accidentally make breaking changes so they configure their project to instruct package validation tooling to run API compatibility against the previous version of the package.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+```csproj
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <PackageVersion>2.0.0</PackageVersion>
+    <EnablePackageBaselineValidation>true</EnablePackageBaselineValidation>
+  </PropertyGroup>
+
+  <Sdk Name="Microsoft.DotNet.PackageValidation" Version="1.0.0-preview4.21111.2" />
```
joperezr, line 94: Same here, let's try to move the SDK tag right after Project tag.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+</Project>
+```
+
+A few weeks later, Tom is tasked with adding support for a connection timeout to their library. The Connect method looks like this right now:
```
joperezr, line 98:
```suggestion
A few weeks later, Tom is tasked with adding support for a connection timeout to their library. The Connect method currently looks like this:
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+```C#
+public static HttpClient Connect(string url)
+{
+    // ...
+}
+```
+
+Since a connection timeout is an advanced configuration setting they reckon they can just add an optional parameter:
```
joperezr, line 107:
```suggestion
Since a connection timeout is an advanced configuration setting, they reckon they can just add an optional parameter:
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+```c#
+public static HttpClient Connect(string url, TimeSpan? timeout = default)
+{
+    // ...
+}
+```
+
+However, when they rebuild the package they are getting an error:
```
joperezr, line 116:
```suggestion
However, when they rebuild the package they are getting the following validation error:
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+```c#
+error : There are breaking changes between the versions. Please add or modify the apis in the recent version or suppress the intentional breaking changes. 
+error : API compatibility errors in between lib/net6.0/A.dll (left) and lib/net6.0/A.dll (right) for versions 1.0.0 and 2.0.0 respectively:
+error CP0002: Member 'A.B.Connect(string)' exists on the left but not on the right 
+```
+
+Tom realizes that while this is not a source breaking change, it's a binary breaking change. They solve this problem by adding an overload instead:
```
joperezr, line 123: Perhaps consider adding links for what a [source breaking change](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#source-breaking-change) and [binary breaking change](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change) mean

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+public static HttpClient Connect(string url, TimeSpan timeout)
+{
+    // ...
+}
+```
+
+## Validation Against Different Runtimes
+
+.Net Packages can contain different implementation assemblies for different runtimes. The developer needs to make sure that these assemblies are compatible with the compile time assemblies.
```
joperezr, line 138:
```suggestion
.Net Packages may choose to have different implementation assemblies for different runtimes. In that case, developers need to make sure that these assemblies are compatible with the compile-time assemblies.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+Rohan is working on a library involving some interop calls to unix and windows api respectively. The source code looks like
```
joperezr, line 140:
```suggestion
For example, consider the following scenario: Rohan is working on a library involving some interop calls to unix and windows api respectively. The source code looks like
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+        // call windows specific stuff
+    }
+#endif
+```
+
+The resuting package structure looks like
+
+```
+lib -> net6.0 -> A.dll 
+runtimes -> unix -> lib -> net6.0 -> A.dll
```
joperezr, line 160:
```suggestion
lib/
  net6.0/
          A.dll
runtimes/
            unix/
                 lib/
                   net6.0/
                           A.dll
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+```lib\net6.0\A.dll``` will always be used at compile time irrespective of the underlying operating system. ```lib\net6.0\A.dll``` will be used at runtime for Non-Unix systems and ```runtimes\unix\lib\net6.0\A.dll``` will be used at runtime for Unix system.
```
joperezr, line 163:
```suggestion
```lib\net6.0\A.dll``` will always be used at compile time irrespective of the underlying operating system. ```lib\net6.0\A.dll``` will also be used at runtime for Non-Unix systems, but ```runtimes\unix\lib\net6.0\A.dll``` will be used at runtime for Unix systems.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+
+When Rohan tries to pack this project, he gets an error:
+
+```
+error : The compile time assemblies public api surface area should be compatible with the runtime assemblies for all target frameworks and RIDs. Please add the following apis to the runtime assemblies.
+error : API Compatibility errors between lib/net6.0/A.dll (left) and runtimes/unix/lib/net6.0/A.dll (right): 
+error CP0002: Member 'A.B.Open(string)' exists on the left but not on the right.
+```
+
+Rohan quickly realizes his mistake and add ```A.B.Open(string)``` to the unix runtime as well.
```
joperezr, line 173:
```suggestion
Rohan quickly realizes his mistake and adds ```A.B.Open(string)``` to the unix runtime as well.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,198 @@
...
+    {
+        // call windows specific stuff
+    }
+#endif
+```
+
+## Whats Next For Package Validation
+
+- Error Suppression for Package Validation Errors eg Intentional breaking changes between versions.
+- More Compatibilty Rules eg compatible assembly versions, nullability.
```
joperezr, line 197:
```suggestion
- Error suppression for package validation errors like intentional breaking changes between versions.
- More API compatibilty rules like compatible assembly versions, nullability, attributes.
```

joperezr:
Most of the comments I left are nits but this looks great, thanks @Anipik! I have a general question: have we finalized on the design of the baseline comparison onboarding? I ask because talking to the SDK and MSBuild teams, they suggested against trying to add the PackageDownload automatically and instead have part of the instructions be asking the user to add the PackageDownload directly on their project.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+
+Package Validation tooling will allow library developers to validate that their packages are well-formed and they they didn't make unintentional breaking changes.
+
+# How To Add Package Validation To Your Projects
+
+Package Validation is currently being shipped as a msbuild sdk package which can be consumed by a project. It is a set of tasks and targets that run after generating the package when calling dotnet pack.
+
+The project should look like the following:
+
+```C#
```
joperezr, line 17:
```suggestion
```xml
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+    public void DoStringManipulation(string input)
+    {
+        // Do some string operations.
+    }
+#endif
+```
+
+When Finley packs the project it fails with the following error:
+
+```c#
```
joperezr, line 57:
```suggestion
```cmd
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+public static HttpClient Connect(string url, TimeSpan? timeout = default)
+{
+    // ...
+}
+```
+
+However, when they rebuild the package they are getting the following validation error:
+```c#
```
joperezr, line 118:
```suggestion
```cmd
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+error : There are breaking changes between the versions. Please add or modify the apis in the recent version or suppress the intentional breaking changes. 
+error : API compatibility errors in between lib/net6.0/A.dll (left) and lib/net6.0/A.dll (right) for versions 1.0.0 and 2.0.0 respectively:
+error CP0002: Member 'A.B.Connect(string)' exists on the left but not on the right 
+```
+
+Tom realizes that while this is not a source breaking change, it's a binary breaking change. They solve this problem by adding an overload instead:
+```
```
joperezr, line 125:
```suggestion
```c#
```

Anipik:
> I have a general question: have we finalized on the design of the baseline comparison onboarding? I ask because talking to the SDK and MSBuild teams, they suggested against trying to add the PackageDownload automatically and instead have part of the instructions be asking the user to add the PackageDownload directly on their project.

i am leaning more towards that as well as our restore mechanism is currently broken as well

joperezr:
>i am leaning more towards that as well as our restore mechanism is currently broken as well

@dsplaisted and @rainersigwald have voiced some opinions around that, so we should probably ask them once we think have a design we want to go with. **TL;DR** for both of  you guys: we are considering taking @dsplaisted suggestion made offline which was to not try and add a `PackageDownload` dynamically with our validation infrastructure, and instead just ask users that want to onboard on this to manually add PackageDownload themselves. @Anipik if we do this, then that means that this feature doesn't really need to be an SDK any longer, and we can instead just be a regular PackageReference.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+    }
+#else
+    public static void Open(string path)
+    {
+        // call windows specific stuff
+    }
+#endif
+```
+
+The resuting package structure looks like
```
ericstj, line 157:
```suggestion
The resulting package structure looks like
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
ericstj: IMHO a flat structure is easier to read
```suggestion
lib/net6.0/A.dll 
runtimes/unix/lib/net6.0/A.dll
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+    {
+        // call windows specific stuff
+    }
+#endif
+```
+
+## Whats Next For Package Validation
+
+- Error suppression for package validation errors like intentional breaking changes between versions.
+- More API compatibilty rules like compatible assembly versions, nullability, attributes.
```
ericstj, line 198:
```suggestion
- More API compatibility rules like compatible assembly versions, nullability, attributes.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
+## Package Validation
+
+With .NET Core & Xamarin we have made cross-platform a mainstream requirement for library authors. However, we lack validation tooling which can result in packages that don't work well which in turn hurts our ecosystem. This is especially problematic for emerging platforms where adoption isn't high enough to warrant special attention by library authors.
```
ericstj, line 3: Minor rewording suggestion
```suggestion
With .NET Core & Xamarin we have made cross-platform a mainstream requirement for library authors. However, we lack validation tooling for cross targeting packages, which can result in packages that don't work well, which in turn hurts our ecosystem. This is especially problematic for emerging platforms where adoption isn't high enough to warrant special attention by library authors.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+
+The tooling we provide as part of the SDK has close to zero validation that multi-targeted packages are well-formed. For example, a package that multi-targets for .NET 6.0 and .NET Standard 2.0 needs to ensure that code compiled against the .NET Standard 2.0 binary can run against the binary that is produced for .NET 6.0. We have seen this issue in the wild, even with 1st parties, for example, the Azure AD libraries.
```
ericstj, line 5:
```suggestion
The tooling we provide as part of the SDK has close to zero validation that multi-targeted packages are well-formed. For example, a package that multi-targets for .NET 6.0 and .NET Standard 2.0 needs to ensure that code compiled against the .NET Standard 2.0 binary can run against the .NET 6.0 binary. We have seen this issue in the wild, even with 1st parties, for example, the Azure AD libraries.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+
+Another common issue in the .NET ecosystem is that binary breaking changes aren't well understood. Developers tend to think of "does the source still compile" as the bar to determine whether an API is compatible. However, at the API level certain changes that are fine in C# aren't compatible at the level of IL, for example, adding a defaulted parameter or changing the value of a constant.
```
ericstj, line 7: Rather than writing this in a way that puts blame on the developer consider the following.

```suggestion
It's easy to think that a change is safe and compatible if source consuming that change continues to compile without changes. However, certain changes may work fine in C# but can cause problems at runtime if the consumer wasn't recompiled, for example, adding a defaulted parameter or changing the value of a constant.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
```diff
@@ -0,0 +1,199 @@
...
+
+Package Validation tooling will allow library developers to validate that their packages are well-formed and they they didn't make unintentional breaking changes.
```
ericstj, line 9:
```suggestion
Package Validation tooling will allow library developers to validate that their packages are well-formed and have no breaking changes.
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
Anipik: > is this the final preview version we are planning to go with? If not, should we not include Version and let MSBuild pick the latest one in NuGet? (Not sure if that works with preview versions)

we will probably need to specify it as we might be shipping multiple versions and latest version may not be the one that we want people to test. This version might change for the final post.

src/Compatibility/Microsoft.DotNet.PackageValidation/ReadMe.md
Anipik: The only issue there is that there will be other extra info like project paths and validation package path 
```
E:\A>dotnet pack
Microsoft (R) Build Engine version 16.10.0-preview-21181-07+073022eb4 for .NET

  Restored E:\A\A.csproj (in 222 ms).
  A -> E:\A\bin\Debug\netstandard2.0\A.dll
  A -> E:\A\bin\Debug\net6.0\A.dll
  Successfully created package 'E:\AllHands\CheckIfZero\bin\Debug\A.1.0.0.nupkg'.
C:\Users\.nuget\packages\microsoft.dotnet.packagevalidation\1.0.0-preview4\build\Microsoft.DotNet.PackageValidation.targets(28,5): error : .NETStandard,Version=v2.0 assembly api surface area should be compatible with net6.0 assembly surface area so we can compile against .NETStandard,Version=v2.0 and run on net6.0 .framework. [E:\A\A.csproj]
C:\Users\.nuget\packages\microsoft.dotnet.packagevalidation\1.0.0-preview4\build\Microsoft.DotNet.PackageValidation.targets(28,5): error : API Compatibility errors between lib/netstandard2.0/A.dll (left) and lib/net6.0/A.dll (right): [E:\AllHands\A\A.csproj]
C:\Users\.nuget\packages\microsoft.dotnet.packagevalidation\1.0.0-preview4\build\Microsoft.DotNet.PackageValidation.targets(28,5): error CP0002: Member 'A.B.DoStringManipulation(string)' exists on the left but not on the right [E:\A\A.csproj]
```
do u think it adds enough value ?

PR #17833, opened 2021-05-24
Account for case sensitive file systems in manifest updates

sfoslund:


src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -238,6 +239,20 @@ public void InstallWorkloadManifest(ManifestId manifestId, ManifestVersion manif
             }
         }
 
+        private string GetCaseSensitiveManifestDir(string manifestPath, ManifestId manifestId)
```
joeloff, line 242: Would it perhaps make sense to just name this GetManifestDir and pass an additional StringComparison parameter?

sfoslund: Maybe? I'm not sure this is going to be reused regardless. We normalize all our manifest/ pack ids to lower case so this is only going to be used here, when we're installing a new manifest and we want to ensure that we delete the old one (whatever casing it has)

wli3: We should make sure everything on disk is normalized instead of account for that. And since SDK put these files on disk. SDK should have total control. Or any interaction has a chance to have casing mismatch. Why casing mismatch happens in the beginning?

sfoslund: It's coming from the baseline manifest. We could update the existing logic for laying down the manifests in the installer repo but we'll also have to account for this in @joeloff 's tooling that will eventually replace that.

sfoslund: @wli3 do you want me to close this and change the way we lay down baseline manifests instead? My concern with that approach is that we could run into this bug again in the future when we change/ have different methods for laying down the baseline manifests.

wli3: I think we should close this. Even account for "different methods for laying down the baseline manifests". This solution will only fix 1 place of mismatch. We will still have unexpected mismatch in new method of laying down files.

The alternative is we could try to make every part of the system being case insensitive -- we at least need to build an abstraction for the file system to tolerate different casing, and always use the abstraction to access File IO.

But I think for now, SDK controls all path of laying files on disk. And if there are new ways, we need to ensure the new method will preserve the exact casing.

Issue #17879, opened 2021-05-26
[Package Validation]Add tests for msbuild tasks and targets


Anipik:
We dont have tests test our msbuild targets. We should add proper infra to test these and avoid this as  a point of failure.

cc @ericstj @ViktorHofer @safern @joperezr

PR #17880, opened 2021-05-26
Allow user to manually set the PackageValidationBaselineVersion

Anipik:
This will allow users to manually set PackageValidationBaselineVersion and will removal the code to automatically calculate the versions for us.

This is done to avoid the extra network calls/
We will also throw errors if the PackageValidationBaselineVersion or PackageValidationBaselinePath is not valid.

Tested locally for different cases. https://github.com/dotnet/sdk/issues/17879 issue to track the addition of tests.


src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -3,27 +3,22 @@
...
 
   <Target Name="RunPackageValidation"
           AfterTargets="Pack"
           Condition="$(IsPackable) == 'true'">
-    <PropertyGroup Condition="'$(EnablePackageBaselineValidation)' == 'true' and
-                              '$(PackageValidationBaselinePath)' == ''">
-      <PackageValidationBaselineName Condition="'$(PackageValidationBaselineName)' == ''">$(_PackageValidationBaselineName)</PackageValidationBaselineName>
-      <PackageValidationBaselineVersion Condition="'$(PackageValidationBaselineVersion)' == ''">$(_PackageValidationBaselineVersion)</PackageValidationBaselineVersion>
-      <PackageValidationBaselinePath Condition="'$(PackageValidationBaselineVersion)' != ''">$([MSBuild]::NormalizePath('$(NuGetPackageRoot)', '$(PackageValidationBaselineName.ToLower())', '$(PackageValidationBaselineVersion)', '$(PackageValidationBaselineName.ToLower()).$(PackageValidationBaselineVersion).nupkg'))</PackageValidationBaselinePath>
-      <EnablePackageBaselineValidation Condition="'$(PackageValidationBaselinePath)' == ''">false</EnablePackageBaselineValidation>
```
ericstj, line 24: Per @ViktorHofer's feedback in email we can set `EnablePackageBaselineValidation` to true when it is not set and either `PackageValidationBaselineVersion` or `PackageValidationBaselinePath` is set.

Basically minimize the number of things that the user needs to set in order to opt-into the comparison.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -3,27 +3,22 @@
...
+
+    <Error Condition="'$(EnablePackageBaselineValidation)' == 'true' and '$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' == ''" Text="Please set PackageValidationBaselineVersion or PackageValidationBaselinePath property and restore the project again."/>
```
ViktorHofer, line 14: Now that we aren't calculating the basline version anymore, why can't this error be in the `ValidatePackage` task instead?

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -3,27 +3,22 @@
...
+
+    <PropertyGroup >
+      <PackageValidationBaselinePath Condition="'$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">$([MSBuild]::NormalizePath('$(NuGetPackageRoot)', '$(PackageId.ToLower())', '$(PackageValidationBaselineVersion)', '$(PackageId.ToLower()).$(PackageValidationBaselineVersion).nupkg'))</PackageValidationBaselinePath>
     </PropertyGroup>
 
+    <Error Condition="'$(EnablePackageBaselineValidation)' == 'true' and !Exists('$(PackageValidationBaselinePath)')" Text="$(PackageValidationBaselinePath) does not exist. Please check the PackageValidationBaselinePath or PackageValidationBaselineVersion." />
```
ViktorHofer, line 20: Why do we need this error? Shouldn't the underlying IO call that loads the package be responsible for throwing the error? What about cases where the package is deleted (for whatever reason) between this read operation and the read operation which later happens in the task?

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -3,27 +3,22 @@
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.GetLastStablePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
```
ViktorHofer, line 4: Remove this UsingTask

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -35,7 +33,7 @@ protected override void ExecuteCore()
 
             new CompatibleTfmValidator(NoWarn, null, RunApiCompat, logger).Validate(package);
             new CompatibleFrameworkInPackageValidator(NoWarn, null, logger).Validate(package);
-            if (BaselineValidation)
+            if (!string.IsNullOrEmpty(BaselinePackageTargetPath))
```
joperezr, line 36: Since we also allow the user to set this (as opposed to just using the version) we should validate that the path actually exists before we call CreatePackage. I'm fine with doing that check either here or in the targets file, but we should do it at some point in order to provide diagnostics in case the path doesn't exist

Anipik: Nuget parser will do that for us while loading it and throw that the package doesnot exist.

Anipik: https://github.com/dotnet/sdk/pull/17880#discussion_r639905957

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,27 +1,17 @@
...
-    <!-- For a non static graph restore with a single tfm, hook onto CollectPackageDownloads which will then be invoked just once. -->
-    <AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets Condition="'$(AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets)' == '' and '$(TargetFrameworks)' == '' and '$(TargetFramework)' != ''">CollectPackageDownloads</AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets>
-    <!-- For a non static graph restore with multiple tfms, hook onto the target that fans out to the individual tfm restore graph walk operations. -->
-    <AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets Condition="'$(AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets)' == ''">_GenerateProjectRestoreGraphAllFrameworks</AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets>
-  </PropertyGroup>
+  <ItemGroup Condition="'$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
+    <PackageDownload Include="$(PackageId)" Version="[$(PackageValidationBaselineVersion)]" />
```
joperezr, line 6: I wonder if it's worth checking if the user for whatever reason added PackageDownload for that packageId already. Also, should we allow folks to specify a different packageID? For example, what happens if someone renamed ther package across releases, but still wants to perform validation to make sure their new package supports all of the previous frameworks and is api compatible?

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
joperezr: Have we tested how that looks for the user? If it is just an exception from the nuget parser with a callstack, then I'd argue that is not a great experience and instead opt to catch that case early and provide an error message with a human-readable message telling them what is wrong. If NuGet parser already throws a message instead of a callstack, then I'm fine.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: > Also, should we allow folks to specify a different packageID? For example, what happens if someone renamed ther package across releases, but still wants to perform validation to make sure their new package supports all of the previous frameworks and is api compatible?

They can use PackageValidationBaselinePath property to use against versions of different name.


> I wonder if it's worth checking if the user for whatever reason added PackageDownload for that packageId already.

I don't think we see any performance degradation or adding extra network calls. So i think we can avoid warning user about it.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
Anipik: it returns with the call stack. personally i prefer the call stack as it gives some more info.

ViktorHofer: I don't have a strong opinion here but avoiding an unnecessary read operation that will performed soon after sounds better IMO. I'll defer to @ericstj in this case.

joperezr: In the past in similar areas, we have opted to use detailed error messages as opposed to call stacks as most of the time that includes noise which is not super actionable and might confuse the user where the real thing we want to point out is just "The path that you passed in for us to check against the baseline doesn't exist".

If you feel strongly about keeping the callstack I'm fine, but in general I think it is better to have a human-readable message with less noise as that tends to be easier to take an action on.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
ViktorHofer: I think being able to specify a different PackageId when using PackageDownload does make sense and we already had that feature before. My suggestion would be to add that part of the code back that defines `PackageValidationBaselineName` and default it to PackageId.

joperezr: >They can use PackageValidationBaselinePath property to use against versions of different name.

Good point.

> I don't think we see any performance degradation or adding extra network calls. So i think we can avoid warning user about it.

I wasn't too concerned about that, but instead more about what happens if the PackageDownload specified by the user is a different version than the one we are using, will that still work?

ViktorHofer: > I wasn't too concerned about that, but instead more about what happens if the PackageDownload specified by the user is a different version than the one we are using, will that still work?

Yes, PackageDownload is specifically designed for that scenario which is being able to restore different versions of the same package.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
joperezr: >I don't have a strong opinion here but avoiding an unnecessary read operation that will performed soon after sounds better IMO.

If what we want is to avoid the read, then we could always just add a try- catch block on the call to CreatePackage, and catch the exception and provide a better message in case the file is not found.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: > My suggestion would be to add that part of the code back that defines PackageValidationBaselineName and default it to PackageId

What are the cases where we want packageValidationBaselineName to be different from packageID ? For jose scenario where a user decides to change names between releases, they can directly use PackageValidationBaselinePath

joperezr: You can male the same argument for the baseline version number. The idea of providing a way to override the package name is just convenience, so user doesn’t have to do packagedownload themselves and construct the path when we can do that for them.

Anipik: > . The idea of providing a way to override the package name is just convenience, so user doesn’t have to do packagedownload themselves and construct the path when we can do that for them.

we already have the packageId for most of the cases where we want to override the packageName. A person can set packageId if they want.

ViktorHofer: > where a user decides to change names between releases, they can directly use PackageValidationBaselinePath

How would that work in CI? Imagine we rename a package in dotnet/runtime but still would like to validate against the baseline? How would we accomplish that in CI then?

Anipik: > You can make the same argument for the baseline version number. 

Version is different in the sense we can have different versions that a user might want to test against.

joperezr: I think @ViktorHofer suggestion of allowing users to optionally override package baseline id is a good one we should take, but that is also a new feature which is not exactly tied to this PR so I'm fine with adding that later and wouldn't want to block this PR on that discussion so I'm fine with the way this code is for now, and we should just consider adding the package id override in a subsequent PR.

Anipik:
cc @ViktorHofer @ericstj  @joperezr can you guys review this one ?

joperezr (approved):

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,97 +1,31 @@
...
-      <PackageValidationBaselinePath Condition="'$(PackageValidationBaselineVersion)' != ''">$([MSBuild]::NormalizePath('$(NuGetPackageRoot)', '$(PackageValidationBaselineName.ToLower())', '$(PackageValidationBaselineVersion)', '$(PackageValidationBaselineName.ToLower()).$(PackageValidationBaselineVersion).nupkg'))</PackageValidationBaselinePath>
-      <EnablePackageBaselineValidation Condition="'$(PackageValidationBaselinePath)' == ''">false</EnablePackageBaselineValidation>
+
+    <PropertyGroup >
+      <PackageValidationBaselinePath Condition="'$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">$([MSBuild]::NormalizePath('$(NuGetPackageRoot)', '$(PackageValidationBaselineName.ToLower())', '$(PackageValidationBaselineVersion)', '$(PackageValidationBaselineName.ToLower()).$(PackageValidationBaselineVersion).nupkg'))</PackageValidationBaselinePath>
     </PropertyGroup>
 
+    <Error Condition="!Exists('$(PackageValidationBaselinePath)')" Text="$(PackageValidationBaselinePath) does not exist. Please check the PackageValidationBaselinePath or PackageValidationBaselineVersion." />
```
ViktorHofer, line 21: This means that you can never use the Microsoft.DotNet.PackageValidation without the baseline feature. Is that intentional?

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,97 +1,31 @@
...
-    <!-- For a non static graph restore with a single tfm, hook onto CollectPackageDownloads which will then be invoked just once. -->
-    <AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets Condition="'$(AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets)' == '' and '$(TargetFrameworks)' == '' and '$(TargetFramework)' != ''">CollectPackageDownloads</AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets>
-    <!-- For a non static graph restore with multiple tfms, hook onto the target that fans out to the individual tfm restore graph walk operations. -->
-    <AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets Condition="'$(AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets)' == ''">_GenerateProjectRestoreGraphAllFrameworks</AddPackageValidationBaselinePackageAsPackageDownloadBeforeTargets>
+    <PackageValidationBaselineName>$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))</PackageValidationBaselineName>
```
ViktorHofer, line 6: Is PackageId already defined when this property is evaluated? I see it defined here https://github.com/NuGet/NuGet.Client/blob/6aae596beda5e60106e0a6475124bfb85a49caef/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L30 but can't remember what is imported first.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,97 +1,31 @@
...
 
+    <Error Condition="!Exists('$(PackageValidationBaselinePath)')" Text="$(PackageValidationBaselinePath) does not exist. Please check the PackageValidationBaselinePath or PackageValidationBaselineVersion." />
+
     <!-- PackageTargetPath isn't exposed by NuGet: https://github.com/NuGet/Home/issues/6671. -->
     <Microsoft.DotNet.PackageValidation.ValidatePackage
       PackageTargetPath="$([MSBuild]::ValueOrDefault('$(PackageTargetPath)', '$(PackageOutputPath)$(PackageId).$(PackageVersion).nupkg'))"
       RuntimeGraph="$(RuntimeIdentifierGraphPath)"
       NoWarn="$(NoWarn)"
       RunApiCompat="$([MSBuild]::ValueOrDefault('$(RunApiCompat)', 'true'))"
-      BaselineValidation="$(EnablePackageBaselineValidation)"
```
ViktorHofer, line 33: We currently set this property in dotnet/runtime to disable validation. How will this work without this property?

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: PackageId is set first and imported.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: we wont set PackageValidationBaselineVersion and PackageValidationBaselineFilePath for those projects.

ViktorHofer: Sounds like a project author who creates a new package needs to unset the version i.e. this way: `<PackageValidationBaselineVersion />`. Do you believe that's the right user experience? Curious on your thoughts here .

Anipik: > Sounds like a project author who creates a new package needs to unset the version i.e. this way:

no he doesnt need to unset, because we are not setting it anywhere to a default value. we only set PackageValidationBaselineName to packageId.

ViktorHofer: Not following. I thought we would set PackageValidationBaselineVersion to a default value in dotnet/runtime. How would you disable package baseline validation in a specific project?

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
ericstj: Agree, we shouldn't block using this feature for brand-new packages.  This error should be constrained to only apply when the `PackageValidationBaselinePath` is non empty.

I wonder if we can come up with a heuristic to lead folks who use this package to discover that they need to specify the baseline while not getting in the way of folks building the first version of a package.

I could imagine such a heuristic looking like a suppressible warning when someone is building a package that is not version `1.0.0` that tells folks to specify a baseline.

joperezr: Great catch, I missed this in my review, we should add the extra condition.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: I was replying for a customer facing standalone project. A customer using this in his project and who doesnt want to run the baseline validation would not need to unset any variable.

yes for runtime repo, we default it 5.0.0. we are able to do that because our package versions are mostly synced.
 we will probably need to do some gymnastics to support where we dont want to run baseline validation.

But i think we should not add code to this customer feature which is specific to the runtime.

The way i imagine our customer scenarios are
- single standalone projects.
- multiple projects where probably the package versions (last released package versions) are not synced up.

Thats my thoughts :P

ViktorHofer: > But i think we should not add code to this customer feature which is specific to the runtime.

You mean specific to repositories which use the same `PackageVersion` for their libraries. I'm sure that's not just dotnet/runtime.

> we will probably need to do some gymnastics to support where we dont want to run baseline validation.

With the current change, yes. That's why I'm asking if you think it's a good user experience in dotnet/runtime and similar repositories to unset the `PackageValidationBaselineVersion` or if `DisablePackageBaselineValidation` should be supported as well as an option.

PR #17894, opened 2021-05-26
Allowing workload update and list without environment variable

sfoslund:


sfoslund:
@marcpopMSFT right now workload list and update commands are behind an environment variable. I assume we want to remove that now that they're implemented in prev 5?

marcpopMSFT:
Yes please.

PR #17916, opened 2021-05-27
MSBuild server V1

### LINKED ISSUE #17308: MSBuild Service - using MSBULDNOINPROCNODE env var
Part of .NET 6.0 user story dotnet/msbuild#5873 as issue dotnet/msbuild#6392

Use existing MSBULDNOINPROCNODE escape hatch to improve consequent and incremental build scenarios.

Entry node will not compile any projects and will use reusable msbuild node(s) for it. 
This way recompile or incremental compile might happen with hot caches resulting in considerably faster build time.

rokonec:
To improve inner loop experience of building small project fast in scenarios:

dotnet new console
dotnet build/run
We force MSBuild to always use external reusable nodes/processes to build projects.

Measured improvement: #6467

Implementation:

Modify SDK to set MSBUILDNOINPROCNODE = 1
Implement escape hatches DOTNET_DO_NOT_USE_MSBUILDNOINPROCNODE = 1

Also see: https://github.com/dotnet/msbuild/issues/6494

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
@@ -100,6 +100,8 @@ namespace Microsoft.DotNet.Cli.Utils
 
             if (value == string.Empty || value == "\0")
             {
+                // Do not allow MSBuild NOIPROCNODE as null env vars are not properly transferred to build nodes
+                _msbuildRequiredEnvironmentVariables["MSBUILDNOINPROCNODE"] = "0";
                 // Unlike ProcessStartInfo.EnvironmentVariables, Environment.SetEnvironmentVariable can't set a variable
                 // to an empty value, so we just fall back to calling MSBuild out-of-proc if we encounter this case.
                 // https://github.com/dotnet/runtime/issues/50554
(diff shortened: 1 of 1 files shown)


src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -16,6 +16,7 @@ namespace Microsoft.DotNet.Cli.Utils
     internal class MSBuildForwardingAppWithoutLogging
     {
         private static readonly bool AlwaysExecuteMSBuildOutOfProc = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_RUN_MSBUILD_OUTOFPROC");
+        private static readonly bool DoNotUseMSBUILDNOINPROCNODE = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_DO_NOT_USE_MSBUILDNOINPROCNODE");
```
ladipro, line 19: nit: Maybe remove the double negation: `DOTNET_CLI_USE_MSBUILD_INPROC_NODE` ?

ladipro (approved):

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
rokonec: I was trying to keep name same as is the MSBUILD escape hatch toi avoid confusion between your OUTOFPROC and mine NOINPROC.  People might get impression that OUTOFPROC == NOINPROC and get confused why we have two escapes for same thing. Ithough about `DOTNET_CLI_DONT_USE_MSBUILD_SERVER_V1` but I believe it is confusing for anyone without knowledge of msbuild server background....

PR #17926, opened 2021-05-27
Wire up some of the MSI installer

joeloff:
- Update the installer factory to create the MSI workload installer
- Add workload installation record management for MSI installer + unit tests


src/Cli/dotnet/commands/dotnet-workload/InstallType.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
sfoslund, line 1: This file probably belongs alongside the other common install code. So far I've been putting all of it under dotnet-workload/install, but if we want to move it to another common dir, that works too (also doesn't have to be done in this PR).

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -20,7 +22,32 @@ internal class WorkloadInstallerFactory
             string dotnetDir = null, 
             PackageSourceLocation packageSourceLocation = null)
         {
+            var installType = GetWorkloadInstallType(sdkFeatureBand);
+
+            if (installType == InstallType.Msi && OperatingSystem.IsWindows())
```
sfoslund, line 27: Should we error if the install type is MSI and the OS isn't windows? It seems like we've gotten into a bad state if that's the case.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -20,7 +22,32 @@ internal class WorkloadInstallerFactory
...
         }
+
+        /// <summary>
+        /// Determines the <see cref="InstallType"/> associated with a specific SDK version.
+        /// </summary>
+        /// <param name="sdkFeatureBand">The SDK version to check.</param>
+        /// <returns>The <see cref="InstallType"/> associated with the SDK.</returns>
+        public static InstallType GetWorkloadInstallType(SdkFeatureBand sdkFeatureBand)
+        {
+            string installerTypePath = Path.Combine(Path.GetDirectoryName(Environment.ProcessPath), "metadata",
```
sfoslund, line 42: Can we use the dotnetDir parameter [here](https://github.com/dotnet/sdk/pull/17926/files#diff-30590c656d6aa0fc5b561cb5ef14bd838f7ade7aa2e7c008cda0aa2cf4f79c5aR22) instead of checking the process path? This will make testing easier

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -20,7 +22,32 @@ internal class WorkloadInstallerFactory
...
+                "workloads", $"{sdkFeatureBand}", "installertype");
```
sfoslund, line 43: nit: `$"{sdkFeatureBand}"` is the same as `sdkFeatureBand`?

src/Tests/dotnet-workload-list.Tests/GivenAnMsiInstallation.cs
```diff
@@ -0,0 +1,88 @@
...
+                CreateWorkloadRecord("6.0.100", "workload.B");
+                CreateWorkloadRecord("6.0.200", "workload.C");
+
+                var records = RecordManager.GetInstalledWorkloads(new SdkFeatureBand("6.0.200"));
+
+                Assert.Contains(new WorkloadId("workload.C"), records);
+            }
+            finally
+            {
+                Registry.CurrentUser.DeleteSubKeyTree(@"SOFTWARE\Microsoft\dotnet-test");
```
sfoslund, line 35: This is to clean up the registry, right? Isn't there a built in way to add a dispose method to run after tests?

src/Tests/dotnet-workload-list.Tests/GivenAnMsiInstallation.cs
```diff
@@ -0,0 +1,88 @@
...
+    [Collection("MsiWorkloadRecords")]
+    public class GivenAnMsiInstallation
+    {
+        // Override HKLM to HKCU so we can run tests without needing elevation
+        private static MsiWorkloadInstallationRecordManager RecordManager = new MsiWorkloadInstallationRecordManager(
+            Registry.CurrentUser,
+            @"SOFTWARE\Microsoft\dotnet-test\InstalledWorkloads\Standalone");
+
+        [WindowsOnlyFact]
+        public void ItCanReadWorkloadRecords()
```
sfoslund, line 21: nit: We've been using the `Given... it...` test name pattern for all our workload tests so far

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
joeloff: Is it? The interpolation does an implicit ToString(), otherwise I need to do sdkFeatureBand.ToString() to convert it I think

src/Tests/dotnet-workload-list.Tests/GivenAnMsiInstallation.cs
joeloff: I'll fix this up

src/Tests/dotnet-workload-list.Tests/GivenAnMsiInstallation.cs
joeloff: you mean in the SDK test framework? I don't know, I'll look around

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
joeloff: Yes, we should

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
joeloff: Right now it should just fall through to the file based installer. I did have code for it and then removed it, but we could throw something like an InvalidOperationException. This would be a weird state where someone had to create the sentinel file themselves on a non-Windows OS.

I can revive the code

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
sfoslund: Oh gotcha, I'm not sure.

src/Tests/dotnet-workload-list.Tests/GivenAnMsiInstallation.cs
sfoslund: @wli3 might know?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
sfoslund: Personally I think it would be better to explicitly fail if we're in that state, but I don't have strong feelings.

joeloff: I don't either :) But I'll add the message back for now.

src/Tests/dotnet-workload-list.Tests/GivenAnMsiInstallation.cs
joeloff: Looks like I can just implement IDisposable based on what I read

src/Cli/dotnet/commands/dotnet-workload/InstallType.cs
joeloff: I'll probably leave it where it is for now until I get the rest in place and see what it looks like. I might very well end up with a whole separate Windows folder

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -20,7 +22,37 @@ internal class WorkloadInstallerFactory
             string dotnetDir = null, 
             PackageSourceLocation packageSourceLocation = null)
         {
+            var installType = GetWorkloadInstallType(sdkFeatureBand, string.IsNullOrWhiteSpace(dotnetDir) ?  Environment.ProcessPath : dotnetDir);
```
sfoslund, line 25: Environment.ProcessPath -> Path.GetDirectoryName(Environment.ProcessPath)

sfoslund (approved):

PR #17930, opened 2021-05-27
[release/6.0-preview5] Add support for Pgo Mibc files to be used by the SDK for Preview5

davidwrighton:
In order to upgrade the RuntimeList.xml file schema for building the RuntimeList.xml in the runtime repo, we must update the SDK to be able to consume them. (The SDK update must happen before the runtime repo can generate the new data as the runtime repo has a testbed which depends on the LKG of the SDK to be able to process existing files.)

Pgo Mibc file consumption from within a RuntimeList.xml file is necessary to enable end customers to take advantage our our static pgo system when producing trimmed, R2R'd published self-contained applications, or self-contained R2R composite images.

## Customer Impact
By adding this work to Preview5, we will be able to expose static pgo data to consumers in long-term serviceable manner in Preview7. In addition, this will enable customers in Preview5 to use their own locally generated static pgo data. However, as we do not intend to advertise that feature in Preview5, it is not expected that customers will see any impact.

## Testing
Local ad-hoc testing of runtime repo.

## Risk
Minimal. No new logic will activate until the runtime produces a RuntimeList.xml with the new content.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolveRuntimePackAssets.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolveRuntimePackAssets.cs
@@ -118,6 +118,10 @@ namespace Microsoft.NET.Build.Tasks
                 {
                     assetType = "native";
                 }
+                else if (typeAttributeValue.Equals("PgoData", StringComparison.OrdinalIgnoreCase))
+                {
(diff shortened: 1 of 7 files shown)


jeffschwMSFT:
Adding @marcpopMSFT

trylek (approved):
LGTM, thank you!

PR #17933, opened 2021-05-27
Implement cross feature band workload update

sfoslund:


src/Tests/dotnet-workload-install.Tests/MockPackWorkloadInstaller.cs
```diff
@@ -113,6 +114,9 @@ public IEnumerable<WorkloadId> GetInstalledWorkloads(SdkFeatureBand sdkFeatureBa
             return InstalledWorkloads;
         }
 
-        public IEnumerable<SdkFeatureBand> GetFeatureBandsWithInstallationRecords() => throw new NotImplementedException();
+        public IEnumerable<SdkFeatureBand> GetFeatureBandsWithInstallationRecords()
+        {
+            return Array.Empty<SdkFeatureBand>();
```
joeloff, line 119: Consider using Enumerable.Empty<>

src/Tests/dotnet-workload-install.Tests/MockPackWorkloadInstaller.cs
```diff
@@ -7,6 +7,7 @@
 using static Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadResolver;
 using Microsoft.DotNet.Workloads.Workload.Install.InstallRecord;
 using Microsoft.Extensions.EnvironmentAbstractions;
+using System.Linq;
 
 namespace Microsoft.DotNet.Cli.Workload.Install.Tests
```
wli3, line 12: We should test and handle (better error message) when older sdk cannot read higher version manifest file

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -124,7 +124,7 @@ public void UpdateWorkloads(bool includePreviews = false, DirectoryPath? offline
             _reporter.WriteLine();
             var featureBand = new SdkFeatureBand(string.Join('.', _sdkVersion.Major, _sdkVersion.Minor, _sdkVersion.SdkFeatureBand));
 
-            var workloadIds = _workloadInstaller.GetWorkloadInstallationRecordRepository().GetInstalledWorkloads(featureBand);
+            var workloadIds = GetUpdatableWorkloads();
             _workloadManifestUpdater.UpdateAdvertisingManifestsAsync(includePreviews, offlineCache).Wait();
             var manifestsToUpdate = _workloadManifestUpdater.CalculateManifestUpdates();
```
wli3, line 129: Should advertising manifest use the high feature band too? We should use advertising manifest at all?

src/Tests/dotnet-workload-install.Tests/MockPackWorkloadInstaller.cs
sfoslund: Can you clarify the scenario you're talking about? I don't think we need to worry about an older SDK reading a newer manifest. The way it's implemented right now, we only ever look at the current feature band's manifests. The steps are basically figure out which workloads are installed with the prior feature band (using workload install records) and then determine which packs we need for those workloads based on the current band's manifests.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
sfoslund: Just double checked, the WorkloadManifestUpdater uses whatever feature band is used by the manifest provider, which is always the current version/ version provided the SDK version option, which should be correct here.

src/Tests/dotnet-workload-install.Tests/MockPackWorkloadInstaller.cs
sfoslund: Just pushed test coverage/ better checks for valid sdk-version option values. Is this along the lines of what you were thinking?

sfoslund:
@wli3 @dsplaisted @joeloff I've implemented the `--from-previous-sdk` option we discussed earlier today and this should be caught up on feedback. Mind doing another review pass?

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -38,10 +37,8 @@ internal class WorkloadListCommand : CommandBase
             _machineReadableOption = result.ValueForOption<bool>(WorkloadListCommandParser.MachineReadableOption);
             _verbosity = result.ValueForOption<VerbosityOptions>(WorkloadListCommandParser.VerbosityOption);
 
-            _sdkVersion = string.IsNullOrEmpty(result.ValueForOption<string>(WorkloadListCommandParser.VersionOption)) ?
-                new ReleaseVersion(version ?? Product.Version) :
-                new ReleaseVersion(result.ValueForOption<string>(WorkloadListCommandParser.VersionOption));
             var dotnetPath = Path.GetDirectoryName(Environment.ProcessPath);
+            _sdkVersion = WorkloadInstallCommand.GetValidatedSdkVersion(result.ValueForOption<string>(WorkloadUpdateCommandParser.VersionOption), version, dotnetPath);
```
wli3, line 41: nit: If the method is static, there is no harm reaching to a different class. But for the code organization point of view maybe create a separate class, or make it an extension method to sdkverion. Or even create a SDKVersion class if you want to do very OO style

wli3 (approved):
Sorry for the delay :p

Issue #17645, opened 2021-05-15
Could not copy the file deps.json because it was not found. Microsoft.Common.CurrentVersion.targets 4919


komdil:
I updated VS 2019 to 16.9.4 version. Sometimes I am getting three errors on rebuilding solution:

Could not copy the file deps.json because it was not found. Microsoft.Common.CurrentVersion.targets 4919
Could not copy the file runtimeconfig.json because it was not found. Microsoft.Common.CurrentVersion.targets 4919 
Could not copy the file runtimeconfig.dev.json because it was not found. Microsoft.Common.CurrentVersion.targets 4919

To reprodue this issue, 
get this project:  https://github.com/komdil/MyProject/tree/reproduceSDKIssue
Rebuild twice

I created issue on Visual Studio support: https://developercommunity.visualstudio.com/t/could-not-copy-the-file-depsjson-because-it-was-no/1400396
They mantion that it is started happening after this PR: https://github.com/dotnet/sdk/pull/14488

For workarround I removed these lines from Microsoft.NET.Sdk.targets:

```
    <ItemGroup Condition="'$(GenerateDependencyFile)' == 'true'">
      <AllItemsFullPathWithTargetPath Include="$(ProjectDepsFilePath)"
                                      TargetPath="$([System.IO.Path]::GetFileName($(ProjectDepsFilePath)))"
                                      CopyToOutputDirectory="PreserveNewest" />
    </ItemGroup>
    
    <ItemGroup Condition=" '$(GenerateRuntimeConfigurationFiles)' == 'true'">
      <AllItemsFullPathWithTargetPath Include="$(ProjectRuntimeConfigFilePath)"
                                TargetPath="$([System.IO.Path]::GetFileName($(ProjectRuntimeConfigFilePath)))"
                                CopyToOutputDirectory="PreserveNewest" />
      <AllItemsFullPathWithTargetPath Include="$(ProjectRuntimeConfigDevFilePath)"
                                TargetPath="$([System.IO.Path]::GetFileName($(ProjectRuntimeConfigDevFilePath)))"
                                CopyToOutputDirectory="PreserveNewest" />
    </ItemGroup>
```

komdil:
@dsplaisted It will be good if you give us any workarround. It is slowing down our productivity. Removing lines from Microsoft.NET.Sdk.targets works but I am not sure it is good solution. Can it affect to other places?

dsplaisted:
Hi @komdil,

Let me try to explain what is happening.

[...]

PR #14488, opened 2020-11-10
Fix references to Exe projects

dsplaisted:
Fix #1675 

Include runtimeconfig.json and deps.json in files that are copied from a referenced Exe project.

Note that I also included a refactoring of how you can create test PublishCommands, which touched a lot of files.  So it should be easier to review this commit by commit.

Currently, it still fails if a self-contained project references a framework-dependent project, or vice versa.  When a self-contained project references a framework-dependent project, running the referenced project fails with the following error:

> It was not possible to find any compatible framework version
> The framework 'Microsoft.NETCore.App', version '5.0.0' was not found.
>  - No frameworks were found.
>
> You can resolve the problem by installing the specified framework and/or SDK.

I suspect this may be because some of the host files (hostfxr or something) are different between self-contained and framework-dependent apps, and are interfering with each other.  @vitek-karas @agocke What do you think?

When a framework-dependent project references a self-contained project, it fails because the runtime pack assets aren't copied.  However, if we were to fix that I think we might run into the same issues that we get when referencing a framework-dependent project from a self-contained project.

@jaredpar @ericstj @ViktorHofer

PR #17943, opened 2021-05-28
Could not copy the file deps.json because it was not found. Microsoft.Common.CurrentVersion.targets

komdil:
This PR fixes issue: #17645

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -292,7 +292,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Target Name="AddDepsJsonAndRuntimeConfigToCopyItemsForReferencingProjects"
           BeforeTargets="GetCopyToOutputDirectoryItems;_GetCopyToOutputDirectoryItemsFromThisProject"
-          Condition="'$(HasRuntimeOutput)' == 'true'">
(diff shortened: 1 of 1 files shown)


BenVillalobos:
When investigating https://developercommunity.visualstudio.com/t/Could-not-copy-the-file-SomeFiledepsj/1380752 this solution actually resolved the issue. Not sure if it's the right path or why this PR is failing, but thought it was worth mentioning.

komdil:
@BenVillalobos  That workaround does not work on my case. In my opinion, there should be another solution because it looks like a bug. It is working when I rebuild once, when I rebuild twice it is failing. I think there should be a small change somewhere, but I don't know where. 
Can someone review that PR: #14488. That PR fixes issue related to publishing project: #1675
Maybe we should add publish conditions or add another condition there?

komdil:
@BenVillalobos  any comments?

BenVillalobos:
@komdil I'm not familiar with the details of applying `_UseBuildDependencyFile`, but applying `UseCommonOutputDirectory` (as suggested in the original issue) worked as well and that seemed ideal for the scenario of multiple projects pointing to the same output directory.

komdil:
> @komdil I'm not familiar with the details of applying `_UseBuildDependencyFile`, but applying `UseCommonOutputDirectory` (as suggested in the original issue) worked as well and that seemed ideal for the scenario of multiple projects pointing to the same output directory.

Unfortunately, `UserCommandOutputDirectory` not helped. I want to know why this issue is not marked as a bug. The reason is I am available to rebuild my solution only once and getting errors on rebuilding twice.  If my project configurations are wrong, why I am not getting these errors on the first time rebuilding?

PR #17956, opened 2021-05-28
Use _GlobalPropertiesToRemoveFromProjectReferences GetCopyToPublishDirectoryItems to exclude global properties from P2P project references during publish

abpiskunov:
Use _GlobalPropertiesToRemoveFromProjectReferences GetCopyToPublishDirectoryItems to exclude global properties from P2P project references during publish


vijayrkn:
Adding @dsplaisted & @wli3 since this target is owned by the SDK team.

vijayrkn (approved):

abpiskunov:
@dsplaisted i have test failure in on of the configurations which fails to connect to nuget feed - do you see it in other PRs - is it known?

PR #17982, opened 2021-05-31
Don't omit Microsoft.NETCore.App from list of frameworks in runtimeconfig

dsplaisted:
In .NET Core 3.0, we added the ability to have multiple shared frameworks, which are listed in the runtimeconfig.json file.  However, due to [issues in the host](https://github.com/dotnet/runtime/issues/3402), we omitted the base Microsoft.NETCore.App framework from the list in the runtimeconfig if there were any other frameworks.

The bugs in the host have long since been fixed (as far as I know).  However, omitting the Microsoft.NETCore.App entry from the frameworks was causing issues in dotnet/sdk, where we override the version of the base framework.  Projects that used a higher-level framework such as Microsoft.AspNetCore.App were loading the wrong version of Microsoft.NETCore.App, as the version compiled against wasn't listed in the runtimeconfig.

PR where we hit this: #17948
Commit with workaround: 150c6fa0aae7347be7e3a452492faad5ddb34f7e

This PR currently changes the logic so that Microsoft.NETCore.App will by default always be listed in the runtimeconfig when targeting .NET 6 or higher.  In any case the `AlwaysIncludeCoreFrameworkInRuntimeConfig` property can be used to override the default behavior.

I think the issues were fixed before .NET Core 3.0 released, so we might be able to always make the default to list all frameworks if the breaking change risk is low.

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenFrameworkReferences.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenFrameworkReferences.cs
@@ -40,7 +40,7 @@ namespace FrameworkReferenceTest
             var testProject = new TestProject()
             {
                 Name = "MultipleFrameworkReferenceTest",
-                TargetFrameworks = "netcoreapp3.0",
+                TargetFrameworks = "net6.0",
                 IsExe = true
             };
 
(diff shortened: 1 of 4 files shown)


vitek-karas (approved):
As far as I can tell the bug has been fixed in 3.0 release. So unless we officially support 2.* apps with the .NET 6 SDK it should be OK to always do this without a way to opt-out.

I think we should just change this.

wli3 (approved):

src/Tests/Microsoft.NET.Build.Tests/GivenFrameworkReferences.cs
```diff
@@ -40,7 +40,7 @@ public void Multiple_frameworks_are_written_to_runtimeconfig_when_there_are_mult
             var testProject = new TestProject()
             {
                 Name = "MultipleFrameworkReferenceTest",
```
wli3, line 42: test for lower TFM still having the Microsoft.NETCore.App?

PR #17999, opened 2021-06-01
 Add workload list with available update workloads

wli3:


src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -65,9 +66,13 @@ internal class WorkloadUpdateCommand : CommandBase
             _includePreviews = parseResult.ValueForOption<bool>(WorkloadUpdateCommandParser.IncludePreviewsOption);
             _downloadToCacheOption = parseResult.ValueForOption<string>(WorkloadUpdateCommandParser.DownloadToCacheOption);
             _verbosity = parseResult.ValueForOption<VerbosityOptions>(WorkloadUpdateCommandParser.VerbosityOption);
-            _sdkVersion = string.IsNullOrEmpty(parseResult.ValueForOption<string>(WorkloadUpdateCommandParser.VersionOption)) ?
+            Option versionOption = new Option<string>("--sdk-version", Install.LocalizableStrings.VersionOptionDescription)
+            {
+                ArgumentHelpName = Install.LocalizableStrings.VersionOptionName
+            };
```
sfoslund, line 72: Here too- why do we need to create a new option?

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -5,69 +5,97 @@
...
             _machineReadableOption = result.ValueForOption<bool>(WorkloadListCommandParser.MachineReadableOption);
             _verbosity = result.ValueForOption<VerbosityOptions>(WorkloadListCommandParser.VerbosityOption);
 
-            var sdkVersion = new ReleaseVersion(version ?? Product.Version);
-            var dotnetPath = Path.GetDirectoryName(Environment.ProcessPath);
-            var workloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(dotnetPath, sdkVersion.ToString());
-            var workloadResolver = WorkloadResolver.Create(workloadManifestProvider, dotnetPath, sdkVersion.ToString());
-            _sdkFeatureBand = new SdkFeatureBand(sdkVersion);
+            _dotnetPath = dotnetDir ?? Path.GetDirectoryName(Environment.ProcessPath);
+            ReleaseVersion currentSdkReleaseVersion = new(currentSdkVersion ?? Product.Version);
```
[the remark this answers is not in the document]
wli3: it is few lines below

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -66,9 +67,13 @@ internal class WorkloadInstallCommand : CommandBase
             _downloadToCacheOption = parseResult.ValueForOption<string>(WorkloadInstallCommandParser.DownloadToCacheOption);
             _workloadIds = parseResult.ValueForArgument<IEnumerable<string>>(WorkloadInstallCommandParser.WorkloadIdArgument).ToList().AsReadOnly();
             _verbosity = parseResult.ValueForOption<VerbosityOptions>(WorkloadInstallCommandParser.VerbosityOption);
-            _sdkVersion = string.IsNullOrEmpty(parseResult.ValueForOption<string>(WorkloadInstallCommandParser.VersionOption)) ?
+            Option versionOption = new Option<string>("--sdk-version", LocalizableStrings.VersionOptionDescription)
```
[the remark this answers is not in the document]
wli3: My mistake, a bad inline method caused everything to be inlined

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
```diff
@@ -57,7 +51,10 @@ public static Command GetCommand()
             var command = new Command("install", LocalizableStrings.CommandDescription);
 
             command.AddArgument(WorkloadIdArgument);
-            command.AddOption(VersionOption);
+            command.AddOption(new Option<string>("--sdk-version", LocalizableStrings.VersionOptionDescription)
+            {
+                ArgumentHelpName = LocalizableStrings.VersionOptionName
+            });
```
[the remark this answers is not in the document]
wli3: same, reverting

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
wli3: same, reverting

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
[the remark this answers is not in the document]
sfoslund: Why do we need to separate these? Can't we use the target SDK version and default to `currentSdkVersion ?? Product.Version` is that's not provided?

Edit: Ah, I see you're basically doing that later in the constructor. But why do we have to save both versions as fields then? Also, wouldn't we want to consider the target version here? https://github.com/dotnet/sdk/pull/17999/files#diff-927d9977a1d2302c4b31778c2949410c7afe077ef1ec0b842061241dca72c43fR85

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommandParser.cs
```diff
@@ -13,14 +13,23 @@ internal static class WorkloadListCommandParser
 
         public static readonly Option VerbosityOption = CommonOptions.VerbosityOption();
 
-        public static readonly Option VersionOption = WorkloadUpdateCommandParser.VersionOption;
+        public static readonly Option TargetSdkVersionOption = new Option<string>("--target-sdk-version", Workloads.Workload.Install.LocalizableStrings.VersionOptionDescription)
+        {
```
sfoslund, line 17: Other commands use `--sdk-version`, can we use that here too to be consistent?

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -79,39 +108,36 @@ public override int Execute()
...
+            foreach ((ManifestId _, ManifestVersion existingVersion, ManifestVersion newVersion,
+                Dictionary<WorkloadDefinitionId, WorkloadDefinition> workloads) in manifestsToUpdate)
             {
-                updateList.Add(new UpdateAvailableEntry("6.0.100", "6.0.101",
-                    _mockIosDescription,
-                    "microsoft-ios-sdk-full"));
+                foreach ((WorkloadDefinitionId workloadDefinitionId, WorkloadDefinition workloadDefinition) in
+                    workloads)
+                {
+                    if (installedWorkloads.Contains(new WorkloadId(workloadDefinitionId.ToString())))
```
sfoslund, line 126: Won't we get "updates" advertised that are already installed with this logic? Should we compare manifest versions as well to make sure that this is really an update and not the version we already have installed?

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommandParser.cs
```diff
@@ -12,7 +12,10 @@ internal static class WorkloadUninstallCommandParser
 
         public static readonly Option VerbosityOption = WorkloadInstallCommandParser.VerbosityOption;
         
-        public static readonly Option VersionOption = WorkloadInstallCommandParser.VersionOption;
+        public static readonly Option VersionOption = new Option<string>("--sdk-version", Workloads.Workload.Install.LocalizableStrings.VersionOptionDescription)
```
sfoslund, line 15: Why is this needed? I think this is just a longer version of what we already had

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommandParser.cs
```diff
@@ -34,7 +32,7 @@ public static Command GetCommand()
 
             command.AddOption(ConfigOption);
             command.AddOption(AddSourceOption);
-            command.AddOption(VersionOption);
+            command.AddOption(WorkloadInstallCommandParser.VersionOption);
```
sfoslund, line 35: I think these changes should be reverted too

src/Tests/dotnet-workload-list.Tests/GivenWorkloadInstallerAndWorkloadsInstalled.cs
```diff
@@ -0,0 +1,129 @@
...
+                        new Dictionary<WorkloadDefinitionId, WorkloadDefinition>
+                        {
+                            [new WorkloadDefinitionId("other-manifest-workload")] = new(
+                                new WorkloadDefinitionId("other-manifest-workload"), false,
+                                "other-manifest-workload description",
+                                WorkloadDefinitionKind.Dev, null, null, null)
+                        })
+                };
+
+            // Update workload
```
sfoslund, line 69: nit: old comment

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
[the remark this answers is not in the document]
wli3: > But why do we have to save both versions as fields then?

Don't really need to, but we mostly make commandline parameter a field. Just for the consistency.

I think the bigger issue is putting the right version to SdkDirectoryWorkloadManifestProvider is not a dependency injection concern. It is part of the business logic. So ideally sdkversion should be a parameter for the method call.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommandParser.cs
wli3: `--sdk-version` is too vague, people would think it is the current sdk version.

sfoslund: Can we normalize all the other commands to use that name then? And then we just have to declare it once and then reuse it, like the other options?

wli3: `--sdk-version` is the right thing to use

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
wli3: Talked offline. Updater will do a no op, but manifest will be updated. So still counts as an update.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
sfoslund, line 72: One more leftover version option

sfoslund (approved):

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
wli3: Fixed

PR #18039, opened 2021-06-02
Prevent infinite loops in wasm hot reload

### LINKED ISSUE #33209: dotnet-watch: WebAssembly check that verifies if delta is applied can spin indefinitely
The VerifyDeltaApplied check spins indefinitely if a browser does not establish a websocket connection back to the app. This also means the delta failed to apply which should cause the app to be treated as a rude edit.

pranavkm:
* If the WebSocket connection to the browser failed to be established,
dotnet-watch will loop indefinitely as WaitAsync never gets scheduled. This change
adds a limit to how long the app waits, and also terminates early if no client is currently connected

* In addition, it's possible for the TLS connection to fail to be established even though the dev-certs is installed.
This is particularly apparent when developing in WSL. In such an event, attempt to fallback to the non-tls WS endpoint.

Fixes https://github.com/dotnet/aspnetcore/issues/33209

### DIFF
--- a/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
+++ b/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
@@ -1,10 +1,17 @@
-setTimeout(function () {
+setTimeout(async function () {
   // dotnet-watch browser reload script
+  const webSocketUrls = '{{hostString}}'.split(',');
   let connection;
-  try {
-    connection = new WebSocket('{{hostString}}');
-  } catch (ex) {
-    console.debug(ex);
+  for (const url of webSocketUrls) {
+    try {
+      connection = await getWebSocket(url);
+      break;
+    } catch (ex) {
+      console.debug(ex);
+    }
+  }
+  if (!connection) {
+    console.debug('Unable to establish a conection to the browser refresh server.');
     return;
   }
 
@@ -105,9 +112,9 @@ setTimeout(function () {
       } catch (error) {
         console.warn(error);
         applyFailed = true;
-      } 
+      }
     });
-    
+
     fetch('_framework/blazor-hotreload', { method: 'post', headers: { 'content-type': 'application/json' }, body: JSON.stringify(deltas) })
       .then(response => {
         if (response.status == 200) {
@@ -115,7 +122,7 @@ setTimeout(function () {
           window.sessionStorage.setItem('blazor-webasssembly-cache', { etag, deltas });
         }
       });
-    
+
     if (applyFailed) {
       sendDeltaNotApplied();
     } else {
@@ -158,5 +165,28 @@ setTimeout(function () {
 
   function sendDeltaNotApplied() {
     connection.send(new Uint8Array([0]).buffer);
+  }
+
+  function getWebSocket(url) {
+    return new Promise((resolve, reject) => {
+      const webSocket = new WebSocket(url);
+      function onOpen() {
+        clearEventListeners();
+        resolve(webSocket);
+      }
+
+      function onError(error) {
+        clearEventListeners();
+        reject(error);
+      }
+
+      function clearEventListeners() {
+        webSocket.removeEventListener('open', onOpen);
+        webSocket.removeEventListener('error', onError);
+      }
+
+      webSocket.addEventListener('open', onOpen);
+      webSocket.addEventListener('error', onError);
+    });
   }
 }, 500);
--- a/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
+++ b/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
@@ -170,23 +170,38 @@ setTimeout(async function () {
   function getWebSocket(url) {
     return new Promise((resolve, reject) => {
       const webSocket = new WebSocket(url);
+      let opened = false;
+
       function onOpen() {
+        opened = true;
         clearEventListeners();
         resolve(webSocket);
       }
 
-      function onError(error) {
+      function onClose(event) {
+        if (opened) {
+          // Open completed successfully. Nothing to do here.
+          return;
+        }
+
+        let error = 'WebSocket failed to connect.';
+        if (event instanceof ErrorEvent) {
+          error = event.error;
+        }
+
         clearEventListeners();
         reject(error);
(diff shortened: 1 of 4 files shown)


src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs
```diff
@@ -38,18 +38,25 @@ public BrowserRefreshServer(IReporter reporter)
...
 
-            var useTls = await ShouldUseHttps();
+            var supportsTLS = await SupportsTLS();
 
             _refreshServer = new HostBuilder()
                 .ConfigureWebHost(builder =>
                 {
                     builder.UseKestrel();
-                    builder.UseUrls(useTls ? $"https://{hostName}:0" : $"http://{hostName}:0");
+                    if (supportsTLS)
```
captainsafia, line 52: Do we need to bother with this check if we can fallback on the logic that makes the WebSocket connection to the valid URL on the client-side?

Feels like we can avoid making the additional process call here.

src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs
```diff
@@ -61,23 +68,26 @@ public async ValueTask<string> StartAsync(CancellationToken cancellationToken)
...
+                    .Replace("https://127.0.0.1", "wss://localhost", StringComparison.Ordinal));
             }
 
-            return serverUrl
-                .Replace("https://", "wss://", StringComparison.Ordinal)
-                .Replace("http://", "ws://", StringComparison.Ordinal);
+            return new[]
+            {
+                serverUrls
+                    .First()
```
captainsafia, line 87: Why do we only pick the first URL here?

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
```diff
@@ -72,43 +72,43 @@ public async ValueTask ReportDiagnosticsAsync(DotNetWatchContext context, IEnume
...
-                return IsDeltaApplied(result);
+                    var result = await context.BrowserRefreshServer.ReceiveAsync(_receiveBuffer, cancellationToken);
+                    if (result is null)
+                    {
+                        // A null result indicates no clients are connected. No deltas could have been applied in this state.
+                        _reporter.Verbose("No client is connected to ack deltas");
+                        return false;
+                    }
+
+                    if (IsDeltaApplied(result.Value))
```
BrennanConroy, line 92: nit: return IsDeltaApplied

src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
```diff
@@ -159,4 +166,27 @@ setTimeout(function () {
...
+        clearEventListeners();
+        reject(error);
+      }
+
+      function clearEventListeners() {
+        webSocket.removeEventListener('open', onOpen);
+        webSocket.removeEventListener('error', onError);
+      }
+
+      webSocket.addEventListener('open', onOpen);
```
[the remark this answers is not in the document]
pranavkm: We shouldn't get to line44 until `getWebSocket` is awaited before which point the event listener is removed.

src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs
pranavkm: Kestrel does not like it if you ask it to bind to a TLS port without the certs installed. So we do need to do some due-diligence here.

src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs
pranavkm: Users can only provide one URL via an environment variable. We previously had this .First https://github.com/dotnet/sdk/pull/18039/files#diff-2696395053559822320fce8c67b5b71b00f693255b29906d02c6427be3e9533fL69 here.

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
pranavkm: It's inside a loop though. We want it to return true the first time it succeeds.

BrennanConroy: Ah didn't notice it was in the loop

src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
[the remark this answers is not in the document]
BrennanConroy: Don't use the error event prefer close. The docs aren't great for it, but through experience we've seen that close will always get called whereas error isn't. And if error is called, closed will be called.

You can see in SignalR that we just log in the error handler:
https://github.com/dotnet/aspnetcore/blob/e0139a1f55fa7eb7b7d700e27aa0f9d1a5720041/src/SignalR/clients/ts/signalr/src/WebSocketTransport.ts#L89-L97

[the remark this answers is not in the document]
pranavkm: Aren't you using the error event in the code that you linked to? Is this what I should be using as reference: https://github.com/dotnet/aspnetcore/blob/e0139a1f55fa7eb7b7d700e27aa0f9d1a5720041/src/SignalR/clients/ts/signalr/src/WebSocketTransport.ts#L112-L130?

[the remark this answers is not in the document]
BrennanConroy: > Aren't you using the error event in the code that you linked to?

Logging only.

> Is this what I should be using as reference

Yeah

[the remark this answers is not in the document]
pranavkm: Updated

BrennanConroy (approved):

captainsafia (approved):

captainsafia (approved):

PR #18045, opened 2021-06-02
Allow opting out of Razor Source Generator

pranavkm:
For 6.0, there are some scenarios primarily around other
source generators that do not work well with RSG. This change
introduces makes the UseRazorSourceGenerator an option that can be configured by projects
to opt back in to the rzc based compilation. This acts as a stop-gap until the compiler
can support robust ordering of source generators in a future release.


captainsafia (approved):
Thanks! I assume this worked fine with MVC with no changes?

pranavkm:
I was able to verify MVC, Blazor Wasm (hosted) and Blazor Server apps work just fine with this change. That flag is very well factored 👍🏽

PR #18000, opened 2021-06-01
MSBuild property for the AutoreleasePool feature.

AaronRobinsonMSFT:
This should go into .NET 6 release and is for enabling https://github.com/dotnet/runtime/pull/47592 and https://github.com/dotnet/runtime/pull/52023.

/cc @wli3 @dsplaisted

PR #18050, opened 2021-06-03
Set a consistent default behavior for AutoreleasePool during linking.

AaronRobinsonMSFT:
Missed setting the linker default in https://github.com/dotnet/sdk/pull/18000

/cc @elinor-fung


elinor-fung (approved):
Looks good to me.
cc @sbomer @vitek-karas

sbomer:
Not sure I have all the necessary context, but shouldn't this be done by setting `featuredefault="true"` here instead?https://github.com/dotnet/runtime/blob/911640b3a891f92ff66e9c82ce65f71d203f11a2/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.OSX.xml#L3
This will let it get trimmed without creating a runtimeconfig entry (see https://github.com/mono/linker/blob/main/docs/data-formats.md#conditional-substitutions-and-descriptors).

Usually when the linker defaults match runtime defaults we use that approach. We set linker-specific defaults in illink.targets when the runtime defaults aren't linker-friendly.

AaronRobinsonMSFT:
@sbomer Okay. I submitted a PR for that suggestion and will revert this PR.

PR #18052, opened 2021-06-03
Use task to get around unexpected batching behavior

wli3:
### Description

%(Identity) is not unique in this batching. If there are duplicated relative path or identity,  _PublishItemsOutputGroupOutputs_ would get bad _relativePath_ metadata with semicolons in the middle.

### Customer Impact

If there are duplicated file in _ResolvedFileToPublish_, relativePath metadata will have bad output and later fail the build. This impacted a large customer and this also caused WinUI CI build to fail after upgrading to 16.10.

### Regression?

Yes. By https://github.com/dotnet/sdk/commit/ba1bce0e0c437f0e0305eb12534f9e37ff1c6b5f

### Risk

Low risk.

### Test changes in this PR

Both unit test and functional tests

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/GetPublishItemsOutputGroupOutputs.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/GetPublishItemsOutputGroupOutputs.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.NET.Build.Tasks
+{
+    public sealed class GetPublishItemsOutputGroupOutputs : TaskBase
+    {
+        public ITaskItem[] ResolvedFileToPublish { get; set; }
+        public string PublishDir { get; set; }
+
+        [Output]
+        public ITaskItem[] PublishItemsOutputGroupOutputs { get; private set; }
+
+        protected override void ExecuteCore()
+        {
+            var list = new List<ITaskItem>();
+            if (ResolvedFileToPublish != null)
+            {
+                foreach (var r in ResolvedFileToPublish)
+                {
+                    var newItem = new TaskItem(r.GetMetadata("FullPath"));
+                    r.CopyMetadataTo(newItem);
+                    newItem.SetMetadata(metadataName: MetadataKeys.TargetPath, metadataValue: r.GetMetadata(MetadataKeys.RelativePath));
+                    newItem.SetMetadata(metadataName: "OutputPath", (PublishDir ?? "") + r.GetMetadata(MetadataKeys.RelativePath));
+                    newItem.SetMetadata(metadataName: "OutputGroup", "PublishItemsOutputGroup");
+                    list.Add(newItem);
+                }
+            }
+
+            PublishItemsOutputGroupOutputs = list.ToArray();
+        }
+    }
+}
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
@@ -1169,13 +1169,19 @@ Copyright (c) .NET Foundation. All rights reserved.
     </PublishItemsOutputGroupDependsOn>
   </PropertyGroup>
 
+  <UsingTask TaskName="Microsoft.NET.Build.Tasks.GetPublishItemsOutputGroupOutputs"
+          AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
+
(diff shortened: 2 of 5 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/GetPublishItemsOutputGroupOutputs.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public sealed class GetPublishItemsOutputGroupOutputs : TaskBase
+    {
+        public ITaskItem[] ResolvedFileToPublish { get; set; }
+        public string PublishDir { get; set; }
+
+        [Output]
+        public ITaskItem[] PublishItemsOutputGroupOutputs { get; private set; }
+
+        protected override void ExecuteCore()
```
wli3, line 20: @rainersigwald could you help me double check if the task is 1:1 translation? I think I get most of the edge case by the unit test
```xml
      <PublishItemsOutputGroupOutputs Include="@(ResolvedFileToPublish->'%(FullPath)')"
                                      TargetPath="%(ResolvedFileToPublish.RelativePath)"
                                      IsKeyOutput="%(ResolvedFileToPublish.IsKeyOutput)"
                                      OutputPath="@(ResolvedFileToPublish->'$(PublishDir)%(relativePath)')"
                                      OutputGroup="PublishItemsOutputGroup" />
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -1169,13 +1169,19 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <Output TaskParameter="PublishItemsOutputGroupOutputs" ItemName="PublishItemsOutputGroupOutputs" />
+    </GetPublishItemsOutputGroupOutputs>
+
     <ItemGroup>
-      <PublishItemsOutputGroupOutputs Include="@(ResolvedFileToPublish->'%(FullPath)')"
-                                      TargetPath="%(ResolvedFileToPublish.RelativePath)"
-                                      IsKeyOutput="%(ResolvedFileToPublish.IsKeyOutput)"
-                                      OutputPath="@(ResolvedFileToPublish->'$(PublishDir)%(relativePath)')"
-                                      OutputGroup="PublishItemsOutputGroup" />
       <PublishItemsOutputGroupOutputs Include="$(PublishedSingleFilePath)"
```
wli3, line 1185: And here too. The following itemgroup add for "PublishedSingleFilePath" should work right?

The first part is to calculate non PublishedSingleFile, and that was converted to a task. And there is a following PublishedSingleFilePath calculation untouched. If there is a PublishedSingleFile, it would be added to PublishItemsOutputGroupOutputs

src/Tasks/Microsoft.NET.Build.Tasks/GetPublishItemsOutputGroupOutputs.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            var list = new List<ITaskItem>();
+            if (ResolvedFileToPublish != null)
+            {
+                foreach (var r in ResolvedFileToPublish)
+                {
+                    var newItem = new TaskItem(Path.GetFullPath(r.ItemSpec));
+                    r.CopyMetadataTo(newItem);
+                    newItem.SetMetadata(metadataName: MetadataKeys.TargetPath, metadataValue: r.GetMetadata(MetadataKeys.RelativePath));
+                    newItem.SetMetadata(metadataName: "IsKeyOutput", r.GetMetadata("IsKeyOutput"));
```
rainersigwald, line 30: Doesn't `CopyMetadataTo` do this?

src/Tasks/Microsoft.NET.Build.Tasks/GetPublishItemsOutputGroupOutputs.cs
```diff
@@ -0,0 +1,40 @@
...
+        public ITaskItem[] PublishItemsOutputGroupOutputs { get; private set; }
+
+        protected override void ExecuteCore()
+        {
+            var list = new List<ITaskItem>();
+            if (ResolvedFileToPublish != null)
+            {
+                foreach (var r in ResolvedFileToPublish)
+                {
+                    var newItem = new TaskItem(Path.GetFullPath(r.ItemSpec));
```
rainersigwald, line 27: I think it's preferable to

```suggestion
                    var newItem = new TaskItem(r.GetMetadata("FullPath"));
```

because this will apply the original context of the item to expand the path, rather than relative to current working directory as `Path.GetFullPath` will do. This _rarely_ matters but can be interesting if a non-full-path item is passed between projects.

src/Tasks/Microsoft.NET.Build.Tasks/GetPublishItemsOutputGroupOutputs.cs
rainersigwald: LGTM

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
rainersigwald: Looks fine.

rainersigwald (approved):

PR #18069, opened 2021-06-03
Use task to get around unexpected batching behavior

wli3:
%(Identity) is not unique in this case. Would get ; if there are
duplicated relative path or identity


rainersigwald (approved):

DHowett:
QQ: Was the MSBuild `Distinct` function considered instead of a new task? It seems like it would serve this purpose.

PR #18072, opened 2021-06-03
Support for preview features

pgovind:
Add support for a `<RequiresPreviewFeatures>` property in the csproj. See https://github.com/dotnet/designs/blob/main/accepted/2021/preview-features/preview-features.md for more information.

TLDR:
.NET 6 will introduce preview features. To enable them, the sdk needs to
1.  Support a `<EnablePreviewFeatures>` property
2. Automatically generate an assembly level `RequiresPreviewFeatures` attribute when preview features are turned on.
3. Automatically change language version to preview when preview features are turned on


src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,6 +370,30 @@ public void It_includes_internals_visible_to()
             AssemblyInfo.Get(assemblyPath)["InternalsVisibleToAttribute"].Should().Be("Tests");
         }
 
+        [Fact]
```
pgovind, line 373: I'm having trouble building locally on windows and testing locally on WSL2. 

On WSL2, I'm submitting this command line:
``` csharp
~/sdk/artifacts/bin/redist/Debug/dotnet/dotnet test --filter "Microsoft.NET.Build.Tests.GivenThatWeWantToControlGeneratedAssemblyInfo.It_includes_requires_preview_features*"
```

But that fails with the message that It_includes_requires_preview_features is not found. I inspected the test dll and I do see It_includes_requires_preview_features as one of the unit tests. Anyone know what I'm doing wrong?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,6 +370,110 @@ public void It_includes_internals_visible_to()
...
```
pgovind, line 373: This was originally a theory with 3 test cases, but that seems to fail in CI with the following error:

```
/private/tmp/helix/working/AE680940/w/A5DF08DC/e/testExecutionDirectory/It_includes_r---61EE9334 already exists
```

I'm guessing that is because each test tries to create a tmp directory while running.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
```diff
@@ -114,6 +115,8 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_Parameter1>%(AssemblyMetadata.Identity)</_Parameter1>
         <_Parameter2>%(AssemblyMetadata.Value)</_Parameter2>
       </AssemblyAttribute>
+      <AssemblyAttribute Include="System.Runtime.Versioning.RequiresPreviewFeaturesAttribute" Condition="'$(EnablePreviewFeatures)' == 'true' and '$(GeneratePreviewFeaturesAttribute)' == 'true'">
```
pgovind, line 118: I'm not super familiar with MSBuild or the sdk repo. If this is the wrong spot to make these changes, let me know!

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
dsplaisted: Yes, typically we handle this by passing in the theory parameters as the `identifier` argument to `CopyTestAsset` or similar methods.

pgovind:
Just making it clear that this PR is now ready for review (in case folks were used to seeing the "DO NOT REVIEW" in the title)

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
```diff
@@ -48,6 +48,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RuntimeIdentifiers Condition=" '$(PackAsTool)' == 'true' ">$(RuntimeIdentifiers);$(PackAsToolShimRuntimeIdentifiers)</RuntimeIdentifiers>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(EnablePreviewFeatures)' == 'true'">
```
dsplaisted, line 51: Is it possible to only set `LangVersion` here if it's not already explicitly set, or does this happen after the default is already set in the targets?

```suggestion
  <PropertyGroup Condition="'$(EnablePreviewFeatures)' == 'true' And '$(LangVersion)' == ''">
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,6 +370,110 @@ public void It_includes_internals_visible_to()
...
+            TestPreviewFeatures(true, false);
+        }
+
+        [Fact]
+        public void Requires_preview_features_not_specified()
+        {
+            TestPreviewFeatures(false, false);
+        }
+
+        private void TestPreviewFeatures(bool addPreviewFeatureProperty, bool enablePreviewFeatures)
```
dsplaisted, line 391: Nit: Now that you're passing the right `identifier` in to `CopyTestAsset`, you can convert this back to a `Theory`.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,6 +370,110 @@ public void It_includes_internals_visible_to()
...
+            foreach (var attribute in parameterlessAttributes)
+            {
+                if (attribute.Equals("RequiresPreviewFeaturesAttribute", System.StringComparison.Ordinal))
+                {
+                    contains = true;
+                    break;
+                }
+            }
+
+            var getValuesCommand = new GetValuesCommand(testAsset, "LangVersion", targetFramework: "net6.0");
```
dsplaisted, line 425: Nit: Don't hardcode this both here and when calling `WithTargetFramework`.  Declare it once as a variable or constant and then use it in both places.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
dsplaisted: I'm curious, what ended up being the issue here?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,6 +370,110 @@ public void It_includes_internals_visible_to()
...
+            }
+            else
+            {
+                Assert.True(contains);
+                Assert.True(langVersion.Equals("Preview", System.StringComparison.Ordinal));
+            }
+        }
+
+        [Fact]
+        public void It_doesnt_includes_requires_preview_features()
```
dsplaisted, line 443: Is this a duplicate of the coverage already there in the previous tests?

src/Tests/Microsoft.NET.TestFramework/AssemblyInfo.cs
```diff
@@ -12,6 +12,45 @@ namespace Microsoft.NET.TestFramework
...
+
+                    const ushort prolog = 1; // two-byte "prolog" defined by ECMA-335 (II.23.3) to be at the beginning of attribute value blobs
+                    if (value.ReadUInt16() != prolog || header.Kind != SignatureKind.Method || header.IsGeneric)
+                    {
+                        throw new BadImageFormatException();
+                    }
+
+                    var paramCount = signature.ReadCompressedInteger();
+                    if (paramCount <= 0 || // must have at least 1 parameter
+                        signature.ReadSignatureTypeCode() != SignatureTypeCode.Void) // return type must be void
```
dsplaisted, line 44: The comment in the first line here is no longer correct.  And is the check for void return type correct here?  The previous code was skipping attributes that didn't have parameters OR that weren't void.  Now we want parameterless attributes, but it seems like maybe we still want to skip them if they aren't void returns.  (I don't know the metadata format at all, so I'm not sure.)

dsplaisted (approved):

dsplaisted:
What about the analyzer that's supposed to go along with this.  Is that in progress?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
pgovind: It was. I didn't know if there was a difference between defining a property as false (but not using it) vs not defining it at all. Looks like there isn't, so I removed this unit test.

src/Tests/Microsoft.NET.TestFramework/AssemblyInfo.cs
pgovind: Removed the comment and the void check now

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
pgovind: Ah you're right, we can set it only when `LangVersion` has not been set explicitly. Fixed now.

pgovind:
> What about the analyzer that's supposed to go along with this. Is that in progress?

Yup I'm working on the analyzer locally. I should have it up for review some time this week.

PR #18078, opened 2021-06-03
Workload install elevation error

### LINKED ISSUE #17087: error message and user experience when the user are not elevated
zip install | admin install * windows | mac | linux

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/17087


sfoslund:
I've tested manually but I don't know of a way to add test coverage. Let me know if you have any thoughts.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -36,6 +36,11 @@ internal class WorkloadInstallerFactory
                 return new NetSdkMsiInstaller();
             }
 
+            if (!CanWriteToDotnetRoot(dotnetDir))
```
joeloff, line 39: Don't know about other platforms, but for Windows, UnauthorizedException is typically used since that also sets up the appropriate E_ACCESS_DENIED HResult.

joeloff:
> I've tested manually but I don't know of a way to add test coverage. Let me know if you have any thoughts.

You could try creating a folder using DirectorySecurity and setting access control rules. Ideally you'd want the test to create a directory that demands admin rights, then create a directory inside the first and it should fail.

If you restricted the test to Windows, you could try creating a folder and setting the owner to LOCAL\SYSTEM. Creating a sub-folder or file should then fail. But I'm not 100% sure you can create rules demanding higher access than what you have.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -56,5 +61,27 @@ public static InstallType GetWorkloadInstallType(SdkFeatureBand sdkFeatureBand,
...
+                {
+                    using (FileStream fs = File.Create(testPath, 1, FileOptions.DeleteOnClose)) { }
+                }
+                else
+                {
+                    Directory.CreateDirectory(Path.GetDirectoryName(testPath));
+                }
+                return true;
+            }
+            catch
```
joeloff, line 81: Definitely catch unauthorized exception to return false. Any other exception doesn't necessarily mean you don't have permission, e.g. you could get an IOException if something else failed in the OS that's not permission related and you can't be sure it will definitely fail again, unless it's a permission issue

sfoslund:
> But I'm not 100% sure you can create rules demanding higher access than what you have.

I haven't been able to find a way to do this, I don't think there's a easy way to add test coverage.

joeloff (approved):

PR #18081, opened 2021-06-04
Revert "MSBuild server V1"

dsplaisted:
Reverts dotnet/sdk#17916

Switching to use the build server seems to be causing failures (timeouts) in the installer repo: https://github.com/dotnet/installer/pull/10804.  So this PR reverts that change until we can understand why it is failing.


brianrob:
cc: @stephentoub, @DamianEdwards

rokonec (approved):

stephentoub:
Thanks.  Hopefully we can root cause it quickly and get it back in.  The teased wins were mouth watering ;-)

PR #18102, opened 2021-06-04
Revert "Set a consistent default behavior for AutoreleasePool during linking."

AaronRobinsonMSFT:
Reverts dotnet/sdk#18050

The default behavior should be contained in the substitutions XML rather than the SDK in this case. See https://github.com/dotnet/sdk/pull/18050#issuecomment-854771229 for details. The PR to replace this logic is https://github.com/dotnet/runtime/pull/53741.


sbomer (approved):
Thanks!

sbomer (commented):
Un-approving since my suggestion doesn't work

AaronRobinsonMSFT:
I am abandoning this. The conversation I had with @sbomer offline indicated this could be noisy given it is only applicable to macOS platforms. I agree with this but the built-in COM feature is also here and that will only ever apply to Windows. This means the current state is okay but less desirable. Going forward the linker team should be proactively consulted about these semantics.

/cc @eerhardt @sbomer @vitek-karas

PR #18136, opened 2021-06-07
Consider workload level platforms on workload install

sfoslund:


sfoslund:
> CSC : error CS8630: Invalid 'nullable' value: 'Enable' for C# 7.3. Please use language version '8.0' or greater.

It looks like this error was caused by a stage 0 update, it was fixed when I reverted it. @jaredpar is this a new error?

sfoslund:
@joeloff this is the result of the discussion we had last week about pack vs workload platforms as defined in the manifests.

jaredpar:
> It looks like this error was caused by a stage 0 update, it was fixed when I reverted it. @jaredpar is this a new error?

This is an old error. It's essentially saying that you specified a value for `/nullable` but had `/langVersion 7.3`. That version of the language does not support nullable reference types at all. Hence the compiler is essentially saying that you have incompatible options here.

src/Cli/dotnet/commands/dotnet-workload/install/xlf/LocalizableStrings.zh-Hans.xlf
```diff
@@ -242,6 +242,16 @@
         <target state="new">Workload installation failed: {0}</target>
         <note />
       </trans-unit>
+      <trans-unit id="WorkloadNotRecognized">
+        <source>Workload with id {0} is not recognized.</source>
+        <target state="new">Workload with id {0} is not recognized.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="WorkloadNotSupportedOnPlatform">
+        <source>Workload with id {0} is not supported on this platform.</source>
```
joeloff, line 251: Do we have a value for "this platform" that can maybe be included in in the message?

joeloff (approved):

src/Cli/dotnet/commands/dotnet-workload/install/xlf/LocalizableStrings.zh-Hans.xlf
sfoslund: There's not an easy way to get the current platform but we should provide the supported platforms? Would this be helpful to users, as it would be a list of rids?

joeloff: I'm always up for providing people with more information, but not if the list is like 20 items as that might be overwhelming.

sfoslund: Right now I'm only seeing max 5-ish, but that was my concern with including it- the list could get pretty big

PR #18156, opened 2021-06-08
Implement workload repair

### LINKED ISSUE #16783: Repair workloads installation
If workloads installation ends up with inconsistency between installation record and the disk state. We need a command to repair. Alternatively `--force` option could be considered.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/16783

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/repair/WorkloadRepairCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/repair/WorkloadRepairCommand.cs
@@ -1,0 +1,118 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.CommandLine.Parsing;
+using Microsoft.Deployment.DotNet.Releases;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.NET.Sdk.WorkloadManifestReader;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Workloads.Workload.Install.InstallRecord;
+using Microsoft.DotNet.Cli.NuGetPackageDownloader;
+using Microsoft.Extensions.EnvironmentAbstractions;
+using Microsoft.DotNet.Workloads.Workload.Install;
+using NuGet.Common;
+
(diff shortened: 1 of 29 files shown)


src/Cli/dotnet/commands/dotnet-workload/repair/WorkloadRepairCommand.cs
```diff
@@ -0,0 +1,118 @@
...
+
+                var packsToInstall = workloadIds
+                    .SelectMany(workloadId => _workloadResolver.GetPacksInWorkload(workloadId.ToString()))
+                    .Distinct()
+                    .Select(packId => _workloadResolver.TryGetPackInfo(packId))
+                    .Where(pack => pack != null);
+
+                foreach (var packId in packsToInstall)
+                {
+                    installer.InstallWorkloadPack(packId, sdkFeatureBand);
```
wli3, line 109: This will run the following underneath. Will the following works when there is existing files? File.Copy would throw when there are existing files at least. I think existing tests cannot validate that since they are mocks
```c#

if (IsSingleFilePack(packInfo))
                            {
                                File.Copy(packagePath, packInfo.Path);
                            }
                            else
                            {
                                var tempExtractionDir = Path.Combine(_tempPackagesDir.Value, $"{packInfo.Id}-{packInfo.Version}-extracted");
                                tempDirsToDelete.Add(tempExtractionDir);
                                Directory.CreateDirectory(tempExtractionDir);
                                var packFiles = _nugetPackageDownloader.ExtractPackageAsync(packagePath, new DirectoryPath(tempExtractionDir)).Result;

                                FileAccessRetrier.RetryOnMoveAccessFailure(() => Directory.Move(tempExtractionDir, packInfo.Path));
                            }
```

wli3 (approved):
Rest is good. But please check if copy over existing files and , nuget extract over existing files will work

wli3:
btw. I saw some existing code has `.Result`. It is not a big problem, but `GetAwaiter().GetResult()` is preferred. https://stackoverflow.com/questions/17284517/is-task-result-the-same-as-getawaiter-getresult

src/Cli/dotnet/commands/dotnet-workload/repair/WorkloadRepairCommand.cs
sfoslund: It checks `if (!PackIsInstalled(packInfo))` on line 87, so it won't get to this logic if there are existing files.

PR #18160, opened 2021-06-08
Update ILLink tests to run on latest frameworks

agocke:
Many of the ILLink tests had hardcoded frameworks so that they only ran
on the current framework when the test was added. This change uses new
infrastructure to automatically run tests on multiple frameworks, especially
newer ones.

### DIFF
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
@@ -481,7 +481,7 @@ namespace Microsoft.NET.Publish.Tests
             var rid = EnvironmentInfo.GetCompatibleRid(targetFramework);
             var testAssetName = "TrimmedAppWithReferences";
             var testAsset = _testAssetsManager
-                .CopyTestAsset(testAssetName)
+                .CopyTestAsset(testAssetName, identifier: Path.GetRandomFileName())
                 .WithSource();
 
             var publishCommand = new PublishCommand(testAsset, "App");
@@ -504,7 +504,7 @@ namespace Microsoft.NET.Publish.Tests
             var rid = EnvironmentInfo.GetCompatibleRid(targetFramework);
             var testAssetName = "TrimmedAppWithReferences";
             var testAsset = _testAssetsManager
-                .CopyTestAsset(testAssetName)
+                .CopyTestAsset(testAssetName, identifier: Path.GetRandomFileName())
                 .WithSource();
 
             var publishCommand = new PublishCommand(testAsset, "App");
@@ -524,7 +524,7 @@ namespace Microsoft.NET.Publish.Tests
             var rid = EnvironmentInfo.GetCompatibleRid(targetFramework);
             var testAssetName = "TrimmedAppWithReferences";
             var testAsset = _testAssetsManager
-                .CopyTestAsset(testAssetName)
+                .CopyTestAsset(testAssetName, identifier: Path.GetRandomFileName())
                 .WithSource()
                 .WithProjectChanges(project => {
                     SetMetadata(project, "PackageReference", "TrimmerSingleWarn", "false");
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
@@ -244,7 +244,7 @@ namespace Microsoft.NET.Publish.Tests
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
+        [InlineData(nameof(LatestTfm))]
         public void ILLink_respects_TrimmableAssembly(string targetFramework)
         {
             var projectName = "HelloWorld";
@@ -344,7 +344,7 @@ namespace Microsoft.NET.Publish.Tests
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [MemberData(nameof(Latest), MemberType = typeof(PublishTestUtils))]
+        [InlineData(nameof(LatestTfm))]
         public void ILLink_can_set_TrimmerDefaultAction(string targetFramework)
         {
             string projectName = "HelloWorld";
@@ -456,8 +456,7 @@ namespace Microsoft.NET.Publish.Tests
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [InlineData("net5.0")]
-        [InlineData(LatestTfm)]
+        [MemberData(nameof(Net5Plus), MemberType = typeof(PublishTestUtils))]
         public void ILLink_accepts_option_to_enable_analysis_warnings_without_PublishTrimmed(string targetFramework)
         {
             var projectName = "AnalysisWarnings";
@@ -475,7 +474,7 @@ namespace Microsoft.NET.Publish.Tests
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
+        [InlineData(nameof(LatestTfm))]
         public void ILLink_shows_single_warning_for_packagereferences_only(string targetFramework)
         {
             var rid = EnvironmentInfo.GetCompatibleRid(targetFramework);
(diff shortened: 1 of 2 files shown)


tlakollo (commented):
The changes look great, I keep wondering though what will happen when we need a combination of certain TFMs, like instead of saying "Use from this TFM up" to say "Use from this TFM down" for example:
ILLink_analysis_warnings_are_enabled_by_default tests only apply to Net6Plus, which means that we could have a test for the other TFMs that test that analysis warnings are disabled by default. 
My guess is that these are such rare scenarios that we will end up using the InlineData again because they are fixed values that won't change in the future, right?

tlakollo (commented):
We may as well include some of this changes in sdk\src\Tests\Microsoft.NET.Publish.Tests\GivenThatWeWantToPublishASingleFileApp.cs

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -108,33 +108,34 @@ public void ILLink_runs_and_creates_linked_app(string targetFramework, bool refe
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [InlineData("netcoreapp3.0", "copyused")]
-        [InlineData("net5.0", "copyused")]
-        [InlineData("net5.0", "link")]
-        public void ILLink_links_simple_app_without_analysis_warnings_and_it_runs(string targetFramework, string trimMode)
+        [MemberData(nameof(SupportedTfms), MemberType = typeof(PublishTestUtils))]
```
sbomer, line 111: I'm curious how you decided whether to run a given test for the full matrix or just some configurations.

sbomer (approved):
Generally LGTM - I would only be worried about increasing the ci time with the extra permutations.

Some tests are failing because a test directory already exists - I think the ones that call CreateTestProjectForILLinkTesting with different theory data just need to pass in a unique referenceProjectIdentifier for each theory instance.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
agocke: My read of the test was that it should pass on all configurations -- not sure whether or not that's correct

agocke:
Note: the tests are currently expected to fail due to https://github.com/dotnet/runtime/issues/53898.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
sbomer: I think it should - my question was more about the criteria for testing all configs. This testcase in particular seems like a good idea as a smoke test, but a lot of them just test MSBuild logic which behaves the same for all TFMs and might not be worth it (this one for example https://github.com/dotnet/sdk/pull/18160/files#diff-f12d20c2aaef217fcb831a91991d26ddc5faac172673d49f4e59cb9db0015252R988).

agocke: Yeah, that makes sense -- in that case we could consider making those `LatestTfm` only

agocke: If you want to point out in the diff which ones you think are SDK-only, I can change those. Otherwise I'll err on the side of running all the frameworks.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -159,8 +160,7 @@ public void PrepareForILLink_can_set_IsTrimmable(string targetFramework)
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [InlineData("net5.0")]
-        [InlineData("net6.0")]
+        [MemberData(nameof(SupportedTfms), MemberType = typeof(PublishTestUtils))]
```
sbomer, line 163:
```suggestion
        [MemberData(nameof(Net5Plus), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -220,8 +220,7 @@ public void ILLink_respects_global_TrimMode(string targetFramework, string trimM
...
```
sbomer, line 223:
```suggestion
        [MemberData(nameof(Net5Plus), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -245,7 +244,7 @@ public void ILLink_roots_IntermediateAssembly(string targetFramework)
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [InlineData("net6.0")]
+        [MemberData(nameof(SupportedTfms), MemberType = typeof(PublishTestUtils))]
```
sbomer, line 247:
```suggestion
        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -345,7 +344,7 @@ public void ILLink_TrimMode_applies_to_IsTrimmable_assemblies(string targetFrame
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [InlineData("net6.0")]
+        [MemberData(nameof(Net6Plus), MemberType = typeof(PublishTestUtils))]
```
sbomer, line 347:
```suggestion
        [MemberData(nameof(Latest), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -475,13 +475,13 @@ public void ILLink_accepts_option_to_enable_analysis_warnings_without_PublishTri
         }
 
         [RequiresMSBuildVersionTheory("16.8.0")]
-        [InlineData("net6.0")]
+        [MemberData(nameof(SupportedTfms), MemberType = typeof(PublishTestUtils))]
```
sbomer, line 478:
```suggestion
        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -498,13 +498,13 @@ public void ILLink_shows_single_warning_for_packagereferences_only(string target
...
```
sbomer, line 501:
```suggestion
        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -518,13 +518,13 @@ public void ILLink_accepts_option_to_show_all_warnings(string targetFramework)
...
```
sbomer, line 521:
```suggestion
        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -581,7 +581,7 @@ public void ILLink_errors_fail_the_build(string targetFramework)
...
```
sbomer, line 584:
```suggestion
        [MemberData(nameof(LatestTfm), MemberType = typeof(PublishTestUtils))]
```

sbomer (commented):
I left a few suggestions. I suppose most of the tests really only need to run for LatestTfm, and there are a few that should be run on .NET 5 and Latest since the msbuild logic changed between 5/6. But I didn't want to reduce testing for netcoreapp3.0 where we already do it today.

PR #18185, opened 2021-06-09
Add Validation suppression engine

joperezr:
Adding suppression engine to be used for all validation errors (Package validation as well as APICompat errors).

Setting it as a draft for now, will turn into a real PR once tests are in.

cc: @safern @Anipik @ericstj @terrajobst 

The idea is that once this PR is in, I plan on adding a design doc with documentation on how the experience should be for customers.


src/Compatibility/Microsoft.DotNet.ValidationSuppression/Microsoft.DotNet.ValidationSuppression.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
joperezr: I initially separated it since I considered it to be a concept separate to APICompat since theoretically it could be used by PackageValidation on it's own, so I tried to keep ApiCompat project to only have APICompat-specific things. That said, as discussed offline, I don't have a strong feeling about this so I'm fine if we decide to collapse even though in my opinion, the conceptual layering does ask for this concept to be separate.

src/Compatibility/Compatibility.sln
```diff
@@ -0,0 +1,50 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
joperezr, line 2: We don't need to check in the solution if we don't want to but I found it easier and seems like everytime that somebody works on PackageValidation will have to touch ApiCompat or viceversa.

safern: It seems like the sdk pattern is to instead add an `slnf` rather than a new solution. The Validation projects are already part of the sdk.sln.

src/Compatibility/Microsoft.DotNet.ValidationSuppression/Microsoft.DotNet.ValidationSuppression.csproj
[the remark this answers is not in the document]
safern: I think it makes sense to have it live the lowest on the stack as possible as API Compat will endup using this at some point when we add an MSBuild Task or dotnet global tool.

src/Compatibility/Compatibility.sln
joperezr: oh good call, I forgot about the slnf, I'll remove the sln then

joperezr:
> eng\CIBuild.cmd -configuration Debug    $(_Test)  /p:RunTestsAsTool=true
========================== Starting Command Output ===========================
"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\workspace\_work\_temp\4c307c6a-2ded-4293-9556-42d32561bf56.cmd""
_Test : The term '_Test' is not recognized as the name of a cmdlet, function, script file, or operable program. Check

CI Failure seems a problem with the yml, seems like it is trying to pass in a variable which is not correct. I tried running that same command locally without that variable and build is green.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Microsoft.DotNet.ValidationSuppression
+{
+    /// <summary>
+    /// Represents a Suppression for a validation error.
+    /// </summary>
+    public class Suppression : IEquatable<Suppression>
+    {
+        /// <summary>
+        /// The DiagnosticId represenging the error to be suppressed.
```
Anipik, line 16: nit: representing

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        public string? Target { get; set; }
+
+        /// <summary>
+        /// Left operand of an APICompat comparison.
+        /// </summary>
+        public string? Left { get; set; }
+
+        /// <summary>
+        /// Right operand of an APICOmpat comparison.
```
Anipik, line 31: nit: APICompat

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+            {
+                if(_validationSuppressions.Contains(error))
+                {
+                    return true;
+                }
+                else
+                {
+                    // See if the error is globally suppressed by checking if the same diagnosticid and target are entered
+                    // without any left and right.
+                    if ((error.DiagnosticId == null || !error.DiagnosticId.StartsWith("pkv", StringComparison.InvariantCultureIgnoreCase))
```
Anipik, line 71: can you explain this condition ```((error.DiagnosticId == null || !error.DiagnosticId.StartsWith("pkv", StringComparison.InvariantCultureIgnoreCase))```

joperezr: Sure, the idea here is that imagine you have the case of an API Compat validation error that applies to multiple left and multiple right, for example: a member that used to exist in your previous package in netstandard2.0 got removed in your new package for both netstandard2.0 and net6.0 assets. This means you will get 2 APICompat validation errors, one for each pair. What this condition supports, is that if you decide  in your suppression file to add a suppression that only has a diagnostic id and a target (in this case the member that was removed) then it will be treated as a global suppression and will be suppressed for both of the cases.

joperezr: You can check out this unit test as an example, which is exercising this if check:
https://github.com/dotnet/sdk/pull/18185/files/107da235e17bce8a0fcd18460c15f18782b703ee#diff-df90c3b499c9c47bff81aaffef3af26830f68ae2b57ae6efb4a9059064021ecdR74-R81

joperezr: Finally, this only relates to ApiCompat errors (since they are the only ones that compare left's and right's) which is why I have the second part of the condition which checks that the diagnosticId is not a PKV (package validation error)

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
```diff
@@ -0,0 +1,59 @@
...
+        public bool Equals(Suppression? other)
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            StringComparer stringComparer = StringComparer.Create(CultureInfo.InvariantCulture, ignoreCase: true);
+
+            if (AreDifferent(DiagnosticId, other.DiagnosticId) ||
```
safern, line 45: You could simplify this check by making the helper method return positive result for equality instead (`AreEqual`) and then you could change this to just be `return AreEqual(...) && AreEqual(...) && ...`

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
```diff
@@ -0,0 +1,59 @@
...
+                AreDifferent(Left, other.Left) ||
+                AreDifferent(Right, other.Right))
+            {
+                return false;
+            }
+
+            return true;
+
+            bool AreDifferent(string? first, string? second) 
+                => !(string.IsNullOrEmpty(first?.Trim()) && string.IsNullOrEmpty(second?.Trim()) || stringComparer.Equals(first?.Trim(), second?.Trim()));
```
safern, line 56: You don't need to check for null or empty right? string.Equals should just be able to handle that for equality?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionComparer.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <summary>
+    /// Defines methods to support the comparison of Suppression objects for equality.
+    /// </summary>
+    internal class SuppressionComparer : IEqualityComparer<Suppression>
+    {
+        /// <inheritdoc/>
+        public bool Equals(Suppression x, Suppression y) => x.Equals(y);
+
+        /// <inheritdoc/>
+        public int GetHashCode(Suppression obj) => HashCode.Combine(obj.DiagnosticId?.ToLowerInvariant(), obj.Target?.ToLowerInvariant(), obj.Left?.ToLowerInvariant(), obj.Right?.ToLowerInvariant());
```
safern, line 18: WHy not override `GetHashCode` on the suppression itself and just call that here?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            // Do nothing. Used for tests.
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="SuppressionEngine"/> based on the contents of a given suppression file.
+        /// </summary>
+        /// <param name="validationSuppressionFile">The path to the suppressions file to be used for initialization.</param>
+        /// <returns>An instance of <see cref="SuppressionEngine"/>.</returns>
+        public static SuppressionEngine CreateFromSuppressionFile(string validationSuppressionFile)
```
safern, line 168: I would name this, `FromFile` that is the pattern we usually use at other places.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Xml.Serialization;
+
+namespace Microsoft.DotNet.ValidationSuppression
```
safern, line 12: Maybe we could name this, `Microsoft.DotNet.Compatibility.Suppression`?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
joperezr: It is able to handle null, but what I want in there is for null == string.Empty which the comparer won't mark as equal. The reason I want null to be equal to string.Empty, is because I want the following two suppressions to be considered equal:
```xml
  <Suppression>
    <DiagnosticId>CP0001</DiagnosticId>
    <Target>T:A.B</Target>
    <Left></Left>
    <Right></Right>
  </Suppression>
  <Suppression>
    <DiagnosticId>CP0001</DiagnosticId>
    <Target>T:A.B</Target>
  </Suppression>
```

joperezr: This case is covered by this unit test case: https://github.com/dotnet/sdk/pull/18185/files#diff-3e46c16064bf399d853838ec1bedde39d07377d9a5fa4a2f61671d49f47beccdR14

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionComparer.cs
joperezr: You are right, I'll do that and that way we can just use the GenericEqualityComparer

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
joperezr: will flip.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: I tried looking for precedent in the framework and find both instances present pretty much equally across the framework (half use FromFile, other half use CreateFromFile). No strong feeling from my side except that I lake all factory methods start with 'Create' which makes them easy to find when using intellisense, but if removing it and having only the `FromFile` as the signature is preferred I can change it.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: I can add the Compatibility as part of the namespace, but the thing I don't love about having `Suppression` as the namespace is that now we will have class `Microsoft.DotNet.Compatibility.Suppression.Suppression`, which is why I originally added ValidationSuppression as the namespace instead.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
safern: A ok, makes sense.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
safern: Ok, makes sense to have `Create` prefix but, why not we remove the `Supression` from the naming so that it is not redundant and that long?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
safern: Yeah that's not ideal. I just don't like `ValidationSuppression` cause `Validation` could apply to any form of validation and this is very specific to compatibility.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
```diff
@@ -0,0 +1,50 @@
...
+#nullable enable
+using System;
+using System.Globalization;
+
+namespace Microsoft.DotNet.ValidationSuppression
+{
+    /// <summary>
+    /// Represents a Suppression for a validation error.
+    /// </summary>
+    public class Suppression : IEquatable<Suppression>
```
safern, line 13: What are your thoughts of adding a field to mark if it is a baseline suppression (running against previous package version) instead of encoding it as part of the `Left` string? I think that way when we include this as part of API Compat, we wouldn't need to use the same encoding for the error messages when using `Left` ID.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+{
+    /// <summary>
+    /// Collection of Suppressions which is able to add suppressions, check if a specific error is suppressed, and write all suppressions
+    /// down to a file. The engine is thread-safe.
+    /// </summary>
+    public class SuppressionEngine
```
safern, line 18: Should the `SupressionEngine` also support `NoWarn`? Or the story for `NoWarn` should be handled individually by every consumer?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
safern: Wouldn't this never be hit? I thought that `Equality` of the suppressions encounters for null/empty `Left` and `Right`? So the previous if would be true?

safern: Ah, I see with the test example. Never mind, cool that we support this "global" suppression.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: The way it is used today it is not needed, because this is today used only by PackageLogger (and in the future by ApiCompat) both of which already take noWarn. For example, the Logger today will only be called with an error after it has validated that the error wasn't specified in NoWarn, so no need for the engine to check again. ApiCompat also takes NoWarn, so we could in the same way perform some filtering for NoWarn first, and then check if the suppression engine is suppressing the specific error.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/Suppression.cs
joperezr: That is a good point, I can try doing that and see how it looks.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+            catch (Exception)
+            {
+                result = new HashSet<Suppression>();
+            }
+            return result;
+        }
+
+        protected virtual Stream GetReadableStream(string baselineFile) => new FileStream(baselineFile, FileMode.Open);
+
+        protected virtual Stream GetWritableStream(string validationSuppressionFile) => new FileStream(validationSuppressionFile, FileMode.OpenOrCreate);
```
safern, line 206: nit, use the same argument name for these two methods, it threw me off a little bit as I got confused thinking there was two different kind of files the suppression engine could accept.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -41,17 +43,31 @@ public void RunApiCompat()
...
+                        // Only include package version when comparing assets from different packages.
+                        string leftOperand = apicompatTuples.leftAssemblyRelativePath;
+                        string rightOperand = apicompatTuples.rightAssemblyRelativePath;
+                        if (!apicompatTuples.leftAssemblyPackagePath.Equals(apicompatTuples.rightAssemblyPackagePath, System.StringComparison.InvariantCultureIgnoreCase))
+                        {
+                            leftOperand = $"[{Path.GetFileNameWithoutExtension(apicompatTuples.leftAssemblyPackagePath)}]{leftOperand}";
+                            rightOperand = $"[{Path.GetFileNameWithoutExtension(apicompatTuples.rightAssemblyPackagePath)}]{rightOperand}";
+                        }
+
+                        _log.LogError(
```
safern, line 61: NIT: Why not add an override that takes an IDifference plus the operands, then just call the API that checks if the error should be suppressed taking each individual field and creates the suppression object for you?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
safern: That way we would avoid having separate logic to filter, currently `NoWarn` is supported via `DiagnosticBag.Filter`

src/Compatibility/Microsoft.DotNet.PackageValidation/PackageValidationLogger.cs
```diff
@@ -1,22 +1,48 @@
...
+        private readonly bool _baselineAllErrors;
 
-        public PackageValidationLogger(Logger log)
+        public PackageValidationLogger(Logger log, string baselineFile)
+            : this(log, baselineFile, false) {}
+
+        public PackageValidationLogger(Logger log, string baselineFile, bool baselineAllErrors)
         {
             _log = log;
+            _suppressionEngine = SuppressionEngine.CreateFromSuppressionFile(baselineFile);
```
safern, line 22: Ideally, `APICompat` should accept a `SuppressionEngine` that way no matter the consumer of `APICompat` always gets the differences baselined?  but that can be done later once `APICompat` has information about the left and right targets.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -0,0 +1,183 @@
...
+    {
+        [Fact]
+        public void AddingASuppressionTwiceDoesntThrow()
+        {
+            var testEngine = SuppressionEngine.Create();
+            AddSuppression(testEngine);
+            AddSuppression(testEngine);
+
+            static void AddSuppression(SuppressionEngine testEngine) => testEngine.AddSuppression("PKG004", "A.B()", "ref/net6.0/mylib.dll", "lib/net6.0/mylib.dll");
+
```
safern, line 20: nit: 
```suggestion
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                Target = "T:A.B",
+                Left = "ref/netstandard2.0/tempValidation.dll",
+                Right = "lib/net6.0/tempValidation.dll"
+            }));
+        }
+
+        [Fact]
+        public void SuppressionEngineDoesNotNeedFileToBeCreated()
+        {
+            SuppressionEngine engine = SuppressionEngine.CreateFromSuppressionFile("AFileThatDoesNotExist.xml");
```
safern, line 53: Shouldn't it ideally throw? This could potentially lead to bugs where the file path is wrong and we just don't know why the errors are being baselined?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
safern: Yeah, however with the suppression engine we would centralize all the suppression logic, so I think that's better. Rather than having multiple places where we filter diagnostics?

safern: That said, we can do that centralization later, just wanted to get your thoughts.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+            {
+                if(_validationSuppressions.Contains(error))
+                {
+                    return true;
+                }
+                else
+                {
+                    // See if the error is globally suppressed by checking if the same diagnosticid and target are entered
+                    // without any left and right.
+                    if ((error.DiagnosticId == null || !error.DiagnosticId.StartsWith("pkv", StringComparison.InvariantCultureIgnoreCase))
```
safern, line 71: Could we instead check for API Compat diagnostic prefix? Since that is logic tight to API Compat, if we added more diagnostic IDs later with a different prefix, we've covered those as well.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+                else
+                {
+                    // See if the error is globally suppressed by checking if the same diagnosticid and target are entered
+                    // without any left and right.
+                    if ((error.DiagnosticId == null || !error.DiagnosticId.StartsWith("pkv", StringComparison.InvariantCultureIgnoreCase)) 
+                        && _validationSuppressions.Contains(new Suppression { DiagnosticId = error.DiagnosticId, Target = error.Target }))
+                    {
+                        return true;
+                    }
+                    else
```
safern, line 76: You could simplify this by making the if condition a return statement instead of an if/else.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: Sure, I'll probably won't have time to finish this PR but feel free to rename as appropriate.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: Sounds good to me, We can leave it as `CreateFromFile`

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: sure, it will be easy to do this by just passing in the NoWarns to the Engine, and then the engine would just return `true` in `IsErrorSuppressed` if the error passed in has the DiagnosticId that matches one of the elements in NoWarn

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: good catch, these are only for tests so not public surface area, but I agree that names should match.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
joperezr: You will eventually create the Suppression object, whether that be at the SuppressionEngine level or at the Log callsite. I'm fine with any design, but what I like about the current one is that it is very easy to see which are all the different suppression types that we log today by just looking for all the callsites of the Suppression constructor.

src/Compatibility/Microsoft.DotNet.PackageValidation/PackageValidationLogger.cs
joperezr: yeah I think we can have a model where both the Logger and ApiCompat take in a suppressionEngine, and they both suppress at different levels. I also agree that this should be done in a separate PR.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
joperezr: The idea here in terms of the experience, is that if you are running validation for the first time and want to baseline all of the errors, you will set the property that get's passed to here in your .csproj even when the file doesn't exist, which will just cause validation to create the file if it doesn't already exist. For example, one minimal project that will consume this should look like:

```xml
  <Sdk Name="Microsoft.DotNet.PackageValidation" Version="1.0.0-dev" />

  <PropertyGroup>
    <TargetFrameworks>netstandard2.0;netstandard1.3;net6.0</TargetFrameworks>
    <!--Extra properties to be set for suppressions-->
    <BaselineAllPackageValidationErrors>true</BaselineAllPackageValidationErrors>
    <PackageValidationSuppressionBaselinePath>PackageValidationSuppressions.xml</PackageValidationSuppressionBaselinePath>
  </PropertyGroup>

</Project>
```

And in this example, even though PackageValidationSuppressions.xml file doesn't exist initially, after the first run of the validation the file will be generated for you.

joperezr: One thing to address before we merge the PR though, will be the name of those two properties: `BaselineAllPackageValidationErrors` `PackageValidationSuppressionBaselinePath`. Even though they are technically correct, the term 'baseline' is overloaded in package validation, as it already means checking against an old package, and in here I was more going for 'suppress all existing errors'.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
joperezr: Makes sense, will do.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
safern: Yeah that makes sense; but wouldn't it make sense to throw if the file doesn't exist and we are note baselining errors? That would be the scenario where a user could potentially think that their suppressions are not working, so an error might help them figure out what the issue is, getting paths correctly on MSBuild can sometime cause trouble.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
safern: Yeah, I know we will create the `Suppression` object but it might make sense to just do that in a central location rather than every time I want to log an error I would have to create an object. Either way is fine I was trying to avoid duplication. Also to look at which are the difference suppression types we could just look at the `LogError` references.

src/Compatibility/Microsoft.DotNet.PackageValidation/CompatibleFrameworkInPackageValidator.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.DotNet.ValidationSuppression;
```
Anipik, line 7: why is this needed ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
Anipik: would it make sense to divide this into 2 different properties, the input where we always check if the path exists and a another one to provide the path to the generated file.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -0,0 +1,183 @@
...
+            (stream) =>
+            {
+                stream.Position = 0;
+                using StreamReader reader = new StreamReader(stream);
+                output = reader.ReadToEnd();
+            });
+            engine.WriteSuppressionsToFile("DummyFile");
+
+            Assert.True(StringComparer.OrdinalIgnoreCase.Equals(engine.sampleSuppressionFile.Trim(), output.Trim()));
+
```
Anipik, line 70: nit : extra line

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -0,0 +1,183 @@
...
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
+    <Target>T:tempValidation.Class1</Target>
+    <Left>[tempvalidation.1.0.0]lib/netstandard1.3/tempValidation.dll</Left>
+    <Right>[tempValidation.2.0.0]lib/netstandard1.3/tempValidation.dll</Right>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>PKV004</DiagnosticId>
+    <Target>.NETFramework,Version=v4.8</Target>
```
Anipik, line 146: do we support the short folder name like "net48" as well ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        public static IEnumerable<object[]> GetEqualData()
+        {
+            yield return new object[] { new Suppression(), new Suppression { DiagnosticId = null, Left = null, Right = null, Target = null} };
+            yield return new object[] { new Suppression(), new Suppression { DiagnosticId = string.Empty, Left = string.Empty, Right = string.Empty, Target = string.Empty} };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004" }, new Suppression { DiagnosticId = "pk004" } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004" }, new Suppression { DiagnosticId = " pk004 " } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B" }, new Suppression { DiagnosticId = " pk004 ", Target = "A.b " } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B", Left = "ref/net6.0/myLib.dll" }, new Suppression { DiagnosticId = " pk004 ", Target = "A.B", Left = "ref/net6.0/mylib.dll" } };
+
```
Anipik, line 19: nit : extra line

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B()", Left = "ref/net6.0/myLib.dll" }, new Suppression { DiagnosticId = "PK004", Target = "A.B()" } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B()", Left = "ref/net6.0/myLib.dll" }, new Suppression { DiagnosticId = "PK004", Target = "A.B()", Left = "lib/net6.0/myLib.dll" } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B()", Right = "ref/net6.0/myLib.dll" }, new Suppression { DiagnosticId = "PK004", Target = "A.B()" } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B()", Right = "ref/net6.0/myLib.dll" }, new Suppression { DiagnosticId = "PK004", Target = "A.B()", Right = "lib/net6.0/myLib.dll" } };
+            yield return new object[] { new Suppression { DiagnosticId = "PK004", Target = "A.B()", Left = "ref/net6.0/mylib.dll", Right = "lib/net6.0/myLib.dll" }, new Suppression { DiagnosticId = "PK004", Target = "A.B()", Left = "ref/netstandard2.0/mylib.dll", Right = "lib/net6.0/myLib.dll" } };
```
Anipik, line 33: can you add some tests for the baselineValiditor.

Anipik (commented):
We need some end to end tests here.  But it can be done separately after we have the infra required to do so.

safern:
Thanks for the review, @Anipik, working through your feedback and mine to update the branch. 

> We need some end to end tests here.

With that you mean running PackageValidation using the suppression engine? Yeah I agree, we can open an issue for that if needed, I'll try to get something though.

Anipik:
> With that you mean running PackageValidation using the suppression engine? 

yes.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
safern: I think `PackageValidationSuppressionBaselinePath` should be the one that we use to generate when `BaselineAllPackageValidationErrors` is set to true. That way you don't need multiple settings.

Also, I think we should have a default, which would be a file at the project level with a convenient name.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
safern: Good call I think we should support that.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ValidationSuppressions/SuppressionEngine.cs
safern: Will address on a separate PR.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
safern: I added the default to the project directory, made it to throw if the file doesn't exist, but don't throw if the path is null or empty, because that means, the user didn't specify a suppressions file and the default path doesn't exist.

I also renamed the properties to: `CompatibilitySuppressionFilePath` and `GenerateCompatibilitySuppressionFile`

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
safern: I'm going to open an issue for this. Given that the package validation errors contain the extended folder name, I think it is fine to not support the short form at the moment. Supporting it involves parsing and translating from one form to the other, so I don't think is worth the effort at the moment.

safern:
> yes.

I'm going to open an issue for that as well. Once we have the e2e infrastructure we can address that.

safern:
@Anipik @ericstj @terrajobst @ViktorHofer could you please take another look?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -61,13 +70,12 @@ public void SuppressionEngineSuppressionsRoundTrip()
             (stream) =>
             {
                 stream.Position = 0;
-                using StreamReader reader = new StreamReader(stream);
+                using StreamReader reader = new(stream);
                 output = reader.ReadToEnd();
             });
             engine.WriteSuppressionsToFile("DummyFile");
```
Anipik, line 76: do we overwrite the file after each call ?

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -12,12 +12,13 @@
 
   <Target Name="RunPackageValidation"
           AfterTargets="Pack"
-          Condition="$(IsPackable) == 'true'">
+          Condition="'$(IsPackable)' == 'true'">
```
Anipik, line 15: should we add the outputs argument here ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
safern: Actually we should generate a tmp file everytime as this could lead to race conditions.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
safern: Could you elaborate, I'm not understanding what do you mean?

Anipik (approved):

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: Disregard this.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
Anipik: are we planning to do it in this pr or in the future ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -0,0 +1,227 @@
...
+            (stream) =>
+            {
+                stream.Position = 0;
+                using StreamReader reader = new(stream);
+                output = reader.ReadToEnd();
+            });
+            string filePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName(), "DummyFile.xml");
+            engine.WriteSuppressionsToFile(filePath);
+
+            Assert.True(StringComparer.OrdinalIgnoreCase.Equals(engine.suppressionsFile.Trim(), output.Trim()));
```
ericstj, line 79: This could be `Assert.Equal(engine.suppressionsFile.Trim(), output.Trim(), ignoreCase: true)` to make this more debuggable.

safern: True. I'll update it when I do a PR.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ErrorSuppression/SuppressionEngine.cs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                if (_validationSuppressions.Contains(error))
+                {
+                    return true;
+                }
+                else
+                {
+                    // See if the error is globally suppressed by checking if the same diagnosticid and target are entered
+                    // without any left and right.
+                    return (error.DiagnosticId == null || error.DiagnosticId.StartsWith("cp", StringComparison.InvariantCultureIgnoreCase)) &&
```
ViktorHofer, line 73: @joperezr do you remember why you explicitly checked for diagnostics that start with the `cp` prefix? Why would we explicitly block to globally suppress package validation diagnostics?

joperezr: Yes, this is because we only allowed global suppressions for API Compatibility diagnostics that were detected. Package validation has two different types of diagnostics: PKVxxxxx which are the ones caused by an issue with how your package is authored, and CPxxxxxx which are the issues where compatibility problems where detected. We only allowed global suppressions for the latter.

ViktorHofer: What's the underlying reason that you can suppress PKV issues via the NoWarn property or when you have a direct match suppression but not via "global suppressions"?

joperezr: I think that when we did this we wanted to be able to avoid a left and a right, that is what we called "global supression", so basically the support to only have a diagnostic and target and have that apply to all left's and right's. The thing that is special with PKV, is that IIRC they don't have a left and right already. A PKV is just a package violation, but you are not really comparing left and right. Even when you have a pkv diagnostic that is comparing a baseline package to the new one, the diagnostic you would get would still not have left and right, it would only have IsBaselineValidation set to true. So essentially, there is no harm in also doing PKV global suppressions as well, but there is no benefit either since left and right are never a thing for PKV errors, and what global suppression is doing in this next line is to see if there is an error matching without a left and right.

ViktorHofer: That makes perfect sense. Right, package validation doesn't compare a specific left to a specific right. In most cases it actually complains that a specific left or right cannot be found. Thanks for the detailed explanation.

ViktorHofer: I added comments to better describe the existing behavior in my PR: https://github.com/dotnet/sdk/pull/25938.

PR #18191, opened 2021-06-09
add end to end test for package validation

### LINKED ISSUE #17879: [Package Validation]Add tests for msbuild tasks and targets

### LINKED ISSUE #17936: [Package-Validation]Add support of package validation on system having sdk version less than net6.0
One of our customers were trying to use Microsoft.Dotnet.PackageValidation package without having .net 6.0 sdk installed.
This lead to failures where the tool was not able to find the depedencies like system.runtime.

This is because currently we ship only net6.0 and net472 assemblies.
I suggest to move net6.0 -> netcoreapp3.1 which  is out LTS

cc @ericstj @ViktorHofer @safern @joperezr

Anipik:
Fixes https://github.com/dotnet/sdk/issues/17879
Fixes https://github.com/dotnet/sdk/issues/17936

These tests protects us from breaking msbuild props and targets file

- simply running package validation with no baseline
- running baseline with baselineVersion
- running baseline with baseline package path
- throwing error when path is not valid.

Found and Fixed a bug where the PackageOutputPath is not normalized.

### DIFF
--- a/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
+++ b/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
@@ -3,5 +3,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <DefineConstants Condition="'$(ForceValidationProblem)' == 'true'">$(DefineConstants);ForceValidationProblem</DefineConstants>
   </PropertyGroup>
 </Project>
--- a/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
+++ b/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
@@ -1,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Sdk Name="Microsoft.DotNet.PackageValidation" Version="1.0.0-test" />
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+  </PropertyGroup>
+</Project>
--- a/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
+++ b/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
@@ -1,8 +1,24 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <Sdk Name="Microsoft.DotNet.PackageValidation" Version="1.0.0-test" />
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <DefineConstants Condition="'$(ForceValidationProblem)' == 'true'">$(DefineConstants);ForceValidationProblem</DefineConstants>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildProjectDirectory)\..\..\..\..\src\Compatibility\Microsoft.DotNet.PackageValidation\Microsoft.DotNet.PackageValidation.csproj" />
+    <PackageReference Include="NuGet.Frameworks" Version="6.0.0-preview.1.66" />
+    <PackageReference Include="NuGet.Packaging" Version="6.0.0-preview.1.66" />
+    <PackageReference Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21277.15" />
+  </ItemGroup>
+
+  <Import Project="$(MSBuildProjectDirectory)\..\..\..\..\src\Compatibility\Microsoft.DotNet.PackageValidation\build\Microsoft.DotNet.PackageValidation.targets" />
+  
+  <PropertyGroup>
+    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)\bin\$(Configuration)\netstandard2.0\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
+    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)\bin\$(Configuration)\net6.0\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
+  </PropertyGroup>
+
 </Project>
--- a/src/Assets/TestProjects/PackageValidationTestProject/Program.cs
+++ b/src/Assets/TestProjects/PackageValidationTestProject/Program.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace PackageValidationTestProject
+{
+    public class Program
+    {
+    }
+}
--- a/src/Assets/TestProjects/PackageValidationTestProject/Program.cs
+++ b/src/Assets/TestProjects/PackageValidationTestProject/Program.cs
@@ -7,5 +7,12 @@ namespace PackageValidationTestProject
 {
     public class Program
     {
+#if ForceValidationProblem
+#if !NET6_0
+  public void SomeAPINotIn6_0()
+  {
+  }
+#endif
+#endif
     }
 }
--- a/src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
+++ b/src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
@@ -13,15 +13,6 @@ namespace Microsoft.DotNet.PackageValidation.Tests
     {
         public ValidatePackageTargetTests(ITestOutputHelper log) : base(log)
         {
-            // clearing the cache
-            string localSdkNugetCacheValidationPath = Path.Combine(TestContext.Current.NuGetCachePath, "microsoft.dotnet.packagevalidation");
-            if (Directory.Exists(localSdkNugetCacheValidationPath))
-                Directory.Delete(localSdkNugetCacheValidationPath, recursive: true);
-
-            // Packing and copying the local version package validation
-            string packageValidationProjectPath = Path.Combine(Path.GetDirectoryName(_testAssetsManager.TestAssetsRoot), "Compatibility", "Microsoft.DotNet.PackageValidation", "Microsoft.DotNet.PackageValidation.csproj");
-            new PackCommand(Log, packageValidationProjectPath)
-                .Execute($"-p:PackageVersion=1.0.0-test;PackageOutputPath={TestContext.Current.TestPackages}");
         }
 
         [Fact]
@@ -86,7 +77,7 @@ namespace Microsoft.DotNet.PackageValidation.Tests
             Assert.Equal(0, result.ExitCode);
 
             result = new PackCommand(Log, Path.Combine(testAsset.TestRoot, "PackageValidationTestProject.csproj"))
-                .Execute($"-p:PackageVersion=2.0.0;PackageValidationBaselineVersion=1.0.0");
+                .Execute($"-p:PackageVersion=2.0.0;PackageValidationBaselineVersion=1.0.0;PackageValidationBaselineName=PackageValidationTestProject");
 
             // No failures while running the package validation on a simple assembly.
             Assert.Equal(0, result.ExitCode);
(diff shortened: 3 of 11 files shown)


src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void ValidatePackageTargetRunsSuccessfully()
+        {
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("PackageValidationTestProject")
+                .WithSource();
+
+            var result = new PackCommand(Log, Path.Combine(testAsset.TestRoot, "PackageValidationTestProject.csproj"))
+                .Execute();
+
+            // No failures while running the package validation on a simple assembly.
```
safern, line 36: is there a way to ensure package validation was indeed executed?

Anipik: Not in the current form. We can add some logging at then end of the task saying that the package validation run successfully.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -38,6 +38,11 @@ protected override void ExecuteCore()
                 Package baselinePackage = NupkgParser.CreatePackage(BaselinePackageTargetPath, runtimeGraph);
                 new BaselinePackageValidator(baselinePackage, NoWarn, null, RunApiCompat, logger).Validate(package);
             }
+
+            if (!Log.HasLoggedErrors)
+            {
+                Log.LogMessage(MessageImportance.High, Resources.SuccessfulPackageRun);
```
joperezr, line 44: Not super relevant to the PR, but I'm not sure we should be setting this message with high importance.

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("PackageValidationTestProject")
+                .WithSource();
+
+            var result = new PackCommand(Log, Path.Combine(testAsset.TestRoot, "PackageValidationTestProject.csproj"))
+                .Execute();
+
+            Assert.Equal(0, result.ExitCode);
+            Assert.Contains(Resources.SuccessfulPackageRun, result.StdOut);
```
joperezr, line 38: Perhaps instead of checking for a string in the output to validate if we actually ran, we should have the validation target to output a  semaphore file on the obj directory which could also be used as inputs/outputs of the target so that we can skip it if the package didn't really change. Then we could just check here in the test if the semaphore was written.

src/Assets/TestProjects/PackageValidationTestProject/Program.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace PackageValidationTestProject
+{
+    public class Program
```
joperezr, line 8: Should we also add #ifdefs in here so that we can test for non-success validation errors? Or is that not intended as part of these tests? I was thinking that if you wanted to fail the tests we could add here something very simple like:
```c#
public class Program
{
#if ForceValidationProblem
#if !NET6_0
  public void SomeAPINotIn6_0()
  {
  }
#endif
#endif
```
That way when you want to cause a validation error you just pass in a /p:ForceValidationProblem=true which you should have in the csproj a defineconstant then so that the APICompat issue is triggered.

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                .WithSource();
+
+            var result = new PackCommand(Log, Path.Combine(testAsset.TestRoot, "PackageValidationTestProject.csproj"))
+                .Execute($"-p:PackageOutputPath={testAsset.TestRoot}");
+
+            Assert.Equal(0, result.ExitCode);
+
+            string packageValidationBaselinePath = Path.Combine(testAsset.TestRoot, "PackageValidationTestProject.1.0.0.nupkg");
+            result = new PackCommand(Log, Path.Combine(testAsset.TestRoot, "PackageValidationTestProject.csproj"))
+                .Execute($"-p:PackageVersion=2.0.0;PackageValidationBaselinePath={packageValidationBaselinePath}");
```
joperezr, line 55: Do we plan to allow baseline validation against a previous package with the same version? Meaning comparing against package version 1.0.0 when you are still building 1.0.0? If so we should add a test for it as regression to make sure this is not broken.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
safern: Yeah I was going to say importance normal or low and it will show on the binlog.

Anipik: we wont see normal or low importance in vs or console. I think it will be nice to log some info for the user.

src/Assets/TestProjects/PackageValidationTestProject/Program.cs
joperezr: The purpose of adding a test that fails validation would be to ensure that a failed validation would fail the msbuild.exe invocation. I know that validation itself has it's own tests so it's ok not to have an exhaustive testing in here, but at least also ensure that we catch regressions where validation failure doesn't trigger msbuild failure would be good.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
safern: IMHO I don't think the user cares much about seeing a log that says the validation ran successfully, that should be implicit with the build being successful.

joperezr: Agreed, the minimal output is super reduced, today for a packed app it is not more than 10 lines, so I don't think that a message like this (specially something saying that validation was successful) is high enough in importance to have it.

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        public ValidatePackageTargetTests(ITestOutputHelper log) : base(log)
+        {
+            // clearing the cache
+            string localSdkNugetCacheValidationPath = Path.Combine(TestContext.Current.NuGetCachePath, "microsoft.dotnet.packagevalidation");
+            if (Directory.Exists(localSdkNugetCacheValidationPath))
+                Directory.Delete(localSdkNugetCacheValidationPath, recursive: true);
+
+            // Packing and copying the local version package validation
+            string packageValidationProjectPath = Path.Combine(Path.GetDirectoryName(_testAssetsManager.TestAssetsRoot), "Compatibility", "Microsoft.DotNet.PackageValidation", "Microsoft.DotNet.PackageValidation.csproj");
```
safern, line 22: Should we be concerned that there might be locked files and that might cause the test to fail? Or is this the pattern that the SDK tests use?

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
safern: Are we adding this logging?

Anipik: Logging wont help here unless we print something to the output screen.

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
Anipik: sdk doesnt produce any packages so there is no precedent here, but yes there could be a race condition here. i will add the lock.

ericstj: We shouldn't be polluting the nuget cache with build output, nor should we be deleting things from the nuget cache.  Come up with a pattern to import the props/targets as if the actual package was built and installed.

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
safern: Ah ok. So a low importance message can't be captured into the test outcome?

Anipik: no

src/Tests/Microsoft.DotNet.PackageValidation.Tests/ValidatePackageTargetTests.cs
Anipik: > We shouldn't be polluting the nuget cache with build output, nor should we be deleting things from the nuget cache.

this is not the machine build output. its specific to the repo. eg the path is ```C:\git\sdk\artifacts\.nuget\packages```

ericstj: This repo produces an SDK.  Also, consumption via package isn’t the mainline scenario. We want this to be part of the .NET SDK, right?  Can’t we leverage the existing hooks that the SDK uses to test itself?  @dsplaisted

dsplaisted: I think I was under the impression that initially this would be consumed as a package.  If it's already going in the .NET SDK, then yes, we should just test it as built-in functionality.

dsplaisted:
Who is going to own the package validation functionality?  Can we update the CODEOWNERS file to cover the package validation and tests?

safern:
Thanks, @dsplaisted, we indeed are missing an entry for the package validation tests. 

@Anipik could you add it?

dsplaisted:
I think we need a CODEOWNERS entry not just for the package validation tests, but for the package validation functionality itself.

safern:
> I think we need a CODEOWNERS entry not just for the package validation tests, but for the package validation functionality itself.

We already have it. We have an entry for all of `src/Compatibility/`, which is where package validation source lives.

src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
```diff
@@ -1,8 +1,24 @@
...
+    <PackageReference Include="NuGet.Frameworks" Version="6.0.0-preview.1.66" />
+    <PackageReference Include="NuGet.Packaging" Version="6.0.0-preview.1.66" />
+    <PackageReference Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21277.15" />
+  </ItemGroup>
+
+  <Import Project="$(MSBuildProjectDirectory)\..\..\..\..\src\Compatibility\Microsoft.DotNet.PackageValidation\build\Microsoft.DotNet.PackageValidation.targets" />
+  
+  <PropertyGroup>
+    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)\bin\$(Configuration)\netstandard2.0\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
```
safern, line 20: Instead of hardcoding paths that contain tfms (that could change), you could instead figure that out at runtime in the test setup and pass that as a property globally?

Anipik: These paths depend on the test project and not on Microsoft.Dotnet.PackageValidation tfms

safern: Actually could you use OutDir or TargetPath to figure out this path?

Anipik: pack runs as  part of outerbuild so both of these props are not set.

ViktorHofer: What about `BaseOutputPath` then?

Anipik: ```BaseOutputPath``` is just bin and output path is ```bin/$(configuration)```. we still have to specify the tfm.

safern: I think it is still worth to use `OutputPath` and then just hardcode the tfm? That way we avoid hardcoding configuration?

Anipik: I forget these props are set in .targets file i.e after the csproj so we wont be able to use them

safern (approved):

src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
safern: Ok, then I guess we can leave with this on the tests.

PR #18204, opened 2021-06-10
Workaround duplicate manifest errors

sfoslund:
First commit taken from https://github.com/dotnet/sdk/pull/17833- fixes dup manifest error caused by manifests with different casing left behind by old previews
Second commit works around manifest id naming change between previews 4 and 5, which also causes a dup manifest error. Verified manually by installing prev 4 manifests on top of this build and ensuing workload install succeeds.

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
@@ -13,6 +13,8 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
         private readonly string _sdkRootPath;
         private readonly string _sdkVersionBand;
         private readonly string [] _manifestDirectories;
+        private static string[] _preview4ManifestIds = new string[] { "Microsoft.NET.Workload.Android", "Microsoft.NET.Workload.BlazorWebAssembly", "Microsoft.NET.Workload.iOS",
+            "Microsoft.NET.Workload.MacCatalyst", "Microsoft.NET.Workload.macOS", "Microsoft.NET.Workload.tvOS" };
 
         public SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVersion)
             : this(sdkRootPath, sdkVersion, Environment.GetEnvironmentVariable)
@@ -81,7 +83,10 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
                 {
                     foreach (var workloadManifestDirectory in Directory.EnumerateDirectories(_manifestDirectories[0]))
                     {
-                        yield return workloadManifestDirectory;
+                        if (!_preview4ManifestIds.Any(prev4Manifest => workloadManifestDirectory.ToLowerInvariant().Contains(prev4Manifest.ToLowerInvariant())))
+                        {
+                            yield return workloadManifestDirectory;
+                        }
                     }
                 }
             }
@@ -102,7 +107,10 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
 
                 foreach (var workloadManifestDirectory in directoriesWithManifests.Values)
                 {
-                    yield return workloadManifestDirectory;
+                    if (!_preview4ManifestIds.Any(prev4Manifest => workloadManifestDirectory.ToLowerInvariant().Contains(prev4Manifest.ToLowerInvariant())))
+                    {
+                        yield return workloadManifestDirectory;
+                    }
                 }
             }
         }
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
@@ -13,8 +13,8 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
         private readonly string _sdkRootPath;
         private readonly string _sdkVersionBand;
         private readonly string [] _manifestDirectories;
-        private static string[] _preview4ManifestIds = new string[] { "Microsoft.NET.Workload.Android", "Microsoft.NET.Workload.BlazorWebAssembly", "Microsoft.NET.Workload.iOS",
-            "Microsoft.NET.Workload.MacCatalyst", "Microsoft.NET.Workload.macOS", "Microsoft.NET.Workload.tvOS" };
+        private static HashSet<string> _outdatedManifestIds = new HashSet<string>() { "microsoft.net.workload.android", "microsoft.net.workload.blazorwebassembly", "microsoft.net.workload.ios",
+            "microsoft.net.workload.maccatalyst", "microsoft.net.workload.macos", "microsoft.net.workload.tvos" };
 
         public SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVersion)
             : this(sdkRootPath, sdkVersion, Environment.GetEnvironmentVariable)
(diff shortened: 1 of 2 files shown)


marcpopMSFT:
BTW, in with VS now to find out if we have more time.  Will know within the hour.

sfoslund:
> BTW, in with VS now to find out if we have more time. Will know within the hour.

Any updates? If not, I'll retarget to main

marcpopMSFT:
They wanted more discussion so that's the updated mail I just sent.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -178,26 +178,27 @@ public void InstallWorkloadManifest(ManifestId manifestId, ManifestVersion manif
...
-                   {
+                    action: () =>
+                    {
                        packagePath = _nugetPackageDownloader.DownloadPackageAsync(WorkloadManifestUpdater.GetManifestPackageId(sdkFeatureBand, manifestId), new NuGetVersion(manifestVersion.ToString())).Result;
                        tempExtractionDir = Path.Combine(_tempPackagesDir.Value, $"{manifestId}-{manifestVersion}-extracted");
                        Directory.CreateDirectory(tempExtractionDir);
                        var manifestFiles = _nugetPackageDownloader.ExtractPackageAsync(packagePath, new DirectoryPath(tempExtractionDir)).Result;
 
-                       if (Directory.Exists(manifestPath) && Directory.GetFileSystemEntries(manifestPath).Any())
+                       var caseSensitiveManifestPath = GetCaseSensitiveManifestDir(Path.GetDirectoryName(manifestPath), manifestId);
```
dsplaisted, line 188: The name `manifestPath` is being used in different places to refer to different things.  In this scope it's the path including the feature band and the manifest ID, but the argument to `GetCaseSensitiveManifestDir` uses the same name but it refers to the feature band path without the manifest ID.  Then in the `GetCaseSensitiveManifestDir` method, the same variable name is re-used inside a lambda to refer to the path including the manifest ID again.

We should try to use different names to make this less confusing.  Maybe `manifestDirectory` for the directory which includes the manifest ID, and `manifestFeatureBandDirectory` for the directory one level up?  `manifestPath` could refer to the path to the actual manifest json file.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -238,6 +239,20 @@ public void InstallWorkloadManifest(ManifestId manifestId, ManifestVersion manif
             }
         }
 
+        private string GetCaseSensitiveManifestDir(string manifestPath, ManifestId manifestId)
```
dsplaisted, line 242: I think naming this `GetExistingManifestDir` would be clearer.  Then add comments explaining that it does a case insensitive search for the existing manifest directory, and that the reason is because preview 4 and preview 5 used different casing, leading to both manifests being present and causing errors.

```suggestion
        private string GetExistingManifestDir(string manifestFeatureBandDirectory, ManifestId manifestId)
```

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -13,6 +13,8 @@ public class SdkDirectoryWorkloadManifestProvider : IWorkloadManifestProvider
         private readonly string _sdkRootPath;
         private readonly string _sdkVersionBand;
         private readonly string [] _manifestDirectories;
+        private static string[] _preview4ManifestIds = new string[] { "Microsoft.NET.Workload.Android", "Microsoft.NET.Workload.BlazorWebAssembly", "Microsoft.NET.Workload.iOS",
+            "Microsoft.NET.Workload.MacCatalyst", "Microsoft.NET.Workload.macOS", "Microsoft.NET.Workload.tvOS" };
```
dsplaisted, line 17: Just want to check that this list is correct.  How did we come up with the list?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -13,6 +13,8 @@ public class SdkDirectoryWorkloadManifestProvider : IWorkloadManifestProvider
         private readonly string _sdkRootPath;
         private readonly string _sdkVersionBand;
         private readonly string [] _manifestDirectories;
+        private static string[] _preview4ManifestIds = new string[] { "Microsoft.NET.Workload.Android", "Microsoft.NET.Workload.BlazorWebAssembly", "Microsoft.NET.Workload.iOS",
```
dsplaisted, line 16: Make this a `HashSet` for faster lookup.  Also instead of `_preview4ManifestIds`, let's call it something like `_outdatedManifestIds`, as that's more descriptive and also applies if we need to do the same thing again.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -81,7 +83,10 @@ public IEnumerable<string> GetManifestDirectories()
                 {
                     foreach (var workloadManifestDirectory in Directory.EnumerateDirectories(_manifestDirectories[0]))
                     {
-                        yield return workloadManifestDirectory;
+                        if (!_preview4ManifestIds.Any(prev4Manifest => workloadManifestDirectory.ToLowerInvariant().Contains(prev4Manifest.ToLowerInvariant())))
```
dsplaisted, line 86: Instead of just checking if the string contains the manifest ID, we should be parsing out the manifest ID from `workloadManifestDirectory` (I think we can do this just with `Path.GetFileName`), and comparing that manifest ID to the list of outdated manifest IDs.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -102,7 +107,10 @@ public IEnumerable<string> GetManifestDirectories()
 
                 foreach (var workloadManifestDirectory in directoriesWithManifests.Values)
                 {
-                    yield return workloadManifestDirectory;
+                    if (!_preview4ManifestIds.Any(prev4Manifest => workloadManifestDirectory.ToLowerInvariant().Contains(prev4Manifest.ToLowerInvariant())))
```
dsplaisted, line 110: Logic should be changed here too.  Perhaps it would be a good idea to factor the logic out into a helper method so it's not repeated.

dsplaisted (commented):
Can we add a test for the case where we have manifests from preview 4 and preview 5 (the test doesn't need to include all of them, just one of them that was renamed and would cause an error if both were used).

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
sfoslund: Taking this part of the change out anyways since there aren't any overlapping manifest ids between prev 4 and 5 so casing shouldn't be an issue.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
sfoslund: I downloaded the prev 4 zip and took all the manifest id dirs in the manifests folder.

sfoslund:
@dsplaisted I've applied all of your feedback and changed the base.

dsplaisted (approved):
Have we tested this fix?  Can we add test coverage?

sfoslund:
> Have we tested this fix? Can we add test coverage?

Just pushed some test coverage, I've also tested manually by overlaying the prev 4 and 5 manifest dirs and running workload install

PR #14093, opened 2020-10-10
Add --use-current-runtime option

am11:
It is a common use-case to use the same runtime during the build and publish. During the development of .NET apps, it is painful to remember the entire - ever growing - `--runtime <baseos>-<arch>` matrix.

`--use-current-runtime`, backed by MSBuild property `UseCurrentRuntimeIdentifier=True/False`, will allow users to keep things neutral in their documentation and deployment scripts.

Fixes #10449
cc @eerhardt @tmds, @dagood

PR #18216, opened 2021-06-10
Add property for opting out of apphost signing in macOS

mateoatr:
Add support for the new `CreateAppHost` argument, `enableMacOSCodeSign`. This PR adds property `_EnableMacOSCodeSign` to allow users to opt-out of apphost signing in macOS.

Dependent upon https://github.com/dotnet/runtime/pull/53913.


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -509,6 +509,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                      Exists('$(AppHostSourcePath)')">
     <PropertyGroup>
       <_UseWindowsGraphicalUserInterface Condition="($(RuntimeIdentifier.StartsWith('win')) or $(DefaultAppHostRuntimeIdentifier.StartsWith('win'))) and '$(OutputType)'=='WinExe'">true</_UseWindowsGraphicalUserInterface>
+      <_EnableMacOSCodeSign Condition="$([MSBuild]::IsOSPlatform(`OSX`)) and Exists('/usr/bin/codesign')">true</_EnableMacOSCodeSign>
```
vitek-karas, line 512: What happens if I try to build a linux binary on mac? (Obviously we must not try to sign it)

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -114,14 +114,13 @@ Copyright (c) .NET Foundation. All rights reserved.
...
     <SelfContained Condition="'$(SelfContained)' == ''">false</SelfContained>
-    <_OnOsx>$(NETCoreSdkRuntimeIdentifier.StartsWith('osx'))</_OnOsx>
     <_RuntimeIdentifierUsesAppHost Condition="$(RuntimeIdentifier.StartsWith('ios')) or $(RuntimeIdentifier.StartsWith('tvos')) or $(RuntimeIdentifier.StartsWith('maccatalyst')) or $(RuntimeIdentifier.StartsWith('android')) or $(RuntimeIdentifier.StartsWith('browser'))">false</_RuntimeIdentifierUsesAppHost>
     <_RuntimeIdentifierUsesAppHost Condition="'$(_RuntimeIdentifierUsesAppHost)' == ''">true</_RuntimeIdentifierUsesAppHost>
     <UseAppHost Condition="'$(UseAppHost)' == '' and
                            '$(_RuntimeIdentifierUsesAppHost)' == 'true' and
                            ('$(SelfContained)' == 'true' or
                             ('$(RuntimeIdentifier)' != '' and '$(_TargetFrameworkVersionWithoutV)' >= '2.1') or
-                            ('$(_TargetFrameworkVersionWithoutV)' >= '3.0' and $(_OnOsx) != 'true'))">true</UseAppHost>
+                            '$(_TargetFrameworkVersionWithoutV)' >= '3.0')">true</UseAppHost>
```
vitek-karas, line 123: While I support this change I think we should make the change more visible. I don't know what's the process for this - technically it could be seen as a mild breaking change. @agocke or @marcpopMSFT  might know how we track these. Short version:
* .NET 6 SDK will generate apphost executable for macOS apps by default for all TFMs that support apphost
* It will now codesign it with the default cert when running on macOS
* In .NET 5 we disabled apphosts on macOS because we didn't sign them and default settings OS would not run such executables.

(This will help a lot with making things like `dotnet run` work on Apple M1 when targetting downlevel TFMs)

marcpopMSFT: Minor changes can go into release notes but full breaking changes get filed as issues in the doc's repo to go into the breaking change documentation.  Since it's a change in default behavior, I'd probably lean towards breaking changes so filing an issue with the details using this link: https://github.com/dotnet/docs/issues/new?assignees=&labels=&template=dotnet-breaking-change.md&title=

agocke: I agree, this is probably not a breaking change for most people, but it seems good to have it documented as such, just in case.

mateoatr: Opened the breaking change issue: https://github.com/dotnet/docs/issues/24766

vitek-karas: There is another issue with this change unfortunately:
On Apple M1 when running with native SDK (so arm64), if I try to build a 3.1 app, it would:
* Try to generate apphost (as per the above change)
* It would imply implicit RID which is the SDK's RID, which would be `osx-arm64`

This will FAIL - there's no osx-arm64. It's possible to try this today:
```
> dotnet new console -f netcoreapp3.1
> dotnet publish -r osx-arm64 --self-contained false
error NETSDK1084: There is no application host available for the specified RuntimeIdentifier 'osx-arm64'.
```

Related to the discussion here: https://gist.github.com/richlander/8cd7b4570775d40f9024561883d79516#overriding-the-implict-rid

@richlander for context

vitek-karas (changes requested):
I think we will have to somehow address the implicit RID on Apple M1 problem before merging this, as it would make what is a very mild breaking change a very hard break (`dotnet build` doesn't work by default on 3.1 app on osx-arm64 SDK).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -62,7 +62,10 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == 'true'">
-    <RuntimeIdentifier>$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="$(NETCoreSdkRuntimeIdentifier.EndsWith('arm64')) and
+                                  $(NETCoreSdkRuntimeIdentifier.StartsWith('osx')) and
+                                  '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0'">$([System.String]::Copy('$(NETCoreSdkRuntimeIdentifier)').Replace("arm64", "x64"))</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
```
vitek-karas, line 68: @wli3 or @dsplaisted could you please review this piece? The goal is when running on osx-arm64 and working with downlevel app (TFM < 6.0) to switch to osx-x64 since the first version which supports osx-arm64 is .NET 6 (downlevel versions don't have support and only run on x64, also macOS arm64 can emulate x64, so it's the right thing to do to run the app as x64).

The main concern we had was how to construct the new RID - specifically if it should maintain the potential OS version and such in the RID or not.

vitek-karas (approved):
Looks good to me, but this should also be reviewed by somebody from SDK to validate mainly the implicit RID calculation.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
dsplaisted: This will only apply for projects that opt in to `UseCurrentRuntimeIdentifier`, or when `--use-current-runtime` is passed on the command line.  Is that the intention?

I thought that we were going to use this type of logic for the the `DefaultAppHostRuntimeIdentifier`.

It's not great that we have to do string manipulation to check the RuntimeIdentifier, but I don't think there's a better option.

I do think when setting the value you don't need to use `String.Copy`.  You can just call replace directly, String.Copy is a workaround to call replace on item metadata, and shouldn't be needed for a property value.

vitek-karas: I don't know what's the difference between `UseCurrentRuntimeIdentifier` and `DefaultAppHostRuntimeIdentifier`. The intent of this change is that typing `dotnet run` on a 3.1 project when running on macOS arm64 SDK will produce macOS x64 app output (apphost) and thus will be able to execute the app. This then also means that `dotnet build` should behave that way.

In short - SDK should not default to the RID of the SDK when targeting downlevel TFMs and such TFMs are not supported on the RID. It should default to whatever RID is actually supported for the TFM on the current machine.

dsplaisted: `DefaultAppHostRuntimeIdentifier` is the primary property you should be concerned with.  That's what determines which apphost gets used if a `RuntimeIdentifier` (or `AppHostRuntimeIdentifier`) isn't otherwise specified.

`UseCurrentRuntimeIdentifier` is a new-ish option to allow building a self-contained app without explicitly specifying the RuntimeIdentifier.  See #14093.  It is probably a good idea for `UseCurrentRuntimeIdentifier` to switch from arm64 to x64 when targeting .NET 5.0 or earlier, but that's not the mainline scenario you're trying to address.

mateoatr: That makes sense. I've changed things so we modify the `DefaultAppHostRuntimeIdentifier`.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -136,6 +133,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'x86'">win-x86</DefaultAppHostRuntimeIdentifier>
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'ARM'">win-arm</DefaultAppHostRuntimeIdentifier>
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'ARM64'">win-arm64</DefaultAppHostRuntimeIdentifier>
+    <DefaultAppHostRuntimeIdentifier Condition="$(NETCoreSdkRuntimeIdentifier.EndsWith('arm64')) and
```
agocke, line 136: Do we need `EndsWith('arm64')`? Isn't it good enough to say `osx`+`<6.0`?

And why use `NETCoreSdkRuntimeIdentifier` instead of `DefaultAppHostRuntimeIdentifier`?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
agocke: If `UseCurrentRuntimeIdentifier` isn't set to `DefaultAppHostRuntimeIdentifier` I think we should set that too. Otherwise I think that scenario will be broken for Mac.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
mateoatr: Changed to use `DefaultAppHostRuntimeIdentifier` instead of `NETCoreSdkRuntimeIdentifier`. I don't think we need the `Ends...` here, it's enough to check for `osx & < 6.0`, but left it here to make things clearer (since we are replacing that same string later).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
marcpopMSFT: To clarify per Teams chat, the plan is for the implicit RID when targeting downlevel runtimes when using the arm64 SDK is to cross-target to x64 for emulation.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -136,6 +133,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'x86'">win-x86</DefaultAppHostRuntimeIdentifier>
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'ARM'">win-arm</DefaultAppHostRuntimeIdentifier>
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'ARM64'">win-arm64</DefaultAppHostRuntimeIdentifier>
+    <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.EndsWith('arm64')) and
+                                  $(DefaultAppHostRuntimeIdentifier.StartsWith('osx')) and
+                                  '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0'">$(DefaultAppHostRuntimeIdentifier.Replace("arm64", "x64"))</DefaultAppHostRuntimeIdentifier>
```
dsplaisted, line 138: I think it would be good to have a comment explaining the reasoning behind this logic.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -509,6 +509,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                      Exists('$(AppHostSourcePath)')">
     <PropertyGroup>
       <_UseWindowsGraphicalUserInterface Condition="($(RuntimeIdentifier.StartsWith('win')) or $(DefaultAppHostRuntimeIdentifier.StartsWith('win'))) and '$(OutputType)'=='WinExe'">true</_UseWindowsGraphicalUserInterface>
+      <_EnableMacOSCodeSign Condition="$([MSBuild]::IsOSPlatform(`OSX`)) and Exists('/usr/bin/codesign') and $(RuntimeIdentifier.StartsWith('osx'))">true</_EnableMacOSCodeSign>
```
dsplaisted, line 512: For framework-dependent apps, `RuntimeIdentifier` won't be set.  Don't you still want to sign the apphost in that case?

src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs
```diff
@@ -21,12 +21,40 @@ namespace Microsoft.NET.Build.Tests
...
+                $"{testProjectName}.dll",
+                $"{testProjectName}.pdb",
+                $"{testProjectName}.deps.json",
+                $"{testProjectName}.runtimeconfig.json"
+            };
+
+            var tfm = testAsset.TargetFrameworkMoniker;
+            if (!string.IsNullOrEmpty(tfm) && (tfm.Contains("netcoreapp") || tfm.Contains("net5") || tfm.Contains("net6")))
+            {
+                var netCoreTFM = new Version(new string(tfm.Where(c => !char.IsLetter(c)).ToArray()));
```
dsplaisted, line 39: I think we should use NuGet TargetFramework parsing APIs (`NuGetFramework.Parse` I believe) here instead of ad-hoc manual parsing.

src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs
```diff
@@ -76,35 +96,85 @@ public void It_builds_a_runnable_apphost_if_opt_in_on_mac()
...
+            {
+                Arguments = $"-d {appHostFullPath}",
+                FileName = @"/usr/bin/codesign",
+                RedirectStandardError = true
+            };
+
+            using (var codesign = Process.Start(psi))
+            {
+                codesign.Start();
+                codesign.StandardError.ReadToEnd()
```
dsplaisted, line 119: I'd suggest using `RunExeCommand` here, which handles running the process and capturing the output.

src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs
```diff
@@ -76,35 +96,85 @@ public void It_builds_a_runnable_apphost_if_opt_in_on_mac()
...
+                })
+                .Should()
+                .Pass();
+
+            var outputDirectory = buildCommand.GetOutputDirectory(targetFramework, runtimeIdentifier: rid);
+            var hostExecutable = $"HelloWorld{(rid.StartsWith("win") ? ".exe" : string.Empty)}";
+            var appHostFullPath = Path.Combine(outputDirectory.FullName, hostExecutable);
+
+            // Check that the apphost was not signed
+            var psi = new ProcessStartInfo()
```
dsplaisted, line 154: Also here for using `RunExeCommand`

src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
```diff
@@ -91,6 +97,7 @@ public TestAsset WithSource()
 
         public TestAsset WithTargetFramework(string targetFramework, string projectName = null)
         {
+            TargetFrameworkMoniker = targetFramework;
```
dsplaisted, line 100: Most of our test assets don't set the `TargetFramework` via `WithTargetFramework`.  So this new property would only be set for a small portion of our tests, while it looks like it should work generally.  Can you try to find another way to solve the problem?  For example, you could add a `targetFramewokr` parameter to the `GetExpectedFilesFromBuild` method.

dsplaisted (commented):
There are tests that verify that we don't sign the apphost, but it doesn't look like there are any that verify that the apphost is signed correctly.  It also looks like it's not currently being signed in the framework-dependent case.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
vitek-karas: We definitely want to sign the apphost in all cases

lewing:
This is approved and green, is there something that needs additional review?

vitek-karas:
We need sign-off from somebody from SDK (for example Daniel).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -136,6 +133,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'x86'">win-x86</DefaultAppHostRuntimeIdentifier>
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'ARM'">win-arm</DefaultAppHostRuntimeIdentifier>
     <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.StartsWith('win')) and '$(PlatformTarget)' == 'ARM64'">win-arm64</DefaultAppHostRuntimeIdentifier>
+    <!-- If we are running on an M1 with a native SDK and the TFM is < 6.0, we have to use a x64 apphost since there are no osx-arm64 apphosts previous to .NET 6.0. -->
+    <DefaultAppHostRuntimeIdentifier Condition="$(DefaultAppHostRuntimeIdentifier.EndsWith('arm64')) and
+                                  $(DefaultAppHostRuntimeIdentifier.StartsWith('osx')) and
+                                  '$(_TargetFrameworkVersionWithoutV)' &lt; '6.0'">$(DefaultAppHostRuntimeIdentifier.Replace("arm64", "x64"))</DefaultAppHostRuntimeIdentifier>
```
dsplaisted, line 139: Use `[MSBuild]::VersionLessThan` here instead of `&lt;` for version comparison.

src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs
```diff
@@ -16,17 +17,45 @@
...
+                $"{testProjectName}.dll",
+                $"{testProjectName}.pdb",
+                $"{testProjectName}.deps.json",
+                $"{testProjectName}.runtimeconfig.json"
+            };
+
+            if (!string.IsNullOrEmpty(targetFramework))
+            {
+                var parsedTargetFramework = NuGetFramework.Parse(targetFramework);
+                if (parsedTargetFramework.Version >= new Version(5, 0, 0, 0))
```
dsplaisted, line 41: The `Version` type has somewhat unexpected comparison rules (5.0 is not the same as 5.0.0.0), so I would suggest just comparing against the `Major` version.
```suggestion
                if (parsedTargetFramework.Version.Major >= 5)
```

src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs
```diff
@@ -16,17 +17,45 @@
...
+                    expectedFiles.Add($"ref/{testProjectName}.dll");
+
+                if (parsedTargetFramework.Version < new Version(6, 0, 0, 0))
```
dsplaisted, line 44:
```suggestion
                if (parsedTargetFramework.Version.Major < 6)
```

src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs
```diff
@@ -16,17 +17,45 @@
...
+            return expectedFiles.ToArray();
+        }
+
         public AppHostTests(ITestOutputHelper log) : base(log)
         {
         }
 
-        [PlatformSpecificTheory(TestPlatforms.Windows | TestPlatforms.Linux | TestPlatforms.FreeBSD)]
-        [InlineData("netcoreapp3.0")]
+        [PlatformSpecificTheory(TestPlatforms.Windows | TestPlatforms.Linux | TestPlatforms.FreeBSD | TestPlatforms.OSX)]
```
dsplaisted, line 55: Now that this runs on macOS, can this just be a `Theory` instead of a `PlatformSpecificTheory`?

dsplaisted (approved):
A few comments, but looks good.

PR #18231, opened 2021-06-11
Support preview Windows SDK Ref packs, and simplify how they're defined

dsplaisted:
- Instead of duplicating Windows SDK Reference `KnownFrameworkReference` items for each target framework, they can be defined in BundledVersions.props like the following:

```xml
<WindowsSdkSupportedTargetPlatformVersion Include="10.0.19041.0" WindowsSdkPackageVersion="10.0.19041.16" MinimumNETVersion="5.0" />
<WindowsSdkSupportedTargetPlatformVersion Include="10.0.18362.0" WindowsSdkPackageVersion="10.0.18362.16" MinimumNETVersion="5.0" />
<WindowsSdkSupportedTargetPlatformVersion Include="10.0.17763.0" WindowsSdkPackageVersion="10.0.17763.16" MinimumNETVersion="5.0" />
<WindowsSdkSupportedTargetPlatformVersion Include="8.0" />
<WindowsSdkSupportedTargetPlatformVersion Include="7.0" />
```

- Support `UseWindowsSDKPreview` property to opt in to preview Windows SDKs
- Support `WindowsSdkPackageVersion` to explicitly set the Windows SDK Ref package version

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
@@ -17,9 +17,34 @@ Copyright (c) .NET Foundation. All rights reserved.
     </ResolveAssemblyReferencesDependsOn>
   </PropertyGroup>
 
+  <UsingTask TaskName="CreateWindowsSdkKnownFrameworkReferences" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="CheckForDuplicateFrameworkReferences" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="ProcessFrameworkReferences" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="ResolveAppHosts" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
+
+  <Target Name="AddWindowsSdkKnownFrameworkReferences"
+          Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetPlatformIdentifier)' == 'Windows'">
+
+    <!-- Remove Windows SDK KnownFrameworkReference items from BundledVersions.props (they will eventually be removed, but that is in a different repo so
+         we can't do the change atomically -->
+    <ItemGroup>
+      <KnownFrameworkReference Remove="Microsoft.Windows.SDK.NET.Ref" />
+    </ItemGroup>
+    
+    <!-- Generate KnownFrameworkReference items for the Windows SDK pack -->
+    <CreateWindowsSdkKnownFrameworkReferences
+      UseWindowsSDKPreview="$(UseWindowsSDKPreview)"
+      WindowsSdkPackageVersion="$(WindowsSdkPackageVersion)"
+      TargetFrameworkIdentifier="$(TargetFrameworkIdentifier)"
+      TargetFrameworkVersion="$(_TargetFrameworkVersionWithoutV)"
+      TargetPlatformIdentifier="$(TargetPlatformIdentifier)"
+      TargetPlatformVersion="$(TargetPlatformVersion)"
+      WindowsSdkSupportedTargetPlatformVersions="@(WindowsSdkSupportedTargetPlatformVersion)">
+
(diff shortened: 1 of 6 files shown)


dsplaisted:
> Is there a reason the package versions here are *.16? The latest .NET SDK should be on *.17

@angelazhangmsft Those definitions are just temporary.  The real numbers will come from BundledVersions.props generated in the dotnet/installer repo, but we can't atomically commit to two different repos :-)

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
```diff
@@ -17,10 +17,35 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   <UsingTask TaskName="ProcessFrameworkReferences" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="ResolveAppHosts" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
+  <Target Name="AddWindowsSdkKnownFrameworkReferences"
+          Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetPlatformIdentifier)' == 'Windows'">
+
+    <!-- Remove Windows SDK KnownFrameworkReference items from BundledVersions.props (they will eventually be removed, but that is in a different repo so
+         we can't do the change atomically -->
+    <ItemGroup>
+      <KnownFrameworkReference Remove="Microsoft.Windows.SDK.NET.Ref" />
```
joeloff, line 31: I take it there are other items that need to remain or would ```Remove="@(KnownFrameworkReference)"``` be better to clear the whole item group?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
```diff
@@ -342,6 +345,60 @@ public void SupportedOSPlatformVersionCanBeSetInDirectoryBuildTargets()
         }
 
 
+        [WindowsOnlyTheory]
+        //  Basic Windows TargetFramework
+        [InlineData("net5.0-windows10.0.19041.0", false, null, "10.0.19041.*")]
+        //  Basic UseWindowsSdkPreview usage
+        [InlineData("net5.0-windows10.0.99999.0", true, null, "10.0.99999-preview")]
```
joeloff, line 352: Do these packages ever use full semantic versioning with build numbers and metadata? It sounded from the comments in the build task that the patterns are fixed, e.g. you add "preview" and that will never change.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
```diff
@@ -11,9 +11,12 @@
 using System.Xml.Linq;
 using System.IO;
 using System.Linq;
+using System;
 
 namespace Microsoft.NET.Build.Tests
 {
+    //  Add Windows SDK tests here?
```
wli3, line 18: ?

wli3 (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
dsplaisted: Yes, there are other items for all of the shared frameworks we have.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
dsplaisted: Yes, the release versions of the packages use 4 digit build numbers, but NuGet doesn't allow you to use prerelease information in a package version if it has 4 numbers.  So the preview packages will only have 3 numbers, and they plan to always just have them marked as `-preview`.  If they need something else, then the `WindowsSdkPackageVersion` property can be used to explicitly specify the full package version.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
dsplaisted: oops!

joeloff (approved):

PR #18234, opened 2021-06-11
[5.0.3xx] Workaround duplicate manifest errors

sfoslund:
#18204

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
@@ -13,7 +13,7 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
         private readonly string _sdkRootPath;
         private readonly string _sdkVersionBand;
         private readonly string [] _manifestDirectories;
-        private static HashSet<string> _outdatedManifestIds = new HashSet<string>() { "microsoft.net.workload.android", "microsoft.net.workload.blazorwebassembly", "microsoft.net.workload.ios",
+        private static HashSet<string> _outdatedManifestIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "microsoft.net.workload.android", "microsoft.net.workload.blazorwebassembly", "microsoft.net.workload.ios",
             "microsoft.net.workload.maccatalyst", "microsoft.net.workload.macos", "microsoft.net.workload.tvos" };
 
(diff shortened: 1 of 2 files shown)


dsplaisted:
@sfoslund When you're porting changes to another branch, can you link the original PR they came from in the description?  IE #18204 in this case

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -13,6 +13,8 @@ public class SdkDirectoryWorkloadManifestProvider : IWorkloadManifestProvider
         private readonly string _sdkRootPath;
         private readonly string _sdkVersionBand;
         private readonly string [] _manifestDirectories;
+        private static HashSet<string> _outdatedManifestIds = new HashSet<string>() { "microsoft.net.workload.android", "microsoft.net.workload.blazorwebassembly", "microsoft.net.workload.ios",
```
nil4, line 16:
```suggestion
        private static HashSet<string> _outdatedManifestIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "microsoft.net.workload.android", "microsoft.net.workload.blazorwebassembly", "microsoft.net.workload.ios",
```

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -102,9 +107,18 @@ public IEnumerable<string> GetManifestDirectories()
...
+                    {
+                        yield return workloadManifestDirectory;
+                    }
                 }
             }
         }
+
+        private bool IsManifestIdOutdated(string workloadManifestDir)
+        {
+            var manifestId = Path.GetFileName(workloadManifestDir).ToLowerInvariant();
```
nil4, line 120: If `_outdatedManifestIds` would be created with `StringComparer.OrdinalIgnoreCase`, this would no longer be necessary:
```suggestion
            var manifestId = Path.GetFileName(workloadManifestDir);
```

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
wli3: better way should be convert them to ManifestId type and then do comparison.

wli3:
@dsplaisted merge when done?

PR #18241, opened 2021-06-11
Set file permission according to UnixPermissions.xml

wli3:


src/Cli/dotnet/FilePermissionSetter.cs
```diff
@@ -19,9 +19,9 @@ public void SetUserExecutionPermission(string path)
         /// (G)roup can read, can't write and can execute.
         /// (O)thers can read, can't write and can execute.
         /// </summary>
-        public void Set755Permission(string path)
+        public void SetPermission(string path, string chmodArgument)
         {
-            RunCommand(path, "755");
+            RunCommand(path, chmodArgument.ToString());
```
joeloff, line 24: What happens if the permission you're trying to set is higher than the permission of the user executing chmod? I'm guess chmod just fails? Should we check for that, or is this a "best effort/fire-and-forget" operation?

wli3: run command will fail and throw. And that's good enough. Validate on SDK level won't help too much. We will still get an error during installation anyway.

wli3: @joeloff could you approve the PR?

joeloff (approved):

PR #18301, opened 2021-06-16
Deprecate -p option (in short form) to dotnet run

dsplaisted:
See https://github.com/dotnet/designs/pull/229 and https://github.com/dotnet/docs/issues/24727

This PR implements the deprecation of the `-p` option.  Supporting `-p` to set properties will come in a later preview or release.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
@@ -218,6 +218,6 @@ The current {1} is '{2}'.</value>
     <value>The property '{0}' must be an object if it is specified.</value>
   </data>
   <data name="RunCommandProjectAbbreviationDeprecated" xml:space="preserve">
-    <value>WARNING: The abbreviation of -p for --project is deprecated. Please use --project.</value>
+    <value>Warning NETSDK1174: The abbreviation of -p for --project is deprecated. Please use --project.</value>
   </data>
 </root>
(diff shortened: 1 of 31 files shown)


src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
```diff
@@ -217,4 +217,7 @@ The current {1} is '{2}'.</value>
   <data name="ValueMustBeAnObject" xml:space="preserve">
     <value>The property '{0}' must be an object if it is specified.</value>
   </data>
-</root>
+  <data name="RunCommandProjectAbbreviationDeprecated" xml:space="preserve">
+    <value>WARNING: The abbreviation of -p for --project is deprecated</value>
```
joeloff, line 221: nit: Missing "."

timheuer: The deprecation message doesn't match the spec of `The abbreviation of -p for --project is deprecated. Please use --project.`  The latter part might seem redundant but makes it clear that `--project` is still supported and the replacement.

joeloff (approved):

src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
dsplaisted: Yes, this was originally based on an earlier version of the spec and has now been updated.

PR #18336, opened 2021-06-17
Deprecate -p option (in short form) to dotnet watch

pranavkm:
Contributes to https://github.com/dotnet/designs/pull/229

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Program.cs
+++ b/src/BuiltInTools/dotnet-watch/Program.cs
@@ -105,11 +105,11 @@ Examples:
 
         internal async Task<int> RunAsync(string[] args)
         {
-            var rootCommand = CreateRootCommand(HandleWatch);
+            var rootCommand = CreateRootCommand(HandleWatch, _reporter);
             return await rootCommand.InvokeAsync(args);
         }
 
-        internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>> handler)
+        internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>> handler, IReporter reporter)
         {
             var quiet = new Option<bool>(
                 new[] { "--quiet", "-q" },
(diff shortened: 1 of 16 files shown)


src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -147,6 +147,11 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
             root.TreatUnmatchedTokensAsErrors = false;
             root.Handler = CommandHandler.Create((CommandLineOptions options, ParseResult parseResults) =>
             {
+                if (!string.IsNullOrEmpty(parseResults.ValueForOption<string>("-p")))
```
dsplaisted, line 150: I don't know how the command line parsing works here.  Have you validated that if `--project` is used, the warning isn't printed (since they are both the same `Option`)?  Should we have an automated test for that?

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -147,6 +147,11 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
...
+                {
+                     reporter.Warn(Resources.Warning_ProjectAbbreviationDeprecated);
```
dsplaisted, line 152: What does this actually look like in the console output?  It looks like we're going to go with the following (in yellow text) for `dotnet run`:

> Warning NETSDKNNNN: The abbreviation of -p for --project is deprecated. Please use --project.

Where NNNN is replaced with an SDK error code, so you can better search for information about the error.

dsplaisted (approved):

src/BuiltInTools/dotnet-watch/Program.cs
pranavkm: You're right. It prints a warning in either case.

src/BuiltInTools/dotnet-watch/Program.cs
pranavkm: Here's what this looks like now sans the SDK error code:

image

Have you decided on what the error code is?

dsplaisted: NETSDK1174 - https://github.com/dotnet/sdk/pull/18301

PR #18404, opened 2021-06-21
Fix splitting of DOTNETSDK_WORKLOAD_PACK_ROOTS environment variable

dsplaisted:


sfoslund (approved):

dsplaisted:
@tmds @omajid FYI, this fix will be needed to support user-local workload installs.  I'm working on a tool to help set that up.

PR #18440, opened 2021-06-22
[Preview 6] Add Package Validation suppression engine

safern:
## Description
Currently with Package Validation there is no way to add granular error suppressions, the only way is to add a NoWarn​ entry to the project, which disables the whole diagnostic ID. So if a customer has an intentional breaking change by removing a method named A.B but also accidentally deleted a method A.C, they would get two errors with diagnostic id CP002​, one for each member that was removed, with NoWarn, they would disable the whole CP002​ diagnostic ID, so the accidental deletion of A.C​ would also be suppressed. 

With this change they can add an XML file listing the suppressions they care to suppress, so they could now disable just CP002 for method A.B by adding this to the XML:

```xml
<Suppressions>
  <Suppression>
    <DiagnosticId>CP002</DiagnosticId>
    <Target>M:A.B</Target>
  </Suppression>
</Suppressions>
```
This is a critical feature for Preview 6 in order to get feedback on granular suppressions story. 
Direct port of: https://github.com/dotnet/sdk/pull/18185

## Customer Impact

It gives the ability to suppress errors on package validation in a granular form rather than just using `NoWarn` and suppressing all errors with the same diagnostic ID.

## Testing

Added unit tests and also manually tested with sample projects.

## Risk
Low, this just adds new functionality.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/ErrorSuppression/Suppression.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/ErrorSuppression/Suppression.cs
@@ -1,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#nullable enable
+using System;
+using System.Xml.Serialization;
+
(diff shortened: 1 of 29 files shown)


Anipik (approved):

safern:
Approved by tactics offline.

PR #18451, opened 2021-06-22
Add StrictMode to MembersMustExist rule

safern:
If we are going to enable strict mode for compile asset vs runtime asset comparison on Preview 6 we might want to port this as well.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
@@ -138,6 +138,9 @@
   <data name="MemberExistsOnLeft" xml:space="preserve">
     <value>Member '{0}' exists on the left but not on the right</value>
   </data>
+  <data name="MemberExistsOnRight" xml:space="preserve">
+    <value>Member '{0}' exists on the right but not on the left</value>
+  </data>
   <data name="ProvidedPathToLoadBinariesFromNotFound" xml:space="preserve">
     <value>Could not find the provided path '{0}' to load binaries from.</value>
   </data>
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/MembersMustExist.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/MembersMustExist.cs
@@ -35,11 +35,11 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
         {
             if (left != null && right == null)
             {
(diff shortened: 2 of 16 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/MembersMustExist.cs
```diff
@@ -35,17 +35,11 @@ private void RunOnTypeSymbol(ITypeSymbol left, ITypeSymbol right, IList<CompatDi
         {
             if (left != null && right == null)
             {
-                AddDifference(left, DifferenceType.Removed, Resources.TypeExistsOnLeft);
+                differences.Add(CreateDifference(left, DiagnosticIds.TypeMustExist, DifferenceType.Removed, "Type '{0}' exists on the left but not on the right"));
```
danmoseley, line 38: Shouldn't these all be in the resx?

danmoseley: Looks like we have TypeExistsOnLeft or MemberExistsOnLeft etc.

safern: Yes. This was because I had started the work before I added the resx. Good catch.

safern: Fixed, thanks for catching this 😄

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -138,6 +138,9 @@
   <data name="MemberExistsOnLeft" xml:space="preserve">
     <value>Member '{0}' exists on the left but not on the right</value>
   </data>
+  <data name="MemberExistsOnRight" xml:space="preserve">
+    <value>Member '{0}' exists on the right but not on the left</value>
```
ericstj, line 142: Unreleated to this PR, but are we planning on parameterizing the names for right/left so that the diagnostics are more readable?

ericstj (approved):

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
safern: Yes. I'm working on that. Should have a PR out soon. That's on the timeline for Preview 6.

PR #18482, opened 2021-06-23
Allow copying across mounts on workload install

### LINKED ISSUE #18450: Workload install copying across mounts error
The files on disk workload install needs to copy files into both the user's home (advertising manifests) and the dotnet directory (packs, installed manifests). This can cause problems with those are under different mounts, for example, the default user home on linux is `/home/...` and you could potentially install dotnet to `/mnt/...`. This is problematic because install uses one temp directory and has to be able to copy from that temp directory into both the home and dotnet directories. When the mounts for these locations conflict we run into a `Invalid cross-device link.` error (https://github.com/dotnet/runtime/issues/31149). 

Repro: Workload install using WSL

Error output:
> /mnt/c/code/tmp2$ ./dotnet workload install microsoft-net-sdk-blazorwebassembly-aot
Skip NuGet package signing validation. NuGet signing validation is not available on Linux or macOS https://aka.ms/workloadskippackagevalidation .
Updated advertising manifest microsoft.net.sdk.android.
Updated advertising manifest microsoft.net.sdk.ios.
Updated advertising manifest microsoft.net.sdk.maccatalyst.
Updated advertising manifest microsoft.net.sdk.macos.
Failed to update the advertising manifest microsoft.net.sdk.maui: microsoft.net.sdk.maui.manifest-6.0.100 is not found in NuGet feeds https://api.nuget.org/v3/index.json"..
Updated advertising manifest microsoft.net.sdk.tvos.
Updated advertising manifest microsoft.net.workload.mono.toolchain.
Installing workload manifest microsoft.net.sdk.android version 30.0.100-preview.5.28.
Workload installation failed, rolling back installed packs...
Installing workload manifest microsoft.net.sdk.android version 11.0.200-ci.main.256.
Installation roll back failed: Failed to install manifest microsoft.net.sdk.android version 11.0.200-ci.main.256: The transaction has aborted..
Workload installation failed: Failed to install manifest microsoft.net.sdk.android version 30.0.100-preview.5.28: Invalid cross-device link.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/18450

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -193,7 +193,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                     Directory.Delete(adManifestPath, true);
                 }
                 Directory.CreateDirectory(Path.GetDirectoryName(adManifestPath));
-                FileAccessRetrier.RetryOnMoveAccessFailure(() => CopyDirectory(Path.Combine(extractionPath, "data"), adManifestPath));
+                FileAccessRetrier.RetryOnMoveAccessFailure(() => DirectoryPath.MoveDirectory(Path.Combine(extractionPath, "data"), adManifestPath));
 
                 _reporter.WriteLine(string.Format(LocalizableStrings.AdManifestUpdated, manifestId));
 
@@ -271,24 +271,5 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
         internal static PackageId GetManifestPackageId(SdkFeatureBand featureBand, ManifestId manifestId) =>
             new PackageId($"{manifestId}.Manifest-{featureBand}");
-
-        // Note: cannot use Directory.Move because it errors when copying across mounts
-        internal static void CopyDirectory(string sourcePath, string destPath)
-        {
-            if (!Directory.Exists(destPath))
-            {
-                Directory.CreateDirectory(destPath);
-            }
-
-            foreach (var dir in Directory.GetDirectories(sourcePath))
-            {
-                CopyDirectory(dir, Path.Combine(destPath, Path.GetFileName(dir)));
-            }
-
-            foreach (var file in Directory.GetFiles(sourcePath))
-            {
-                new FileInfo(file).CopyTo(Path.Combine(destPath, Path.GetFileName(file)), true);
-            }
-        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -271,5 +271,24 @@ private ManifestVersion GetInstalledManifestVersion(ManifestId manifestId)
 
         internal static PackageId GetManifestPackageId(SdkFeatureBand featureBand, ManifestId manifestId) =>
             new PackageId($"{manifestId}.Manifest-{featureBand}");
+
+        // Note: cannot use Directory.Move because it errors when copying across mounts
+        internal static void CopyDirectory(string sourcePath, string destPath)
```
joeloff, line 276: Would it makes sense to make this an extension method  we could reuse in other places in the CLI if we ever needed something like this again?

dsplaisted, line 276: It looks like this changes behavior from moving to copying.  It's also no longer atomic, which I think was the reason for extracting to a temp folder and then moving in the first place.

So I think we should use `Directory.Move` when it's possible.  When it's not possible, we probably want to do something more resilient / atomic, such as copying it to the metadata folder (which we can probably assume is on the same volume) and then moving it from there.

I'm not sure if there's a way to detect up-front whether Directory.Move will work or whether we have to catch the failure.

cc @wli3

radical:
Should there be a new test to ensure that this doesn't regress?

sfoslund:
> Should there be a new test to ensure that this doesn't regress?

This line causes existing tests to use different mounts/ drives, which repros the issue: https://github.com/dotnet/sdk/pull/18482/files#diff-81861764e6652051f255fa90e0aba2c3bc22f8bb020931fb7de60a5c5650174cR433

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
sfoslund: Makes sense, I'll make that change.

sfoslund: I agree on using Directory.Move when possible, I'll make that change. However I don't think we can always depend on the metadata dir being in the same mount- for example, when we're copying advertising manifests into the user home I believe that's on a different mount on linux by default.

joeloff: Is it possible that one of the target folders can be a symlink that crosses mount points?

wli3 (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
wli3: we kind of giving up the atomic move when we decided on using OS temp dir. So this is expected.

Although I am surprised Directory.Move will just fail

PR #18502, opened 2021-06-24
Allow `dotnet sln list` to display solution folder paths

### LINKED ISSUE #10308: Add solution folder argument to dotnet-sln show-folders  (show solution folders)
Feature request:

If possible, can this fucntionality be added: list of solution folders?
Use case:

1. List of solution folders (dotnet sln show folders)
2. Add new project under on of this solution folder.  (dotnet sln add project)

It must be related to PR: [https://github.com/dotnet/cli/pull/11000](url)

ZeroErrors:
Fixes https://github.com/dotnet/sdk/issues/10308

## Change log
- Fix `NoProjectsFound` not being displayed if there is only solution folders
- Adds extension method `GetFullSolutionFolderPath` for `SlnProject`
- Adds `--solution-folders` option to `dotnet sln list` to display solution folder paths.
- Adds unit test for `dotnet sln list --solution-folders`

## Description
When manipulating solutions via `dotnet sln`, such as adding new projects inside of solution folders, it would be useful to also be able to view the solution folders projects are in.

## Testing
Unit test included.

To test manually setup a solution and a project something like
```
dotnet new sln
dotnet new console --output src/todo-app
dotnet sln add src/todo-app --solution-folder todo
```

Then running `dotnet sln list --solution-folders` should output the path `todo/todo-app`.


sfoslund:
Can you clarify what the expected output is for this option? You expect the output to contain a list of directories that contain projects in the specified solution? In your testing example in the description, you say that the output should be `todo/todo-app`, where is this result coming from?

marcpopMSFT:
Coming back to old PRs, I merged with main and updated the PR to specifically list the solution folders. The original PR listed the projects within folders which is a bit different than what I'd expect with this flag.

That caused confusion for me because if I had foo\bar\bar.csproj and bar\bar\bar.csproj with foo and bar as solution folders, the original output was just foo\bar and bar\bar. I think potentially there is an interesting case for showing which projects are within each solution folder but that can be a separate issue. For now, this will just fetch the solution folders themselves and show their hierarchy.

From our test:
Solution Folder(s)
------------------
EmptyFolder
EmptyFolder\NestedEmptyFolder
NestedSolution
NestedSolution\NestedFolder
NestedSolution\NestedFolder\NestedFolder
NewFolder1
NewFolder1\NewFolder2
Root Empty Folder

image

src/Cli/dotnet/SlnProjectExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Microsoft.DotNet.Tools.Common
+{
+    internal static class SlnProjectExtensions
+    {
+        public static string GetFullSolutionFolderPath(this SlnProject slnProject)
+        {
+            var slnFile = slnProject.ParentFile;
+            var nestedProjects = slnFile.Sections
+                .GetOrCreateSection("NestedProjects", SlnSectionType.PreProcess)
+                .Properties;
```
nagilson, line 17: Is there any chance the project will be in `.NestedPropertySets` vs `.Properties`? I couldn't figure out what the NestedPropertySets were for. If you aren't sure, please feel free to ping me or those on the SLN team so they can provide an answer here.

nagilson, line 17: Do we know that all of the things in `_properties` under a solution folder will be projects? To me it looks like it, but I don't know for sure.

nagilson (approved):
LGTM. The code is very clean, so thank you. There are 2 things I would like clarity on, but otherwise from testing it and analyzing it, I'm glad to merge this.

src/Cli/dotnet/commands/dotnet-sln/list/SlnListParser.cs
```diff
@@ -11,6 +12,8 @@ namespace Microsoft.DotNet.Cli
 {
     public static class SlnListParser
     {
+        public static readonly Option<bool> SolutionFolderOption = new Option<bool>(new string[] { "-s", "--solution-folders" }, LocalizableStrings.ListSolutionFoldersArgumentDescription);
```
nagilson, line 15: I kind of prefer `-f` because it enunciates the 'folder' aspect of it, but this is also good.

nagilson:
@ZeroErrors I must say I didn't expect to review something from someone from Hypixel Studios. Very cool. ^^ Thank you for contributing to the SDK, this is a cool feature add. I'm sorry it took us so long to get to it. Thanks for the thumbs up on Marc's reply, if this doesn't meet your needs/expectations though feel free to let us know.

src/Cli/dotnet/commands/dotnet-sln/list/SlnListParser.cs
marcpopMSFT: @baronfel to comment on the choice of short parameter.

src/Cli/dotnet/SlnProjectExtensions.cs
marcpopMSFT: I confirmed with RichardS that everything in the nested project section is a project (either a solution folder or project but in the solution file itself, they are all considered projects).

marcpopMSFT: From talking to Stanton and reviewing the code, I think that the NestedPropertySets are only used for the `ProjectConfigurationPlatforms` section. You can have the configuration and the ActiveCfg property nested in a single line. https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Cli/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs#L88

You can see where we load property sets we expect to have them in a form that includes a . which fits with the above assumption: https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Cli/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs#L581

So in the case of the NestedProjects section, using Properties and ignoring NestedPropertySets makes sense.

PR #18533, opened 2021-06-25
[API Compat] Add support to specify custom names for left and right on errors

### LINKED ISSUE #17495: API Compat should support custom names for left and right in diagnostic messages
Currently all the API Compat diagnostic messages hardcode `left` and `right` to refer to the compared assets. However we should support custom naming so that messages can be clearer to the customer of what is that it is actually comparing and make the diagnostic more actionable. 

When we add this support, the Package Validation should use this setting and remove `left` and `right` from some of the error messages we have in order to help make this clearer, i.e
https://github.com/dotnet/sdk/blob/10df104084a8e9c60ce66a2d2518d38a18170a3d/src/Compatibility/Microsoft.DotNet.PackageValidation/Resources.resx#L130

cc: @ericstj @Anipik

safern:
Fixes: https://github.com/dotnet/sdk/issues/17495

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
@@ -41,7 +41,7 @@ namespace Microsoft.DotNet.ApiCompatibility
         /// Callback function to get the <see cref="ComparingSettings"/> to be used when creating the settings to get the differences.
         /// This callback is called at the beginning of every <see cref="GetDifferences"/> overload.
         /// </summary>
-        public Func<ComparingSettings> GetComparingSettings { get; set; }
+        public Func<string, string[], ComparingSettings> GetComparingSettings { get; set; }
 
         /// <summary>
         /// Get's the differences when comparing Left vs Right based on the settings at the moment this method is called.
@@ -50,7 +50,7 @@ namespace Microsoft.DotNet.ApiCompatibility
         /// <param name="left">Left symbols to compare against.</param>
         /// <param name="right">Right symbols to compare against.</param>
         /// <returns>List of found differences.</returns>
-        public IEnumerable<CompatDifference> GetDifferences(IEnumerable<IAssemblySymbol> left, IEnumerable<IAssemblySymbol> right)
+        public IEnumerable<CompatDifference> GetDifferences(IEnumerable<IAssemblySymbol> left, IEnumerable<IAssemblySymbol> right, string leftName = null, string rightName = null)
         {
             if (left == null)
             {
@@ -62,7 +62,7 @@ namespace Microsoft.DotNet.ApiCompatibility
                 throw new ArgumentNullException(nameof(right));
             }
 
-            AssemblySetMapper mapper = new(GetComparingSettingsCore());
+            AssemblySetMapper mapper = new(GetComparingSettingsCore(leftName, new[] { rightName }));
             mapper.AddElement(left, ElementSide.Left);
             mapper.AddElement(right, ElementSide.Right);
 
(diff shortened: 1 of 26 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -41,7 +41,7 @@ public class ApiComparer
         /// Callback function to get the <see cref="ComparingSettings"/> to be used when creating the settings to get the differences.
         /// This callback is called at the beginning of every <see cref="GetDifferences"/> overload.
         /// </summary>
-        public Func<ComparingSettings> GetComparingSettings { get; set; }
+        public Func<string, string[], ComparingSettings> GetComparingSettings { get; set; }
```
Anipik, line 44: add comment about the first 2 type parameters

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/CustomSideNameTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            IList<ElementContainer<IAssemblySymbol>> right = SymbolFactory.GetElementContainersFromSyntaxes(rightSyntaxes);
+
+            IEnumerable<(MetadataInformation, MetadataInformation, IEnumerable<CompatDifference>)> differences =
+                differ.GetDifferences(left, right);
+
+            int i = 0;
+            foreach ((MetadataInformation, MetadataInformation, IEnumerable<CompatDifference> differences) diff in differences)
+            {
+                Assert.Single(diff.differences);
+                AssertNames(diff.differences.First(), left.MetadataInformation.AssemblyId, right[i++].MetadataInformation.AssemblyId);
```
Anipik, line 202: what will be the value of left and right assemblyId ? will it be the default value ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/CustomSideNameTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+
+        private void AssertNames(CompatDifference difference, string expectedLeftName, string expectedRightName, bool leftFirst = true)
```
Anipik, line 206: would it make sense for compatDifference to have  left and right assembly names as fields ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/CustomSideNameTests.cs
safern: No, `left.MetadataInformation.AssemblyId` is set in this test in L191: https://github.com/dotnet/sdk/pull/18533/files#diff-3aba641bc1a70ca2e3c5f1781db2e432c9e17665dd52d91a5f2de0214bf4bf7fR191

`right.MetadataInformation.AssemblyId` will be `runtime-n`: https://github.com/dotnet/sdk/blob/f2b190a84a31e7f836129b6c6524974af2e5bbc2/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs#L37

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/CustomSideNameTests.cs
safern: I thought about that, and we might need it if we use the suppression engine at API Compat level, but without that, I don't think it is very relevant as that would just be information that we duplicate across all compat differences for the same tree. 

I think I would rather not store it there for the time being as the consumer already know what the values of left and right are for all the compat differences they get are.

Anipik (approved):

PR #18537, opened 2021-06-25
Pull language version from compilation parse options

captainsafia:
Since we set the lang version in the SDK, we don't need to reconstruct it here.

Follow-up to undo all this later: https://github.com/dotnet/aspnetcore/issues/33848

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
@@ -3,6 +3,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.CSharp;
 
 namespace Microsoft.NET.Sdk.Razor.SourceGenerators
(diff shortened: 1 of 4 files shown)


TanayParikh (commented):
LGTM

TanayParikh (approved):

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
```diff
@@ -35,5 +36,10 @@ internal class RazorSourceGenerationOptions
         /// </para>
         /// </summary>
         public bool SuppressRazorSourceGenerator { get; set; } = false;
+
+        /// <summary>
+        /// Gets the CSharp language version currently used by the compilation.
+        /// </summary>
```
pranavkm, line 42: LangVersion.10? Latest preview is a bad default to pick no?

captainsafia: Generally, yes. I only set it as the default here since that's what the incremental RSG depends on. Follow-up to undo this is at: https://github.com/dotnet/aspnetcore/issues/33848

PR #18545, opened 2021-06-25
Fix null ref in Apply if a WebSocket connection to the browser does not exist

pranavkm:


pranavkm:
FYI @stephentoub

stephentoub:
Thanks, @pranavkm.  I assume this is the cause of the exception I was seeing from dotnet watch?

pranavkm:
>  I assume this is the cause of the exception I was seeing 

Yup!

PR #18579, opened 2021-06-28
Re-enabled full framework tests for some Razor scenarios

captainsafia:


captainsafia:
Need to do this once Helix machines have the latest VS 2020 builds.

PR #18587, opened 2021-06-28
[release/6.0-preview6] [API Compat] Add support to specify custom names for left and right on errors

safern:
Port of: https://github.com/dotnet/sdk/pull/18533

## Description
Currently with Package Validation and API Compat if there is a breaking change the error would hardcode `left` and `right` instead of the path to the assemblies being compared within the package. i.e:

```
error CP002: Member 'F:A.B.C' exists on left but not on right
```

When PackageValidation uses API Compat it compares multiple assemblies within the package depending on which runtime asset applies to a compile asset, so all errors would say there is a member missing on `right` that is present on `left` when `left` and `right` could mean different things depending on the assembly. 

With this change the errors would now be shown as:
```
error CP002: Member 'F:A.B.C' exists on lib/net6.0/a.dll but not on runtimes/win-x64/a.dll
```

## Customer Impact

This makes it easier for customers to understand where the errors are and how to fix them. Also, it aligns with the error suppression which uses the assembly paths within the package to suppress these errors. 

## Testing

Added unit tests and also manually tested with sample projects.

## Risk
Low, test coverage is good and it just changes some error strings.


safern:
Closing as we decided not to take this change to not reset the build.

Issue #18607, opened 2021-06-29
`dotnet sln add` and `dotnet sln remove` can fail with exit code 0 and can partially succeed


ZeroErrors:
Related to: https://github.com/dotnet/sdk/issues/17686

## Description
The implementation of `dotnet sln add` and `dotnet sln remove` allows them to exit with an exit code of 0 even when failing and also allows them to partially succeed.

For example if you run something like `dotnet sln add File.txt Lib/Lib.csproj` it will succeed adding the `Lib/Lib.csproj` to the solution file and print an message about `File.txt` not being a valid project and then will exit with an exit code of 0.

## Steps to reproduce
```
dotnet new sln
dotnet new console --output src/todo-app
```
Create `some-file.txt`

Fails but exits with exit code 0: `dotnet sln add some-file.txt`
Partially succeeds: `dotnet sln add some-file.txt src/todo-app`

Fails but exits with exit code 0: `dotnet sln remove some-file.txt`
Partially succeeds: `dotnet sln remove some-file.txt src/todo-app`
Running `dotnet sln remove src/todo-app` twice will fail on the second run but exit with exit code 0.

Tested on both main (https://github.com/dotnet/sdk/commit/2c011f2aa7a91a386430233d5797452ca0821ed3) and release 6.0.100-preview.5.21302.13

eriawan:
@ZeroErrors 
I agree, this is closely related to my previously submitted issue of #17686

PR #18608, opened 2021-06-29
dotnet sln: Add error message for misplaced solution argument

### LINKED ISSUE #17686: Invalid parameter order when adding a project to sln using `dotnet sln add` should be validated
## General Information
The `dotnet sln add` should respect the order of parameter passed. If the order of the passed parameters are incorrect or invalid, it should display error and the command should not be executed successfully.
Current `dotnet sln add` executed successfully when invalid order of parameters are passed.

Based on the current help from `dotnet sln add -h` the parameters are ordered using this syntax:
`dotnet sln <SLN_FILE> add [options] <PROJECT_PATH>`
taken from:
```
D:\samplenet50>dotnet sln add -h
Usage: dotnet sln <SLN_FILE> add [options] <PROJECT_PATH>

Arguments:
  <SLN_FILE>       The solution file to operate on. If not specified, the command will search the current directory for one.
  <PROJECT_PATH>   The paths to the projects to add to the solution.

Options:
  --in-root               Place project in root of the solution, rather than creating a solution folder.
  -s, --solution-folder   The destination solution folder path to add the projects to.
  -h, --help              Show command line help.
```

## SDK versions used

- [x] 3.1 SDK: 3.1.403
- [x] 5.0 SDK: 5.0.103, 5.0202, 5.0.203

## Operating system

- Windows 10 1909
- Windows 10 2004
- Ubuntu 20 on WSL

## Steps to reproduce

The steps are:

1. Create new sln using `dotnet new sln`, example: `dotnet new sln Solution50`
2. Create any kind of new project. For example: `dotnet new mstest -n UnitTest01`
3. Add the created new project into the previously created sln above, with invalid order of the parameters passed. Example: `dotnet sln add Solution50.sln UnitTest01\UnitTest01.csproj`

Expected result:
The project should not be added and it should display error with meaningful error message.

Actual result:
The project added into the solution, but it still display error message.
```
Invalid project `D:\samplenet50\Solution50.sln`. The project file could not be loaded. Data at the root level is invalid. Line 2, position 1.  D:\samplenet50\Solution50.sln.
Project `UnitTest01\UnitTest01.csproj` added to the solution.
```

Sample:

image

ZeroErrors:
Fixes https://github.com/dotnet/sdk/issues/17686
(Created a new issue for one of the causes not solved in this PR https://github.com/dotnet/sdk/issues/18607)

## Change log
- Added basic check to provide more descriptive message when solution argument is misplaced in `dotnet sln add` and `dotnet sln remove`.
- Added basic command suggestion output to help guide the user.

## Testing
Unit tests included.

To test manually setup a solution and a project something like
```
dotnet new sln
dotnet new console --output src/todo-app
dotnet sln add src/todo-app --solution-folder todo
```

Then running `dotnet sln add <project>.sln src/todo-app` will output
```
Solution argument 'App.sln' is misplaced.
Did you mean?
  dotnet sln <project>.sln add src/todo-app
```
and then the help text


src/Cli/dotnet/commands/dotnet-sln/remove/Program.cs
```diff
@@ -18,16 +18,27 @@ internal class RemoveProjectFromSolutionCommand : CommandBase
...
 
-            _fileOrDirectory = parseResult.ValueForArgument<string>(SlnCommandParser.SlnArgument);
+            var slnFile = _arguments.FirstOrDefault(path => path.EndsWith(".sln"));
+            if (slnFile != null)
+            {
+                var projectArgs = string.Join(" ", _arguments.Where(path => !path.EndsWith(".sln")));
+                throw new GracefulException(new string[]
+                {
+                    string.Format(CommonLocalizableStrings.SolutionArgumentMisplaced, slnFile),
+                    CommonLocalizableStrings.DidYouMean,
```
sfoslund, line 38: @KathleenDollard what do you think of the user experience here? The 'did you mean...' message feels odd to me, I think if anything we should provide an error message educating the user on the difference between a sln and proj as well as print the help here.

src/Tests/dotnet-sln.Tests/GivenDotnetSlnRemove.cs
```diff
@@ -735,5 +736,26 @@ public void WhenProjectIsRemovedThenDependenciesOnProjectAreAlsoRemoved()
             File.ReadAllText(solutionPath)
                 .Should().BeVisuallyEquivalentTo(ExpectedSlnContentsAfterRemoveProjectWithDependencies);
         }
+
+        [Fact]
+        public void WhenSolutionIsPassedAsProjectItPrintsSuggestionAndUsage()
+        {
+            var projectDirectory = _testAssetsManager
+                .CopyTestAsset("TestAppWithSlnAndCsprojFiles", identifier: "WhenSolutionIsPassedAsProject")
```
sfoslund, line 744: I don't think the identifier is needed here (or any of the other tests)

src/Cli/dotnet/commands/dotnet-sln/add/Program.cs
```diff
@@ -43,27 +49,39 @@ internal class AddProjectToSolutionCommand : CommandBase
             {
                 _relativeRootSolutionFolders = null;
             }
+
+            var slnFile = _arguments.FirstOrDefault(path => path.EndsWith(".sln"));
```
sfoslund, line 53: We can factor this and above arguments parsing and checking count into a private method called something like "ParseAndValidateArguments"

src/Cli/dotnet/commands/dotnet-sln/add/Program.cs
```diff
@@ -43,27 +49,39 @@ internal class AddProjectToSolutionCommand : CommandBase
...
+            if (slnFile != null)
+            {
+                string args;
+                if (_inRoot) args = $"--{SlnAddParser.InRootOption.Name} ";
```
sfoslund, line 57: nit: avoid one line if statements to keep code style consistency in the repo (Although I think we'll probably want to remove this anyways and instead give an error that doesn't include these args)

src/Cli/dotnet/commands/dotnet-sln/add/Program.cs
```diff
@@ -43,27 +49,39 @@ internal class AddProjectToSolutionCommand : CommandBase
...
 
-            var fullProjectPaths = arguments.Select(p =>
+            var fullProjectPaths = _arguments.Select(p =>
             {
                 var fullPath = Path.GetFullPath(p);
                 return Directory.Exists(fullPath) ?
                     MsbuildProject.GetProjectFileFromDirectory(fullPath).FullName :
                     fullPath;
-            }).ToList();
+            }).ToArray();
```
sfoslund, line 84: Why is this change necessary? I think it should be reverted.

src/Cli/dotnet/commands/dotnet-sln/remove/Program.cs
```diff
@@ -18,16 +18,27 @@ internal class RemoveProjectFromSolutionCommand : CommandBase
...
+
+            _arguments = parseResult.ValueForArgument(SlnRemoveParser.ProjectPathArgument)?.ToArray() ?? (IReadOnlyCollection<string>)Array.Empty<string>();
             if (_arguments.Count == 0)
             {
                 throw new GracefulException(CommonLocalizableStrings.SpecifyAtLeastOneProjectToRemove);
             }
 
-            _fileOrDirectory = parseResult.ValueForArgument<string>(SlnCommandParser.SlnArgument);
+            var slnFile = _arguments.FirstOrDefault(path => path.EndsWith(".sln"));
+            if (slnFile != null)
```
sfoslund, line 32: Same as above, we can factor this section into a method. In fact, we can probably reuse the same method across both commands.

src/Cli/dotnet/commands/dotnet-sln/remove/Program.cs
```diff
@@ -43,7 +54,7 @@ public override int Execute()
                         MsbuildProject.GetProjectFileFromDirectory(fullPath).FullName :
                         fullPath
                 );
-            });
+            }).ToArray();
```
sfoslund, line 57: nit: unnecessary change, revert

sfoslund (commented):
Thanks for your contribution! Overall, I think it makes sense for us to give a better error for this scenario but I'd like @KathleenDollard to give her input on the specific wording of the error. Otherwise the code quality looks good to me, thanks for adding tests!

src/Tests/dotnet-sln.Tests/GivenDotnetSlnRemove.cs
ZeroErrors: Ok will do, wasn't sure. I couldn't find any documentation on how this is supposed to work so have just been guessing.

src/Cli/dotnet/commands/dotnet-sln/add/Program.cs
ZeroErrors: It was part of changes I was working on trying to deal with https://github.com/dotnet/sdk/issues/18607 but it became a deeper problem than I wanted to address in this PR which is why I made a new issue.

I'll revert it and the same change in `dotnet sln remove` since its not needed for this PR.

src/Cli/dotnet/commands/dotnet-sln/remove/Program.cs
ZeroErrors: I personally just find it faster to resolve problems when a tool can figure out and tell me what it expected instead of needing to spend time to figure out what it wanted but I can just remove the "Did you mean?" part if that's not what is wanted.
If you think its worth talking about it can be split out into an issue and discussed there instead maybe?

The help usage gets printed on stdout as part of `GracefulException` getting thrown.

skundlony (approved):

marcpopMSFT:
Catching up on old PRs. I merged with main, made the refactoring that was suggested in the last PR review, fixed the tests, and reviewed the text with PM and aligned it with templating engine text in a similar "suggested command" scenario: https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx#L799

src/Tests/dotnet-sln.Tests/GivenDotnetSlnRemove.cs
nagilson: The identifier is a parameter used for uniquely identifying [Theory] tests vs [Fact] tests because they get put into a flat folder with the test name into a machine that runs them in parallel. Thanks for resolving it. I added documentation for this stuff a few months back as I had the same confusion!

src/Tests/dotnet-sln.Tests/GivenDotnetSlnAdd.cs
```diff
@@ -1225,5 +1225,68 @@ public void WhenSolutionFolderIsPassedWithDirectorySeparatorFolderStructureIsCor
 
             return slnContents;
         }
+
+        [Fact]
+        public void WhenSolutionIsPassedAsProjectItPrintsSuggestionAndUsage()
```
nagilson, line 1230: This is somewhat of a nit.. but all 15 lines of these three (really four) tests are identical minus the arguments passed in to them. I think we should instead make a helper method that just takes in the execute arguments and expected error message? Or you could turn this into a theory. I prefer the helper method because it lets you keep the better test names.

src/Cli/dotnet/commands/dotnet-sln/SlnArgumentValidator.cs
```diff
@@ -0,0 +1,59 @@
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.Cli;
```
nagilson, line 3: I appreciate that you refactored this into its own class. Thank you.

nagilson (commented):
This is a really nice to have UX improvement. Thank you for your contribution once again. :) Even if we are pretty late to getting to it! And thanks Marc for giving it a refresh.

nagilson (approved):

marcpopMSFT:

PR #18617, opened 2021-06-29
Show trimmer warnings for WinForms and WPF

### LINKED ISSUE #18474: PublishTrimmed for WinForms must warn during build
Currently the SDK will issue a warning when publishing an app which has `PublishTrimmed=true` and is a WinForms app.
But setting `PublishTrimmed=true` in the project file applies feature switch changes to the app even when building with `dotnet build` and some of the feature switches will cause the app to break.

Repro:
* `dotnet new console`
* Add `PublishTrimmed=true` to the `.csproj`
* Add a text box, this can be done by changing the `Form1.Designer.cs` to look like this:
```C#
        private void InitializeComponent()
        {
            this.textBox1 = new System.Windows.Forms.TextBox();
            this.SuspendLayout();
            // 
            // textBox1
            // 
            this.textBox1.Location = new System.Drawing.Point(223, 105);
            this.textBox1.Name = "textBox1";
            this.textBox1.Size = new System.Drawing.Size(100, 23);
            this.textBox1.TabIndex = 0;

            this.components = new System.ComponentModel.Container();
            this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
            this.ClientSize = new System.Drawing.Size(800, 450);
            this.Controls.Add(this.textBox1);
            this.Text = "Form1";
            this.ResumeLayout(false);
            this.PerformLayout();
        }

        #endregion

        private System.Windows.Forms.TextBox textBox1;
```
* `dotnet run`

Observe that there are no warnings reported.
The app starts but immediately fails with exception that Built-in COM support is disabled.

LakshanF:
WinForms and WPF applications can have the `PublishTrimmed `property set in the project file. This will cause some feature switches to be triggered, one of which is to turn off built-in COM support, since that is not compatible with trimming. But this results in a poor developer experience since building the project will not generate any warnings but trying to run almost any WinForms or WPF applications will generate an exception indicating that built-in COM is turned off.

We have been generating warnings during the publish phase. The fix here is to move the warnings to the build phase

- We will generate separate warnings for WinForms and WPF applications to indicate trimming is not compatible
- The warnings (NETSDK1175, NETSDK1168) mention that trimming is not compatible and give aka.ms links to find out more information
- Created a doc [bug](https://github.com/dotnet/docs/issues/24903) to give more information on why trimming is not a good idea and perhaps directions for the determined (and brave) developers to set the turned off feature switches
- Updated the tests to reflect both build and publish options

Fixes #18474

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -163,6 +163,12 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkWarning Condition="'$(PublishTrimmed)' == 'true' and '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0'"
                    ResourceName="PublishTrimmedRequiresVersion30" />
 
+    <!-- Generate Trimming warnings for WinForms and Wpf applications-->
+    <NetSdkWarning Condition="'$(UseWindowsForms)' == 'true' and '$(PublishTrimmed)' == 'true'"
+                 ResourceName="TrimmingWindowsFormsIsNotSupported" />
+    <NetSdkWarning Condition="'$(UseWpf)' == 'true' and '$(PublishTrimmed)' == 'true'"
+                 ResourceName="TrimmingWpfIsNotSupported" />
+    
   </Target>
 
   <Target Name="_CheckForUnsupportedHostingUsage"
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
@@ -111,10 +111,6 @@ Copyright (c) .NET Foundation. All rights reserved.
                             '$(EnableCompressionInSingleFile)' == 'true' And
                             '$(SelfContained)' != 'true'"
                  ResourceName="CompressionInSingleFileRequiresSelfContained" />
-    <NetSdkWarning Condition="'$(UseWindowsForms)' == 'true' and '$(PublishTrimmed)' == 'true'"
-                 ResourceName="TrimmingWindowsFormsIsNotSupported" />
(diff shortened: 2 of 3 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -163,6 +163,12 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkWarning Condition="'$(PublishTrimmed)' == 'true' and '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0'"
                    ResourceName="PublishTrimmedRequiresVersion30" />
 
+    <!-- Generate Trimming warnings for WinForms and Wpf applications-->
+    <NetSdkWarning Condition="'$(UseWindowsForms)' == 'true' and '$(PublishTrimmed)' == 'true'"
```
agocke, line 167: We wanted to make this an error, right? And make it conditional on an additional internal property, `_SuppressWindowsFormsTrimError` and `_SuppressWpfTrimError`?

LakshanF: I was not sure if we decided definitely on the error/warning option (we've gone back and forth in the earlier discussion on this). 

Do we need an additional property to suppress this when the actual warning can be suppressed (NoWarn: NETSDK1168) if we go with the warning?

This behavior seems a reasonable compromise to me,

- We clearly indicate our recommendation that trimming is not compatible (warning and a hard block with built-in COM disabled) 
- Allow developers who still want to go ahead, to suppress the warning and information to turn on feature switches via our docs (after again stating other options recommending to use `COMWrappers `instead of built-in COM in our docs). Making this option too painful by erroring out seems overly cautious to me given we protect developers who are unaware of trimming problems with our default option above.

sbomer (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
agocke: I agree we don't need an extra property if it's just a warning. Let's at least move forward with the warning since that's the high-order bit.

agocke (approved):

PR #18620, opened 2021-06-29
Update System.CommandLine version

### LINKED ISSUE #16312: Help columns are not distributed in a nice way
Repro:

```
dotnet build -h
```

Produces:
image

`--verbosity` makes this look awful. 

I know this is a fix to System.CommandLine, but I hope we can contribute this fix back to that repo. 

Suggest it looks at the width of the options and moves them to the end of the description if longer than a certain amount. Or just always move to the end of the description. 

Also, I though we were going to get wrapping here. Let's check if we can easily access the System.CommandLine smarts on wrapping. 

@jonsequitur

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/16312

WrappedHelp

### DIFF
--- a/src/Cli/dotnet/Parser.cs
+++ b/src/Cli/dotnet/Parser.cs
@@ -160,9 +160,20 @@ namespace Microsoft.DotNet.Cli
 
         internal class DotnetHelpBuilder : HelpBuilder
         {
-            public DotnetHelpBuilder(IConsole console) : base(console) { }
+            public DotnetHelpBuilder(IConsole console, int maxWidth = int.MaxValue) : base(console, maxWidth) { }
 
-            public static Lazy<HelpBuilder> Instance = new Lazy<HelpBuilder>(() => new DotnetHelpBuilder(new CommandLineConsole()));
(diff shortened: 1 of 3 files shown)


src/Cli/dotnet/Parser.cs
```diff
@@ -160,9 +160,20 @@ internal class CommandLineConsole : IConsole
...
-            public static Lazy<HelpBuilder> Instance = new Lazy<HelpBuilder>(() => new DotnetHelpBuilder(new CommandLineConsole()));
+            public static Lazy<HelpBuilder> Instance = new Lazy<HelpBuilder>(() => {
+                int windowWidth;
+                try
+                {
+                    windowWidth = System.Console.WindowWidth;
+                }
+                catch
+                {
+                    windowWidth = int.MaxValue;
```
joeloff, line 173: This is the recommended value? That's an awfully wide window

joeloff (approved):

src/Cli/dotnet/Parser.cs
sfoslund: Yes, that's the default value system command line uses.

PR #18639, opened 2021-06-30
Import RuntimeIdentifierInference targets after workloads

### LINKED ISSUE #18354: It's not possible for a workload to set a default RuntimeIdentifier
### Description

In Xamarin workloads, we'd like to [set a default RuntimeIdentifier](https://github.com/xamarin/xamarin-macios/issues/10861) for each platform we support.

The problem is that the `Microsoft.NET.RuntimeIdentifierInference.targets` file is loaded before our `WorkloadManifest.targets` file, and executes some logic depending on whether the RuntimeIdentifier is set or not, which means that it's too late if we try to set `RuntimeIdentifer` in our WorkloadManifest.targets file.

And we can't set it earlier (`AutoImports.props`), because that would be a global default, not a platform-specific default.

The workaround I've found is to replicate the behavior of `Microsoft.NET.RuntimeIdentifierInference.targets` that we want in our `WorkloadManifest.targets`, which comes down to something like this:

```xml
  <!-- First set the default RuntimeIdentifier if not already specified. -->
  <PropertyGroup Condition=" '$(RuntimeIdentifier)' == '' And '$(RuntimeIdentifiers)' == '' ">
    <RuntimeIdentifier Condition="'$(_PlatformName)' == 'iOS'">iossimulator-x64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="'$(_PlatformName)' == 'tvOS'">tvossimulator-x64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="'$(_PlatformName)' == 'macOS'">osx-x64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="'$(_PlatformName)' == 'MacCatalyst'">maccatalyst-x64</RuntimeIdentifier>
    <!-- Workaround/hack:
        The Microsoft.NET.RuntimeIdentifierInference.targets file is loaded
        before this file, and executes some logic depending on whether the
        RuntimeIdentifier is set or not. Since RuntimeIdentifier isn't set at
        that point (we're setting it here), we need to replicate the logic in
        the Microsoft.NET.RuntimeIdentifierInference.targets file to make sure
        things work as expected.
    -->
    <_RuntimeIdentifierUsesAppHost>false</_RuntimeIdentifierUsesAppHost>
    <IntermediateOutputPath>$(IntermediateOutputPath)$(RuntimeIdentifier)\</IntermediateOutputPath>
    <OutputPath>$(OutputPath)$(RuntimeIdentifier)\</OutputPath>
  </PropertyGroup>
```

Something more appropriate would be nice...

dsplaisted:
Reorder RuntimeIdentifier inference to go after workload imports, to allow workloads to set a default RuntimeIdentifier.  Fixes #18354

@rolfbjarne @jonathanpeppers @akoeplinger This *looks* like a simple change, let's see if it breaks anything.

@pranavkm this reorders the import of Microsoft.NET.RuntimeIdentifierInference.targets to happen after the import for Microsoft.NET.Sdk.Razor.BeforeCommon.targets.  Do you think this is likely to break the Razor SDK?

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
@@ -60,12 +60,6 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
 
   <!--
-    Use RuntimeIdentifier to determine PlatformTarget.
-    Also, enforce that RuntimeIdentifier is always specified for .NETFramework executables.
-  -->
-  <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.RuntimeIdentifierInference.targets" />
-
-  <!--
     Import targets from RazorSDK if referenced
     Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.BeforeCommon.targets
     Sdks\Microsoft.NET.Sdk.Razor\build\netstandard2.0\Microsoft.NET.Sdk.Razor.BeforeCommon.targets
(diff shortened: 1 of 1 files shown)


pranavkm:
> this reorders the import of Microsoft.NET.RuntimeIdentifierInference.targets to happen after the import for Microsoft.NET.Sdk.Razor.BeforeCommon.targets

We currently don't use / recommend the `$TFM-$rid` pattern in any of the Web projects, so this seems fine.

lewing:
cc @radical @steveisok

steveisok:
@jonathanpeppers @rolfbjarne Have you tried testing this yet w/ your workloads?

jonathanpeppers:
I think we can go with this, and if something breaks we should be able to use `$(AfterMicrosoftNETSdkTargets)` to make parts of our targets run later.

I don't immediately see anything that would break in the Android workload.

joeloff (approved):
The change looks fine. The workload owners seem to be the most impacted by the change so I'll defer to their judgement

lewing (approved):

radical (approved):

akoeplinger (approved):

PR #18654, opened 2021-06-30
Add workload rollback option

sfoslund:


sfoslund:
Open questions on this PR:

- What should the syntax be here? Right now they are options on the update command (`--print-rollback-definition-only` and `--from-rollback-definition`)
- I've made `--print-rollback-definition-only` hidden since I assume we will be the ones providing the rollback definition when there's a break. Is there any reason we want to expose this option to the user?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
```diff
@@ -473,5 +473,19 @@ public string GetManifestVersion(string manifestId)
...
         }
+
+        public IDictionary<string, string> GetInstalledManifests()
+        {
+            var manifests = new Dictionary<string, string>();
+            foreach ((string manifestId, Stream manifestStream) in _manifestProvider.GetManifests())
+            {
+                using (manifestStream)
+                {
+                    var manifest = WorkloadManifestReader.ReadWorkloadManifest(manifestId, manifestStream);
```
wli3, line 484: @mhutch mentioned we should not directly use WorkloadManifestReader. Is there an API in manifest resolver to do this? if not, maybe we should add this to manifest resolver. And also there should be some nuance in resolver to find manifests, please check with Mikayla on that too.

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -143,6 +145,126 @@ public void GivenWorkloadManifestUpdateItCanCalculateUpdates()
...
+            var manifestDirs = expectedManifestUpdates.Select(manifest => manifest.Item1)
+                .Select(manifest => Path.Combine(installedManifestDir, manifest.ToString(), _manifestFileName))
+                .ToArray();
+            var workloadManifestProvider = new MockManifestProvider(manifestDirs);
+            var nugetDownloader = new MockNuGetPackageDownloader(dotnetRoot);
+            var workloadResolver = WorkloadResolver.CreateForTests(new MockManifestProvider(Array.Empty<string>()), new string[] { dotnetRoot });
+            var manifestUpdater = new WorkloadManifestUpdater(_reporter, workloadManifestProvider, workloadResolver, nugetDownloader, testDir, testDir);
+
+            var manifestUpdates = manifestUpdater.CalculateManifestRollbacks(rollbackDefPath);
+            manifestUpdates.Should().BeEquivalentTo(expectedManifestUpdates);
```
wli3, line 182: should packs be reinstalled too?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -133,6 +134,22 @@ public async Task BackgroundUpdateAdvertisingManifestsWhenRequiredAsync()
             return manifestUpdates;
         }
 
+        public IEnumerable<(ManifestId manifestId, ManifestVersion existingVersion, ManifestVersion newVersion)> CalculateManifestRollbacks(string rollbackDefinitionFilePath)
```
wli3, line 137: And i am not sure if we could end up adding or removing manifests

wli3 (commented):
Could you write a high level doc on how this feature should be used?

I assume the user need to run  `dotnet workload update --print-rollback-definition-only` and pipe the result to a file? And later to run ` dotnet workload update --from-rollback-definition c:\rollbackdef.json` to restore?

We should put the experience in on the topic of Friday's meeting

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
sfoslund: My understanding was that the manifest reader should only be used directly in the workload resolver, but please correct me if I'm wrong.

sfoslund:
> I assume the user need to run dotnet workload update --print-rollback-definition-only and pipe the result to a file? And later to run dotnet workload update --from-rollback-definition c:\rollbackdef.json to restore?

Yes, although I've made `--print-rollback-definition-only` private because my understanding was that we (or the workloads teams) would be providing the rollback definitions when there's a bug, then supply that to the user. So we would be the only ones using that option and the user would just download the file we provide and use it with --from-rollback-definition

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
sfoslund: I account for that here: https://github.com/dotnet/sdk/pull/18654/files#diff-2a3be4a6207f3297255b699d440fe93813a50baad9c37cf883ffe371853adbe0R144, so we error if the set of manifests defined in the rollback definition doesn't match the set that's currently installed.

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
sfoslund: This test is just covering the manifest updater, so packs aren't actually being installed here at all.

wli3 (approved):

wli3:
--print-rollback
--from-rollback-file

and make them both hidden

wli3 (approved):

PR #18655, opened 2021-06-30
Suppress future ILLink analyzer warnings

### LINKED ISSUE #18581: SuppressTrimAnalysisWarnings should also suppress warnings from the Roslyn trim analyzer
We have the property `SuppressTrimAnalysisWarnings` which tells the linker not to warn when `PublishTrimmed=true`. However, the Roslyn analyzer still warns because it is enabled by this code:

https://github.com/dotnet/sdk/blob/1ee1d244257a379b04b3cf8ca8915e2698144ce5/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L59-L60

Setting `SuppressTrimAnalysisWarnings=true` has no effect on the Roslyn analyzer, but it should. We shouldn't have two settings to disable trim analysis warnings. If someone wants to disable the warnings, they should just need to set one property.

cc @vitek-karas @sbomer @agocke @mateoatr @tlakollo

sbomer:
Fixes https://github.com/dotnet/sdk/issues/18581

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
@@ -78,6 +78,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NoWarn>$(NoWarn);IL2103</NoWarn>
     <!-- Multiple methods associated with state machine type -->
     <NoWarn>$(NoWarn);IL2107</NoWarn>
+    <!-- Unannotated type derived from base type with RequiresUnreferencedCode -->
+    <NoWarn>$(NoWarn);IL2109</NoWarn>
     <!-- Reflection access to members with DynamicallyAccessedMembers requirements -->
     <NoWarn>$(NoWarn);IL2110;IL2111;IL2114;IL2115</NoWarn>
     <!-- Reflection access to members with RequiresUnreferencedCode -->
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
```diff
@@ -52,13 +52,36 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <!-- DynamicallyAccessedMembersAttribute used on unsupported member -->
+    <NoWarn>$(NoWarn);IL2097;IL2098;IL2099;IL2106</NoWarn>
+    <!-- Unknown value passed to Expression.Property -->
+    <NoWarn>$(NoWarn);IL2103</NoWarn>
+    <!-- Multiple methods associated with state machine type -->
+    <NoWarn>$(NoWarn);IL2107</NoWarn>
+    <!-- Reflection access to members with DynamicallyAccessedMembers requirements -->
+    <NoWarn>$(NoWarn);IL2110;IL2111;IL2114;IL2115</NoWarn>
+    <!-- Reflection access to members with RequiresUnreferencedCode -->
+    <NoWarn>$(NoWarn);IL2112;IL2113</NoWarn>
```
eerhardt, line 84: Why not `2109`?

```
IL2109 Trim analysis: Type 'type' derives from 'BaseType' which has 'RequiresUnreferencedCodeAttribute'. [message]. [url]
```

sbomer: I left it out because that one is going away based on recent discussions, but on second thought it's probably better to suppress it until that work is finished.

eerhardt (approved):
Thanks!

PR #18656, opened 2021-07-01
    Rename --add-source to source

wli3:
source param in workloads are exclusive. so it should be --source instead of --add-source

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
@@ -21,10 +21,10 @@ namespace Microsoft.DotNet.Cli
                 ArgumentHelpName = LocalizableStrings.ConfigFileOptionName
             };
 
-        public static readonly Option AddSourceOption =
-            new Option<string[]>("--add-source", LocalizableStrings.AddSourceOptionDescription)
+        public static readonly Option SourceOption =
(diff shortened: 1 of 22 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
```diff
@@ -21,10 +21,10 @@ internal static class WorkloadInstallCommandParser
                 ArgumentHelpName = LocalizableStrings.ConfigFileOptionName
             };
 
-        public static readonly Option AddSourceOption =
-            new Option<string[]>("--add-source", LocalizableStrings.AddSourceOptionDescription)
+        public static readonly Option SourceOption =
+            new Option<string[]>(new string[] { "-s", "--source" }, LocalizableStrings.SourceOptionDescription)
```
sfoslund, line 25: If it only accepts a single source, should we change this type to `string` instead of `string[]` so the parser enforces this? With the current config it would allow `--source foo --source bar`- would that be valid?

sfoslund (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
wli3: no, it accepts a list of sources. and use that set of sources exclusively.

--source foo --source bar is allowed. we have that for dotnet restore a long time

PR #18657, opened 2021-07-01
Rename --add-source to source

wli3:
source param in workloads are exclusive. so it should be --source instead of --add-source


wli3:
@marcpopMSFT do we still have time to check in preview 6?

wli3:
this is good to have

marcpopMSFT:
Too late for preview 6. We have the candidate build already.

PR #18658, opened 2021-07-01
Set PackageValidationBaselineName to assemblyName when packageID is not set

Anipik:
Fixes issue in https://github.com/dotnet/runtime/pull/54250

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
@@ -11,7 +11,7 @@
           Condition="'$(IsPackable)' == 'true'">
 
     <PropertyGroup>
-      <PackageValidationBaselineName>$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))</PackageValidationBaselineName>
+      <PackageValidationBaselineName Condition="'$(PackageValidationBaselineName)' == ''">$(PackageId)</PackageValidationBaselineName>
       <PackageValidationBaselinePath Condition="'$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">$([MSBuild]::NormalizePath('$(NuGetPackageRoot)', '$(PackageValidationBaselineName.ToLower())', '$(PackageValidationBaselineVersion)', '$(PackageValidationBaselineName.ToLower()).$(PackageValidationBaselineVersion).nupkg'))</PackageValidationBaselinePath>
       <GenerateCompatibilitySuppressionFile Condition="'$(GenerateCompatibilitySuppressionFile)' == ''">false</GenerateCompatibilitySuppressionFile>
       <_compatibilitySuppressionFilePath>$([MSBuild]::NormalizePath('$(MSBuildProjectDirectory)', 'CompatibilitySuppressions.xml'))</_compatibilitySuppressionFilePath>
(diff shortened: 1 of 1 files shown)


src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -3,6 +3,7 @@
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
 
   <PropertyGroup>
+    <PackageValidationBaselineName Condition="'$(PackageId)' == ''">$(AssemblyName)</PackageValidationBaselineName>
     <PackageValidationBaselineName>$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))</PackageValidationBaselineName>
```
ViktorHofer, line 7:
```suggestion
    <PackageValidationBaselineName Condition="'$(PackageId)' == ''">$(AssemblyName)</PackageValidationBaselineName>
    <PackageValidationBaselineName>$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$([MSBuild]::ValueOrDefault('$(PackageId)'))'))</PackageValidationBaselineName>
```

ViktorHofer (approved):

ViktorHofer:
In which cases is PackageId not set?

Anipik:
> In which cases is PackageId not set?

Its being set late during the sdk evaluation.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -3,7 +3,8 @@
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
 
   <PropertyGroup>
-    <PackageValidationBaselineName>$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))</PackageValidationBaselineName>
+    <PackageValidationBaselineName Condition="'$(PackageId)' == ''">$(AssemblyName)</PackageValidationBaselineName>
```
ViktorHofer, line 6:
```suggestion
```

ViktorHofer:
> Its being set late during the sdk evaluation.

Doesn't that mean that PackageId is never used then?

Anipik:
> Doesn't that mean that PackageId is never used then?

unless its a global property

ViktorHofer:
> unless its a global property

which is never the case usually.

Anipik:
> which is never the case usually.

setting the packageID in the csproj would be another scenario.

ViktorHofer:
> setting the packageID in the csproj would be another scenario.

That's also very rarely the case for projects unless they would want to override it.

You could easily avoid this sequencing issue by inlining the `PackageValidationBaselineName` value into the two locations where it's used:
- The PackageDownload item
- In the task as a property later

By inlining it, you will only access the `PackageId` value after it is set by NuGet.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -2,19 +2,16 @@
...
-    <PackageDownload Include="$(PackageValidationBaselineName)" Version="[$(PackageValidationBaselineVersion)]" />
+    <PackageDownload Include="$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))" Version="[$(PackageValidationBaselineVersion)]" />
   </ItemGroup>
 
   <Target Name="RunPackageValidation"
           AfterTargets="Pack"
           Condition="'$(IsPackable)' == 'true'">
 
     <PropertyGroup>
+      <PackageValidationBaselineName>$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))</PackageValidationBaselineName>
```
ViktorHofer, line 14:
```suggestion
      <PackageValidationBaselineName Condition="'$(PackageValidationBaselineName)' == ''">$(PackageId)</PackageValidationBaselineName>
```

ViktorHofer:
LGTM module the one remaining feedback item.

ViktorHofer (approved):

safern (approved):

PR #18670, opened 2021-07-01
Add MetadataUpdaterSupport property

sbomer:
While writing some docs I noticed that the feature switch added in https://github.com/dotnet/runtime/pull/54590 hadn't been defined in the SDK yet. Also adds some existing switches to the test.


eerhardt (approved):
Looks good. Thanks for catching this and adding it.

LakshanF (approved):
Thanks for adding the COM and native host properties to this test as well. I'm a little confused with the difference between this test and `GivenThatWeWantToRunILLink `and am happy you are covering this one.

sbomer:
I think this test is supposed to cover all RuntimeHostConfigurationOptions, whether trim-enabled or not.

PR #18672, opened 2021-07-01
Initial Support for PredefinedCultureOnly property

tarekgh:
We are changing the Invariant Globalization mode behavior in .NET 6.0 to force it to throw exception by default when creating non-Invariant cultures https://github.com/dotnet/runtime/pull/54247. The behavior can be controlled by a config switch `System.Globalization.PredefinedCulturesOnly` that we need to support it through the SDK to have msbuild property.

### DIFF
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -13,6 +13,7 @@
     <DebuggerSupport>true</DebuggerSupport>
     <EventSourceSupport>false</EventSourceSupport>
     <InvariantGlobalization>true</InvariantGlobalization>
+    <PredefinedCulturesOnly>true</PredefinedCulturesOnly>
     <ConcurrentGarbageCollection>false</ConcurrentGarbageCollection>
     <ServerGarbageCollection>true</ServerGarbageCollection>
     <RetainVMGarbageCollection>false</RetainVMGarbageCollection>
@@ -59,7 +60,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <CopyToPublishDirectory>Never</CopyToPublishDirectory>
     </Content>
-    
+
     <None Include="NoneCopyOutputAlways.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
@@ -74,7 +75,7 @@
     <ProjectReference Include="../TestLibrary/TestLibrary.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <!-- 
+    <!--
       The TestLibrary has a hard dependency on Newtonsoft.Json.
       The TestApp has a PrivateAssets=All dependency on Microsoft.Extensions.DependencyModel.
       Microsoft.Extensions.DependencyModel depends on Newtonsoft.Json.
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -59,7 +59,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <CopyToPublishDirectory>Never</CopyToPublishDirectory>
     </Content>
-
+    
     <None Include="NoneCopyOutputAlways.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
@@ -74,7 +74,7 @@
     <ProjectReference Include="../TestLibrary/TestLibrary.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <!--
+    <!-- 
       The TestLibrary has a hard dependency on Newtonsoft.Json.
       The TestApp has a PrivateAssets=All dependency on Microsoft.Extensions.DependencyModel.
       Microsoft.Extensions.DependencyModel depends on Newtonsoft.Json.
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -13,7 +13,6 @@
     <DebuggerSupport>true</DebuggerSupport>
     <EventSourceSupport>false</EventSourceSupport>
     <InvariantGlobalization>true</InvariantGlobalization>
-    <PredefinedCulturesOnly>true</PredefinedCulturesOnly>
     <ConcurrentGarbageCollection>false</ConcurrentGarbageCollection>
     <ServerGarbageCollection>true</ServerGarbageCollection>
     <RetainVMGarbageCollection>false</RetainVMGarbageCollection>
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -25,6 +25,7 @@
     <_EnableConsumingManagedCodeFromNativeHosting>false</_EnableConsumingManagedCodeFromNativeHosting>
     <EnableCppCLIHostActivation>false</EnableCppCLIHostActivation>
     <EnableUnsafeBinaryFormatterSerialization>false</EnableUnsafeBinaryFormatterSerialization>
+    <PredefinedCulturesOnly>true</PredefinedCulturesOnly>
     <TieredCompilation>true</TieredCompilation>
     <TieredCompilationQuickJit>true</TieredCompilationQuickJit>
     <TieredCompilationQuickJitForLoops>true</TieredCompilationQuickJitForLoops>
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -25,6 +25,7 @@
     <_EnableConsumingManagedCodeFromNativeHosting>false</_EnableConsumingManagedCodeFromNativeHosting>
     <EnableCppCLIHostActivation>false</EnableCppCLIHostActivation>
     <EnableUnsafeBinaryFormatterSerialization>false</EnableUnsafeBinaryFormatterSerialization>
+    <PredefinedCulturesOnly>true</PredefinedCulturesOnly>
     <TieredCompilation>true</TieredCompilation>
     <TieredCompilationQuickJit>true</TieredCompilationQuickJit>
     <TieredCompilationQuickJitForLoops>true</TieredCompilationQuickJitForLoops>
@@ -59,7 +60,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <CopyToPublishDirectory>Never</CopyToPublishDirectory>
     </Content>
-    
+
     <None Include="NoneCopyOutputAlways.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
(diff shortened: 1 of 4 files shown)


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAProjectWithAllFeatures.cs
```diff
@@ -101,7 +101,7 @@ public void It_publishes_the_project_correctly(string targetFramework, string []
     }
 }");
             baselineConfigJsonObject["runtimeOptions"]["tfm"] = targetFramework;
-            baselineConfigJsonObject["runtimeOptions"]["framework"]["version"] = 
+            baselineConfigJsonObject["runtimeOptions"]["framework"]["version"] =
```
dsplaisted, line 104: Can you revert the whitespace changes in this file, since nothing else is changing in it?

src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
```diff
@@ -59,7 +59,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <CopyToPublishDirectory>Never</CopyToPublishDirectory>
     </Content>
-    
+
```
dsplaisted, line 62: Please revert the whitespace changes in this file.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
```diff
@@ -841,6 +841,62 @@ public void It_marks_package_references_as_externally_resolved(bool? markAsExter
...
+
+            testProj.AdditionalProperties["InvariantGlobalization"] = invariantValue ? "true" : "false";
+
+            if (definePredefinedCulturesOnly)
+            {
+                testProj.AdditionalProperties["PredefinedCulturesOnly"] = predefinedCulturesOnlyValue ? "true" : "false";
+            }
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProj, targetFramework);
+            File.WriteAllText(Path.Combine(testAsset.Path, testProj.Name, $"{testProj.Name}.cs"), @"
```
dsplaisted, line 872: Instead of writing directly to the file, you can do this to the `testProj` before you create it:

```c#
testProj.SourceFiles["Program.cs"] = @"
// Your code here
";
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
```diff
@@ -841,6 +841,62 @@ public void It_marks_package_references_as_externally_resolved(bool? markAsExter
...
+                using System;
+                using System.Reflection;
+                class Program
+                {
+                    static void Main(string[] args)
+                    {
+                        bool invariant = false;
+                        bool predefinedCulturesOnly = false;
+                        try { invariant = (bool) typeof(object).Assembly.GetType(""System.Globalization.GlobalizationMode"").GetProperty(""Invariant"", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null); } catch {}
+                        try { predefinedCulturesOnly = (bool) typeof(object).Assembly.GetType(""System.Globalization.GlobalizationMode"").GetProperty(""PredefinedCulturesOnly"", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null); } catch {}
```
dsplaisted, line 882: Is there a better way to check this than by using private reflection?  Would reading the runtimeconfig file work?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
```diff
@@ -841,6 +841,62 @@ public void It_marks_package_references_as_externally_resolved(bool? markAsExter
...
+
+            var buildCommand = new BuildCommand(Log, Path.Combine(testAsset.Path, testProj.Name));
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
+
+            var runCommand = new RunExeCommand(Log, Path.Combine(buildCommand.GetOutputDirectory(targetFramework).FullName, $"{testProj.Name}.exe"));
+            var stdOut = runCommand.Execute().StdOut.Split(Environment.NewLine.ToCharArray()).Where(line => !string.IsNullOrWhiteSpace(line));
+            stdOut.Should().BeEquivalentTo(expectedPredefinedCulturesOnlyValue);
```
dsplaisted, line 897: `BeEquivalentTo` ignores the ordering of the elements in the collections, which is probably not what you want.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
```diff
@@ -841,6 +841,62 @@ public void It_marks_package_references_as_externally_resolved(bool? markAsExter
...
+                        bool predefinedCulturesOnly = false;
+                        try { invariant = (bool) typeof(object).Assembly.GetType(""System.Globalization.GlobalizationMode"").GetProperty(""Invariant"", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null); } catch {}
+                        try { predefinedCulturesOnly = (bool) typeof(object).Assembly.GetType(""System.Globalization.GlobalizationMode"").GetProperty(""PredefinedCulturesOnly"", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null); } catch {}
+
+                        Console.WriteLine($""{invariant},{predefinedCulturesOnly}"");
+                    }
+                }
+            ");
+
+            var buildCommand = new BuildCommand(Log, Path.Combine(testAsset.Path, testProj.Name));
```
dsplaisted, line 889:
```suggestion
            var buildCommand = new BuildCommand(testAsset);
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
```diff
@@ -841,6 +841,62 @@ public void It_marks_package_references_as_externally_resolved(bool? markAsExter
...
+            };
+
+            testProj.AdditionalProperties["InvariantGlobalization"] = invariantValue ? "true" : "false";
+
+            if (definePredefinedCulturesOnly)
+            {
+                testProj.AdditionalProperties["PredefinedCulturesOnly"] = predefinedCulturesOnlyValue ? "true" : "false";
+            }
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProj, targetFramework);
```
dsplaisted, line 871: This will create a separate test folder for each instance of the theory, and probably fix test failures in CI.

```suggestion
            var testAsset = _testAssetsManager.CreateTestProject(testProj, identifier: $"{targetFramework}{invariantValue}{predefinedCulturesOnlyValue}{definePredefinedCulturesOnly}");
```

src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
eerhardt, line 62: We should be setting the `<PredefinedCulturesOnly>` property in the top PropertyGroup of this file.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -72,6 +72,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <UserProfileRuntimeStorePath Condition="'$(UserProfileRuntimeStorePath)' == ''">$(_DefaultUserProfileRuntimeStorePath)</UserProfileRuntimeStorePath>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0'))">
+    <PredefinedCulturesOnly Condition="'$(InvariantGlobalization)' == 'true' And '$(PredefinedCulturesOnly)' == ''">true</PredefinedCulturesOnly>
```
eerhardt, line 76: (nit) it is normal convention to put the "has this property already been set" check first:

```xml
<PredefinedCulturesOnly Condition="'$(PredefinedCulturesOnly)' == '' and '$(InvariantGlobalization)' == 'true'">true</PredefinedCulturesOnly>
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
tarekgh: Would it matter if I have just one element in the collection? what you suggest to use instead?

dsplaisted: Never mind, I thought you were splitting on commas instead of newlines.  So the code you have works.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -70,6 +70,12 @@ Copyright (c) .NET Foundation. All rights reserved.
     <_DefaultUserProfileRuntimeStorePath Condition="$([MSBuild]::IsOSPlatform(`Windows`))">$(USERPROFILE)</_DefaultUserProfileRuntimeStorePath>
     <_DefaultUserProfileRuntimeStorePath>$([System.IO.Path]::Combine($(_DefaultUserProfileRuntimeStorePath), '.dotnet', 'store'))</_DefaultUserProfileRuntimeStorePath>
     <UserProfileRuntimeStorePath Condition="'$(UserProfileRuntimeStorePath)' == ''">$(_DefaultUserProfileRuntimeStorePath)</UserProfileRuntimeStorePath>
+    <!-- Prevent including the PredefinedCulturesOnly in the runtime config file in .NET 5.0 and below even if it is defined -->
+    <PredefinedCulturesOnly Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' Or $([MSBuild]::VersionLessThan($(TargetFrameworkVersion), '6.0'))"></PredefinedCulturesOnly>
```
tarekgh, line 74: @dsplaisted I added this line too. Please let me know if it make sense to you.

eerhardt: I don't believe this is necessary and think it should be removed. If someone is setting this property in a previous TFM, it won't do any harm. It can get in the runtime config for a .NET 5 app, and nothing will read it.

tarekgh: That make sense too. I was trying to weigh if it will be useful to have. I just run into that when I was adding the test case. I'll remove it then.

src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
tarekgh: @dsplaisted I have returned back the changes in GivenThatWeWantToPublishAProjectWithAllFeatures.cs and TestApp as @eerhardt suggested to keep it to test the behavior when enabling the property in down-level version even if it will not have any runtime effect.

eerhardt (approved):
:shipit:

tarekgh:
@dsplaisted are we good to merge? or you have more comments?

dsplaisted:
@tarekgh I've merged it

PR #18688, opened 2021-07-02
Donot attempt to download baseline package when DisablePackageBaselineValidation is set 

Anipik:
- Donot download package if DisablePackageBaselineValidation is set.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
@@ -2,7 +2,7 @@
 <Project>
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
 
-  <ItemGroup Condition="'$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
+  <ItemGroup Condition="'$(DisablePackageBaselineValidation)' != 'true' and '$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
     <PackageDownload Include="$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))" Version="[$(PackageValidationBaselineVersion)]" />
   </ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
```diff
@@ -40,7 +40,7 @@ public BaselinePackageValidator(Package baselinePackage, string noWarn, (string,
         /// <param name="package">Nuget Package that needs to be validated.</param>
         public void Validate(Package package)
         {
-            foreach (ContentItem baselineCompileTimeAsset in _baselinePackage.RefAssets)
```
safern, line 43: Why do you need this change?

safern (approved):

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -2,7 +2,7 @@
 <Project>
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
 
-  <ItemGroup Condition="'$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
+  <ItemGroup Condition="$(DisablePackageBaselineValidation) != 'true' and '$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
```
safern, line 5: You are missing quotes around the property. Without the quotes this will always be true.

safern (changes requested):

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
Anipik: For most of the cases ref assets should work fine, but as we are moving away from having ref assets in package, i thought it might be best to replace it by compileTimeAssets here.

Anipik: On second thought adding it will decrease the performance as it will add duplicates in case of no ref assemblies in the package. reverted it back.

ViktorHofer (approved):

Issue #18661, opened 2021-07-01
Fail to install workload "microsoft-net-sdk-blazorwebassembly-aot" 


Andy-Win:
**Steps**:
1. Install latest Dev17.0 Preview2 from main branch.
2. Install .NET SDK 6.0 Preview 6 : build 6.0.100-preview.6.21328.10 from [Pipelines - Run 20210628.3-6.0.100-preview.6.21328.10-95778 artifacts (azure.com)](https://dev.azure.com/dnceng/internal/_build/results?buildId=1210635&view=artifacts&type=publishedArtifacts).
3. Add feed: https://pkgs.dev.azure.com/dnceng/public/_packaging/6.0.100-preview.6.21328.10-shipping/nuget/v3/index.json.
4. Run "dotnet workload install microsoft-net-sdk-blazorwebassembly-aot" as administrator.

**ACTUAL**
Failed to update the advertising manifest microsoft.net.workload.mono.toolchain: The file is not a valid nupkg. File path: C:\Users\v-wajian\AppData\Local\Temp\dotnet-sdk-advertising-temp\microsoft.net.workload.mono.toolchain.manifest-6.0.100\6.0.0-preview.6.21325.12\microsoft.net.workload.mono.toolchain.manifest-6.0.100.6.0.0-preview.6.21325.12.nupkg.
Installing pack Microsoft.NET.Runtime.WebAssembly.Sdk version 6.0.0-preview.6.21325.12...
Workload installation failed, rolling back installed packs...
Rolling back pack Microsoft.NET.Runtime.WebAssembly.Sdk installation...
Workload installation failed: The file is not a valid nupkg. File path: C:\Users\v-wajian\AppData\Local\Temp\dotnet-sdk-advertising-temp\microsoft.net.runtime.webassembly.sdk\6.0.0-preview.6.21325.12\microsoft.net.runtime.webassembly.sdk.6.0.0-preview.6.21325.12.nupkg
![image](https://user-images.githubusercontent.com/35762055/124095612-6b92ee00-da8c-11eb-9aa8-799e8300e1a5.png)

**Full Error Text**
![image](https://user-images.githubusercontent.com/8246794/124131315-4c508c80-da9d-11eb-9cc0-9d1892f6a332.png)

steveisok:
"The file is not a valid nupkg" appears for every single manifest the sdk tries to update.  Pure speculation, but did something go wrong with the signing?

Andy-Win:
[...]

PR #18701, opened 2021-07-02
Fix signing verification locking issues

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/18661

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
@@ -145,6 +145,9 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             }
             finally
             {
+                GC.Collect();
+                GC.WaitForPendingFinalizers();
+
                 // Delete leftover dirs and files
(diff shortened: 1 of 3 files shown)


joeloff (approved):

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -145,6 +145,9 @@ public void InstallWorkloadPack(PackInfo packInfo, SdkFeatureBand sdkFeatureBand
             }
             finally
             {
+                GC.Collect();
+                GC.WaitForPendingFinalizers();
```
Pilchie, line 149: Note that in general, this pattern isn't sufficient, if you have chains of finalizable objects, you might need to do this `n` times where `n` is the length of the longest possible chain.

Is there no disposable based system that can be used here instead?

joeloff: yeah, this is not ideal. We're seeing lots of issues with quick, successive file operations where handles are not being released, e.g. we extract a file, read it, close it, delete it, and then delete fails, even after explicitly calling close() before deleting it.

PR #18709, opened 2021-07-03
Add assembly identity Api Check

Anipik:
Assembly Identity rules

- assemblies should be not null
- names must be exactly same
- cultures must be same
- version of the right should be >= left (rightVersion == leftVersion under strict mode)
- Non strict mode
   leftKeyTokenNull or retargetable flag set => always true
   else leftKeyTokenNull == rightKeyTokenNull
- same rules for right under strict mode

added tests for each of these cases in strict and non-strict mode.


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -117,9 +117,24 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="AssemblyNameDoesNotExist" xml:space="preserve">
+    <value>Assembly with name "{0}" doesnot exist at {1}.</value>
```
safern, line 121: Let's use single quotes for the names, to be consistent with other errors.

```suggestion
    <value>Assembly with name '{0}' does not exist at {1}.</value>
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,98 @@
...
+        public override void Initialize(RuleRunnerContext context)
+        {
+            context.RegisterOnAssemblySymbolAction(RunOnAssemblySymbol);
+        }
+
+        private void RunOnAssemblySymbol(IAssemblySymbol left, IAssemblySymbol right, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            if (left == null && right != null)
+            {
+                differences.Add(new CompatDifference(DiagnosticIds.AssemblyIdentityMustMatch, string.Format(Resources.AssemblyNameDoesNotExist, leftName, right.Identity.Name), DifferenceType.Removed, right.Identity.GetDisplayName()));
```
safern, line 24: Do you think this must be a different diagnostic ID?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,98 @@
...
+                return;
+            }
+
+            if (left != null && right == null)
+            {
+                differences.Add(new CompatDifference(DiagnosticIds.AssemblyIdentityMustMatch, string.Format(Resources.AssemblyNameDoesNotExist, rightName, left.Identity.Name), DifferenceType.Added, left.Identity.GetDisplayName()));
+                return;
+            }
+
+            string leftAssemblyName = left.Identity.Name;
```
safern, line 34: Is it worth checking for the name? If the assembly name doesn't match, the mapper won't map them as the same assembly, what I'm trying to say is that when this rule runs, if both left and right != null, that means they have the same name.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            bool isLeftRetargetable = (left.Identity.Flags & AssemblyNameFlags.Retargetable) != 0;
+            bool isRightRetargetable = (right.Identity.Flags & AssemblyNameFlags.Retargetable) != 0;
+            if (!leftAssemblyPublicKeyToken.IsEmpty && !isLeftRetargetable && !leftAssemblyPublicKeyToken.SequenceEqual(rightAssemblyPublicKeyToken))
+            {
+                differences.Add(
+                    new CompatDifference(
+                        DiagnosticIds.AssemblyIdentityMustMatch,
+                        string.Format(Resources.AssemblyPublicKeyTokenDoesNotMatch, GetStringRepresentation(leftAssemblyPublicKeyToken), GetStringRepresentation(rightAssemblyPublicKeyToken), leftName, rightName),
+                        DifferenceType.Changed,
```
safern, line 71: since the differenceType and diagnostic ID is the same across all of these differences, you could add a helper method that takes the message and the symbol/target in, then return the compat difference. Share that method and just call it everywhere to get a new compat difference. Something like: https://github.com/dotnet/sdk/blob/56a6239ba8959df4e7d851923e5022a0fce78805/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/MembersMustExist.cs#L70

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
safern: Never mind, with the entrypoints for single assemblies, this could happen.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,98 @@
...
+                return;
+            }
+
+            string leftAssemblyName = left.Identity.Name;
+            string leftAssemblyCulture = left.Identity.CultureName == string.Empty ? "neutral" : left.Identity.CultureName;
+            Version leftAssemblyVersion = left.Identity.Version;
+            ReadOnlySpan<byte> leftAssemblyPublicKeyToken = left.Identity.PublicKeyToken.AsSpan();
+
+            string rightAssemblyName = right.Identity.Name;
+            string rightAssemblyCulture = right.Identity.CultureName == string.Empty ? "neutral" : right.Identity.CultureName;
```
safern, line 40:
```suggestion
            string rightAssemblyCulture = string.IsNullOrEmpty(right.Identity.CultureName) ? "neutral" : right.Identity.CultureName;
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            if (left != null && right == null)
+            {
+                differences.Add(new CompatDifference(DiagnosticIds.AssemblyIdentityMustMatch, string.Format(Resources.AssemblyNameDoesNotExist, rightName, left.Identity.Name), DifferenceType.Added, left.Identity.GetDisplayName()));
+                return;
+            }
+
+            string leftAssemblyName = left.Identity.Name;
+            string leftAssemblyCulture = left.Identity.CultureName == string.Empty ? "neutral" : left.Identity.CultureName;
```
safern, line 35:
```suggestion
            string leftAssemblyCulture = string.IsNullOrEmpty(left.Identity.CultureName) ? "neutral" : left.Identity.CultureName;
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
safern, line 34: it seems like might be worth storing `left.Identity` and `right.Identity` in local variables? They are used a lot.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                differences.Add(new CompatDifference(DiagnosticIds.MatchingAssemblyDoesNotExist, string.Format(Resources.AssemblyNameDoesNotExist, rightName, left.Identity.Name), DifferenceType.Added, left.Identity.GetDisplayName()));
+                return;
+            }
+
+            AssemblyIdentity leftIdentity = left.Identity;
+            AssemblyIdentity rightIdentity = right.Identity;
+
+            string leftAssemblyName = leftIdentity.Name;
+            string leftAssemblyCulture = string.IsNullOrEmpty(leftIdentity.CultureName) ? "neutral" : left.Identity.CultureName;
```
safern, line 38:
```suggestion
            string leftAssemblyCulture = string.IsNullOrEmpty(leftIdentity.CultureName) ? "neutral" : leftIdentity.CultureName;
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,106 @@
...
+            Version leftAssemblyVersion = leftIdentity.Version;
+            ReadOnlySpan<byte> leftAssemblyPublicKeyToken = leftIdentity.PublicKeyToken.AsSpan();
+
+            string rightAssemblyName = rightIdentity.Name;
+            string rightAssemblyCulture = string.IsNullOrEmpty(rightIdentity.CultureName) ? "neutral" : right.Identity.CultureName;
```
safern, line 43:
```suggestion
            string rightAssemblyCulture = string.IsNullOrEmpty(rightIdentity.CultureName) ? "neutral" : rightIdentity.CultureName;
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            StringBuilder sb = new StringBuilder();
+            foreach (var b in publicKeyToken)
+            {
+                sb.Append(b.ToString("x2"));
+            }
+            return sb.ToString();
+        }
+
+        private CompatDifference CreateDifference(string format, string leftProperty, string rightProperty, string leftName, string rightName, AssemblyIdentity identity) =>
```
safern, line 103: Maybe rename this to, `CreateIdentityDifference`? as this is not used by the name difference?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
```diff
@@ -0,0 +1,247 @@
...
+        [Fact]
+        public static void AssemblyNamesDoNotMatch()
+        {
+            IAssemblySymbol left = CSharpCompilation.Create("AssemblyA").Assembly;
+            IAssemblySymbol right = CSharpCompilation.Create("AssemblyB").Assembly;
+            ApiComparer differ = new();
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(left, right);
+
+            Assert.Single(differences);
+            Assert.Equal("CP0003 : left assembly name 'AssemblyA' does not match with right assembly name 'AssemblyB'.", differences.First().ToString());
```
safern, line 32: For these tests we are not asserting the messages because they are localized. Instead do a contains: `AssemblyA` and COntaines `AssemblyB'. 

You can look at the other tests, but there is a CompatDifferenceComparer that doesn't compare the message. In order to assert the other properties of the difference. 

Also, shouldn't this diagnostic ID be `CP0004`?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -117,9 +117,24 @@
...
+    <value>{2} assembly token '{0}' does not match with {3} '{1}'.</value>
+  </data>
   <data name="AssemblySearchDirectoryDoesNotExist" xml:space="preserve">
     <value>Provided assembly search directory '{0}' does not exist.</value>
   </data>
+  <data name="AssembyCultureDoesNotMatch" xml:space="preserve">
+    <value>{2} assembly culture '{0}' does not match with {3} assembly culture '{1}'.</value>
+  </data>
+  <data name="AssembyVersionDoesNotMatch" xml:space="preserve">
+    <value>{3} assembly version '{1}' should be not higher than {2} version '{0}'.</value>
```
safern, line 136:
```suggestion
    <value>{3} assembly version '{1}' should be equal to {2} version '{0}'.</value>
```

In this case the versions should be equal, right?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            if (leftAssemblyCulture != rightAssemblyCulture)
+            {
+                differences.Add(CreateDifference(Resources.AssembyCultureDoesNotMatch, leftAssemblyCulture, rightAssemblyCulture, leftName, rightName, rightIdentity));
+            }
+ 
+            if (rightAssemblyVersion < leftAssemblyVersion)
+            {
+                differences.Add(CreateDifference(Resources.AssembyVersionDoesNotMatch, rightAssemblyVersion.ToString(), leftAssemblyVersion.ToString(), rightName, leftName, rightIdentity));
```
safern, line 59: Maybe add another resource string so that the error message eludes to the version needing to be greater or equal?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
```diff
@@ -0,0 +1,247 @@
...
+
+            ApiComparer strictDiffer = new();
+            strictDiffer.StrictMode = true;
+            leftSymbols = new AssemblySymbolLoader().LoadAssembly(leftDllPath);
+            rightSymbols = new AssemblySymbolLoader().LoadAssembly(rightDllPath);
+
+            // Not strictly compatible
+            differences = strictDiffer.GetDifferences(leftSymbols, rightSymbols);
+            Assert.Single(differences);
+            Assert.Equal("CP0003 : right assembly version '2.0.0.0' should be not higher than left version '1.0.0.0'.", differences.First().ToString());
```
safern, line 136: instead of doing `differences.First().ToString()` since the other properties in CompatDifference matter, let's assert for those as well?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
Anipik: CP0004 is thrown when mapper doesnot find the exact assembly and set one of the inputs as null

safern: A right.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -117,9 +117,27 @@
...
+    <value>{2} assembly token '{0}' does not match with {3} '{1}'.</value>
+  </data>
   <data name="AssemblySearchDirectoryDoesNotExist" xml:space="preserve">
     <value>Provided assembly search directory '{0}' does not exist.</value>
   </data>
+  <data name="AssembyCultureDoesNotMatch" xml:space="preserve">
+    <value>{2} assembly culture '{0}' does not match with {3} assembly culture '{1}'.</value>
+  </data>
+  <data name="AssembyVersionDoesNotMatch" xml:space="preserve">
+    <value>{3} assembly version '{1}' should be equal to  {2} version '{0}'.</value>
```
safern, line 136:
```suggestion
    <value>{3} assembly version '{1}' should be equal to {2} version '{0}'.</value>
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -117,9 +117,27 @@
...
+  </data>
+  <data name="AssembyVersionIsNotCompatible" xml:space="preserve">
+    <value>{3} assembly version '{1}' should be not higher than {2} version '{0}'.</value>
```
safern, line 139: Assembly version can be equal or higher when in strict mode, this error string still seems wrong.

safern (approved):
LGTM

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
Anipik: I dont think its worth it. I am expecting the mapper scenario to be more common and  we will hit that and version, culture and token are more interesting scenrios here.

PR #18714, opened 2021-07-03
Enable running of strict mode api compat in packageValidation

### LINKED ISSUE #18702: [Package Validation] Should enable API Compat strict mode when applicable
Package Validation should enable strict mode when running API Compat between a compile asset and its applicable runtime asset.

cc: @Anipik @ericstj

Anipik:
This feature has been requested by the aspnetcore as they want to run all the validators in strict mode (except the baseline)

This pr provide an option for running of all the apicompat runs (except with previous version) in strict mode.

Fixes: #18702


src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -24,6 +24,7 @@
       RuntimeGraph="$(RuntimeIdentifierGraphPath)"
       NoWarn="$(NoWarn)"
       RunApiCompat="$([MSBuild]::ValueOrDefault('$(RunApiCompat)', 'true'))"
+      EnableStrictMode="$([MSBuild]::ValueOrDefault('$(RunApiCompat)', 'false'))"
```
safern, line 27: Is the RunApiCompat on by default? I think strict mode should have its own flag and should be on by default when running between compile and lib assets.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -24,6 +24,8 @@
       RuntimeGraph="$(RuntimeIdentifierGraphPath)"
       NoWarn="$(NoWarn)"
       RunApiCompat="$([MSBuild]::ValueOrDefault('$(RunApiCompat)', 'true'))"
+      EnableStrictModeForCompatibleTfms="$([MSBuild]::ValueOrDefault('$(EnableStrictModeForCompatibleTfms)', 'true'))"
```
safern, line 27: Can you elaborate what's the difference between these two and why is on by default and the other is not?

Anipik: eg net6.0 assembly has some features which are not present in ns2.0, enabling the strict mode will mean that  everything in net6.0 needs to be in ns2.0 by default

safern: Right, but what behavior does compatible tfms and what does compatible frameworks cover? I feel like naming is confusing, they are very similar.

Anipik: CompatibleTargetFrameworks => covers cases that compileTime and runtimeTime (runtimeSpecific or runtime independent) are compatible with each other for any framework in the package.
CompatibleFrameworksInPackage => covers cases where a package has compatible frameworks inside the packge eg net5.0 & net6.0, ns2.0 and net5.0 and make sure that those assemblies satisfy the same compatibility rules.

safern (approved):
Should we add some tests that exercise these properties being set?

Anipik:
> Should we add some tests that exercise these properties being set?

I am working on a change on what checks have been run. we can add these tests to that change.

PR #18740, opened 2021-07-06
Don't check for elevation on non-elevated workload options

### LINKED ISSUE #18738: `dotnet workload install  xxxx --print-download-link-only` should not require elevated permissions
```
➜ jasonimison@Jasons-iMac  ~/src/vsmac git:(dev/nosami/updater-dotnet-workload) dotnet --version
6.0.100-preview.7.21356.1
➜ jasonimison@Jasons-iMac  ~/src/vsmac git:(dev/nosami/updater-dotnet-workload) dotnet workload install maui-ios --print-download-link-only 
Inadequate permissions to access, please elevate to run this command.
install

```

This command doesn't actually install anything, it just prints the location of any nupkgs that are needed, so it shouldn't require elevated permissions. The command works without elevation in 6.0 preview 5

This is a problem for the VSMac updater as it downloads the packages in the background without user intervention and only the actual installation prompts for elevation. This is a blocker for VSMac integration.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/18738

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
@@ -22,7 +22,8 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             string dotnetDir = null, 
             string tempDirPath = null,
             PackageSourceLocation packageSourceLocation = null,
-            RestoreActionConfig restoreActionConfig = null)
+            RestoreActionConfig restoreActionConfig = null, 
+            bool elevationRequired = true)
         {
             var installType = GetWorkloadInstallType(sdkFeatureBand, string.IsNullOrWhiteSpace(dotnetDir) 
                 ? Path.GetDirectoryName(Environment.ProcessPath)
@@ -38,7 +39,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                 return new NetSdkMsiInstaller();
             }
 
-            if (!CanWriteToDotnetRoot(dotnetDir))
+            if (elevationRequired && !CanWriteToDotnetRoot(dotnetDir))
             {
                 throw new GracefulException(LocalizableStrings.InadequatePermissions);
             }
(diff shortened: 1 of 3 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -38,7 +39,7 @@ internal class WorkloadInstallerFactory
                 return new NetSdkMsiInstaller();
             }
 
-            if (!CanWriteToDotnetRoot(dotnetDir))
+            if (elevationRequired && !CanWriteToDotnetRoot(dotnetDir))
```
wli3, line 42: I think there are too many flags for the logic. Any reason we cannot catch the exception and point the error only when we failed to write?

sfoslund: The `CanWriteToDotnetRoot` method does catch the error and it returns false only when we failed to write. The issue here is that there are some options (namely the print download like option) that don't need to write to the dotnet root, so they don't need to be elevated and this is erroring unnecessarily.

sfoslund: @wli3 do you have an alternative suggestion here?

wli3: Just don't do any explicit check, rely on catch and rethrow better error message?

sfoslund: I originally added this explicit check because we wanted to error right away if we had inadequate permissions, if we wait until we try to access the dotnet root in the actual installation then we've already done a lot of work (updating ad manifests, downloading/ extracting packages). Also, if we want to catch and rethrow with a better error message whenever we run into a permission issue, we'll have to duplicate that across quite a few places, since there are lots of places we could run into that issue.

wli3 (approved):

wli3:
file an issue to see if we split read API and write API later.
timebox to see if we can add test around it

sfoslund:
> timebox to see if we can add test around it

Didn't come up with anything for this, will follow up with this issue: https://github.com/dotnet/sdk/issues/18809

PR #18750, opened 2021-07-06
Change manifest root hook to work across SDK versions

dsplaisted:
This changes the workload resolver so that it will append the SDK feature band to the paths defined in `DOTNETSDK_WORKLOAD_MANIFEST_ROOTS`.  This means that setting this environment variable for one SDK feature band won't pollute other bands with workload manifests for the wrong band.

This is a change in behavior for this environment variable.  The only usage of it that I could find so far was [in xamarin-macios](https://github.com/xamarin/xamarin-macios/blob/3b6f059863e2435422769cbe2fa11c41cd095ff8/dotnet/Makefile#L347), which would need to be updated if we make this change.

@rolfbjarne @jonathanpeppers @Redth Does this sound OK?  Are you aware of any other uses of this environment variable?


jonathanpeppers (commented):
xamarin-android and dotnet/maui are not using `$DOTNETSDK_WORKLOAD_MANIFEST_ROOTS`.

They both take the approach of setting up a local dotnet install and copying files inside.

joeloff (approved):

PR #18754, opened 2021-07-06
Dispose of all file streams in workload install

### LINKED ISSUE #18703: Investigate use of GC calls in workload install
We have had many file locked errors when developing workload install that have been fixed by calling:
```
                GC.Collect();
                GC.WaitForPendingFinalizers();
```

Example: https://github.com/dotnet/sdk/pull/18701

These calls should not be necessary, it's possible the locked files are a result of a runtime bug. We should investigate and potentially look in the runtime team. 

cc @joeloff

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/18703

### DIFF
--- a/src/Cli/dotnet/NugetPackageDownloader/FirstPartyNuGetPackageSigningVerifier.cs
+++ b/src/Cli/dotnet/NugetPackageDownloader/FirstPartyNuGetPackageSigningVerifier.cs
@@ -47,14 +47,18 @@ namespace Microsoft.DotNet.Cli.NuGetPackageDownloader
 
         internal bool IsFirstParty(FilePath nupkgToVerify)
         {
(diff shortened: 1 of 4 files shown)


joeloff:
Do we know why the using statements are not triggering .close() / dispose() calls when exiting? streams should call dispose which should call close and the release the underlying file handles.

joeloff (approved):

sfoslund:
> Do we know why the using statements are not triggering .close() / dispose() calls when exiting?

I think the issue here was that we weren't using `using` statements at all. I wasn't aware that File.Create opened a stream, so that caused some of the locks, and there were some streams that were being opened in the nuget package verification that weren't being closed or wrapped in usings.

PR #18755, opened 2021-07-06
.NET Workloads: Admin installs

joeloff:
Basic infrastructure to support MSI based workload installations using the CLI. Right now only installs, uninstalls and repairs are supported. Most of the code are just to support some plumbing needed to deal with elevation, logging, etc. that's common to most of the remaining operations.

Once manifest updates are implemented, the detections/planning logic will require a partial redesign since we have to deal with supersedence of packages and do additional checks to avoid downgrade errors etc.

The current implementation supports two execution models

1. The user is using MSIs and executes the workload command from an elevated prompt.
2. The user is using MSIs and executes the workload command from a non-elevated prompt.

In the latter case, a hidden command ```dotnet workload elevate``` is executed. This starts up an elevated server and establishes a number of secure pipes that allows for IPC between the server. The client will only elevate when performing privileged operations and should only do so once per executed command, e.g. if a workload requires installing 7 MSIs, the install command will elevate when it hits the first point where additional permissions are required. Once elevated, the remaining MSIs will not prompt for elevation again. However separate commands will/may require elevation.


src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
```diff
@@ -356,6 +363,11 @@ public IEnumerable<(string, string)> GetInstalledPacks(SdkFeatureBand sdkFeature
                 .Select(packVersionPath => (Path.GetFileName(Path.GetDirectoryName(packVersionPath)), Path.GetFileName(packVersionPath)));
         }
 
+        public void Shutdown()
+        {
+
```
sfoslund, line 368: Can we include a comment here that there's no shutdown needed for this install type?

src/Cli/dotnet/Installer/Windows/InstallerBase.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// 
+        /// The<see cref="Restart"/> property will be set to <see langword="true" /> if the error is either <see cref="Error.SUCCESS_REBOOT_INITIATED"/>
+        /// or <see cref="Error.SUCCESS_REBOOT_REQUIRED"/>.
+        /// </summary>
+        /// <param name="error">The error code to check.</param>
+        /// <param name="message">The message to include the exception.</param>
+        /// <exception cref="WorkloadException" />
+        protected void ExitOnError(uint error, string message)
+        {
+            if (!Error.Success(error))
```
sfoslund, line 128: nit: these method/ var names make this line very confusing to read

src/Cli/dotnet/Installer/Windows/InstallerBase.cs
```diff
@@ -0,0 +1,212 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Throws an exception if the HRESULT of the response message indicates
+        /// a failure.
+        /// </summary>
+        /// <param name="response">The response message to examine.</param>
+        /// <exception cref="WorkloadException"/>
+        protected void ExitOnFailure(InstallResponseMessage response, string message)
```
sfoslund, line 168: In this context what's the difference between an error and a failure?

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -0,0 +1,240 @@
...
+            if (!Directory.Exists(path))
+            {
+                CreateSecureDirectory(Directory.GetParent(path).FullName);
+
+                DirectorySecurity directorySecurity = new();
+                directorySecurity.SetAccessRule(s_AdministratorRule);
+                directorySecurity.SetGroup(s_LocalSystemSid);
+                directorySecurity.CreateDirectory(path);
+                SecureDirectory(path);
+            }
```
sfoslund, line 111: If the dir already exists we assume that it's secure without checking any access rules? Is that okay?

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// <summary>
+        /// Moves a file from one location to another if the destination file does not already exist.
+        /// </summary>
+        /// <param name="sourceFile">The source file to move.</param>
+        /// <param name="destinationFile">The destination where the source file will be moved.</param>
+        protected void MoveFile(string sourceFile, string destinationFile)
+        {
+            if (!File.Exists(destinationFile))
+            {
+                FileAccessRetrier.RetryOnMoveAccessFailure(() => File.Move(sourceFile, destinationFile));
```
sfoslund, line 182: I believe File.Move will error when moving across mounts, is there a situation where that will be a problem here?

src/Cli/dotnet/Installer/Windows/PipeStreamMessageDispatcherBase.cs
```diff
@@ -0,0 +1,118 @@
...
+                    {
+                        throw new TimeoutException("Timed out waiting for connection.");
+                    }
+                }
+                else if (_pipeStream is NamedPipeClientStream)
+                {
+                    ((NamedPipeClientStream)_pipeStream).Connect(ConnectionTimeout);
+                    // We cannot set the transmission mode of a pipe client until it's connected.
+                    _pipeStream.ReadMode = PipeTransmissionMode.Message;
+                }
```
sfoslund, line 80: Else throw? Or is there a case where the pipe stream is neither the client nor the server?

src/Cli/dotnet/Installer/Windows/TimestampedFileLogger.cs
```diff
@@ -0,0 +1,208 @@
...
+        public TimestampedFileLogger(string path, int flushThreshold, params string[] logPipeNames)
+        {
+            Directory.CreateDirectory(Path.GetDirectoryName(path));
+            _stream = File.CreateText(path);
+            LogPath = Path.GetFullPath(path);
+            FlushThreshold = flushThreshold;
+
+            // Capture control events. While console applications do support the CancelKeyPres event, users can
+            // terminate a CLI command by simply closing the command window, rebooting or logging off.
+            //NativeMethods.Windows.SetConsoleCtrlHandler(CtrlHandler, true);
```
sfoslund, line 79: nit: remove commented out code

src/Cli/dotnet/Installer/Windows/WorkloadPackRecord.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Installer.Windows
+{
+    /// <summary>
+    /// Represents a single workload pack installation record in the registry created
+    /// by a workload pack MSI.
+    /// </summary>
+    internal class WorkloadPackRecord
```
sfoslund, line 14: Can we specify in the name that this is for the msi/ registry install, something line WorkloadPackRegistryRecord?

src/Cli/dotnet/commands/dotnet-workload/elevate/WorkloadElevateCommand.cs
```diff
@@ -0,0 +1,46 @@
...
+                {
+                    _server?.Shutdown();
+                }
+            }
+            else
+            {
+                throw new GracefulException(LocalizableStrings.RequiresWindows);
+            }
+
+            return 0;
```
sfoslund, line 43: Exit code always zero for this command?

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// <summary>
+        /// Moves a file from one location to another if the destination file does not already exist.
+        /// </summary>
+        /// <param name="sourceFile">The source file to move.</param>
+        /// <param name="destinationFile">The destination where the source file will be moved.</param>
+        protected void MoveFile(string sourceFile, string destinationFile)
+        {
+            if (!File.Exists(destinationFile))
+            {
+                FileAccessRetrier.RetryOnMoveAccessFailure(() => File.Move(sourceFile, destinationFile));
```
sfoslund, line 263: Same as above, File.Move may cause problems across mounts.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -0,0 +1,598 @@
...
+        public void GarbageCollectInstalledWorkloadPacks()
+        {
+            try
+            {
+                Log?.LogMessage("Starting garbage collection.");
+                IEnumerable<SdkFeatureBand> installedFeatureBands = GetInstalledFeatureBands();
+                IEnumerable<WorkloadId> installedWorkloads = RecordRepository.GetInstalledWorkloads(_sdkFeatureBand);
+                IEnumerable<PackInfo> expectedWorkloadPacks = installedWorkloads
+                    .SelectMany(workload => _workloadResolver.GetPacksInWorkload(workload.ToString()))
+                    .Select(pack => _workloadResolver.TryGetPackInfo(pack))
```
sfoslund, line 102: This will result in dups, so we should probably also call .Distinct().

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -0,0 +1,598 @@
...
+
+        private IWorkloadResolver _workloadResolver;
+
+        private readonly PackageSourceLocation _packageSourceLocation;
+
+        private readonly string _dependent;
+
+        private bool _disposed;
+
+        public int ExitCode => Restart ? unchecked((int)Error.SUCCESS_REBOOT_REQUIRED) : unchecked((int)Error.SUCCESS);
```
sfoslund, line 40: It seems like this is never set again? Do we not set the exit code on failures?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -371,7 +372,15 @@ private async Task<bool> NewerManifestPackageExists(ManifestId manifest)
         private string GetAdvertisingManifestPath(SdkFeatureBand featureBand, ManifestId manifestId) =>
             Path.Combine(_userHome, ".dotnet", "sdk-advertising", featureBand.ToString(), manifestId.ToString());
 
+        //internal static PackageId GetManifestPackageId(SdkFeatureBand featureBand, ManifestId manifestId) =>
+        //    new PackageId($"{manifestId}.Manifest-{featureBand}");
+
         internal static PackageId GetManifestPackageId(SdkFeatureBand featureBand, ManifestId manifestId) =>
-            new PackageId($"{manifestId}.Manifest-{featureBand}");
+            WorkloadInstallerFactory.GetWorkloadInstallType(featureBand, Path.GetDirectoryName(Environment.ProcessPath)) switch
```
sfoslund, line 379: IMO it might make more sense to factor these out into separate methods (GetManifestPackageId and GetManifestMsiPackageId)

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -0,0 +1,598 @@
...
+            {
+                LogException(e);
+                throw;
+            }
+        }
+
+        public void RollBackWorkloadPackInstall(PackInfo packInfo, SdkFeatureBand sdkFeatureBand)
+        {
+            Log?.LogMessage($"Rolling back workload pack.");
+            LogPackInfo(packInfo);
```
sfoslund, line 311: So nothing is actually done here besides logging?

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -0,0 +1,598 @@
...
+            return installedFeatureBands;
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
```
sfoslund, line 419: What is the disposing param for? It looks like it's never used and the name is unclear.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
```diff
@@ -0,0 +1,115 @@
...
+                string workloadInstallationKeyName = Path.Combine(BasePath, $"{sdkFeatureBand}", $"{workloadId}");
+                Log?.LogMessage($"Deleting {workloadInstallationKeyName}");
+                _baseKey.DeleteSubKeyTree(workloadInstallationKeyName, throwOnMissingSubKey: false);
+            }
+            else if (IsClient)
+            {
+                Dispatcher.SendWorkloadRecordRequest(InstallRequestType.DeleteWorkloadInstallationRecord,
+                    workloadId, sdkFeatureBand);
+
+                //LogAndExitOnFailure(response, $"Failed to delete workload record key.");
```
sfoslund, line 70: nit: remove commented code

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -119,44 +119,52 @@ private void ValidateWorkloadIdsInput()
...
+                    }
+                    catch (Exception e)
+                    {
+                        // Don't show entire stack trace
+                        throw new GracefulException(string.Format(LocalizableStrings.WorkloadInstallationFailed, e.Message), e);
+                    }
                 }
             }
-            else
+            catch (Exception)
```
sfoslund, line 161: Should this be a catch or a finally? Do we not want to shutdown on success?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -119,44 +119,52 @@ private void ValidateWorkloadIdsInput()
...
-                        _skipManifestUpdate,
-                        _includePreviews, 
-                        string.IsNullOrWhiteSpace(_fromCacheOption) ? null : new DirectoryPath(_fromCacheOption));
-                }
-                catch (Exception e)
-                {
-                    // Don't show entire stack trace
-                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadInstallationFailed, e.Message), e);
-                }
+                _workloadInstaller.Shutdown();
```
sfoslund, line 163: How reliable is the shutdown method? If it throws here it will cover up the original installation failure.

sfoslund (commented):
LGTM overall! I don't see any blockers but lots of questions and I would like for someone else to review as well.

src/Cli/dotnet/Installer/Windows/InstallClientElevationContext.cs
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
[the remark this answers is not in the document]
joeloff: Yup, I chatted with @wli3 about this and he felt this should be at the same level like the telemetry APIs. Originally I had it under dotnet-workloads.

src/Cli/dotnet/Installer/Windows/ElevationContextBase.cs
```diff
@@ -0,0 +1,38 @@
...
+        public bool HasElevated
+        {
+            get;
+            protected set;
+        }
+
+        /// <summary>
+        /// <see langword="true"/> if the the current user has elevated permissions.
+        /// </summary>
+        public abstract bool IsElevated
```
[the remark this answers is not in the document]
joeloff: HasElevated is used in the ElevationContext to avoid elevating more than once. That way, all the methods that end up on paths requiring privileged operations can just always call Elevate() and leave the work to the ElevationContext.

If we're not elevated, we may need to elevated, but if we are elevated, we won't need to elevate again. The other difference is that IsElevated always queries the security principal, whereas HasElevated is a flag we specifically set in the code. We could have tried to elevate, and the user could have rejected elevation

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// </summary>
+        public InstallRequestType RequestType
+        {
+            get;
+            set;
+        }
+
+        public string SdkFeatureBand
+        {
+            get;
```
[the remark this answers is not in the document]
joeloff: Not all message include the feature band, I had lots of issues with null references because the backing field of the struct couldn't be initialized during deserialization. And because structs are value types, they're not null, but the private field that held the ReleaseVersion was null.

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
```diff
@@ -0,0 +1,101 @@
...
+namespace Microsoft.DotNet.Installer.Windows
+{
+    /// <summary>
+    /// Defines the IPC message structure used to send commands from the unelevated client to the elevated server instance of
+    /// the installer
+    /// </summary>
+    internal class InstallRequestMessage : InstallMessageBase
+    {
+        /// <summary>
+        /// The dependent value to modify.
```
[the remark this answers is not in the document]
joeloff: That's used for reference counting. Typically, it would be ```Microsoft.NET.Sdk,<feature-band>,<SDK host architecture>```.

image

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
joeloff: yup, rollbacks aren't implemented yet. I'm still mulling over how I want to track the actual actions, e.g. if we installed a workload, but we have a shared pack that was installed by another workload, a rollback probably shouldn't remove the pack, but the reference counting is done against the same feature band. Similarly we have issues with modifying registry keys where we may no-op, but then need to remember whether we actually wrote the key during the current transaction, etc.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
joeloff: I cheated a bit here. I didn't want to add shutdown calls in the separate try blocks, but we can

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
joeloff: That is a possibility. I'm open to suggestions. For example, if the user starts a command and hits ctrl+break or starts to logoff. We can handle those, but they're very platform specific.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
joeloff: Actually that's a bug, good catch. I changed some of the error reporting a while back and I think I missed this class. I'll update the PR.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
joeloff: the overload with the bool is required for non-sealed classes to handle managed/unmanaged resources. I added this long ago when I was trying to figure out when a command was completed. I figured once the command itself was being disposed it might be a good indication to shutdown the server.

I've removed this

src/Cli/dotnet/Installer/Windows/InstallClientElevationContext.cs
```diff
@@ -0,0 +1,72 @@
...
+                    // ordered in a single file. 
+                    _log.AddNamedPipe(WindowsUtils.CreatePipeName(serverProcess.Id, "log"));
+
+                    HasElevated = true;
+
+                    _log.LogMessage("Elevated command instance started.");
+                }
+                else
+                {
+                    throw new Exception("Failed to start the server.");
```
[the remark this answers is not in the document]
joeloff: Possibly. I deliberately held off on the strings as they changed so rapidly and figured I'd do a separate pass to finalize all the strings.

src/Cli/dotnet/Installer/Windows/ElevationContextBase.cs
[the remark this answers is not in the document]
sfoslund: Makes sense, can we include some of that context in the comments?

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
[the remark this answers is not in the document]
sfoslund: Ah, dependent in the registry? Can we include that in the comment?

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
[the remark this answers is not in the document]
sfoslund: Gotcha, no big deal, no need then

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
sfoslund: Personally I'm less concerned about those specific scenarios, since the user is getting themselves into a bad state if they take those actions and it can be fixed with repair. I just want to make sure that we don't have issues where our bug in shutdown covers up real install errors

src/Cli/dotnet/Installer/Windows/ElevationContextBase.cs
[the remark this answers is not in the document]
joeloff: I've pushed a fix with updated comments

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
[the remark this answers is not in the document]
joeloff: I've pushed an update for the comments

wli3:
We should schedule a recorded brownbag meeting for these code. How they work in general and what manual test we should perform when we want to make change in these code

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -0,0 +1,240 @@
...
+using Microsoft.DotNet.Cli.Utils;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.Installer.Windows
+{
+    /// <summary>
+    /// Manages caching workload pack MSI packages.
+    /// </summary>
+    [SupportedOSPlatform("windows")]
+    internal class MsiPackageCache : InstallerBase
```
wli3, line 21: this class looks like interop class too. Could you comment the counter part document?

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -0,0 +1,574 @@
...
+using Microsoft.Extensions.EnvironmentAbstractions;
+using Microsoft.NET.Sdk.WorkloadManifestReader;
+using Microsoft.Win32.Msi;
+using NuGet.Versioning;
+using static Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadResolver;
+
+namespace Microsoft.DotNet.Workloads.Workload.Install
+{
+    [SupportedOSPlatform("windows")]
+    internal class NetSdkMsiInstallerClient : MsiInstallerBase, IWorkloadPackInstaller
```
wli3, line 26: this class should be testable by mocking out the windows dependencies

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
wli3: > shutdown calls in the separate try blocksshutdown calls in the separate try blocks

let's do that.

A catch all over these doesn't looks right at all. At least there need to be some test

wli3:
Had one round of review. Still don't have a good idea.

All the interop code can be considered instant "legacy" code, considering they are not easy to test. But I see many places we can isolate "driver code" from business logic and put business logic into test.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
wli3: and we should have the brownbag on how we test this class with server class

src/Cli/dotnet/Installer/Windows/DetectState.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Installer.Windows
+{
+    internal enum DetectState
```
wli3, line 6: We should have a documentation link for all classes like this

src/Cli/Microsoft.DotNet.Cli.Utils/Uuid.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /// <summary>
+        /// Generates a version 3 UUID given a namespace UUID and name. This is based on the algorithm described in
+        /// RFC 4122 (https://tools.ietf.org/html/rfc4122), section 4.3.
+        /// </summary>
+        /// <param name="namespaceUuid">The UUID representing the namespace.</param>
+        /// <param name="name">The name for which to generate a UUID within the given namespace.</param>
+        /// <returns>A UUID generated using the given namespace UUID and name.</returns>
+        public static Guid CreateUuid(Guid namespaceUuid, string name)
```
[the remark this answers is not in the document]
joeloff: This is code I wrote over 10 years ago for ASP.NET., but we can actually remove it. We use the v3 format to generate GUIDs in the Arcade tooling as it matches what WiX uses. I did use this to derive product codes in the CLI when I only have a workload pack ID to avoid some caching operations to obtain the msi JSON manifests.

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
joeloff: Don't think there's any interop or p/invoke calls here. Could you clarify what you mean by the counter part?

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
joeloff: This class relies heavily on Windows specific operations, including IPC, the package cache and subsequent ACLs, registry and MSI APIs in the OS. I'll sync with you offline

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkManagedInstaller.cs
joeloff: Yes

src/Cli/dotnet/Installer/Windows/DetectState.cs
joeloff: I added more documentation

joeloff:
The merge obviously didn't work.

wli3 (approved):
For the schedule sake, let's merge it for now and get some manual test coverage

PR #18794, opened 2021-07-07
Adding PrepareForBundle target

VSadov:
As described in https://github.com/dotnet/designs/pull/228

The testing strategy:
- `User_can_get_bundle_info_before_bundling`
A synthetic target is hooked up to run where signing tool should run. 
In the target we observe that `AppHostFile` is set as expected and that among the `FilesToBundle` items we have `System.Private.CoreLib.dll`
- `User_can_move_file_before_bundling`
A synthetic target is hooked up to run where signing tool should run. 
The target renames `SingleFileTest.dll` file and removes its entry from `FilesToBundle`
If the changes to `FilesToBundle` are not propagated to Bundler, the publish step will fail. 
(NOTE: we do not expect a runnable app here, just expect that publishing succeeds, proving that `FilesToBundle` is functional)

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
@@ -863,7 +863,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     </PropertyGroup>
   </Target>
 
-  <Target Name="BeforeBundle"
+  <Target Name="PrepareForBundle"
       DependsOnTargets="_ComputeFilesToBundle"
       Condition="'$(PublishSingleFile)' == 'true'">
 
@@ -879,8 +879,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="GenerateBundle" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <Target Name="GenerateSingleFileBundle"
           Condition="'$(PublishSingleFile)' == 'true'"
-          DependsOnTargets="_ComputeFilesToBundle;BeforeBundle"
-          Inputs="@(_FilesToBundle)"
+          DependsOnTargets="_ComputeFilesToBundle;PrepareForBundle"
+          Inputs="@(FilesToBundle)"
           Outputs="$(PublishedSingleFilePath)">
 
     <PropertyGroup>
@@ -894,7 +894,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkError Condition="'$(IncludeAllContentForSelfExtract)' == 'true' And '$(IncludeNativeLibrariesForSelfExtract)' != 'true'"
                  ResourceName="CannotIncludeAllContentButNotNativeLibrariesInSingleFile" />
 
-    <GenerateBundle FilesToBundle="@(_FilesToBundle)"
+    <GenerateBundle FilesToBundle="@(FilesToBundle)"
                     AppHostName="$(PublishedSingleFileName)"
                     IncludeSymbols="$(IncludeSymbolsInSingleFile)"
                     EnableCompressionInSingleFile="$(EnableCompressionInSingleFile)"
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
@@ -861,13 +861,25 @@ Copyright (c) .NET Foundation. All rights reserved.
       <PublishedSingleFileName>$(AssemblyName)$(_NativeExecutableExtension)</PublishedSingleFileName>
       <PublishedSingleFilePath>$(PublishDir)$(PublishedSingleFileName)</PublishedSingleFilePath>
     </PropertyGroup>
-
   </Target>
 
+  <Target Name="BeforeBundle"
+      DependsOnTargets="_ComputeFilesToBundle"
+      Condition="'$(PublishSingleFile)' == 'true'">
+
+    <ItemGroup>
+      <FilesToBundle Include="@(_FilesToBundle)"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <AppHostFile>$(PublishedSingleFileName)</AppHostFile>
+    </PropertyGroup>
+  </Target>
+  
   <UsingTask TaskName="GenerateBundle" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <Target Name="GenerateSingleFileBundle"
           Condition="'$(PublishSingleFile)' == 'true'"
-          DependsOnTargets="_ComputeFilesToBundle"
+          DependsOnTargets="_ComputeFilesToBundle;BeforeBundle"
           Inputs="@(_FilesToBundle)"
           Outputs="$(PublishedSingleFilePath)">
 
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileApp.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileApp.cs
@@ -911,8 +911,6 @@ class C
 
             void VerifyPrepareForBundle(XDocument project)
             {
-                System.Console.WriteLine(project.ToString());
-
                 var ns = project.Root.Name.Namespace;
                 var targetName = "CheckPrepareForBundleData";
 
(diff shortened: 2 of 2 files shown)


VSadov:
@vitek-karas @hoyosjs - please take a look.

Also - is there a way to test the target that would actually have some validation value?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -861,13 +861,25 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <AppHostFile>$(PublishedSingleFileName)</AppHostFile>
+    </PropertyGroup>
+  </Target>
+  
   <UsingTask TaskName="GenerateBundle" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <Target Name="GenerateSingleFileBundle"
           Condition="'$(PublishSingleFile)' == 'true'"
-          DependsOnTargets="_ComputeFilesToBundle"
+          DependsOnTargets="_ComputeFilesToBundle;BeforeBundle"
           Inputs="@(_FilesToBundle)"
```
hoyosjs, line 883: The inputs now need to correspond to the exposed extension ItemGroup, so `FilesToBundle`

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -861,13 +861,25 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <PropertyGroup>
+      <AppHostFile>$(PublishedSingleFileName)</AppHostFile>
+    </PropertyGroup>
+  </Target>
+  
   <UsingTask TaskName="GenerateBundle" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <Target Name="GenerateSingleFileBundle"
           Condition="'$(PublishSingleFile)' == 'true'"
-          DependsOnTargets="_ComputeFilesToBundle"
+          DependsOnTargets="_ComputeFilesToBundle;BeforeBundle"
```
hoyosjs, line 882: What's the intended consumption pattern here? Add a Target that is before `GenerateSingleFileBundle` and depends on `BeforeBundle`? Would something like `GenerateSingleFileBundle` depending on `$(GenerateSingleFileBundleDependsOn)` and have it just be `_ComputeFilesToBundle;BeforeBundle` + user provided targets where they can transform the itemgroup as needed? @rainersigwald, what's the most idiomatic/user safe way you'd recommend for this extensibility?

hoyosjs: Something a little like https://github.com/dotnet/msbuild/blob/HEAD/src/Tasks/Microsoft.Common.CurrentVersion.targets#L871-L883

vitek-karas: I agree that we should also define the way to "plugin". For build it's enough to just define a target with the right name `BeforeBuild` - maybe we should do the same.

vitek-karas:
The test should validate that it can be used for the things we designed it for. So something like:
* Write a custom target which gets inserted into the right place (this will be hard to validate, so we probably won't)
* In the target validate that the inputs are as expected (`FilesToBundle`, `AppHost`, ...)
* Modify one of the files (maybe build two versions and replace one with the other, or something like that)
After the build validate that the bundle contains the modified file instead of the original.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
rainersigwald: Common targets generally use this pattern:

1. Target Foo `DependsOnTargets="$(FooDependsOn)"`
2. The default `FooDependsOn` is `BeforeFoo;CoreFoo;AfterFoo`
3. Stub implementations of `BeforeFoo` and `AfterFoo` that do nothing are defined.
4. Users may override the definitions of `BeforeFoo` or `AfterFoo` with customizations.
5. Users may also define their own targets that hook to any of those targets with `BeforeTargets` or `AfterTargets`.

The "override" extension strategy was recommended and used in many projects, but is harder to deal with in SDK-style projects, because the last definition of a target in MSBuild's expanded project wins, and a stub target defined in SDK/common .targets will override a target of the same name defined in the project file. As a result, **we now recommend creating a new target and hooking to the appropriate target** over the override approach. That also lets you give the target a meaningful name (like `SignEmbeddedPayload`).

Here, I don't think `AfterBundle` makes much sense so just adding `BeforeBundle` seems reasonable. I wonder though if it might be better to rename this target `PrepareForBundle` and document something like:

```xml
<Target Name="SignBundledFile" BeforeTargets="GenerateSingleFileBundle" DependsOnTargets="PrepareForBundle">
```

?

That's basically what the linked project is doing, the "bad" thing they're doing is modifying the "private" `@(_FilesToBundle)`. Here you're making that "public".

VSadov:
@vitek-karas - do we have any existing tests for similar functionality?

vitek-karas:
I don't know. @sbomer - did we add any similar extension point trimmer - maybe we have tests for those?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
VSadov: So the suggestions are, if I get it right:
- `GenerateSingleFileBundle` should use `FilesToBundle`, not internal `_FilesToBundle`.
I had small concern about overrider adding/removing items to the list. Changing items is basically the purpose here, but I was not sure whether changing the list itself.
Anyways, if it seems preferable to use the same `FilesToBundle`, I will change to that.
- Rename `BeforeBundle` --> `PrepareForBundle`.
I do not have an opinion here. Either seem ok to me. Just want to be sure that people more invested in the name of the target are in agreement.
- and test this somehow.

Is this right?

hoyosjs: > * `GenerateSingleFileBundle` should use `FilesToBundle`, not internal `_FilesToBundle`.
>   I had small concern about overrider adding/removing items to the list. Changing items is basically the purpose here, but I was not sure whether changing the list itself.

Yeah, it's an odd thing, but necessary. The user might deposit the signed files elsewhere. So yeah, the inputs and implementation need to use the exposed ItemGroup.

> * Rename `BeforeBundle` --> `PrepareForBundle`.

I am happy with either name. Maybe we should add the example to the design, say before `GenerateSingleFileBundle` and depend on the name we chose.

vitek-karas: And then document that the way to plug in is to do what @rainersigwald suggested:

```XML
<Target Name="MySignBundledFile" BeforeTargets="GenerateSingleFileBundle" DependsOnTargets="PrepareForBundle">
```

VSadov: > Yeah, it's an odd thing, but necessary. The user might deposit the signed files elsewhere.

Right, if files are outside of the build's control and effectively or actually read-only. 
I assumed that changing the items metadata would be sufficient to repoint to the new location, but I am probably wrong about that and some scenarios will call for having access to the actual list.

VSadov: I have noticed a small advantage in calling this `PrepareForBundle` - so that noone would ask "why there is no corresponding `AfterBundle` ?".

So I will go with `PrepareForBundle` then.

sbomer:
The linker has a similar extension point `PrepareForILLink`. There are some tests for it here: https://github.com/dotnet/sdk/blob/aaa6580c9cbbed53200c86151dfc9f2041ca9aa9/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs#L139
Note that unlike the bundle target this is not intended to be used to add/remove files from the linker input - it's just there to configure the linker behavior on the predefined inputs (see https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options#trimmed-assemblies for more info).

hoyosjs (approved):
Sans the tests, LGTM.

VSadov:
hmm, the test should have failed. Perhaps it did not run?

VSadov:
I think this is complete (including tests)

VSadov:
I think I will merge this, so that signing infra could start using this. 
If there are suggestions on the testing side we can follow up separately.

PR #18812, opened 2021-07-08
Default EnableUnsafeBinaryFormatterSerialization and EnableUnsafeUTF7Encoding in trimmed applications

### LINKED ISSUE #14475: Default values for feature switches in trimmed console apps
Runtime repo has been introducing feature switches which among other things help reduce application size when the app is trimmed. They also help avoid trim analysis warnings. This issue is to determine and implement the default values for various feature switches when trimming is turned on. 

Ideally the experience would be that trimming a template console app doesn't produce any trim analysis warnings and the size of the app is as small as possible. Downside of turning off features by default is obviously potentially unexpected change of behavior of the app.

For description of the feature switches please refer to [feature-switches.md](https://github.com/dotnet/runtime/blob/master/docs/workflow/trimming/feature-switches.md)

| MSBuild Property Name | Default | Default when `PublishTrimmed=true` | Notes |
| - | - | - | - |
| StartupHookSupport | `true` | `false` | https://github.com/dotnet/sdk/pull/14432/files |
| DebuggerSupport | `true` | `true` (unchanged) | Don't want to break diagnostics |
| EnableUnsafeUTF7Encoding | `false` | `false` (unchanged) | Already turned off by default, so no-op |
| EnableUnsafeBinaryFormatterSerialization | `true` / `false` in ASP.NET | `false` | The plan is to turn this off in future releases, so doing so for trimming in 6.0 seems OK |
| EventSourceSupport | `true` | `true` (unchanged) | Don't want to break diagnostics |
| InvariantGlobalization | `false` | ??? | Needs investigation how much it actually helps for console apps |
| UseSystemResourceKeys | `false` | ??? |  |
| HttpActivityPropagationSupport | `true` | `true` (unchanged) | Don't want to break diagnostics |

TODO:
- Measure size impact of each of these on a sample console app (so that we have data to drive the decisions)
- Measure how many warnings are produced if these are left turned on

eerhardt:
Fix #14475

FYI - @jkotas

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
@@ -37,6 +37,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <StartupHookSupport Condition="'$(StartupHookSupport)' == ''">false</StartupHookSupport>
     <CustomResourceTypesSupport Condition="'$(CustomResourceTypesSupport)' == ''">false</CustomResourceTypesSupport>
     <EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization Condition="'$(EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization)' == ''">false</EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>
+    <EnableUnsafeBinaryFormatterSerialization Condition="'$(EnableUnsafeBinaryFormatterSerialization)' == ''">false</EnableUnsafeBinaryFormatterSerialization>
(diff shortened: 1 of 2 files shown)


jkotas (approved):

pgovind (approved):
Defaults look good to me.

sbomer (approved):

sbomer:
https://github.com/dotnet/sdk/issues/14475 mentions that `EnableUnsafeUTF7Encoding` is already turned off by default, but I only see it turned off in the blazor SDK. Should that one be defaulted in the .NET SDK?

eerhardt:
> #14475 mentions that EnableUnsafeUTF7Encoding is already turned off by default, but I only see it turned off in the blazor SDK. Should that one be defaulted in the .NET SDK?

It's value is defaulted to `false` in the runtime if it isn't specified in the runtimeconfig.json. So there's really no reason to default again in the .NET SDK. There's no behavior difference, and all we get is an extra, unnecessary entry in the runtimeconfig.json.

The value we get out of this change (set it during PublishTrimmed=true) is that the linker sees its value, and trims away the UTF7 code in the runtime.

PR #18814, opened 2021-07-08
Adding cleanup command

jmarolf:
Currently this just forwards to dotnet format. Will follow up with a PR that does more custom command handling

### DIFF
--- a/src/Cli/dotnet/BuiltInCommandsCatalog.cs
+++ b/src/Cli/dotnet/BuiltInCommandsCatalog.cs
@@ -49,6 +49,12 @@ namespace Microsoft.DotNet.Cli
                 // aka.ms target: https://docs.microsoft.com/dotnet/articles/core/tools/dotnet-clean
                 DocLink = "https://aka.ms/dotnet-clean"
             },
+            ["cleanup"] = new BuiltInCommandMetadata
+            {
+                Command = CleanupCommand.Run,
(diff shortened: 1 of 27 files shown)


src/Cli/dotnet/BuiltInCommandsCatalog.cs
```diff
@@ -49,6 +49,12 @@ public static class BuiltInCommandsCatalog
                 // aka.ms target: https://docs.microsoft.com/dotnet/articles/core/tools/dotnet-clean
                 DocLink = "https://aka.ms/dotnet-clean"
             },
+            ["cleanup"] = new BuiltInCommandMetadata
+            {
+                Command = CleanupCommand.Run,
+                // aka.ms target: https://docs.microsoft.com/dotnet/articles/core/tools/dotnet-cleanup
```
dsplaisted, line 55: @jmarolf Currently this aka.ms link routes to https://docs.microsoft.com/en-us/dotnet/core/tools/elevated-access?tabs=windows

jmarolf: [This](https://docs.microsoft.com/dotnet/articles/core/tools/dotnet-cleanup) now 404s (which is believe is the correct behavior for now). Adding the documentation in [this PR](https://github.com/dotnet/docs/pull/25082)

PR #18815, opened 2021-07-08
Set attribute and LangVersion properly for a TFM

pgovind:
https://github.com/dotnet/designs/pull/232 has the latest consensus on setting the assembly level RequiresPreviewFeature attribute. 
LangVersion will be overwritten to Preview **iff** `EnablePreviewFeatures` is true and a project targets the latest TFM.

cc @terrajobst

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
@@ -271,7 +271,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </ItemGroup>
 
   <PropertyGroup>
-    <IsNetCoreAppTargetingLatestTFM Condition="'$(IsNetCoreAppTargetingLatestTFM)' == '' and '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(_TargetFrameworkVersionWithoutV)' == '6.0'">true</IsNetCoreAppTargetingLatestTFM>
+    <IsNetCoreAppTargetingLatestTFM Condition="'$(IsNetCoreAppTargetingLatestTFM)' == '' and '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(_TargetFrameworkVersionWithoutV)' == '$(NETCoreAppMaximumVersion)'">true</IsNetCoreAppTargetingLatestTFM>
   </PropertyGroup>
 
 </Project>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
@@ -270,4 +270,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="GenerateDocumentationFile" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <IsNetCoreAppTargetingLatestTFM Condition="'$(IsNetCoreAppTargetingLatestTFM)' == '' and '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(_TargetFrameworkVersionWithoutV)' == '6.0'">true</IsNetCoreAppTargetingLatestTFM>
+  </PropertyGroup>
+
 </Project>
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
@@ -370,13 +370,48 @@ namespace Microsoft.NET.Build.Tests
             AssemblyInfo.Get(assemblyPath)["InternalsVisibleToAttribute"].Should().Be("Tests");
         }
 
+        private static string _cachedCurrentTFM = null;
+
+        private static string CurrentTFM
+        {
+            get
+            {
+                if (_cachedCurrentTFM == null)
+                {
+                    var logger = new StringTestLogger();
+                    var testAssetsManager = new TestAssetsManager(logger);
+                    TestDirectory testDirectory = testAssetsManager.CreateTestDirectory();
+                    string path = testDirectory.Path;
+                    var cmd = new DotnetCommand(logger)
+                        .WithWorkingDirectory(path)
+                        .Execute("new", "console");
+
+                    string lines = File.ReadAllText(Path.Combine(path, "CurrentTFM.csproj"));
+                    string startTargetFrameworkText = @"<TargetFramework>";
+                    string endTargetFrameworkText = @"</TargetFramework>";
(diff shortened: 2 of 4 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
```diff
@@ -270,4 +270,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="GenerateDocumentationFile" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <IsNetCoreAppTargetingLatestTFM Condition="'$(IsNetCoreAppTargetingLatestTFM)' == '' and '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(_TargetFrameworkVersionWithoutV)' == '6.0'">true</IsNetCoreAppTargetingLatestTFM>
```
dsplaisted, line 274: We should use `NETCoreAppMaximumVersion` here so that we don't have to change this code for each major version.

```suggestion
    <IsNetCoreAppTargetingLatestTFM Condition="'$(IsNetCoreAppTargetingLatestTFM)' == '' and '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(_TargetFrameworkVersionWithoutV)' == '$(NETCoreAppMaximumVersion)'">true</IsNetCoreAppTargetingLatestTFM>
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -419,11 +419,22 @@ public void TestPreviewFeatures(bool enablePreviewFeatures, bool generateRequire
             var values = getValuesCommand.GetValues();
             var langVersion = values.FirstOrDefault() ?? string.Empty;
 
-            if (generateRequiresPreviewFeaturesAttribute)
+            if (enablePreviewFeatures && generateRequiresPreviewFeaturesAttribute)
             {
-                Assert.True(contains);
+                if (targetFramework == "net6.0")
```
dsplaisted, line 424: This test is going to need to be updated every time we bump the major version.  Possible ways to avoid this:

- Have a way for a test to get the "current" version of .NET.  We've had a bit of prior discussion on this, it would help us to keep our test coverage more current.
- Run `dotnet new console` to and read the TargetFramework from the created project.  This should be the latest .NET version (once the templates are updated after we start work on a new version).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets
pgovind: Ah I was looking for a similar variable but didn't find anything in my search. Fixed now!

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
pgovind: > Run dotnet new console to and read the TargetFramework from the created project.

I ended up going down this path and hacked up a static getter. I _think_ it'll work across all our environments. Open to feedback here!

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,13 +370,48 @@ public void It_includes_internals_visible_to()
...
+                    var testAssetsManager = new TestAssetsManager(logger);
+                    TestDirectory testDirectory = testAssetsManager.CreateTestDirectory();
+                    string path = testDirectory.Path;
+                    var cmd = new DotnetCommand(logger)
+                        .WithWorkingDirectory(path)
+                        .Execute("new", "console");
+
+                    string lines = File.ReadAllText(Path.Combine(path, "CurrentTFM.csproj"));
+                    string startTargetFrameworkText = @"<TargetFramework>";
+                    string endTargetFrameworkText = @"</TargetFramework>";
```
dsplaisted, line 391: It would be better to use XML APIs to parse the project XML.  Here's an example:

https://github.com/dotnet/installer/blob/44a674908e37892dc1833f8cc7f8f0d9ec48eeb3/test/EndToEnd/ProjectBuildTests.cs#L407-L410

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -370,13 +370,48 @@ public void It_includes_internals_visible_to()
             AssemblyInfo.Get(assemblyPath)["InternalsVisibleToAttribute"].Should().Be("Tests");
         }
 
+        private static string _cachedCurrentTFM = null;
+
+        private static string CurrentTFM
```
dsplaisted, line 375: Nit: `LatestTargetFramework` would be a better name here, TFM is often used this way but technically refers to a different form of the TargetFramework (such as `.NETCoreApp,Version=v5.0`).

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -421,17 +456,17 @@ public void TestPreviewFeatures(bool enablePreviewFeatures, bool generateRequire
...
-                    Assert.True(contains);
-                }
-                else
+                if (targetFramework == "net5.0")
                 {
                     // The assembly level attribute is generated only for the latest TFM for the given sdk
                     Assert.False(contains);
                     Assert.NotEqual("Preview", langVersion);
                 }
+                else
```
dsplaisted, line 465: Nit: I think this would be clearer if there were an explicit comparison between `targetFramework` and the latest target framework (`CurrentTFM` or `LatestTargetFramework` or whatever the property ends up being).

dsplaisted (approved):

dsplaisted:
Note that we've just branched for Preview 7, so if you want to get this in Preview 7 you'll need to retarget to that branch, and I think get approval.

pgovind:
Hmm maybe the backport bot doesn't exist in this repo. All this PR does is that it generates the `RequiresPreviewFeatures` assembly level attribute only for apps targeting .NET 6. Without this PR, apps targeting older versions of .NET will also have the attribute applied to them. This PR is not critical for P7, and no projects currently use this attribute, so I _think_ it'd be ok get this into RC1.  Pinging @jeffhandley and @tannergooding for thoughts? Is this critical to get into P7 or can it wait for RC1?

jeffhandley:
To make sure I understand, right now in .NET 6 Preview 7:

- If a project gets marked with `<EnablePreviewFeatures>`
- And it's targeting < `net6.0`
- The SDK will generate the `[assembly:RequiresPreviewFeatures]` attribute in `AssemblyInfo.cs`
- And that will lead to a build error since the attribute doesn't exist

With the fix, the attribute would not be generated, avoiding the build error.

If this understanding is correct, this can wait for RC1. The only preview features folks would be enabling are `net6.0`.

PR #18817, opened 2021-07-09
Bump RZL and don't use source generator if  not Dev17

captainsafia:
Part of https://github.com/dotnet/aspnetcore/issues/32672.

### DIFF
--- a/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
+++ b/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
@@ -52,8 +52,6 @@ Copyright (c) .NET Foundation. All rights reserved.
         <UseRazorSourceGenerator Condition="'$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
         <!-- The incremental source generator API is disabled on VS 16.11 and below so we disable it there. -->
         <UseRazorSourceGenerator Condition="'$(VisualStudioVersion)' &lt; '17.0'">false</UseRazorSourceGenerator>
-        <!-- Required to support incremental source generator -->
-        <LangVersion Condition="'$(UseRazorSourceGenerator)' != '' ">preview</LangVersion>
         <RazorLangVersion Condition="'$(RazorLangVersion)' == '' ">6.0</RazorLangVersion>
       </PropertyGroup>
     </When>
--- a/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
+++ b/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
@@ -50,9 +50,11 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_TargetingNET50OrLater>true</_TargetingNET50OrLater>
         <_TargetingNET60OrLater>true</_TargetingNET60OrLater>
         <UseRazorSourceGenerator Condition="'$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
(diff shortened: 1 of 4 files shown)


src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
```diff
@@ -50,9 +50,11 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_TargetingNET50OrLater>true</_TargetingNET50OrLater>
         <_TargetingNET60OrLater>true</_TargetingNET60OrLater>
         <UseRazorSourceGenerator Condition="'$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
+        <!-- The incremental source generator API is disabled on VS 16.11 and below so we disable it there. -->
+        <UseRazorSourceGenerator Condition="'$(VisualStudioVersion)' &lt; '17.0'">false</UseRazorSourceGenerator>
```
pranavkm, line 54: This feels very iffy. What happens if we don't do this and have the user figure this out? We're already claiming 6.0 requires VS 2022. (Also I don't think the comparison works - it's comparing two strings).

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
```diff
@@ -50,9 +50,11 @@ Copyright (c) .NET Foundation. All rights reserved.
...
         <!-- Required to support incremental source generator -->
         <LangVersion Condition="'$(UseRazorSourceGenerator)' != '' ">preview</LangVersion>
```
pranavkm, line 56: Can we remove this now? C# 10 is enabled by default for net6.0 projects.

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
captainsafia: I based this off the examples in https://dotnet.grep.app/search?q=%27%24%28VisualStudioVersion%29%27%20%26lt%3B&filter[lang][0]=XML. I believe it works because it does a lexographic comparison.

> What happens if we don't do this and have the user figure this out? 

Users will end up getting weird errors about types not being available when they build in VS. Since the errors aren't meant for end-users their pretty hostile (e.g. "Cannot load type IIncrementalGenerator from assembly").

I think it's localized enough that it's worth doing as a quality of life improvement.

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
captainsafia: Yep!

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
pranavkm: Could we leave this out for now? We can always add it back if we get a lot of feedback. The problem is that turning off the source generator also has a noticeable runtime behavior difference. It would be very spooky to see if this just because an older version of VS opened this.

pranavkm (approved):
Thanks!

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
captainsafia: This configuration doesn't look like its propogating over to the incremental generator so it's failing to run. Reverting for now. Will ping compiler team to get a sense of when it is safe to remove this (or if maybe there is another thing keeping the generator from running).

PR #18834, opened 2021-07-09
[API Compat] Add rule to validate members are not added to interfaces

### LINKED ISSUE #17928: [ApiCompatibility] Add a new rule to detect breaking changes on updating contract of abstract classes/ interfaces
Add a new rule to detect breaking changes when:

-  a new `abstract` method is added to  an already released `abstract` class
-  a new member is added to an already release `Interface`

safern:
Fixes: https://github.com/dotnet/sdk/issues/17928

I decided to not add a flag to specify if default implementations should be breaking as those can only appear on a runtime that supports them. If the target runtime that the code is being compiled for doesn't support that when adding a default implementation the compiler will error.

FYI: @terrajobst @utpilla

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
@@ -192,4 +192,7 @@
   <data name="VisitorRightCountShouldMatchMappersSetSize" xml:space="preserve">
     <value>The provided right count when creating the visitor should match the right set size specified for the mappers.</value>
   </data>
+  <data name="CannotAddMemberToInterface" xml:space="preserve">
+    <value>Cannot add interface member '{0}' to {1} because does not exist on {2}</value>
+  </data>
 </root>
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
@@ -130,7 +130,7 @@
     <value>Provided assembly search directory '{0}' does not exist.</value>
   </data>
   <data name="CannotAddAbstractMember" xml:space="preserve">
-    <value>Cannot add abstract member '{0}' to {1} because does not exist on {2}</value>
+    <value>Cannot add abstract member '{0}' to {1} because it does not exist on {2}</value>
   </data>
   <data name="AssembyCultureDoesNotMatch" xml:space="preserve">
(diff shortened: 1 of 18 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -192,4 +192,7 @@
   <data name="VisitorRightCountShouldMatchMappersSetSize" xml:space="preserve">
     <value>The provided right count when creating the visitor should match the right set size specified for the mappers.</value>
   </data>
+  <data name="CannotAddMemberToInterface" xml:space="preserve">
+    <value>Cannot add interface member '{0}' to {1} because does not exist on {2}</value>
```
Anipik, line 196: should it be ```Cannot add interface member '{0}' to {1} because it does not exist on {2}```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddMemberToInterface.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
Anipik, line 3: nit extra //

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddMemberToInterface.cs
```diff
@@ -0,0 +1,52 @@
...
+            // StrictMode scenario should be handled by MembersMustExist rule.
+            if (!Settings.StrictMode)
+            {
+                context.RegisterOnMemberSymbolAction(RunOnMemberSymbol);
+            }
+        }
+
+        private void RunOnMemberSymbol(ISymbol left, ISymbol right, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            if (left == null && right != null && right.ContainingType.TypeKind == TypeKind.Interface)
```
Anipik, line 26: why is ```left == null``` here ?

safern: That means the member is added to right. So it doesn't exist on left. That is what this rule is trying to catch, when a member is added to right (does not exist on left) and the type is an interface.

Anipik (approved):

ericstj (commented):
LGTM

Issue #18832, opened 2021-07-09
Architecture and OS targeting improvements for .NET 6


marcpopMSFT:
https://github.com/dotnet/designs/blob/d248ad68b41866eba17e80225a08518b8202ddf6/accepted/2021/architecture-targeting.md#net-6

For the shorthand RID, the proposal is to go with --arch and -a and --os (no need for a short form of os). 

### .NET 6

The following changes should be included in .NET 6, motivated by the x64 emulation scenario. They are all additive and non-breaking.

- [x]  Enable shorthand RID syntax. Add to the following verbs (priority order): -- https://github.com/dotnet/sdk/pull/18889
   - `build`
   - `publish`
   - `tool install`
   - `test`
   - `run`
   - `watch run`
   - `watch test`
- [x] Enable shorthand RID syntax for `tool install` (separate PR from other verbs)-- #19370
- [x]  Do NOT enable the `-r` RID syntax on these same verbs. Doing so would cause a breaking change in .NET 7, which seems like a unforgivable mistake.-- No-op
- [x] Enable implicit RID where there is one known choice:
   - The macOS Arm64 SDK should treat pre .NET 6 versions as X64 by default.-- Fixed by https://github.com/dotnet/sdk/pull/18216
   - The Windows Arm64 SDK should treat pre .NET 5 versions as X64 by default.-- https://github.com/dotnet/sdk/pull/19052
- [x]  Parity syntax between `dotnet build` and `dotnet publish` for pivoting between self-contained and framework-dependent modalities, for example:-- https://github.com/dotnet/sdk/pull/18837
   - `dotnet build -r win-x64 --self-contained`
   - `dotnet build -r win-x64 --self-contained true`
   - `dotnet build -r win-x64 --self-contained false`
   - `dotnet build -r win-x64 --no-self-contained`
- [x]  Add warning when using `-r` without a `--self-contained` or `--no-self-contained`. The warning is for .NET 6+ apps only.-- https://github.com/dotnet/sdk/pull/18837
- [x] Add a warning when using `-r` with either `--os` or `--arch`-- https://github.com/dotnet/sdk/pull/18889

The addition of the shorthand RID syntax and the parity syntax between `build` and `publish` will provide a satisfactory set of gestures to enable migration with .NET 6 to a non-breaking syntax with respect to .NET 7. In particular, users need to migrate any uses of `-r` to always be accompanied by one of the `--self-contained` or `--no-self-contained` switches.

[...]

PR #18837, opened 2021-07-09
Update self contained/ runtime command line options for build and publish

sfoslund:
Fixes 4th and 5th bullets of https://github.com/dotnet/sdk/issues/18832

## Description
This change adds `--self-contained` and `--no-self-contained` options to the `dotnet build` command to achieve parity with `dotnet publish`. It also adds a warning for .NET 6+ apps when `--runtime` is specified on the CLI without `--self-contained` or `--no-self-contained`.
These changes are all additive and do not impact the existing CLI. Note that this doesn't change functionality, as users can achieve the same result by specifying self contained as a MSBuild property on the command line. This change will make it more straight forward to achieve the same result. 

## Regression?
No

## Risk
Low, this is a new feature and won't effect existing CLI. 

## Verification
[X] Manual (required)
[x] Automated


src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -674,4 +674,17 @@ setx PATH "%PATH%;{0}"
   <data name="CommandPrereleaseOptionDescription" xml:space="preserve">
     <value>Allows prerelease packages to be installed.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SelfContainedOptionDescription" xml:space="preserve">
+    <value>Publish the .NET runtime with your application so the runtime doesn't need to be installed on the target machine.
+The default is 'true' if a runtime identifier is specified.</value>
+  </data>
+  <data name="NoSelfContainedOptionDescription" xml:space="preserve">
```
richlander, line 681: Can you call this `FrameworkDependent` instead of `NoSelfContained`? Framework dependent is the product default so it seems odd to pivot on self-contained. I know this is just a property name, but seems worth some consideration.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -674,4 +674,17 @@ setx PATH "%PATH%;{0}"
...
+    <value>Publish your application as a framework dependent application without the .NET runtime. A supported .NET runtime must be installed to run your application.</value>
```
richlander, line 682:
```suggestion
    <value>Publish your application as a framework dependent application without the .NET runtime. A compatible .NET runtime must be installed to run your application.</value>
```

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -674,4 +674,17 @@ setx PATH "%PATH%;{0}"
...
+  </data>
+  <data name="SelfContainAndNoSelfContainedConflict" xml:space="preserve">
+    <value>The '--self-contained' and '--no-self-contained' options conflict with each other. Specify only one of the options.</value>
+  </data>
+  <data name="RuntimeOptionShouldBeUsedWithSelfContained" xml:space="preserve">
```
richlander, line 687: This property has the opposite polarity it should have, I believe. I think it should be: `SelfContainedOptionShouldBeUsedWithRuntime`.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -674,4 +674,17 @@ setx PATH "%PATH%;{0}"
...
+    <value>Warning: The '-r|--runtime' option should be used with '--self-contained' or '--no-self-contained'.</value>
```
richlander, line 688:
```suggestion
    <value>Warning: The  '--self-contained' or '--no-self-contained' option should be used with '-r|--runtime'.</value>
```

richlander: Similar problem.

sfoslund:
@dsplaisted can I get a quick review here?

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -93,13 +98,53 @@ public static Argument DefaultToCurrentDirectory(this Argument arg)
 
         public static Option DebugOption() => new Option<bool>("--debug");
 
+        public static Option SelfContainedOption() =>
+            new ForwardedOption<bool>(
+                "--self-contained",
+                CommonLocalizableStrings.SelfContainedOptionDescription)
+            .ForwardAsSingle(o => $"-property:SelfContained={o}");
+
+        public static Option NoSelfContainedOption() =>
```
dsplaisted, line 107: Do we really want to support both `--no-self-contained` and `--self-contained false`?

@KathleenDollard

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -93,13 +98,53 @@ public static Argument DefaultToCurrentDirectory(this Argument arg)
...
+            }
+
+            if (!(hasSelfContainedOption || hasNoSelfContainedOption) && hasRuntimeOption)
+            {
+                projectArgs = projectArgs.Any() ? projectArgs : new string[] { Directory.GetCurrentDirectory() };
+                foreach (var project in projectArgs)
+                {
+                    try
+                    {
+                        var msbuildProj = MsbuildProject.FromFileOrDirectory(new ProjectCollection(), project, false);
```
dsplaisted, line 135: Should we be evaluating the project in the CLI here?  What kind of a perf impact will that have?

@rainersigwald

Perhaps we should try to generate this warning from the project build itself.  That would mean it would be a NETSDK warning message with an error code, and could be ignored, which would also be nice.

sfoslund: I added the error here because we wouldn't be able to detect when these properties were set on the command line vs by targets, but I suppose I could add an extra property with the CLI options to mark that it was set on the cli then move the error into targets. Do you want me to update with that approach?

rainersigwald: Evaluation is expensive enough that it seems like a good idea to push this into targets (but I don't have all the context on this particular thing fresh in my mind).

dsplaisted: Sounds like it would be a good idea to move it into the targets.  Though I think it's likely to generate the warning for each project built, as the properties we pass from the CLI are global properties which by default flow to referenced projects.

sfoslund: Updated, moved to targets. @dsplaisted do you mind re-reviewing?

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -815,4 +815,8 @@ To install these workloads, run the following command: {1}</value>
     <value>NETSDK1177: Failed to sign apphost with error code {1}: {0}</value>
     <comment>{StrBegin="NETSDK1177: "}</comment>
   </data>
+  <data name="SelfContainedOptionShouldBeUsedWithRuntime" xml:space="preserve">
+    <value>NETSDK1178: The  '--self-contained' or '--no-self-contained' option should be used with '-r|--runtime'.</value>
```
dsplaisted, line 819: @richlander @KathleenDollard To review the warning text here.  Perhaps something like the following would be clearer:

> When the target runtime is specified with '--runtime' or '-r', the deployment type should be specified with '--self-contained' or '--no-self-contained'.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1073,6 +1073,23 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   </Target>
 
+  <!--
+  ============================================================
+  Command Line Confirugation Validation
+  ============================================================
+  -->
+
+  <Target Name="ValidateCommandLineProperties"
+          AfterTargets="PreBuildEvent" >
```
dsplaisted, line 1083: For a lot of similar checks, we do `BeforeTargets="_CheckForInvalidConfigurationAndPlatform"`.  Is there any reason this is different?

src/Tests/dotnet-build.Tests/GivenDotnetBuildBuildsCsproj.cs
```diff
@@ -172,5 +172,54 @@ public void DotnetBuildDoesNotPrintCopyrightInfo()
...
+               .Pass()
+               .And
+               .HaveStdOutContaining("NETSDK1178");
+        }
+
+        [Fact]
+        public void It_does_not_warn_on_rid_with_self_contained_options()
+        {
+            var testInstance = _testAssetsManager.CopyTestAsset("HelloWorld")
+                .WithSource()
```
dsplaisted, line 197: This needs to target .NET 6, or it won't generate the warning anyway, right?

dsplaisted (commented):
The code changes look good.

Personally I don't think we should have a `--no-self-contained` option.  I think we should instead just use `--self-contained false`.  I'd like @KathleenDollard to weigh in on this, and am happy to defer to whatever she thinks is best here.

I also believe that we will now repeat this warning for each (transitively) referenced project.  Is that OK, or should we try to make the warning just show up once?

sfoslund:
> Personally I don't think we should have a --no-self-contained option

For what its worth, `--no-self-contained is already an option for publish, so this PR just adds it to build for parity.

src/Tasks/Common/Resources/Strings.resx
sfoslund: @richlander suggested this string in the previous PR reviews.

KathleenDollard: That seems a bit long, and I like to reduce wrapping where possible. Maybe: 

```suggestion
    <value>NETSDK1178: The  '--self-contained' or '--no-self-contained' option should be used when  '--runtime' is specified.</value>
```

Daniel, is there a reason you want to lead with the conditional?

KathleenDollard (commented):
And I had an old pending comment as well on this 

I'd like Rich's call on the string in question. I can live with any of the options as I think they are all clear.

src/Tasks/Common/Resources/Strings.resx
dsplaisted: @KathleenDollard It sounded a bit better to me, but it could certainly be reversed.

src/Cli/dotnet/CommonLocalizableStrings.resx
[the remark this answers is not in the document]
sfoslund: This is not a new string, it's just showing up in the diff because it was moved to a different resx file. This is the existing help text.

src/Cli/dotnet/CommonLocalizableStrings.resx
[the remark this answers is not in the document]
sfoslund: This string has been removed in subsequent updates anyways, resolving this...

src/Tasks/Common/Resources/Strings.resx
richlander: I like shorter version better.

How about this one?

> One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.

sfoslund:
@richlander @dsplaisted @KathleenDollard I've updated based on your feedback, thanks! We're hoping to get this in for preview 7 so can I please get another review?

src/Tasks/Common/Resources/Strings.resx
KathleenDollard: I hate to keep this going, but...

> Either '--self-contained' or '--no-self-contained' is required when '--runtime' is used.

Aligning "One of" and "are" grammatically seems hard ;-)

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -93,13 +95,33 @@ public static Argument DefaultToCurrentDirectory(this Argument arg)
...
+
         public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosity)
         {
             return verbosity.Equals(VerbosityOptions.diag) ||
                 verbosity.Equals(VerbosityOptions.diagnostic) ||
                 verbosity.Equals(VerbosityOptions.d) ||
                 verbosity.Equals(VerbosityOptions.detailed);
         }
+
+        public static void ValidateSelfContainedOptions(bool hasSelfContainedOption, bool hasNoSelfContainedOption, bool hasRuntimeOption, IEnumerable<string> projectArgs)
```
dsplaisted, line 118: Seems like we don't need the last two arguments anymore.

dsplaisted (approved):

PR #18838, opened 2021-07-09
Use MetadataInformation object and left  vs multiple right optimization

Anipik:
- add displayString to metadataInformation object to allow custom assembly names.
- closing streams early to avoid locking of package.
-  some code clean up

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
@@ -18,9 +19,13 @@ namespace Microsoft.DotNet.PackageValidation
     /// </summary>
     public class ApiCompatRunner
     {
-        private List<(string leftAssemblyPackagePath, MetadataInformation leftAssembly, string rightAssemblyPackagePath, MetadataInformation rightAssembly, string compatibilityReason, string header)> _queue = new();
+        private Dictionary<MetadataInformation, List<(MetadataInformation rightAssembly, string header)>> _dict = new();
         private readonly ApiComparer _differ = new();
         private readonly IPackageLogger _log;
+
+        private bool _isBaselineSuppression = false;
+        private string _leftPackagePath;
+        private string _rightPackagePath;
 
         public ApiCompatRunner(string noWarn, (string, string)[] ignoredDifferences, bool enableStrictMode, IPackageLogger log)
         {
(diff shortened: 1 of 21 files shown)


src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -31,63 +36,88 @@ public ApiCompatRunner(string noWarn, (string, string)[] ignoredDifferences, boo
...
-        /// <param name="compatibilityReason">The reason for assembly compatibilty.</param>
+        /// <param name="leftMetadataInfo">Metadata information for left assembly.</param>
+        /// <param name="rightMetdataInfo">Metadata information for right assembly.</param>
         /// <param name="header">The header for the api compat diagnostics.</param>
-        public void QueueApiCompat(string leftPackagePath, string leftRelativePath, string rightPackagePath, string rightRelativePath, string assemblyName, string compatibilityReason, string header)
+        public void QueueApiCompat(MetadataInformation leftMetadataInfo, MetadataInformation rightMetdataInfo, string header)
+        {
+            if (_dict.ContainsKey(leftMetadataInfo))
+            {
+                _dict[leftMetadataInfo].Add((rightMetdataInfo, header));
```
safern, line 98: Maybe the header can just be a resource formatted string and we can create it with the info we have when logging it out? This way we don't have to pass it around?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -31,63 +36,88 @@ public ApiCompatRunner(string noWarn, (string, string)[] ignoredDifferences, boo
...
         {
-            foreach (var apicompatTuples in _queue.Distinct())
+            foreach (MetadataInformation left in _dict.Keys)
             {
-                // TODO: Add optimisations tuples.
-                using (Stream leftAssemblyStream = GetFileStreamFromPackage(apicompatTuples.leftAssemblyPackagePath, apicompatTuples.leftAssemblyRelativePath))
-                using (Stream rightAssemblyStream = GetFileStreamFromPackage(apicompatTuples.rightAssemblyPackagePath, apicompatTuples.rightAssemblyRelativePath))
+                Stream leftAssemblyStream = GetFileStreamFromPackage(_leftPackagePath, left.AssemblyId);
+                IAssemblySymbol leftSymbols = new AssemblySymbolLoader().LoadAssembly(left.AssemblyName, leftAssemblyStream);
+                leftAssemblyStream.Close();
```
safern, line 47: Maybe wrap the stream on a using statement for it to get disposed rather than closing it? That would help dispose it if there is an exception?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -31,63 +36,88 @@ public ApiCompatRunner(string noWarn, (string, string)[] ignoredDifferences, boo
...
+                    _differ.GetDifferences(leftContainer, rightContainerList).ToList();
 
-                    string leftName = apicompatTuples.leftAssemblyRelativePath;
-                    bool isBaselineSuppression = false;
-                    if (!apicompatTuples.leftAssemblyPackagePath.Equals(apicompatTuples.rightAssemblyPackagePath, System.StringComparison.InvariantCultureIgnoreCase))
-                    {
-                        isBaselineSuppression = true;
-                        leftName = Resources.Baseline + " " + leftName;
-                    }
+                for (int i = 0; i < differences.Count(); i++)
```
safern, line 62: Instead use the Count property rather than the Linq one.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -31,63 +36,88 @@ public ApiCompatRunner(string noWarn, (string, string)[] ignoredDifferences, boo
...
-                    IAssemblySymbol rightSymbols = new AssemblySymbolLoader().LoadAssembly(apicompatTuples.assemblyName, rightAssemblyStream);
+                    Stream rightAssemblyStream = GetFileStreamFromPackage(_rightPackagePath, rightTuple.rightAssembly.AssemblyId);
+                    IAssemblySymbol rightSymbols = new AssemblySymbolLoader().LoadAssembly(rightTuple.rightAssembly.AssemblyName, rightAssemblyStream);
+                    rightAssemblyStream.Close();
+                    rightContainerList.Add(new ElementContainer<IAssemblySymbol>(rightSymbols, rightTuple.rightAssembly));
+                }
 
-                    _log.LogMessage(MessageImportance.Low, apicompatTuples.header);
+                IList<(MetadataInformation, MetadataInformation, IEnumerable<CompatDifference>)> differences =
+                    _differ.GetDifferences(leftContainer, rightContainerList).ToList();
```
safern, line 60: Do we really need to convert this to a List? or can we just enumerate the result and have a counter to access the dictionary?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -80,6 +82,7 @@ public void RunApiCompat()
                             difference.DiagnosticId,
                             difference.Message);
                     }
+                    counter++;
```
safern, line 85: nit: you can inline this to the usage of the variable as you only use it once inside the while. 

```cs
(MetadataInformation rightAssembly, string header) rightTuple = _dict[left][counter++];
```

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -31,63 +35,92 @@ public ApiCompatRunner(string noWarn, (string, string)[] ignoredDifferences, boo
...
+            else
+            {
+                _dict.Add(leftMetadataInfo, new List<(MetadataInformation rightAssembly, string header)>() { (rightMetdataInfo, header) });
+            }
+        }
+
+        internal void InitializePaths(string leftPackagePath, string rightPackagePath)
+        {
+            if (string.IsNullOrEmpty(leftPackagePath))
+                throw new ArgumentException(string.Format(Resources.InvalidPackagePath, leftPackagePath));
```
safern, line 112: what is the point of including the value of `leftPackagePath` and the value of `rightPackagePath` on these exceptions if it will be null or empty? I think we should instead pass the param name.

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
```diff
@@ -57,13 +59,11 @@ public void Validate(Package package)
...
-                        package.PackagePath,
-                        latestCompileTimeAsset.Path,
-                        package.PackageId,
-                        Resources.BaselineVersionValidatorHeader,
-                        string.Format(Resources.ApiCompatibilityBaselineHeader, baselineCompileTimeAsset.Path, latestCompileTimeAsset.Path, _baselinePackage.Version, package.Version));
+                    MetadataInformation left = new(package.PackageId, ((NuGetFramework)baselineCompileTimeAsset.Properties["tfm"]).GetShortFolderName(), baselineCompileTimeAsset.Path, Resources.Baseline + " " + baselineCompileTimeAsset.Path);
+                    MetadataInformation right = new(package.PackageId, ((NuGetFramework)latestCompileTimeAsset.Properties["tfm"]).GetShortFolderName(), latestCompileTimeAsset.Path);
+
+                    string header = string.Format(Resources.ApiCompatibilityBaselineHeader, baselineCompileTimeAsset.Path, latestCompileTimeAsset.Path, _baselinePackage.Version, package.Version);
+                    _apiCompatRunner.QueueApiCompat(left, right, header);
```
safern, line 66: This code is repeated three times, maybe extract it to a local/shared method?

src/Compatibility/Microsoft.DotNet.PackageValidation/CompatibleTFMValidator.cs
```diff
@@ -84,13 +87,11 @@ public void Validate(Package package)
...
-                            package.PackagePath,
-                            runtimeAsset.Path,
-                            package.PackageId,
-                            Resources.CompatibleTfmValidatorHeader,
-                            string.Format(Resources.ApiCompatibilityHeader, compileTimeAsset.Path, runtimeAsset.Path));
+                        MetadataInformation left = new(package.PackageId, ((NuGetFramework)compileTimeAsset.Properties["tfm"]).GetShortFolderName(), compileTimeAsset.Path);
+                        MetadataInformation right = new(package.PackageId, ((NuGetFramework)runtimeAsset.Properties["tfm"]).GetShortFolderName(), runtimeAsset.Path);
+
+                        string header = string.Format(Resources.ApiCompatibilityHeader, compileTimeAsset.Path, runtimeAsset.Path);
+                        _apiCompatRunner.QueueApiCompat(left, right, header);
```
safern, line 94: Duplicated code, can we share it somehow?

safern (approved):
Modulo some nits.

src/Compatibility/Microsoft.DotNet.PackageValidation/Helpers.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.ApiCompatibility.Abstractions;
+using NuGet.ContentModel;
+using NuGet.Frameworks;
+
+namespace Microsoft.DotNet.PackageValidation
+{
+    internal static class Helpers
```
safern, line 10: I would name this Extensions or something like that.

Anipik: i will do the name change in other open pr

Issue #18825, opened 2021-07-09
Additional C# global imports


JunTaoLuo:
There were some additional discussions on what global imports should or should not be included in the base .NET SDK. Specifically, there were concerns about:

✅ = currently included
❌ = not currently included

1. ✅  System.Net.Http: The contention is that this is used for web concerns so maybe it belongs in the Web SDK instead. However, it's also very frequently used in general, even console apps.
2. ❌  System.Net: This is often used in combination with System.Net.Http. We should put this in where-ever we end up putting System.Net.Http
3. ❌ System.Diagnostics: This has been used in some of the Web templates. Is this a common enough concern to be included in an SDK, be it the base .NET SDK or Web SDK?
4. ❌ System.Text: This is another commonly used namespace that we can consider for the base .NET SDK. Historically, this has been used 

In the interest of ensuring all discussion is happening in the same place, let's use this issue as much as possible to track discussions. @DamianEdwards @RussKie. 

FYI @davidfowl @halter73 @KathleenDollard @pranavkm @terrajobst

Feel free to tag anyone else who would be interested in this dicsussion. I've only included folks who have had additional input on the current import list.

DamianEdwards:
Here's the [PR that removes `using` statements from the web templates](https://github.com/dotnet/aspnetcore/pull/34219/files) based on what's implicitly added by the SDK now. The namespaces that are left are a potential place to start WRT to what to consider for further inclusion.

JunTaoLuo:
Also, as a reference here's the currently included list:

1. Base .NET SDK
```
    <Import Include="System" />
    <Import Include="System.Collections.Generic" />
    <Import Include="System.IO" />
    <Import Include="System.Linq" />
    <Import Include="System.Net.Http" />
    <Import Include="System.Threading" />
    <Import Include="System.Threading.Tasks" />
```
2. Web SDK (also includes everying in the base .NET SDK)
```
    <Import Include="System.Net.Http.Json" />
    <Import Include="Microsoft.AspNetCore.Builder" />
    <Import Include="Microsoft.AspNetCore.Hosting" />
    <Import Include="Microsoft.AspNetCore.Http" />
    <Import Include="Microsoft.AspNetCore.Routing" />
    <Import Include="Microsoft.Extensions.Configuration" />
    <Import Include="Microsoft.Extensions.DependencyInjection" />
    <Import Include="Microsoft.Extensions.Hosting" />
    <Import Include="Microsoft.Extensions.Logging" />
```
[...]

Issue #18839, opened 2021-07-09
Naming of the properties that control implicit namespaces included by the SDK


JunTaoLuo:
Follow up to the conversation started in https://github.com/dotnet/sdk/pull/18459#discussion_r664925999.

We don't like the underscore in the naming of the properties that control the list of implicit imports. For example, we have the name `<DisableImplicitNamespaceImports_Web>`. However, we also need to ensure that these options are very specific and unlikely to collide with other properties. 

FYI @mhutch @DamianEdwards @davidfowl

pranavkm:
Closing this since it's been shipped. 👋🏽 @JunTaoLuo!

PR #18850, opened 2021-07-10
Added Microsoft.Extensions.* default using to blazor wasm

davidfowl:
- This is a minimum set to fix the default template having removed these extension based usings.
- We'll want to add more in the future

See https://github.com/dotnet/sdk/issues/18825#issuecomment-877668329 for more details


src/BlazorWasmSdk/Sdk/Sdk.props
```diff
@@ -19,4 +19,12 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
 
   <Import Project="$(_BlazorWebAssemblyPropsFile)" />
+
+  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports_BlazorWasm)' != 'true'
+      And '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+      And $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '6.0'))">
+    <Import Include="Microsoft.Extensions.Configuration" />
+    <Import Include="Microsoft.Extensions.DependencyInjection" />
+    <Import Include="Microsoft.Extensions.Logging" />
```
javiercn, line 28: Ugh, is `Import` the itemgroup that we chose for this?

`Import` is already a keyword in MSBuild (see [here](https://docs.microsoft.com/en-us/visualstudio/msbuild/import-element-msbuild?view=vs-2019)), we should reconsider the name to avoid confusion.

This will just make navigating msbuild files more miserable

src/BlazorWasmSdk/Sdk/Sdk.targets
```diff
@@ -13,6 +13,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <PropertyGroup>
     <_BlazorWebAssemblyTargetsFile Condition="'$(_BlazorWebAssemblyTargetsFile)' == ''">$(MSBuildThisFileDirectory)..\targets\Microsoft.NET.Sdk.BlazorWebAssembly.Current.targets</_BlazorWebAssemblyTargetsFile>
+    <DisableImplicitNamespaceImports_BlazorWasm Condition="'$(DisableImplicitNamespaceImports_BlazorWasm)' == '' and '$(DisableImplicitNamespaceImports)' != 'true'">false</DisableImplicitNamespaceImports_BlazorWasm>
```
javiercn, line 16: `BlazorWebassemblyDisableImplicitNamespaceImports` if you expect this to be a `public` property. We don't use `_` anywhere afaik on our SDKs

davidfowl: See https://github.com/dotnet/sdk/issues/18839

src/BlazorWasmSdk/Sdk/Sdk.props
davidfowl: Yea it's aligned with VB

JunTaoLuo: Yes, I brought up the same concern, but we felt aligning the behaviour is more important. Also, global using was partially inspired by the `Import` mechanism of VB it turns out.

src/BlazorWasmSdk/Sdk/Sdk.targets
JunTaoLuo: Wouldn't it be nice if we had namespaces for MSBuild properties 😄

JunTaoLuo (commented):
Changes look good but add some tests: https://github.com/dotnet/sdk/blob/main/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateImplicitNamespaceImports_WebApp.cs

JunTaoLuo (approved):
Approving to unblock https://github.com/dotnet/installer/pull/11100. We should file an issue to add the necessary tests though.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateImplicitNamespaceImports_BlazorWasm.cs
```diff
@@ -0,0 +1,105 @@
...
+            var importFileName = $"{testAsset.TestProject.Name}.ImplicitNamespaceImports.cs";
+
+            var buildCommand = new BuildCommand(testAsset);
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
+
+            var outputDirectory = buildCommand.GetIntermediateDirectory(tfm);
+
```
JunTaoLuo, line 34:
```suggestion
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateImplicitNamespaceImports_BlazorWasm.cs
```diff
@@ -0,0 +1,104 @@
...
+                IsExe = true,
+                TargetFrameworks = tfm,
+                ProjectSdk = "Microsoft.NET.Sdk.BlazorWebAssembly"
+            };
+            testProject.PackageReferences.Add(new TestPackageReference("Microsoft.AspNetCore.Components.WebAssembly", "5.0.0"));
+            testProject.SourceFiles["Program.cs"] = @"
+using Microsoft.AspNetCore.Components.WebAssembly.Hosting;
+
+var builder = WebAssemblyHostBuilder.CreateDefault(args);
+
```
davidfowl, line 96: I removed adding the component so I could avoid making the template more complex.

JunTaoLuo (approved):
When tests pass.

captainsafia (approved):

Issue #18824, opened 2021-07-09
Move VB Import items to props files


JunTaoLuo:
This would allow the project files themselves to remove imports.

This pattern is used for Import items for C# projects in https://github.com/dotnet/sdk/pull/18459. We will need to add a test like: https://github.com/dotnet/sdk/blob/a6b38f0b1f4a6829e327fbb6c7d692c1ed125d1a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateImplicitNamespaceImports_DotNet.cs#L71

JunTaoLuo:
The PR that resolves this has been merged.

PR #18853, opened 2021-07-11
Relocate VB import items to props file

JunTaoLuo:
Fixes #https://github.com/dotnet/sdk/issues/18824

Is `GivenThatICareAboutVBApps` the right place to add the relevant tests?

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.props
@@ -30,29 +30,4 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <DefineTrace Condition=" '$(DefineTrace)' == '' ">true</DefineTrace>
   </PropertyGroup>
-
-  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework'">
-    <!-- These namespaces are present in 2.0 Framework assemblies -->
-    <Import Include="Microsoft.VisualBasic" />
-    <Import Include="System" />
-    <Import Include="System.Collections" />
-    <Import Include="System.Collections.Generic" />
-    <Import Include="System.Data" />
-    <Import Include="System.Diagnostics" />
-    <!-- These namespaces are introduced in 3.5 Framework assemblies -->
-    <Import Include="System.Linq" Condition=" '$(_TargetFrameworkVersionWithoutV)' >= '3.5' "/>
-    <Import Include="System.Xml.Linq" Condition=" '$(_TargetFrameworkVersionWithoutV)' >= '3.5' "/>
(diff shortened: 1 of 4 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.props
```diff
@@ -31,4 +31,28 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DefineTrace Condition=" '$(DefineTrace)' == '' ">true</DefineTrace>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework'">
```
dsplaisted, line 34: Instead of moving these to Microsoft.NET.Sdk.VisualBasic.props, can we put them in Microsoft.NET.Sdk.DefaultItems.props, and unify them with the C# ones?  As is a VB project will get duplicate `Import` items for the namespaces used by both C# and Visual Basic.

JunTaoLuo: Good catch regarding the duplication.

I'm a little concerned with combining them since some of the namespaces wouldn't make sense. For example, we definitely wouldn't include Microsoft.VisualBasic in DefaultItems.props. I can also see System.Collections, System.Data and System.Xml.Linq being controversial. Also, the TFM requirements are different.

An alternative is to remove the duplication here but keep VB specific imports in this file. I can see that being an issue since the imports will be split over two file which aren't as clear cut as dotnet/web divide. I'm not a fan of this either. 

I think what might make more sense is to move the imports in `Microsoft.NET.Sdk.DefaultItems.props` to CSharp.props. I think that might make more sense since those are intended to be consumed for C# projects. It also avoids splitting the namespaces imported here.

dsplaisted: Yes, what I'm suggesting is that you wouldn't change any of the default `Import` items for C# or VB with this PR.  I think it would be good to have them all in one file though, with conditions based on the `Language` property.

JunTaoLuo: Hmm is it a good idea to put everything in a single file? There will be lots of conditions and I think it'll look messy. Also, will this extend to other languages in the future beyond just C# and VB?

dsplaisted: My thinking is that if it's messy in one file, then it's also messy in two files, it's just harder to see and to reason about.

JunTaoLuo: Sure I can put it in a file and show what it would look like. Personally, I lean toward language specific props files so at least it's somewhat more manageable.

JunTaoLuo:
This is ready for a final review.

dsplaisted (approved):

PR #18865, opened 2021-07-12
add strict mode tests & fix assembly token strict mode bug

Anipik:
- add some strict mode tests
- while updating the version in the runtime, i discovered a bug where we were missing a token equality check.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
@@ -75,7 +75,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
                     rightName,
                     rightIdentity));
             }
-            else if (Settings.StrictMode && !rightAssemblyPublicKeyToken.IsEmpty && !isRightRetargetable)
+            else if (Settings.StrictMode && !rightAssemblyPublicKeyToken.IsEmpty && !isRightRetargetable && !rightAssemblyPublicKeyToken.SequenceEqual(leftAssemblyPublicKeyToken))
             {
                 differences.Add(CreateIdentityDifference(
                     Resources.AssemblyPublicKeyTokenDoesNotMatch,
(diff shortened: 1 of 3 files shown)


src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
```diff
@@ -181,12 +187,27 @@ public void LeftAssemblyKeyTokenNull()
             IAssemblySymbol rightSymbols = new AssemblySymbolLoader().LoadAssembly(rightDllPath);
 
             ApiComparer differ = new();
-            IEnumerable<CompatDifference> differences = differ.GetDifferences(leftSymbols, rightSymbols);
-            Assert.Empty(differences);
+            differ.StrictMode = strictMode;
+
+            if (strictMode)
+            {
+                IEnumerable<CompatDifference> differences = differ.GetDifferences(leftSymbols, rightSymbols);
```
safern, line 194: You could call this before the if/else and then just assert inside the if/else.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -75,7 +75,7 @@ private void RunOnAssemblySymbol(IAssemblySymbol left, IAssemblySymbol right, st
                     rightName,
                     rightIdentity));
             }
-            else if (Settings.StrictMode && !rightAssemblyPublicKeyToken.IsEmpty && !isRightRetargetable)
+            else if (Settings.StrictMode && !rightAssemblyPublicKeyToken.IsEmpty && !isRightRetargetable && !rightAssemblyPublicKeyToken.SequenceEqual(leftAssemblyPublicKeyToken))
```
safern, line 78: if the rightAssemblyPublicToken.IsEmpty and the left is not and the right assembly is not retargetable this would not add a difference, right?

Same for the if above with left pkt.

Anipik: right (empty and NonRetargetable left (non-empty and non-retargetable) will result in mismatch in normal mode only. we wont need to check for this condition.

safern: Right but if the public key token is empty and the assembly is not retargetable, then we would not error the difference, correct?

Anipik: yes if "base" dll token is empty then comparable dll can be null or not null and we wont error the difference.

safern (approved):
LGTM

PR #18874, opened 2021-07-12
[API Compat] Make sure test syntax trees don't have any syntax errors

safern:
Since the API Compat tests didn't have any validation that the syntax trees were actually syntactically correct, they had a bunch of syntax errors. This fixes it and prevents from happening in the future.

### DIFF
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs
@@ -8,27 +8,34 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using Xunit;
 
 namespace Microsoft.DotNet.ApiCompatibility.Tests
 {
     internal static class SymbolFactory
     {
-        internal static IAssemblySymbol GetAssemblyFromSyntax(string syntax, bool enableNullable = false, bool includeDefaultReferences = false, [CallerMemberName] string assemblyName = "")
+        internal static IAssemblySymbol GetAssemblyFromSyntax(string syntax, bool enableNullable = false, [CallerMemberName] string assemblyName = "")
         {
-            CSharpCompilation compilation = CreateCSharpCompilationFromSyntax(syntax, assemblyName, enableNullable, includeDefaultReferences);
+            CSharpCompilation compilation = CreateCSharpCompilationFromSyntax(syntax, assemblyName, enableNullable);
+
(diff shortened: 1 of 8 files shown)


src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs
```diff
@@ -46,27 +53,48 @@ internal static IList<ElementContainer<IAssemblySymbol>> GetElementContainersFro
...
+                new string[] { "NETFRAMEWORK" }
+#else
+                Array.Empty<string>()
+#endif
+        ); 
+
+        private static IEnumerable<KeyValuePair<string, ReportDiagnostic>> DiagnosticOptions { get; } = new[]
+        {
+            // Suppress warning for unused fields.
+            new KeyValuePair<string, ReportDiagnostic>("CS0067", ReportDiagnostic.Suppress)
```
ericstj, line 95: Where do we need this?  I wouldn't expect the compiler to ever complain about publicly visible unused fields.

ericstj (approved):
Thanks for fixing!

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs
safern: I messed up the comment it is not unused fields it was unused events in the declaring class. https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0067 

I'll update the comment.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs
```diff
@@ -46,27 +53,48 @@ internal static IList<ElementContainer<IAssemblySymbol>> GetElementContainersFro
...
+#if NETFRAMEWORK
+                new string[] { "NETFRAMEWORK" }
+#else
+                Array.Empty<string>()
+#endif
+        ); 
+
+        private static IEnumerable<KeyValuePair<string, ReportDiagnostic>> DiagnosticOptions { get; } = new[]
+        {
+            // Suppress warning for unused fields.
```
safern, line 94:
```suggestion
            // Suppress warning for unused events.
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SymbolFactory.cs
ericstj: I see, that one makes sense to suppress.

PR #18877, opened 2021-07-13
Fix matching analyzers from packages to TargetFramework

### LINKED ISSUE #18148: Conditional PackageReference version splitting analyzer presence fails
If a project references two different versions of the same package in two different TargetFrameworks, and the package has added or removed a Roslyn analyzer, the analyzer will be added from _both_ versions of the package, even the one where it's not present.

```xml

  <PropertyGroup>
    <TargetFrameworks>net6.0;net472</TargetFrameworks>
    <RootNamespace>two_systemtextjson</RootNamespace>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.Text.Json" Version="4.7.0" Condition=" '$(TargetFramework)' == 'net472' " />
    <PackageReference Include="System.Text.Json" Version="6.0.0-preview.4.21253.7" Condition=" '$(TargetFramework)' == 'net6.0' " />
  </ItemGroup>

</Project>
```

```sh-session
two-systemtextjson on master .NET v6.0.100-preview.5.21302.13 🎯 net6.0;net472 
❯ dotnet build  
Microsoft (R) Build Engine version 17.0.0-preview-21302-02+018bed83d for .NET

  Restored S:\work\two-systemtextjson\two-systemtextjson.csproj (in 135 ms).
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
CSC : error CS0006: Metadata file 'C:\Users\raines\.nuget\packages\system.text.json\4.7.0\analyzers\dotnet\cs\System.Text.Json.SourceGeneration.dll' could not be found [S:\work\two-systemtextjson\two-systemtextjson.csproj]
  two-systemtextjson -> S:\work\two-systemtextjson\bin\Debug\net6.0\two-systemtextjson.dll

Build FAILED.

CSC : error CS0006: Metadata file 'C:\Users\raines\.nuget\packages\system.text.json\4.7.0\analyzers\dotnet\cs\System.Text.Json.SourceGeneration.dll' could not be found [S:\work\two-systemtextjson\two-systemtextjson.csproj]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:01.20
```

Because 4.7.0 of that package didn't have an analyzer/source generator, the file isn't found in that package version. The reference is only there because it was found in the other version.

dsplaisted:
Fixes #18148

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
@@ -14,6 +14,7 @@ using Microsoft.Build.Utilities;
 using NuGet.Common;
 using NuGet.Frameworks;
 using NuGet.ProjectModel;
+using NuGet.Versioning;
 
 namespace Microsoft.NET.Build.Tasks
 {
@@ -824,7 +825,7 @@ namespace Microsoft.NET.Build.Tasks
 
             private void WriteAnalyzers()
             {
-                Dictionary<string, LockFileTargetLibrary> targetLibraries = null;
+                Dictionary<(string, NuGetVersion), LockFileTargetLibrary> targetLibraries = null;
 
                 foreach (var library in _lockFile.Libraries)
                 {
@@ -842,12 +843,10 @@ namespace Microsoft.NET.Build.Tasks
 
                         if (targetLibraries == null)
                         {
-                            targetLibraries = _runtimeTarget
-                                .Libraries
-                                .ToDictionary(l => l.Name, StringComparer.OrdinalIgnoreCase);
+                            targetLibraries = _compileTimeTarget.Libraries.ToDictionary(l => (l.Name, l.Version));
                         }
 
-                        if (targetLibraries.TryGetValue(library.Name, out var targetLibrary))
+                        if (targetLibraries.TryGetValue((library.Name, library.Version), out var targetLibrary))
                         {
                             WriteItem(_packageResolver.ResolvePackageAssetPath(targetLibrary, file), targetLibrary);
                         }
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -842,12 +843,10 @@ private void WriteAnalyzers()
 
                         if (targetLibraries == null)
                         {
-                            targetLibraries = _runtimeTarget
-                                .Libraries
-                                .ToDictionary(l => l.Name, StringComparer.OrdinalIgnoreCase);
```
rainersigwald, line 847: Is it a problem to lose the case-insensitivity?

rainersigwald (approved):

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
dsplaisted: I was wondering the same thing.  I think that it's probably safe to assume that package IDs use consistent casing in the assets file.  @nkolev92 @zkat Does that sound correct?

nkolev92: It really depends.
Some sections of the assets file represent user input and those honor the user casing, targets/libraries sections should be ok, but I don't think we have any code that assumes that.

So it really depends on what you are comparing.

dsplaisted: @nkolev92 In this case we're comparing `LockFileTargetLibrary` and `LockFileLibrary`.

nkolev92: I think it should be safe then. 

You can check that easily by testing with `newtonsoft.json` or `nuget.common`.

PR #18889, opened 2021-07-13
Adding OS and arch command line options

sfoslund:
Fixes first bullet of #18832

## Description
This change adds `--os` and `--arch` options to build, publish, run, and test commands. This is a shorthand syntax for the runtime identifier, where the provided values will be combined with the default RID. For example, on a win-x64 machine, specifying `--os os` will result in running with the RID set to `os-x64` and specifying `--arch arch` will result in `win-arch`. Full spec here: https://github.com/dotnet/designs/blob/d248ad68b41866eba17e80225a08518b8202ddf6/accepted/2021/architecture-targeting.md#enable-shorthand-rid-targeting-syntax
These changes are all additive and do not impact the existing CLI.

## Regression?
No

## Risk
Low, this is a new feature and won't effect existing CLI. 

## Verification
[X] Manual (required)
[x] Automated

### DIFF
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -9,6 +9,7 @@ using Microsoft.DotNet.Tools.Common;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.DotNet.Cli.Utils;
+using System.CommandLine.Parsing;
 
 namespace Microsoft.DotNet.Cli
 {
@@ -116,18 +117,33 @@ namespace Microsoft.DotNet.Cli
                 verbosity.Equals(VerbosityOptions.detailed);
         }
 
-        internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg)
+        internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg, ParseResult parseResult)
         {
-            var currentRid = GetCurrentRuntimeId();
-            var os = currentRid.Substring(0, currentRid.LastIndexOf("-"));
-            return new string[] { $"-property:RuntimeIdentifier={os}-{arg}", "-property:SelfContained=true" };
+            if (parseResult.BothArchAndOsOptionsSpecified())
+            {
+                return Array.Empty<string>();
+            }
+
+            return ResolveRidShorthandOptions(null, arg);
         }
 
-        internal static IEnumerable<string> ResolveOsOptionToRuntimeIdentifier(string arg)
+        internal static IEnumerable<string> ResolveOsOptionToRuntimeIdentifier(string arg, ParseResult parseResult)
+        {
+            if (parseResult.BothArchAndOsOptionsSpecified())
+            {
+                return ResolveRidShorthandOptions(arg, parseResult.ValueForOption<string>(CommonOptions.ArchitectureOption().Aliases.First()));
+            }
+
+            return ResolveRidShorthandOptions(arg, null);
+        }
+
+        private static IEnumerable<string> ResolveRidShorthandOptions(string os, string arch)
         {
             var currentRid = GetCurrentRuntimeId();
-            var arch = currentRid.Substring(currentRid.LastIndexOf("-") + 1, currentRid.Length - currentRid.LastIndexOf("-") - 1);
-            return new string[] { $"-property:RuntimeIdentifier={arg}-{arch}", "-property:SelfContained=true" };
+            os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            var properties = new string[] { $"-property:RuntimeIdentifier={os}-{arch}" };
+            return properties;
         }
 
         private static string GetCurrentRuntimeId()
@@ -147,6 +163,10 @@ namespace Microsoft.DotNet.Cli
             }
             return currentRuntimeIdentifiers[0]; // First rid is the most specific (ex win-x64)
         }
+
+        private static string GetOsFromRid(string rid) => rid.Substring(0, rid.LastIndexOf("-"));
+
+        private static string GetArchFromRid(string rid) => rid.Substring(rid.LastIndexOf("-") + 1, rid.Length - rid.LastIndexOf("-") - 1);
     }
 
     public enum VerbosityOptions
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -6,6 +6,9 @@ using Microsoft.DotNet.Tools;
 using System.CommandLine;
 using System.IO;
 using Microsoft.DotNet.Tools.Common;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
 
 namespace Microsoft.DotNet.Cli
 {
@@ -35,7 +38,7 @@ namespace Microsoft.DotNet.Cli
                 description)
             {
                 ArgumentHelpName = CommonLocalizableStrings.FrameworkArgumentName
-                    
+
             }.ForwardAsSingle(o => $"-property:TargetFramework={o}")
             .AddSuggestions(Suggest.TargetFrameworksFromProjectFile());
 
@@ -91,6 +94,18 @@ namespace Microsoft.DotNet.Cli
                 "--interactive",
                 CommonLocalizableStrings.CommandInteractiveOptionDescription);
 
+        public static Option ArchitectureOption(bool includeShortVersion = true) =>
+            new ForwardedOption<string>(
+                includeShortVersion ? new string[] { "--arch", "-a" } : new string[] { "--arch" },
+                CommonLocalizableStrings.ArchitectureOptionDescription)
+            .SetForwardingFunction(ResolveArchOptionToRuntimeIdentifier);
+
+        public static Option OperatingSystemOption() =>
+            new ForwardedOption<string>(
+                "--os",
+                CommonLocalizableStrings.OperatingSystemOptionDescription)
+            .SetForwardingFunction(ResolveOsOptionToRuntimeIdentifier);
+
         public static Option DebugOption() => new Option<bool>("--debug");
 
         public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosity)
@@ -99,6 +114,38 @@ namespace Microsoft.DotNet.Cli
                 verbosity.Equals(VerbosityOptions.diagnostic) ||
                 verbosity.Equals(VerbosityOptions.d) ||
                 verbosity.Equals(VerbosityOptions.detailed);
+        }
+
+        internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg)
+        {
+            var currentRid = GetCurrentRuntimeId();
+            var os = currentRid.Substring(0, currentRid.LastIndexOf("-"));
+            return new string[] { $"-property:RuntimeIdentifier={os}-{arg}", "-property:SelfContained=true" };
+        }
+
+        internal static IEnumerable<string> ResolveOsOptionToRuntimeIdentifier(string arg)
+        {
+            var currentRid = GetCurrentRuntimeId();
+            var arch = currentRid.Substring(currentRid.LastIndexOf("-") + 1, currentRid.Length - currentRid.LastIndexOf("-") - 1);
+            return new string[] { $"-property:RuntimeIdentifier={arg}-{arch}", "-property:SelfContained=true" };
+        }
+
+        private static string GetCurrentRuntimeId()
+        {
+            var dotnetRootPath = Path.GetDirectoryName(Environment.ProcessPath);
+            dotnetRootPath = dotnetRootPath.Contains("dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
+            var ridFileName = "NETCoreSdkRuntimeIdentifierChain.txt";
+            string runtimeIdentifierChainPath = string.IsNullOrEmpty(Product.Version) ?
+                Path.Combine(Directory.GetDirectories(Path.Combine(dotnetRootPath, "sdk"))[0], ridFileName) :
+                Path.Combine(dotnetRootPath, "sdk", Product.Version, ridFileName);
+            string[] currentRuntimeIdentifiers = File.Exists(runtimeIdentifierChainPath) ?
+                File.ReadAllLines(runtimeIdentifierChainPath).Where(l => !string.IsNullOrEmpty(l)).ToArray() :
+                new string[] { };
+            if (currentRuntimeIdentifiers == null || !currentRuntimeIdentifiers.Any() || !currentRuntimeIdentifiers[0].Contains("-"))
+            {
+                throw new GracefulException(CommonLocalizableStrings.CannotResolveRuntimeIdentifier);
+            }
+            return currentRuntimeIdentifiers[0]; // First rid is the most specific (ex win-x64)
         }
     }
 
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -119,8 +119,14 @@ namespace Microsoft.DotNet.Cli
 
         internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg, ParseResult parseResult)
         {
+            if (parseResult.HasOption(RuntimeOption(string.Empty).Aliases.First()))
+            {
+                throw new GracefulException(CommonLocalizableStrings.CannotSpecifyBothRuntimeAndArchOptions);
+            }
+
             if (parseResult.BothArchAndOsOptionsSpecified())
             {
+                // ResolveOsOptionToRuntimeIdentifier handles resolving the RID when both arch and os are specified
                 return Array.Empty<string>();
             }
 
@@ -129,6 +135,11 @@ namespace Microsoft.DotNet.Cli
 
         internal static IEnumerable<string> ResolveOsOptionToRuntimeIdentifier(string arg, ParseResult parseResult)
         {
+            if (parseResult.HasOption(RuntimeOption(string.Empty).Aliases.First()))
+            {
+                throw new GracefulException(CommonLocalizableStrings.CannotSpecifyBothRuntimeAndOsOptions);
+            }
+
             if (parseResult.BothArchAndOsOptionsSpecified())
             {
                 return ResolveRidShorthandOptions(arg, parseResult.ValueForOption<string>(CommonOptions.ArchitectureOption().Aliases.First()));
@@ -139,11 +150,16 @@ namespace Microsoft.DotNet.Cli
 
         private static IEnumerable<string> ResolveRidShorthandOptions(string os, string arch)
         {
+            var properties = new string[] { $"-property:RuntimeIdentifier={ResolveRidShorthandOptionsToRuntimeIdentifier(os, arch)}" };
+            return properties;
+        }
+
+        internal static string ResolveRidShorthandOptionsToRuntimeIdentifier(string os, string arch)
+        {
             var currentRid = GetCurrentRuntimeId();
             os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
             arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
-            var properties = new string[] { $"-property:RuntimeIdentifier={os}-{arch}" };
-            return properties;
+            return $"{os}-{arch}";
         }
 
         private static string GetCurrentRuntimeId()
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -165,7 +165,8 @@ namespace Microsoft.DotNet.Cli
         private static string GetCurrentRuntimeId()
         {
             var dotnetRootPath = Path.GetDirectoryName(Environment.ProcessPath);
-            dotnetRootPath = dotnetRootPath.Contains("dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
+            // When running under test the path does not always contain "dotnet" and Product.Version is empty.
+            dotnetRootPath = Path.GetFileName(dotnetRootPath).Contains("dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
             var ridFileName = "NETCoreSdkRuntimeIdentifierChain.txt";
             string runtimeIdentifierChainPath = string.IsNullOrEmpty(Product.Version) ?
                 Path.Combine(Directory.GetDirectories(Path.Combine(dotnetRootPath, "sdk"))[0], ridFileName) :
--- a/src/Cli/dotnet/commands/dotnet-run/RunCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/RunCommandParser.cs
@@ -47,6 +47,8 @@ namespace Microsoft.DotNet.Cli
             command.AddOption(InteractiveOption);
             command.AddOption(NoRestoreOption);
             command.AddOption(CommonOptions.VerbosityOption());
+            command.AddOption(CommonOptions.ArchitectureOption());
+            command.AddOption(CommonOptions.OperatingSystemOption());
             command.TreatUnmatchedTokensAsErrors = false;
 
             return command;
--- a/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
@@ -138,6 +138,8 @@ namespace Microsoft.DotNet.Cli
             command.AddOption(NoRestoreOption);
             command.AddOption(CommonOptions.InteractiveMsBuildForwardOption());
             command.AddOption(CommonOptions.VerbosityOption());
+            command.AddOption(CommonOptions.ArchitectureOption(false));
+            command.AddOption(CommonOptions.OperatingSystemOption());
 
             return command;
         }
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
@@ -55,6 +55,8 @@ namespace Microsoft.DotNet.Cli
             command.AddOption(ToolCommandRestorePassThroughOptions.NoCacheOption);
             command.AddOption(ToolCommandRestorePassThroughOptions.InteractiveRestoreOption);
             command.AddOption(VerbosityOption);
+            command.AddOption(CommonOptions.ArchitectureOption());
+            command.AddOption(CommonOptions.OperatingSystemOption());
 
             return command;
         }
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
@@ -55,8 +55,6 @@ namespace Microsoft.DotNet.Cli
             command.AddOption(ToolCommandRestorePassThroughOptions.NoCacheOption);
             command.AddOption(ToolCommandRestorePassThroughOptions.InteractiveRestoreOption);
             command.AddOption(VerbosityOption);
-            command.AddOption(CommonOptions.ArchitectureOption());
-            command.AddOption(CommonOptions.OperatingSystemOption());
 
             return command;
         }
--- a/src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
+++ b/src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
@@ -100,7 +100,7 @@ namespace Microsoft.DotNet.Cli.MSBuild.Tests
                 var expectedArch = Environment.Is64BitOperatingSystem ? "x64" : "x86";
                 command.GetArgumentsToMSBuild()
                     .Should()
-                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=os-{expectedArch} -property:SelfContained=true");
+                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=os-{expectedArch}");
             });
         }
 
@@ -113,7 +113,20 @@ namespace Microsoft.DotNet.Cli.MSBuild.Tests
                 var command = BuildCommand.FromArgs(new string[] { "--arch", "arch" }, msbuildPath);
                 command.GetArgumentsToMSBuild()
                     .Should()
-                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=win-arch -property:SelfContained=true");
+                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=win-arch");
+            });
+        }
+
+        [Fact]
+        public void OSAndArchOptionsCanBeCombined()
+        {
+            CommandDirectoryContext.PerformActionWithBasePath(WorkingDirectory, () =>
+            {
+                var msbuildPath = "<msbuildpath>";
+                var command = BuildCommand.FromArgs(new string[] { "--arch", "arch", "--os", "os" }, msbuildPath);
+                command.GetArgumentsToMSBuild()
+                    .Should()
+                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=os-arch");
             });
         }
     }
--- a/src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
+++ b/src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
@@ -4,6 +4,10 @@
 using Microsoft.DotNet.Tools.Build;
 using FluentAssertions;
 using Xunit;
+using Microsoft.DotNet.Cli.Utils;
+using System;
+using Microsoft.NET.TestFramework;
+using Microsoft.DotNet.Tools;
 
 namespace Microsoft.DotNet.Cli.MSBuild.Tests
 {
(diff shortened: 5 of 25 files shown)


vitek-karas:
Can you please add a short description of what the change is supposed to do? (Basically a mini-spec for the change)

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
```diff
@@ -84,7 +88,33 @@ public void MsbuildInvocationIsCorrect(string[] args, string expectedAdditionalA
...
+        }
 
+        [Fact]
+        public void OsOptionIsCorrectlyResolved()
+        {
+            CommandDirectoryContext.PerformActionWithBasePath(WorkingDirectory, () =>
+            {
+                var msbuildPath = "<msbuildpath>";
+                var command = BuildCommand.FromArgs(new string[] { "--os", "os" }, msbuildPath);
+                var expectedArch = Environment.Is64BitOperatingSystem ? "x64" : "x86";
```
joeloff, line 100: What if the OS is arm64?

sfoslund: This is just a test, so it's not accounting for that right now.

sfoslund:
>Can you please add a short description of what the change is supposed to do? (Basically a mini-spec for the change)

Sorry, forgot to link the issue/ spec: https://github.com/dotnet/sdk/issues/18832, this PR covers the first bullet in that issue.

richlander:
@sfoslund -- it would be best to put that information (in response to Vitek) in the PR description. That makes it much easier for others coming to this PR.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +115,38 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
                 verbosity.Equals(VerbosityOptions.d) ||
                 verbosity.Equals(VerbosityOptions.detailed);
         }
+
+        internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg)
+        {
+            var currentRid = GetCurrentRuntimeId();
+            var os = currentRid.Substring(0, currentRid.LastIndexOf("-"));
+            return new string[] { $"-property:RuntimeIdentifier={os}-{arg}", "-property:SelfContained=true" };
```
richlander, line 123: Per my read, the default when using the new shorthand combinations is `--selfcontained true`. That's not the intention, but the opposite.

So, two things:

- The default should the `--self-contained false`
- It should be possible for someone to set `--self-contained true`.

For example, the first of these should be framework-dependent and the other self-contained:

- `dotnet build -a x64`
- `dotnet build -a x64 --self-contained`

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +115,38 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+        }
+
+        internal static IEnumerable<string> ResolveOsOptionToRuntimeIdentifier(string arg)
+        {
+            var currentRid = GetCurrentRuntimeId();
+            var arch = currentRid.Substring(currentRid.LastIndexOf("-") + 1, currentRid.Length - currentRid.LastIndexOf("-") - 1);
+            return new string[] { $"-property:RuntimeIdentifier={arg}-{arch}", "-property:SelfContained=true" };
+        }
+
+        private static string GetCurrentRuntimeId()
```
richlander, line 133: There is a lot of logic here. It seems like this method would be a lot easier to use if you added a tiny bit more logic, split the RID and then returned a tuple or a struct. This line would be longer be needed with each usage: https://github.com/dotnet/sdk/pull/18889/files#diff-99e0f27176e2b38367e3898ec98c6d1a67f2cb629195156f4bc1c57b2ef6670bR129.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +115,38 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
                 verbosity.Equals(VerbosityOptions.d) ||
                 verbosity.Equals(VerbosityOptions.detailed);
         }
+
+        internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg)
```
richlander, line 119: What happens if both `-a` and `--os` are specified? I'm not seeing how that case is handled. There should be a test for that.

richlander:
Modulo my feedback, this looks good. Thanks!

I'm still left wondering how specifying a RID will behave for new scenarios, like `dotnet tool install`. Will that be a subsequent change?

I'm more than happy to test these changes to validate that they play out correctly in an actual build. If you share a build with me, I'll test it (in a timely manner).

sfoslund:
> I'm still left wondering how specifying a RID will behave for new scenarios, like dotnet tool install. Will that be a subsequent change?

Can you clarify what you mean here? I included it in this change because it was in your spec.

richlander:
> Can you clarify what you mean here? I included it in this change because it was in your spec.

Yes, it is in the spec. What I mean is, if you type `dotnet tool install dotnet-runtimeinfo -a x64`, is an x64 apphost generated instead of an Arm64 one (assuming you are using the Arm64 SDK)?

sfoslund:
>Yes, it is in the spec. What I mean is, if you type dotnet tool install dotnet-runtimeinfo -a x64, is an x64 apphost generated instead of an Arm64 one (assuming you are using the Arm64 SDK)?

Right now these options aren't hooked up to anything because I wasn't clear based on the spec what the intended behavior was. Is that the intended behavior? If so I can implement it.

richlander:
I could have been more clear/prescriptive on this point. This text (from the spec) is the closest I got to on that.

> Tool installation with `dotnet tool install` is a particularly problematic scenario. At the time .NET 6 is released, one can expect that close to 100% of .NET tools (packaged via NuGet) will target a version prior to .NET 6 and therefore be incapable on running on Arm64 (at least on macOS). However, when installed, the SDK will generate an apphost that matches the SDK RID. As a result, the tool will consistently fail to launch (since the required runtime will be missing). The SDK needs to apply the same implicit RID rules for .NET tools that it will apply for `dotnet build`. In addition, `dotnet tool install` should enable explicit RID targeting.

> Note: One can think of .NET tools as being distributed as portable apps with no apphost, with the most favorable apphost being provided just-in-time at installation. It's a pretty good model and aligns with the spirit of the overall design.

The same behavior is expected for `dotnet test` and `dotnet run` (and `dotnet watch`, which is a conceptual wrapper over those too).

sfoslund:
> The same behavior is expected for dotnet test and dotnet run 

I see, dotnet run and test are already enabled by this PR but I'll integrate this into dotnet tool install. 

> (and dotnet watch, which is a conceptual wrapper over those too).

This PR does not cover dotnet watch as my team doesn't own that code. @captainsafia are you the right person for that?

captainsafia:
> This PR does not cover dotnet watch as my team doesn't own that code. @captainsafia are you the right person for that?

@pranavkm can help out with dotnet watch issues.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,57 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
                 verbosity.Equals(VerbosityOptions.d) ||
                 verbosity.Equals(VerbosityOptions.detailed);
         }
+
+        internal static IEnumerable<string> ResolveArchOptionToRuntimeIdentifier(string arg, ParseResult parseResult)
+        {
+            if (parseResult.BothArchAndOsOptionsSpecified())
+            {
+                return Array.Empty<string>();
```
vitek-karas, line 124: I think it's worth adding a comment here that we return empty array since we rely on the `ResolveOsOptionToRuntimeIdentifier` to be called which will handle the case of both specified at the same time.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
vitek-karas: But is should, right? I know the SDK repo doesn't run tests on arm64, and unfortunately the installer doesn't do that either right now. But we're trying to enable the tests on installer (at least I would hope so, we can't ship a supported platform without ever running SDK tests on it...), once that happens, this will likely break.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
```diff
@@ -84,7 +88,46 @@ public void MsbuildInvocationIsCorrect(string[] args, string expectedAdditionalA
...
+                command.GetArgumentsToMSBuild()
+                    .Should()
+                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=os-{expectedArch}");
+            });
+        }
+
+        [WindowsOnlyFact]
```
vitek-karas, line 107: Can we support the test on all platforms? Specifically macOS as that is currently the most important platform where this has to work.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
```diff
@@ -84,7 +88,46 @@ public void MsbuildInvocationIsCorrect(string[] args, string expectedAdditionalA
...
+        {
+            CommandDirectoryContext.PerformActionWithBasePath(WorkingDirectory, () =>
+            {
+                var msbuildPath = "<msbuildpath>";
+                var command = BuildCommand.FromArgs(new string[] { "--arch", "arch", "--os", "os" }, msbuildPath);
+                command.GetArgumentsToMSBuild()
+                    .Should()
+                    .Be($"{ExpectedPrefix} -restore -consoleloggerparameters:Summary -property:RuntimeIdentifier=os-arch");
+            });
         }
```
vitek-karas, line 131: Other tests to add:
* Specify the same option twice (not sure if the command line parsing handles this implicitly) - `dotnet build -a x86 -a x64`
* Test on all the other commands - not just build - at least test that both options are accepted and that they work - can we make this into a matrix test instead?
* Test what happens if the new options are combined with `-r`
* Test what happens if the project specifies `RuntimeIdentifier` as a property in `.csproj` - which one will win?

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,57 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+            os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            var properties = new string[] { $"-property:RuntimeIdentifier={os}-{arch}" };
+            return properties;
+        }
+
+        private static string GetCurrentRuntimeId()
+        {
+            var dotnetRootPath = Path.GetDirectoryName(Environment.ProcessPath);
+            dotnetRootPath = dotnetRootPath.Contains("dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
```
dsplaisted, line 152: I'm not following what this line is doing.  Under what circumstances would you need to append "dotnet" to the path?

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,57 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+            var ridFileName = "NETCoreSdkRuntimeIdentifierChain.txt";
+            string runtimeIdentifierChainPath = string.IsNullOrEmpty(Product.Version) ?
```
dsplaisted, line 154: When would Product.Version not be defined here?

src/Cli/dotnet/commands/dotnet-run/RunCommandParser.cs
```diff
@@ -47,6 +47,8 @@ public static Command GetCommand()
             command.AddOption(InteractiveOption);
             command.AddOption(NoRestoreOption);
             command.AddOption(CommonOptions.VerbosityOption());
+            command.AddOption(CommonOptions.ArchitectureOption());
+            command.AddOption(CommonOptions.OperatingSystemOption());
```
dsplaisted, line 51: Does this work correctly for `dotnet run`?  The output path is going to be different if you specify the OS or architecture, does the run command take that into account?

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -138,6 +138,8 @@ public static Command GetCommand()
             command.AddOption(NoRestoreOption);
             command.AddOption(CommonOptions.InteractiveMsBuildForwardOption());
             command.AddOption(CommonOptions.VerbosityOption());
+            command.AddOption(CommonOptions.ArchitectureOption(false));
+            command.AddOption(CommonOptions.OperatingSystemOption());
```
dsplaisted, line 142: Similar to `dotnet run`, we should make sure that `dotnet test` works correctly with these options.  The test should run with the specified architecture.  I'm not sure what should happen if you try to specify a different OS for `dotnet test`, it doesn't support remotely running the tests on a different OS.

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
```diff
@@ -55,6 +55,8 @@ public static Command GetCommand()
             command.AddOption(ToolCommandRestorePassThroughOptions.NoCacheOption);
             command.AddOption(ToolCommandRestorePassThroughOptions.InteractiveRestoreOption);
             command.AddOption(VerbosityOption);
+            command.AddOption(CommonOptions.ArchitectureOption());
+            command.AddOption(CommonOptions.OperatingSystemOption());
```
dsplaisted, line 59: Tool install is probably a lot more complex and we may want to leave it for a separate PR.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,57 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+        private static IEnumerable<string> ResolveRidShorthandOptions(string os, string arch)
+        {
+            var currentRid = GetCurrentRuntimeId();
+            os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            var properties = new string[] { $"-property:RuntimeIdentifier={os}-{arch}" };
+            return properties;
+        }
+
+        private static string GetCurrentRuntimeId()
```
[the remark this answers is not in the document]
dsplaisted: Per the [spec](https://github.com/dotnet/designs/pull/233/files#diff-c3c573f67f5f781df30107752d1fad11567a0daed7e3008b6468fb9cdf5410e1R230):

> The shorthand syntax is intended as a CLI and not MSBuild concept. For scenarios that today require specifying RIDs in project file, the intent is that users will continue to specify RIDs. If we find that there is a need for a shorthand RID syntax in project files, then we can consider extending this syntax to MSBuild.

We could pass "private" properties to MSBuild if we had to, but if we can I think we should avoid it, as we'd have to prevent them from flowing across project references as global properties the same way we do with RuntimeIdentifier.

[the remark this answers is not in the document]
vitek-karas: Alternatively - if we have existing C# code which does this today, we should share that between MSBuild and CLI. If we don't, maybe we should consider doing so. Otherwise this will be a prime target for unintentional breaks and inconsistencies.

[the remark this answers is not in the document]
dsplaisted: For MSBuild, it's a property (`NETCoreSdkRuntimeIdentifier`) which we [generate into](https://github.com/dotnet/installer/blob/c7b0e76ac551a908c14797c3b8af809e285f0cc4/src/redist/targets/GenerateBundledVersions.targets#L280) the Microsoft.NETCoreSdk.BundledVersions.props file when we're building the SDK.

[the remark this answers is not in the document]
vitek-karas: That sounds like something we could do for CLI as well - but it would require us to build the CLI as RID specific, which is not the case today.
Alternative idea - can we instead rely on the `dotnet.dll` location? It would be easier/safer than trying to compute the path. Options:
* Simply look "next to dotnet.dll" - not as robust as the other options, but definitely much better than the code here
* Through a post-build step bake the RID into the binary. This is technically tricky and could have problems with signing - but it's an option.
* Build CLI as RID specific - the best approach I think (we plan to do that eventually anyway for size reasons)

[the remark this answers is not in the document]
dsplaisted: We already also use the `NETCoreSdkRuntimeIdentifierChain.txt` file for the [workload MSBuild SDK resolver](https://github.com/dotnet/sdk/blob/main/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs#L29).  So I think it's fine to rely on that file.  I think looking next to dotnet.dll is a great idea to simplify this code, in what ways do you think it wouldn't be robust?

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,57 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+
+            return ResolveRidShorthandOptions(arg, null);
+        }
+
+        private static IEnumerable<string> ResolveRidShorthandOptions(string os, string arch)
+        {
+            var currentRid = GetCurrentRuntimeId();
+            os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            var properties = new string[] { $"-property:RuntimeIdentifier={os}-{arch}" };
```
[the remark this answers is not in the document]
richlander: From the spec:

> Specifying both a RID and the short-hand syntax would be an error, for example with the following commands:
> - `dotnet build -r win-x64 -a arm64`
> - `dotnet build -r linux-arm64 --os win -a x64`

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -91,6 +95,18 @@ public static Argument DefaultToCurrentDirectory(this Argument arg)
                 "--interactive",
                 CommonLocalizableStrings.CommandInteractiveOptionDescription);
 
+        public static Option ArchitectureOption(bool includeShortVersion = true) =>
+            new ForwardedOption<string>(
+                includeShortVersion ? new string[] { "--arch", "-a" } : new string[] { "--arch" },
+                CommonLocalizableStrings.ArchitectureOptionDescription)
```
Nirmal4G, line 101: How about a proper long and short form for the Architecture option?

```suggestion
            new ForwardedOption<string>(
                includeShortVersion ? new string[] { "--architecture", "-a" } : new string[] { "--architecture" },
                CommonLocalizableStrings.ArchitectureOptionDescription)
```

##### NOTES

- If there's a way to add `--arch` as an alias of the long form `--architecture`, that would make this more friendly and intuitive!!!
- See examples below…

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -91,6 +95,18 @@ public static Argument DefaultToCurrentDirectory(this Argument arg)
...
+            .SetForwardingFunction(ResolveArchOptionToRuntimeIdentifier);
+
+        public static Option OperatingSystemOption() =>
+            new ForwardedOption<string>(
+                "--os",
+                CommonLocalizableStrings.OperatingSystemOptionDescription)
```
Nirmal4G, line 107: How about a proper long and short form for the Operating System option?

```suggestion
            new ForwardedOption<string>(
                includeShortVersion ? new string[] { "--operating-system", "-s" } : new string[] { "--operating-system" },
                CommonLocalizableStrings.OperatingSystemOptionDescription)
```

#### Notes

- If there's a way to add `--os` as an alias of the long form `--operating-system`, that would make this more friendly and intuitive!!!
- We could also do the shorter `--system` instead of the longer `--operating-system` but either is okay!

---

We could also call it as a platform but that's asking too much, since, platform means different in the MSBuild world!

```suggestion
            new ForwardedOption<string>(
                includeShortVersion ? new string[] { "--platform", "-p" } : new string[] { "--platform" },
                CommonLocalizableStrings.PlatformOptionDescription)
```

---

#### Examples

| Short Form | Long Form |
| --- | --- |
| `dotnet build -s win -a x64` | `dotnet build --os Windows --arch x64` |
| `dotnet build -s win -a x64` | `dotnet build --operating-system Windows --architecture x64` |
| `dotnet build -s win -a x64` | `dotnet build --system Windows --architecture x64` |
| `dotnet build -p win -a x64` | `dotnet build --platform Windows --architecture x64` |

Nirmal4G (commented):
Some suggestions.

I know that the whole point of having short forms is to write in the CLI faster but the long form should be more beginner friendly and should have a proper word along with its shorter alias. The **very long form** I suggested could be useful in scripts which could make them more readable and better understandable.

src/Cli/dotnet/CommonOptions.cs
Nirmal4G: I just read "AndOs" as "Andos" and `ResolveOs` as `Resolves`. When skimming through these things happen. It would be better to expand `OS` to `OperatingSystem` everywhere in the sources especially in the variable names. It might be long but will be readable and more legible.

src/Cli/dotnet/CommonOptions.cs
sfoslund: We've decided to stick with the short versions per the spec.

src/Cli/dotnet/CommonOptions.cs
sfoslund: In testing. Adding this line allows us to pick up the rid file when running unit tests, without it we fail and can't test.

src/Cli/dotnet/CommonOptions.cs
sfoslund: Same as above, in testing.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,57 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+                Path.Combine(Directory.GetDirectories(Path.Combine(dotnetRootPath, "sdk"))[0], ridFileName) :
+                Path.Combine(dotnetRootPath, "sdk", Product.Version, ridFileName);
+            string[] currentRuntimeIdentifiers = File.Exists(runtimeIdentifierChainPath) ?
+                File.ReadAllLines(runtimeIdentifierChainPath).Where(l => !string.IsNullOrEmpty(l)).ToArray() :
+                new string[] { };
+            if (currentRuntimeIdentifiers == null || !currentRuntimeIdentifiers.Any() || !currentRuntimeIdentifiers[0].Contains("-"))
+            {
+                throw new GracefulException(CommonLocalizableStrings.CannotResolveRuntimeIdentifier);
+            }
+            return currentRuntimeIdentifiers[0]; // First rid is the most specific (ex win-x64)
```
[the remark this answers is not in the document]
sfoslund: That logic is to account for the slightly different configuration when we're testing, this is how we're able to pick up the RID file in tests.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
sfoslund: The first item is covered with command line parsing implicitly and the last item is implicitly handled by MSBuild (since command line args are global properties). But I'll add tests for the other two, thanks.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
sfoslund: Updated in latest commit

sfoslund:
@dsplaisted @vitek-karas I've updated based on your feedback, let me know if you have any further suggestions. @richlander I'm leaving dotnet tool install out of this PR for now as that command will be more complex. I'll make a separate PR once this is merged.

src/Cli/dotnet/CommonOptions.cs
vitek-karas: Why - do we need to modify the tests to better mimic the actual runtime environment?

src/Cli/dotnet/CommonOptions.cs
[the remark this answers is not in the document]
vitek-karas: Does it mean that we're not testing how the product will work when installed for real? That would be unfortunate... and would be really good to fix.
What is the layout on disk the tests use? And why it can't be changed to look like if the product is correctly installed.

src/Cli/dotnet/CommonOptions.cs
sfoslund: Environment.ProcessPath gives us the location of dotnet.exe when we're really running commands, but in tests it gives us testhost.exe (which is on dir above the location of dotnet.exe). The same difference is what causes Product.Version to be an empty string when testing.

vitek-karas: Good point - can we make it more specific then?
Basically switch the code around instead of looking for "dotnet" look for "testhost" and then do "the other thing". It would make the code cleaner in my opinion. And I would still add a comment describing how it looks under test (basically what you wrote above in your reply here).

sfoslund: Sure, just pushed a commit with that change.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,81 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            return $"{os}-{arch}";
+        }
+
+        private static string GetCurrentRuntimeId()
+        {
+            string runtimeIdentifierChainPath;
+            var ridFileName = "NETCoreSdkRuntimeIdentifierChain.txt";
+            if (Path.GetFileName(Environment.ProcessPath).Equals("dotnet.exe"))
+            {
```
vitek-karas, line 170: * This will only work on Windows
* Why not do it the other way round - single out the testing scenario - I understand why we want test-only behavior in the product, and I think it's fine to have that. But I think the code should look something like:
```C#
if (Path.GetFileNameWithoutExtension(Environment.ProcessPath).Equals("testhost")
{
    // Do test specific discovery
}
else 
{
   // Do normal product code path
}
```

I think it's better to be very specific about supporting tests. One other way (even cleaner) would be to have for example an environment variable which has to be set in order to activate the test only behavior.

The end goal being that it should be basically impossible to get the test behavior by accident.

sfoslund: Investigating the CI failures I realized that we only see the test host being run when we're running with VS, in CI tests are run with dotnet.exe but we still have the same issues, so I don't think it'll be possible to completely separate out this logic.

vitek-karas: Yeah - I was looking at it as well.. The test stuff is somewhat of a mess. I guess it will have to be "not nice" :-). Please add a comment why it's doing what it's doing and what is the expected environment - just so that a year from now this doesn't look like dark magic.

And thanks for bearing with me 👼

sfoslund: Just pushed the change

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,74 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+            os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            return $"{os}-{arch}";
+        }
+
+        private static string GetCurrentRuntimeId()
+        {
+            var dotnetRootPath = Path.GetDirectoryName(Environment.ProcessPath);
+            // When running under test the path does not always contain "dotnet" and Product.Version is empty.
+            dotnetRootPath = dotnetRootPath.Contains("dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
```
vitek-karas, line 169: I still think that doing `path.Contains("dotnet")` is not the best. If I close the SDK repo to `F:\dotnet\sdk` (which is actually where I have it on my machine), that condition will be true for everything from the repo - including the `testhost` which comes from the repo-local SDK. But that doesn't mean that the path will point to `dotnet.exe`.

sfoslund: Updated to check the file name in the latest commit.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -100,6 +116,74 @@ public static bool VerbosityIsDetailedOrDiagnostic(this VerbosityOptions verbosi
...
+            os = string.IsNullOrEmpty(os) ? GetOsFromRid(currentRid) : os;
+            arch = string.IsNullOrEmpty(arch) ? GetArchFromRid(currentRid) : arch;
+            return $"{os}-{arch}";
+        }
+
+        private static string GetCurrentRuntimeId()
+        {
+            var dotnetRootPath = Path.GetDirectoryName(Environment.ProcessPath);
+            // When running under test the path does not always contain "dotnet" and Product.Version is empty.
+            dotnetRootPath = Path.GetFileName(dotnetRootPath).Contains("dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
```
vitek-karas, line 169:
```suggestion
            dotnetRootPath = (Path.GetFileNameWithoutExtension(dotnetRootPath) == "dotnet") ? dotnetRootPath : Path.Combine(dotnetRootPath, "dotnet");
```

vitek-karas (approved):

sfoslund:
@dsplaisted we'd like to get this in for preview 7, can I get a review so we can merge?

src/Cli/dotnet/commands/dotnet-run/RunCommandParser.cs
dsplaisted, line 51: Does does the fact that these options have a forwarding function interfere here, since the runtime is passed to the RunCommand as a parameter instead of as additional MSBuild arguments like the other commands?

dsplaisted (approved):
I think we should also have some coverage of the new options that work by actually starting a `dotnet` process, since the code to get the current RuntimeIdentifier works differently when running as an in-process unit test.

src/Cli/dotnet/commands/dotnet-run/RunCommandParser.cs
sfoslund: We account for these options here: https://github.com/dotnet/sdk/pull/18889/files#diff-c59ee7ae48403cce44671ddc3549eb921e7c23bda26840c883b29d9f7e7942e5R38, the forwarding function shouldn't interfere.

wli3:
> I could have been more clear/prescriptive on this point. This text (from the spec) is the closest I got to on that.
> 
> > Tool installation with `dotnet tool install` is a particularly problematic scenario. At the time .NET 6 is released, one can expect that close to 100% of .NET tools (packaged via NuGet) will target a version prior to .NET 6 and therefore be incapable on running on Arm64 (at least on macOS). However, when installed, the SDK will generate an apphost that matches the SDK RID. As a result, the tool will consistently fail to launch (since the required runtime will be missing). The SDK needs to apply the same implicit RID rules for .NET tools that it will apply for `dotnet build`. In addition, `dotnet tool install` should enable explicit RID targeting.
> 
> > Note: One can think of .NET tools as being distributed as portable apps with no apphost, with the most favorable apphost being provided just-in-time at installation. It's a pretty good model and aligns with the spirit of the overall design.
> 
> The same behavior is expected for `dotnet test` and `dotnet run` (and `dotnet watch`, which is a conceptual wrapper over those too).

@richlander. I suggest we take a pause for now and discuss further in the meeting.

> that close to 100% of .NET tools (packaged via NuGet) will target a version prior to .NET 6 and therefore be incapable on running on Arm64

That is always the case, .NET5 tools cannot run on .NET6, .NET3.1 tools cannot run on .NET5. There is nothing special on arm64. Even if we generate the right rid apphost, it will still fail since runtime does not move forward.

And if we generate different apphost on the PATH, will the shell resolve both architecture executable for the user?

And the engineering work is not trivial at all for the work. RID select involves reading the read graph which global tools does not do. We also need to consider prepackage apphost in addition at SDK provided non-signed apphost.

wli3:
> Modulo my feedback, this looks good. Thanks!
> 
> I'm still left wondering how specifying a RID will behave for new scenarios, like `dotnet tool install`. Will that be a subsequent change?
> 
> I'm more than happy to test these changes to validate that they play out correctly in an actual build. If you share a build with me, I'll test it (in a timely manner).

And what is the benefit of allowing user to specific a rid? Tools are framework-dependent, and once it is on user's machine, all we need to do is to find a convenient way to run it. Unlike website-docker issue, SDK control both end of the deployment.

richlander:
Let me turn this around to you. How would this scenario work?

- Install .NET Tool (built with .NET Core 3.1) with the .NET 6 Arm64 SDK.
- Run tool.

What do you expect to happen?

wli3:
> Let me turn this around to you. How would this scenario work?
> 
> * Install .NET Tool (built with .NET Core 3.1) with the .NET 6 Arm64 SDK.
> * Run tool.
> 
> What do you expect to happen?

Regardless arm64 or not, it will not work. Basically installing  .NET Tool (built with .NET Core 3.1) with the .NET **5** SDK. will not work **today**.

wli3:
And I talked you about making runtime move forward before, but runtime team decided not to.

Even that worked. I prefer runtime to be able to resolve different runtime architecture from any apphost to any architecture. Tools are just a symptom of not having our framework dependent scenario work for arm64 vs x64 yet

richlander:
> Regardless arm64 or not, it will not work. Basically installing .NET Tool (built with .NET Core 3.1) with the .NET 5 SDK. will not work today.

It works fine. Explain why it doesn't.

vitek-karas:
We can't make arm64 apphost run x64 runtime - it's just not possible. On macOS there are some dirty tricks possible which might allow this, but I'm now aware of anything similar on Windows (and I don't think anything like that is planned).
It will work if you run the app via `dotnet app.dll` (and you pick the right `dotnet`, but that is another problem on its own).

I think the basic ask here is that `dotnet tool install` will know about the fact that .NET 3.1 doesn't have arm64 runtime and default to x64 apphost for those tools (assuming 3.1 runtime is installed).

richlander:
Exactly.

wli3:
Talked to @richlander offline. The immediate issue need to address is that:

Have 6.0 arm installed on the machine. install older tool targeting 3.1. The apphost will be arm based. When the user run the app, the app will tell the user to download 3.1 intel runtime. But due to the apphost, even intel runtime is installed, the user still cannot run the app. And runtime team has not plan to create such apphost can run both arm and non arm application.

Scoped the experience to make the development cheaper:
only allow -architecture, so that the valid user input can only be win-x64, win-arm64, osx-x64, osx-arm (OS is the current machine OS). Then we don't need rid graph. We can derive the apphost package id and version by rid and version (inserted to SDK), and download using SDK nuget API directly.

Technical note:

There is an issue for resolving "dotnet [space] blah", it would require current SDK to run first to resolve the executable and then process.start a potential different architecture binary. This likely not works. We might ask the user to run dotnet-format instead.

Figure out what the tool is targeting (if < 6.0) is another unknown. Maybe SDK could get it from asset.json by nuget restore. The worst case scenario SDK could parse and read from runtime.config.json

Be aware of the packaged apphost code path. Although first party should retarget or auto roll forward to avoid this situation.

vitek-karas:
Technically there is a way to ask the tool what framework version it needs/wants - sort of. Once we extract it on disk (before the executable creation), we could use native hosting APIs to ask `hostfxr` to load the app and try to resolve frameworks (given some runtime location, probably the SDK-native location). This would either work (in which case it's probably OK to create executable with the same arch as the SDK) or it would fail, in which case the SDK could either guess that it's downlevel app, or somehow read it from the error reported from `hostxr`.

But I understand that this would be a lot of work to get right, so probably something to consider for 7 (in which case we could add a new API into `hostfxr` for this for example).

richlander:
Is this something we could consider now? I know that `hostfxr` is a risky component to change. That said, we're still working on this scenario. I wonder if there is an opportunity to "do it right" from the beginnining.

vitek-karas:
I think we can get the desired behavior without `hostfxr` change today (Since we really only have .NET 6 and newer or .NET 5 and older split - so a binary "it worked or not" is enough to decide between the two). The new API would make the code robust to future changes, but that can be done when we actually need it.

Adding a new API to `hostfxr` is risky and I think we're pass the point in .NET 6 where I would feel comfortable doing it - especially since we don't have a strong reason to do so.

/cc @agocke for context

agocke:
Agreed, I wouldn't consider a hostfxr change at this point. .NET 7, sure, but for now parsing JSON is the best approach.

PR #18910, opened 2021-07-14
Add workload restore command

wli3:
Adding _FindAllReferenceWorkload target to find workloads in project.
Discover the projects and then pipe all arguements to workload install
command

### DIFF
--- a/src/Cli/dotnet/Parser.cs
+++ b/src/Cli/dotnet/Parser.cs
@@ -72,7 +72,7 @@ namespace Microsoft.DotNet.Cli
         // Argument
         public static readonly Argument<string> DotnetSubCommand = new Argument<string>() { Arity = ArgumentArity.ExactlyOne, IsHidden = true };
 
-        private static Command ConfigureCommandLine(Command rootCommand, bool includeWorkloadCommands = false)
+        private static Command ConfigureCommandLine(Command rootCommand)
         {
             // Add subcommands
             foreach (var subcommand in Subcommands)
@@ -80,11 +80,10 @@ namespace Microsoft.DotNet.Cli
                 rootCommand.AddCommand(subcommand);
             }
 
-            // Workload command is behind a feature flag during development
-            rootCommand.AddCommand(WorkloadCommandParser.GetCommand(includeWorkloadCommands || Env.GetEnvironmentVariableAsBool("DEVENABLEWORKLOADCOMMAND", defaultValue: false)));
+            rootCommand.AddCommand(WorkloadCommandParser.GetCommand());
 
             //Add internal commands
-           rootCommand.AddCommand(InstallSuccessCommand);
+            rootCommand.AddCommand(InstallSuccessCommand);
 
             // Add options
             rootCommand.AddOption(DiagOption);
@@ -106,16 +105,6 @@ namespace Microsoft.DotNet.Cli
         }
 
         public static System.CommandLine.Parsing.Parser Instance { get; } = new CommandLineBuilder(ConfigureCommandLine(RootCommand))
-            .UseExceptionHandler(ExceptionHandler)
-            .UseHelp()
-            .UseHelpBuilder(context => new DotnetHelpBuilder(context.Console))
-            .UseResources(new CommandLineValidationMessages())
-            .UseParseDirective()
-            .UseSuggestDirective()
-            .DisablePosixBinding()
-            .Build();
-
-        public static System.CommandLine.Parsing.Parser GetWorkloadsInstance { get; } = new CommandLineBuilder(ConfigureCommandLine(new RootCommand(), true))
             .UseExceptionHandler(ExceptionHandler)
             .UseHelp()
             .UseHelpBuilder(context => new DotnetHelpBuilder(context.Console))
--- a/src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
@@ -2,21 +2,28 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.CommandLine.Parsing;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Construction;
+using Microsoft.Build.Execution;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Logging;
 using Microsoft.DotNet.Cli;
 using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Workloads.Workload.Install;
 using Microsoft.Extensions.EnvironmentAbstractions;
+using Microsoft.NET.Sdk.WorkloadManifestReader;
 
 namespace Microsoft.DotNet.Workloads.Workload.Restore
 {
     internal class WorkloadRestoreCommand : CommandBase
     {
-        private readonly string _configFilePath;
-        private readonly IReporter _errorReporter;
-        private readonly IFileSystem _fileSystem;
+        private readonly ParseResult _result;
         private readonly IReporter _reporter;
-        private readonly string[] _sources;
-        private readonly string _verbosity;
+        private readonly IEnumerable<string> _slnOrProjectArgument;
 
         public WorkloadRestoreCommand(
             ParseResult result,
@@ -24,21 +31,102 @@ namespace Microsoft.DotNet.Workloads.Workload.Restore
             IReporter reporter = null)
             : base(result)
         {
-            _fileSystem = fileSystem ?? new FileSystemWrapper();
-
+            _result = result;
             _reporter = reporter ?? Reporter.Output;
-            _errorReporter = reporter ?? Reporter.Error;
-
-            _configFilePath = result.ValueForOption<string>(WorkloadRestoreCommandParser.ConfigOption);
-            _sources = result.ValueForOption<string[]>(WorkloadRestoreCommandParser.SourceOption);
-            _verbosity =
-                Enum.GetName(result.ValueForOption<VerbosityOptions>(WorkloadRestoreCommandParser.VerbosityOption));
+            _slnOrProjectArgument =
+                result.ValueForArgument<IEnumerable<string>>(RestoreCommandParser.SlnOrProjectArgument);
         }
 
         public override int Execute()
         {
-            _reporter.WriteLine("WIP workload restore stub");
+            var allProjects = DiscoverAllProjects(Directory.GetCurrentDirectory(), _slnOrProjectArgument).Distinct();
+            List<WorkloadId> allWorkloadId = RunTargetToGetWorkloadIds(allProjects);
+            _reporter.WriteLine(string.Format(LocalizableStrings.InstallingWorkloads, string.Join(" ", allWorkloadId)));
+
+            var workloadInstallCommand = new WorkloadInstallCommand(_result,
+                workloadIds: allWorkloadId.Select(a => a.ToString()).ToList().AsReadOnly());
+
+            workloadInstallCommand.Execute();
             return 0;
+        }
+
+        private List<WorkloadId> RunTargetToGetWorkloadIds(IEnumerable<string> allProjects)
+        {
+            var globalProperties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                {"SkipResolvePackageAssets", "true"}
+            };
+
+            var allWorkloadId = new List<WorkloadId>();
+            foreach (string projectFile in allProjects)
+            {
+                var project = new ProjectInstance(projectFile, globalProperties, null);
+
+                bool buildResult = project.Build(new[] {"_GetRequiredWorkloads"},
+                    loggers: new ILogger[]
+                    {
+                        new ConsoleLogger(_result
+                            .ValueForOption(WorkloadInstallCommandParser.VerbosityOption)
+                            .ToLoggerVerbosity())
+                    },
+                    remoteLoggers: Enumerable.Empty<ForwardingLoggerRecord>(),
+                    targetOutputs: out var targetOutputs);
+
+                if (buildResult == false)
+                {
+                    throw new GracefulException(
+                        string.Format(
+                            LocalizableStrings.FailedToRunTarget,
+                            projectFile));
+                }
+
+                var targetResult = targetOutputs["_GetRequiredWorkloads"];
+                allWorkloadId.AddRange(targetResult.Items.Select(item => new WorkloadId(item.ItemSpec)));
+            }
+
+            allWorkloadId = allWorkloadId.Distinct().ToList();
+            return allWorkloadId;
+        }
+
+
+        internal static List<string> DiscoverAllProjects(string currentDirectory,
+            IEnumerable<string> slnOrProjectArgument = null)
+        {
+            var slnFiles = new List<string>();
+            var projectFiles = new List<string>();
+            if (slnOrProjectArgument == null || !slnOrProjectArgument.Any())
+            {
+                slnFiles = Directory.GetFiles(currentDirectory, "*.sln").ToList();
+                projectFiles.AddRange(Directory.GetFiles(currentDirectory, "*.*proj"));
+            }
+            else
+            {
+                slnFiles = slnOrProjectArgument
+                    .Where(s => Path.GetExtension(s).Equals(".sln", StringComparison.OrdinalIgnoreCase))
+                    .Select(Path.GetFullPath).ToList();
+                projectFiles = slnOrProjectArgument
+                    .Where(s => Path.GetExtension(s).EndsWith("proj", StringComparison.OrdinalIgnoreCase))
+                    .Select(Path.GetFullPath).ToList();
+            }
+
+            foreach (string file in slnFiles)
+            {
+                var solutionFile = SolutionFile.Parse(file);
+                var projects = solutionFile.ProjectsInOrder;
+                foreach (var p in projects)
+                {
+                    projectFiles.Add(p.AbsolutePath);
+                }
+            }
+
+            if (projectFiles.Count == 0)
+            {
+                throw new GracefulException(
+                    LocalizableStrings.CouldNotFindAProject,
+                    currentDirectory, "--project");
+            }
+
+            return projectFiles;
         }
     }
 }
--- a/src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommandParser.cs
@@ -8,21 +8,12 @@ namespace Microsoft.DotNet.Cli
 {
     internal static class WorkloadRestoreCommandParser
     {
-        public static readonly Option<string> ConfigOption = WorkloadInstallCommandParser.ConfigOption;
-
-        public static readonly Option<string[]> SourceOption = WorkloadInstallCommandParser.SourceOption;
-
-        public static readonly Option<VerbosityOptions> VerbosityOption = WorkloadInstallCommandParser.VerbosityOption;
-
         public static Command GetCommand()
         {
             Command command = new Command("restore", LocalizableStrings.CommandDescription);
 
-            command.AddOption(ConfigOption);
-            command.AddOption(SourceOption);
-            command.AddWorkloadCommandNuGetRestoreActionConfigOptions();
-            command.AddOption(VerbosityOption);
-
+            command.AddArgument(RestoreCommandParser.SlnOrProjectArgument);
+            WorkloadInstallCommandParser.AddWorkloadInstallCommandOptions(command);
             return command;
         }
     }
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.GetRequiredWorkloads.CrossTargeting.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.GetRequiredWorkloads.CrossTargeting.targets
@@ -1,0 +1,33 @@
+<!--
+***********************************************************************************************
+Microsoft.NET.Sdk.Workloads.CrossTargeting.targets
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
+<Project>
+  <!--
+============================================================
+                                     _GetRequiredWorkloads
+
+ Collecting all Workload requirement for each inner build.
+============================================================
+-->
+
+  <Target Name="_GetRequiredWorkloads"
+          DependsOnTargets="_ComputeTargetFrameworkItems"
+          Returns="@(_ResolvedSuggestedWorkload)">
+
+    <MSBuild Projects="@(_InnerBuildProjects)"
+             Condition="'@(_InnerBuildProjects)' != '' "
+             Targets="_GetRequiredWorkloads"
+             BuildInParallel="$(BuildInParallel)">
+      <Output ItemName="_ResolvedSuggestedWorkload" TaskParameter="TargetOutputs"/>
+
+    </MSBuild>
+  </Target>
+</Project>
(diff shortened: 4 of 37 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
```diff
@@ -35,4 +35,32 @@ Copyright (c) .NET Foundation. All rights reserved.
...
                           GenerateErrorsForMissingWorkloads="true"/>
   </Target>
+
+  <Target Name="_SetSkipResolvePackageAssets" >
+    <PropertyGroup>
+      <SkipResolvePackageAssets>true</SkipResolvePackageAssets>
+    </PropertyGroup>
+  </Target>
+  
+  <Target Name="_FindAllReferenceWorkload" DependsOnTargets="_SetSkipResolvePackageAssets;GetSuggestedWorkloads;PrepareProjectReferences" Returns="@(_ResolvedSuggestedWorkloads)">
```
wli3, line 45: @dsplaisted help me double check the RemoveProperties are right

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloadsCrossTarget.targets
```diff
@@ -0,0 +1,37 @@
...
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+============================================================
+                                     _FindAllReferenceWorkload
+
+ Collecting all Workload requirement for each inner build.
+============================================================
+-->
+
+  <Target Name="_FindAllReferenceWorkload"
```
wli3, line 21: @dsplaisted I end up adding a outer build version instead using the same one

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
```diff
@@ -2,43 +2,119 @@
...
+            {
+                {Constants.MSBuildExtensionsPath, AppContext.BaseDirectory}
+            };
+
+            var allWorkloadId = new List<WorkloadId>();
+            foreach (string projectFile in allProjects)
+            {
+                var project = new ProjectInstance(projectFile, globalProperties, null);
+
+                project.Build(new[] {"_FindAllReferenceWorkload"},
```
wli3, line 67: @rainersigwald please help me double check that's all I need for using msbuild API calling a target

rainersigwald: Can you check for success?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
```diff
@@ -35,4 +35,32 @@ Copyright (c) .NET Foundation. All rights reserved.
                           NETCoreSdkVersion="$(NETCoreSdkVersion)"
                           GenerateErrorsForMissingWorkloads="true"/>
   </Target>
+
+  <Target Name="_SetSkipResolvePackageAssets" >
+    <PropertyGroup>
+      <SkipResolvePackageAssets>true</SkipResolvePackageAssets>
+    </PropertyGroup>
+  </Target>
```
rainersigwald, line 43: Why have this mutate state within the build rather than pass it in?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
```diff
@@ -35,4 +35,32 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  
+  <Target Name="_FindAllReferenceWorkload" DependsOnTargets="_SetSkipResolvePackageAssets;GetSuggestedWorkloads;PrepareProjectReferences" Returns="@(_ResolvedSuggestedWorkloads)">
+    <MSBuild
+    Projects="@(_MSBuildProjectReferenceExistent)"
+    Targets="GetSuggestedWorkloads"
+    BuildInParallel="$(BuildInParallel)"
+    Properties="%(_MSBuildProjectReferenceExistent.SetConfiguration); %(_MSBuildProjectReferenceExistent.SetPlatform); GenerateErrorsForMissingWorkloads=false; SkipResolvePackageAssets=true"
```
rainersigwald, line 50: Likewise here, are these properties avoidable?

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
rainersigwald: You could probably improve perf here by doing a `BeginBuild`/[PendBuildRequest](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.execution.buildmanager.pendbuildrequest?view=msbuild-16-netcore)/`EndBuild` flow so MSBuild can deduplicate some work and do a bunch of stuff concurrently. Not critical for this operation IMO.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
wli3: just set global state? If so the target will not be able to function by itself and there is no good way to tell you need to set SkipResolvePackageAssets.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloadsCrossTarget.targets
```diff
@@ -0,0 +1,37 @@
...
+          DependsOnTargets="_ComputeTargetFrameworkItems"
+          Returns="@(_ResolvedSuggestedWorkloads)">
+
+    <MSBuild Projects="@(_InnerBuildProjects)"
+             Condition="'@(_InnerBuildProjects)' != '' "
+             Targets="_FindAllReferenceWorkload"
+             BuildInParallel="$(BuildInParallel)">
+      <Output ItemName="_ResolvedSuggestedWorkloadsWithDuplicate" TaskParameter="TargetOutputs"/>
```
wli3, line 29: @rainersigwald _ResolvedSuggestedWorkloadsWithDuplicate should collect all output of batches right?

rainersigwald: Correct.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
rainersigwald: I'm mostly worried about calling this target somehow during a real build and it being a nightmare to debug. It is "private" though . . .

src/Tests/dotnet-workload-restore.Tests/DiscoverAllProjectsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                    });
+
+            // current directory is ignored
+            result.Should().Contain(f => Path.GetFileName(f) == "App.csproj", "from checking the sln file");
+            result.Should().Contain(f => Path.GetFileName(f) == "Lib.csproj", "from directly pass in");
+        }
+
+        [Fact]
+        public void WhenCallWithDirectoryWith2ProjectItShouldFindAll()
+        {
```
sfoslund, line 60: I don't see any tests confirming we can get from the project/sln arg in the parse result to a resolved list of projects

wli3: WhenCallWithSlnOrProjectArgumentItCollectProjectsFromSolution is the pass in example

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommandParser.cs
```diff
@@ -8,21 +8,12 @@ namespace Microsoft.DotNet.Cli
...
         {
             Command command = new Command("restore", LocalizableStrings.CommandDescription);
 
-            command.AddOption(ConfigOption);
-            command.AddOption(SourceOption);
-            command.AddWorkloadCommandNuGetRestoreActionConfigOptions();
-            command.AddOption(VerbosityOption);
-
+            command.AddArgument(RestoreCommandParser.SlnOrProjectArgument);
+            WorkloadInstallCommandParser.AddWorkloadInstallCommandOptions(command);
```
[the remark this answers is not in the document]
wli3: I checked the help text. They still make sense

src/Tests/dotnet-workload-restore.Tests/DiscoverAllProjectsTests.cs
sfoslund: But it's not covering the parsing/ reading the parse result value

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
wli3: Set globalproperty instead

src/Cli/dotnet/Parser.cs
```diff
@@ -112,7 +111,7 @@ private static CommandLineBuilder DisablePosixBinding(this CommandLineBuilder bu
             .DisablePosixBinding()
             .Build();
 
-        public static System.CommandLine.Parsing.Parser GetWorkloadsInstance { get; } = new CommandLineBuilder(ConfigureCommandLine(new RootCommand(), true))
+        public static System.CommandLine.Parsing.Parser GetWorkloadsInstance { get; } = new CommandLineBuilder(ConfigureCommandLine(new RootCommand()))
```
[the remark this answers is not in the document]
wli3: Replaced

sfoslund (approved):

wli3:
@dsplaisted Rainer checked msbuild part too, so I think it won't be too bad. I'll get it in once test passes.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
```diff
@@ -35,4 +35,27 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   </Target>
+
+  <!--  This target is not part of the build. Only used by dotnet workload restore command. Global property "SkipResolvePackageAssets"
+        need to be set to "true" to avoid requiring restore.-->
+  <Target Name="_FindAllReferenceWorkload" DependsOnTargets="GetSuggestedWorkloads;PrepareProjectReferences" Returns="@(_ResolvedSuggestedWorkloads)">
+    <MSBuild
+      Projects="@(_MSBuildProjectReferenceExistent)"
+      Targets="GetSuggestedWorkloads"
+      BuildInParallel="$(BuildInParallel)"
+      Properties="GenerateErrorsForMissingWorkloads=false; SkipResolvePackageAssets=true"
```
dsplaisted, line 46: I don't think you need to set either of these properties here.  `SkipResolvePackageAssets` should already be set as a global property, which means it should flow through to project references.  And `GenerateErrorsForMissingWorkloads` is a parameter to the `ShowMissingWorkloads` task, but setting it as a property doesn't do anything.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
```diff
@@ -35,4 +35,27 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      Condition="'%(_MSBuildProjectReferenceExistent.BuildReference)' == 'true' and '@(ProjectReferenceWithConfiguration)' != '' and '@(_MSBuildProjectReferenceExistent)' != ''"
```
dsplaisted, line 47: What's the purpose of the `'@(ProjectReferenceWithConfiguration)' != ''` clause?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
```diff
@@ -35,4 +35,27 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      RemoveProperties="%(_MSBuildProjectReferenceExistent.GlobalPropertiesToRemove)$(_GlobalPropertiesToRemoveFromProjectReferences)">
```
dsplaisted, line 48: I think we also need to remove `TargetFramework` here.  Otherwise, I think we'll end up doing the wrong thing in the following scenario:

Project A multitargets `net6.0-ios;net6.0-android`, and references Project B.
Project B targets `net6.0`.

If we don't remove the `TargetFramework`, then I think we'd end up passing `net6.0-ios` and `net6.0-android` as the `TargetFramework` to project B when the project A inner builds traverse the project references.  So we'd end up building the project with a TargetFramework that it doesn't target.

src/Tests/Microsoft.NET.Build.Tests/WorkloadTests.cs
```diff
@@ -237,5 +237,67 @@ public void It_should_import_aliased_pack()
...
+            getValuesCommand.DependsOnTargets = "_FindAllReferenceWorkload";
+            getValuesCommand.ShouldRestore = false;
+
+            getValuesCommand.Execute("/p:SkipResolvePackageAssets=true")
+                .Should()
+                .Pass();
+
+            getValuesCommand.GetValues()
+                .Should()
+                .BeEquivalentTo("microsoft-android-sdk-full", "microsoft-net-runtime-mono-tooling");
```
dsplaisted, line 263: I think these names are changing, you may want to update stage 0 and see if that includes the rename so that this test doesn't start failing when it happens.

src/Tests/Microsoft.NET.Build.Tests/WorkloadTests.cs
```diff
@@ -237,5 +237,67 @@ public void It_should_import_aliased_pack()
                 .Should()
                 .BeEquivalentTo("true");
         }
+
+        [Fact]
+        public void It_should_get_suggested_workload_by_FindAllReferenceWorkload_target()
+        {
+            var mainProject = new TestProject()
+            {
+                Name = "MainProject", TargetFrameworks = "net6.0-android", IsSdkProject = true, IsExe = true
```
dsplaisted, line 246: Any reason these properties are all on one line?

src/Tests/Microsoft.NET.Build.Tests/WorkloadTests.cs
```diff
@@ -237,5 +237,67 @@ public void It_should_import_aliased_pack()
...
+                .Should()
+                .Pass();
+
+            getValuesCommand.GetValues()
+                .Should()
+                .BeEquivalentTo("microsoft-android-sdk-full", "microsoft-net-runtime-mono-tooling");
+        }
+
+        [Fact]
+        public void Given_multi_target_It_should_get_suggested_workload_by_FindAllReferenceWorkload_target()
```
dsplaisted, line 267: This tests a multi-targeted project referencing a multi-targeted project, but they both require the same workloads.  I think we should cover more cases, such as:

- `net6.0` referencing `net6.0;net6.0-android;net6.0-ios`.  With the way we've implemented workload restore, this will require the ios and android workloads, even though they are not strictly required to build the first project.
- `net6.0;net6.0-ios` referencing `net6.0;net6.0-android`.  This is similar but has different workloads returned by each project
- `net6.0` referencing `net6.0`, where the referenced project has `RunAOTCompilation` set to true.  This validates when a referenced project requires a workload and there's no multi-targeting.

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommandParser.cs
```diff
@@ -8,21 +8,12 @@ namespace Microsoft.DotNet.Cli
...
         public static Command GetCommand()
         {
             Command command = new Command("restore", LocalizableStrings.CommandDescription);
 
-            command.AddOption(ConfigOption);
-            command.AddOption(SourceOption);
-            command.AddWorkloadCommandNuGetRestoreActionConfigOptions();
-            command.AddOption(VerbosityOption);
-
+            command.AddArgument(RestoreCommandParser.SlnOrProjectArgument);
```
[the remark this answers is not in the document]
wli3: I use an msbuild API to read sln to projects

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
```diff
@@ -2,43 +2,128 @@
...
+                var targetResult = targetOutputs["_FindAllReferenceWorkload"];
+                allWorkloadId.AddRange(targetResult.Items.Select(item => new WorkloadId(item.ItemSpec)));
+            }
+
+            allWorkloadId = allWorkloadId.Distinct().ToList();
+            return allWorkloadId;
+        }
+
+        internal static List<string> DiscoverAllProjects(string currentDirectory,
+            IEnumerable<string> slnOrProjectArgument = null)
```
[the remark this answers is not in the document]
wli3: I thought about, but why make it different. I started with refactoring RunCommand. But when there are 2 projects in the folder what the user should do? It is valid, and workload restore can handle it (unlike runcommand), why fail and ask the user to input one by one?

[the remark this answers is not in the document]
dsplaisted: All other MSBuild and dotnet commands require you to specify the project if there is more than one in the folder.  It seems weird to make `dotnet workload restore` the one exception.

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommandParser.cs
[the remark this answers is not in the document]
rainersigwald: Since this PR parses the sln it's not critical but

> I don't know off the top of my head what will happen if you call an arbitrary target in a solution.

Since https://github.com/dotnet/msbuild/pull/1497, MSBuild will synthesize a target that calls the arbitrary target on each project in the solution (but today we don't roll up the results into an observable target output).

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
```diff
@@ -2,43 +2,128 @@
...
+
+            workloadInstallCommand.Execute();
             return 0;
         }
+
+        private static List<WorkloadId> RunTargetToGetWorkloadIds(IEnumerable<string> allProjects)
+        {
+            var globalProperties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                {Constants.MSBuildExtensionsPath, AppContext.BaseDirectory},
```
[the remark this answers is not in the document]
wli3: Removed the line. It still works.

src/Cli/dotnet/commands/dotnet-workload/restore/WorkloadRestoreCommand.cs
```diff
@@ -2,43 +2,128 @@
...
+
+            var allWorkloadId = new List<WorkloadId>();
+            foreach (string projectFile in allProjects)
+            {
+                var project = new ProjectInstance(projectFile, globalProperties, null);
+
+                bool buildResult = project.Build(new[] {"_FindAllReferenceWorkload"},
+                    loggers: Enumerable.Empty<ILogger>(),
+                    remoteLoggers: Enumerable.Empty<ForwardingLoggerRecord>(),
+                    targetOutputs: out var targetOutputs);
```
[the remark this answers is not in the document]
wli3: Added console logger with quite as default

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
wli3: Good catch. Removed

src/Tests/Microsoft.NET.Build.Tests/WorkloadTests.cs
wli3: due to https://github.com/dotnet/sdk/issues/19021 I will update stage 0 in a different PR

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -208,7 +208,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </Target>
 
   <Target Name="ResolvePackageAssets"
-          Condition=" '$(DesignTimeBuild)' != 'true' Or Exists('$(ProjectAssetsFile)')"
+          Condition=" ('$(DesignTimeBuild)' != 'true' Or Exists('$(ProjectAssetsFile)' ) ) And '$(SkipResolvePackageAssets)' != 'true'"
```
Nirmal4G, line 211:
```suggestion
          Condition="('$(DesignTimeBuild)' != 'true' Or Exists('$(ProjectAssetsFile)')) And '$(SkipResolvePackageAssets)' != 'true'"
```

extra internal whitespace

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.GetRequiredWorkloads.CrossTargeting.targets
```diff
@@ -0,0 +1,33 @@
...
+Microsoft.NET.Sdk.GetRequiredWorkloads.CrossTargeting.targets
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
Nirmal4G, line 12:
```suggestion
<Project>
```

I think we don't need the xmlns as of MSBuild v15.0

Nirmal4G (commented):
Some suggestions for the targets!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.GetRequiredWorkloads.CrossTargeting.targets
```diff
@@ -0,0 +1,33 @@
+<!--
+***********************************************************************************************
+Microsoft.NET.Sdk.GetRequiredWorkloads.CrossTargeting.targets
```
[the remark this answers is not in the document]
wli3: see Daniel's comment above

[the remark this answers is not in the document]
Nirmal4G: How about simply `Workloads`?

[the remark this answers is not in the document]
Nirmal4G: How about simply `Workloads` instead of `GetRequiredWorkloads`? It's expected that we'll be adding logic in the future. We can't rename the file then, so might as well give a generic name, IMHO!

Also, the file `Microsoft.NET.Sdk.ImportWorkloadsCrossTarget.targets` still remains after rename, is it intentional?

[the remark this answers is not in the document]
dsplaisted: We can rename the file in the future if we want to, though we don't often do so.

I do like the suggested name of `Microsoft.NET.Sdk.Workloads.CrossTargeting.targets`.

wli3:
@dsplaisted all issue addressed

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.ImportWorkloads.targets
wli3: Removed

wli3:
@dsplaisted I'll merge once tests pass

PR #18921, opened 2021-07-14
Rationalize default Imports for C# and VB

dsplaisted:
This updates the list of `Import` items for VB to include the Imports that C# by default gets for .NET 6 and higher.  This means that VB will now additionally get the following when targeting .NET 6:

- System.IO
- System.Net.Http
- System.Threading

This also reorganizes the imports so that it's more likely that if we expand the list in the future, we will apply that to both languages.

@KathleenDollard Do you think this is the right thing to do?

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props
@@ -87,17 +87,15 @@ Copyright (c) .NET Foundation. All rights reserved.
                       and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
                       and $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))">
 
-    <!-- These imports are already defined as VB implicit imports, so only add them for C# here -->
-    <Import Include="System" Condition="'$(Language)' == 'C#'" />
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props
```diff
@@ -82,18 +82,22 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-    <Import Include="System.Collections.Generic" />
+
+    <!-- These imports are already defined as VB implicit imports, so only add them for C# here -->
+    <Import Include="System" Condition="'$(Language)' == 'C#'" />
+    <Import Include="System.Collections.Generic" Condition="'$(Language)' == 'C#'" />
+    <Import Include="System.Linq" Condition="'$(Language)' == 'C#'" />
+    <Import Include="System.Threading.Tasks" Condition="'$(Language)' == 'C#'" />
+    
+    <!-- These imports were not included by default for VB projects prior to .NET 6, so they apply to both languages -->
     <Import Include="System.IO" />
```
JunTaoLuo, line 97: Minor nit.

It would be nice to sort the `Import` items when generating the file now that we organize them logically. We used to preserve the alphabetical order as a side effect of how they are organized but it will no longer be the case after this change. I know it's rare for users to browse to this file, but the fact that it's a visible .cs file for C# projects in the obj directory makes me want to strive for more clean looking file.

JunTaoLuo (approved):

PR #18922, opened 2021-07-14
Update System.CommandLine version and add version subscription

sfoslund:
Update System.CommandLine version and add darc subscription


sfoslund:
@MichaelSimons has asked us to start taking automatic version updates of System.CommandLine so we stay in sync with the templating repo. 

@dotnet/domestic-cat @wli3 once the subscription is enabled these insertions will likely fail pretty regularly as System.CommandLine is still making many breaking changes. Resolving these is not trivial (take for example how many files have had to be touched in this PR) so we'll likely want to treat these insertions mostly as reminders to update before release time, which is when source build will require us to have a matching version with templating. I'll plan on resolving the breaking changes when they arise. 

@jmarolf @JoeRobich @pranavkm these updates will effect dotnet format and watch as well since we all share a System.CommandLine version.

src/Tests/dotnet.Tests/.editorconfig
```diff
@@ -0,0 +1,4 @@
+[*.cs]
+
+# CS0618: Type or member is obsolete
+dotnet_diagnostic.CS0618.severity = suggestion
```
sfoslund, line 4: I'm adding this config in a few places because System.CommandLine has recently marked a few of their APIs as obsolete. We use them across quite a few places in the CLI and after discussing with @jonsequitur there doesn't seem to be a simple fix for this. I'm tracking making a better fix here: https://github.com/dotnet/sdk/issues/18952 but I don't think this work is feasible for .NET 6 timeframe.

dsplaisted (approved):

jmarolf (approved):
:shipit:

PR #18929, opened 2021-07-15
[API Compat] Add rule that validates a base type or interface is not removed

### LINKED ISSUE #18680: [API Compat] Add rule that validates a base type or interface is not removed
This rule should check that the base types and interfaces match on both the implementation and in the contract.

This rule should handle the special case of moving an interface declaration to a base type, which is not a breaking change. 

This should consider strict mode and run both ways.

cc: @ericstj @Anipik

safern:
Fixes: #18680


src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
```diff
@@ -0,0 +1,294 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
Anipik, line 3: nit: extra ```//```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -0,0 +1,85 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
Anipik, line 3: nit: extra ```//```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -0,0 +1,85 @@
...
+                if (Settings.StrictMode)
+                    ValidateBaseTypeNotRemoved(right, left, rightName, leftName, differences);
+            }
+
+            ValidateInterfaceNotRemoved(left, right, leftName, rightName, differences);
+
+            if (Settings.StrictMode)
+                ValidateInterfaceNotRemoved(right, left, rightName, leftName, differences);
+        }
+        private void ValidateBaseTypeNotRemoved(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
```
Anipik, line 41: nit : no new line

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        private void ValidateInterfaceNotRemoved(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            HashSet<ITypeSymbol> rightInterfaces = new(right.GetAllBaseInterfaces(), Settings.SymbolComparer);
+
+            foreach (ITypeSymbol leftInterface in left.GetAllBaseInterfaces())
+            {
+                // Ignore internal interfaces
+                if (!leftInterface.IsVisibleOutsideOfAssembly())
```
Anipik, line 70: should we add a flag here if ```Includeinternals``` is set.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        private void ValidateBaseTypeNotRemoved(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            // Base type order matters in the hierarchy chain so therefore use a list.
+            List<ITypeSymbol> rightBaseTypes = right.GetAllBaseTypes().ToList();
+
+            int lastTypeIndex = 0;
+            foreach (ITypeSymbol leftBaseType in left.GetAllBaseTypes())
+            {
+                lastTypeIndex = rightBaseTypes.FindIndex(lastTypeIndex, r => Settings.SymbolComparer.Equals(leftBaseType, r));
```
Anipik, line 49: can you explain why the start point for search is lastTypeIndex ? how does it ensures the hierarchy

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
safern: I thought about it, however removing an internal interface is not a breaking change. Maybe if we do that we should do the same for removing base types?

cc: @dougbu @ericstj thoughts?

Anipik: we hit this one for old api compat for open telemetry

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
```diff
@@ -0,0 +1,294 @@
...
+
+            CompatDifference[] expected = new[]
+            {
+                new CompatDifference(DiagnosticIds.CannotRemoveBaseType, string.Empty, DifferenceType.Changed, "T:CompatTests.First"),
+                new CompatDifference(DiagnosticIds.CannotRemoveBaseInterface, string.Empty, DifferenceType.Changed, "T:CompatTests.First"),
+            };
+
+            Assert.Equal(expected, differences, CompatDifferenceComparer.Default);
+
+            differences[0].Message.Contains("CompatTests.FirstBase");
```
Anipik, line 91: any specific reason on here we are testing the messages as well ?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
```diff
@@ -0,0 +1,294 @@
...
+  public class First { }
+  public class FirstBase { }
+  public interface IFirstInterface { }
+}
+";
+
+            string rightSyntax = @"
+namespace CompatTests
+{
+  public class First : FirstBase, IFirstInterface { }
```
Anipik, line 138: are you worried that this wont allow developers to derive from new classes while running in the strict mode, should we allow disabling the strict mode for particular rules.

safern: > are you worried that this wont allow developers to derive from new classes while running in the strict mode, should we allow disabling the strict mode for particular rules.

It would allow them to derive from new classes, but they should derive from those new classes on both sides. `StrictMode` is intended for a compile vs runtime asset scenario.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
safern: They are localized and that can cause tests to fail based on the culture settings of the computer executing the tests.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
safern: Sure. The base class hierarchy order can not change, as in order to change the order you would be dropping a base class from another class at some point and that is a breaking change, so we are guaranteed that we can check the base classes from bottom to top on the tree skipping the indexes we already checked.

i.e if you have:

Left:
```cs
public class Foo : FirstBase { }
public class FirstBase : SecondBase { }
public class SecondBase : ThirdBase { }
public class ThirdBase { }
```

if I were to introduce a new base class I would have to honor order and push the previous base class up in the hierarchy:
Right: 
```cs
public class Foo : FirstBase { }
public class FirstBase : NewSecondBase { }
public class NewSecondBase : SecondBase { }
public class SecondBase : ThirdBase { }
public class ThirdBase { }
```

That would be flattened as
Right: `[FirstBase, NewSecondBase, SecondBase, ThirdBase]`
Left: `[FirstBase, SecondBase, ThirdBase]`.

So that way you when we are looking for `ThirdBase` in `Right`, `lastTypeIndex=2`, so we skip the base types we don't care and that way it is faster to find all the base types.

Doing something like the following is a breaking change:
```cs
public class First : FirstBase { }
public class FirstBase : SecondBase { }
public class SecondBase { }
```
changed to:
```cs
public class First : FirstBase, SecondBase { }
public class FirstBase { }
public class SecondBase { }
```
Is breaking as `FirstBase` no longer has `SecondBase` as it's base class.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
safern: Yeah I think you are right we should consider that flag as if internals are included that means a lot of types and interfaces  that are internal will only be able to implement internal interfaces. I don't know if we should for the base type scenario though, don't know what the policy is there.

safern: I added a change to include internals if the flag is set, however abstract classes check is always including non visible outside of the assembly, I don't know if we should filter that out. Don't know if removing an internal or private class from a hierarchy is breaking (my assumption is it is as it could inherit from public base types and we would loose those in the hierarchy?).

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
```diff
@@ -0,0 +1,310 @@
...
+  public interface IFirstInterface { }
+}
+";
+
+            string rightSyntax = @"
+namespace CompatTests
+{
+  public class First : NewBase { }
+  public class Second : NewSecondBase { }
+  public class NewBase : FirstBase, INewInterface { }
```
ericstj, line 34: Can we also test pushing a method down while inserting a base class in the heirarchy?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -0,0 +1,83 @@
...
+                ValidateInterfaceNotRemoved(right, left, rightName, leftName, differences);
+        }
+
+        private void ValidateBaseTypeNotRemoved(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            // Base type order matters in the hierarchy chain so therefore use a list.
+            List<ITypeSymbol> rightBaseTypes = right.GetAllBaseTypes().ToList();
+
+            int lastTypeIndex = 0;
+            foreach (ITypeSymbol leftBaseType in left.GetAllBaseTypes())
```
ericstj, line 45: Do we really need to iterate every base type, or can we just make sure that the immediate base type that left had is present in the heirarchy for right and then count on transitivity catching the rest of the heirarchy?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
safern: You would expect that to not be flagged right?

ericstj: Right, it's releated to this.  Often times the reason for introducing a base class in the heirarchy is so that you can share some (but not all) of the members of a type.  In such a case you create a new base type, and move some of the members down from your original type.  It's a binary compatible change but one I expect we need to handle in the member comparison as well.

safern: I'll make sure that works then. Thanks for the suggestion!

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
safern: I think we need to iterate all left base types if we want to report all removals that affect right in the chain, as this test captures:

Left: https://github.com/dotnet/sdk/blob/a6011df3ec4fa79535da9cb982f647456d1b1e66/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs#L231-L239
Right: https://github.com/dotnet/sdk/blob/a6011df3ec4fa79535da9cb982f647456d1b1e66/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs#L255-L265
Result: https://github.com/dotnet/sdk/blob/a6011df3ec4fa79535da9cb982f647456d1b1e66/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs#L297-#L299

However, we could safely assume that the only base type on left that we need to make sure exists in right is the first immediate base type in left. Reasons to assume that is safe, because if any base type is removed on the chain other than the first one on left, that will be caught when we compare the type which it's base type was removed.

So in the example above, the only difference we would get, it would be:

_SecondBase doesn't inherit from `ThirdBase` anymore._

Rather than getting the error of `ThirdBase` being remove on both, `First` and `SecondBase`.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotRemoveBaseTypeOrInterfaceTests.cs
safern: Added test and found a bug on `MembersMustExist` I was doing a bad assumption using left type to search on the base types' members when I should've right containing type base types.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
ericstj: Using the diagnostics of the break is a good measurement of which is more correct.  Does one approach seem more accurate than the other based on the diagnostics raised?  I'm neutral on preference here, but raised this as a possibility in case it might simplify the code at all.

ericstj (approved):
LGTM.  Thanks for addressing my feedback.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -38,24 +38,25 @@ private void RunOnTypeSymbol(ITypeSymbol left, ITypeSymbol right, string leftNam
...
-                                            DifferenceType.Changed,
-                                            right));
+                // If we found the immediate left base type on right we can assume
+                // that any removal of a base type up on the hierarchy will be handled
+                // when validating the type which it's base type was actually removed.
+                if (Settings.SymbolComparer.Equals(leftBaseType, rightBaseType))
                     return;
-                }
+
+                rightBaseType = rightBaseType.BaseType;
```
ericstj, line 52: Should strict mode break and fallthrough to difference diagnostic?

safern: Ah that's a good point. I think StrictMode in this case would let right to drop a base type that left has and not catch it. I think StrictMode should match and should just look for `left.BaseType == right.BaseType` rather than searching the tree for the presence of right on left.

safern: Hmm actually I tried this and didn't like the outcome of the diagnostics, because if one side drops the basetype, i.e right drops it, then the message would echo saying that left doesn't inherit from object but right does. 

I think it is better to try and search the tree and if right adds a new base type it would be handled correctly the way it is.

PR #18931, opened 2021-07-15
Rewrite assembly identity must match tests to use in memory assemblies

safern:
While adding some rules and running tests I noticed that the new tests for assembly identity rule bumped the execution time from ~10s to ~30-40s in the console and to 1+min in VS. Also depending on machine state sometimes they failed when switching branches. 

This was causing tests to generate a project and build it to get some dlls out and then load them, which for testing purposes is not worth it as it brings some flakiness, we should test rules with in memory assemblies as we just want to test that the rules are interpreting the metadata correctly.

Also I did some cleanup on the rule and improved a diagnostic message.


src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
```diff
@@ -30,170 +36,106 @@ public static void AssemblyNamesDoNotMatch()
...
-            IAssemblySymbol leftSymbols = new AssemblySymbolLoader().LoadAssembly(leftDllPath);
-            IAssemblySymbol rightSymbols = new AssemblySymbolLoader().LoadAssembly(rightDllPath);
+            IAssemblySymbol leftSymbol = SymbolFactory.GetAssemblyFromSyntax(syntax, publicKey: _publicKey);
+            IAssemblySymbol rightSymbol = SymbolFactory.GetAssemblyFromSyntax(syntax, publicKey: _publicKey);
 
             // public key tokens must match
             ApiComparer differ = new();
             differ.StrictMode = strictMode;
 
-            IEnumerable<CompatDifference> differences = differ.GetDifferences(leftSymbols, rightSymbols);
```
Anipik, line 168: can we add a check here that leftAssemblyKeyToken is not null

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
```diff
@@ -202,57 +144,54 @@ public void LeftAssemblyKeyTokenNull(bool strictMode)
...
         }
 
-        [RequiresMSBuildVersionTheory("17.0.0.32901")]
+        [Theory]
         [InlineData(true)]
         [InlineData(false)]
         public void RightAssemblyKeyTokenNull(bool strictMode)
         {
-            var testAsset = _testAssetsManager
-                .CopyTestAsset("PublicKeyTokenValidation", allowCopyIfPresent: true)
```
Anipik, line 211: can you delete this asset as well from the assets folder

safern: Ah, yeah I forgot to do that. Thanks 😄

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AssemblyIdentityMustMatchTests.cs
safern: Yup. I'll add a check in all tests for that kind of information.

Anipik (approved):
left 2 comments, other than that looks good

PR #16777, opened 2021-04-07
Mark source generator file as generated

martincostello:
Give the output file for the `ProvideApplicationPartFactoryAttribute` a `.g` suffix so that it is treated as a generated file to prevent it being flagged by source analyzers for violations the application developer cannot fix themselves. See reactiveui/refit#1067 for a similar fix.

Otherwise an application using a package such as [StyleCop.Analyzers](https://www.nuget.org/packages/StyleCop.Analyzers) can receive warnings/errors for the code emitted by the source generator as it is not determined to have been generated by tooling and to be user code.

For example, using `6.0.100-preview.3.21202.5` of the SDK errors such as the following are emitted;

```
C:\Coding\martincostello\api\src\API\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator\API.UnifiedAssembly.Info.cs(1,221): error SA1518: File is required to end with a single newline character [C:\Coding\martincostello\api\src\API\API.csproj]
C:\Coding\martincostello\api\src\API\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator\API.UnifiedAssembly.Info.cs(1,1): error SA1633: The file header is missing or not located at the top of the file. [C:\Coding\martincostello\api\src\API\API.csproj]
```

I wasn't sure of the best way to add a unit test for this, or how to go about validating this locally.

PR #18943, opened 2021-07-15
Mark Razor source generator files as generated

martincostello:
Give the output files from the Razor Source Generator a `.g.cs` suffix so that they are treated as a generated file to prevent it being flagged by source analyzers for violations the application developer cannot fix themselves and so that tools such as coverlet ignore them.

I _think_ this is part of the fix (and is the same as #16777 which seems to have gone as part of somet refactoring), but I couldn't get the tests to debug through the source generator in Visual Studio 2022 preview 2, so I wasn't able to actually inspect my changes and check they were actually being applied.

I think there's _also_ an issue with coverlet, as if I try and exclude the RSG-generated files from coverage anyway using a filter of `[*]AspNetCoreGeneratedDocument*`, it still refuses to cover any of the code due to it being unable to find the files from the source-generator on disk.

I haven't gotten as far as trying to debug coverlet yet, but I noticed from structured logging that the source files from the RSG didn't have the `.g.cs` suffix:

```
[coverlet] Unable to instrument module: C:\Coding\martincostello\api\tests\API.Tests\bin\debug\net6.0\API.dll, pdb without local source files, [C:\Coding\martincostello\api\src\API\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator\_Views_Docs_Index_cshtml.cs]
```

Relates to dotnet/aspnetcore#31719.

/cc @captainsafia

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerator.Helpers.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerator.Helpers.cs
@@ -37,17 +37,6 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
             return builder.ToString();
         }
 
-        private static void HandleDebugSwitch(bool waitForDebugger)
-        {
-            if (waitForDebugger)
-            {
-                while (!Debugger.IsAttached)
-                {
-                    Thread.Sleep(3000);
-                }
-            }
-        }
-
         private static RazorProjectEngine GetDiscoveryProjectEngine(StaticCompilationTagHelperFeature tagHelperFeature, IEnumerable<MetadataReference> references, IEnumerable<SourceGeneratorProjectItem> items, RazorSourceGenerationOptions razorSourceGeneratorOptions)
         {
             var fileSystem = new VirtualRazorProjectFileSystem();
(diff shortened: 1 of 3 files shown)


captainsafia (commented):
Looks good!

There are some tests that need to be updated. They look for a particular file path. Doing a search of "Razor.SourceGenerator" should find the test cases that need to be updated to look for files with the `.g.cs` extension.

martincostello:
Great, thanks. I'll get those updated 👍

martincostello:
Passing now - I'm not sure what the dev workflow should be for these, but I couldn't seem to get the tests locally to consume the changes I was actually making, so I had to guess at whether it would fix and let AzDo run them for me to get the feedback.

captainsafia:
@martincostello I typically use the instructions here for running tests from the command line: https://github.com/dotnet/sdk/blob/main/documentation/project-docs/developer-guide.md#using-the-built-dotnet

martincostello:
@captainsafia Ah, I didn't find that, thanks 🤦 - I made the assumption it was like the aspnetcore repo, and used the restore, build and test `.cmd` files to do it. I'll know for the future now 😄

PR #18957, opened 2021-07-16
Don't re-run discovery for files that haven't changed

chsienki:
During testing we identified a bottle neck where this is always running, even though it doesn't need to.

This fix groups the source items by name, only re-running discovery if the name of the files change (or when a file is added/removed).

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
@@ -31,7 +31,9 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
                     (c1, c2) => c1 != null && c2 != null && c1.References != c2.References)
                 .Select((compilation, _) => compilation.References);
 
-            var sourceItemsByName = sourceItems.Collect().WithLambdaComparer((@new, old) => @new.SequenceEqual(old, new LambdaComparer<SourceGeneratorProjectItem>((l, r) => string.Equals(l?.FilePath, r?.FilePath, System.StringComparison.OrdinalIgnoreCase))));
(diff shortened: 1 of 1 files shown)


src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -31,9 +31,11 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                     (c1, c2) => c1 != null && c2 != null && c1.References != c2.References)
                 .Select((compilation, _) => compilation.References);
 
+            var sourceItemsByName = sourceItems.Collect().WithLambdaComparer((@new, old) => @new.SequenceEqual(old, new LambdaComparer<SourceGeneratorProjectItem>((l, r) => string.Equals(l?.FilePath, r?.FilePath, System.StringComparison.OrdinalIgnoreCase))));
```
pranavkm, line 34:
```suggestion
            var sourceItemsByName = sourceItems.Collect()
                .WithLambdaComparer((@new, old) => @new.SequenceEqual(old, new LambdaComparer<SourceGeneratorProjectItem>(
                    (l, r) => string.Equals(l?.FilePath, r?.FilePath, System.StringComparison.OrdinalIgnoreCase))));
```

pranavkm: Minor formatting nit. But this looks fine by me.

pranavkm (approved):

chsienki:
Note I also found a bug in the generator driver itself, where it doesn't consider a `.Collect()` to be cached even if all its inputs are. Filed https://github.com/dotnet/roslyn/issues/54855 to track, and means we can make this logic nicer when its fixed.

PR #18985, opened 2021-07-16
Improve caching in RazorSourceGenerator

pranavkm:


src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -20,114 +26,143 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+            var importFiles = sourceItems.Where(static file =>
+            {
+                var path = file.FilePath;
+                if (path.EndsWith(".razor"))
+                {
+                    return Path.GetFileNameWithoutExtension(path) == "_Imports";
+                }
+                else if (path.EndsWith(".cshtml"))
+                {
+                    return Path.GetFileNameWithoutExtension(path) == "_ViewImports";
```
captainsafia, line 45: Are imports files the only scenario where we care about things being in the virtual file system?

If so, making sure that Razor works with global usings (https://github.com/dotnet/aspnetcore/issues/34217#issuecomment-882001435) might help us in removing the virtual filesystem concept all together.

pranavkm: We can't remove _Imports / _ViewImports since they also support other directives and there are existing projects that already use it.

captainsafia: Yeah, I realized that we can't get rid of this shortly after typing it. Alas....

PR #18986, opened 2021-07-16
Don't produce errors when an Exe project references a test project

dsplaisted:
This fixes another case where the NETSDK1130 error would be generated when it should not be.  When an Exe project references a test project, we should not generate this error.

This case was discovered here: https://github.com/dotnet/sdk/issues/17579#issuecomment-881646782

This is similar to #17594, except this fixes the case when the reference is in the other direction.


marcpopMSFT:
@dsplaisted if this is for p7, can you add the template and email tactics?  This seems to meet the bar.

sfoslund (approved):

jamers99:
When will this be fixed in .NET 5?

PR #18990, opened 2021-07-16
Revert "Rationalize default Imports for C# and VB"

marcpopMSFT:
Reverts dotnet/sdk#18921 and #18853


JunTaoLuo (approved):

dsplaisted:
Looks like this [wasn't the cause](https://github.com/dotnet/installer/pull/11122#issuecomment-881970074) for the failure when flowing sdk into installer, and that has now flowed successfully, so I think we can close this.

PR #19078, opened 2021-07-20
Merge preview 7 to main

sfoslund:


mhutch:
The merge commit looks good. Regarding the failures, I think `Given_multi_target_It_should_get_suggested_workload_by_GetRequiredWorkloads_target` was depending on bugs that were fixed in my recent PR.

The workload suggestion logic should never have suggested `microsoft-net-runtime-mono-tooling`, `microsoft-net-runtime-android-aot` or `microsoft-net-runtime-ios`, as these are abstract workloads. That is definitely fixed - only concrete workloads can be suggested now.

However, they should _also_ not have been suggested because they are redundant (they are subsets of the  `-full` workloads) and that seems to indicate a bug in the suggestion algorithms.

I'm a little confused that a `net6.0` project was bringing in `microsoft-net-runtime-mono-tooling`. That suggests a bug on the workload targets being used?

sfoslund:
Thanks Mikayla, @wli3 do you mind reviewing the latest commit since you originally added those tests?

mhutch (approved):

joeloff (approved):
I mostly looked at the admin install changes around the pack IDs

PR #19080, opened 2021-07-20
[Static web assets] Fixes issues we discovered while integrating the static web assets improvements on to the ASP.NET Core repo

javiercn:
## Description

As part of integrating the changes we did to static web assets into the ASP.NET Core repo, we ran into some issues and this PR Contains the fixes to them.

* Filter doc XML files from blazor static web assets
* Fallback to Identity metadata when we try to match project references
  * This is not technically necessary because regular projects have the right metadata, however, ASP.NET Core uses some 
     special project types and its easy/cheap for us to be more lenient here and fallback.
* Correct empty entry for manifest pattern on generation
  * The dotnet.js manifest file was not being passed on to GenerateBlazorBootJson manifest.
* When generating the manifest we use for development we were including empty entries "" as well as the base path for assets 
   in the current project, which is not necessary.
* Only include discovery patterns when the folder exists on disk
  * For example, avoid creating a pattern when the wwwroot folder doesn't exist, since we can't map a non existing folder at runtime.
* Correctly compute the asset path for assets that are copied to the output folder.

**The PR is divided in three commits:**
* The first commit is the important one and contains the fixes mentioned above.
* The second commit contains test improvements we've done to ensure we are solid.
* The third commit updates the test baselines for the different builds we perform.

## Customer Impact

Customers won't be able to load Blazor apps unless they switch to the previous Static Web Assets implementation (we included a flag for this) however, we want to make sure this is the default experience in Preview7 to ensure that we uncover as many issues as we can, since these improvements are the foundation for many of the new features we are building

## Regression?
- [X] Yes
- [ ] No

6.0 Preview6

## Risk
- [ ] High
- [ ] Medium
- [X] Low

The changes are complex, however we put a fallback mechanism in place to force the older implementation to be used. In case a scenario doesn't work for a user, they can resort to the preview6 implementation setting a value on their csproj file.

## Verification
- [X] Manual (required)
- [X] Automated

Manual verification is problematic and limited due to the fact that there is a build time component and a runtime component and that they live on different repositories.

I've been able to use a freshly built "dev sdk" to build one of the samples and i've used the runtime changes (paired with this ones) to run the project and run the published project.

- [ ] Yes
- [ ] No
- [ ] N/A


src/RazorSdk/Tasks/StaticWebAssets/GenerateStaticWebAssetsDevelopmentManifest.cs
```diff
@@ -8,6 +8,7 @@
 using System.Security.Cryptography;
 using System.Text.Encodings.Web;
 using System.Text.Json;
+using System.Text.Json.Serialization;
```
pranavkm, line 11: Do you need this? It looks like the PR removed some of the JSON serialization, but did not add new ones.

src/RazorSdk/Tasks/StaticWebAssets/GenerateStaticWebAssetsDevelopmentManifest.cs
```diff
@@ -64,8 +68,8 @@ public override bool Execute()
 
             var discoveryPatternsByBasePath = DiscoveryPatterns
                 .Select(ComputeDiscoveryPattern)
-                .GroupBy(p => p.BasePath,
-                 (key, values) => (key.Split('/'), values));
+                .GroupBy(p => p.BasePath == p.BasePath ? "" : p.BasePath,
```
pranavkm, line 71: This looks wrong. did you mean `p.BasePath == BasePath`?

pranavkm:
```suggestion
                .GroupBy(p => p.BasePath == p.BasePath ? string.Empty : p.BasePath,
```

pranavkm (approved):

src/RazorSdk/Tasks/StaticWebAssets/GenerateStaticWebAssetsDevelopmentManifest.cs
javiercn: Yes, I noticed too and fixed it.

The goal of this is to make sure that we don't apply the BasePath to patterns in the current project since the (base path for the project will match that from the pattern)

That said, I'll change this after preview7 to include the Source (essentially the PackageId) and do it based on that, so that we make sure we only do it for the current project.

mkArtakMSFT:
Approved by tactics.

PR #19124, opened 2021-07-21
Advertise workload updates in list command

### LINKED ISSUE #17484: Workload list should show update available
For command line, just show if there is any workload available is enough.

But for --machine-readable, we need to surface all workload manifest level and also have the description piped out

### LINKED ISSUE #18994: Add message notifying of available workload updates

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/18994
Fixes https://github.com/dotnet/sdk/issues/17484


dsplaisted:
With this PR, you pretty much have to explicitly check for updated workloads by running `dotnet workload list`.  That seems to make it moot that we are automatically updating the advertising manifest in the background, because we'll never tell you updates are available until you run a command to check.  I think the idea behind automatically checking for updated workloads in the background was that we would automatically advertise the updates in commands that are typically run (ie `dotnet build` and other similar commands).

This PR seems fine, but I don't think it really fixes #18994.

src/Assets/TestProjects/SampleManifest/MockListSampleUpdated.json
```diff
@@ -0,0 +1,45 @@
+﻿{
+    "version": "5.0.0",
```
dsplaisted, line 2: This version should probably be incremented from the baseline.

dsplaisted (approved):
This looks good.  As we discussed, we want to add a notification that workloads updates are available to more commands.

Also, we could consider marking (with an asterisk, maybe) which workloads have updates available in the list of installed workloads, instead of having a separate message at the end listing the ones with updates.

sfoslund:
@dsplaisted I pushed the updates we discussed in the meeting on Friday and they're significant, mind taking another look?

sfoslund:
Ping @dsplaisted should I block on you re-reviewing this PR or go ahead and merge?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -104,6 +112,41 @@ public async Task BackgroundUpdateAdvertisingManifestsWhenRequiredAsync()
             }
         }
 
+        private void WriteUpdatableWorkloadsFile()
```
dsplaisted, line 115: Once you get an update advertised, and then you run `dotnet workload update`, is there anything that turns off the update advertising?  I'm not seeing where this file is deleted or updated once you update workloads.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -104,6 +112,41 @@ public async Task BackgroundUpdateAdvertisingManifestsWhenRequiredAsync()
...
+                    if (updatableWorkloads != null && updatableWorkloads.Any())
+                    {
+                        Console.WriteLine();
+                        Console.WriteLine(LocalizableStrings.WorkloadUpdatesAvailable);
+                    }
+                }
+            }
+            catch (Exception)
+            {
+                // Never surface errors
```
dsplaisted, line 146: I think it would be a good idea to display the error if we are in debug (`-d` I think) or verbose logging mode.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -104,6 +112,41 @@ public async Task BackgroundUpdateAdvertisingManifestsWhenRequiredAsync()
...
+            }
+            File.WriteAllText(filePath, jsonContent);
+        }
+
+        public static void AdvertiseWorkloadUpdates()
+        {
+            try
+            {
+                var backgroundUpdatesDisabled = bool.TryParse(Environment.GetEnvironmentVariable("DOTNET_CLI_WORKLOAD_UPDATE_NOTIFY_DISABLE"), out var disableEnvVar) && disableEnvVar;
+                var adUpdatesFile = Path.Combine(CliFolderPathCalculator.DotnetHomePath, ".dotnet", ".workloadAdvertisingUpdates");
```
dsplaisted, line 133: Should this use the `GetAdvertisingWorkloadsFilePath` method?  Should it be factored out into a static method so the logic can be shared?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
```diff
@@ -561,6 +561,11 @@ public IWorkloadResolver CreateOverlayResolver(IWorkloadManifestProvider overlay
             return overlayResolver;
         }
 
+        public string GetSdkFeatureBand()
+        {
+            return _manifestProvider?.GetSdkFeatureBand() ?? string.Empty;
```
dsplaisted, line 566: Is it ever correct to return an empty string, or should this throw an Exception in that case?

dsplaisted (approved):

PR #19128, opened 2021-07-21
Improve caching in RazorSourceGenerator (#18985)

pranavkm:
* Cache tag helpers when the declaration build of razor files does not change
* Cache generated output invariant of other non _Import files

## Description

As part of .NET preview7, the compiler added new sets of APIs to support incremental source generators. We updated the Razor Source Generator use these APIs to benefit from improved performance. We've since discovered issues in the compiler's implementation of incrementalism (that are being addressed for p7) as well as issues with our source generator. The latter is what this PR is being addressed. In the absence of these changes, there's significant degradation (up to 3x slower) in the inner loop perf for compiling razor files. This degradation affects VS and hot reload.

## Customer Impact

Improvements here would benefit compilation in Visual Studio as well as hot reload in VS and dotnet-watch, particularly in projects containing a large number of Razor (.razor / .cshtml) files.

## Regression?
- [ ] Yes
- [x] No

[If yes, specify the version the behavior has regressed from]

## Risk
- [ ] High
- [x] Medium
- [ ] Low

We have tests for correctness in build incrementalism in the SDK, unfortunately source generator incrementalism is currently disabled in CLI's implementation of the compiler. This is [being addressed](https://github.com/dotnet/roslyn/pull/55023) for p7, but it's possible in my current manual verification I missed scenarios where we are over-caching (bad for correctness) or under-caching (bad for perf, but no biggie since the results are still better).

[Justify the selection above]

## Verification
- [x] Manual (required)
- [ ] Automated

- [ ] Yes
- [ ] No
- [x] N/A

Addresses [issue number]

### DIFF
--- a/src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
+++ b/src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
@@ -10,15 +10,9 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
 {
     internal static class IncrementalValuesProviderExtensions
     {
-        internal static IncrementalValuesProvider<T> WithLambdaComparer<T>(this IncrementalValuesProvider<T> source, Func<T?, T?, bool> equal)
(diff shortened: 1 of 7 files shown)


mkArtakMSFT:
This was approved by Tactics!

PR #19130, opened 2021-07-21
[trying to repro a crash] reenable a test

VSadov:
Reenabling a test for (Windows and Linux only)  in an attempt to capture a crash dump.


VSadov:
@agocke - I think it could be useful to merge this. 

By looking at statistics this test fails mostly on OSX, where dumps are not very useful, so I keep it disabled on OSX to reduce noise.
It would be useful though to get dumps from failures on Linux, which we did not have lately.

agocke (approved):
I think we can merge this and wait for another a failure

VSadov:
Thanks!!

Issue #18813, opened 2021-07-08
Crossgen2 perfmap format version support


trylek:
As of

https://github.com/dotnet/runtime/pull/53792

Crossgen2 newly supports a new command-line option <code>--perfmap-format-version</code> that currently has two legal values, 0 and 1, where 0 corresponds to the "old style Crossgen1" perfmap naming format and 1 is the "new style Crossgen2" naming format. In particular, historically Crossgen1 used to name perfmap files like this:

<code>&lt;assembly-name&gt;.ni.{&lt;MVID of the MSIL file&gt;}.map</code>

This is insufficient and cumbersome in Crossgen2 composite mode where multiple assemblies get compiled into a single native executable; moreover the pre-existing scheme was insufficient anyway as symbol indexation needs to take targeting OS and architecture into consideration and that wasn't reflected in the <code>MVID</code>. The new naming in version 1 is like this:

<code>&lt;assembly-name&gt;.ni.r2rmap</code>

The current POR is that we drop the <code>{MVID}</code> bit and change <code>.map</code> to <code>.r2rmap</code> to avoid clashes with traditional .map files; this will need adjusting the following SDK source files:

https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/RunReadyToRunCompiler.cs

We also need to adjust the script calling them:

https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets

These changes should let SDK make Crossgen2 produce the new perfmap format version 1 and ultimately switch over installer build in the runtime repo to produce the new perfmap style files. According to my understanding the bulk of the follow-up work will involve consuming the metadata in the perfmap file for the purpose of symbol indexation; I believe that should be tracked by a separate work item.

This change may need a counterpart perfview change and possibly additional downstream changes I'm not yet aware of; these will probably merit creating additional specific work items as we discover them.

Thanks

Tomas

/cc @dotnet/crossgen-contrib
@dotnet/dotnet-diag 
@tommcdon 
@brianrob 
@hoyosjs 
@mikem8361

[...]

PR #19132, opened 2021-07-21
Add SDK support for perfmap format version

trylek:
This PR addresses the remaining work outlined in the issue

https://github.com/dotnet/sdk/issues/18813

As of .NET 6 Preview 7 Crossgen2 supports the option
--perfmap-format-version that currently supports two values, 0
(being the legacy value / default) and 1; the only difference is
that version 1 uses a different naming scheme for the output
perfmap files, dropping the {MVID} part and using the extension
".ni.r2rmap".

This PR adds support for the new option in the SDK Crossgen2
publishing logic; for .NET 5 and prior, we naturally need to stick
to the legacy naming; for .NET 6, SDK honors a new property
PublishReadyToRunPerfmapFormatVersion that can be used to override
the default.

As of the 1st commit on this PR, the SDK default is set to 1 i.e.
the "new format". This basically means that from the point of
merging this PR onwards SDK will by default produce the new
perfmap file names on Linux.

Similarly, once the current SDK combines with the runtime repo
(which will likely happen as part of the RC1 fork), the installer
partition will start producing the new naming style for the
Crossgen2-compiled framework.

As this is technically a breaking change, we need to discuss whether
that's acceptable (e.g. if there was no prior support for perfmap
indexation, we probably don't need to care much); if we need tighter
control over the perfmap versioning / naming, there are two different
things we can do:

1) For now change the PublishReadyToRunPerfmapFormatVersion default
to 0 in the SDK Microsoft.NET.CrossGen.targets script. This means that
perfmap files will continue to use the legacy naming scheme until someone
explicitly switches over runtime or the SDK itself later on in the future.

2) We can make a counterpart check-in to the runtime repo setting
PublishReadyToRunPerfmapFormatVersion to 0 in the installer
ReadyToRun.targets script; once we're confident to switch over,
we'll just delete the property from the script.

I'm still hitting a couple of errors in local testing but I'm
publishing the PR anyway as I'm hoping to solicit early feedback for the
change.

Thanks

Tomas

/cc @dotnet/crossgen-contrib

### DIFF
--- a/src/Tasks/Common/MetadataKeys.cs
+++ b/src/Tasks/Common/MetadataKeys.cs
@@ -124,5 +124,6 @@ namespace Microsoft.NET.Build.Tasks
         public const string EmitSymbols = "EmitSymbols";
         public const string IsVersion5 = "IsVersion5";
         public const string CreateCompositeImage = "CreateCompositeImage";
+        public const string PerfmapFormatVersion = "PerfmapFormatVersion";
     }
 }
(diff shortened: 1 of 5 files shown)


AntonLapounov (approved):
Looks good!

brianrob (approved):

davmason (approved):
Thanks Tomas!

PR #19139, opened 2021-07-22
WinRT Component Support.

j0shuams:

I am proposing we add a property in an existing target, which was made for WinRT support recently. As well as add a new target that uses the property set in the existing target to add a reference. We need this to support C# project reference consumers. Because C# consumers only get a .winmd via GetTargetPath, they do not have the assembly containing the *managed* implementation, so we annotate the .winmd to have the path to the implementation, and we add it as a reference here. 

@dsplaisted

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Windows.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Windows.targets
@@ -68,4 +68,31 @@ Copyright (c) .NET Foundation. All rights reserved.
     </ItemGroup>
     
   </Target>
+
+    <Target Name="RemoveManagedWinRTComponentWinMDReferences"
+          AfterTargets="ResolveProjectReferences">
+
+    <PropertyGroup>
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=19139) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=19139) before we can accept your contribution.<br/><br/>:x: j0shuams [sign now](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=19139)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/sdk?pullRequest=19139) it.</sub>

j0shuams:
One thing we should consider is how to handle multiple project references, this assumes there is just one. Will try to use an itemgroup instead of a property.

marcpopMSFT:
Which months release were you looking to get this in for?  We're essentially done with 5.0.2xx and 5.0.3xx for August already and August is the last release of 5.0.3xx.

j0shuams:
@marcpopMSFT Sounds like earliest we can plan for is September then, and I'll rebase to 5.0.4xx

PR #19172, opened 2021-07-22
[Static web assets] Fixes issue with 3.1 and 5.0 wasm apps

javiercn:
## Description

* Fixes an issue discovered by CTI when building old applications with the new SDK
  * The 3.1 wasm apps were targeting the wrong static web assets version.
    * Added a check to validate that RazorLangVersion must be greater than 3 for the new static web assets to apply.
  * The 5.0 apps where using an updated version of GenerateWebassemblyBlazorBootJson
    * Restored and renamed the 5.0 version of GenerateWebassemblyBlazorBootJson and made the 5.0 targets use that.
* Fixes an issue where the scoped CSS bundle was being generated before the sibling scoped css bundles were discovered.
  * This only affected projects having a companion RazorClassLibrary (except for hosted blazor webassembly apps that weren't affected).
  * Packages including scoped CSS bundles were not affected.
  * The fix was to changing the scoped css target dependencies in favor of the standard targets introduced as part of the static web assets improvements.

## Customer Impact

* For the first issue: Customers in Blazor webassembly 3.2 (now out of support) would be unable to build their application with the 6.0 SDK
  * There is an msbuild property they can setup in their project to force static web assets to use the older implementation (which would unblock them).
* Customers using the 6.0 SDK to build 5.0 applications would be hit by an incompatible update we did to one of our tasks.
  * There is no workaround for this other than using an older SDK version.

* For the second issue, Scoped CSS was computing the bundles before collecting the bundles for referenced projects. 
  * This only affects projects referencing a razor class library that is not a Blazor webassembly project.
  * As a workaround, some MSbuild can probably be concocted to force the right order of execution in the user project, but its 
  less than ideal.

## Regression?
- [X] Yes
- [ ] No

6.0 preview6

## Risk
- [ ] High
- [ ] Medium
- [X] Low

For the scoped CSS issue I've identified the gap and added a new test to explicitly cover the scenario.

For the build issues with 3.1 and 5.0 I've built the dogfood version of the SDK with the fix, performed a build and run it with a stock SDK (without rebuilding the project).

## Verification
- [X] Manual (required)
- [X] Automated


src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
```diff
@@ -99,7 +99,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Controls whether or not the static web assets feature is enabled. By default is enabled for netcoreapp3.0
          applications and RazorLangVersion 3 or above. -->
     <StaticWebAssetsEnabled Condition="'$(StaticWebAssetsEnabled)' == ''">$(_Targeting30OrNewerRazorLangVersion)</StaticWebAssetsEnabled>
-    <UseStaticWebAssetsV1 Condition="'$(UseStaticWebAssetsV1)' == '' And ('$(_TargetingNET60OrLater)' == 'true' Or '$(_TargetingNETStandard21OrLater)' == 'true')">false</UseStaticWebAssetsV1>
+    <UseStaticWebAssetsV1 Condition="'$(UseStaticWebAssetsV1)' == '' And ('$(_TargetingNET60OrLater)' == 'true' Or '$(_TargetingNETStandard21OrLater)' == 'true') And '$(RazorLangVersion)' > '3.0'">false</UseStaticWebAssetsV1>
```
pranavkm, line 102: Why are netstandard and 6.0 projects considered equivalent? Can we simply use net6.0 as a way to determine if it needs to use the v2 version of the manifest?

javiercn: blazor webassembly 3.1 is netstandard 2.1 and needs to use the old targets. However, I want netstandard2.1 libraries to use the new static web assets so that existing RCLs produce the assets in the new package format.

The lang requirement discriminates between old blazor webassembly apps and regular RCLs

pranavkm: > However, I want netstandard2.1 libraries to use the new static web assets 

If the two formats are entirely compatible, can we leave it be? It seems problematic because the users may have different experiences depending on what SDK they are using to build their app.

pranavkm: Also, doesn't this affect legacy apps? Using a ns2.1 library with a 3.1 or 5.0 app is a supported scenario.

javiercn: I'm changing the logic here on your suggestion to just use 6.0 and onwards.
For now, we can require rcls that target a previous version to opt-in with the flag.
And in RC1 I have an idea on how to make this backwards compatible.

pranavkm (approved):

PR #19209, opened 2021-07-23
Move PackageValidation to sdk with opt-in flag

Anipik:
- Add EnablePackageValidation opt-in flag
- Use roslyn version that comes with sdk.
- remove any dependencies from the package.
- add infra to override the package validation tasks and targets.

tests conducted
- works with core-msbuild & inbox compiler
- works with desktop-msbuild
- works with core-msbuild & out of band compiler
- works with core-msbuild, inbox compiler and package reference to validation package.


src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -81,7 +82,15 @@
   <Target Name="_AddBuildOutputToPackageDesktop" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
     <ItemGroup>
       <!-- Include .NET Framework build outputs in the package under tools directory. -->
-      <TfmSpecificPackageFile Include="$(OutputPath)**" PackagePath="tools/$(TargetFramework)/%(RecursiveDir)%(FileName)%(Extension)"/>
+      <BuildOutputFiles Include="$(OutputPath)**" 
+                              PackagePath="tools/$(TargetFramework)/%(RecursiveDir)%(FileName)%(Extension)" />
+
+      <TfmSpecificPackageFile Include="@(BuildOutputFiles)" 
+                              Condition="'%(FileName)' == 'Microsoft.DotNet.PackageValidation' or
```
ericstj, line 89: Why is this still necessary with ExcludeAssets above?

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -15,10 +15,11 @@
...
     <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" ExcludeAssets="Runtime" />
-    <PackageReference Include="NuGet.Frameworks" Version="$(NuGetBuildTasksPackageVersion)" />
-    <PackageReference Include="NuGet.Packaging" Version="$(NuGetBuildTasksPackageVersion)" />
-    <PackageReference Include="NuGet.Protocol" Version="$(NuGetBuildTasksPackageVersion)" />
-    <ProjectReference Include="..\Microsoft.DotNet.ApiCompatibility\Microsoft.DotNet.ApiCompatibility.csproj" />
+    <PackageReference Include="NuGet.Frameworks" Version="$(NuGetBuildTasksPackageVersion)" ExcludeAssets="Runtime" />
+    <PackageReference Include="NuGet.Packaging" Version="$(NuGetBuildTasksPackageVersion)" ExcludeAssets="Runtime" />
+    <PackageReference Include="NuGet.Protocol" Version="$(NuGetBuildTasksPackageVersion)" ExcludeAssets="Runtime" />
+    <PackageReference Include="System.IO.Compression" Version="4.3.0" ExcludeAssets="Runtime" />
```
ericstj, line 21: I would expect ExcludeAssets to be transitive.  Why did not need to add this?

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -30,30 +33,58 @@ public class ValidatePackage : TaskBase
 
         public string CompatibilitySuppressionFilePath { get; set; }
 
+        public string RoslynDirectory { get; set; }
+
         protected override void ExecuteCore()
         {
-            RuntimeGraph runtimeGraph = null;
-            if (!string.IsNullOrEmpty(RuntimeGraph))
+            if (string.IsNullOrEmpty(RoslynDirectory))
```
ericstj, line 40: Instead of placing this in `ExecuteCore` can we place it in an override of `Execute`?  That way we get this as close to the start of the task as possible.  It's also beneficial to put it in a method body that does little else, that way it reduces the chance that the JIT might go and load assemblies as part of compiling the code where we hook assembly resolution.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.targets
```diff
@@ -0,0 +1,40 @@
...
+  <ItemGroup Condition="'$(EnablePackageValidation)' == 'true' and '$(DisablePackageBaselineValidation)' != 'true' and '$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
+    <PackageDownload Include="$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))" Version="[$(PackageValidationBaselineVersion)]" />
+  </ItemGroup>
+
+  <Target Name="RunPackageValidation"
+          AfterTargets="Pack"
+          Condition="'$(IsPackable)' == 'true' and '$(EnablePackageValidation)' == 'true'">
+
+    <PropertyGroup Condition="'$(RoslynDirectory)' == ''">
+      <RoslynDirectory>$([System.IO.Path]::GetDirectoryName($(CSharpCoreTargetsPath)))</RoslynDirectory>
```
ericstj, line 14: Can you double check that `CSharpCoreTargetsPath` is defined for VB, and F# projects?

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -30,30 +33,58 @@ public class ValidatePackage : TaskBase
...
         }
+
+        private Assembly ResolverForRoslyn(object sender, ResolveEventArgs args)
+        {
+            AssemblyName name = new(args.Name);
+
+            return name.Name switch
+            {
+                "Microsoft.CodeAnalysis" or "Microsoft.CodeAnalysis.CSharp" =>
+                    Assembly.LoadFrom(Path.Combine(RoslynDirectory, $"{name.Name}.dll")),
```
ericstj, line 85: We ought to check the version we loaded is sufficient to satisfy our reference.  If it isn't, the runtime will "trust" us and then we'd be downgrading the assembly reference and could run into missing members later on.

src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj
```diff
@@ -50,10 +50,13 @@
     <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
```
ericstj, line 53: Looks like other SDKs have avoided this, eg: https://github.com/dotnet/sdk/blob/9ed247b28cba4501185dad4e546dc9c51e1c5711/src/RazorSdk/Tasks/StaticWebAssets/StaticWebAsset.cs#L427

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.targets
```diff
@@ -0,0 +1,40 @@
...
+
+  <ItemGroup Condition="'$(EnablePackageValidation)' == 'true' and '$(DisablePackageBaselineValidation)' != 'true' and '$(PackageValidationBaselinePath)' == '' and '$(PackageValidationBaselineVersion)' != ''">
+    <PackageDownload Include="$([MSBuild]::ValueOrDefault('$(PackageValidationBaselineName)', '$(PackageId)'))" Version="[$(PackageValidationBaselineVersion)]" />
+  </ItemGroup>
+
+  <Target Name="RunPackageValidation"
+          AfterTargets="Pack"
+          Condition="'$(IsPackable)' == 'true' and '$(EnablePackageValidation)' == 'true'">
+
+    <PropertyGroup Condition="'$(RoslynDirectory)' == ''">
```
ericstj, line 13: Double check this naming with Roslyn team.  Ideally we get some agreement for them to define this in future targets (inbox and package).

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
Anipik: how does this look ?
```c#
        public override bool Execute()
        {
            if (string.IsNullOrEmpty(RoslynDirectory))
            {
                throw new ArgumentNullException(nameof(RoslynDirectory));
            }

            AppDomain.CurrentDomain.AssemblyResolve += ResolverForRoslyn;
            try
            {
                base.Execute();
            }
            finally
            {
                AppDomain.CurrentDomain.AssemblyResolve -= ResolverForRoslyn;
            }

            return !Log.HasLoggedErrors;
        }
```

ericstj: Can you move the `return` statement into the `try` as `return base.Execute()`?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/MetadataInformation.cs
```diff
@@ -29,6 +30,13 @@ public MetadataInformation(string name, string targetFramework, string assemblyI
             string.Equals(TargetFramework, other.TargetFramework, StringComparison.OrdinalIgnoreCase) &&
             string.Equals(AssemblyId, other.AssemblyId, StringComparison.Ordinal);
 
-        public override int GetHashCode() => HashCode.Combine(AssemblyName, TargetFramework, AssemblyId);
+        public override int GetHashCode()
```
safern, line 33: Why did we need to do this change?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticBag.cs
```diff
@@ -28,7 +28,11 @@ public DiagnosticBag(string noWarn, (string diagnosticId, string referenceId)[]
 
         public DiagnosticBag(IEnumerable<string> noWarn, (string diagnosticId, string referenceId)[] ignoredDifferences)
         {
-            _noWarn = new HashSet<string>(noWarn);
+            if (_noWarn != null)
```
safern, line 31: Hopefully I'll get time to get rid of this logic on the DiagnosticBag, this should be handled by the suppression engin going forward.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ErrorSuppression/Suppression.cs
```diff
@@ -54,6 +55,14 @@ static bool AreEqual(string? first, string? second)
                 => string.IsNullOrEmpty(first?.Trim()) && string.IsNullOrEmpty(second?.Trim()) || StringComparer.InvariantCultureIgnoreCase.Equals(first?.Trim(), second?.Trim());
         }
 
-        public override int GetHashCode() => HashCode.Combine(DiagnosticId?.ToLowerInvariant(), Target?.ToLowerInvariant(), Left?.ToLowerInvariant(), Right?.ToLowerInvariant(), IsBaselineSuppression);
+        public override int GetHashCode()
+        {
+            int hashCode = 1447485498;
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(DiagnosticId ?? string.Empty);
```
safern, line 61: This is changing the source value, it needs to use ToLowerInvariant as the suppressions are not case sensitive?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.targets
```diff
@@ -0,0 +1,45 @@
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. -->
```
safern, line 1: Should we name these `Compatibility.targets`? I would expect to share this with `API Compat` task when I add it?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.props
```diff
@@ -2,6 +2,6 @@
 <Project>
   <PropertyGroup>
     <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tools\net472\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
-    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tools\netcoreapp3.1\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
+    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tools\net6.0\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
```
safern, line 5: Are we going to support overriding these tools via a NuGet package? Are you planning on adding such support on a separate PR?

Anipik: > Are we going to support overriding these tools via a NuGet package? Are you planning on adding such support on a separate PR?

yes the support is already there, ```Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets ``` is overriding this property in case of package reference.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.targets
Anipik: i think we will need to have the same name as the package name to automatically port the package files. we can change the package name to Microrost.DotNet.Compatibility ?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/MetadataInformation.cs
Anipik: This is in order to remove the Microsoft.Bcl.HashCode dependency and this is in custom with what sdk is doing at other places.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -55,5 +80,20 @@ protected override void ExecuteCore()
...
             }
         }
+
+        private Assembly ResolverForRoslyn(object sender, ResolveEventArgs args)
+        {
+            AssemblyName name = new(args.Name);
+            if (name.Name == "Microsoft.CodeAnalysis" || name.Name == "Microsoft.CodeAnalysis.CSharp")
+            {
+                Assembly asm = Assembly.LoadFrom(Path.Combine(RoslynAssemblyPath, $"{name.Name}.dll"));
+                if (asm.GetName().Version < new Version(RoslynMinVersion))
```
ericstj, line 90: This is unnecessary, just use the version passed in on `ResolveEventArgs`

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -9,8 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="$(MicrosoftCodeAnalysisCSharpPackageVersion)" />
+    <!-- Please update RoslynMinVersion in Microsoft.NET.PackageValidation.targets when we use newer version of roslyn.-->
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-3.21372.2" />
```
ericstj, line 13: I don’t think you should do this. See other comment

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.targets
```diff
@@ -0,0 +1,45 @@
...
+      NoWarn="$(NoWarn)"
+      RunApiCompat="$([MSBuild]::ValueOrDefault('$(RunApiCompat)', 'true'))"
+      EnableStrictModeForCompatibleTfms="$([MSBuild]::ValueOrDefault('$(EnableStrictModeForCompatibleTfms)', 'true'))"
+      EnableStrictModeForCompatibleFrameworksInPackage="$([MSBuild]::ValueOrDefault('$(EnableStrictModeForCompatibleFrameworksInPackage)', 'false'))"
+      GenerateCompatibilitySuppressionFile="$(GenerateCompatibilitySuppressionFile)"
+      CompatibilitySuppressionFilePath="$(CompatibilitySuppressionFilePath)"
+      BaselinePackageTargetPath="$(PackageValidationBaselinePath)"
+      DisablePackageBaselineValidation="$(DisablePackageBaselineValidation)"
+      RoslynAssemblyPath="$(RoslynAssemblyPath)"
+      RoslynMinVersion="$(RoslynMinVersion)" />
```
ericstj, line 43: Please remove this parameter. It is not an external decision. This is determined by what we compile against and should be observed by the version requested in the assembly resolve handler.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,34 +1,7 @@
...
-      RuntimeGraph="$(RuntimeIdentifierGraphPath)"
-      NoWarn="$(NoWarn)"
-      RunApiCompat="$([MSBuild]::ValueOrDefault('$(RunApiCompat)', 'true'))"
-      EnableStrictModeForCompatibleTfms="$([MSBuild]::ValueOrDefault('$(EnableStrictModeForCompatibleTfms)', 'true'))"
-      EnableStrictModeForCompatibleFrameworksInPackage="$([MSBuild]::ValueOrDefault('$(EnableStrictModeForCompatibleFrameworksInPackage)', 'false'))"
-      GenerateCompatibilitySuppressionFile="$(GenerateCompatibilitySuppressionFile)"
-      CompatibilitySuppressionFilePath="$(CompatibilitySuppressionFilePath)"
-      BaselinePackageTargetPath="$(PackageValidationBaselinePath)"
-      DisablePackageBaselineValidation="$(DisablePackageBaselineValidation)"/>
-  </Target>
```
ViktorHofer, line 33: By moving all this code into the SDK, we can't just update the package without updating the SDK. Is that intentional?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackageValidation.props
safern: But you moved these files into the SDK, I don't see how these are being packed anymore.

ViktorHofer: When using the package you don't just want to consume newer assemblies but also updates to the msbuild files which isn't possible with the current design.

safern: Yes.. these files should be included in the package and should either override the targets or set a property on the props file that causes the SDK to not longer include the inbox targets if ordering is an issue.

ViktorHofer: @Anipik if you need inspiration on how to accomplish that, look at the ILLink.Tasks integration:
- https://github.com/mono/linker/blob/main/src/ILLink.Tasks/sdk/Sdk.props
- https://github.com/mono/linker/blob/b888d672e11588e2bbca1cd3eeaee30d53416897/src/ILLink.Tasks/build/Microsoft.NET.ILLink.Tasks.props#L15
- https://github.com/dotnet/sdk/blob/3983460336101a26a2304a5ccfc34ba379a015f6/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.props#L14

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -55,5 +78,20 @@ protected override void ExecuteCore()
...
+
+        private Assembly ResolverForRoslyn(object sender, ResolveEventArgs args)
+        {
+            AssemblyName name = new(args.Name);
+            if (name.Name == "Microsoft.CodeAnalysis" || name.Name == "Microsoft.CodeAnalysis.CSharp")
+            {
+                Assembly asm = Assembly.LoadFrom(Path.Combine(RoslynAssemblyPath, $"{name.Name}.dll"));
+                if (asm.GetName().Version < name.Version)
+                {
+                    throw new Exception(Resources.UpdateSdkVersion);
```
ericstj, line 90: Might be helpful to list more details about what assembly was being loaded, what version was expected and what version was found?  Also, the action taken would be to use a newer compiler.  In the common case this assembly would come with the SDK and the compiler version in the SDK would always be sufficient.  Users would hit this error if they tried to use the OOB compilers, or tried to consume the OOB PackageValidation.

Anipik: It will also throw a load error with information regarding the assembly  its trying to load.

ericstj: That won't list why we rejected it though.  Might be helpful to include at least the version info.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -0,0 +1,36 @@
...
+<Project>
+  <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
+
+  <Target Name="RunPackageValidation"
+          AfterTargets="Pack"
+          Condition="'$(IsPackable)' == 'true' and '$(EnablePackageValidation)' == 'true'">
+
+    <PropertyGroup Condition="'$(RoslynAssemblyPath)' == ''">
+      <RoslynAssemblyPath>$(RoslynTargetsPath)</RoslynAssemblyPath>
+      <RoslynAssemblyPath Condition="'$(MSBuildRuntimeType)' == 'Core'">$([System.IO.Path]::Combine('$(RoslynAssemblyPath)', bincore))</RoslynAssemblyPath>
```
RikkiGibson, line 11: Can/should Roslyn own initialization of this property? IIRC Jared has talked about how the paths to our assemblies are subject to change in the future. For example the net472 assemblies may move into a `bin/` folder next to the `bincore/` folder that the .NET Core assemblies currently live in. Wondering what you think @chsienki.

Anipik: > Can/should Roslyn own initialization of this property?

yeah thats the idea for future. But this is the best we can do for now.

chsienki: Agreed that Roslyn should own this, but in the short term it seems fine to put this in the SDK. 

Even if 3rd parties start taking a dependency, then I think it should still work as this target is going to have to import the roslyn ones, meaning it'll flow through when we move it. I opened https://github.com/dotnet/roslyn/issues/55186 to track this.

ericstj: @RikkiGibson @chsienki are you OK with the property name `RoslynAssemblyPath` as it has been defined here, so that when Roslyn defines it we don't need to react?

chsienki: @ericstj Yeah, I guess my only nit would be make it Assemb**lies** path as there is more than one and it matches with the `RoslynTargetsPath`. Otherwise I'm not sure what else we would call it.

src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs
```diff
@@ -30,6 +33,26 @@ public class ValidatePackage : TaskBase
 
         public string CompatibilitySuppressionFilePath { get; set; }
 
+        public string RoslynAssemblyPath { get; set; }
+
+        public override bool Execute()
+        {
+            if (string.IsNullOrEmpty(RoslynAssemblyPath))
```
safern, line 40: Should we add the `Required` attribute to the parameter so that if it is empty rather than throwing an exception we get an MSBuild error saying that parameter is required?

src/Compatibility/Microsoft.DotNet.PackageValidation/buildMultiTargeting/Microsoft.DotNet.PackageValidation.props
```diff
@@ -1,4 +0,0 @@
-<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. -->
```
safern, line 1: Do we still need the multitargeting support?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -0,0 +1,36 @@
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. -->
+<Project>
+  <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
+
+  <Target Name="RunPackageValidation"
+          AfterTargets="Pack"
+          Condition="'$(IsPackable)' == 'true' and '$(EnablePackageValidation)' == 'true'">
```
safern, line 7: Should the package set `EnablePackageValidation` to `true` on it's props file? It seems weird to reference the package and still needing to set the opt-in flag.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.props
```diff
@@ -2,6 +2,6 @@
 <Project>
```
safern, line 2: Let's not rename this file yet as the package hasn't been renamed, so this is breaking the NuGet convention and therefore will not be imported.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.props
```diff
@@ -2,6 +2,6 @@
...
   <PropertyGroup>
     <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tools\net472\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
-    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tools\netcoreapp3.1\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
+    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tools\net6.0\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
```
safern, line 5: It seems like you are missing to update the csproj to build for `net6.0` rather than `netcoreapp3.1`.

safern: Also, are these going to be under `tools` in the SDK? Is that how the SDK is composed?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
Anipik: Resolved offline

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,34 +1,10 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. -->
 <Project>
-  <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
+  <PropertyGroup>
+    <DotNetPackageValidationAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tools\net472\Microsoft.DotNet.PackageValidation.dll</DotNetPackageValidationAssembly>
```
safern, line 4: IT seems like there is a property that points to the directory where the tasks are already, it is defined in the Common targets and we are imported after them so it should work?

https://github.com/dotnet/sdk/blob/3c35324c260eb11143a1c647169f21b8e4632d45/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets#L20

src/Compatibility/Microsoft.DotNet.PackageValidation/buildMultiTargeting/Microsoft.DotNet.PackageValidation.targets
```diff
@@ -1,4 +1,4 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. -->
 <Project>
   <Import Project="$(MSBuildThisFileDirectory)..\build\$(MSBuildThisFile)" />
-</Project>
\ No newline at end of file
+</Project>
```
safern, line 4:
```suggestion
</Project>
```

safern (approved):
LGTM.

src/Compatibility/Microsoft.DotNet.PackageValidation/build/Microsoft.DotNet.PackageValidation.targets
Anipik: It doesnt get defined for multi-targeting.

safern: This file gets imported in both multi target and single target:

https://github.com/dotnet/sdk/blob/3c35324c260eb11143a1c647169f21b8e4632d45/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CrossTargeting.targets#L14

https://github.com/dotnet/sdk/blob/3c35324c260eb11143a1c647169f21b8e4632d45/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L12

PR #19229, opened 2021-07-24
Fix incremental in RunProduceContentAssets target.

### LINKED ISSUE #10914: MSBuild NuGet Content Asset handling alternately creates and deletes content every other build
SDK version: .NET Core SDK 3.1.200 (and also 3.1.102)

When a .NET project has a reference to a NuGet package that includes a content asset to be compiled into a project (e.g., the `Nullable` NuGet package adds a `Nullable.cs` source file to the compilation for projects for some target frameworks), if you run builds twice in succession, they alternate between copying the file into the `obj` folder and deleting it.

This causes a problem for other parts of the build that want to use the file if they happen to run at the wrong time. For example, the problem described at https://github.com/tonerdo/coverlet/issues/714#issuecomment-579704498 appears to arise from this.

In that case, Coverlet wants to be able to read the source file as part of its code coverage handling. But because of this issue, alternate builds can end up failing to produce any code coverage information, because on every other build, the `Nullable.cs` file is missing.

The cause appears to be the behaviour of the `IncrementalClean` target (in `Microsoft.Common.CurrentVersion.targets`). This seems to be trying to remove any intermediate build outputs that were produced by a previous build but which it believes are no longer required. Unfortunately, in cases where the content asset in question was put in place by an earlier build, the `RunProduceContentAssets` target (in `Microsoft.PackageDependencyResolution.targets`) does not list the file in its `FileWrites` output (because the file was already there so it didn't write it), with the effect that the `IncrementalClean` has no way of knowing that this content asset is still wanted.

The result is that `IncrementalClean` removes a file because it mistakenly thinks it is a hangover from an older build that is no longer required.

This often goes unnoticed because the relevant cleanup happens after compilation, so it typically doesn't greatly matter if the file vanishes. But when parts of the build want to inspect source code later on (as Coverlet does—it wants to inspect the source as part of its instrumentation process) this overly aggressive cleanup causes problems.

(This is causing some of our CI builds to fail to produce code coverage information, because we run the test with a separate `dotnet test` step after the `dotnet build`, and since that `dotnet test` is the 2nd build to occur, it falls foul of this problem. And as far as we can tell, the obvious workaround of specifying `--no-build` won't work because the coverage process relies on instrumentation that is performed as part of the build.)

yuehuang010:
Fix an issue where FileWrites item will "forget" that it produced a folder.  This leads to incremental clean step to delete it between runs.  With the file deleted, this target will run again creating new files and breaking incremental.  The fix is to always add the file to FileWrites.  ProduceContentAssets task is already incremental and won't write to the same location twice.

Fix #10914.


wli3:
@yuehuang010 could you add a test?

yuehuang010:
@wli3 Is there an existing build test where I could extend?

wli3:
"IncrementalBuild_WithP2P_WorksWhenBuildProjectReferencesIsDisabled" would be a good example. Do you know a good nuget package to test on?

yuehuang010:
Hi @wli3, can you help me with the error?  I don't understand it and it doesn't repro locally.  Thanks.

yuehuang010:
@wli3 Ping.

sfoslund:
> Expected File C:\h\w\AD7A0949\t\dotnetSdkTests\nfuah40l.l3t\It_handles_co---209447AF\ContentFiles\obj\Debug\netcoreapp3.0\NuGet\61139A41364EEBD4B8F4B70445B190E5E43DDAC3\ContentFilesExample\1.0.2\ExampleReader.cs to exist, but it does not.

@yuehuang010 this is the error we're seeing in the tests. So you aren't able to repro this locally? Note that you may have to do a full rebuild of the SDK before changes in the props/targets files are reflected.

yuehuang010:
@sfoslund Test Fixed, Thanks.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToUseContentFiles.cs
```diff
@@ -1,21 +1,18 @@
...
-using FluentAssertions;
-using System.Xml.Linq;
-using System.Collections.Generic;
-using System;
-using Xunit.Abstractions;
-using Microsoft.NET.TestFramework.ProjectConstruction;
-
 namespace Microsoft.NET.Build.Tests
 {
+    using System.IO;
```
sfoslund, line 5: nit: please leave the usings as they were for the sake of consistency

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToUseContentFiles.cs
```diff
@@ -26,7 +23,7 @@ public GivenThatWeWantToUseContentFiles(ITestOutputHelper log) : base(log)
         [Fact]
         public void It_handles_content_files_correctly()
         {
-            const string targetFramework = "netcoreapp2.0";
+            const string targetFramework = "netcoreapp3.1";
```
sfoslund, line 26: Does this issue repro with the 6.0? If so, we prefer to use the latest TFMs in our tests.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -397,15 +397,14 @@ Copyright (c) .NET Foundation. All rights reserved.
 
       <Output TaskParameter="CopyLocalItems" ItemName="_ContentCopyLocalItems" />
       <Output TaskParameter="ProcessedContentItems" ItemName="_ProcessedContentItems" />
-      <Output TaskParameter="FileWrites" ItemName="FileWrites" />
+      <Output TaskParameter="ProcessedContentItems" ItemName="FileWrites" />
```
sfoslund, line 400: I'm not following what this chance achieves, maybe someone more familiar with this task could take a look? @dsplaisted does this change look okay to you?

yuehuang010: The itemgroup "FileWrites" contains a list of items that the this project has or would produce.  The task output "FileWrites" is a list of files that has copied.  On incremental build, this task copied no files, thus "FileWrites" itemgroup receives less files.  At the end of the build, incremental clean step will delete files not in itemgroup "FileWrites".  This delete the files copied by this task.  The next build will re-copy the files and the cycle repeats.

Incremental build would break every other build.

dsplaisted: I think this change makes sense.  Can we delete the `FileWrites` output parameter from the `ProduceContentAssets` task, since it's not used anymore and if it is used it's probably going to cause incremental build bugs.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToUseContentFiles.cs
```diff
@@ -26,7 +25,7 @@ public GivenThatWeWantToUseContentFiles(ITestOutputHelper log) : base(log)
         [Fact]
         public void It_handles_content_files_correctly()
```
dsplaisted, line 26: Rather than updating the existing test to cover more scenarios, I'd prefer adding a test that specifically covers incremental build.

dsplaisted (approved):

dsplaisted:
Note that the main branch now targets .NET 7, so you may want to retarget this to 6.0.2xx (once it exists), or 6.0.1xx-rc2 (which I think is still in tell mode for SDK changes).

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToUseContentFiles.cs
yuehuang010: I consider it, but this saves an extra build for an already long run time for a unit test.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
yuehuang010: If I was main owner, I would agree.  My change is focused on low risk and low chance of regression.  I don't want to trackdown everyone else that uses it, even though it's probably zero. Feel free to take over.

yuehuang010:
> Note that the main branch now targets .NET 7, so you may want to retarget this to 6.0.2xx (once it exists), or 6.0.1xx-rc2 (which I think is still in tell mode for SDK changes).

I couldn't get "net7.0" to pass even after syncing.  Is that the right TFV?

dsplaisted:
> > Note that the main branch now targets .NET 7, so you may want to retarget this to 6.0.2xx (once it exists), or 6.0.1xx-rc2 (which I think is still in tell mode for SDK changes).
> 
> @dsplaisted I couldn't get "net7.0" to pass even after syncing. Is that the right TFV?

I don't think using `net7.0` as a `TargetFramework` is working yet.  What I meant is that changes that go in the main branch won't ship until .NET 7.

dsplaisted:
@yuehuang010 I've merged this into main, but it will need to be ported to a 6.0 branch if we want to ship it in 6.0

PR #19270, opened 2021-07-26
[ASP.NET] Add support for JS modules and initializers

javiercn:
This PR adds support for JS Modules and initializers:
https://github.com/dotnet/aspnetcore/issues/26145
https://github.com/dotnet/aspnetcore/issues/32353

JS modules are discovered based on the following patterns:
```console
// For components
**.razor.js"
// For Mvc pages and views
**.cshtml.js"
```

We discover these files and copy them to the output folder in the right location (we strip the pages, shared, views, components, prefix)

Blazor applications can load these modules in the same way they load other modules, the only addition is that we define this convention instead of you having to do it manually.

For JS initializers:
* Each project/package can define a file with the name `$(PackageId).lib.module.js` which will be detected as a library initializer.
* Blazor webassembly will embed the list of initializers within blazor.boot.json and will import them automatically during the boot process. In the Blazor case, the initializers can export a couple of well-known functions to customize the blazor loading process, but that is outside of the scope of this PR.
* For other types of applications a json file is automatically generated as part of the build that includes the list of initializers for the application.
  * Blazor server will read this file and emit its contents to the generated HTML when it writes the first component on to the page or if no component was emitted, it will automatically load it on start.
  * Blazor desktop will just load the file as part of the boot process.

MVC/Razor pages applications won't automatically load any of the library initializers, however, it is trivial to include a small script to fetch the manifest and trigger the load of all the library initializers.

Finally, this PR includes a small set of changes to the Blazor publish process to enable customizing the blazor build process. We do this by collecting the list of Blazor files and writing a file to disk containing their full paths. That new file is included in the blazor.boot.json manifest in a new `extensions` section that contains a dictionary of custom extensions by name.

This is enough to prove that users can customize the Blazor build process inside their app or from an external nuget package to suit your requirements. Upon boot (outside of the scope of this PR), blazor will load the blazor.boot.json manifest, load the library initializers and invoke them giving them their options, from which they can customize the remaining loading process.

### This PR has been updated based on our discussions


src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.JSModules.targets
```diff
@@ -0,0 +1,325 @@
...
+  <Target Name="_ResolveJsModuleInputs" DependsOnTargets="ResolveProjectStaticWebAssets">
+
+    <ItemGroup>
+      <_JSModuleCandidates Include="@(StaticWebAsset)" Condition="'%(SourceType)' == 'Discovered'" />
+    </ItemGroup>
+
+    <!-- This runs after we've discovered the original set of files, we already found the assets in the wwwroot, and here we just need
+         to identify them and correctly clasify them. Modules from other projects or packages will already be correctly tagged when we
+         retrieve them.
+    -->
```
SteveSandersonMS, line 74: I have strong reservations about us hard-coding filename conventions like `Pages/`, `Shared/`, `Components/`, etc. into the SDK. These are arbitrary details of the project templates and not part of the framework. I personally would not have folder names like those in any project I created (since I would group things by feature, not by file type), and most likely teams working in other languages would have different names for anything they add.

If we need to have a filter rule, can we force the developer to declare it? Something like:

```xml
<ItemGroup>
    <JavaScriptModule Include="MyApp/**/*.js" />
</ItemGroup>
```

... or something like that?

Since it's not obvious that the SDK is going to bundle and publicly ship arbitrary `.js` files from the project, it's ideal for developers to have to declare this explicitly and not have a default. But if there is a default, it definitely shouldn't include predefined strings like `Pages`, `Shared`, etc.

javiercn: Hmm, Pages, Shared, Views, etc. Are at least well-known conventions in the MVC world. Our template ships with a Pages folder that most people use. 

If you want to have a different convention, we can provide a way for users to do that, but I think we should have default conventions that match the patterns we normally follow.

SteveSandersonMS: I'd love to fix our templates not to have things like `Pages`, `Shared`, etc. On the assumption that we're going to do that (for Blazor at least) it doesn't seem wise to bake them as rules into the SDK.

javiercn:
This PR is broken down into three commits:
* The first commit has the code and the tests.
* The second commit updates existing baselines as a result of these changes.
* The third commit includes the new baselines.

src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.JSModules.targets
javiercn: This is done now

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -319,7 +313,54 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      PathPrefix=""
+      UseAlternatePathDirectorySeparator="true"
+    >
+      <Output TaskParameter="AssetsWithTargetPath" ItemName="_BlazorJsModuleCandidatesForBuildWithTargetPath" />
+    </ComputeStaticWebAssetsTargetPaths>
+
+    <GetFileHash Files="@(_BlazorJsModuleCandidatesForBuildWithTargetPath)" Algorithm="SHA256" HashEncoding="base64">
+      <Output TaskParameter="Items" ItemName="_BlazorOutputWithHash" />
+    </GetFileHash>
+
```
pranavkm, line 347:
```suggestion
```

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -368,8 +409,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     </ItemGroup>
   </Target>
 
-  <Target Name="_ProcessPublishFilesForBlazor" DependsOnTargets="LoadStaticWebAssetsBuildManifest" AfterTargets="ILLink">
-    
+  <Target Name="ProcessPublishFilesForBlazor" DependsOnTargets="LoadStaticWebAssetsBuildManifest" AfterTargets="ILLink">
```
pranavkm, line 412: Is this meant to be a user extensibility point? Is there a reason they cannot configure itemgroups that are consumed by this globally (outside of a target)? We've had a terrible record of telling people to hook up to targets and taking it away when we refactor or change our implementation.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -408,15 +449,42 @@ Copyright (c) .NET Foundation. All rights reserved.
...
       <Output TaskParameter="FilesToRemove" ItemName="_PublishResolvedFilesToRemove" />
-    </ComputeBlazorPublishAssets>    
-    
-    <ItemGroup>      
+    </ComputeBlazorPublishAssets>
+
+    <ItemGroup>
       <ResolvedFileToPublish Remove="@(_PublishResolvedFilesToRemove)" />
       <StaticWebAsset Include="@(_NewBlazorPublishStaticWebAssets)" />
+      <PublishBlazorBootStaticWebAsset Include="@(StaticWebAsset)" Condition="'%(AssetKind)' != 'Build' and (('%(StaticWebAsset.AssetTraitName)' == 'BlazorWebAssemblyResource' and '%(StaticWebAsset.AssetTraitValue)' != 'manifest' and '%(StaticWebAsset.AssetTraitValue)' != 'boot') or '%(StaticWebAsset.AssetTraitName)' == 'Culture')" />
```
pranavkm, line 457: Could we split this across multiple lines?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.Current.props
```diff
@@ -39,6 +39,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <StaticWebAssetsAdditionalPublishProperties>$(StaticWebAssetsAdditionalPublishProperties);BuildProjectReferences=false;ResolveAssemblyReferencesFindRelatedSatellites=true;_PublishingBlazorWasmProject=true</StaticWebAssetsAdditionalPublishProperties>
     <StaticWebAssetsAdditionalPublishPropertiesToRemove>$(StaticWebAssetsAdditionalPublishPropertiesToRemove);NoBuild;RuntimeIdentifier</StaticWebAssetsAdditionalPublishPropertiesToRemove>
 
+    <!-- JS Modules -->
+    <!-- We disable the manifest generation because we are going to inline the modules in the blazor.boot.json manifest -->
+    <GenerateJSModuleManifest>false</GenerateJSModuleManifest>
```
pranavkm, line 44: Can we put this in the targets file if it's being unconditionally enabled? Putting this in the props file makes it much harder to change it in the future, if for e.g. we'd like to base this on a value provided in the proj file or based on ambient values such as TFM.

src/RazorSdk/Tasks/GenerateJSModuleManifest.cs
```diff
@@ -0,0 +1,64 @@
...
+        public override bool Execute()
+        {
+            var modules = JsModules.Select(StaticWebAsset.FromTaskItem).Select(s => s.ComputeTargetPath("", '/')).ToArray();
+            Array.Sort(modules, StringComparer.Ordinal);
+
+            PersistModules(modules);
+            return !Log.HasLoggedErrors;
+        }
+
+        private void PersistModules(string [] modules)
```
pranavkm, line 37:
```suggestion
        private void PersistModules(string[] modules)
```

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/WasmJsModulesIntegrationTests.cs
```diff
@@ -0,0 +1,342 @@
...
+        public void JsModules_CanCustomizeBlazorInitialization()
+        {
+            // Arrange
+            var testAsset = "BlazorWasmMinimal";
+            ProjectDirectory = CreateAspNetSdkTestAsset(testAsset);
+            File.WriteAllText(Path.Combine(ProjectDirectory.TestRoot, "wwwroot", "blazorwasm-minimal.lib.module.js"), "console.log('Hello initializer')");
+
+            ProjectDirectory.WithProjectChanges(document =>
+            {
+                document.Root.Add(new XElement("Target",
```
pranavkm, line 202: I think you're better off using `XElement.Parse` at this point. It's a bit difficult to follow what's happening here.

pranavkm (approved):

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: We need to compute the list of Blazor assets, which happens in this target, and then we pick up any potential result from modifying it, and we add it to the list of assets.

I went this route becasue I believe its the simplest for customers, do you have any other way in mind that keeps the simplicity and gives us more flexibility?

The other thing I can think of is to get them to add themselves to a property group, and we'll make sure they are always called at the right time, how does that sound?

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/WasmJsModulesIntegrationTests.cs
javiercn: done

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.Current.props
javiercn: done

StevenRasmussen:
Thanks for such a quick reply!  So just to clarify: while debugging I shouldn't see them placed in the `wwwroot` folder or anything? Blazor just knows that I'm debugging and so it uses the original file or something?  But I would still load them like:
``` csharp
var module = await JS.InvokeAsync<IJSObjectReference>("import", "./Overlay.razor.js");
```

StevenRasmussen:
Hmmm.... I must be doing something wrong.  Here's my C#:
```csharp
protected override async Task OnAfterRenderAsync(bool firstRender)
{
	await base.OnAfterRenderAsync(firstRender);
	_module = await JS.InvokeAsync<IJSObjectReference>("import", "./Overlay.razor.js");
}
```
Error message in the browser after the call to `JS.InvokeAsync` to try and load the module:
image

Also, I'm happy to open a new ticket to further the discussion as I know this is getting off-topic for the PR itself.  Was just hoping that there was something simple I was missing (which could still be the case as this is my first foray into loading modules in js).

PR #19392, opened 2021-07-29
Add links to NETSDK1150 and NETSDK1151 messages

dsplaisted:
Add links to error messages generated when there's a mismatch between self-contained values when an Exe project references another Exe.

Currently these links point to the breaking change documentation here: https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error

This should make it a lot easier for people that hit this error message to turn it off.

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -704,11 +704,11 @@ To install these workloads, run the following command: {1}</value>
     <comment>{StrBegin="NETSDK1149: "}</comment>
   </data>
   <data name="SelfContainedExeCannotReferenceNonSelfContained" xml:space="preserve">
-    <value>NETSDK1150: The referenced project '{0}' is a non self-contained executable.  A non self-contained executable cannot be referenced by a self-contained executable.</value>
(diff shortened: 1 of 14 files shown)


sfoslund (approved):

beatcracker:
Any chance of backporting this to .NET 5 SDK? The fe6e41152cc26a132e2993e14a5c5c4723b2ab3b is tagged with  `v6.0.100-rc.2.21505.11` and `v6.0.100-rc.1.21458.71`.

I'm  on the `5.0.402` and the links are not there:

```none
/usr/share/dotnet/sdk/5.0.402/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(993,5):
Error NETSDK1150: The referenced project '../foo/bar.csproj' is a non self-contained executable. 
A non self-contained executable cannot be referenced by a self-contained executable.
```

This message above is from the container-based build using .NET SDK 5 image:

- `mcr.microsoft.com/dotnet/sdk:5.0-buster-slim`
- `sha256:5d6aff01076f269ecf5b9d9e5d708d1cb21636636f54c47d929efde9e89bb197`

PR #19394, opened 2021-07-29
[ASP.NET Core] Fine tuning some of the targets for Blazor Desktop

javiercn:
* Adds a couple of targets to standardize the point on the build where assets get generated if they haven't been generated at an earlier point in the build.
* This helps features like CSS isolation standarize on where to generate assets and enables platforms like Blazor desktop to move the generation to an earlier point in the build without specific knowledge of the individual build features.


javiercn:
This has been superseeded by another PR

PR #19432, opened 2021-07-30
Update SDK to support new DI and Attribute feature switches

eerhardt:
See
* https://github.com/dotnet/runtime/pull/55102
* https://github.com/dotnet/runtime/pull/56475

@pranavkm @captainsafia - note the changes to the Blazor WASM targets file.

### DIFF
--- a/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
+++ b/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
@@ -47,6 +47,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <UseSystemResourceKeys Condition="'$(UseSystemResourceKeys)' == ''">true</UseSystemResourceKeys>
     <EnableUnsafeUTF7Encoding Condition="'$(EnableUnsafeUTF7Encoding)' == ''">false</EnableUnsafeUTF7Encoding>
     <HttpActivityPropagationSupport Condition="'$(HttpActivityPropagationSupport)' == ''">false</HttpActivityPropagationSupport>
+    <NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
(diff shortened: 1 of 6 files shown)


src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -47,6 +47,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <UseSystemResourceKeys Condition="'$(UseSystemResourceKeys)' == ''">true</UseSystemResourceKeys>
     <EnableUnsafeUTF7Encoding Condition="'$(EnableUnsafeUTF7Encoding)' == ''">false</EnableUnsafeUTF7Encoding>
     <HttpActivityPropagationSupport Condition="'$(HttpActivityPropagationSupport)' == ''">false</HttpActivityPropagationSupport>
+    <NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
+    <_AggressiveAttributeTrimming Condition="'$(_AggressiveAttributeTrimming)' == ''">true</_AggressiveAttributeTrimming>
```
pranavkm, line 51: Is this "private" because we don't think it's ready for prime time?

pranavkm (approved):

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
eerhardt: Right. It is a bit dangerous removing all these attributes as some code could be reading them, and now behavior has changed. To mitigate this, we are making this setting "private" for Blazor WASM (and Maui if needed) to use to get as small as possible, but we don't expect other app models to use this.

With https://github.com/dotnet/runtime/pull/56475, Blazor WASM devs who are broke by the attribute trimming can now turn it off to get unblocked, at the cost of a slightly larger app.

sbomer (approved):

PR #19439, opened 2021-07-30
Add error for publishing single-file targeting win7

agocke:
Unfortunately we can't produce the same error for `win-*` since that would produce
too many false positives, but we can guard at least this one obvious instance.


VSadov (approved):
LGTM

PR #19452, opened 2021-07-31
Fixed GenerateImplicitNamespaceImports bug.

### LINKED ISSUE #19446: ImplicitNamespaceImports generated codefile generates an invalid import if you try to import as an alias.
Let's say for example I try to import these:

```xml
<Import Include="System" />
<Import Include="System.Collections.Generic" />
<Import Include="System.CommandLine" />
<Import Include="System.CommandLine.Help" />
<Import Include="System.CommandLine.Invocation" />
<Import Include="System.CommandLine.IO" />
<Import Include="System.Globalization" />
<Import Include="System.IO" />
<Import Include="System.Linq" />
<Import Include="System.Reflection" />
<Import Include="System.Text" />
<Import Include="System.Threading.Tasks" />
<Import Include="Elskom.Generic.Libs" />
<Import Include="Microsoft.Diagnostics.NETCore.Client" />
<Import Include="newsmakeResources = Elskom.Generic.Libs.Properties" />
```

The resulting generated ``<project name>.ImplicitNamespaceImports.cs`` file will contain:

```cs
// <autogenerated />
global using global::System;
global using global::System.CommandLine;
global using global::System.CommandLine.Help;
global using global::System.CommandLine.Invocation;
global using global::System.CommandLine.IO;
global using global::System.Globalization;
global using global::System.Reflection;
global using global::System.Text;
global using global::Elskom.Generic.Libs;
global using global::Microsoft.Diagnostics.NETCore.Client;
global using global::newsmakeResources = Elskom.Generic.Libs.Properties;
```

Which results in: ``<project name>/obj/Debug/net6.0/<project name>.ImplicitNamespaceImports.cs(18,40): error CS1525: Invalid expression term '='``

I feel like whereever in the .NET SDK that generates that file that it should actully explicitly look for things like ``=``, ``static``, etc *before* it tries to insert ``global::`` then insert it *after* it finds ``static`` and/or ``=`` so that way the compilation would work.

My .NET SDK version currently is: ``6.0.100-rc.1.21366.58``.

AraHaan:
Fixes https://github.com/dotnet/sdk/issues/19446.


AraHaan:
@ JunTaoLuo if this build fails mind helping me figure out why said changes are wanting to fail?

JunTaoLuo:
This design for this feature needs more discussion before proceeding. The import is intended for namespaces only and not aliases as described in the issue so it's working as designed at the moment.

cc @DamianEdwards who owns the feature.

AraHaan:
I think it's worth this fix as then when .NET 6 does go live most users (including me) will then assume "this better support static usings.", "this better support alias usings.", "this does however support normal usings for sure."

Plus must people even the BCL (in dotnet/runtime) at a lot of places use alias usings as well, if those could become global usings that is then auto generated that can also clean those out from the actual codebase as well too. I think they might also make extensive use for static usings as well, at least dotnet/winforms makes use of it though so having this support both alias's and static usings would also improve the dotnet org repositories as well in terms of the usings in the codebase.

It's very common in dotnet/winforms for this to occur in the codebase in the ``System.Windows.Forms`` project:

```cs
using static Interop.User32;
```

And maybe even others.

AraHaan:
Interesting it works locally on my machine 👀.

AraHaan:
Now that I think back I think the GenerateImplicitNamespaceImports feature that @DamianEdwards owns should become a source generator instead so that way it will become part of the compilation.

Also I will see if I can source dive in more parts of the sdk to find other similar things and see if other things (that generates code in the msbuild tasks) and see what I can do to port those tasks to become source generators instead (like the one that writes the AssemblyAttributes.cs file and AssemblyInfo.cs files).

AraHaan:
Closing this as this has already been fixed in a much better way.

PR #19503, opened 2021-08-04
Update workload list output strings

### LINKED ISSUE #19192: dotnet workload list returns an empty list
I installed Visual Studio 2022 preview 2, including the "Mobile Development with .NET" workload, to try out MAUI using .NET 6 preview 6 - which is shipped with VS2022 preview 2.

I ran "dotnet new maui" at the command prompt, and an error indicated that the template was unknown. That suggested that the maui workload wasn't installed, which made sense as it is still in preview. No problem so far.

I started a command prompt and ran "dotnet workload list" - it returned an empty list - this suggested that no workloads were available. 

This was unexpected behaviour. I don't think the .NET MAUI workload was installed, based on the fact that the template was not available. As a result, I would have expected "dotnet workload list" to return a list of available workloads, including MAUI - but it didn't.

I ran "dotnet workload install maui-desktop" and this showed an error indicating that elevation of privileges was required. I had run the command prompt without running as administrator. Aha, so maybe that was the problem.

I started a new command prompt with elevated privileges and ran "dotnet workload install maui-desktop" and this time the installation worked.

The main reason for raising this issue is to help anyone who gets an empty list when running "dotnet workload list". I couldn't find anyone saying that they had seen this and what to do about it if it happened. My suspicion is that it needs to be run at a command prompt with elevated privileges, but I have seen nobody say that, and it doesn't tell you that itself when you run it. Maybe this should be a warning emitted by the "dotnet workload list" command, if it is true?

I hope, at the least, that the existence of this issue helps other people who encounter this behaviour, even if no fix is considered necessary.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/19192

### DIFF
--- a/src/Tests/dotnet-workload-list.Tests/GivenDotnetWorkloadList.cs
+++ b/src/Tests/dotnet-workload-list.Tests/GivenDotnetWorkloadList.cs
@@ -41,7 +41,7 @@ namespace Microsoft.DotNet.Cli.Workload.List.Tests
             command.Execute();
 
             // Expect 3 lines for table headers
-            _reporter.Lines.Count.Should().Be(3);
+            _reporter.Lines.Count.Should().Be(5);
         }
 
         [Fact]
@@ -68,7 +68,7 @@ namespace Microsoft.DotNet.Cli.Workload.List.Tests
 
             foreach (var workload in expectedWorkloads)
             {
-                _reporter.Lines.Should().Contain(workload.ToString());
+                _reporter.Lines.Select(line => line.Trim()).Should().Contain(workload.ToString());
             }
         }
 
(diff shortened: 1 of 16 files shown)


src/Tests/dotnet-workload-list.Tests/GivenDotnetWorkloadList.cs
```diff
@@ -41,7 +41,7 @@ public void GivenNoWorkloadsAreInstalledListIsEmpty()
             command.Execute();
 
             // Expect 3 lines for table headers
```
joeloff, line 43: Update the comment to 5 or maybe change to something neutral like "Expected number of lines for table headers" in case we change this again

joeloff (approved):

src/Tests/dotnet-workload-list.Tests/GivenDotnetWorkloadList.cs
sfoslund: Good catch, thanks!

Issue #803, opened 2017-02-03
Issue error or warning if BaseIntermediateOutputPath is modified after it's used to calculate MSBuildProjectExtensionsPath


dsplaisted:
Filing this in the SDK repo for now, though we may want to do this in MSBuild itself.

If you try to override the `BaseIntermediateOutputPath` property in a project that uses the `<Project Sdk="Microsoft.NET.Sdk">` syntax, you will be setting the property after it's already been used to calculate the `MSBuildProjectExtensionsPath` property, which is where the auto-generated NuGet .props and .targets files are imported from.  So you won't get any imports from NuGet packages.

The solution to this is to set the `BaseIntermediateOutputPath` in a Directory.Import.props file, which gets evaluated beforehand, or not to use the `Sdk` attribute on the project and instead have a pair of .props and .targets imports in the project to import from the Sdk, and to set the `BaseIntermediateOutputPath` before the .props import.

This is likely to be a common issue people run into, and it will not be obvious what is wrong or how to fix it.  So we should consider detecting that the `BaseIntermediateOutputPath` property was modified after it was used to calculate `MSBuildProjectExtensionsPath`, and generate an error (or at least a warning).

Related issue for us to fix this for the SDK build itself: #802

dsplaisted:
@rainersigwald @AndyGerlicher What do you folks think about adding a warning or error for this in MSBuild compared to doing it in the .NET SDK?

srivatsn:
I'd filed https://github.com/Microsoft/msbuild/issues/1603 in MSBuild about this.

rainersigwald:
I don't see a good way to catch this in common targets alone. The usual (admittedly terrible) approach to this sort of check is to set a shadow property (maybe `_OriginalBaseIntermediateOutputPath`) to the value of the real property at or before the point of first use, and then have a target run to validate that the two properties match. Since the point of first use is in the Sdk, MSBuild can't help there.

In the past, folks have brought up notions like write-once or defaulted properties (I couldn't find an issue for this, so I filed https://github.com/Microsoft/msbuild/issues/1673). That could help here but would be a serious language change.

dsplaisted:
> Since the point of first use is in the Sdk

That doesn't seem to be true.  I think the first point of use is [in Microsoft.Common.props](https://github.com/Microsoft/msbuild/blob/ca7b47d21088068aa7f8055b5637c5eb5ab500db/src/XMakeTasks/Microsoft.Common.props#L46-L48).

[...]

PR #19512, opened 2021-08-04
Append instead of override DefaultExcludesInProjectFolder

EvanCahill:
The definition of `DefaultExcludesInProjectFolder` in Microsoft.NET.Sdk.DefaultItems.targets overrides the
existing definition of the property due to a typo that references `DefaultItemExcludesInProjectFolder` instead
of `DefaultExcludesInProjectFolder`.

This appears to have been introduced in #803 where the property was renamed from 
`DefaultItemExcludesInProjectFolder` to `DefaultExcludesInProjectFolder` but the inner reference was not
updated.


marcpopMSFT:
Fixed in https://github.com/dotnet/sdk/pull/24063

PR #19518, opened 2021-08-04
Fix round tripping of xml file with xml comments.

Anipik:
Currently rewriting the file leaves some extra characters  at the end of the file. This will rewrite the file.

### DIFF
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
@@ -76,7 +76,7 @@ namespace Microsoft.DotNet.Compatibility.ErrorSuppression.Tests
             string filePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName(), "DummyFile.xml");
             engine.WriteSuppressionsToFile(filePath);
 
-            Assert.Equal(engine.suppressionsFile.Trim(), output.Trim(), ignoreCase: true);
+            Assert.Equal(engine.suppressionsFileWithoutComment.Trim(), output.Trim(), ignoreCase: true);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/SuppressionTests/SuppressionEngineTests.cs
```diff
@@ -76,7 +76,7 @@ public void SuppressionEngineSuppressionsRoundTrip()
             string filePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName(), "DummyFile.xml");
             engine.WriteSuppressionsToFile(filePath);
 
-            Assert.Equal(engine.suppressionsFile.Trim(), output.Trim(), ignoreCase: true);
+            Assert.Equal(engine.suppressionsFileWithoutComment.Trim(), output.Trim(), ignoreCase: true);
```
safern, line 79: The comments would be lost when generating the file again, right?

Anipik: yes.

safern: Is there a way to preserve comments or that is just not possible with the xml serializer?

danmoseley: If they have to be XML comments then this sounds like it would do it
https://stackoverflow.com/a/7386148/849053

But programmatically written comments could presumably just be another field written out as a `<comment>` element.

safern: Nice. I just want to preserve comments that people might add to the suppression file? But we could also use a `<comment>` element and we could add notes when generating the suppression file to make the file more readable,  maybe the rule name?

ericstj: I’m not sure we can round trip comments placed at arbitrary places in the document unless we employ a strictly additive rebaseline implementation. Comments could appear anywhere in the file and I don’t think xml serialization supports reading and writing (roundtripping) the comments.  @danmoseley your link is helpful but that’s about writing comments, not round tripping them.

safern (approved):

PR #19545, opened 2021-08-05
Add workload fallback to previous feature band

### LINKED ISSUE #18489: Fall back to preview feature band when there is no current workload manifest matching the current SDK
During development, we want to be able to reversion for a new feature band without forcing an update to the workload.
Additionally, a workload may choose not to make changes and so may not want to be forced to update.

As part of this, we might want to design a system for knowing what workloads we expect so we can deprecate the workload and not load it.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/18489


sfoslund:
@dsplaisted the issue mentions that we want to add the ability to depreciate workloads when we fallback, do you have any thoughts on that implementation? I haven't included it here.

dsplaisted:
The fallback should be per manifest, instead of for the top-level manifest folder.  IE there might be a 6.0.200 version of Microsoft.NET.Workload.Mono.ToolChain and Microsoft.NET.Workload.Emscripten, but not 6.0.200 versions of Microsoft.NET.Sdk.Android or Microsoft.NET.Sdk.iOS.  In that case it should fall back to the 6.0.100 folder for iOS and Android, but not for the Mono toolchain or Emscripten workload manifests.

This also should be based on a list of known manifests that are included in the SDK.  That's the way to deprecate workloads too, if we don't include them in the SDK, then it wouldn't fall back to previous folders for that workload manifest.

We probably want to generate the list of known workload manifests in the installer repo, and put that list in a data file in the SDK directory for the workload resolver logic to pick up.

sfoslund:
@dsplaisted okay I've updated the implementation to fallback per manifest and when a serialized list of known manifests is located at `dotnet/sdk/version/.knownWorkloadIds`, can you take a look?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -113,10 +131,41 @@ public IEnumerable<string> GetManifestDirectories()
...
+                .Where(featureBand => featureBand < Version.Parse(_sdkVersionBand));
+            var matchingManifestFatureBands = candidateFeatureBands
+                .Where(featureBand => Directory.Exists(Path.Combine(_sdkRootPath, "sdk-manifests", featureBand.ToString(), manifestId)));
+            if (matchingManifestFatureBands.Any())
+            {
+                return Path.Combine(_sdkRootPath, "sdk-manifests", matchingManifestFatureBands.Max()!.ToString(), manifestId);
+            }
+            else
+            {
+                throw new Exception(string.Format(Strings.KnownManifestIdDoesNotExist, manifestId));
```
dsplaisted, line 167: I'm not sure whether we should be throwing an exception here.  Right now a workaround for when your manifests are messed up is to just delete the sdk-manifests folder.  At a minimum, I think we should have a way to suppress the error and ignore the missing manifests.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/Strings.resx
```diff
@@ -186,4 +186,7 @@
   <data name="CyclicWorkloadRedirect" xml:space="preserve">
     <value>Cyclic workload redirect '{0}' in manifest '{1}' [{2}]</value>
   </data>
+  <data name="KnownManifestIdDoesNotExist" xml:space="preserve">
+    <value>Known manifest id {0} does not exist.</value>
```
dsplaisted, line 190:
```suggestion
    <value>Workload manifest not found: {0}</value>
```

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -247,6 +247,46 @@ var sdkDirectoryWorkloadManifestProvider
...
+            var manifestDirectory7 = Path.Combine(fakeDotnetRootDirectory, "sdk-manifests", "7.0.100");
+            Directory.CreateDirectory(manifestDirectory7);
+            Directory.CreateDirectory(Path.Combine(manifestDirectory7, "Mock"));
+            File.WriteAllText(Path.Combine(manifestDirectory7, "Mock", "WorkloadManifest.json"), string.Empty);
+
+            var knownWorkloadsFilePath = Path.Combine(fakeDotnetRootDirectory, "sdk", "6.0.100", ".knownWorkloadIds");
+            Directory.CreateDirectory(Path.GetDirectoryName(knownWorkloadsFilePath)!);
+            File.WriteAllText(knownWorkloadsFilePath, "[\"Android\",\"iOS\"]");
+
+            var sdkDirectoryWorkloadManifestProvider
```
dsplaisted, line 282: I think tests for this should cover resolving to manifests that are in the current version as well as previous versions.  So I'd do something like this:

- Have iOS and Android manifests in the 7.0.100 folder
- Have an iOS manifest in the 6.0.100 folder
- Have iOS and Android manifests in the 5.0.100 folder
- Have an Android manifest in the 4.0.100 folder

Each manifest should have different contents depending on the feature band it's in so we can tell which folder it was resolved from.  The test should verify that the iOS manifest was resolved from 6.0.100 and Android was resolved from 5.0.100.

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -266,6 +267,33 @@ public void GivenWorkloadInstallItDoesNotRemoveOldInstallsOnRollback()
             installPacks.Count().Should().Be(3);
         }
 
+        [Fact]
+        public void GivenWorkloadInstallItFallsBackToPriorFeatureBands()
```
dsplaisted, line 271: Is this giving us any extra coverage that's not already going to be covered by the SdkDirectoryWorkloadManifestProviderTests?  I'm not sure whether we need this test or not.

dsplaisted (approved):

sfoslund:
FYI @wli3 since I think you had some concerns about this feature, do you want to review too?

wli3:
checking...

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -113,10 +131,41 @@ public IEnumerable<string> GetManifestDirectories()
...
+
+            return manifestIdsToDirectories.Values;
+        }
+
+        private string FallbackForMissingManifest(string manifestId)
+        {
+            var candidateFeatureBands = Directory.GetDirectories(Path.Combine(_sdkRootPath, "sdk-manifests"))
+                .Select(dir => Path.GetFileName(dir))
+                .Where(featureBand => Version.TryParse(featureBand, out _))
+                .Select(featureBand => Version.Parse(featureBand))
```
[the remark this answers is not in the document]
sfoslund: Yes, I'm using version here to match how we use it on line 40 of this file.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -80,6 +97,7 @@ public IEnumerable<(string manifestId, string? informationalPath, Func<Stream> o
 
         public IEnumerable<string> GetManifestDirectories()
```
wli3, line 88: implementation side, we just need to make sure GetManifestDirectories and GetManifests is the only method used to search for manifest. Or other caller might get the old behavior

sfoslund: Agreed, I think that's the intention regardless of this change.

PR #19548, opened 2021-08-05
Fix the duplicates and self comapare in api compat runner

### LINKED ISSUE #19547: [Package-Validation] Remove Duplicate right assembly tuples and also remove tuples where left == right.
Currently we generating a lot of duplicate right assemblies and also we are running the apicompat between same assemblies(which is completely useless)

cc @safern @ericstj

Anipik:
Fixes https://github.com/dotnet/sdk/issues/19547

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
@@ -18,7 +18,7 @@ namespace Microsoft.DotNet.PackageValidation
     /// </summary>
     public class ApiCompatRunner
     {
-        private Dictionary<MetadataInformation, List<(MetadataInformation rightAssembly, string header)>> _dict = new();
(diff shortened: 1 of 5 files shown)


src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -97,7 +97,8 @@ public void QueueApiCompat(MetadataInformation leftMetadataInfo, MetadataInforma
         {
             if (_dict.ContainsKey(leftMetadataInfo))
             {
-                _dict[leftMetadataInfo].Add((rightMetdataInfo, header));
+                if (!_dict[leftMetadataInfo].Contains((rightMetdataInfo, header)))
+                    _dict[leftMetadataInfo].Add((rightMetdataInfo, header));
```
safern, line 101: We could simplify this to:
```cs
if (_dict.TryGetValue(leftMetadataInfo, out var value)
{
  if (value.Contains(...))
    value.Add(...);
}
```

safern (approved):

PR #19551, opened 2021-08-06
Only allow dlls and executable to be passed to the ReadyToRun infrastructure

davidwrighton:
Protect crossgen2 from attempting to compile non-dll and non-exe binaries
- Shouldn't have an impact on normal scenarios but when we enable mibc presence in the RuntimeList.xml file this will prevent them attempting to be compiled


AntonLapounov (approved):
Looks great, thank you!

PR #19562, opened 2021-08-06
Don't show .NET 6 in VS2019 TargetFramework dropdown

dsplaisted:
Building .NET 6 projects in VS 2019 is not supported.  (For example, you are likely to get errors for global implicit usings, as the version of Roslyn in VS 2019 doesn't fully support global usings).

So this PR hides .NET 6 from the Target Framework dropdown in the project properties in VS 2019.

I tested this manually in VS 2019 and VS 2022.


sfoslund (approved):

KalleOlaviNiemitalo:
Is there a risk that MSBuild is so old that it does not even support VersionGreaterThanOrEquals? Maybe not, if Visual Studio 2017 uses its private copy of .NET Core 2.1 SDK instead of this newer SDK.

dsplaisted:
> Is there a risk that MSBuild is so old that it does not even support VersionGreaterThanOrEquals? Maybe not, if Visual Studio 2017 uses its private copy of .NET Core 2.1 SDK instead of this newer SDK.

No, the 6.0 SDK requires at least VS 16.10, and that is enforced via the SDK resolver, not MSBuild logic.

PR #19570, opened 2021-08-06
Cache results of ResolveTargetingPackAssets across builds

### LINKED ISSUE #19282: Consider caching in ResolveTargetingPackAssets
`ResolveTargetingPackAssets` is called in every build, including incremental builds, because it populates items for later targets. However, that means that it reads its input files for every project in every build, which can add up.

In a no-op incremental command-line build of OrchardCore it takes

```
     3631 ms  ResolveTargetingPackAssets               156 calls
```

in a 50 second build (though spread across multiple processes).

We've also seen this show up in scenarios ([Microsoft-internal link](https://microsoft.sharepoint.com/teams/snappy/_layouts/OneNote.aspx?id=%2Fteams%2Fsnappy%2FSiteAssets%2FVS%20Perf%20Notebook&wd=target%28SnappyJr%20Clusters.one%7C14604092-56E1-4364-9A06-93957270E2A0%2F21-07-21%20-%20Switch%20Build%20Configuration%7CBE54E9CA-34E3-480D-8733-8E444DFED855%2F%29)) as file reads with a high "duplication factor" -- in other words we read `FrameworkList.xml` end to end 30 times or so per MSBuild worker node.

rainersigwald:
This PR uses MSBuild's RegisteredTaskObject machinery to cache (for the lifetime of the process) the results of resolving targeting pack assets. This task runs on every build because it must produce items to populate `@(Reference)`, but its output basically never changes since the targeting packs are static.

The caching is two-layered:

1. Cache the parsing of the targeting pack (`FrameworkList.xml` and friends) so that those files need to be read only once per MSBuild worker node per build.
2. Cache the full results of the task, including deduplication in the case of multiple targeting packs.

Fixes #19282.

I tested this with a no-op incremental build scenario (of https://github.com/OrchardCMS/OrchardCore/commit/ec2154ae099913ee81751f30c4f7660952fe63c9). Worker nodes were reused for the benchmarks.

Filtering to just this task, you can see a significant reduction in total time spent:

image

The overall time is improved but not dramatically (on the order of 2%):

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `S:\benchmarkscripts\orchardcore\Build.bat S:\sdk\sdk-main` | 22.056 ± 1.190 | 20.107 | 23.662 | 1.02 ± 0.09 |
| `S:\benchmarkscripts\orchardcore\Build.bat S:\sdk\cacheharder` | 21.725 ± 1.508 | 19.727 | 24.428 | 1.00 |


src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -46,6 +46,8 @@ public class ResolveTargetingPackAssets : TaskBase
         [Output]
         public ITaskItem[] UsedRuntimeFrameworks { get; set; }
 
+        private static readonly bool s_AllowCacheLookup = System.Environment.GetEnvironmentVariable("DOTNET_SDK_ALLOWTARGETINGPACKCACHING") != "0";
```
rainersigwald, line 49: It doesn't look like there are a lot of SDK escape hatches like this. Does this look ok?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -208,10 +266,37 @@ private void AddNetStandardTargetingPackAssets(ITaskItem targetingPack, string t
...
+
+            string frameworkListKey = hashKey.ToString();
+
+            if (s_AllowCacheLookup &&
+                BuildEngine4.GetRegisteredTaskObject(
+                  frameworkListKey, 
+                  RegisteredTaskObjectLifetime.AppDomain /* TODO: should this be build? */) 
+                is FrameworkListCacheEntry cacheEntry)
+            {
+                // TODO: probe files to see if they've been modified? or just assume not
```
rainersigwald, line 289: The meta-question here is: is there any supported scenario where the contents of the targeting pack will change? Or can we cache it forever?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -208,10 +266,37 @@ private void AddNetStandardTargetingPackAssets(ITaskItem targetingPack, string t
...
+
+                // TODO: is this safe? do we need to clone them or something?
```
rainersigwald, line 291: This is on me and I now suspect the answer is "it's fine".

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -61,6 +63,47 @@ protected override void ExecuteCore()
...
+
+            if (s_AllowCacheLookup &&
+                BuildEngine4?.GetRegisteredTaskObject(
+                    hashKey,
+                    RegisteredTaskObjectLifetime.AppDomain /* TODO: should this be build? */)
+                  is ResolvedAssetsCacheEntry cacheEntry)
+            {
+                // TODO: probe files to see if they've been modified? or just assume not
+
+                // TODO: is this safe? do we need to clone them or something?
```
rainersigwald, line 94: A task can safely cache `TaskItem`s as far as I can tell: dotnet/msbuild#6754

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
rainersigwald: Discussed this with @dsplaisted offline and we decided "process lifetime" is probably ok, as we don't expect targeting packs to change.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -208,13 +267,43 @@ private void AddNetStandardTargetingPackAssets(ITaskItem targetingPack, string t
             string targetingPackDllFolder,
             ITaskItem targetingPack, List<TaskItem> referenceItems, List<TaskItem> analyzerItems)
         {
-            XDocument frameworkListDoc = XDocument.Load(frameworkListPath);
-
             string profile = targetingPack.GetMetadata("Profile");
 
+            StringBuilder hashKey = new(nameof(ResolveTargetingPackAssets) + nameof(AddItemsFromFrameworkList));
```
dsplaisted: Same feedback here about making sure we're caching all the inputs and adding test coverage for that if possible.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -61,6 +63,47 @@ protected override void ExecuteCore()
...
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.ItemSpec);
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.Path));
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata("TargetingPackFormat"));
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata("TargetFramework"));
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.NuGetPackageId));
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.NuGetPackageVersion));
+                hashKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.PackageConflictPreferredPackages));
+            }
+
+            string hashKey = hashKeyBuilder.ToString();
```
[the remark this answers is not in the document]
rainersigwald: That was my inclination as well but I can turn it off if anyone feels strongly.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -61,6 +63,47 @@ protected override void ExecuteCore()
 
             var resolvedTargetingPacks = ResolvedTargetingPacks.ToDictionary(item => item.ItemSpec, StringComparer.OrdinalIgnoreCase);
 
+            StringBuilder hashKeyBuilder = new(nameof(ResolveTargetingPackAssets) + nameof(ExecuteCore));
+            hashKeyBuilder.AppendLine();
```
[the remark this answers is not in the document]
rainersigwald: That test is slick and I love it. I'll adopt one much like it.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
rainersigwald: This one is harder since it's internal but I'll see what I can do.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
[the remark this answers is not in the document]
rainersigwald: Done.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
rainersigwald: It was indeed harder but a bit of refactoring made it work ok. Look good now?

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/ResolveTargetingPackAssetsTests.cs
```diff
@@ -12,15 +16,82 @@ public class ResolveTargetingPackAssetsTests
...
+
+            task.Execute().Should().BeTrue();
+
+            ((MockBuildEngine4)task.BuildEngine).RegisteredTaskObjectsQueries.Should().Be(2,
+                because: "there should be a lookup for the overall and the specific targeting pack");
+
+            ((MockBuildEngine4)task.BuildEngine).RegisteredTaskObjects.Count.Should().Be(2,
+                because: "there should be a cache entry for the overall lookup and for the specific targeting pack");
+
+            task.Execute().Should().BeTrue();
```
dsplaisted, line 49: Should we be creating a new task instance here instead of reusing the existing one?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -162,6 +190,69 @@ protected override void ExecuteCore()
...
+            cacheKeyBuilder.AppendLine(nameof(ResolvedTargetingPacks));
+            foreach (var resolvedTargetingPack in ResolvedTargetingPacks)
+            {
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.ItemSpec);
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.Path));
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata("TargetingPackFormat"));
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata("TargetFramework"));
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.NuGetPackageId));
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.NuGetPackageVersion));
+                cacheKeyBuilder.AppendLine(resolvedTargetingPack.GetMetadata(MetadataKeys.PackageConflictPreferredPackages));
```
dsplaisted, line 238: Ideally the tests would verify that the key depends on any metadata items that are read.  The way to do that would probably be to use a strongly-typed wrapper for all metadata access.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -208,13 +299,34 @@ private void AddNetStandardTargetingPackAssets(ITaskItem targetingPack, string t
             string targetingPackDllFolder,
             ITaskItem targetingPack, List<TaskItem> referenceItems, List<TaskItem> analyzerItems)
         {
```
dsplaisted, line 301: Can this be a static method that takes a `FrameworkListDefinition` as an argument?  Then it won't be able to accidentally access data that's not included in the key.

dsplaisted (approved):
Looks good.  I think it would be nice to also cover the item metadata in the tests.

rainersigwald:
Test failure was completely legit: https://github.com/dotnet/sdk/compare/4d159ab1d7e8d0560509e8176f1127eac299de3d..8fb8739032f205dc88cfdb66bb95f828ef84e17a

I will add a smaller UT for this task to catch that too.

rainersigwald:
Updated individual-task results (OrchardCore, 24 build nodes):

image

Overall time is much noisier. I suspect my initial 1-2% improvement was noise there. But I can't see a reason why speeding up this task would _increase_ overall time.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs
```diff
@@ -0,0 +1,424 @@
...
+            buildEngine.RegisteredTaskObjectsQueries.Should().Be(4,
+                because: "there should be no hits on the overall or targeting pack lookup this time through");
+
+            buildEngine.RegisteredTaskObjects.Count.Should().Be(4,
+                because: "there should be distinct results for C# and F#");
+        }
+
+        private ResolveTargetingPackAssets InitializeTask(out string mockPackageDirectory)
+        {
+            mockPackageDirectory = Path.Combine(Path.GetTempPath(), "dotnetSdkTests", Path.GetRandomFileName());
```
dsplaisted, line 118: I'm not a fan of putting test state in temporary directories that don't get cleaned up and are hard to find for debugging.  Can we make this test class derive from `SdkTest` and then use `_testAssetsManager.CreateTestDirectory()` or something?

Most of the task unit tests don't derive from `SdkTest`, but `GivenAResolveRuntimePackAssetsTask` does, so it should be OK.

dsplaisted (approved):
Great job, this is really cool!

AraHaan:
Anyone able to test if this can also speed up say “compiling roslyn” itself or say compiling everything in dotnet/runtime locally.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs
rainersigwald: This was preexisting but sure.

rainersigwald: Is 289ee935ca what you were thinking of?

rainersigwald:
> Anyone able to test if this can also speed up say “compiling roslyn” itself or say compiling everything in dotnet/runtime locally.

@AraHaan I'd expect it to speed it up but not significantly at those scales. This task isn't a huge proportion of build time, just easy to fix.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs
dsplaisted: Pretty much, but I think you'll need to update it to use a different directory each time it's called as part of the same test.  We generally do this with an `identifier` parameter that you can pass through to CreateTestDirectory.

When running in CI, we will fail if it tries to create a test directory that already exists.  This helps prevent flakiness where different tests end up using the same test directory.  We don't do that locally, since otherwise every time you re-ran a test it would fail.

rainersigwald: Done in 5f64f82fa9 but right now no test calls it twice so I didn't actually call it. Do you want me to pass it in in every call anyway?

dsplaisted: OK, I got confused by the different overloads of `InitializeTask`: https://github.com/dotnet/sdk/blob/5f64f82fa9faf86654b837be4c0b695e64f4caf2/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs#L74-L87

If no test actually calls it twice then you don't even need to take the identifier as an argument, I just thought the tests were re-creating the same directory and that was going to cause failures in CI.

rainersigwald: Pushed a revert + some strategic renames to hopefully clarify for the future.

dsplaisted: Thanks.  I already signed off so feel free to merge.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs
```diff
@@ -125,10 +125,9 @@ public void Given_Passing_ResolvedTargetingPacks_A_Different_Language_Parses_Aga
         }
 
         private ResolveTargetingPackAssets InitializeTask(out string mockPackageDirectory,
-            [CallerMemberName] string testName = nameof(GivenAResolvePackageAssetsTask),
-            string identifier = null)
```
AraHaan: Why did this get reverted?

rainersigwald: It's not needed today: https://github.com/dotnet/sdk/pull/19570#discussion_r703831303

PR #19571, opened 2021-08-06
Fix flaky parser tests

### LINKED ISSUE #19544: More tests flaky due to same issue with distributed logger on the end of the command
https://github.com/dotnet/sdk/runs/3252226972
https://github.com/dotnet/sdk/pull/19528

eg 
```
Error message
Expected string to be \n"-maxcpucount -verbosity:m -restore -target:Publish <project> <extra-args>" with a length of 73, but \n"-maxcpucount -verbosity:m -restore -target:Publish <project> <extra-args> -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,/datadisks/disk1/workspace/_work/1/s/artifacts/bin/redist/Release/dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,/datadisks/disk1/workspace/_work/1/s/artifacts/bin/redist/Release/dotnet.dll" has a length of 346.
```

Note that these passed on rerun. It would be nice to know why the parameter is getting sporadically added. Seems worth investigating as a bug?

Note there's quite a few eg 
all of Microsoft.DotNet.Cli.MSBuild.Tests.GivenDotnetPublishInvocation.MsbuildInvocationIsCorrect
MsbuildInvocationIsCorrectForSeparateRestore
CommandAcceptsMultipleCustomProperties
MsbuildInvocationIsCorrectForNoBuild

@sfoslund @dotnet/domestic-cat

sfoslund:
Some parser tests that check what arguments are being passed to MSBuild are failing because an extra parameter along the lines of `-distributedlogger:Microsoft.DotNet....` is being added occasionally. This argument comes from telemetry logic and can be removed by setting the current session id to null in tests: https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs#L26

Fixes https://github.com/dotnet/sdk/issues/19544

### DIFF
--- a/src/Cli/dotnet/CommandDirectoryContext.cs
+++ b/src/Cli/dotnet/CommandDirectoryContext.cs
@@ -33,6 +33,7 @@ namespace Microsoft.DotNet.Cli
                     $"Calls to {nameof(CommandDirectoryContext)}.{nameof(PerformActionWithBasePath)} cannot be nested.");
             }
             _basePath = basePath;
+            Telemetry.Telemetry.CurrentSessionId = null;
             try
             {
                 action();


joeloff (approved):

joeloff:
We saw this yaml file failure earlier today and then it magically worked when we reran another SDK build

PR #19596, opened 2021-08-09
Runtime pack resolution

### LINKED ISSUE #14044: Allow runtime packs to be resolved from the packs folder
For targeting packs, the SDK first checks if they are in the dotnet packs folder.  If they are it uses them from there, otherwise it generates a `PackageDownload` item so that they will be downloaded to the packages folder as part of NuGet restore.

Currently, we don't check for runtime packs in the packs folder.  This is because we haven't shipped any with the .NET SDK, as they are only used for self-contained apps and there are many different runtime packs for different RuntimeIdentifiers which represent different OS/architecture combinations.

For .NET 6, iOS and Android apps will need runtime packs, and the set of target RuntimeIdentifiers will be more limited.  So they would like to deliver runtime packs as part of their optional SDK workloads.

This means we should update the SDK runtime pack resolution logic to first look in the dotnet packs folder and resolve from there if possible, before generating a `PackageDownload` item to download the runtime pack if necessary.

### LINKED ISSUE #18064: Resolve runtime pack versions based on workload manifest
For workloads that ship with runtime packs, we want to resolve the runtime pack that ships with that workload.  One way to do this is to have the workload include KnownRuntimePack items in its targets that specify the matching version of the runtime pack.  However, it would be nicer if the ProcessFrameworkReference task (or similar) could automatically choose the right runtime pack version by reading it from the workload manifest.

dsplaisted:
Fixes #14044
Fixes #18064

- Resolves runtime packs from packs directory if present
- Supports using DOTNETSDK_WORKLOAD_PACK_ROOTS environment variable to set additional packs folders for targeting and runtime packs
  - The packs should be under a `packs` subfolder of the directory specified by the environment variable
- Supports resolving targeting or runtime pack versions from workload manifest
  - To use this, the version in the `KnownFrameworkReference` or `KnownRuntimePack` item should be set to `**FromWorkload**`

### DIFF
--- a/src/Tasks/Common/Resources/xlf/Strings.ja.xlf~RF1ea49cc.TMP
+++ b/src/Tasks/Common/Resources/xlf/Strings.ja.xlf~RF1ea49cc.TMP
@@ -1,40 +1,40 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
-  <file datatype="xml" source-language="en" target-language="pt-BR" original="../Strings.resx">
+  <file datatype="xml" source-language="en" target-language="ja" original="../Strings.resx">
     <body>
       <trans-unit id="AddResourceWithNonIntegerResource">
         <source>NETSDK1076: AddResource can only be used with integer resource types.</source>
-        <target state="translated">NETSDK1076: AddResource só pode ser usado com os tipos de recursos inteiros.</target>
+        <target state="translated">NETSDK1076: AddResource は、整数のリソースの種類でのみ使用できます。</target>
         <note>{StrBegin="NETSDK1076: "}</note>
       </trans-unit>
       <trans-unit id="AppConfigRequiresRootConfiguration">
         <source>NETSDK1070: The application configuration file must have root configuration element.</source>
-        <target state="translated">NETSDK1070: o arquivo de configuração do aplicativo deve ter um elemento de configuração raiz.</target>
+        <target state="translated">NETSDK1070: アプリケーション構成ファイルには、ルート構成要素が必要です。</target>
         <note>{StrBegin="NETSDK1070: "}</note>
       </trans-unit>
       <trans-unit id="AppHostCreationFailedWithRetry">
         <source>NETSDK1113: Failed to create apphost (attempt {0} out of {1}): {2}</source>
-        <target state="translated">NETSDK1113: falha ao criar appHost (tentativa {0} de {1}): {2}</target>
+        <target state="translated">NETSDK1113: apphost を作成できませんでした ({1} 回中 {0} 回目の試行): {2}</target>
         <note>{StrBegin="NETSDK1113: "}</note>
       </trans-unit>
       <trans-unit id="AppHostCustomizationRequiresWindowsHostWarning">
         <source>NETSDK1074: The application host executable will not be customized because adding resources requires that the build be performed on Windows (excluding Nano Server).</source>
-        <target state="translated">NETSDK1074: o host do aplicativo executável não será personalizado porque a adição de recursos requer que o build seja executado no Windows (excluindo Nano Server).</target>
+        <target state="translated">NETSDK1074: リソースの追加ではビルドが Windows 上で実行される必要があるため、アプリケーション ホストの実行可能ファイルはカスタマイズされません (Nano Server を除く)。</target>
         <note>{StrBegin="NETSDK1074: "}</note>
       </trans-unit>
       <trans-unit id="AppHostHasBeenModified">
         <source>NETSDK1029: Unable to use '{0}' as application host executable as it does not contain the expected placeholder byte sequence '{1}' that would mark where the application name would be written.</source>
-        <target state="translated">NETSDK1029: Não é possível usar '{0}' como executável do host de aplicativo, porque ele não contém a sequência de bytes de espaço reservado esperada '{1}' que marcaria o local em que o nome do aplicativo seria gravado.</target>
+        <target state="translated">NETSDK1029: '{0}' は、本来アプリケーション名が書き込まれる場所を示す、必要なプレースホルダー バイト シーケンス '{1}' が含まれていないため、実行可能アプリケーション ホストとして使用できません。</target>
         <note>{StrBegin="NETSDK1029: "}</note>
       </trans-unit>
       <trans-unit id="AppHostNotWindows">
         <source>NETSDK1078: Unable to use '{0}' as application host executable because it's not a Windows PE file.</source>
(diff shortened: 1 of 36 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -609,6 +614,36 @@ private bool AddCrossgen2Package(Version normalizedTargetFrameworkVersion, List<
...
+                    {
+                        yield return Path.Combine(packRoot, "packs");
+                    }
+                }
+            }
+
+            foreach (var packFolder in GetPackFolders())
+            {
+                string packPath = Path.Combine(packFolder, packName, packVersion);
+                if (Directory.Exists(packPath))
```
joeloff, line 638: Do we care if there are old roots that no longer exist? Right now we're skipping over them. Would users care is perhaps a better question

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -609,6 +614,36 @@ private bool AddCrossgen2Package(Version normalizedTargetFrameworkVersion, List<
...
 
+        private string GetPackPath(string packName, string packVersion)
+        {
+            IEnumerable<string> GetPackFolders()
+            {
+                if (!string.IsNullOrEmpty(TargetingPackRoot))
+                {
+                    yield return TargetingPackRoot;
+                }
+                var packRootEnvironmentVariable = Environment.GetEnvironmentVariable("DOTNETSDK_WORKLOAD_PACK_ROOTS");
```
joeloff, line 625: Do we have a central class where we can keep all these environmental variable constants? That way we don't need to worry about typo's in the code.

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
dsplaisted: This is more of a test hook than a user-facing feature, so I don't think we care right now.

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
joeloff: Thank you. I think this will save us a lot of potential pain in the future

joeloff (approved):

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
dsplaisted: I've added one now (`EnvironmentVariableNames`) and added constants for the workload environment variables.  There are a bunch of other CLI environment variables which I didn't refactor into this class yet.

dsplaisted:
@joeloff I've updated this PR to apply your feedback and also to now address #18064.  So a re-review would be good :-)

joeloff:
> @joeloff I've updated this PR to apply your feedback and also to now address #18064. So a re-review would be good :-)

I'll take another pass tonight

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -609,6 +630,61 @@ private bool AddCrossgen2Package(Version normalizedTargetFrameworkVersion, List<
...
+
+            return null;
+        }
+
+        SdkDirectoryWorkloadManifestProvider _workloadManifestProvider;
+        WorkloadResolver _workloadResolver;
+
+        private string GetResolvedPackVersion(string packID, string packVersion)
+        {
+            if (!packVersion.Equals("**FromWorload**", StringComparison.OrdinalIgnoreCase))
```
joeloff, line 668: Should be "FromWorkload"?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -152,5 +154,197 @@ public void It_succeeds_when_RuntimeIdentifier_and_PlatformTarget_mismatch_but_P
...
+            {
+                var itemGroup = XElement.Parse(@"
+  <ItemGroup>
+    <KnownFrameworkReference Include='Microsoft.NETCore.App.Test'
+                          TargetFramework='net6.0'
+                          RuntimeFrameworkName='Microsoft.NETCore.App.Test'
+                          DefaultRuntimeFrameworkVersion='**FromWorload**'
+                          LatestRuntimeFrameworkVersion='**FromWorload**'
+                          TargetingPackName='Microsoft.NETCore.App.Test.Ref'
+                          TargetingPackVersion='**FromWorload**'
```
joeloff, line 275: Looks like a typo - should be FromWorkload?

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
dsplaisted: Oops!  Thanks for catching that :-)

joeloff:
I think this looks good.

src/Tasks/Common/Resources/xlf/Strings.ja.xlf~RF1ea49cc.TMP
```diff
@@ -0,0 +1,914 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
KalleOlaviNiemitalo, line 1: src/Tasks/Common/Resources/xlf/Strings.ja.xlf~RF1ea49cc.TMP looks like a name for a temporary file; did you add this by accident?

There is the older Strings.es.xlf~RF13d378ed.TMP as well.

akoeplinger: Yeah this looks like a mistake, filed https://github.com/dotnet/sdk/pull/19688 to delete them.

dsplaisted: Thanks for catching that.  I've submitted #19694 to fix.

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -609,6 +630,61 @@ private bool AddCrossgen2Package(Version normalizedTargetFrameworkVersion, List<
...
+            if (_workloadManifestProvider == null)
+            {
+                _workloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(NetCoreRoot, NETCoreSdkVersion);
+                _workloadResolver = WorkloadResolver.Create(_workloadManifestProvider, NetCoreRoot, NETCoreSdkVersion);
+            }
+
+            var packInfo = _workloadResolver.TryGetPackInfo(new WorkloadPackId(packID));
+            if (packInfo == null)
+            {
+                Log.LogError("NETSDKZZZZ: Error getting pack version: Pack '{0}' was not present in workload manifests.", packID);
```
mhutch, line 682: Placeholder error code?

Issue #19521, opened 2021-08-05
Update implicit global usings feature to address issues


DamianEdwards:
The new implicit global usings feature added to the SDK in 6.0.100-preview.7 has caused a few issues in some important scenarios such that some changes will be made to mitigate them and generally improve its usability.

These changes are targeting 6.0.100-rc.1

Items:
- [x] Rename the MSBuild property from `<DisableImplicitNamespaceImports>` to `<ImplicitUsings>`
   - **Note this is for C# only**
  - Visual Basic will continue to use the `<DisableImplicitNamespaceImports>` property for disabling implicit namespace imports
- [x] Add the `<ImplicitUsings>` property to the schema file so it appears in statement completion
- [x] Change the defined conditions to only enable adding the implicit usings from the SDK when the `<ImplictUsings>` property is set to "true" or "enable"
  - The existing conditions based on TFM should be removed
  - The only condition for adding the implicit usings as defined in the SDK will be the check that `<ImplicitUsings>` is equal to "true" or "enable"
- [x] Rename the item type that specifies the global namespaces to emit to the generated .cs file from `<Import>` to `<Using>`
  - **Note this is for C# only**
  - Visual Basic will continue to use the `<Import>` item type for specifying namespaces to import
- [x] dotnet/msbuild#6745
- [x] Ensure the generated file is still emitted if any `<Using>` items are declared (even if `<ImplicitUsings>` is false)
- [x] Ensure that implicit `<Import>` items for Visual Basic are not changed from what they were in .NET 5
- [x] Update processing of `<Using>` items to support defining `using [alias]` and `using static [type]` as well as `using [namespace]`, e.g.:
  - `<Using Include="Microsoft.AspNetCore.Http.Results" Alias="Results" />`
    emits `global using Results = global::Microsoft.AspNetCore.Http.Results;`
  - `<Using Include="Microsoft.AspNetCore.Http.Results" Static="True" />`
    emits `global using static global::Microsoft.AspNetCore.Http.Results;`
  - `<Using Include="Microsoft.AspNetCore.Http" />`
    emits `global using global::Microsoft.AspNetCore.Http;`
- [x] Update the C# project templates to enable implicit usings for new .NET 6 projects, i.e. include `<ImplicitUsings>enable</ImplicitUsings>` in the `.csproj` file
  - _This will be tracked by issues in the relevant repos for each template_
  - [x] dotnet/aspnetcore#35131
  - [x] dotnet/templating#3619
[...]

PR #18459, opened 2021-06-23
Add global usings for .NET projects (.NET SDK, Web and Worker)

JunTaoLuo:
Fixes https://github.com/dotnet/aspnetcore/issues/32451.

TODOs:
- [x] Verify that preview condition should be removed
  - [x] Ensure SDK is updated to support C# 10
- [x] Add test
- [x] Review additional usings:
  - System.ComponentModel.DataAnnotations
  - Microsoft.Extensions.Caching.Memory
  - Microsoft.AspNetCore.Mvc\
- [x] Checkin with VB expert
- [x] Perf numbers using sample minimal API Empty Web project: https://github.com/JunTaoLuo/GlobalUsings/tree/main/App

|   | Mean | Median | Std. Dev. |
| ------------- | ------------- | ------------- | ------------- |
| Before (usings in Program.cs)  | 2253.425ms | 2266.458ms | 107.8083ms |
| After (usings in generated App.ImplicitNamespaceImports.cs):  | 2221.974ms | 2214.453ms | 104.5122ms |

Followups:

- Underscore in naming: e.g. DisableImplicitNamespaceImports_Web
- Inclusion of System.Net.Http in the base SDK
- Relocation of VB Imports to props files
- Breaking change announcements
  - dotnet/Docs
  - blog post

PR #19599, opened 2021-08-09
Update implicit global usings feature

pranavkm:
* Undo changes to VB projects
* For C#, projects use the `Using` itemgroup to generate global usings. Support aliasing and static imports with attributes.
* Use ImplicitUsings to determine if SDK-specific namespaces are imported by default.

Contributes to https://github.com/dotnet/sdk/issues/19521


DamianEdwards:
Can we get the schema files updated as part of this too please?

pranavkm:
> Can we get the schema files updated as part of this too please?

Is it in this repo?

DamianEdwards:
@pranavkm ah I think you're right and that it's over at https://github.com/dotnet/msbuild/blob/main/src/MSBuild/MSBuild/Microsoft.Build.CommonTypes.xsd

DamianEdwards:
@pranavkm logged dotnet/msbuild#6745 to track the schema updates

DamianEdwards:
@pranavkm when you merge this please change it to "Contributes to #19521" as it doesn't fix everything in that issue yet.

AraHaan:
In this case cant all projects generate usings for ``System``, ``System.Collections.Generic``, ``System.Threading.Tasks``, ``System.Linq``, etc by default?

There might be a few others as well that even my projects use commonly as well too (exposed in https://github.com/Elskom/Sdk/pull/248/).

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
```diff
@@ -0,0 +1,125 @@
...
+                if (@using.Static)
+                {
+                    builder.Append("static ");
+                }
+
+                if (!string.IsNullOrEmpty(@using.Alias))
+                {
+                    builder.Append(@using.Alias)
+                        .Append(" = ");
+                }
```
dsplaisted, line 47: What if a `Using` item has both `Static` and `Alias` metadata.  This will generate C# code that won't compile, right?  Should we error here in that case?

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+
+            public bool Equals(UsingInfo x, UsingInfo y)
+            {
+                return Compare(x, y) == 0;
+            }
+
+            public int GetHashCode(UsingInfo obj)
+            {
+                return StringComparer.Ordinal.GetHashCode(obj);
```
dsplaisted, line 121: Shouldn't this be hashing the namespace string instead of the object itself, which is likely to be either slow or incorrect?

```suggestion
                return StringComparer.Ordinal.GetHashCode(obj.Namespace);
```


Ideally it would use `HashCode.Combine` to include the other properties, but this code needs to run on .NET Framework which doesn't have that.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
```diff
@@ -10,6 +10,30 @@ Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework'">
```
dsplaisted, line 13: Can we move the VB `Import` definitions to `Microsoft.NET.Sdk.VisualBasic.props` instead of the targets file?  This will make it possible to override / remove these in the project file, and will probably make the project properties page that manages these work better too.

/cc @KathleenDollard

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_DotNet.cs
```diff
@@ -64,17 +65,18 @@ public void It_can_disable_dotnet_imports()
 
             var outputDirectory = buildCommand.GetIntermediateDirectory(tfm);
 
-            outputDirectory.Should().NotHaveFile(importFileName);
+            outputDirectory.Should().NotHaveFile(globalUsingsFileName);
         }
 
         [RequiresMSBuildVersionFact("17.0.0.32901")]
         public void It_can_remove_specific_imports_in_project_file()
```
dsplaisted, line 72:
```suggestion
        public void It_can_remove_implicit_usings_in_project_file()
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_DotNet.cs
```diff
@@ -103,10 +105,21 @@ public void It_can_generate_custom_imports()
         {
```
dsplaisted, line 105: Test name update suggestion: `It_can_generate_custom_global_usings`

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_DotNet.cs
```diff
@@ -103,10 +105,21 @@ public void It_can_generate_custom_imports()
...
+                var itemGroup = new XElement(ns + "ItemGroup");
+                projectXml.Root.Add(XElement.Parse(
+@"<ItemGroup>
+    <Using Include=""CustomNamespace"" />
+    <Using Include=""TestStaticNamespace"" Static=""True"" />
+    <Using Include=""System.Biology"" Alias=""AppliedChemistry"" />
+</ItemGroup>"));
+            });
+
+            testProject.AdditionalItems["Using"] = new Dictionary<string, string> { ["Include"] = "CustomNamespace" };
```
dsplaisted, line 120: FYI: #19596 changes the API for adding additional items to a TestProject, so if that PR merges first this will need to be updated.

Also, it looks like you're including a `Using` for `CustomNamespace` twice, is that intentional?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_DotNet.cs
```diff
@@ -155,28 +189,8 @@ public void It_can_disable_import_generation()
         {
```
dsplaisted, line 189: Do we need this test at all anymore?  Is it now just testing the same thing as `Implicit_Usings_Are_Not_Enabled_By_Default`?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_WebApp.cs
```diff
@@ -12,18 +12,19 @@
...
 {
-    public class GivenThatWeWantToGenerateImplicitNamespaceImports_WebApp : SdkTest
+    public class GivenThatWeWantToGenerateGlobalUsings_WebApp : SdkTest
     {
 
-        public GivenThatWeWantToGenerateImplicitNamespaceImports_WebApp(ITestOutputHelper log) : base(log) { }
+        public GivenThatWeWantToGenerateGlobalUsings_WebApp(ITestOutputHelper log) : base(log) { }
 
         [RequiresMSBuildVersionFact("17.0.0.32901")]
         public void It_generates_web_imports_and_builds_successfully()
```
dsplaisted, line 21:
```suggestion
        public void It_generates_web_implicit_usings_and_builds_successfully()
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_WebApp.cs
```diff
@@ -62,9 +62,9 @@ public void It_can_disable_web_imports()
         {
```
dsplaisted, line 62: Nit: This test should be renamed

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_Worker.cs
```diff
@@ -13,17 +13,18 @@
...
 namespace Microsoft.NET.Build.Tests
 {
-    public class GivenThatWeWantToGenerateImplicitNamespaceImports_Worker : SdkTest
+    public class GivenThatWeWantToGenerateGlobalUsings_Worker : SdkTest
     {
-        public GivenThatWeWantToGenerateImplicitNamespaceImports_Worker(ITestOutputHelper log) : base(log) { }
+        public GivenThatWeWantToGenerateGlobalUsings_Worker(ITestOutputHelper log) : base(log) { }
 
         [RequiresMSBuildVersionFact("17.0.0.32901")]
         public void It_generates_worker_imports_and_builds_successfully()
```
dsplaisted, line 21:
```suggestion
        public void It_generates_worker_implicit_usings_and_builds_successfully()
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_Worker.cs
```diff
@@ -56,9 +57,9 @@ public void It_can_disable_worker_imports()
         {
```
dsplaisted, line 57: Nit: rename this test.

src/BlazorWasmSdk/Sdk/Sdk.props
```diff
@@ -20,11 +20,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Import Project="$(_BlazorWebAssemblyPropsFile)" />
 
-  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports_BlazorWasm)' != 'true'
-      And '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
-      And $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '6.0'))">
-    <Import Include="Microsoft.Extensions.Configuration" />
-    <Import Include="Microsoft.Extensions.DependencyInjection" />
-    <Import Include="Microsoft.Extensions.Logging" />
+  <ItemGroup Condition="'$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable'">
```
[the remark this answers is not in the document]
DamianEdwards: I put it in the spec for three reasons, one of them was to align with `<Nullable>`, another was to better match the naming (boolean properties typically start with a verb like `Enable` or `Disable`), the other was to leave space for expanding the supported values in the future, as was suggested during the review meeting last week, so the feature could support more than just a simple, on/off.

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
DamianEdwards: Yeah we should error as `global using static Foo = global::My.Namespace.Value` isn't valid AFAIK.

src/BlazorWasmSdk/Sdk/Sdk.props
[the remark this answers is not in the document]
davidfowl: Not to scope creep but there's a potential future with being able to opt into a particular set of default usings:

```xml
<ImplicitUsings>web</ImplicitUsings>
<ImplicitUsings>console</ImplicitUsings>
<ImplicitUsings>wpf</ImplicitUsings>
```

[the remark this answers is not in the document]
pranavkm: > Not to scope creep

I think it's much easier to remove a bunch of usings from the itemgroup than try and figure out what set of usings each of these names gets you.

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
pranavkm: My plan was to let the compiler complain (it produces an error saying that you cannot alias a static using). It felt less complicated to do this in case a newer langversion started supporting this in the future.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
pranavkm: Should we do that separately? That way if figure out there are issues with doing it, we can revert it independent of this change.

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
```diff
@@ -0,0 +1,125 @@
...
+                }
+
+                if (!string.IsNullOrEmpty(@using.Alias))
+                {
+                    builder.Append(@using.Alias)
+                        .Append(" = ");
+                }
+
+                builder.Append("global::")
+                    .Append(@using.Namespace)
```
KalleOlaviNiemitalo, line 50: If the namespace name contains C# reserved keywords, then is the `Using` item expected to contain the @ escapes already (thus making it specific to C#), or should the task add them?

src/BlazorWasmSdk/Sdk/Sdk.props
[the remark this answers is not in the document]
DamianEdwards: Perhaps, but consider then instead the values representing where they're declared:
```
<ImplicitUsings>sdk</ImplicitUsings>
<ImplicitUsings>packages</ImplicitUsings>
<ImplicitUsings>project</ImplicitUsings>
<ImplicitUsings>other</ImplicitUsings>
```

Point is, there are possibilities here, and the other reasonings still stand IMO.

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
pranavkm: I want to treat the `Include` as an opaque C# string. I really don't want to add language rules to the SDK

src/Tasks/Microsoft.NET.Build.Tasks/GenerateGlobalUsings.cs
```diff
@@ -0,0 +1,125 @@
...
+            var outputFilePath = Path.GetFullPath(OutputFile.ItemSpec);
+            if (File.Exists(outputFilePath))
+            {
+                var existing = File.ReadAllText(outputFilePath);
+
+                if (string.Equals(existing, contentToWrite, StringComparison.Ordinal))
+                {
+                     Log.LogMessage(MessageImportance.Low, Strings.SkippingUnchangedFile, OutputFile.ItemSpec);
+                    return;
+                }
```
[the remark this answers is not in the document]
AraHaan: I tried that before on my failed PR and it would result in test failures, I also tried consuming that task directly in a task like this and it would pass when ran on it's own on the specific test assembly for this dll, but then fail on other tests or at least the same tests in CI vs when ran locally.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1134,7 +1133,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.DisableStandardFrameworkResolution.targets" Condition="'$(DisableStandardFrameworkResolution)' == 'true'" />
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.DesignerSupport.targets" />
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.GenerateAssemblyInfo.targets" Condition="'$(UsingNETSdkDefaults)' == 'true'"/>
-  <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.GenerateImplicitNamespaceImports.targets" Condition="'$(DisableImplicitNamespaceImports)' != 'true' and '$(Language)' == 'C#'"/>
+  <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.GenerateGlobalUsings.targets" Condition="'$(Language)' == 'C#'"/>
```
AraHaan, line 1136: Why not import this in the C# file instead so you do not have to check the ``$(Language)``?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_DotNet.cs
```diff
@@ -13,17 +14,18 @@
...
 {
-    public class GivenThatWeWantToGenerateImplicitNamespaceImports_DotNet : SdkTest
+    public class GivenThatWeWantToGenerateImpliciUsings_DotNet : SdkTest
     {
-        public GivenThatWeWantToGenerateImplicitNamespaceImports_DotNet(ITestOutputHelper log) : base(log) { }
+        public GivenThatWeWantToGenerateImpliciUsings_DotNet(ITestOutputHelper log) : base(log) { }
 
         [RequiresMSBuildVersionFact("17.0.0.32901")]
-        public void It_generates_dotnet_imports_and_builds_successfully()
+        public void It_can_generates_global_usings_and_builds_successfully()
```
AraHaan, line 22:
```suggestion
        public void It_can_generate_global_usings_and_builds_successfully()
```

AraHaan (changes requested):

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_WebApp.cs
```diff
@@ -54,17 +46,25 @@ public void It_generates_web_imports_and_builds_successfully()
...
 global using global::Microsoft.Extensions.Hosting;
 global using global::Microsoft.Extensions.Logging;
+global using global::System;
+global using global::System.Collections.Generic;
+global using global::System.IO;
+global using global::System.Linq;
+global using global::System.Net.Http;
+global using global::System.Net.Http.Json;
+global using global::System.Threading;
+global using global::System.Threading.Tasks;
```
RussKie, line 56: Please confirm the new order won't trigger [SA1208](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1208.md)
image

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
pranavkm: https://github.com/dotnet/sdk/issues/19697 to track this.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToGenerateGlobalUsings_WebApp.cs
pranavkm: As long as the analyzer does not run on generated files, this shouldn't be a concern.

dsplaisted (approved):

RussKie (commented):
👍

AraHaan:
Thanks.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
RussKie, line 13: May I please get a clarification why VB's implicit imports are now hidden behind `DisableImplicitNamespaceImports`, which is `on` by default? VB always had implicit imports on by default AFAIK.

This has broken https://github.com/dotnet/winforms/pull/5437

/cc: @KathleenDollard @KlausLoeffelmann @DamianEdwards

RussKie: IIUIC, https://github.com/dotnet/sdk/pull/18459/files#diff-5dfdd31992820273a25b399353a85ec6de6bdd620424a8adab3238c6ffd5df38 has removed a condition which hasn't been restored:
image

RussKie: I think this may be [the culprit](https://github.com/dotnet/sdk/pull/18459/files#diff-49cf43685a82b532902ba9b23390b53738fe901f67664b7a1aae71c0444e0dec), which needs to be undone:
image

RussKie: For a test I built a .NET 5.0 VB app, and neither `DisableImplicitNamespaceImports` nor `DisableImplicitFrameworkReferences` are defined.

E.g.
```
dotnet new winforms --language vb -f net5.0
dotnet build /bl
start msbuild.binlog
```

image
image

And it has all imports:
image

DamianEdwards: Looks like this might have been missed. @pranavkm?

KalleOlaviNiemitalo: Why does the screen shot show "net6.0-windows" if this is a " .NET 5.0 VB app"?

pranavkm: The issue that @RussKie is running in to is because of an Arcade change - https://github.com/dotnet/arcade/pull/7745. That said, I missed undoing one of the VB changes that is covered as part of https://github.com/dotnet/sdk/pull/19840

PR #19604, opened 2021-08-09
Warn instead of error on workload garbage collection exception

sfoslund:


joeloff (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -181,9 +181,17 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
...
+                if (_workloadInstaller.GetInstallationUnit().Equals(InstallationUnit.Packs))
+                {
+                    _workloadInstaller.GetPackInstaller().GarbageCollectInstalledWorkloadPacks();
+                }
+            }
+            catch (Exception e)
             {
-                _workloadInstaller.GetPackInstaller().GarbageCollectInstalledWorkloadPacks();
+                // Garbage collection failed, warn user
+                _reporter.WriteLine(string.Format(LocalizableStrings.GarbageCollectionFailed, e.Message).Yellow());
```
dsplaisted, line 194: The `Message` property of an exception is often not enough to tell what really went wrong.  Calling `ToString()` will include the stack trace and is better for diagnosibility.  However, it's not as user-friendly.  So I'd suggest printing the `Message` by default, but using `ToString()` if we're in verbose logging mode.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -181,9 +181,17 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
 
             InstallWorkloadsWithInstallRecord(workloadIds, _sdkFeatureBand, manifestsToUpdate, offlineCache);
 
-            if (_workloadInstaller.GetInstallationUnit().Equals(InstallationUnit.Packs))
+            try
+            {
+                if (_workloadInstaller.GetInstallationUnit().Equals(InstallationUnit.Packs))
+                {
+                    _workloadInstaller.GetPackInstaller().GarbageCollectInstalledWorkloadPacks();
```
dsplaisted, line 188: Is this the only place we call garbage collection?  Doesn't it happen on uninstall and update too?  Shouldn't we also be warning the same way anywhere it's called?

PR #19646, opened 2021-08-11
support --prerelease for dotnet tool

### LINKED ISSUE #12131: Support --prerelease for dotnet tools
Since https://github.com/NuGet/Home/issues/912 is ready

wli3:
Fix https://github.com/dotnet/sdk/issues/12131

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine.Parsing;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.Utils;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Tools.Tool.Install
+{
+    internal static class ParseResultExtension
+    {
+        public static VersionRange GetVersionRange(this ParseResult parseResult)
+        {
+            string packageVersion = parseResult.ValueForOption<string>(ToolInstallCommandParser.VersionOption);
+            bool prerelease = parseResult.ValueForOption<bool>(ToolInstallCommandParser.PrereleaseOption);
+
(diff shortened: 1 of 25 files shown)


src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    internal static class ParseResultExtension
+    {
+        public static VersionRange GetVersionRange(this ParseResult parseResult)
+        {
+            string packageVersion = parseResult.ValueForOption<string>(ToolInstallCommandParser.VersionOption);
+            bool prerelease = parseResult.ValueForOption<bool>(ToolInstallCommandParser.PrereleaseOption);
+            if (prerelease)
+            {
+                packageVersion = (packageVersion ?? "") + "*-*";
```
wli3, line 19: @nkolev92 it is more on the input side. This would work most of the time. 
"--version 1.3 --prerelease" would convert to 1.3*-*. However, if the input is "--version (1, 2] --prerelease", this won't work. Any suggestion? Do parsed VersionRange has something like versionRange.AllowPrerelease ?

dsplaisted (approved):

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
nkolev92: There's no means of selecting prerelease only. Prerelease tends to be in addition to stable. If you want a prerelease range you'd need to specify a prerelease version in it.

When combining a version  `--prerelease`, things are ambiguous.
I'd suggest to consider not combining them. 

You either want the latest or an explicit version. Having an in between seems like a potential world of pain. 
`--version 1.3 --prerelease` being `1.3*-*` would mean that things like `1.3.0-preview.2` are included as well. 

Note that *not* combining them is how `dotnet add package` works.

WeihanLi: Would `1.3*-*` match `1.30.0`?

wli3: Blocked combined

Issue #16563, opened 2021-03-28
Revert `DisableWinExeOutputInference`


batzen:
With https://github.com/dotnet/sdk/issues/11398 the `OutputType` a user explicitly sets in his projects is ignored, except when he sets `DisableWinExeOutputInference`.

This is not acceptable and was not asked for by the community.
https://github.com/dotnet/sdk/issues/11398 was not discussed and the feedback given there was simply ignored.

We now have a very, very confusing behavior where the users explicit choice of setting `OutputType` to `exe` is completely ignored.

Please see the following issues for reference:
- https://github.com/dotnet/sdk/issues/13331
- https://github.com/dotnet/sdk/issues/13624
- https://github.com/dotnet/wpf/issues/4314
- https://github.com/dotnet/wpf/issues/4138
- https://developercommunity2.visualstudio.com/t/WPF-NET-Core---Console-Application-does/1289479

There wasn't even an explanation why this should be changed in the first place.
The reasoning in the documentation at https://docs.microsoft.com/de-de/dotnet/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type states "It's assumed that most users don't want a console window to open when a WPF or Windows Forms app is executed".
May i ask who assumes this because of which issues?
Was this requested by someone? https://github.com/dotnet/sdk/issues/11398 did not mention any users/issues identifying this as an issue.

jogibear9988:
This was the Pull req. https://github.com/dotnet/sdk/pull/12152

marcpopMSFT:
Thanks for the feedback.

We should have limited this change to only projects targeting 5.0 and newer so we did not affect existing projects. We have rectified that going forward: https://github.com/dotnet/sdk/pull/16472. This was an attempt to fix existing apps that used templates that incorrectly defaulted to OutputType=Exe for a Windows Desktop application.  That application type should not have default behavior for including the console window spawning. There are many scenarios that customers would still want the console window but we did not want that to be the default behavior.  As mentioned in the link you included, we also want to align around OutputType going forward for multi-targeting to other platforms that don't support winexe.

jogibear9988:
@sfoslund @marcpopMSFT 
have you read all the complains? for example also the ones in my ticket? https://github.com/dotnet/wpf/issues/4314
Also if it is not the default any more, how could the user switch to console via the visual studio UI? what is the user experience if he changes a dropdown, and after that the value jumps back?

batzen:
[...]

PR #19651, opened 2021-08-11
Remove WinExe OutputType inference

dsplaisted:
In .NET 5, we [automatically converted](https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/automatically-infer-winexe-output-type) an `OutputType` of `Exe` to `WinExe` for WPF and Windows forms apps.  This was intended to simplify .NET MAUI apps (so that the OutputType wouldn't need to be conditioned on the TargetFramework).

However:

- This broke user expectations and frustrated some folks.  See #16563 and other linked issues
- .NET MAUI apps are going to by default use WinUI, not Windows Forms or WPF, so this logic won't even apply to them

So this PR reverts the previous automatic `OutputType` conversion.

@mattleibow @jonathanpeppers @Redth The .NET MAUI targets for Windows should probably do something similar to the logic being removed here, ie change the `OutputType` from `Exe` to `WinExe` when targeting Windows.


joeloff (approved):

RussKie:
To confirm - .NET 5.0 apps are still affected, but .NET 6.0+ aren't; and we don't intend to service .NET 5.0. Is this correct?

PR #19669, opened 2021-08-11
Make PackageValidation uses the right roslyn when the Microsoft.Net.Compilers.Toolset is used 

### LINKED ISSUE #19615: PackageValidation uses the wrong roslyn when the Microsoft.Net.Compilers.Toolset is used
PackageValidation uses the `RoslynTargetsPath` property which always points to the compiler in the SDK. When using the toolset package, [the CSharpCoreTargetsPath and VisualBasicCoreTargetsPath properties are overwritten](https://github.com/dotnet/roslyn/blob/06495a2bf3813e13cff59d2cdf30c24c99966d8e/src/NuGet/Microsoft.Net.Compilers.Toolset/build/Microsoft.Net.Compilers.Toolset.props#L10-L11) but never the `RoslynTargetsPath`.

Also, as the CodeAnalysis task assembly is bundled with the toolset package, does this logic need to run at all when using the Microsoft.Net.Compilers.Toolset package? https://github.com/dotnet/sdk/blob/a6e14e80ca216d6a394387a628f0e3abbadd6f28/src/Compatibility/Microsoft.DotNet.PackageValidation/ValidatePackage.cs#L78-L92

Anipik:
Fixes https://github.com/dotnet/sdk/issues/19615
depends on https://github.com/dotnet/roslyn/pull/55552

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
@@ -8,6 +8,8 @@
 
     <PropertyGroup Condition="'$(RoslynAssembliesPath)' == ''">
       <RoslynAssembliesPath>$(RoslynTargetsPath)</RoslynAssembliesPath>
+      <RoslynAssembliesPath Condition="'$(MSBuildProjectExtension)' == '.csproj'">$([System.IO.Path]::GetDirectoryName($(CSharpCoreTargetsPath)))</RoslynAssembliesPath>
+      <RoslynAssembliesPath Condition="'$(MSBuildProjectExtension)' == '.vbproj'">$([System.IO.Path]::GetDirectoryName($(VisualBasicCoreTargetsPath)))</RoslynAssembliesPath>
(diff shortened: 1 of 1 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -8,6 +8,8 @@
 
     <PropertyGroup Condition="'$(RoslynAssembliesPath)' == ''">
       <RoslynAssembliesPath>$(RoslynTargetsPath)</RoslynAssembliesPath>
+      <RoslynAssembliesPath Condition="'$(MSBuildProjectExtension)' == '.csproj'">$([System.IO.Path]::GetDirectoryName($(CSharpCoreTargetsPath)))</RoslynAssembliesPath>
```
safern, line 11: should we use the `Language` property instead? What if the project is an fsharp project?

Anipik: > should we use the Language property instead? 

i decided against it because the sdk was using this MSBuildProjectExtension property to make this decision while setting this *CoreTargetsPath properties from RoslynTargetsPath.

> What if the project is an fsharp project

we dont overrride this property in the compilers package hence will use the default sdk version from RoslynAssembliesPath

safern: A ok, sounds good.

safern (approved):

PR #19681, opened 2021-08-12
Add self contained short flag

### LINKED ISSUE #19576: Enable `-s` as a convenience over `--self-contained`
This is a pretty common mode. Today, you can use `-r` to mean self-contained. We're changing that.

Proposal:

- Enable `-s` (or some other short-form like `--sc`) as a convenience over `--self-contained`.
- Use the SDK RID as an implicit RID if non is specified.

Proposed timeframe: 6.0.200.

Context: https://github.com/dotnet/designs/pull/233#discussion_r684318468

WeihanLi:
#19576 

Add `--sc` short for `--self-contained`

### DIFF
--- a/src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
+++ b/src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
@@ -98,7 +98,9 @@ namespace Microsoft.DotNet.Cli.Publish.Tests
 
         [Theory]
         [InlineData(null)]
+        [InlineData("-s")]
         [InlineData("--self-contained")]
+        [InlineData("-s=true")]
         [InlineData("--self-contained=true")]
         public void ItPublishesSelfContainedWithRid(string args)
         {
@@ -115,6 +117,7 @@ namespace Microsoft.DotNet.Cli.Publish.Tests
         }
 
         [Theory]
+        [InlineData("-s=false")]
         [InlineData("--self-contained=false")]
         [InlineData("--no-self-contained")]
         public void ItPublishesFrameworkDependentWithRid(string args)
@@ -142,6 +145,7 @@ namespace Microsoft.DotNet.Cli.Publish.Tests
         }
 
         [Theory]
+        [InlineData("-s=false")]
         [InlineData("--self-contained=false")]
         [InlineData(null)]
         [InlineData("--no-self-contained")]
@@ -165,7 +169,9 @@ namespace Microsoft.DotNet.Cli.Publish.Tests
         }
 
         [Theory]
+        [InlineData("-s --no-self-contained")]
         [InlineData("--self-contained --no-self-contained")]
+        [InlineData("-s=true --no-self-contained")]
         [InlineData("--self-contained=true --no-self-contained")]
         public void ItFailsToPublishWithConflictingArgument(string args)
         {
--- a/src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
+++ b/src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
@@ -98,9 +98,9 @@ namespace Microsoft.DotNet.Cli.Publish.Tests
 
         [Theory]
         [InlineData(null)]
-        [InlineData("-s")]
+        [InlineData("--sc")]
         [InlineData("--self-contained")]
-        [InlineData("-s=true")]
+        [InlineData("--sc=true")]
         [InlineData("--self-contained=true")]
         public void ItPublishesSelfContainedWithRid(string args)
         {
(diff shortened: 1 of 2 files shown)


richlander:
Love it. When we merge this, we should also add support an implicit RID at the same time. That will make the feature from good to ***great***.

/cc @agocke

agocke:
I think we just need a few tests here. There are some existing tests in `GivenDotnetPublishPublishesProject.cs` that test `--self-contained` that could probably be adjusted to test this as well by adding more `InlineData` clauses

WeihanLi:
> I think we just need a few tests here. There are some existing tests in `GivenDotnetPublishPublishesProject.cs` that test `--self-contained` that could probably be adjusted to test this as well by adding more `InlineData` clauses

Great thanks @agocke, would add some test cases

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -165,6 +166,7 @@ public void ItPublishesFrameworkDependentWithoutRid(string args)
         }
 
         [Theory]
+        [InlineData("-s --no-self-contained")]
```
AraHaan, line 169: What about ``-s=true --no-self-contained``?

Also what about a shorthand for ``--no-self-contained`` as well?

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -98,6 +98,7 @@ public void ItDoesNotImplicitlyRestoreAProjectWhenPublishingWithTheNoRestoreOpti
 
         [Theory]
         [InlineData(null)]
+        [InlineData("-s")]
```
AraHaan, line 101: What about ``-s=true``?

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
WeihanLi: I thought `-s` stands for `--self-contained true`, if `-s=false` is needed, I'd like to add more cases.

AraHaan: It probably is, but I seen test cases where it tests ``--self-contained=true``.

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
WeihanLi: Added, thanks

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
WeihanLi: Added, thanks

richlander: I don't think we need a short-hand for `--no-self-contained`. The plan is for that to be the default in .NET 7.

agocke (approved):
LGTM

agocke:
@richlander Any more comments

agocke:
Also @wli3 in-case I'm missing anything

wli3:
@KathleenDollard could you double check that? We need to be careful adding new short hands, since we have only 26 letters to use

wli3:
it is good on the technical side

KathleenDollard:
Unfortunately, `s` is `source`. This is important on the install commands, and it also works on `restore`. 

We avoid having the same abbreviation mean different things, especially for closely related commands. And we work hard to have consistent abbreviations where a particular option is available. `build` and `restore` are obviously closely linked commands and it would seem likely consider them sharing more options in the future. 

Since `build` does a `restore` and supports `s`, taking this would mean that we could not add `--source` to pass through to restore. It would also that we could not specify `--self-contained` for `restore` (to get the right packages). Unfortunately, `c` doesn't work and `-sc` would be confusing for Posix users. We generally avoid two letter abbreviations because of that, but as Will points out, there are only 26 letters and there is value to an abbreviation here.

AraHaan:
I think there is no choice but to abbreviate this as ``-sc``.

agocke:
We could do `--sc`, which is still short but also fits posix convention

richlander:
`--sc` sounds great to me.

I still want to hold on this change until we add an implicit RID by default.

AraHaan:
> `--sc` sounds great to me.
> 
> I still want to hold on this change until we add an implicit RID by default.

Is there a way to do that based on what it detects on the os and what cpu bitness it detects as well?

WeihanLi:
Thanks, will update to `--sc` later. 

@richlander as for the default implicit RID, I'm going to implement it in another PR <https://github.com/dotnet/sdk/pull/19725>, I'm not sure about the changes, maybe @agocke could help?

wli3:
FYI @KathleenDollard is not available today.

marcpopMSFT:
Got @KathleenDollard's signoff offline.  @wli3 , anything else before we approve and merge?

AraHaan:
Is this ready to merge yet?

wli3 (approved):

Issue #19704, opened 2021-08-12
[StaticWebAssets] Tweak the settings when invoking referenced projects


javiercn:
Through integrating static web assets in Maui we've discovered that we need some Tweaks to the project references. When we capture the project references and merge them in a StaticWebAssetConfiguration item, we need to ensure we capture the right set of properties from the associated project reference to ensure later on we invoke them in the same way we did to collect the references originally.

This for example mean capturing thins like UndefineProperties and so on to apply them appropriately in the subsequent calls.

dsplaisted:
Looks like this was fixed with #19720?

PR #19720, opened 2021-08-13
[Static web assets] Additional unit tests and fixes

javiercn:
* Adds unit tests for most of the MSBuild tasks.
* Validate that all asset paths are unique when we build the manifest.
* Fixes a couple of small bugs when we grouped assets by Identity instead of `target path`.
* Fixes the way we merge properties for projects to account for `UndefineProperties`


javiercn:
Fixes https://github.com/dotnet/sdk/issues/19704

src/Tests/Microsoft.NET.Sdk.Razor.Tests/AspNetSdkBaselineTest.cs
```diff
@@ -23,7 +23,7 @@ public class AspNetSdkBaselineTest : AspNetSdkTest
 #if GENERATE_SWA_BASELINES
         public static bool GenerateBaselines = true;
 #else
-        public static bool GenerateBaselines = false;
+        public static bool GenerateBaselines = true;
```
pranavkm, line 26:
```suggestion
        public static bool GenerateBaselines = false;
```

We should add a test for this that says that `Assert.False(GenerateBaselines);`

src/Tests/Microsoft.NET.Sdk.Razor.Tests/StaticWebAssets/ReadStaticWebAssetsManifestFileTest.cs
```diff
@@ -0,0 +1,346 @@
...
+            task.ReferencedProjectsConfiguration.Should().BeNull();
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                if (TempFilePath != null && File.Exists(TempFilePath))
+                {
+                    File.Delete(TempFilePath);
```
pranavkm, line 338: The SDK tests leave these files around so you can see it. They write to the Intermediate dir, and use the same file name for each run, so the number of files doesn't grow. Perhaps we could consider doing the same thing here.

pranavkm (approved):

src/Tests/Microsoft.NET.Sdk.Razor.Tests/AspNetSdkBaselineTest.cs
javiercn: Good catch, I just did.

src/Tests/Microsoft.NET.Sdk.Razor.Tests/StaticWebAssets/ReadStaticWebAssetsManifestFileTest.cs
javiercn: I changed to use the TestContext and put the temp files in the temp path.

PR #19731, opened 2021-08-13
Fix publishing .NET Standard facades

### LINKED ISSUE #18101: [Regression] .NET 6 SDK will no longer copy the netstandard facades to the publish directory for net461 apps
cc: @dsplaisted @marcpopMSFT @andschwa @ericstj 

**Repro Steps**

1. Create a new netstandard lib
```xml

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>

</Project>
```
2. Create a new net461 Console app that references it:
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net461</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\reproLib\reproLib.csproj" />
  </ItemGroup>

</Project>
```

3. Publish the project with `dotnet publish`

**Expected**
Because the net461 project is referencing a netstandard library, tooling should inject all of the netstandard facades onto the publish directory to ensure the application will be able to run as expected. **This is what happens with .NET 5.0 and previous sdks**.

**Actual**
`bin` directory does get the facades injected, but the publish directory won't when you use a 6.0 SDK, which means that if the app tries to run in a machine that only has .NET 461 installed, the app will likely crash at runtime.

**Problem**

I took a look and the issue here seems to be caused by [this change](https://github.com/dotnet/sdk/pull/14020/files#diff-946cba6640de0522dafbdd3ab4294eeb086609079f451068ab0b575ade3442c0R721) which basically changed the way we calculate the items to be copied to the publish directory and added an additional filter to check if the `ResolveCopyLocal` item had the `Private` metadata set to `false`. When we are injecting the netstandard facades to resolveCopyLocal item, we explicitly set Private to false, which is why they are now all getting excludded from the publish directory. Here is where we set the metadata to false and the explanation of why we do it:
https://github.com/dotnet/sdk/blob/203c2ac3eeb99d188d2b439c96f72f93b7b726fd/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets#L91-L109

This regression has a potential of great impact, so we should consider how we can fix it before 6.0 ships, whether the solution is to not set Private item to false or if the solution is instead to change the new filter on the publish items.

dsplaisted:
Fixes #18101

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets
+++ b/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets
@@ -104,9 +104,7 @@ Copyright (c) .NET Foundation. All rights reserved.
            and VS won't show a warning icon on the reference.  See https://github.com/dotnet/sdk/issues/1499
            -->
 
-      <ReferenceCopyLocalPaths Include="@(_NETStandardLibraryNETFrameworkLib)" Condition="'%(FileName)' != 'netfx.force.conflicts'">
-        <Private>false</Private>
-      </ReferenceCopyLocalPaths>
+      <ReferenceCopyLocalPaths Include="@(_NETStandardLibraryNETFrameworkLib)" Condition="'%(FileName)' != 'netfx.force.conflicts'"/>
 
       <_UpdatedReference Remove="@(_UpdatedReference)" />
     </ItemGroup>
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishWithoutConflicts.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishWithoutConflicts.cs
@@ -50,8 +50,9 @@ namespace Microsoft.NET.Publish.Tests
             var files = getValuesCommand.GetValues()
                 .Where(file => file.Contains(reference));
             files.Count().Should().Be(1);
-            // We should choose the file from system.runtime.interopservices.runtimeinformation package version
-            files.FirstOrDefault().Contains(@"system.runtime.interopservices.runtimeinformation\4.3.0").Should().BeTrue();
+
+            // We should choose the system.runtime.interopservices.runtimeinformation file from Microsoft.NET.Build.Extensions as it has a higher AssemblyVersion (4.0.2.0 compared to 4.0.1.0)
+            files.FirstOrDefault().Contains(@"Microsoft.NET.Build.Extensions\net461\lib\System.Runtime.InteropServices.RuntimeInformation.dll").Should().BeTrue();
         }
 
         [Theory]
--- a/src/Tests/Microsoft.NET.Publish.Tests/PublishNetFrameworkApp.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/PublishNetFrameworkApp.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using FluentAssertions;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.NET.TestFramework;
+using Microsoft.NET.TestFramework.Assertions;
+using Microsoft.NET.TestFramework.Commands;
+using Microsoft.NET.TestFramework.ProjectConstruction;
+using Xunit;
+using Xunit.Abstractions;
+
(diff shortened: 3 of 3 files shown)


src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets
```diff
@@ -104,9 +104,7 @@ Copyright (c) .NET Foundation. All rights reserved.
            and VS won't show a warning icon on the reference.  See https://github.com/dotnet/sdk/issues/1499
            -->
 
-      <ReferenceCopyLocalPaths Include="@(_NETStandardLibraryNETFrameworkLib)" Condition="'%(FileName)' != 'netfx.force.conflicts'">
-        <Private>false</Private>
-      </ReferenceCopyLocalPaths>
+      <ReferenceCopyLocalPaths Include="@(_NETStandardLibraryNETFrameworkLib)" Condition="'%(FileName)' != 'netfx.force.conflicts'"/>
```
joperezr, line 107: Perhaps we should also update the comment above since it still says that Private is set to false?

joperezr (approved):
Small nit, but other than updating the comment this LGTM. Thanks for fixing this @dsplaisted!

src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.NETFramework.targets
dsplaisted: Private is still set to false on the `Reference` items which are created in the `AddFacadesToReferences` task.  That's what the comment is supposed to be referring to, as far as I understand.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishWithoutConflicts.cs
```diff
@@ -50,8 +50,9 @@ public void It_solves_conflicts_between_package_and_implicit_references()
             var files = getValuesCommand.GetValues()
                 .Where(file => file.Contains(reference));
             files.Count().Should().Be(1);
-            // We should choose the file from system.runtime.interopservices.runtimeinformation package version
-            files.FirstOrDefault().Contains(@"system.runtime.interopservices.runtimeinformation\4.3.0").Should().BeTrue();
+
+            // We should choose the system.runtime.interopservices.runtimeinformation file from Microsoft.NET.Build.Extensions as it has a higher AssemblyVersion (4.0.2.0 compared to 4.0.1.0)
+            files.FirstOrDefault().Contains(@"Microsoft.NET.Build.Extensions\net461\lib\System.Runtime.InteropServices.RuntimeInformation.dll").Should().BeTrue();
```
dsplaisted, line 55: @joperezr @ericstj I had to change this test to expect the file from Microsoft.NET.Build.Extensions to win instead of the one from the package.  This seems legit, here's the message from the binlog:

> Encountered conflict between 'Reference:C:\Users\daplaist\.nuget\packages\system.runtime.interopservices.runtimeinformation\4.3.0\ref\netstandard1.1\System.Runtime.InteropServices.RuntimeInformation.dll' and 'Reference:System.Runtime.InteropServices.RuntimeInformation'. Choosing 'Reference:System.Runtime.InteropServices.RuntimeInformation' because AssemblyVersion '4.0.2.0' is greater than '4.0.1.0'.

I'm not sure why the test originally expected the other file to win, maybe it was always a bug.

Does this seem OK?

joperezr: How was the test even passing before this change? I'm seeing the same behavior locally as the test after your change as well, and also for the copyLocal item file:
```
Encountered conflict between 'CopyLocal:C:\Users\joe_p\.nuget\packages\system.runtime.interopservices.runtimeinformation\4.3.0\runtimes\win\lib\net45\System.Runtime.InteropServices.RuntimeInformation.dll' and 'CopyLocal:C:\Program Files\dotnet\sdk\5.0.100-rc.2.20479.15\Microsoft\Microsoft.NET.Build.Extensions\net461\lib\System.Runtime.InteropServices.RuntimeInformation.dll'. Choosing 'CopyLocal:C:\Program Files\dotnet\sdk\5.0.100-rc.2.20479.15\Microsoft\Microsoft.NET.Build.Extensions\net461\lib\System.Runtime.InteropServices.RuntimeInformation.dll' because AssemblyVersion '4.0.2.0' is greater than '4.0.1.0'.
```

Given this, I'm vine with fixing the test as well as I agree that it must have been a bug before unless this was written before we serviced the Microsoft.NET.Build.Extensions facades a long time ago and we were actually expecting the package to win.

PR #19733, opened 2021-08-13
Rename package to Microsoft.DotNet.Compatibility and fix regressions

Anipik:
- Builds a new package Microsoft.DotNet.Compatibility
- Moves all the tasks & targets into Microsoft.DotNet.Compatibility
- Fixes the roslyn regression being carried in the package.
- Fixes the roslyn path regression not being correctly set for cross-targeting builds


src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
```diff
@@ -0,0 +1,92 @@
...
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoWarn>$(NoWarn);NU5128</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\..\Tasks\Common\**\*.cs" LinkBase="Common" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)" ExcludeAssets="Runtime" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" ExcludeAssets="Runtime" />
+    <!-- We carry NuGet as part of the package in case the package is used with an older SDKs or with full framework MSBuild. -->
+    <PackageReference Include="NuGet.Frameworks" Version="$(NuGetBuildTasksPackageVersion)" />
```
safern, line 18: Should these instead be transitive by `PackageValidation`?

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
```diff
@@ -0,0 +1,92 @@
...
+    <PackageReference Include="NuGet.Packaging" Version="$(NuGetBuildTasksPackageVersion)" />
+    <PackageReference Include="NuGet.Protocol" Version="$(NuGetBuildTasksPackageVersion)" />
+    <ProjectReference Include="..\Microsoft.DotNet.PackageValidation\Microsoft.DotNet.PackageValidation.csproj" ExcludeAssets="Runtime" />
+    <ProjectReference Include="..\Microsoft.DotNet.ApiCompatibility\Microsoft.DotNet.ApiCompatibility.csproj" ExcludeAssets="Runtime" />
```
safern, line 22: Could you add a comment on why we are setting `ExcludeAssets="Runtime"` to a `P2P`?

src/Compatibility/Microsoft.DotNet.Compatibility/PackageValidationLogger.cs
```diff
@@ -4,9 +4,10 @@
 using System.IO;
 using Microsoft.Build.Framework;
 using Microsoft.DotNet.Compatibility.ErrorSuppression;
+using Microsoft.DotNet.PackageValidation;
 using Microsoft.NET.Build.Tasks;
 
-namespace Microsoft.DotNet.PackageValidation
+namespace Microsoft.DotNet.Compatibility
 {
     internal class PackageValidationLogger : IPackageLogger
```
safern, line 12: Maybe we should rename this to not mention `Package`?, Maybe, `ValidationLogger`? I would expect this to end up being used by `API Compat` task.

src/Compatibility/Microsoft.DotNet.PackageValidation/IPackageLogger.cs
```diff
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Build.Framework;
-using Microsoft.DotNet.ApiCompatibility.Abstractions;
 using Microsoft.DotNet.Compatibility.ErrorSuppression;
 
 namespace Microsoft.DotNet.PackageValidation
```
safern, line 7: Move this file as well and rename it as the API Compat task will probably end up using the logger?

Anipik: The would be a circular dependency Compatibility depends on package validation & api compatibility. we can move the type down to api compat.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -12,6 +11,7 @@
     <!-- We pin the versions of the assets that ship with the SDK in order to be able to run on older SDKs. 
     Currently we support .NET 6 Preview5 or later. -->
     <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21304.5" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)" ExcludeAssets="Runtime" />
```
joperezr, line 14: Why do we have the `ExcludeAssets`? Is it so that we don't carry roslyn with us? If so could you add a comment?

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
```diff
@@ -0,0 +1,87 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
+    <IsPackable>true</IsPackable>
+    <IsShippingPackage>true</IsShippingPackage>
+    <StrongNameKeyId>Open</StrongNameKeyId>
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoWarn>$(NoWarn);NU5128</NoWarn>
```
joperezr, line 10: Can you add a comment of what this NoWarn is and why we need it?

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
```diff
@@ -0,0 +1,87 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
+    <IsPackable>true</IsPackable>
+    <IsShippingPackage>true</IsShippingPackage>
+    <StrongNameKeyId>Open</StrongNameKeyId>
+    <VersionPrefix>1.0.0</VersionPrefix>
```
joperezr, line 8: Why do we need a Version prefix?

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
```diff
@@ -0,0 +1,87 @@
...
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoWarn>$(NoWarn);NU5128</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\..\Tasks\Common\**\*.cs" LinkBase="Common" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" ExcludeAssets="Runtime" />
```
joperezr, line 15: Same here, please explain or add a comment on what is the ExcludeAssets for.

src/Compatibility/Microsoft.DotNet.PackageValidation/IPackageLogger.cs
safern: Hmm I don't think API Compat will use this interface, just the logger class. Let's leave it where it is and we can move it down if needed later on.

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
safern: This is how arcade versioning works. If the build is stable this would be treated as the package version, if not the suffix is added.

Anipik: Its basically use the sdk convention to override the default 6.0 default branding

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -3,15 +3,16 @@
...
     <NoWarn>RS1024;$(NoWarn)</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- ExcludeAssets="Runtime" is being set to avoid adding package references and their dependencies into the package. -->
     <!-- We pin the versions of the assets that ship with the SDK in order to be able to run on older SDKs. 
     Currently we support .NET 6 Preview5 or later. -->
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21304.5" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21304.5"  ExcludeAssets="Runtime" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)" ExcludeAssets="Runtime" />
```
safern, line 15: Why are we adding `Microsoft.Build.Framework` here? `API Compat` shouldn't depend on any MSBuild things at the moment.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -3,15 +3,16 @@
...
     <!-- We need to compare ISymbols in a special way (by name) and roslyn symbol comparers take more heuristics into consideration.-->
     <NoWarn>RS1024;$(NoWarn)</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- ExcludeAssets="Runtime" is being set to avoid adding package references and their dependencies into the package. -->
     <!-- We pin the versions of the assets that ship with the SDK in order to be able to run on older SDKs. 
     Currently we support .NET 6 Preview5 or later. -->
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21304.5" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.0.0-1.21304.5"  ExcludeAssets="Runtime" />
```
safern, line 14: This project should not exclude assets if we were to write another entry point, say a console app on top of it, it won't find Roslyn. We should exclude the assets on the P2P directly, but if we ship this as a package it should carry Roslyn with it.

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -2,93 +2,20 @@
...
     <StrongNameKeyId>Open</StrongNameKeyId>
-    <PackageType>MSBuildSdk</PackageType>
-    <VersionPrefix>1.0.0</VersionPrefix>
-    <IncludeBuildOutput>false</IncludeBuildOutput>
-    <NoWarn>$(NoWarn);NU5128</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="..\..\Tasks\Common\**\*.cs" LinkBase="Common" />
+    <!-- ExcludeAssets="Runtime" is being set to avoid adding package references and dependencies of project and package references. -->
```
safern, line 9: I think the object model packages should carry their own copies of things as if they are referenced as packages they should work fine, but on the MSBuild task project we should exclude the assets.

This goes against my comment of letting NuGet flow transitively, but now that I give it a second thought it makes more sense to just exclude things on the MSBuild task and include the references we want to carry as the task, and leave these object model packages as regular packages and not making them special for the first entry point we added.

src/Compatibility/Microsoft.DotNet.Compatibility/ValidationLogger.cs
```diff
@@ -4,20 +4,21 @@
...
 using Microsoft.Build.Framework;
 using Microsoft.DotNet.Compatibility.ErrorSuppression;
+using Microsoft.DotNet.PackageValidation;
 using Microsoft.NET.Build.Tasks;
 
-namespace Microsoft.DotNet.PackageValidation
+namespace Microsoft.DotNet.Compatibility
 {
-    internal class PackageValidationLogger : IPackageLogger
+    internal class ValidationLogger : ICompatibilityLogger
```
joperezr, line 12: NIT: Validation seems like an overloaded term to me. Perhaps CompatibilityLogger would be more accurate.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -1,15 +1,16 @@
...
   <Target Name="RunPackageValidation"
           AfterTargets="Pack"
           Condition="'$(IsPackable)' == 'true' and '$(EnablePackageValidation)' == 'true'">
 
     <PropertyGroup Condition="'$(RoslynAssembliesPath)' == ''">
       <RoslynAssembliesPath>$(RoslynTargetsPath)</RoslynAssembliesPath>
-      <RoslynAssembliesPath Condition="'$(MSBuildProjectExtension)' == '.csproj'">$([System.IO.Path]::GetDirectoryName($(CSharpCoreTargetsPath)))</RoslynAssembliesPath>
-      <RoslynAssembliesPath Condition="'$(MSBuildProjectExtension)' == '.vbproj'">$([System.IO.Path]::GetDirectoryName($(VisualBasicCoreTargetsPath)))</RoslynAssembliesPath>
+      <_packageReferenceList>@(PackageReference)</_packageReferenceList>
+      <RoslynAssembliesPath Condition="'$(MSBuildProjectExtension)' == '.csproj' and $(_packageReferenceList.Contains('Microsoft.Net.Compilers.Toolset'))">$([System.IO.Path]::GetDirectoryName($(CSharpCoreTargetsPath)))</RoslynAssembliesPath>
```
joperezr, line 12: What happens if _packageReferenceList does not have the Compilers package? Should we default to a different path? Or should we throw an error at this point?

joperezr:
Left 2 small nits, but other than that this looks good.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
Anipik: We will use the default path set here https://github.com/dotnet/sdk/pull/19733/files/1f6b652b649b96ebb9a736e1a384c1e1e03da783#diff-56820c7e920cd3e52b0bb614980ba2ace37edc9b01ae8f4cf8e2620686f998a4R10

joperezr (approved):

PR #19757, opened 2021-08-16
Add CodeSignAppHost Task

mateoatr:
Add a task with the codesigning logic that currently lives on the `HostModel`. For single-file publish, we make sure that we sign the binary only after we've completed generating the bundle and not on apphost creation; otherwise, we'd end up having a malformed binary, since codesigning affects the binary format.


vitek-karas:
@mateoatr @agocke  - please discuss which solution is better. This PR or the one implemented in https://github.com/dotnet/runtime/pull/57532. As far as I can tell both have a good chance of working.

mateoatr:
> @mateoatr @agocke - please discuss which solution is better. This PR or the one implemented in [dotnet/runtime#57532](https://github.com/dotnet/runtime/pull/57532). As far as I can tell both have a good chance of working.

Discussed already -- we are making changes in the runtime. Closing.

PR #19764, opened 2021-08-16
Set the EnableDynamicLoading property before importing Microsoft.PackageDependencyResolution.targets

jkoritzinsky:
The condition for `CopyLocalLockFileAssemblies` in Microsoft.Package.DependencyResolution.targets depends on the `EnableDynamicLoading` property and should be true in the `EnableComHosting` case.

This issue was discovered when working on https://github.com/dotnet/winforms/pull/4685

cc: @ericstj @RussKie

### DIFF
--- a/src/Assets/TestProjects/ComServerWithDependencies/ComServer.cs
+++ b/src/Assets/TestProjects/ComServerWithDependencies/ComServer.cs
@@ -1,0 +1,28 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace COMServer
+{
+    [ComVisible(true)]
+    [Guid("17B6329E-B025-4FC7-A854-97D34600C5A6")]
+    public class Class1
+    {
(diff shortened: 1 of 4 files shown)


dsplaisted (approved):
Could we have a test for this?  It could go next to `GivenThatWeWantToBuildALibrary.It_can_build_with_dynamic_loading_enabled`

dsplaisted:
FYI @elinor-fung @jkoritzinsky, looks like this has been broken since before #3305

RussKie (approved):
👍 

Tested in https://github.com/dotnet/winforms/pull/4685#issuecomment-899882567

ericstj:
So @russkie mentioned this regressed in preview7 https://github.com/dotnet/winforms/pull/4685#issuecomment-899204843

Do we understand why it regressed to make sure there aren't any other properties regressed?

dsplaisted:
No, I don't know why this would have regressed in Preview 7.  I looked at the diff between Preview 5 and Preview 7 for Microsoft.NET.Sdk.targets and Microsoft.PackageDependencyResolution.targets, and it doesn't look like anything changed that would have affected this.

I don't know exactly what regressed or how, but it seems like it may not have been related to the issue being fixed in this PR.

jkoritzinsky:
Based on https://github.com/dotnet/samples/pull/4773, I believe this regressed due to the `.runtimeconfig.dev.json` not being produced by default any more.

PR #19771, opened 2021-08-17
[Package Validation] resolve references when running API Compat

safern:
I am not enabling references for baseline package validation as it is not that trivial to get the references. We should work with nuget to support that on 7.0.

@joperezr im just missing some testing. Also what I’ve been doing is running this on dotnet/runtime, with an allConfigurations build to generate all the package and the property to generate the suppression files and then look at the diff. That has helped spot some bugs that I’m fixing as part of this PR.

cc: @anipik @ericstj


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -33,6 +34,37 @@
       CompatibilitySuppressionFilePath="$(CompatibilitySuppressionFilePath)"
       BaselinePackageTargetPath="$(PackageValidationBaselinePath)"
       DisablePackageBaselineValidation="$(DisablePackageBaselineValidation)"
-      RoslynAssembliesPath="$(RoslynAssembliesPath)" />
+      RoslynAssembliesPath="$(RoslynAssembliesPath)"
+      ReferenceDirectories="@(PackageValidationReferenceDirectories)" />
```
ericstj, line 38: Why do we use directories instead of actual ReferencePath items?  It's possible this allows more references to be seen than we want.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -33,6 +34,37 @@
...
+      <_referenceDirectories Include="@(ReferencePath -> '%(RootDir)%(Directory)')" />
+      <_ReferenceDirectoriesWithTargetFramework Include="%(_referenceDirectories.Identity)"  TargetFramework="$(TargetFramework)" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_GetPackageValidationReferenceDirectories"
+          DependsOnTargets="$(_GetPackageValidationReferenceDirectoriesDependsOn)"
+          Condition="'$(RunPackageValidationWithoutReferences)' != 'true'">
+
+    <MSBuild Projects="@(_InnerBuildProjects)"
```
ericstj, line 59: Can we add a comment about where we copied this technique from (NuGet targets I believe), and what it needs to stay consistent with?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -33,6 +34,37 @@
...
+          Returns="@(_ReferenceDirectoriesWithTargetFramework)">
+    <ItemGroup>
+      <_referenceDirectories Include="@(ReferencePath -> '%(RootDir)%(Directory)')" />
+      <_ReferenceDirectoriesWithTargetFramework Include="%(_referenceDirectories.Identity)"  TargetFramework="$(TargetFramework)" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_GetPackageValidationReferenceDirectories"
+          DependsOnTargets="$(_GetPackageValidationReferenceDirectoriesDependsOn)"
+          Condition="'$(RunPackageValidationWithoutReferences)' != 'true'">
```
ericstj, line 57: So it looks like we have this enabled by default if I'm reading this right.  That seems OK if we are happy with the perf of this.  Might be interesting to measure that to see what it adds for a characteristic package.

ericstj:
> I am not enabling references for baseline package validation as it is not that trivial to get the references. We should work with nuget to support that on 7.0.

If we assume that all the dependencies of the baseline are compatible with the current, can we use the references from the current build for the baseline?  We'd miss cases where the dependency changed, or cases where the latest build no longer depends on something the baseline build did, but wouldn't the net result be less false positives?

If someone wanted to fetch the dependencies for the baseline themselves, eg like this https://github.com/dotnet/machinelearning/blob/526215dc3f057b7ff53adf6014552c29a5407466/tools-local/Microsoft.ML.StableApi/Microsoft.ML.StableApi.csproj#L29-L30 does our task provide a way for them to pass them in?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -37,6 +37,8 @@ public class ApiComparer
         /// </summary>
         public bool StrictMode { get; set; }
 
+        public bool RunningWithReferences { get; set; }
```
joperezr, line 40: NIT: Missing triple slash comments. I'll add these in the next commit.

ericstj:
>  I would just add a way for users to specify their referencePaths in case they want to run APICompat against a baseline and use references

That seems reasonable to unblock the folks who might want to go this route.

IMO we are OK and making forward progress if we reduce false negatives (add new rules) reduce false positives (add dependencies that might let us compare a moved type).  If a change introduces more false negatives (hides real problems) or more false positives (noisy errors that need to be suppressed) it is bad.  There still may be value if a change reduces some, but not all false positives, if the cases it misses were already issues that the user had to deal with (due to the missing dependencies).

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Properties/launchSettings.json
```diff
@@ -0,0 +1,13 @@
+{
```
joperezr, line 1: NIT: we should remove this file.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -87,6 +99,58 @@ public void RunApiCompat()
...
+
+            AssemblySymbolLoader loader = new(resolveAssemblyReferences: shouldResolveReferences);
+            if (shouldResolveReferences)
+            {
+                resolvedReferences = true;
+                loader.AddReferenceSearchDirectories(directories);
+            }
+            else if (!_isBaselineSuppression && _referenceDirectories != null)
+            {
+                _log.LogWarning(
```
joperezr, line 120: So far we have used the warnings mostly for compatibility validation errors. This case is a bit different as this is mainly just a message that says that we are not using Warnings for this specific check. Should this be informational message as opposed to warning? @ericstj thoughts?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -87,6 +99,58 @@ public void RunApiCompat()
...
+
+            IAssemblySymbol symbol = loader.LoadAssembly(assemblyInformation.AssemblyName, assemblyStream);
+
+            if (loader.HasLoadWarnings(out IEnumerable<AssemblyLoadWarning> warnings))
+            {
+                resolvedReferences = false;
+                foreach (AssemblyLoadWarning warning in warnings)
+                {
+                    _log.LogWarning(
+                        new Suppression()
```
joperezr, line 139: Same goes for this warning, should it be message instead?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -87,6 +99,58 @@ public void RunApiCompat()
...
+            else if (!_isBaselineSuppression && _referenceDirectories != null)
+            {
+                _log.LogWarning(
+                    new Suppression()
+                    {
+                        DiagnosticId = ApiCompatibility.DiagnosticIds.SearchDirectoriesNotFoundForTfm,
+                        Target = assemblyInformation.DisplayString
+                    },
+                    ApiCompatibility.DiagnosticIds.SearchDirectoriesNotFoundForTfm,
+                    "Could not find a reference directory in the provided directories for TargetFramework '{0}' when loading '{1}'. For more information see: https://aka.ms/dotnetpackagevalidation",
```
ericstj, line 127: Shouldn't this be a localized resource?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
ericstj: If the user has communicated to us that they wish to use dependencies, but has not provided those dependencies then we need to tell the user about this.  I don't think a message is strong enough.  Error's are not suppressible.  I do think it makes sense to use a suppressible warning.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
ericstj: Messages are effectively never observed except for diagnostics, so I don't think we'd use a message.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -3,6 +3,7 @@
   <UsingTask TaskName="Microsoft.DotNet.PackageValidation.ValidatePackage" AssemblyFile="$(DotNetPackageValidationAssembly)" />
 
   <Target Name="RunPackageValidation"
+          DependsOnTargets="_GetPackageValidationReferenceDirectories"
```
Anipik, line 6: can we use property here like ```PackageValidationDependsOn``` ? We will need to provide this hook for our customers in order for them to do stuff with their packages like aspnetcore.

also do we need to provide an option for users to run without references ? I dont think of a scenario where we will need to run this without references. just bringing it up for discussion.

src/Compatibility/Microsoft.DotNet.PackageValidation/CompatibleFrameworkInPackageValidator.cs
```diff
@@ -16,10 +16,10 @@ namespace Microsoft.DotNet.PackageValidation
     public class CompatibleFrameworkInPackageValidator
     {
         private readonly ApiCompatRunner _apiCompatRunner;
-
-        public CompatibleFrameworkInPackageValidator(string noWarn, (string, string)[] ignoredDifferences, bool enableStrictMode, IPackageLogger log)
+
```
Anipik, line 19: nit: extra space

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -87,6 +99,58 @@ public void RunApiCompat()
...
         }
 
+        private IAssemblySymbol GetAssemblySymbolFromStream(Stream assemblyStream, MetadataInformation assemblyInformation, out bool resolvedReferences)
+        {
+            resolvedReferences = false;
+            HashSet<string> directories = null;
+
+            // In order to enable reference support for baseline suppression we need a better way
+            // to resolve references for the baseline package. Let's not enable it for now.
+            bool shouldResolveReferences = !_isBaselineSuppression && _referenceDirectories != null &&
```
Anipik, line 109: how about the users manually providing the references for the baseline package ? should we add that support ?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -87,6 +99,58 @@ public void RunApiCompat()
...
+                foreach (AssemblyLoadWarning warning in warnings)
+                {
+                    _log.LogWarning(
+                        new Suppression()
+                        {
+                            DiagnosticId = warning.DiagnosticId,
+                            Target = warning.ReferenceId
+                        },
+                        warning.DiagnosticId,
+                        "When loading: '{0}': {1}",
```
Anipik, line 145: nit: needs to be localized ?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
```diff
@@ -12,13 +13,35 @@
 
 namespace Microsoft.DotNet.ApiCompatibility
 {
+    public class AssemblyLoadWarning : IDiagnostic, IEquatable<AssemblyLoadWarning>
```
Anipik, line 16: Missing triple slash comments

Anipik:
We need to add some end to end tests here.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -37,6 +37,11 @@ public class ApiComparer
         /// </summary>
         public bool StrictMode { get; set; }
 
+        /// <summary>
+        /// Flag indicating whether the ApiComparison will be performed using references or not.
+        /// </summary>
+        public bool RunningWithReferences { get; set; }
```
joperezr, line 43: Seems like RunningWithReferences is both here and in ComparingSettings, so perhaps we should remove them from here

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
joperezr: My latest change moves to use ReferencePath. I still have the fallback to go to containing directories in order to a) not brake one API that supports adding directories to search assemblies there and b) to be able to find transitive dependencies that are not part of referencepath but happen to be on the same directories as referencePath assemblies. This latter case is rare, and usually only happens when dealing with a full facade.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
joperezr: Which technique in specific? Is it the fact that we don't build the project references or the DependsOnTargets hooks we picked?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
joperezr: I will measure perf and report back, but so far I haven't seen a big drawback here when running these changes on dotnet/runtime so I suspect the perf hit is minimal.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
joperezr: yeah, @ericstj mentioned adding that extensibility point as well. I agree and we should add it, let's just do it in a separate PR so it's easier to review.

joperezr:
As an FYI, this is the way that dotnet/runtime CompatibilitySuppressions.xml files would look like after taking in this change: https://github.com/joperezr/runtime/commit/df2f3c5faaef163a7c1789b5789270098390b1c8

As you can see, we are cleaning a few CompatibilitySuppressions.xml files which were only there due to not being able to follow type-forwards. We also now detect other compat differences on full facades since we can now follow those type-forwards and detect some diffs between some of our implementations.

Anipik:
```xml 

<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <Suppression>
    <DiagnosticId>CP1002</DiagnosticId>
    <Target>System.Runtime.CompilerServices.Unsafe.dll</Target>
  </Suppression>
</Suppressions>  
```

can we just not add them to the suppressions file and just use noWarn to disable these warning ?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
```diff
@@ -260,7 +316,7 @@ public IEnumerable<IAssemblySymbol> LoadMatchingAssemblies(IEnumerable<IAssembly
                 if (warnOnMissingAssemblies && !found)
                 {
                     string assemblyInfo = validateMatchingIdentity ? assembly.Identity.GetDisplayName() : assembly.Name;
-                    _warnings.Add(string.Format(Resources.MatchingAssemblyNotFound, assemblyInfo));
+                    _warnings.Add(new AssemblyLoadWarning(DiagnosticIds.AssemblyNotFound, assemblyInfo, string.Format(Resources.MatchingAssemblyNotFound, assemblyInfo)));
```
Anipik, line 319: why are we storing these warnings ? wouldnt it better if we just log them

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -54,22 +57,31 @@ public void RunApiCompat()
...
                 int counter = 0;
                 foreach ((MetadataInformation, MetadataInformation, IEnumerable<CompatDifference> differences) diff in differences)
                 {
                     (MetadataInformation rightAssembly, string header) rightTuple = _dict[left][counter++];
-                    _log.LogMessage(MessageImportance.Low, rightTuple.header);
 
+                    bool logHeaderMessage = true;
                     foreach (CompatDifference difference in diff.differences)
                     {
+                        if (logHeaderMessage)
```
Anipik, line 79: Why are we logging this for all the differences ? wouldnt it be the same value for all ?

joperezr:
> can we just not add them to the suppressions file and just use noWarn to disable these warning ?

Do you mean once we consume these changes in dotnet/runtime and just suppress that one? We could if we wanted to, but that means of course that we would suppress all assemblies not found, so I think that individually suppressing is better in order to fully understand and not hide future issues

src/Compatibility/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
joperezr: Well the only reason that I see is that the AssemblySymbolLoader doesn't have a reference to the logger so, at least as of today, the way it works is that main function of the loader is to try to load the assemblies, and the runner which calls it, has the reference to the logger so it is able to gather the warnings and throw them. I don't fundamentally see a problem of passing the logger down to the loader as well, but I also don't see a problem with the way we have it working now, so unless there is a big reason why we would want to change it, I would keep it as is.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
joperezr: Not sure what you mean, what this if-check is doing is making sure that we only print the header once for each group of differences (from the rightTuple) which seems correct to me.

Anipik: The header tells which assemblies are being compared and has no information specific to api compat differences.

joperezr: Correct. What I am saying is that we are still only logging the header once per left-right tuple, since when looping through the differences we log it the first time but then set `logHeaderMessage` to false so that we don't log it more than once. The only thing this change is doing, is delaying the time we log the header to when we actually know that diff.differences is not empty, which again, seems correct to me.

Anipik (approved):

joperezr:
I think this is now ready to go. Things that will be done in a separate PR are:
- Attempt to not warn for missing dependencies when they are actually not needed for the analysis.
- Provide a switch (MSBuild property) to disable the use of references by APICompat
- Enable references for baseline validation.

PR #19786, opened 2021-08-17
Update ValidateExecutableReferences with new schema

Forgind:
Opts into using the new schema and looks for a TF with its first attribute equal to nearestTargetFramework.

There are multiple ways to do this, so I'm open to suggested improvements 🙂

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1082,7 +1082,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="ValidateExecutableReferences" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
   <PropertyGroup>
-    <_UseAttributeForTargetFrameworkInfoPropertyNames>true</_UseAttributeForTargetFrameworkInfoPropertyNames>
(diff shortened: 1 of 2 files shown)


dsplaisted:
This is failing a bunch of tests on full framework, presumably because it's running on an older version of MSBuild which doesn't have dotnet/msbuild#6699.

Can we condition setting `_UseAttributeForTargetFrameworkInfoPropertyNames` on whether the MSBuild version is 17.0 or higher, and also support both schemas in `ValidateExecutableReferences`?

Forgind:
I'm confused—doesn't the SDK come with its own MSBuild? I would've thought that would mean that if your SDK is new enough to have this change, it would be new enough to have 6699, too. Had the main I based this on not been updated with 6699 yet or something?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1089,6 +1093,7 @@ Copyright (c) .NET Foundation. All rights reserved.
       SelfContained="$(SelfContained)"
       IsExecutable="$(_IsExecutable)"
       ReferencedProjects="@(_MSBuildProjectReferenceExistent)"
+      MSBuildVersion="$(MSBuildVersion)"
```
dsplaisted, line 1096: Could we pass the value of `_UseAttributeForTargetFrameworkInfoPropertyNames` in here instead of the MSBuild version?  Then we wouldn't have two places where we're checking whether we're on version 17 or higher.

dsplaisted (approved):

marcpopMSFT:
@dsplaisted should this be in RC2?

dsplaisted:
> @dsplaisted should this be in RC2?

Yes, I would go ahead and merge it.

marcpopMSFT:
@Forgind , can you prep for RC2 then?

PR #19791, opened 2021-08-17
Disable processing in RazorSourceGenerator when no razor files are involved

### LINKED ISSUE #35404: RazorSourceGenerator should no-op more aggressively in web projects that do not contain any razor files
In a project with any razor files, the RSG executes and does some throwaway work calculating things like TagHelperDescriptors. The incrementalism caches these results (and producing a bunch of Razor content that never gets used), but we should be avoiding this entirely by no-oping this.

pranavkm:
We disabled processing the RSG when it's disabled via Razor's tooling editorconfig. Another scenario
where we do a bunch of extra work is when building webapps that do not contain any Razor (.razor / .cshtml) files.

This change updates the generator to no-op and avoid some allocations / work if no razor files are present.

Fixes https://github.com/dotnet/aspnetcore/issues/35404

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
@@ -30,6 +30,9 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
             var sourceItems = sourceItemsWithDiagnostics
                 .ReportDiagnostics(context);
 
+            var hasRazorFiles = sourceItems.Collect()
+                .Select(static (sourceItems, _) => sourceItems.Any());
+
             var importFiles = sourceItems.Where(static file =>
             {
                 var path = file.FilePath;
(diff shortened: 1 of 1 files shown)


src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -101,16 +114,45 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
 
             var tagHelpersFromReferences = context.CompilationProvider
                 .Combine(razorSourceGeneratorOptions)
-                .WithLambdaComparer(static (a, b) => a.Left.References.SequenceEqual(b.Left.References), static a => a.Left.References.GetHashCode())
+                .Combine(hasRazorFiles)
+                .WithLambdaComparer(static (a, b) =>
+                {
+                    var ((compilationA, razorSourceGeneratorOptionsA), hasRazorFilesA) = a;
+                    var ((compilationB, razorSourceGeneratorOptionsB), hasRazorFilesB) = a;
```
chsienki, line 121:
```suggestion
                    var ((compilationB, razorSourceGeneratorOptionsB), hasRazorFilesB) =  b;
```

PR #19804, opened 2021-08-18
Preserve environment variable previously configured on ProcessStartInfo

### LINKED ISSUE #35410: dotnet watch stomps over existing hosting startup assemblies
This is a regression on 6.0 that affects the ability for apps to provide their own hosting startup assemblies:
```csharp
ASPNETCORE_HOSTINGSTARTUPASSEMBLIES=$Microsoft.AspNetCore.Watch.BrowserRefresh
```

When hosting startup assemblies are defined in launchSettings.json (as the SPA templates do) they get stomped over.

### LINKED ISSUE #35427: dotnet-watch: Hot reload will hang indefinitely for blazor wasm if browser does not launch
While the behavior is expected, there's no indication that it's waiting for the browser to launch. We could print something to make it a bit more obvious why it's hanging.

pranavkm:
* dotnet-watch concats values from Environment, but ignores environment values configured on ProcessStartInfo
typically configured by launchSettings.json.

* Print a message that says we're waiting for a browser connection.

Fixes https://github.com/dotnet/aspnetcore/issues/35410
Fixes https://github.com/dotnet/aspnetcore/issues/35427

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs
+++ b/src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs
@@ -121,7 +121,11 @@ namespace Microsoft.DotNet.Watcher.Tools
             await _terminateWebSocket.Task;
         }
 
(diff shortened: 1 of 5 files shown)


javiercn (approved):
Looks great!

PR #19840, opened 2021-08-18
Follow up to global usings changes

### LINKED ISSUE #19696: Remove ImplicitUsings=enable from test after stage 0 SDK is updated
The SDK tests creates a project from a stage 0 SDK which is affected by the ImplicitUsings change. The template has been updated, but we need a new stage 0 for the test to pass. In the meanwhile, the test is tweaked to allow it to continue working without being skipped.

### LINKED ISSUE #19793: Ensure <Using /> declarations from in-box SDKs are conditioned on C#
The new `<Using />` item type is explicitly intended to only be used in C# projects. The [code generation targets](https://github.com/dotnet/sdk/blob/f2cc77b60cdc2a4e8b393adf7b855b478ad79671/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L1131) and [core global usings items](https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props#L27) are already conditioned on C# but we should ensure that the higher-level SDKs (e.g. Web, Worker, WindowsDesktop, etc.) also condition any using item types they bring in by C# too.

- Current Web props: https://github.com/dotnet/sdk/blob/main/src/WebSdk/Web/Sdk/Sdk.props#L61
- Current Worker props: https://github.com/dotnet/sdk/blob/main/src/WebSdk/Worker/Targets/Microsoft.NET.Sdk.Worker.props#L41
- Current BlazorWasm props: https://github.com/dotnet/sdk/blob/main/src/BlazorWasmSdk/Sdk/Sdk.props#L23
- Current WinForms props: https://github.com/dotnet/winforms/blob/main/pkg/Microsoft.Private.Winforms/sdk/dotnet-wpf/Microsoft.NET.Sdk.WindowsDesktop.WindowsForms.props#L15
- Current WPF props: https://github.com/dotnet/wpf/blob/main/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.WPF.props#L5

pranavkm:
* When reverting changes to the VisualBasic.targets file, a property initialization was missed. This would have introduced
a regression for users upgrading to the 6.0 SDK when they had global imports disabled in their apps.
* Only add implicitly enabled imports for C# projects.
* Update the stage 0 SDK and re-enable skipped / altered tests

Fixes https://github.com/dotnet/sdk/issues/19793
Fixes https://github.com/dotnet/sdk/issues/19696

### DIFF
--- a/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
+++ b/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
@@ -3,7 +3,11 @@
 
 #nullable disable
 
+using System;
+using System.Collections.Generic;
+using System.IO;
 using System.IO.Pipes;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
@@ -10,6 +10,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <DisableImplicitNamespaceImports Condition="'$(DisableImplicitNamespaceImports)'==''">$(DisableImplicitFrameworkReferences)</DisableImplicitNamespaceImports>
+  </PropertyGroup>
   <ItemGroup Condition=" '$(DisableImplicitNamespaceImports)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <!-- These namespaces are present in 2.0 Framework assemblies -->
     <Import Include="Microsoft.VisualBasic" />
--- a/src/Tests/Microsoft.DotNet.PackageInstall.Tests/SampleGlobalTool/consoledemo.csproj
+++ b/src/Tests/Microsoft.DotNet.PackageInstall.Tests/SampleGlobalTool/consoledemo.csproj
@@ -10,5 +10,6 @@
      <EnableSourceControlManagerQueries>false</EnableSourceControlManagerQueries>
      <EnableSourceLink>false</EnableSourceLink>
      <DeterministicSourcePaths>false</DeterministicSourcePaths>
+     <CheckEolTargetFramework>false</CheckEolTargetFramework>
   </PropertyGroup>
 </Project>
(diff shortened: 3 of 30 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets
```diff
@@ -10,6 +10,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
```
pranavkm, line 13: This was in the 5.0 SDK - got missed when undoing the changes to VB targets - https://github.com/dotnet/sdk/blob/release/5.0.1xx/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.VisualBasic.targets#L13-L15

src/Tests/Microsoft.DotNet.PackageInstall.Tests/SampleGlobalTool/consoledemo.csproj
```diff
@@ -10,5 +10,6 @@
      <EnableSourceControlManagerQueries>false</EnableSourceControlManagerQueries>
      <EnableSourceLink>false</EnableSourceLink>
      <DeterministicSourcePaths>false</DeterministicSourcePaths>
+     <CheckEolTargetFramework>false</CheckEolTargetFramework>
```
pranavkm, line 13: The latest rc1 SDK complains about EOL for 2.1 apps.

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -3,7 +3,11 @@
 
 #nullable disable
 
+using System;
```
pranavkm, line 6: Re-adding global usings since they're no longer implicitly added.

dsplaisted (approved):

pranavkm:
> Expected collection {"macos"} to be equivalent to {"microsoft-android-sdk-full"}, but it misses {"microsoft-android-sdk-full"}.

Is this because of a workload rename? (@sfoslund?)

sfoslund:
@pranavkm most if not all of the tests should use fake manifests, not the ones actually installed in stage 0, so the rename shouldn't theoretically matter. Do you have a link to this test failure? Is it persistent?

pranavkm:
It looks like it failed a second time with the same error - https://dev.azure.com/dnceng/public/_build/results?buildId=1302241&view=ms.vss-test-web.build-test-results-tab&runId=38369588&paneView=debug&resultId=101122.

sfoslund:
Hmm I don't see any cause right away, but it sounds like @wli3 is going to take a look at the test failures since he's domestic cat and added these tests.

Issue #19514, opened 2021-08-04
Add differentiation between long and short property options in run command


sfoslund:
See https://github.com/dotnet/sdk/pull/19402#discussion_r680184127

sfoslund:
Fixed by https://github.com/dotnet/sdk/pull/19909

PR #19909, opened 2021-08-19
Differentiate between long and short run property options

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/19514

### DIFF
--- a/src/Cli/dotnet/ParseResultExtensions.cs
+++ b/src/Cli/dotnet/ParseResultExtensions.cs
@@ -131,7 +131,7 @@ namespace Microsoft.DotNet.Cli
             return properties.Where(property => !property.Contains("="));
         }
 
-        public static IEnumerable<string> GetRunCommandShorthandPropertyValues(this ParseResult parseResult)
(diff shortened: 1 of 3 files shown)


src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -127,14 +127,25 @@ public static bool UsingRunCommandShorthandProjectOption(this ParseResult parseR
 
         public static IEnumerable<string> GetRunCommandShorthandProjectValues(this ParseResult parseResult)
         {
-            var properties = parseResult.ValueForOption(RunCommandParser.PropertyOption);
+            var properties = GetRunPropertyOptions(parseResult, true);
             return properties.Where(property => !property.Contains("="));
         }
 
         public static IEnumerable<string> GetRunCommandShorthandPropertyValues(this ParseResult parseResult)
```
dsplaisted, line 134: `GetRunCommandShorthandPropertyValues` doesn't seem like the best name here, because it returns both the properties that were passed using the shorthand and longhand form.

dsplaisted (approved):
It looks l

PR #19951, opened 2021-08-20
Avoid assuming RSG in non-C# projects

pranavkm:
* [x] While it's generally no-op to define it, we can be correct and only configure UseRazorSourceGenerator for C# projects where they're supported.

### DIFF
--- a/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
+++ b/src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
@@ -49,9 +49,7 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_TargetingNETCoreApp30OrLater>true</_TargetingNETCoreApp30OrLater>
         <_TargetingNET50OrLater>true</_TargetingNET50OrLater>
         <_TargetingNET60OrLater>true</_TargetingNET60OrLater>
-        <UseRazorSourceGenerator Condition="'$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
-        <!-- Required to support incremental source generator -->
-        <LangVersion Condition="'$(UseRazorSourceGenerator)' != '' ">preview</LangVersion>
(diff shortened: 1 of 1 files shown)


adrianwright109:
Looks like this also fixes https://github.com/dotnet/aspnetcore/issues/33848

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
```diff
@@ -49,9 +49,7 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_TargetingNETCoreApp30OrLater>true</_TargetingNETCoreApp30OrLater>
         <_TargetingNET50OrLater>true</_TargetingNET50OrLater>
         <_TargetingNET60OrLater>true</_TargetingNET60OrLater>
-        <UseRazorSourceGenerator Condition="'$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
-        <!-- Required to support incremental source generator -->
-        <LangVersion Condition="'$(UseRazorSourceGenerator)' != '' ">preview</LangVersion>
+        <UseRazorSourceGenerator Condition="'$(Language)' == 'C#' AND '$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
```
pranavkm, line 52:
```suggestion
        <!-- Required to support incremental source generator -->
        <LangVersion Condition="'$(UseRazorSourceGenerator)' != '' ">preview</LangVersion>
        <UseRazorSourceGenerator Condition="'$(Language)' == 'C#' AND '$(UseRazorSourceGenerator)' == '' ">true</UseRazorSourceGenerator>
```

pranavkm:
@adrianwright109 unfortunately not. The compiler still seems to require a LangVersion=preview for incremental source generators.

@captainsafia could I get a quick review? I'd like to try and get this ported to rc1.

captainsafia (approved):

PR #20063, opened 2021-08-24
Only show VS workload list header on windows

sfoslund:


joeloff (approved):
I assume if we have data to add we'd do that conditionally as well when the header is added

sfoslund:
> I assume if we have data to add we'd do that conditionally as well when the header is added

Yup, I believe so.

KalleOlaviNiemitalo:
Is Visual Studio for Mac able to install .NET SDK workloads? If it is, does `dotnet workload list` display those?

sfoslund:
> Is Visual Studio for Mac able to install .NET SDK workloads? If it is, does dotnet workload list display those?

Yes, VS for mac will also be able to install workloads but due to differences in the way those experiences are implemented, workload list can show the workloads installed via VS for mac but not VS on windows.

PR #20067, opened 2021-08-24
Don't require elevation for workload list

sfoslund:


joeloff (approved):
Is search OK?

sfoslund:
> Is search OK?

Yup, on mac it is

PR #20136, opened 2021-08-25
Dispose open handle on the package

Anipik:
This is dangling reference and we are hitting file contetnion issues because of this while trying to use this package in aspnet core repo

Tested locally and it no longer repros


joperezr (approved):
Small comment left but this looks good otherwise. Good catch.

PR #20153, opened 2021-08-25
Project Reference support for WinRT Components

j0shuams:
## Description

Note: this a rebase of the same PR made for .NET 5 -- https://github.com/dotnet/sdk/pull/19218

I am proposing we add a property in an existing target, which was made for WinRT support recently. As well as add a new target that uses the property set in the existing target to add a reference. We need this to support C# project reference consumers. Because C# consumers only get a .winmd via GetTargetPath, they do not have the assembly containing the managed implementation, so we annotate the .winmd to have the path to the implementation, and we add it as a reference here.

## Customer Impact

Native app (C++) developers should be able to reference WinRT components made with C#/WinRT simply by adding a project reference. This PR removes the need for a project reference and a reference to the winmd made by the project -- the latter of which required an extra build step before the reference could be added. 

## Regression?

 - [ ] Yes
 - [X] No

## Risk 

 - [ ] High
 - [ ] Medium
 - [X] Low

## Verification 

- [x] Manual (required)
- [ ] Automated

I have opened an issue to track adding the automated validation - https://github.com/microsoft/CsWinRT/issues/958

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Windows.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Windows.targets
@@ -58,6 +58,14 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Target Name="RemoveManagedWinRTComponentWinMDReferences"
           AfterTargets="ResolveProjectReferences">
+    
+    <ItemGroup>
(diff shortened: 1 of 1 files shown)


dsplaisted (approved):
It would be good to link the previous PR.  Also FYI, you can change the target branch of an existing PR with the "edit" button, though sometimes that picks up commits you don't want so you need to make sure it did the right thing afterwards.

Issue #20195, opened 2021-08-26
.NET stack temporary folder cannot be accessed after running under sudo


wli3:
Issues like 
https://github.com/dotnet/sdk/issues/19823
https://github.com/dotnet/sdk/issues/19523

.NET stack creates many temporary folder as part of the command execution. In many cases, if the command is running under sudo, the temp directory will be shared with the non-sudo user. If files or folders are shared between sudo and non sudo command. And if the command is first run under sudo, the folder would be default set to ROOT user access only. So then next command run without sudo will not be able to access.

Most of these temporary folder use $HOME as base, while sudo on mac does not change $HOME when run under sudo. https://unix.stackexchange.com/a/91572

Temp folder potentially have the issues:
### runtime 

- dotnet folder under Path.GetTempPath() (Path.GetTempPath() would be changed to /tmp/ vs user dir under sudo. So that is currently fine)

### sdk

- .dotnet folder under $HOME
- sdk-advertising-temp under Path.GetTempPath()

### nuget

- Folders under $HOME https://github.com/NuGet/NuGet.Client/blob/538727480d93b7d8474329f90ccb9ff3b3543714/src/NuGet.Core/NuGet.Configuration/Utility/SettingsUtility.cs#L15-L20
- NuGetScratch under Path.GetTempPath() https://github.com/NuGet/NuGet.Client/blob/1d75910076b2ecfbe5f142227cfb4fb45c093a1e/src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs#L82

### template engine

- .templateengine folder under $HOME

## short term solution

While we have this behavior since 1.0. There are not many scenarios require .NET run under sudo. This issue become significate when we integrate with mac installer for workload install. It need to be run under sudo. We decided to override $HOME directory to `/tmp/dotnet_sudo_home/` when we detect the command is running under sudo. And therefor "simulate" the behavior `sudo -i`. So sudo vs non sudo command won't share the same directory

## long term solution

Not fully discussed. One direction is to have an file system abstraction across the stack.

PR #20196, opened 2021-08-26
Add sudo home override

### LINKED ISSUE #19523: /Users/USERNAME/.dotnet/sdk-advertising-temp/ cannot be accessed
I get this when I run `dotnet workload list --machine-readable` after installing with sudo.

I think it is due to this temp directory was created under sudo, so when run without sudo it cannot be accessed. We might want to separate folder for different elevation or set explicit permission for this folder. Note, on different Linux, $HOME under sudo could be the current user or the root depends on how sudo the application is configed.

### LINKED ISSUE #19823: dotnet workload install creates dir with root permissions
Steps to reproduce:
Clean env
 Install https://dotnet.microsoft.com/download/dotnet/thank-you/sdk-6.0.100-preview.7-macos-x64-installer
Execute sudo `dotnet workload install maui ` or `dotnet workload install wasm-tools`  or any of the .NET 6 workloads  from command line.
Check 'dotnet new mvc --auth Individual -o mvcind` after maui install.

Expected Behaviour:
mvc project gets creates and restores successfully

Actual Behaviour:
mvc project gets created but fails to restore 
Cause: `dotnet workload install`  creates ~/.nuget with  root permissions and when `dotnet new `runs for an individual auth,  permission is denied.

wli3:
https://github.com/dotnet/sdk/issues/20195

Fix #19823
Fix #19523

### DIFF
--- a/src/Cli/dotnet/Program.cs
+++ b/src/Cli/dotnet/Program.cs
@@ -34,6 +34,13 @@ namespace Microsoft.DotNet.Cli
             TimeSpan startupTime = mainTimeStamp - Process.GetCurrentProcess().StartTime;
 
             bool perfLogEnabled = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_PERF_LOG", false);
+
+            // Avoid create temp directory with root permission and later prevent access in non sudo
+            if (SudoEnvironmentDirectoryOverride.IsRunningUnderSudo())
+            {
+                perfLogEnabled = false;
+            }
+
             PerformanceLogStartupInformation startupInfo = null;
             if (perfLogEnabled)
             {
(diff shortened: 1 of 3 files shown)


sfoslund (approved):

GouriKumari:
@wli3  : After installing rc1 in a clean machine, dotnet workload list --machone-readable needs to be executed with elevated privileges. Is that expected?

sfoslund:
@GouriKumari I fixed this issue a few weeks ago, but I believe it missed rc1. You shouldn't see the problem in rc2: https://github.com/dotnet/sdk/pull/20067

GouriKumari:
@sfoslund : Is there time to include it in rc1?  #20196 Is fix from this PR also missed rc1?

sfoslund:
I don't think so, it looks like we already have a candidate build.

PR #20209, opened 2021-08-26
Test driven RazorSourceGenerator incrementality

pranavkm:
* Add some instrumentation and tests to validate RazorSourceGenerator's incrementality
* Do not codegen declaration for .cshtml files.
* Update the declaration caching to be more granular, and also cache SyntaxTrees
* Cache tag helpers from compilation correctly

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGeneratorEventSource.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGeneratorEventSource.cs
@@ -5,7 +5,7 @@ using System.Diagnostics.Tracing;
 
 namespace Microsoft.NET.Sdk.Razor.SourceGenerators
 {
-    [EventSource(Name = "Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator")]
+    [EventSource(Name = "Microsoft-DotNet-SDK-Razor-SourceGenerator")]
     internal sealed class RazorSourceGeneratorEventSource : EventSource
     {
         public static readonly RazorSourceGeneratorEventSource Log = new();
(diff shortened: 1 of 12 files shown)


eerhardt (commented):
Just an FYI that https://github.com/dotnet/roslyn/pull/55892 is adding EventSource events around calling each SourceGenerator. I see your events are more granular, so they both can make sense. Just wanted you to be aware that we'll get events for all source generators.

pranavkm:
@marcpopMSFT could I bother you to merge this in to main: The changes were [merged in to rc2](https://github.com/dotnet/sdk/pull/20264) but main is blocked due to the failing test.

src/RazorSdk/SourceGenerators/RazorSourceGeneratorEventSource.cs
```diff
@@ -0,0 +1,137 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.NET.Sdk.Razor.SourceGenerators
+{
+    [EventSource(Name = "Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator")]
```
brianrob, line 8: The standard pattern is to use dashes in the EventSource name (e.g. Microsoft-DotNet-SDK-Razor-SourceGenerator or something like that).

src/RazorSdk/SourceGenerators/RazorSourceGeneratorEventSource.cs
```diff
@@ -0,0 +1,137 @@
...
+        public static readonly RazorSourceGeneratorEventSource Log = new();
+
+        private RazorSourceGeneratorEventSource() { }
+
+        private const int ComputeRazorSourceGeneratorOptionsId = 1;
+
+        [Event(ComputeRazorSourceGeneratorOptionsId, Level = EventLevel.Informational)]
+        public void ComputeRazorSourceGeneratorOptions()
+        {
+            if (IsEnabled())
```
brianrob, line 20: Each `[Event]` method is going to call `IsEnabled` as soon as it enters `WriteEvent`, so you don't need to bother wrapping the call to `WriteEvent` in the if check.  If the computation of the arguments is expensive, then you should wrap that computation and the call to write the event in a call to `IsEnabled`.

brianrob (approved):
I just looked at the `EventSource` changes.  Overall looks good - a couple of things, but nothing earth shattering.

brianrob (approved):

PR #20256, opened 2021-08-27
[release/6.0.1xx] Disable CA1416 for ProcessExtensions

dseefeld:
This is causing issues for source-build, if a newer version of CA is used.


MichaelSimons (approved):

joeloff (approved):

dseefeld:
> For future reference for myself, where are these issues flagged?

Full source-build build is done in the installer repo.  These are showing up there because when building all repos from src, we use an updated (just built) version of CA when building SDK.

PR #20271, opened 2021-08-27
Hide workload list options only relevant to machine readable

sfoslund:


src/Cli/dotnet/OptionForwardingExtensions.cs
```diff
@@ -44,6 +44,12 @@ public static Option<T> AllowSingleArgPerToken<T>(this Option<T> option)
             return option;
         }
 
+        public static Option<T> MakeHidden<T>(this Option<T> option)
```
joeloff, line 47: Change looks good, wondering if we should call this Hide instead

sfoslund: Sure, just renamed.

sfoslund:
@joeloff anything else to do here or is it good to go?

joeloff (approved):
LGTM, thanks

PR #20367, opened 2021-08-31
Fix null ref and client shutdown

joeloff:
**Description**: MSI payload packages can be missing from the package feeds, e.g. Maui doesn't push regular updates, When a manifest update breaks because of this and a rollback is triggered to perform a downgrade of a package that wasn't installed correctly, the missing registry key can result in a null reference, A side effect of this results in attempting to shutdown the elevated service which runs after the command has completed, causing the exception to be raised by dotnet.exe instead.

**Fixes**: https://github.com/dotnet/sdk/issues/20143

**Risk**: Low

**Validation**: Manual and automated

**Regression**: No, new feature for RC1


joeloff:
@dsplaisted I'll bring this to tactics tomorrow

sfoslund (approved):

joeloff:
Approved by tactics over mail

Issue #20357, opened 2021-08-30
Hide `--sdk-version` parameter to workload commands


dsplaisted:
The `--sdk-version` parameter to workload commands is intended for VS for Mac to use in order to download workloads before a new version of the .NET SDK is installed.

It isn't intended to be used from the command line, and it doesn't work for MSI-based installs.  So we should hide it from the list of options displayed in help.

@tdykstra We should also remove this from the documentation

sfoslund:
Fixed by #20387

Issue #20283, opened 2021-08-28
Improve command-line help for `dotnet workload install` `--download-to-cache`


dsplaisted:
From the `dotnet workload install` help:

```
--download-to-cache <download-to-cache>  Download packages needed to install a workload to a folder that can be used for offline installation.
```

The parameter value should be displayed as `<cache-folder>` or similar instead of `<download-to-cache>`.

`--from-cache` should similarly be updated.

sfoslund:
Fixed by #20387

PR #20387, opened 2021-08-31
Update workload help text

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/20357
Fixes https://github.com/dotnet/sdk/issues/20283


sfoslund:
@tdykstra this PR hides the `--sdk-version` option (which is present on most of the dotnet workload commands) from the help, so it should be removed from the docs as well. Would you be able to make that change?
I also hid all of the options on workload list besides verbosity and help in [this PR](https://github.com/dotnet/sdk/pull/20271), so if you could make those updates as well that would be great.

dsplaisted (approved):

PR #20390, opened 2021-08-31
Fix missing packaged shim after update

wli3:


marcpopMSFT:
FYI, we're not shipping 5.0.3xx anymore.

PR #20415, opened 2021-08-31
Merge release/6.0.1xx rc1

wli3:
Need to make sure all loc are there for before the loc pass

PR #20468, opened 2021-09-01
Fix bad merge re-apply de34d98c4f282888a257b2941c46e5baaa2cc2f9

wli3:
@captainsafia sorry I merge too soon. Here is the fix to bring it back


captainsafia (approved):
Thanks!

cc: @pranavkm Will is fixing a bad merge per https://github.com/dotnet/sdk/pull/20415#discussion_r699729727.

PR #20489, opened 2021-09-01
Update `dotnet watch` description

### LINKED ISSUE #20487: (dotnet watch) - example command is invalid
In the help output for `dotnet watch`:

```
  For example: dotnet watch -- --verbose run

  Even though '--verbose' is an option dotnet-watch supports, the use of '--'
  indicates that '--verbose' should be treated instead as an argument for
  dotnet-run.
```

`dotnet watch -- --verbose run` will fail.  Either:

- This was meant to be an example of an incorrect command.  If that's the case, the help message isn't very clear about that.
- OR, the example is just wrong.

### LINKED ISSUE #40472: (dotnet watch) - example command is invalid
In the help output for `dotnet watch`:

https://github.com/dotnet/sdk/blob/e515ca20d54355f130f4ff02944870875e17ab12/src/BuiltInTools/dotnet-watch/Program.cs#L48-L52

`dotnet watch -- --verbose run` will fail.  Either:

- This was meant to be an example of an incorrect command.  If that's the case, the help message isn't very clear about that.
- OR, the example is just wrong.

Shamus03:
Fixes dotnet/aspnetcore#40472

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Program.cs
+++ b/src/BuiltInTools/dotnet-watch/Program.cs
@@ -49,7 +49,8 @@ Remarks:
 
   Even though '--verbose' is an option dotnet-watch supports, the use of '--'
   indicates that '--verbose' should be treated instead as an argument for
-  dotnet-run.
+  dotnet-run. And because '--verbose' is not a valid argument for dotnet-run,
+  this command would fail.
 
 Examples:
   dotnet watch run


KalleOlaviNiemitalo:
Well, that explanation isn't right either.  On 6.0.100-preview.7.21379.14:

```
$ dotnet watch -- --verbose run
watch : Started
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet---verbose does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
watch : Exited with error code 1
watch : Waiting for a file to change before restarting dotnet...
```

`dotnet watch` tried to run `dotnet --verbose run` rather than `dotnet run --verbose`, so it doesn't even matter whether `dotnet run` supports `--verbose`.  You can get a similar error with `dotnet watch --no-build run` even though `dotnet run` supports `--no-build`.

It seems the only option recognized by `dotnet run` that is also listed in _runtime-options_  or _sdk-options_ for `dotnet` is `-h|--help`. So you could use `dotnet watch -- --help` to run `dotnet --help` repeatedly, instead of asking for help on `dotnet watch`.  Such a command does not look useful, though.

I don't think users ever need the `--` on `dotnet watch` when using this SDK version.  However, if a future SDK version makes `dotnet watch` itself support a `--diagnostics` option, then `dotnet watch --diagnostics run` and `dotnet watch -- --diagnostics run` will mean different things. So there could be two reasons to use `--`:

- To make it easier for people to understand whether the option is intended for `dotnet run` itself or for the repeated `dotnet` command.
- To make the command more likely to keep working in a future version of .NET SDK, where `dotnet run` might recognize more options.

Shamus03:
So should the example command just be updated to `dotnet watch -- run --verbose`?

Or should the documentation be updated to say "`Even though '--verbose' is an option dotnet-watch supports, the use of '--'
  indicates that '--verbose' should be treated instead as an argument for
  dotnet (the top-level command). And because '--verbose' is not a valid argument for dotnet,
  this command would fail.`"

Shamus03:
The specific command that brought up this confusion was:

```
dotnet watch test -v n
```

Intending to watch/run `dotnet test -v n`.  Instead, it was essentially interpreted as `dotnet watch --verbose test n`, which caused an error and much confusion.

pranavkm:
In the event `dotnet run` adds support for `--verbose` at a later time, this description would be incorrect. The current text describes the behavior, which I think is sufficient. Closing this, but thanks for the PR

PR #20567, opened 2021-09-02
Only warn about references that are actually needed at runtime

joperezr:
Today if you are running Package Validation with references enabled, we will warn for any assembly that is part of the full closure and is not found in the provided search directories; independently of it is actually required for the analysis or not. This can sometimes bring noise to the comparison, and emit warnings for assemblies that are not important during the analysis. The changes in this PR will change it so that we not warn for all assemblies that can't be found, but instead only do so when the fact that they are not loaded is preventing the full analysis (for example, can't follow a type forward, or can't get the full type base types and interfaces.)

I will run these changes against dotnet/runtime to see if these changes do remove some of the noise we were seeing.


src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -37,6 +37,11 @@ public class ApiComparer
         /// </summary>
         public bool StrictMode { get; set; }
 
+        /// <summary>
+        /// Flag indicating whether the ApiComparison will be performed using references or not.
+        /// </summary>
+        public bool RunningWithReferences { get; set; }
```
ericstj, line 43: This feels different than other parameters which indicate a preferences, this is named to represent state.  Perhaps we can name it something like "WarnOnMissingReferences" or "WarnOnUnresolvedTypes" or "TreatUnresolvedTypesAsError|Warning"

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
```diff
@@ -85,6 +100,23 @@ private void ValidateInterfaceNotRemoved(ITypeSymbol left, ITypeSymbol right, st
...
+                if (rightInterface.TypeKind == TypeKind.Error && Settings.WithReferences)
+                {
+                    AddAssemblyLoadError(differences, rightInterface);
+                }
+            }
+        }
+
+        private void AddAssemblyLoadError(IList<CompatDifference> differences, ITypeSymbol type)
+        {
+            differences.Add(new CompatDifference(
```
ericstj, line 115: Will this ensure only a single diagnostic per assembly?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
joperezr: Makes sense, will change.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotRemoveBaseTypeOrInterface.cs
joperezr: it won't. Good catch, I'll fix this and add a test for it.

joperezr:
@ericstj @ViktorHofer after these set of changes, I ran a build of dotnet/runtime and enabled generating suppression files, and this is the new diff after these changes are in: https://github.com/joperezr/runtime/commit/34d301617a59e204a0a912782f94c262e885bc79

As you can see, this basically removed all instances of CP1002 diagnostic, as all of the previous missing references in the closure were not required to perform the analysis.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticBag.cs
```diff
@@ -73,7 +73,8 @@ public void Add(T difference)
         {
             if (!Filter(difference.DiagnosticId, difference.ReferenceId))
             {
-                _differences.Add(difference);
+                if (!_differences.Contains(difference))
```
ericstj, line 76: Could you make this a HashSet and implement Equals / GetHashCode? As it stands, this will end up being O(N) for every addition which could add up.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
```diff
@@ -65,5 +65,10 @@ public CompatDifference(string diagnosticId, string message, DifferenceType type
         /// </summary>
         /// <returns><see cref="string"/> describing the difference.</returns>
         public override string ToString() => $"{DiagnosticId} : {Message}";
+
+        public bool Equals(CompatDifference other) => DiagnosticId.Equals(other.DiagnosticId, StringComparison.InvariantCultureIgnoreCase) &
```
ericstj, line 69: When implementing equals always make sure to have fast path for reference equality and handle null.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticBag.cs
joperezr: For sure. When added the 'Contains' call I thought about doing that but forgot about it. Will add in a sec.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
```diff
@@ -65,5 +66,22 @@ public CompatDifference(string diagnosticId, string message, DifferenceType type
         /// </summary>
         /// <returns><see cref="string"/> describing the difference.</returns>
         public override string ToString() => $"{DiagnosticId} : {Message}";
+
+        public bool Equals(CompatDifference other) => other != null &&
```
ericstj, line 70: Do you also need `public bool override Equals(object other)`?

joperezr: It wouldn't be called by HashCode<T> since the (default) GenericEqualityComparer<T> will always call the concrete one in our case, but it also doesn't hurt to add so I can go ahead and add it.

ericstj: I see.  I usually follow this rule religiously as I've had to debug plenty of the times it's been broken:
https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-5.0#remarks
> If you override the GetHashCode method, you should also override Equals, and vice versa. If your overridden Equals method returns true when two objects are tested for equality, your overridden GetHashCode method must return the same value for the two objects.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
```diff
@@ -65,5 +66,22 @@ public CompatDifference(string diagnosticId, string message, DifferenceType type
...
+                                                      (object.ReferenceEquals(this, other) ||
+                                                      (DiagnosticId.Equals(other.DiagnosticId, StringComparison.InvariantCultureIgnoreCase) &
+                                                      Type.Equals(other.Type) &
+                                                      ReferenceId.Equals(other.ReferenceId, StringComparison.InvariantCultureIgnoreCase)));
+
+        public override bool Equals(object obj) => obj is CompatDifference other ? Equals(other) : false;
```
ericstj, line 76: nit: 
```suggestion
        public override bool Equals(object obj) => Equals(other as CompatDifference);
```

ericstj (approved):

ericstj:
I fixed my suggestion, sorry for the typo.  I guess github needs syntax completion in those comments.

ericstj (approved):

safern (commented):
Nice! Thanks for doing this.

PR #20577, opened 2021-09-03
Do not specify a protocol if server does not require secrets

pranavkm:
As part of the hot reload work, we wanted to optionally include a WS (sub)protocol if a value was specified. It was designed to be backwards compatible in that older servers (VS uses this) that did not anticipate a subprotocol would continue operating if this wasn't specified.

The second parameter to `WebSockets` is an array of values which [defaults to an empty array](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket). Passing a `null` value causes it to translate it to an empty string and require that the server acknowledge it with an empty subProtocol. Using an empty array appears to fix the issue.


pranavkm:
@javiercn could you have a quick look at this? I'll get this ported to main once this is merged.

javiercn (approved):

PR #20600, opened 2021-09-03
[Razor] Unskip tests

javiercn:
The tests were skipped but they are working now.


src/Tests/Microsoft.NET.Sdk.Razor.Tests/StaticWebAssetsIntegrationTest.cs
```diff
@@ -564,8 +564,8 @@ public void Publish_NoBuild_GeneratesPublishJsonManifestAndCopiesPublishAssets()
                 intermediateOutputPath);
         }
 
-        [Fact(Skip = "https://github.com/dotnet/sdk/issues/17979")]
-        public void Build_DeployOnPublish_GeneratesPublishJsonManifestAndCopiesPublishAssets()
+        [Fact]
+        public void Build_DeployOnBuild_GeneratesPublishJsonManifestAndCopiesPublishAssets()
         {
             var expectedManifest = LoadBuildManifest();
```
SteveSandersonMS, line 570: Should this line be removed now? Looks like we're no longer using it.

SteveSandersonMS (approved):

src/Tests/Microsoft.NET.Sdk.Razor.Tests/StaticWebAssetsIntegrationTest.cs
javiercn: Yep, will do cleanup in a follow-up PR, don't want to force another build since these tend to fail

SteveSandersonMS: > these tend to fail

I don't really get that. Are you saying you think this test is still flaky? Should we reconsider unskipping it then?

SteveSandersonMS: If you mean some unrelated thing in the SDK build is flaky then that makes sense, and TBH I'm not sure it's important enough to do a follow-up PR just for this unused variable in a test - it's fine really.

javiercn: No, just the SDK CI. Our tests are 100% reliable as far as I can tell. I just didn't want to trigger another build for that small change.

TanayParikh:
Safe to delete branch?

PR #20602, opened 2021-09-03
[Blazor] Fix duplicate entries in blazor.boot.json manifest extensions

javiercn:
When we are generating the extension it gets included into two ItemGroups that go into GenerateBlazorBootJson, so it creates an entry with an empty string `""` on the extension section in the manifest.

* The fix removes the item from the item group it should not be in, avoiding the duplicate.
* It adds a debug.assert to ensure that we don't make a mistake like adding an empty entry to the manifest.
* It improves the integration tests to check specifically that there is only one entry in the manifest

### DIFF
--- a/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
+++ b/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
@@ -542,6 +542,9 @@ Copyright (c) .NET Foundation. All rights reserved.
         Include="@(StaticWebAsset)"
         Condition="'%(StaticWebAsset.AssetTraitName)' == 'JSModule' and '%(StaticWebAsset.AssetTraitValue)' == 'JSLibraryModule' and '%(AssetKind)' != 'Build'" />
 
+      <!-- We remove the extensions since they are added to the list of static web assets but we need to compute the target path for them -->
+      <_BlazorPublishAsset Remove="@(_BlazorExtensionsCandidatesForPublish)" />
+
     </ItemGroup>
 
     <ComputeStaticWebAssetsTargetPaths
--- a/src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
+++ b/src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
@@ -1,8 +1,10 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
@@ -155,7 +157,9 @@ namespace Microsoft.NET.Sdk.BlazorWebAssembly
                         Log.LogMessage("Candidate '{0}' is defined as a library initializer resource.", resource.ItemSpec);
                         resourceData.libraryInitializers ??= new();
                         resourceList = resourceData.libraryInitializers;
-                        AddResourceToList(resource, resourceList, resource.GetMetadata("TargetPath"));
+                        var targetPath = resource.GetMetadata("TargetPath");
+                        Debug.Assert(!string.IsNullOrEmpty(targetPath), "Target path for '{0}' must exist.", resource.ItemSpec);
+                        AddResourceToList(resource, resourceList, targetPath);
                         continue;
                     }
                     else if (string.Equals("BlazorWebAssemblyResource", assetTraitName, StringComparison.OrdinalIgnoreCase) &&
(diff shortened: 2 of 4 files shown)


src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -1,8 +1,10 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
SteveSandersonMS, line 3:
```suggestion
```

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/GenerateBlazorBootJsonTest.cs
```diff
@@ -1,13 +1,11 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
SteveSandersonMS, line 3:
```suggestion
```

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -542,6 +542,9 @@ Copyright (c) .NET Foundation. All rights reserved.
         Include="@(StaticWebAsset)"
         Condition="'%(StaticWebAsset.AssetTraitName)' == 'JSModule' and '%(StaticWebAsset.AssetTraitValue)' == 'JSLibraryModule' and '%(AssetKind)' != 'Build'" />
 
+      <!-- We remove the extensions since they are added to the list of static web assets but we need to compute the target path for them -->
```
SteveSandersonMS, line 545: Is this comment accurate? It might just be because I don't know this area of the code at all, but it seems strange that we remove something because "we need to compute the target path for them".

Does removing them somehow cause their target paths to get computed?

SteveSandersonMS:
For PRs like this would you be able to add more of a description? It's not clear to me what problem this solves (is there a linked issue?) or how the change solves it.

It's more of an issue in the long term when we have to dig back through the change history to understand why things are done as they are.

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
javiercn: Tooling actually generates the extra `//` so I'm going to leave it there (I have a separate PR where I apply the correct license to all our files)

javiercn:
> For PRs like this would you be able to add more of a description? It's not clear to me what problem this solves (is there a linked issue?) or how the change solves it.

Apologies, I thought I added the description last friday. Will include more details

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: I added a description to the PR, the issue is that we were including the user extensions in this item group and that was causing an incorrect entry to be added to the manifest.

Extensions and JS modules are pre-processed below on the file as we need the target path for them

javiercn:
ping

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
SteveSandersonMS: Adding a description to the PR is great but the PR is not where we'd be looking when trying to understand this code in the future. If there's a way of phrasing the comment that's easier to understand that would be preferable 😃 

> Extensions and JS modules are pre-processed below on the file as we need the target path for them

I guess I just don't know why anything about the target path is a relevant thing to have a comment about here. It seems like an unrelated part of the system. But the way the comment in the source is phrased makes it sound causal, i.e., the act of removing them from this itemgroup is what makes the target path get computed later. Perhaps that is true but it's unexpected.

I know this comment is only a small detail and we shouldn't stress about it or delay the PR for it.

SteveSandersonMS (approved):

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: All files in blazor.boot.json (dlls and so on) only include the filename.

JS initializer and extensions use the URL because they can live "anywhere" not just inside the _framework folder. So for example you can have a js initializer that lives in `_content/library/library.module.lib.js` and we need to be able to load that.

In a similar vein, when you transform the output during publish, you should not be forced to put those files inside _framework, so we need to compute the URL for those assets an include them there. As an example, you can grab all the dlls from _framework concatenate them and put them inside `_bin/bundle.bdl`

PR #20609, opened 2021-09-03
[Razor] Fixes packing on multitargeting projects

javiercn:
There are several issues involved in fixing this. It is not a scenario that worked well before for several reasons:
* Scoped css would include both files (project bundle and app bundle on the package)
* The static web asset targets get called multiple times, one time per target and that introduces issues.
* With the move to static web assets V2, when packaging assets on multi-targeting packages tries to use the V1 version of static web assets to do the packing.

The fixes are as follows:
* Make scoped CSS only work in 5.0 and onwards which is when the feature was originally introduced.
  * The feature is active on 3.1 however the convention is so specific that we didn't consider disabling it before.
* Remove the app bundle when we are packing assets using V1 (avoid packing app.styles.css)
* Make sure that the targets for packing assets only run once on multi-target builds
  * We achieve this by writing a sentinel file the first time the target to pack the asset runs.
  * On each build we remove the file if its present.
* Ensure that when there is at least a net6.0 target we use the V2 asset definition to do the packing.

### DIFF
--- a/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.StaticWebAssets.5_0.targets
+++ b/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.StaticWebAssets.5_0.targets
@@ -80,7 +80,7 @@ Copyright (c) .NET Foundation. All rights reserved.
       $(ResolveStaticWebAssetsInputsDependsOn)
     </ResolveStaticWebAssetsInputsDependsOn>
 
-    <ResolveReferencedProjectsStaticWebAssetsDependsOn Condition="$(TargetFramework) != ''">
+    <ResolveReferencedProjectsStaticWebAssetsDependsOn Condition="'$(TargetFramework)' != ''">
       PrepareProjectReferences;
       $(ResolveReferencedProjectsStaticWebAssetsDependsOn)
     </ResolveReferencedProjectsStaticWebAssetsDependsOn>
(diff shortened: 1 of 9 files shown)


src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.StaticWebAssets.5_0.targets
```diff
@@ -80,7 +80,7 @@ Copyright (c) .NET Foundation. All rights reserved.
       $(ResolveStaticWebAssetsInputsDependsOn)
     </ResolveStaticWebAssetsInputsDependsOn>
 
-    <ResolveReferencedProjectsStaticWebAssetsDependsOn>
+    <ResolveReferencedProjectsStaticWebAssetsDependsOn Condition="$(TargetFramework) != ''">
```
pranavkm, line 83:
```suggestion
    <ResolveReferencedProjectsStaticWebAssetsDependsOn Condition="'$(TargetFramework)' != ''">
```

src/RazorSdk/Tasks/GenerateStaticWebAsssetsPropsFile.cs
```diff
@@ -37,6 +39,8 @@ public class GenerateStaticWebAsssetsPropsFile : Task
         public ITaskItem[] StaticWebAssets { get; set; }
 
         public string PackagePathPrefix { get; set; } = "staticwebassets";
+        
+        public bool AllowEmptySourceType { get; set; } = false;
```
pranavkm, line 43:
```suggestion
        public bool AllowEmptySourceType { get; set; }
```

pranavkm (approved):

PR #20630, opened 2021-09-03
Blazor AOT Cache Busting

### LINKED ISSUE #30788: dotnet.js for AOTed apps needs a better cache-busting mechanism
Currently the Blazor SDK cache-busts `dotnet.js` by including the framework version as part of the built or published file name. With AOT, this might not be sufficient since the contents of the file can change depending on what code is being AOT compiled. We need an alternate way to mangle the filename for cache busting. A simple solution might be to include the hash of the file as part of the file name.

TanayParikh:
- Adds a hash to the `dotnet.js` file name to improve cache busting for published AOT assets.
- Stop copying build output dotnet.js to obj folder

> We are doing this to help with another PR from the mono folks, we used to copy the dotnet.js file to the obj folder and rename it (because of limitations on old static web assets).

> In addition to that, the mono folks are introducing support for debugging AoT apps at build time, at which point the dotnet.js file gets copied (by them) to the output folder and updated (for support debugging AoT with native libraries).

> Its simpler to avoid our own local copy and just look up the file and always create an asset that includes the version and the hash. That works for both cases, when AoT is being used and when its not

#### Publish Assets
image

#### Bin Log
image
image

#### Methodology

Created a new blazor wasm project and published using the dev sdk.

```
 ../build.sh && ../artifacts/bin/redist/Debug/dotnet/dotnet workload install wasm-tools && ../artifacts/bin/redist/Debug/dotnet/dotnet publish -c Release -bl
```

Output:
```
...
  Optimizing dotnet.wasm ...
  Compressing Blazor WebAssembly publish artifacts. This may take a while...
  aotwasm -> /workspaces/sdk/artifacts/bin/aotwasm/Release/net6.0/publish/
```

Fixes: https://github.com/dotnet/aspnetcore/issues/30788


src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -251,6 +268,23 @@ public override bool Execute()
             static bool IsDotNetWasm(string key) => string.Equals("dotnet.wasm", Path.GetFileName(key), StringComparison.Ordinal);
         }
 
+        private string GetItemHash(ITaskItem item)
+        {
+            var hash = item.GetMetadata("FileHash");
+            if (string.IsNullOrEmpty(hash))
+            {
+                // Some files that are part of the service worker manifest may not have their hashes previously
```
pranavkm, line 276: Is this comment correct? This is only ever called on dotnet.js which isn't part of the manifest, no?

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -193,6 +194,22 @@ public override bool Execute()
...
+                        // A Base64 encoded hash can contain `/`, `=` and `+` which may be problematic for URLs/file names.
+                        // As we're only using the hash for the purposes of cache busting, we can opt to simply remove those chars 
+                        // from the hash (at the cost of a tiny bit of entropy).
+                        //
+                        // URL encoding is not an option as encoded file names are decoded at another layer and not
+                        // reflected in the final published asset file name (ie. final asset would still contain symbols).
+                        //
+                        // https://en.wikipedia.org/wiki/Base64#Base64_table
+                        var itemHash = GetItemHash(aotDotNetJs).Replace("/", string.Empty).Replace("=", string.Empty).Replace("+", string.Empty);
+                        var newRelativePath = $"{existingRelativePathWithoutExtension}.{itemHash}.js";
```
pranavkm, line 210: is picking a substring (first 8 / 10 chars) sufficient? We don't need to tack on 40+ chars to the string to make it unique.

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -193,6 +194,22 @@ public override bool Execute()
...
                         newDotNetJs.SetMetadata("OriginalItemSpec", aotDotNetJs.ItemSpec);
+
+                        var existingRelativePath = newDotNetJs.GetMetadata("RelativePath");
+                        var existingRelativePathWithoutExtension = Path.Combine(Path.GetDirectoryName(existingRelativePath), Path.GetFileNameWithoutExtension(existingRelativePath));
+
+                        // A Base64 encoded hash can contain `/`, `=` and `+` which may be problematic for URLs/file names.
+                        // As we're only using the hash for the purposes of cache busting, we can opt to simply remove those chars 
+                        // from the hash (at the cost of a tiny bit of entropy).
+                        //
+                        // URL encoding is not an option as encoded file names are decoded at another layer and not
```
pranavkm, line 205: Out of curiosity, what's url decoding these names?

pranavkm (commented):
Changes look good, just a few questions

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
javiercn: Agree

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
TanayParikh: Not entirely sure, didn't want to spend too much time tracking that down given it's not really necessary (especially given we're willing to take a substring anyways).

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
TanayParikh: You're right. This comment isn't accurate in this case. Copied the hashing/comment from another task in the pipeline which did deal with manifests. Removed the comment.

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
javiercn: The final asset would not contain the symbols.

Assets are always copied to `Concat(BasePath, RelativePath)`. Otherwise is a bug

TanayParikh: > final asset would not contain the symbols

Yep, I'm trimming out the symbols below.

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
javiercn: Agree with Pranav, not sure what this comment means.

At this stage nothing has/needs a hash, we compute hashes when we are about to generate the blazor boot json manifest.

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
TanayParikh: > The final asset would not contain the symbols.

Yep, I'm trimming out the symbols below.

javiercn: Actually, use this https://github.com/dotnet/sdk/blob/main/src/RazorSdk/Tasks/ComputeCssScope.cs#L43-L71 (an adapted version)

javiercn: We did all the calculations at the time for the entropy and the risk of collisions and so on, so its better

TanayParikh: Updated.

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
```diff
@@ -97,7 +100,16 @@ public override bool Execute()
                     }
 
                     var destinationSubPath = candidate.GetMetadata("DestinationSubPath");
-                    if (string.IsNullOrEmpty(destinationSubPath))
+                    if (candidate.GetMetadata("FileName") == "dotnet" && candidate.GetMetadata("Extension") == ".js")
+                    {
+                        var dotNetVersion = MicrosoftNETCoreAppRefPackageVersion;
+                        var itemHash = FileHasher.GetFileHash(candidate.ItemSpec);
+                        var cacheBustedDotnetJSFileName = $"dotnet.{dotNetVersion}.{itemHash}";
+                        candidate.SetMetadata("RelativePath", $"_framework/{cacheBustedDotnetJSFileName}.js");
```
TanayParikh, line 108:
```suggestion
                        candidate.SetMetadata("RelativePath", $"_framework/dotnet.{dotNetVersion}.{itemHash}.js");
```

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
```diff
@@ -97,7 +100,16 @@ public override bool Execute()
...
+                        candidate.SetMetadata("AssetTraitName", "BlazorWebAssemblyResource");
+                        candidate.SetMetadata("AssetTraitValue", "native");
+                    }
+                    else if (string.IsNullOrEmpty(destinationSubPath))
```
javiercn, line 112: I'm not 100% sure you should be editing the candidate (don't remember the exact details) but instead creating a new static web asset from scratch (the candidate is not a static web asset, it's a "random thing with properties")

We might also want to do this through StaticWebAsset.cs as that applies a lot of normalization to the item properties.

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
```diff
@@ -97,7 +100,16 @@ public override bool Execute()
...
                     var destinationSubPath = candidate.GetMetadata("DestinationSubPath");
-                    if (string.IsNullOrEmpty(destinationSubPath))
+                    if (candidate.GetMetadata("FileName") == "dotnet" && candidate.GetMetadata("Extension") == ".js")
+                    {
+                        var dotNetVersion = MicrosoftNETCoreAppRefPackageVersion;
+                        var itemHash = FileHasher.GetFileHash(candidate.ItemSpec);
+                        var cacheBustedDotnetJSFileName = $"dotnet.{dotNetVersion}.{itemHash}";
+                        candidate.SetMetadata("RelativePath", $"_framework/{cacheBustedDotnetJSFileName}.js");
+                        candidate.SetMetadata("AssetTraitName", "BlazorWebAssemblyResource");
+                        candidate.SetMetadata("AssetTraitValue", "native");
```
SteveSandersonMS, line 110: Could you clarify why we have to modify these asset trait metadata values? I expected to see code that changes the filename/path, but didn't expect to see other metadata get mutated. I'm sure there's a good reason but am curious to understand!

src/BlazorWasmSdk/Tasks/FileHasher.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Text;
+
+namespace Microsoft.NET.Sdk.BlazorWebAssembly
+{
+    public static class FileHasher
+    {
+        public static string GetFileHash(string filePath)
+        {
+            using var hash = SHA256.Create();
+            var bytes = Encoding.UTF8.GetBytes(filePath);
```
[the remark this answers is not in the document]
javiercn: unfortunately this needs to be net472 compatible

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
javiercn: We are doing this to help with another PR from the mono folks, we used to copy the dotnet.js file to the obj folder and rename it (because of limitations on old static web assets).

In addition to that, the mono folks are introducing support for debugging AoT apps at build time, at which point the dotnet.js file gets copied (by them) to the output folder and updated (for support debugging AoT with native libraries).

I met with Ankit and Larry yesterday to get a better understanding of it and determined that its simpler to avoid our own local copy and just look up the file and always create an asset that includes the version and the hash. That works for both cases, when AoT is being used and when its not

src/BlazorWasmSdk/Tasks/FileHasher.cs
```diff
@@ -0,0 +1,36 @@
...
+            var result = new char[10];
+            var dividend = BigInteger.Abs(new BigInteger(hash.AsSpan().Slice(0, 9).ToArray()));
+            for (var i = 0; i < 10; i++)
+            {
+                dividend = BigInteger.DivRem(dividend, 36, out var remainder);
+                result[i] = chars[(int)remainder];
+            }
+
+            return new string(result);
+        }
```
SteveSandersonMS, line 34: Just in case you're interested - and feel free to ignore this - we have some similar code elsewhere in the product like this:

```cs
            byte[] hashBytes;
            using (var sha1 = SHA1.Create())
            using (var fileStream = File.OpenRead(path))
            {
                hashBytes = sha1.ComputeHash(fileStream);
            }

            var hash = Convert.ToBase64String(hashBytes);
```

I don't really mind, but you could just use the first 10 chars from the base64 representation of the SHA1/SHA256 hash if you wanted. It doesn't seem like it case-sensitivity matters either way in this case.

javiercn: this should use the same thing we use in ComputeCssScope, since we did all the math for collisions on that

javiercn: https://github.com/dotnet/sdk/blob/main/src/RazorSdk/Tasks/ComputeCssScope.cs#L57-L69

src/BlazorWasmSdk/Tasks/FileHasher.cs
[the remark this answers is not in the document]
SteveSandersonMS: Thanks for clarifying!

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
SteveSandersonMS: OK. It's confusing when unrelated changes show up without explanation in a PR. Might be worth having a comment here, or future archaeologists/refactorers will be in for a tough time.

TBH there's so much of this build process that seems impossible to know just by reading the code that maybe I should just stop trying to reason about it. If it were me, I'd absolutely cover it in comments describing what external systems are relying on each of the moving parts here.

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
TanayParikh: > creating a new static web asset from scratch (the candidate is not a static web asset, it's a "random thing with properties")

In that case where would we "put" the newly created static web asset. I ask as the `AssetCandidates` is a collection of `ITaskItem` and thus won't accept a `StaticWebAsset`. Maybe we should be doing something like:

https://github.com/dotnet/sdk/blob/a0d74e39bc7533643b78184d189336ca2b96b2be/src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs#L87-L96

It ensures we don't update the initial asset type, but still are able to add in the desired properties. 

> We might also want to do this through StaticWebAsset.cs as that applies a lot of normalization to the item properties.

Can leverage:
https://github.com/dotnet/sdk/blob/a0d74e39bc7533643b78184d189336ca2b96b2be/src/RazorSdk/Tasks/StaticWebAssets/StaticWebAsset.cs#L46-L54

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
TanayParikh: > OK. It's confusing when unrelated changes show up without explanation in a PR. Might be worth having a comment here, or future archaeologists/refactorers will be in for a tough time.

I totally agree, sorry about that should've left a comment with some more context. I've updated the PR description with Javier's comment above to provide reasoning for these additional changes.

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
javiercn: I've recalled my commment here. This is only setting up some basic metadata on the candidates, there's a call afterwards to DefineStaticWebAssets that makes sure all the items are normalized

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
```diff
@@ -97,7 +103,17 @@ public override bool Execute()
...
-                    if (string.IsNullOrEmpty(destinationSubPath))
+                    if (candidate.GetMetadata("FileName") == "dotnet" && candidate.GetMetadata("Extension") == ".js")
+                    {
+                        var dotNetVersion = BundledNETCoreAppPackageVersion;
+                        var itemHash = FileHasher.GetFileHash(candidate.ItemSpec);
+                        var cacheBustedDotnetJSFileName = $"dotnet.{dotNetVersion}.{itemHash}";
+                        candidate.SetMetadata("RelativePath", $"_framework/{cacheBustedDotnetJSFileName}.js");
+                        candidate.SetMetadata("AssetTraitName", "BlazorWebAssemblyResource");
+                        candidate.SetMetadata("AssetTraitValue", "native");
+                        candidate.SetMetadata("ContentRoot", Path.Combine(IntermediateOutputPath, "blazor"));
```
javiercn, line 114: I don't think we are copying this file to `Path.Combine(IntermediateOutputPath, "blazor")` any longer, so I don't think we need the intermediateouputpath here. You should be able to leave it empty and it will receive the content root from [here](https://github.com/dotnet/sdk/blob/main/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets#L211) and will be copied to the output directory from where it will be served

TanayParikh:
Updated.

image
image


(via `../artifacts/bin/redist/Debug/dotnet/dotnet publish -c Release -bl`)

MSBuild:
ComputeBlazorBuildAssets:
image
image

ComputeBlazorPublishAssets:
image
image

javiercn (approved):
Changes look great!

javiercn:
@TanayParikh you need to backport this to release/6.0.1xx

javiercn:
(And please squash your commits)

TanayParikh:
> @TanayParikh you need to backport this to release/6.0.1xx

https://github.com/dotnet/sdk/pull/20936

> (And please squash your commits)

Sorry about that! I hadn't configured `squash and merge` for this repo so it defaulted to a merge commit. I've updated my defaults, though I don't believe I'll be able to fix the merged commits here (without a force push on main).

PR #20648, opened 2021-09-03
Disable main loc build while release/6.0.1xx is running

wli3:


jonfortescue:
Doing it this way will cause it to be overwritten on next arcade update. Instead you should modify [this line](https://github.com/dotnet/sdk/blob/main/eng/build.yml#L18) to be: 

```yaml
- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), eq(variables['Build.SourceBranch'], 'refs/heads/release/6.0.1xx')) }}:
```

or just comment out that onelocbuild block entirely.

wli3:
@jonfortescue corrected according to your sugggestion

jonfortescue (approved):
LGTM

PR #20719, opened 2021-09-07
Prevent null ref in RazorSourceGenerator

pranavkm:
We had an e2e test that verifies RSG worked correctly when suppressed, but missed
an incrementalism case. Added a test, and verified that hot reload works with this patch.

Fixes https://github.com/dotnet/aspnetcore/issues/36227


TanayParikh (approved):

SteveSandersonMS:
Nice! Thanks for the speedy reaction.

Issue #20617, opened 2021-09-03
When installing or updating workloads, better handle flaky network connectivity


dsplaisted:
Workload packs can be relatively large, so we should more gracefully handle cases when a download fails when we're installing or updating workloads.

Right now, I think we download and install each workload pack serially, and if a download fails we roll back the whole install and delete the packs that we did successfully download.  Ideally, we should do the following:

- Download all workload packs up-front before trying to install any of them.  This should avoid having to roll back workload pack installs due to a download failure
- If there is a download failure, don't delete any workload packs that were successfully downloaded.  The user should be able to re-run the same install or update command, and it should pick up downloading the pack that failed to download, instead of having to start from scratch again.

We should do this for both MSI based and file based installations, I'm not sure if this is implemented separately for each one.

See #19739 for where download failures are causing pain.

joeloff:
It is implemented separately. When not using the offline cache, for MSIs, downloads happen as part of the secure cache we create in ProgramData. If the MSI + json manifest is cached, we won't download and just install, otherwise, the package is downloaded to temp, unzipped, we do some verification on the MSI including verifying the AuthentiCode signature and then move it to the secure location.

wli3:
- [ ] Add retry
- [ ] Separate download and install
- [ ] Be able to resume if not all packages are download successfully
- [ ] Better error message, where the operation fails
- [ ] Logging download and install

sfoslund:
Nuget download retry added with #20727, removing release blocking tag but leaving this open for other enhancements.

PR #20727, opened 2021-09-07
Adding retries to nuget package downloader

sfoslund:
Part of https://github.com/dotnet/sdk/issues/20617

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
@@ -35,17 +35,28 @@ namespace Microsoft.DotNet.Cli.Utils
             foreach (var t in timer())
             {
                 await t;
-                var result = action();
+                T result = default;
+                count++;
+                try
+                {
+                    result = action();
+                }
+                catch
+                {
+                    if (count == maxRetryCount)
+                    {
+                        throw;
+                    }
+                }
+
                 if (shouldStopRetry(result))
                 {
                     return result;
                 }
 
-                count++;
                 if (count == maxRetryCount)
                 {
-                    throw new Exception(
-                        $"Retry failed for {taskDescription} after {count} times with result: {result}");
+                    return result;
                 }
             }
             throw new Exception("Timer should not be exhausted");
@@ -56,24 +67,7 @@ namespace Microsoft.DotNet.Cli.Utils
             Func<IEnumerable<Task>> timer = null)
         {
             timer = timer == null ? () => ExponentialRetry.Timer(ExponentialRetry.Intervals): timer;
-            var count = 0;
-            foreach (var t in timer())
-            {
-                await t;
-                count++;
-                try
-                {
-                    return action();
-                }
-                catch
-                {
-                    if (count == maxRetryCount)
-                    {
-                        throw;
-                    }
-                }
-            }
-            throw new Exception("Timer should not be exhausted");
+            return await ExecuteWithRetry(action, _ => false, maxRetryCount, timer);
         }
 
         public static IEnumerable<Task> Timer(IEnumerable<TimeSpan> interval)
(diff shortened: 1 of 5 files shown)


sfoslund:
@wli3 do you have any thoughts on testing this?

wli3:
https://github.com/dotnet/cli/pull/10073/files#diff-a7de1d0db9a6d051b090e60ccb5f6011ca3688115f779cb2c8a081564171fe16R10

The above code can be tested by input "Intervals" with all 0. It used to have a test, but deleted due to hard to run tests for build itself. At the same time I don't think it is too necessary.

You should have exponential retry, so few retry won't happen at the same time.

And it would be great to unify few implementations of retry in our code base. In test, ExponentialRetry, in IO access RetryOnMoveAccessFailure.

sfoslund:
Not sure what's going on with CI (tests are passing and no build errors but legs are still failing) but this should be ready for review.

src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
```diff
@@ -43,16 +41,40 @@ public static IEnumerable<TimeSpan> Intervals
                     return result;
                 }
 
-                log?.WriteLine($"Operation failed. Retry count: {count}");
                 count++;
                 if (count == maxRetryCount)
                 {
-                    throw new RetryFailedException(
+                    throw new Exception(
                         $"Retry failed for {taskDescription} after {count} times with result: {result}");
```
dsplaisted, line 48: This will need to be localized now.  Previously it was just used to retry tests, so it didn't need to be localized, but now we are using it in the product.

src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
```diff
@@ -43,16 +41,40 @@ public static IEnumerable<TimeSpan> Intervals
...
                 }
             }
             throw new Exception("Timer should not be exhausted");
         }
+        
+        public static async Task<T> ExecuteWithRetryOnFailure<T>(Func<T> action,
```
dsplaisted, line 54: Should this just call the other overload with `_ => false` for `shouldStopRetry`?

dsplaisted (commented):
Looks like this will retry after 5 and then after 50 seconds, and then give up.  That seems good to me.

src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
```diff
@@ -6,9 +6,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
-using Xunit.Abstractions;
 
-namespace Microsoft.NET.TestFramework
+namespace Microsoft.DotNet.Cli.Utils
```
joeloff, line 10: good catch on the namespace

src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
joeloff: Should we perhaps include a message informing the user of the interval? The first might be fine, but jumping exponentially without any feedback might make them think that the command is unresponsive

wli3: Good catch. Although we have no time for loc pass. We can format is differently and considering this as a dev message. At the end of the message

wli3 (approved):

src/Cli/Microsoft.DotNet.Cli.Utils/ExponentialRetry.cs
wli3: ideally the user should not see it all the time

sfoslund: I've refactored so we shouldn't need the error anymore

sfoslund:
@wli3 the helix test leg is failing consistently but there are no test errors or build errors. Have you seen this before?

wli3:
I cannot find the failure log anymore. I'll see if this build still fails

dsplaisted (approved):

wli3:
@sfoslund if that still fails, add some logging for debug.

wli3:
Sorry for the delay. I cannot find good reason by looking at the output

sfoslund:
No worries, I'm working on debugging now...

PR #20742, opened 2021-09-07
[release/6.0.1xx] Update BlazorWebAssembly targets to work with updated wasm targets

radical:
[This runtime PR](https://github.com/dotnet/runtime/pull/57556) adds support for
relinking the native wasm files during the build (instead of only during
publish).

This required changes to the targets available, and how they can be
used, and thus require changes in the blazor targets to work with that.

The new setup is explained in
https://github.com/radical/runtime/blob/wasm-build/src/mono/wasm/build/README.md
.

This will need to be merged *after* https://github.com/dotnet/runtime/pull/57556 goes in.


radical:
@javiercn @pranavkm @captainsafia The changes here might not be the best way to get the results, so please feel free to change that. But with this, I can build blazorwasm projects with native libraries, which should allow debugging with VS.

radical:
cc @lewing

javiercn:
@radical do you think we can have a meeting to discuss these changes?

I think we can probably find a more streamlined way to integrate these changes but I need a bit more context I think.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -406,7 +487,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     </ItemGroup>
   </Target>
 
-  <Target Name="ProcessPublishFilesForBlazor" DependsOnTargets="_ResolveBlazorWasmConfiguration;LoadStaticWebAssetsBuildManifest" AfterTargets="ILLink">
+  <Target Name="ProcessPublishFilesForBlazor" DependsOnTargets="_ResolveBlazorWasmConfiguration;LoadStaticWebAssetsBuildManifest" AfterTargets="ILLink" Condition="'$(WasmBuildingForPublish)' != 'true'">
```
javiercn, line 490: I'm not 100% sure about the condition here.

This target computes the list of publish assets for the app, so it should always run.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -115,12 +124,75 @@ Copyright (c) .NET Foundation. All rights reserved.
...
 
+    <ItemGroup Condition="@(WasmNativeAsset->Count()) != 0">
+      <_DotNetJsItem Remove="@(_DotNetJsItem)" />
+      <_DotNetJsItem Include="@(WasmNativeAsset)" Condition="'%(WasmNativeAsset.FileName)%(WasmNativeAsset.Extension)' == 'dotnet.js'">
+        <RelativePath>_framework/$(_BlazorDotnetJsFileName)</RelativePath>
+        <AssetType>native</AssetType>
+        <CopyLocal>true</CopyLocal>
+      </_DotNetJsItem>
+    </ItemGroup>
+
```
javiercn, line 195: Based on offline discussion, we don't need the scrambling here since @TanayParikh took care of it on his PR

javiercn (approved):
Overall changes look good.

I'm signing off to make sure things are unblocked (provided tests pass and trusting manual verification of the AoT scenarios has happened before merging.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
radical: Does that mean that the two targets `_Scramble*` can be dropped completely?

radical: Ah, the target is no longer there! 👍

javiercn: Yep. We do this all the time now on build and publish if necessary, as it improves the cache busting

marek-safar (approved):

radical:
The corresponding runtime PR https://github.com/dotnet/runtime/pull/58913 has merged. Next is to wait for a version bump PR to be opened.

TanayParikh:
I believe this'll need to be cherry-picked into `main` as well.

PR #20781, opened 2021-09-08
Make ICUDataMode less prone to accidental breakage

SteveSandersonMS:
Currently, the enum `ICUDataMode` relies on compiler-generated numeric values. This is not a great move because we serialize these values and check for specific ones in the JS-side code. Anybody who updates this enum, for example inserting a new value in the middle, would risk breaking applications in subtle ways.

We should use explicit values to make breakage less likely.

### DIFF
--- a/src/BlazorWasmSdk/Tasks/BootJsonData.cs
+++ b/src/BlazorWasmSdk/Tasks/BootJsonData.cs
@@ -100,22 +100,24 @@ namespace Microsoft.NET.Sdk.BlazorWebAssembly
         public Dictionary<string, ResourceHashesByNameDictionary> extensions { get; set; }
     }
 
-    public enum ICUDataMode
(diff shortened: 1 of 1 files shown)


javiercn (approved):

Eilon (approved):

pranavkm:
>  for example inserting a new value in the middle, would risk breaking applications in subtle ways.

Seems super unlikely we would do that, but this change seems like good hygiene.

SteveSandersonMS:
> Seems super unlikely we would do that, but this change seems like good hygiene.

I agree that just having a comment saying "don't do that" would likely be enough. But if we're adding a comment, we might as well make the numbers explicit.

Without even a comment, I think it's totally plausible that somebody might think it's nice to re-sort into alphabetical order or something like that!

PR #20793, opened 2021-09-08
Support multi-targeting for Roslyn components

### LINKED ISSUE #20355: Support multi-targeting for Roslyn components
### Background
As Roslyn implements more features and APIs, the need to target these features/APIs increases in Roslyn components (i.e. Code Analyzers/Fixers and Source Generators). As the components need to target newer versions of Roslyn, it is becoming clear that some components need to target multiple versions of Roslyn at the same time.

Take, for example, the following issues in dotnet/runtime with the System.Text.Json source generator:

* https://github.com/dotnet/runtime/issues/56702
* https://github.com/dotnet/runtime/issues/57880

In both of these issues, the only fix is for the System.Text.Json source generator to take a dependency on version `4.0` of the `Microsoft.CodeAnalysis` assemblies. The first needs `IIncrementalGenerator` and the second needs `FileScopedNamespaceDeclarationSyntax`, both of which were only introduced in `4.0`.

However, once a Roslyn component references a version of `Microsoft.CodeAnalysis`, that Roslyn component can no longer be loaded in an earlier version of the compiler.

This poses two problems:

1. If a Roslyn component still wants to work in a previous version of the compiler, for example in Visual Studio 2019, the component has a problem. It needs to pick which version it targets - either older versions without the new API support, or the new version.
2. If a Roslyn component chooses to target the new version, there is no way to shut off the component when loaded in older versions. Thus, when a developer in VS 2019 references a NuGet package with a component targeting Roslyn `4.0`, the developer gets a warning in their build:

`
Warning	CS8032	An instance of analyzer System.Text.Json.SourceGeneration.JsonSourceGenerator cannot be created from C:\Users\eerhardt\.nuget\packages\system.text.json\7.0.0-dev\analyzers\dotnet\cs\System.Text.Json.SourceGeneration.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.
`

This is not a great experience for our customers.

### Proposal

We will update the current logic that resolves the Analyzers from NuGet packages to support an enhanced `analyzers` folder structure. The enhanced folder structure will allow for the Roslyn version targeted by the component to be expressed. This will allow for multiple versions to be targeted in a single package, as well as being able to express a "minimal" version. This will address the two problems above. When the Analyzers are resolved from the NuGet package, the correct asset(s) will be resolved according to which version of the compiler will be used to compile the assembly.

#### Current folder structure

The following structure is recognized today by the SDK:

```
  analyzers[\dotnet][\cs|vb]\component.dll
```

#### Proposed folder structure

This proposal adds a new, optional folder to the structure: `roslyn{version}`:

```
  analyzers[\dotnet][\roslyn{version}][\cs|vb]\component.dll
```

Where `{version}` is the `{major}.{minor}` version of the `Microsoft.CodeAnalysis` assemblies used during the `Compile` task.

When a NuGet package contains folders under `analyzers` with the pattern, `roslyn<major-version>.<minor-version>`, the folder with the highest version that is less than or equal to the current `Microsoft.CodeAnalysis` `<major>.<minor>` version will be used. This works the same as TFMs work under the `lib` folder. Except instead of TargetFramework versions being evaluated, the Roslyn version will be evaluated.

Also note, since the same Analyzer logic exists in NuGet's [ResolveNuGetPackageAssets Task](https://github.com/NuGet/NuGet.BuildTasks/blob/1d8af3499f94a32c1d128a42faceade39c1f4337/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs#L444-L470). Thus, we will need to update the NuGet MSBuild logic with this support. That way developers using non-SDK .NET projects will get the same Analyzers selected as SDK projects.

#### Earlier SDK versions
Since this proposal is to add support to .NET SDK 6.0, we need to consider what happens when NuGet packages that follow this structure are used in SDK versions before 6.0.

If implemented as indicated above, NuGet packages that support multiple Roslyn versions will have all of their assets loaded in earlier versions of the SDK. To prevent that, we can add a new MSBuild property to the SDK that indicates it supports multi-targeting Roslyn versions.

```
$(SupportsRoslynComponentVersioning)
```

NuGet packages can include MSBuild logic that "lights up" when `$(SupportsRoslynComponentVersioning)` is not present. This indicates that the built-in support isn't there to select the right version. The NuGet package needs to do the selection itself. Typically, NuGet packages will select the component that targets the lowest version of Roslyn version it supports, since newer Roslyn versions can't be guaranteed. Another option would be to opt out of adding a Roslyn component all together.

#### Additional rules

* When a package contains analyzer assets both inside a `roslyn{version}` folder and outside, ex. directly under `analyzers\dotnet\cs\analyzer.dll`, both assets will be selected. The reasoning is that the assets outside of `roslyn{version}` folders are considered to work everywhere. Existing analyzer resolution rules add both assets inside and outside `{language}` folders, if present. This follows the same reasoning.
    - If a NuGet package wants to express a component that should be used when none of the `roslyn{version}` folders apply, it can add that asset to `roslyn1.0`.

### Alternative Solutions

An alternative is to follow the proposal in https://github.com/dotnet/roslyn/issues/54108:

> The current plan is to provide authoring and consumption targets that make this easy to do correctly. Over time we'll aim to get NuGet itself updated to support the feature in-box and remove the need for the targets.

Following this approach would mean that each NuGet package that wants to support multiple versions of Roslyn (or a minimum version) would need to ship special MSBuild `.targets` files with duplicated logic in them.

This approach is not ideal because:

1. If there is a bug in the targets, it needs to be fixed in all NuGet packages that copied the targets file.
2. There could be potential conflicts between the targets in the NuGet packages if they are not named uniquely.
3. The logic to resolve the Roslyn version shouldn't be the responsibility of the NuGet packages that ship Roslyn components.

@chsienki @jaredpar @dsplaisted @jasonmalinowski @jmarolf

eerhardt:
Allows for Roslyn components (analyzers, source generators) to target mulltiple Roslyn API versions in a single package. The highest compatible asset is selected.

Fix #20355


src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
```diff
@@ -114,6 +114,14 @@ public string ProjectLanguage
             get; set;
         }
 
+        /// <summary>
+        /// Optional version of the compiler API (E.g. 'roslyn3.9', 'roslyn4.0')
+        /// </summary>
+        public string CompilerApiVersion
```
dsplaisted, line 120: Most of the `ResolvePackageDependencies` task is legacy behavior which is disabled unless you set `EmitLegacyAssetsFileItems` to true.  See https://docs.microsoft.com/dotnet/core/compatibility/msbuild#design-time-builds-only-return-top-level-package-references.

So I don't think we need to update this task.  No one should be consuming the analyzer information that it generates, and if they do it should be OK that they don't handle multi-targeted analyzers.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -172,12 +176,28 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="Microsoft.NET.Build.Tasks.ResolvePackageAssets"
             AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
+  <!-- Reads the version of the compiler APIs that are currently being used in order to pick the correct Roslyn components. -->
+  <Target Name="_ResolveCompilerVersion"
```
dsplaisted, line 180: Can we change this so that the code to read the compiler API version is done as part of the `ResolvePackageAssets` task instead of computed beforehand and passed in?  That way if we don't have any references to multi-targeted analyzers we don't even have to read the API version from the assembly, so we don't have to worry about the perf impact.

In that case probably you would need to pass in the path to the CodeAnalysis DLL to the task instead of the version, which would make testing harder.

dsplaisted (commented):
I haven't reviewed the code in detail yet, but I did have some quick feedback.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
eerhardt: The idea we had offline was that the Roslyn targets would be setting the `$(CompilerApiVersion)` property statically. Thus this target wouldn't need to run at all.

cc @chsienki 

> That way if we don't have any references to multi-targeted analyzers

We are using this property to figure out if we have any references to multi-targeted analyzers at all. Using the first part of this property's value (`roslyn`), we are looking for folders that start with that name in the `analyzers` folder. I didn't want to hard-code "roslyn" into the C# task. It seems like a better design to put that information up in the .targets file, where it can be changed easily, if needed.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
eerhardt: Ah, that makes sense. Thanks for the information on this. I can revert this change.

eerhardt:
@dsplaisted @marcpopMSFT - any thoughts here? I'd like to get this into 6.0 to unblock the multi-targeting work in https://github.com/dotnet/runtime/pull/58446

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -207,9 +211,24 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   </Target>
 
+  <!-- Reads the version of the compiler APIs that are currently being used in order to pick the correct Roslyn components. -->
+  <Target Name="_ResolveCompilerVersion"
+          Condition="'$(CompilerApiVersion)' == ''">
```
ericstj, line 216: Do we have to think about any non-roslyn project types?  Is there something that tells us that a project type supports analyzers?

ericstj (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
eerhardt: > Do we have to think about any non-roslyn project types?

Yes. We need to consider F# projects for sure. We may also need to think about C++/CLI projects.

> Is there something that tells us that a project type supports analyzers?

I haven't found anything. I did find the `RoslynComponent` capability, but that means this project is itself an analyzer.

https://github.com/dotnet/roslyn/blob/86ff365d5d28fddd4a4738cf1d9d101522dffd50/src/Compilers/Core/MSBuildTask/Microsoft.Managed.Core.targets#L348

I can condition this on `Language == 'C#' or 'VB'` if you think that would be safer..? The Task is written so it can accept an empty/null `CompilerApiVersion`, so not setting that property for non-C#/VB projects will be safe.

Thoughts @chsienki ?

chsienki: `RoslynComponent` is only used for the debugging experience, so can't rely on that. I think gating on Language is probably sensible. I'm not sure what will happen if you try and read from a non-existent assembly? If it errors you could also condition on `Condition="Exists('<tasksassembly.dll>')"` to make sure we only try and read it when it's there.

eerhardt: That's a good call. I'll add both conditions.

src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs
```diff
@@ -47,7 +47,167 @@ public static NuGetFramework ParseFrameworkName(string frameworkName)
...
+        ///
+        /// "analyzers/dotnet/roslyn3.7/analyzer.dll"
+        /// "analyzers/dotnet/roslyn3.8/analyzer.dll"
+        /// "analyzers/dotnet/roslyn4.0/analyzer.dll"
+        ///
+        /// When the <paramref name="compilerApiVersion"/> is 'roslyn3.9', only the assets 
+        /// in the folder with the highest applicable compiler version are picked.
+        /// In this case,
+        /// 
+        /// "analyzers/dotnet/roslyn3.8/analyzer.dll"
```
chsienki, line 66: Is this comment correct? For 3.9 if should pick 4.0 right?

dsplaisted, line 66: I don't think so, if the Roslyn compiler version is 3.9, then it can't necessarily run analyzers written for Roslyn 4.0, so it should use the ones written for Roslyn 3.8.

eerhardt: I believe the comment is correct.  If we are running on Roslyn v3.9, we can't pick the `roslyn4.0` asset, or else we would get a load exception that `Microsoft.CodeAnalysis.dll version=4.0.0.0 is not found`.

Since `v4.0` is higher than the current version `v3.9`, we have to exclude it.

This works very similar to TFM selection: you can't pick a `net6.0` asset out of a NuGet package when the app is trying to run on `net5.0`.

chsienki (approved):

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -848,6 +855,8 @@ public int GetHashCode((string, NuGetVersion) library)
 
             private void WriteAnalyzers()
             {
+                HashSet<string>? excludedAnalyzers = NuGetUtils.GetExcludedAnalyzers(_lockFile, _task.ProjectLanguage, _task.CompilerApiVersion);
```
dsplaisted, line 858: I would like to see the logic to check if an analyzer applies refactored.  However, feel free to merge as is since you're crunched for time.

The way this is structured seems confusing.  `GetExcludedAnalyzers` ends up calling `IsApplicableAnalyzer`, but then returns a HashSet which is passed back in to `IsApplicableAnalyzer`.

I think it would be easier to understand (and possibly faster) if `IsApplicableAnalyzer` simply contained all the logic to determine whether an analyzer was applicable, including the multi-targeting logic, instead of separately computing a list of excluded analyzers.  If needed, that could be wrapped in a class which could maintain state to perform expensive operations once, lazily, as necessary.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -207,9 +211,26 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   </Target>
 
+  <!-- Reads the version of the compiler APIs that are currently being used in order to pick the correct Roslyn components. -->
+  <Target Name="_ResolveCompilerVersion"
```
dsplaisted, line 215: I'd like to see the Roslyn targets updated to set this version number soon so that this target is normally never actually run.

src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs
chsienki: Yeah, ignore me I was just having a slow moment.

dsplaisted (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
eerhardt: @chsienki - any chance you can take care of this? I can open an issue in dotnet/roslyn if you'd like.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
eerhardt: Your idea of wrapping the logic in a class allowed me to find a way to make this work.

It isn't possible to make this decision directly in `IsApplicableAnalyzer` because we need to see all the assets before we know which assets apply and which don't. Creating a class to keep track of the necessary information allowed me to track the potential assets, and then add them at the end.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
```diff
@@ -216,7 +216,7 @@ private void GetPackageAndFileDefinitions()
                     string resolvedPath = ResolveFilePath(file, resolvedPackagePath);
                     fileItem.SetMetadata(MetadataKeys.ResolvedPath, resolvedPath ?? string.Empty);
 
-                    if (NuGetUtils.IsApplicableAnalyzer(file, ProjectLanguage))
+                    if (NuGetUtils.IsApplicableAnalyzer(file, ProjectLanguage, excludedAnalyzers: null))
```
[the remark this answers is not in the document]
eerhardt: This change is no longer necessary.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
chsienki: We have an issue here https://github.com/dotnet/roslyn/issues/52265 

It's easy enough to ship in the .NET 6 timeframe by just hard coding it; my concern has always been around the maintainability and ensuring it never gets out of sync, but we can figure that out after the fact if we really need it here.

PR #20822, opened 2021-09-08
[Release 6.0.1xx] [PackageValidation] Improve warnings about references when running PackageValidation in a project after packaging it.

joperezr:
Port of #20567 Only warn about references that are actually needed at runtime.

When running PackageValidation for possible Compatibility breaks with your package, we currently warn for all assemblies in the full closure that we can't find when performing the analysis. These changes improve that to only warn when we actually required those missing references for the analysis.

## Customer impact

Customer's that started using PackageValidation in RC1 (which now will use References when performing the APICompatibility analysis) might have gotten some warnings/errors due to not being able to find some assemblies that are part of the full closure of the libraries being compared. With this change, those warnings/errors will only show if the missing assemblies are actually required by the analysis. This will reduce the false positives we log.

## Regression

No. This is a new feature we are shipping in 6.0.

## Risk

Low. We have very good test coverage in both unit tests and scenario tests, and we have also consumed these changes in dotnet/runtime repo as a test and found no issues.


src/Compatibility/Microsoft.DotNet.PackageValidation/xlf/Resources.zh-Hant.xlf
```diff
@@ -12,11 +12,6 @@
         <target state="new">API compatibility errors between '{0}' (left) and '{1}' (right):</target>
         <note />
       </trans-unit>
-      <trans-unit id="AssemblyLoadWarning">
```
ericstj, line 15: Unrelated to this PR, but I would have expected all of these to be localized at this point.  Are we missing something WRT loc handoff @dsplaisted?

ericstj (approved):

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/CompatDifference.cs
```diff
@@ -65,5 +66,22 @@ public CompatDifference(string diagnosticId, string message, DifferenceType type
...
+                                                      (object.ReferenceEquals(this, other) ||
+                                                      (DiagnosticId.Equals(other.DiagnosticId, StringComparison.InvariantCultureIgnoreCase) &
+                                                      Type.Equals(other.Type) &
+                                                      ReferenceId.Equals(other.ReferenceId, StringComparison.InvariantCultureIgnoreCase)));
+
+        public override bool Equals(object obj) => Equals(obj as CompatDifference);
+
+        public override int GetHashCode()
+        {
+            int hashCode = 1447485498;
```
danmoseley, line 80: does'nt matter today, but it's good to have comments explaining magic numbers like this. I see this scheme is used in our hashcodes several places in this assembly. A quick web search doesn't throw it up as a typical hashcode parameter..

danmoseley: so it could be a terrible one.. can it use HashCode.Combine?

joperezr: It could, in fact that is what we had in other parts of the repo but this being an SDK package it needs to a) target netstandard2.0 (where HashCode.Combine is not inbox) and b) can't have any package dependencies (which would be the way to get to HashCode.Combine in NS2.0). Because of these two reasons, we implement a version of HashCode.Combine here. I did copy this logic from other places in this SDK, but AFAICT it is just a number that was picked and used to generate a hashcode but there isn't any special about it. I think adding comments to explain is a fair point.

danmoseley: given the other places don't have comments I wouldn't reset for that.

post 6.0 someone probably should verify this is a reasonable scheme though ...

ericstj: Yeah, might be better to start with a prime.

PR #20852, opened 2021-09-09
Pass in position to .GetSuggestions()

DavidKarlas:
This means user can get completion also when caret is not at end of command line.
For example `dotnet tool inst$ abc` if caret position is where I placed `$` it will now correctly suggest `install`.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-complete/CompleteCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-complete/CompleteCommand.cs
@@ -60,7 +60,7 @@ namespace Microsoft.DotNet.Cli
 
             var result = Parser.Instance.Parse(input);
 
-            return result.GetSuggestions()
+            return result.GetSuggestions(position)
                 .Distinct()
                 .ToArray();
         }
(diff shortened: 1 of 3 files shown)


sfoslund (approved):
Thanks @DavidKarlas! Is it possible to add a test for this? This will go into 7.0, is that okay with you? If we want it sooner you can target the release/6.0.1xx branch.

DavidKarlas:
I will add unit test tomorrow, I'm sure how high priority this bug is vs. risks, maybe @jonsequitur can make a comment on that?

sfoslund:
the SDK's 6.0.1xx branch is still fairly open, so this should be fine risk-wise. I think we can retarget.

DavidKarlas:
@sfoslund I'm not fimilar with how to get stuff into 6.0.1xx on this repo... Do I merge this and open new PR for 6.0.1xx, or do I target my change at that branch and it will later automatically be backported int main branch?

sfoslund:
You can target the release/6.0.1xx branch in this PR and we'll merge that branch into main periodically.

PR #20864, opened 2021-09-09
Switch to 1ES servicing pools on 5.0.3xx

jonfortescue:
Per https://github.com/dotnet/core-eng/issues/14276


marcpopMSFT:
@jonfortescue 5.0.3xx is out of support so there shouldn't be changes going in here or more builds in the future.

PR #20867, opened 2021-09-09
Remove MMI dependency

joeloff:
MMI links in a large set of native dependencies and was only used to work around System.Management being unsupported on Windows (arm64).

This change removes that dependency and instead directly call the native APIs providing similar information.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/Extensions/ProcessExtensions.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/Extensions/ProcessExtensions.cs
@@ -1,12 +1,9 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using System.Runtime.Versioning;
-using Microsoft.Management.Infrastructure;
+using Microsoft.Win32.SafeHandles;
 
 namespace Microsoft.DotNet.Cli.Utils
 {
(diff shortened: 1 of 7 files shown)


src/Cli/Microsoft.DotNet.Cli.Utils/Extensions/ProcessExtensions.cs
```diff
@@ -36,27 +33,19 @@ public static Process GetParentProcess(this Process process)
         /// </summary>
         /// <param name="process">The process component.</param>
         /// <returns>The process ID of the parent process, or -1 if the parent process could not be found.</returns>
-        public static int GetParentProcessId(this Process process)
+        public unsafe static int GetParentProcessId(this Process process)
         {
-            CimSession cimSession = CimSession.Create(null);
-            IEnumerable<CimInstance> results = cimSession.QueryInstances(@"root\cimv2", "WQL",
-                $"SELECT ParentProcessId FROM Win32_Process WHERE ProcessId='{process.Id}'");
+            using SafeProcessHandle handle = new(process.Handle, ownsHandle: false);
```
joeloff, line 38: @jkotas Is this call correct or should I duplicate the GetProcessHandle logic?

src/Cli/Microsoft.DotNet.Cli.Utils/NativeMethods.cs
```diff
@@ -63,6 +76,14 @@ internal struct JobObjectExtendedLimitInformation
 
             [DllImport("kernel32.dll", SetLastError = true)]
             internal static extern bool AssignProcessToJobObject(IntPtr hJob, IntPtr hProcess);
+
+            [DllImport("kernel32.dll", CharSet = CharSet.Auto)]
+            [DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
```
jkotas, line 81: None of the existing PInvokes in this file have DefaultDllImportSearchPaths. Do you really need it? (The search path is typically set at the assembly level.)

joeloff: The parent process ID method call is used from an elevated instance, so I started adding it. I can move it to the assembly level. Does that impact anything on non-Windows for the assembly?

src/Cli/Microsoft.DotNet.Cli.Utils/Extensions/ProcessExtensions.cs
jkotas:
```suggestion
            SafeProcessHandle handle = process.SafeHandle;
```

jkotas: >  should I duplicate the GetProcessHandle logic?

I do not think you need to duplicate `GetProcessHandle` logic. SafeHandle property should have all permissions that it can have given how you are using it.

joeloff: fixed

src/Cli/Microsoft.DotNet.Cli.Utils/NativeMethods.cs
jkotas: Ok, it is probably safer to just add it on the few APIs that you adding, given that this is RC fix.

sfoslund (approved):
I don't see any obvious issues but I don't have a deep understanding here, please block on @jkotas's review if possible.

src/Cli/Microsoft.DotNet.Cli.Utils/Microsoft.DotNet.Cli.Utils.csproj
```diff
@@ -43,7 +43,6 @@
     <PackageReference Include="Microsoft.Build" Version="$(MicrosoftBuildPackageVersion)" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCorePackageVersion)" />
     <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
-    <PackageReference Include="Microsoft.Management.Infrastructure" Version="$(MicrosoftManagementInfrastructurePackageVersion)" />
```
jkotas, line 46: Also delete `MicrosoftManagementInfrastructurePackageVersion` from  eng/Versions.props ?

jkotas (approved):
LGTM

PR #20897, opened 2021-09-10
[Static web assets] Ensure subpath is always normalized

javiercn:
Fixes https://github.com/aspnet/AspNetCore-ManualTests/issues/890

I've fixed the issue and revamped the unit tests on the class to be much more thorough.

In essence you can have an asset that is mapped as follows:
obj\something\subfolder\dotnet.js
ContentRoot = obj\something\
RelativePath = _framework\dotnet.version.hash.js

In these situations we check if the identity of the asset starts with the content root and set the subpath to the remainder of the path. Given that it uses the identity of the asset, it uses the OS path separator instead of `/` so it needs to be normalized.

The issue is caused because on docker the manifest gets generated on windows and the path separator there is `\` and is not being normalized, so when we run on linux, it fails to find files in a subfolder.

This doesn't happen when the content root for an asset doesn't match its identity since we expect it to be copied to their final destination which we compute as `<<contentroot>>/<<relativepath>>`


src/RazorSdk/Tasks/StaticWebAssets/GenerateStaticWebAssetsDevelopmentManifest.cs
```diff
@@ -148,11 +150,11 @@ private void PersistManifest(StaticWebAssetsDevelopmentManifest manifest)
                         var matchingAsset = new StaticWebAssetMatch
                         {
                             SubPath = asset.Identity.StartsWith(asset.ContentRoot) ?
-                                asset.Identity.Substring(asset.ContentRoot.Length) :
+                                StaticWebAsset.Normalize(asset.Identity.Substring(asset.ContentRoot.Length)) :
```
javiercn, line 153: The fix is here

pranavkm (approved):

Issue #20889, opened 2021-09-10
With .NET 6 RC2 SDK zip installed, maui workload update failed in CLI.


WenJunJi01:
Environment: 6.0.100-rc.2.21460.1(runtime-6.0.0-rc.2.21458.16)

Repro steps:
1. Set up clean machine with .NET 6 RC2 SDK zip installation from https://github.com/dotnet/installer 
2. Run _dotnet workload install maui --skip-manifest-update --source  "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" --source "https://pkgs.dev.azure.com/azure-public/vside/_packaging/xamarin-impl/nuget/v3/index.json"_ in CLI
3. Succeed install without warning or error
4. Run dotnet workload update 

Expected:
maui workload should update succeed without warning or error 

Actual:
maui workload update failed in CLI.
![image](https://user-images.githubusercontent.com/87680983/132832087-f60b856d-766c-4c30-915e-8315db9659a6.png)

Dotnet info:
.NET SDK (reflecting any global.json):
 Version:   6.0.100-rc.2.21460.1
 Commit:    41cdf7a4b2

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19043
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Users\v-danche\Downloads\dotnet-sdk-6.0.100-rc.2.21460.1-win-x64\sdk\6.0.100-rc.2.21460.1\

Host (useful for support):
  Version: 6.0.0-rc.2.21458.16
  Commit:  4cf387b5af

.NET SDKs installed:
  6.0.100-rc.2.21460.1 [C:\Users\v-danche\Downloads\dotnet-sdk-6.0.100-rc.2.21460.1-win-x64\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.0-rc.2.21459.34 [C:\Users\v-danche\Downloads\dotnet-sdk-6.0.100-rc.2.21460.1-win-x64\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.0-rc.2.21458.16 [C:\Users\v-danche\Downloads\dotnet-sdk-6.0.100-rc.2.21460.1-win-x64\shared\Microsoft.NETCore.App]
[...]

PR #20898, opened 2021-09-10
Ensure dotnet root is not null

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/20889


dsplaisted (approved):
Was this a recent regression?  Could there be other places with a similar issue?

sfoslund:
> Was this a recent regression? Could there be other places with a similar issue?

Yes, it was a result of the user local install PR. Not that I'm aware of.

PR #20920, opened 2021-09-10
Vary the cache key for TagHelpersFromReferences on SuppressRazorSourceGenerator

pranavkm:
RSG is invoked with 2 sets of values for SuppressRazorSourceGenerator first with it being set, and once without.
When set, RSG, does not compute any tag helpers from references and produces an empty result. Prior to this change,
we were caching this value invariant of SuppressRazorSourceGenerator. Consequently the next time the RSG was invoked,
the source generator infrastructure would assume that the empty result is still valid and use it for codegen.

The result of the bug is that in a hot reload compilation, none of the tag helpers were resolved and razor would treat tag helpers
from references as markup text. We fix this scenario by varying TagHelpersFromReferences using SuppressRazorSourceGenerator. Additionally added a test to validate that we're generating the expected source.

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
@@ -101,7 +101,9 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
 
             var tagHelpersFromReferences = context.CompilationProvider
                 .Combine(razorSourceGeneratorOptions)
-                .WithLambdaComparer(static (a, b) => a.Left.References.SequenceEqual(b.Left.References), static a => a.Left.References.GetHashCode())
+                .WithLambdaComparer(
+                    static (a, b) => a.Right.SuppressRazorSourceGenerator == b.Right.SuppressRazorSourceGenerator && a.Left.References.SequenceEqual(b.Left.References),
+                    static a => a.Left.References.GetHashCode())
                 .Select(static (pair, _) =>
                 {
                     var (compilation, razorSourceGeneratorOptions) = pair;
--- a/src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
+++ b/src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
@@ -1,0 +1,306 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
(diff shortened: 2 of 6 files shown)


src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -0,0 +1,306 @@
...
+                result.GeneratedSources,
+                sourceResult =>
+                {
+                    Assert.Contains("public partial class Index", sourceResult.SourceText.ToString());
+                },
+                sourceResult =>
+                {
+                    var sourceText = sourceResult.SourceText.ToString();
+                    Assert.Contains("public partial class Counter", sourceText);
+                    // Regression test for https://github.com/dotnet/aspnetcore/issues/36116. Verify that @onclick is resolved as a component, and not as a regular attribute
```
mkArtakMSFT, line 133: Thanks for adding these tests as well as context about why they're here.

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -101,7 +101,9 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
 
             var tagHelpersFromReferences = context.CompilationProvider
                 .Combine(razorSourceGeneratorOptions)
-                .WithLambdaComparer(static (a, b) => a.Left.References.SequenceEqual(b.Left.References), static a => a.Left.References.GetHashCode())
+                .WithLambdaComparer(
```
mkArtakMSFT, line 104: So the issue was that the options were being messed up in the past? And now this code makes sure that the options are properly merged?

pranavkm: Each `Select` is a step in in the incremental source generator. In this step, our inputs are the `Compilation + RazorSourceGeneratorOptions` and we produce a list of `TagHelperDescriptor` that are discovered from the reference assemblies. 

On Line 111, we skip doing any work if the source generator is suppressed. This happens to avoid runtime code messing up razor tooling during design time builds. The baseline compilation in VS runs with `SuppressRazorSourceGenerator=true`. In a hot reload build, the compiler flips `SuppressRazorSourceGenerator=false` and re-invokes the source generator. The source generator infrastructure in Roslyn would invoke the comparer and figure out if it needs to re-compute. Unfortunately, the previous code only varied by the list of reference assemblies, which did not change between these compilations. So it re-uses the previously cached empty results and the hot reload compilation fails to process any tag helpers from ref assemblies.

mkArtakMSFT (approved):
Thanks @pranavkm!

mkArtakMSFT:
@pranavkm looks like your PR is still in draft. Should it be switched so that we can merge it?

PR #20930, opened 2021-09-10
Fix nuget config not being honored

wli3:
Copy user level nuget config to override folder

Tested end to end locally. Pretty hard to add unit test for permission code

### DIFF
--- a/src/Cli/dotnet/SudoEnvironmentDirectoryOverride.cs
+++ b/src/Cli/dotnet/SudoEnvironmentDirectoryOverride.cs
@@ -6,6 +6,7 @@ using System.CommandLine.Parsing;
 using System.IO;
 using System.Linq;
 using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.Common;
 using NuGet.Common;
 using NuGet.Configuration;
 
@@ -36,6 +37,18 @@ namespace Microsoft.DotNet.Cli
         {
             if (!OperatingSystem.IsWindows() && IsRunningUnderSudo() && IsRunningWorkloadCommand(parseResult))
             {
+                if (!TempHomeIsOnlyRootWritable(SudoHomeDirectory))
+                {
+                    try
+                    {
+                        Directory.Delete(SudoHomeDirectory, recursive: true);
+                    }
+                    catch (DirectoryNotFoundException)
+                    {
+                        // Avoid read after write race condition
+                    }
+                }
+
                 Directory.CreateDirectory(SudoHomeDirectory);
 
                 var homeBeforeOverride = Path.Combine(Environment.GetEnvironmentVariable("HOME"));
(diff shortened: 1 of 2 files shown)


src/Cli/dotnet/SudoEnvironmentDirectoryOverride.cs
```diff
@@ -33,7 +37,57 @@ public static void OverrideEnvironmentVariableToTmp(ParseResult parseResult)
             if (!OperatingSystem.IsWindows() && IsRunningUnderSudo() && IsRunningWorkloadCommand(parseResult))
             {
                 Directory.CreateDirectory(SudoHomeDirectory);
+
+                var homeBeforeOverride = Path.Combine(Environment.GetEnvironmentVariable("HOME"));
                 Environment.SetEnvironmentVariable("HOME", SudoHomeDirectory);
+
+                CopyUserNuGetConfigToOverridenHome(homeBeforeOverride);
```
KalleOlaviNiemitalo, line 44: The spelling is "[overridden](https://en.wiktionary.org/wiki/overridden#English)".

KalleOlaviNiemitalo (commented):
I feel suspicious about the hardcoded `/tmp/dotnet_sudo_home/` — what if that directory has been maliciously created by a different, unprivileged user, who still has write access — but perhaps that was discussed in another PR.

src/Cli/dotnet/SudoEnvironmentDirectoryOverride.cs
```diff
@@ -33,7 +37,57 @@ public static void OverrideEnvironmentVariableToTmp(ParseResult parseResult)
             if (!OperatingSystem.IsWindows() && IsRunningUnderSudo() && IsRunningWorkloadCommand(parseResult))
             {
                 Directory.CreateDirectory(SudoHomeDirectory);
+
+                var homeBeforeOverride = Path.Combine(Environment.GetEnvironmentVariable("HOME"));
```
sfoslund, line 41: Sorry, I must have missed this in the first check in of this change, but directly using the HOME env var doesn't match the logic we use for calculating the home dir here: https://github.com/dotnet/sdk/blob/release/6.0.1xx/src/Common/CliFolderPathCalculatorCore.cs#L35, should we be taking those other env vars into account?

wli3: That's fine. If the user override the home directory, they are in control of the situation. We just need to make sure the default make sense.

wli3:
@sfoslund looks good?

sfoslund (approved):

wli3:
@KalleOlaviNiemitalo thank you. I added logic to ensure the owner is root

KalleOlaviNiemitalo:
@wli3 `lstat` might be more secure for that; wouldn't be tricked by a user-owned symlink to a root-owned directory. But I did not check whether you already prevent that in some other way.

src/Cli/dotnet/SudoEnvironmentDirectoryOverride.cs
```diff
@@ -32,12 +37,93 @@ public static void OverrideEnvironmentVariableToTmp(ParseResult parseResult)
...
+
+        private static bool IsOtherUserWritable(StatInterop.FileStatus fileStat)
+        {
+            return (fileStat.Mode & (int) StatInterop.Permissions.S_IWOTH) == 0;
+        }
+
+        private static bool IsGroupWritable(StatInterop.FileStatus fileStat)
+        {
+            return (fileStat.Mode & (int) StatInterop.Permissions.S_IWGRP) == 0;
+        }
```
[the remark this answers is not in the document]
wli3: could you give me more detail? Any suggestions?

wli3:
@joeloff @sfoslund could you both have a look again? especially on https://github.com/dotnet/sdk/pull/20930/commits/d239c27aac0fc61380ad8f408ad757de286467c3

src/Cli/dotnet/SudoEnvironmentDirectoryOverride.cs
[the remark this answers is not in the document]
wli3: @joeloff @KalleOlaviNiemitalo now I get it. Rename to "...CannotWrite"

joeloff (approved):

PR #20937, opened 2021-09-11
Fix EnableTrimAnalyzer for TFMs < 6.0

agocke:


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
```diff
@@ -46,7 +46,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <VerifyDependencyInjectionOpenGenericServiceTrimmability Condition="'$(VerifyDependencyInjectionOpenGenericServiceTrimmability)' == ''">true</VerifyDependencyInjectionOpenGenericServiceTrimmability>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(SuppressTrimAnalysisWarnings)' == '' And ('$(EnableTrimAnalyzer)' == 'true' Or '$(PublishTrimmed)' == 'true')">
+  <PropertyGroup Condition="'$(SuppressTrimAnalysisWarnings)' == '' And '$(PublishTrimmed)' == 'true'">
```
tlakollo, line 49: So what I get from the change is that if EnableTrimAnalyzer is set, but PublishTrimmed is not set then SuppressTrimAnalysisWarnings will be empty. But since we only check for the condition being true in this file so being empty effectively disables the suppression, in that case, do we actually need to set it to 'false' in https://github.com/dotnet/sdk/pull/20937/files#diff-63c4b1605cd7cef177bfd11abd51ab525581c9076caf29a69bce547b8c72b72cR55
Also, what would happen in this other target file https://github.com/dotnet/sdk/blob/792f76cc4e42ecd260170b3e5ca46097e7bbb384/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets#L19

vitek-karas: There's also https://github.com/dotnet/sdk/blob/fdba056f2196faee047b1cd099f1e9e98525a0a4/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L65-L66
(although I don't know if that one is used for analyzers only, or if it applies to the entire project).
In either case, these should be similar/same (here we condition it on >=6.0, in the analyzers, it's conditions on >=5.0. That doesn't sound right either.

vitek-karas: I think we need to figure out what is the desired behavior. Honestly this is a mess (the lack of annotations in ref assemblies), so we just need to do damage control and try to really fix it in 7:

I'm only going to think about analyzers behavior (linker behavior should ideally be the same and it applies to less scenarios):
* If I target >=6.0 I should get the behavior which was here before the changes (warnings are on by default, unless WinForms/WPF)
* Anything below 6.0 is tricky:
  * The ref assemblies are not annotated.
  * For real runtime TFMS (`net5.0`, `netcoreapp3.1`) that's not a big deal since we didn't support it there in a meaningful way, so disabling them by default makes sense (without annotations in the ref assemblies, analyzers are effectively useless unless you add the attributes into your own code).
  * For netstandard TFMs it's questionable. Ref assemblies are still NOT annotated, so it's probably the same and thus warnings should be disabled by default.

I would say that we should enable warnings if EnableTrimAnalysis==true regardless of TMF. Basing this on the logic that if I manually set `EnableTrimAnalysis=true` (since it's off by default), I probably want to see the warnings, regardless of TFM. Obviously the experience will be sub-par due to the missing annotations in ref assemblies. We introduces EnableTrimAnalysis in 6.0, so it's unlikely people will have it in existing projects.

The one thing which "bothers" me is the interaction with the piece of code in the [analyzers](https://github.com/dotnet/sdk/blob/fdba056f2196faee047b1cd099f1e9e98525a0a4/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L65-L66) which turns on `EnableTrimAnalysis` when `PublishTrimmed` is turned on. Basically if I have an existing project which targets `net5.0` and it has `PublishTrimmed=true` I would suddenly get warnings using 6.0 SDK - not great. (I don't think `netstandard` projects will have `PublishTrimmed=true`, it doesn't do anything in <6.0 for those projects).

vitek-karas: Maybe if we add something like (after the one about PublishTrimmed):
```xml
<PropertyGroup Condition="'$(SuppressTrimAnalysisWarnings)' == '' And '$(EnableTrimAnalysis)' == 'true'">
    <SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>
</PropertyGroup>
```

I guess it will have no observable effect, but it would make the behavior explicit. I agree with Tlaka, that leaving the property empty doesn't feel right for cases where we have an intent in mind (we know what should happen).

vitek-karas: This still leaves open the question of what should happen for `IsTrimmable=true`. But I think that's also new in 6.0, so it should enable warnings by default regardless of TFM (which it probably does after the change?). @sbomer?

vitek-karas (commented):
I "think" this will work as I think it should, it's just very non-obvious. I would at the very least add some comments describing the intent. And I might actually set the property to "false" in cases we want it to be that way (even though "false" and empty have basically the same effect).

agocke:
I'm fine with making the expected behavior be: `EnableTrimAnalysisWarnings` -> `SuppressTrimAnalysisWarnings=false`. I can't be sure that this works for Blazor as I don't think we have any tests for that project type in the SDK.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
vitek-karas: How I think about the properties - and their scope:

* `PublishTrimmed` - "I intend to publish this app as trimmed"
  * Used on apps
  * global impact
  * turns on linker in publish
  * For backward compat:
    * TFM <=5.0 show no warnings and don't enable analyzer
    * TFM >= 6.0 - enable analyzer, turn on all warnings
    
* `IsTrimmed` - "I intend to use this assembly in trimmed apps and I invested into making it compatible with trimming"
  * Used on libraries
  * global impact
  * adds the IsTrimmable metadata to the assembly
  * turns on the analyzer and all warnings (no backward compat as it didn't exist pre-6)
  * currently there's no "linker" for this scenario, something we may want to look into later on
 
* `EnableTrimAnalyzer` - "Manual override to enable trim analysis"
  * analyzer only impact
  * typically not set by the user
  * acts as a manual override - should be defaulted in most cases
  * shows all warnings by default - it makes little sense to run the analyzer if we don't show the warnings (at least right now)

* `SuppressTrimAnalysisWarnings` - "Manual override to enable/disable warnings"
  * global impact - it's important that both analyzer and linker behave the same (either warnings disabled or enabled)
  * typically not set by the user
  * acts as a manual override - should be defaulted in all cases


If we agree on this... I don't know how much of it we can make happen in 6 though.

vitek-karas: One more:

WPF/WinForms ->
* Disable warnings by default
* Disable analyzer (no point if warnings are disabled)
* IsTrimmed or PublishTrimmed should fail regardless of TFM (we will ask people to stick to 5.0 SDK for existing apps).
  * (Note that today we only fail for PublishTrimmed - I don't think we check IsTrimmed, but I think we should - that said, it's not a big deal)

sbomer: Summarizing our offline discussion - we agree on the intent behind all those properties. Mostly this is the case today except for downlevel TFMs. The behavior we want is:
- `PublishTrimmed` defaults for normal .NET6+:
  - `EnableTrimAnalyzer=true`
  - `SuppressTrimAnalysisWarnings` ideally empty, or it could be `false`
- `PublishTrimmed` defaults for downlevel/WinForms/WPF:
  - `EnableTrimAnalyzer=true`
    - (keep analyzer enabled in case it produces non-"trim analysis" warnings)
  - `SuppressTrimAnalysisWarnings=true`
    - _except_ if `EnableTrimAnalyzer` was explicitly set to true
- `IsTrimmable` defaults (independent of TFM):
  - `EnableTrimAnalyzer=true`
  - `SuppressTrimAnalysisWarnings` ideally empty, or could be `false`

I think the following logic roughly captures it:

```
if (PublishTrimmed) {
  if ((downlevel || winforms || wpf) && (EnableTrimAnalyzer != true)) {
    SuppressTrimAnalysisWarnings = true
  }
  EnableTrimAnalyzer = true
}

if (IsTrimmable) {
  EnableTrimAnalyzer = true
}
```

sbomer: We also discussed combining the `SuppressTrimAnalysisWarnings` and `EnableTrimAnalyzer` into one option but that's less clear - let's keep them as separate options for now so that it's possible to enable the trim analyzer and still suppress the trim analysis category of warnings. We could discuss merging them as an orthogonal change.

@vitek-karas @tlakollo please double-check to see if that summary is accurate and matches what we discussed.

vitek-karas: Looks good!

agocke: I think we have to keep `EnableTrimAnalyzer` set to true when `PublishTrimmed` is set because the analyzer checks the global config for the `EnableTrimAnalyzer` property to control scanning. It's too late/risky to do a new insertion to change that out, I think.

agocke: I'm looking at this code now -- just figuring out whether yanking the `EnableTrimAnalyzer` setting out of the group based on `EffectiveAnalysisLevel` is going to have any unforeseen consequences.

Side note: I think it would be useful to pass the `WarningLevel` variable to the linker in the future, so if we want to add new warnings for 7.0 we can do it based on the warning level.

agocke: OK, I've tried to update the code to match the logic above, with some modifications for MSBuild readability.

vitek-karas (approved):
I think this looks good. I know @sbomer wanted to keep the analyzer settings in the analyzer targets file, but if that's too complicated, I'm OK having it all in one place.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
sbomer: We already have a warning level equivalent for the linker even though we didn't use that capability for the 6.0 warnings:
https://github.com/dotnet/sdk/blob/7e0f692d6689c730ce6a22caf3a21444d5b609c7/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L215-L217

Is enabling the analyzer regardless of `EffectiveAnalysisLevel` what we want here? Using an analysis level <5 would disable trim warnings from the linker and I think should do the same for the trim analyzer.

agocke: Wait, so we use both the analysis level and `SuppressTrimAnalysisWarnings` to decide the linker warnings? Wouldn't setting analysis level < 5 only happen in situations where we also set `SuppressTrimAnalysisWarnings`? Or would this be something we change going forward, but is not the case now?

sbomer: Yeah, both the analysis level and the suppressions can disable them, but we're not doing the right thing for analysis level yet.

> Wouldn't setting analysis level < 5 only happen in situations where we also set SuppressTrimAnalysisWarnings

With the change, setting `EnableTrimAnalyzer` on downlevel frameworks could get you into this situation. The analyzer would show warnings but the linker wouldn't.

The other issue is that the trim analyzer doesn't care about the analysis level once it's enabled. I looked at the defaults for code quality warnings for comparison:

- `AnalysisLevel`
  - user-facing property, maps to an implied `EffectiveAnalysisLevel`
  - controls warning waves across the board for analyzers, Roslyn, and ILLink
  - default is set based on the TFM: 5 for .NET5, 'latest' for .NET6+.
  - empty for downlevel TFMs
- `EnableNETAnalyzers`
  - defaults to true when `AnalysisLevel` is 5+
  - can be used as a manual override to turn on various analyzers

It looks like the analyzers enabled via `EnableNETAnalyzers` show all warnings regardless of the analysis level just like the trim analyzer (at least for the one I checked). I don't know if this is the intended behavior, or if nobody has run into it, but I think when they introduce new warning waves the analyzers will need to respect the analysis level.

So maybe it's good enough to enable the trim analyzer and not worry about analysis level for now. It does mean that if you set `EnableTrimAnalyzer=true` on < 5, the linker won't show any warnings unless you also set the analysis level. And if we add warning waves to the analyzer we'll need to revisit this logic to update the warning level when `EnableTrimAnalyzer` is set.

sbomer (approved):
Thanks for fixing this!

PR #20972, opened 2021-09-13
Only throw missing reference path error when running with references

joperezr:
Today, when running package validation (that is not against a package baseline), we resolve references of the assets being compared. We also then provide an optional switch you can set (set `RunPackageValidationWithoutReferences` to true) in order to not use references if you prefer just compare the two assemblies in question without looking at the references. There is an issue with the code since if you decide to run with out references today, you would get a warning similar to this:

```
CP1003 Could not find a reference directory in the provided directories for TargetFramework Foo when loading 'lib/Foo/Bar.dll'. For more information see: https://aka.ms/dotnetpackagevalidation
```

This is because we are not funneling through the value of that property correctly to the analysis, so we are not making it aware that we actually don't care about references. These changes are fixing that bug, as well as adding test coverage for it.

cc: @Anipik @ericstj

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
+++ b/src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
@@ -116,7 +116,7 @@ namespace Microsoft.DotNet.PackageValidation
                 resolvedReferences = true;
                 loader.AddReferenceSearchDirectories(referencePathForTFM);
             }
-            else if (!_isBaselineSuppression && _referencePaths != null && ShouldLogDiagnosticId(ApiCompatibility.DiagnosticIds.SearchDirectoriesNotFoundForTfm))
+            else if (!_isBaselineSuppression && _referencePaths.Count != 0 && ShouldLogDiagnosticId(ApiCompatibility.DiagnosticIds.SearchDirectoriesNotFoundForTfm))
             {
                 _log.LogWarning(
                     new Suppression()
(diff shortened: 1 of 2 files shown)


src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -116,7 +116,7 @@ private IAssemblySymbol GetAssemblySymbolFromStream(Stream assemblyStream, Metad
                 resolvedReferences = true;
                 loader.AddReferenceSearchDirectories(referencePathForTFM);
             }
-            else if (!_isBaselineSuppression && _referencePaths != null && ShouldLogDiagnosticId(ApiCompatibility.DiagnosticIds.SearchDirectoriesNotFoundForTfm))
+            else if (!_isBaselineSuppression && _referencePaths.Count != 0 && ShouldLogDiagnosticId(ApiCompatibility.DiagnosticIds.SearchDirectoriesNotFoundForTfm))
```
ericstj, line 119: Looks like the `_referencePaths != null` check on line 110 can be removed as well.

ericstj (approved):

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
joperezr: Good catch. Fixed.

Anipik (approved):

PR #20979, opened 2021-09-14

v-wuzhai:


dsplaisted (approved):
I would suggest starting to do more of these changes together in the same PR.

PR #20980, opened 2021-09-14
[Blazor] Improves dotnet.js detection on ComputeBlazorPublishAssets

javiercn:
Given that now we do cache busting during build, this logic needs to be updated to account for it.


src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -195,22 +195,12 @@ public override bool Execute()
...
                     {
-                        var itemHash = FileHasher.GetFileHash(aotDotNetJs.ItemSpec);
-                        var cacheBustedDotNetJSFileName = $"dotnet.{BundledNETCoreAppPackageVersion}.{itemHash}.js";
-
-                        var originalFileFullPath = Path.GetFullPath(aotDotNetJs.ItemSpec);
-                        var originalFileDirectory = Path.GetDirectoryName(originalFileFullPath);
-
-                        var cacheBustedDotNetJSFullPath = Path.Combine(originalFileDirectory, cacheBustedDotNetJSFileName);
-
-                        newDotNetJs = new TaskItem(cacheBustedDotNetJSFullPath, asset.CloneCustomMetadata());
```
javiercn, line 206: We didn't/shouldn't need to set the cache busted path as the definition during publish, the original aot asset item spec should be enough

TanayParikh: Ah okay, I believe I'd done this as I wasn't getting any published assets without it, but if this works, all the better.

TanayParikh:
> /backport to main

Nevermind, `dotnet/sdk` doesn't support that. This'll need to be manually backported onto main.

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/BlazorWasmBaselineTests.cs
```diff
@@ -32,12 +32,12 @@ private string TemplatizeCompressedAssets(StaticWebAsset asset, string originalV
                 return null;
             }
 
-            if (originalValue.Replace("[[CustomPackageVersion]]", "__CustomVersion__").Contains("[["))
+            if (originalValue.Replace("[[CustomPackageVersion]]", "__CustomVersion__").Replace("[[hash]]", "__Hash__").Contains("[["))
```
TanayParikh, line 35: Where does this change materialize? Not seeing it in the baselines.

PR #20983, opened 2021-09-14
Hide unused workload verbosity options

sfoslund:


src/Cli/dotnet/CommonOptions.cs
```diff
@@ -25,6 +25,9 @@ internal static class CommonOptions
         public static Option<VerbosityOptions> VerbosityOption() =>
             VerbosityOption(o => $"-verbosity:{o}");
 
+        public static Option<VerbosityOptions> HiddenVerbosityOption() =>
+            VerbosityOption(o => $"-verbosity:{o}").Hide();
```
dsplaisted, line 29: I'm curious, would this work?  Or would that end up hiding both options?

```suggestion
            VerbosityOption().Hide();
```

dsplaisted (approved):

src/Cli/dotnet/CommonOptions.cs
sfoslund: That should work too!

PR #20987, opened 2021-09-14
Warn when Targeting net6.0 with 16.11 and the 6.0 SDK

sfoslund:
`NETSDK1182: Targeting .NET 6.0 in Visual Studio 2019 is not supported.`

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -832,4 +832,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1181: Error getting pack version: Pack '{0}' was not present in workload manifests.</value>
     <comment>{StrBegin="NETSDK1181: "}</comment>
   </data>
+  <data name="Net6NotCompatibleWithDev16" xml:space="preserve">
+    <value>NETSDK1182: Using .NET 6.0 in Visual Studio 2019 is not a supported scenario.</value>
+    <comment>{StrBegin="NETSDK1182: "}</comment>
+  </data>
 </root>
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -833,7 +833,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1181: "}</comment>
   </data>
   <data name="Net6NotCompatibleWithDev16" xml:space="preserve">
-    <value>NETSDK1182: Using .NET 6.0 in Visual Studio 2019 is not a supported scenario.</value>
+    <value>NETSDK1182: Targeting .NET 6.0 in Visual Studio 2019 is not supported.</value>
     <comment>{StrBegin="NETSDK1182: "}</comment>
   </data>
 </root>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
@@ -128,6 +128,12 @@ Copyright (c) .NET Foundation. All rights reserved.
                  FormatArguments="$(SdkResolverGlobalJsonPath)" />
   </Target>
 
+  <Target Name="_WarnWhenUsingNET6WithVS1611"
+        BeforeTargets="_CheckForInvalidConfigurationAndPlatform"
+        Condition="'$(BuildingInsideVisualStudio)' == 'true' and $(VisualStudioVersion.StartsWith('16'))">
+    <NetSdkWarning ResourceName="Net6NotCompatibleWithDev16"/>
+  </Target>
+
   <Target Name="_CheckForInvalidWindowsDesktopTargetingConfiguration"
         BeforeTargets="_CheckForInvalidConfigurationAndPlatform"
         Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0')) and ('$(UseWindowsForms)' == 'true' or '$(UseWPF)' == 'true')">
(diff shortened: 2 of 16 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
```diff
@@ -128,6 +128,12 @@ Copyright (c) .NET Foundation. All rights reserved.
                  FormatArguments="$(SdkResolverGlobalJsonPath)" />
   </Target>
 
+  <Target Name="_WarnWhenUsingNET6WithVS1611"
```
joeloff, line 131: This only checks if you're using VS 16, it doesn't check for specific minor versions, does it?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
```diff
@@ -128,6 +128,12 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+        BeforeTargets="_CheckForInvalidConfigurationAndPlatform"
+        Condition="'$(BuildingInsideVisualStudio)' == 'true' and $(VisualStudioVersion.StartsWith('16'))">
```
dsplaisted, line 133: Shouldn't we be using the MSBuild `VersionLessThan` comparison function here?

Also, I think we should generate this error on the command line too, ie remove the check for `BuildingInsideVisualStudio`.  If you are using the MSBuild version from VS 16.11 from the command line, it is also not supported and you are likely to run into issues, so we should generate the warning there too.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
sfoslund: Yeah, based on the warning text Kathleen suggested I assumed that was the desired behavior, @dsplaisted is that right or do we only want this with specifically 16.11?

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -832,4 +832,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1181: Error getting pack version: Pack '{0}' was not present in workload manifests.</value>
     <comment>{StrBegin="NETSDK1181: "}</comment>
   </data>
+  <data name="Net6NotCompatibleWithDev16" xml:space="preserve">
+    <value>NETSDK1182: Using .NET 6.0 in Visual Studio 2019 is not a supported scenario.</value>
```
dsplaisted, line 836: Have we had PM review of this string?  I'd suggest the following:

```suggestion
    <value>NETSDK1182: Targeting .NET 6.0 in Visual Studio 2019 is not supported.</value>
```

@KathleenDollard

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
joeloff: Yup, that's fine, I was more calling out that the target name contains "WithVS1611", when it's really just VS16

dsplaisted: Good catch Jacques, I think the target should be renamed.

src/Tasks/Common/Resources/Strings.resx
sfoslund: @KathleenDollard suggested this string, but she mentioned that she might update it again. @KathleenDollard I'll let you handle changing the string in this PR if that's okay with you.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
sfoslund: Sounds good, I'll remove the BuildingInsideVS check. As for the version comparison, that makes sense, but do we want to check if the version is exactly 16.11? Or check that it's anything below 17.0? (see this comment: https://github.com/dotnet/sdk/pull/20987#discussion_r708498680)

dsplaisted: We should check anything below 17.0, 16.10 or earlier will be at least as broken as 16.11.

src/Tasks/Common/Resources/Strings.resx
dsplaisted: It's not as simple as updating this in the PR, because you have to run a build to update all of the .xlf files (or manually update a bunch of them).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
```diff
@@ -128,6 +128,12 @@ Copyright (c) .NET Foundation. All rights reserved.
                  FormatArguments="$(SdkResolverGlobalJsonPath)" />
   </Target>
 
+  <Target Name="_WarnWhenUsingNET6AndVSPriorTo17"
```
dsplaisted, line 131: We also need to make this only warn when targeting .NET 6 and higher.

dsplaisted (changes requested):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
sfoslund: Added that condition 👍

dsplaisted (approved):
LGTM.  Since we don't have a test for this, we should make sure to manually test it.

sfoslund:
> we should make sure to manually test it.

Yup, I've tested manually. 

@KathleenDollard @marcpopMSFT I'm leaving this open for now since you mentioned possibly rewording the warning again and running this by tactics. Let me know when you have an update and I'll be happy to make the wording change/merge this.

sfoslund:
@dsplaisted @marcpopMSFT did you bring this up in tactics yesterday? Do we have approval?

dsplaisted:
> did you bring this up in tactics yesterday? Do we have approval?

We didn't have time to cover bugs in the meeting.  We should send an email to the alias.

Issue #20879, opened 2021-09-09
Support URL's in workload update's --from-rollback-file parameter


Redth:
Creating a rollback json file manually is not intuitive, and there is no easy/automated way to create this file otherwise.

If we added support for URL's for the `--from-rollback-file` parameter for `dotnet workload update` we could provide rollback files as part of .NET MAUI builds for example, and use a link to direct to a rollback file.  We would probably do this as part of MAUI's build pipeline, or repo, so that the rollback file always exists for a given branch's configuration/output (ie: `main` would have a rollback with all the workload versions from the latest builds in our main branch).

Redth:
//cc @dsplaisted @wli3

sfoslund:
Fixed in RC2 by #20991

dsplaisted:
Thanks a bunch Sarah and thanks @Redth for filing the issue!

🎉🎈🎉🎈

PR #20991, opened 2021-09-14
Allow specifying url for workload rollback

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/20879

Validated manually that specifying a url with the same option (`--from-rollback-file`) works.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -18,6 +18,8 @@ using NuGet.Common;
 using System.Text.Json;
 using System.Runtime.InteropServices;
 using Microsoft.DotNet.Cli;
+using System.Net;
+using System.Net.Http;
 
 namespace Microsoft.DotNet.Workloads.Workload.Install
 {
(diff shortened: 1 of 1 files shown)


Redth (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -428,11 +430,22 @@ private async Task<bool> NewerManifestPackageExists(ManifestId manifest)
...
+            if (File.Exists(rollbackDefinitionFilePath))
             {
-                throw new ArgumentException(string.Format(LocalizableStrings.RollbackDefinitionFileDoesNotExist, rollbackDefinitionFilePath));
+                fileContent = File.ReadAllText(rollbackDefinitionFilePath);
+            }
+            else
+            {
+                try
+                {
+                    fileContent = (new HttpClient()).GetStringAsync(rollbackDefinitionFilePath).Result;
```
dsplaisted, line 442: I think it would be better to try to check if the string looks like a URL (perhaps by checking if it starts with `http://` or `https://`, though it would be good to get guidance from someone who knows more) rather than just checking whether the file exists and then indiscriminately trying to download it.

Also, what does the error experience look like if the URI doesn't exist, or can't be reached?  I think we would want a somewhat "graceful" exception in that case, ie not printing the stack trace.

Related to that, we might want to print a message that we're trying to download the file before we do it, so that if it takes a while there is some feedback.

joeloff: Take a look at https://docs.microsoft.com/en-us/dotnet/api/system.uri.trycreate?view=net-5.0. I think you can create a URI, then verify the result's URI scheme to be HTTPS - I don't think we should allow for HTTP as that is not secure and we shouldn't trust non-HTTPS.

You could maybe try sending a HEAD request to the URL, but what if you're offline or restricted to an intranet?

sfoslund: Sure, I've converted to check if it can be parsed as a uri using the api @joeloff linked. When the URI doesn't exist or can't be reached we handle the resulting exception elsewhere in the code so it gets formatted before being displayed. 
I'm not adding a message before downloading the file for now since we're past the loc deadline.

sfoslund:
@dsplaisted do you have any other feedback here or will it be good to go in once CI passes?

dsplaisted:
> @dsplaisted do you have any other feedback here or will it be good to go in once CI passes?

No, go ahead and merge! :shipit:

PR #20993, opened 2021-09-14
Add a static cache for GetAssemblyVersion

rainersigwald:
In real-world projects like OrchardCore, ResolvePackageFileConflicts can
run on many files. Since it doesn't cache at all today, that can result
in reading assembly metadata out of the files many, many times.

Instead, cache the results of the call within GetAssemblyVersion and
check only the timestamp of the file on subsequent reads. Even that is
probably too conservative--we could consider caching files in the .NET
SDK directory "forever" (the lifetime of the MSBuild node), and possibly
expanding that to the NuGet packages directory as well, since its
contents should be immutable (but users do occasionally change them).

For this point in the release cycle I chose to be conservative and do
the timestamp check.

### DIFF
--- a/src/Tasks/Common/FileUtilities.MetadataReader.cs
+++ b/src/Tasks/Common/FileUtilities.MetadataReader.cs
@@ -23,25 +23,25 @@ namespace Microsoft.NET.Build.Tasks
 {
     static partial class FileUtilities
     {
-        private static Dictionary<string, (DateTime LastCheckedUtc, Version Version)> s_versionCache = new(StringComparer.OrdinalIgnoreCase /* Not strictly correct on *nix. Fix? */);
+        private static Dictionary<string, (DateTime LastKnownWriteTimeUtc, Version Version)> s_versionCache = new(StringComparer.OrdinalIgnoreCase /* Not strictly correct on *nix. Fix? */);
 
         private static Version GetAssemblyVersion(string sourcePath)
         {
             // TODO: is there a way to assert that it's a full path?
 
+            DateTime lastWriteTimeUtc = File.GetLastWriteTimeUtc(sourcePath);
+
             if (s_versionCache.TryGetValue(sourcePath, out var cacheEntry) 
-                && File.GetLastWriteTimeUtc(sourcePath) < cacheEntry.LastCheckedUtc)
+                && lastWriteTimeUtc < cacheEntry.LastKnownWriteTimeUtc)
             {
                 return cacheEntry.Version;
             }
-
-            DateTime checkTime = DateTime.UtcNow;
 
             Version version = GetAssemblyVersionFromFile(sourcePath);
 
             if (version is not null)
             {
-                s_versionCache[sourcePath] = (checkTime, version);
+                s_versionCache[sourcePath] = (lastWriteTimeUtc, version);
             }
 
             // TODO: clear or prune cache sometimes?
--- a/src/Tasks/Common/FileUtilities.MetadataReader.cs
+++ b/src/Tasks/Common/FileUtilities.MetadataReader.cs
@@ -36,10 +36,7 @@ namespace Microsoft.NET.Build.Tasks
 
             Version version = GetAssemblyVersionFromFile(sourcePath);
 
-            if (version is not null)
-            {
-                s_versionCache[sourcePath] = (lastWriteTimeUtc, version);
-            }
+            s_versionCache[sourcePath] = (lastWriteTimeUtc, version);
 
             // When introducing this cache, we decided that the cached
             // data was small and likely to be basically finite for
--- a/src/Tasks/Common/FileUtilities.MetadataReader.cs
+++ b/src/Tasks/Common/FileUtilities.MetadataReader.cs
@@ -14,7 +14,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Reflection.Metadata;
 using System.Reflection.PortableExecutable;
--- a/src/Tasks/Common/FileUtilities.MetadataReader.cs
+++ b/src/Tasks/Common/FileUtilities.MetadataReader.cs
@@ -32,7 +32,7 @@ namespace Microsoft.NET.Build.Tasks
             DateTime lastWriteTimeUtc = File.GetLastWriteTimeUtc(sourcePath);
 
             if (s_versionCache.TryGetValue(sourcePath, out var cacheEntry) 
-                && lastWriteTimeUtc < cacheEntry.LastKnownWriteTimeUtc)
+                && lastWriteTimeUtc == cacheEntry.LastKnownWriteTimeUtc)
             {
                 return cacheEntry.Version;
             }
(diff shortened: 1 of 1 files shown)


src/Tasks/Common/FileUtilities.MetadataReader.cs
```diff
@@ -21,31 +23,57 @@ namespace Microsoft.NET.Build.Tasks
...
     {
+        private static Dictionary<string, (DateTime LastCheckedUtc, Version Version)> s_versionCache = new(StringComparer.OrdinalIgnoreCase /* Not strictly correct on *nix. Fix? */);
+
         private static Version GetAssemblyVersion(string sourcePath)
         {
-            using (var assemblyStream = new FileStream(sourcePath, FileMode.Open, FileAccess.Read, FileShare.Delete | FileShare.Read))
+            // TODO: is there a way to assert that it's a full path?
+
+            if (s_versionCache.TryGetValue(sourcePath, out var cacheEntry) 
+                && File.GetLastWriteTimeUtc(sourcePath) < cacheEntry.LastCheckedUtc)
```
ladipro, line 33: Why did you choose to use this inequality rather than checking for an exact last-write-time match on the file? If the user overwrites the file with another version and this version was created before the cache entry was added, the condition will hold and `GetAssemblyVersion` will be returning an incorrect version.

FWIW, analogous checks in RAR are based on exact timestamp checks: https://github.com/dotnet/msbuild/blob/ea1d6d99a376ac582216bba06b75f6bd9f3c7c64/src/Tasks/SystemState.cs#L359-L364

rainersigwald: Yeah, that's better.

src/Tasks/Common/FileUtilities.MetadataReader.cs
```diff
@@ -21,31 +23,57 @@ namespace Microsoft.NET.Build.Tasks
...
+
         private static Version GetAssemblyVersion(string sourcePath)
         {
-            using (var assemblyStream = new FileStream(sourcePath, FileMode.Open, FileAccess.Read, FileShare.Delete | FileShare.Read))
+            // TODO: is there a way to assert that it's a full path?
+
+            DateTime lastWriteTimeUtc = File.GetLastWriteTimeUtc(sourcePath);
+
+            if (s_versionCache.TryGetValue(sourcePath, out var cacheEntry) 
+                && lastWriteTimeUtc < cacheEntry.LastKnownWriteTimeUtc)
```
ladipro, line 35: Did you mean

```suggestion
                && lastWriteTimeUtc == cacheEntry.LastKnownWriteTimeUtc)
```
?

src/Tasks/Common/FileUtilities.MetadataReader.cs
```diff
@@ -13,6 +13,8 @@
 #if NETCOREAPP || !EXTENSIONS
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
```
ladipro, line 17: nit: Redundant namespace import. Debugging leftover?

rainersigwald: Yeah, I was trying to put an assert in but couldn't figure out how to phrase it.

rainersigwald:
```suggestion
```

ladipro (approved):

rainersigwald:
Example improvement in "File Information Queries" for OrchardCore scenario "load, then change solution configuration":

```diff
-  1,303 - S:\NugetCache\packages\system.text.json\4.7.2\lib\netcoreapp3.0\System.Text.Json.dll
+  152   - S:\NugetCache\packages\system.text.json\4.7.2\lib\netcoreapp3.0\System.Text.Json.dll
```

src/Tasks/Common/FileUtilities.MetadataReader.cs
```diff
@@ -21,31 +22,57 @@ namespace Microsoft.NET.Build.Tasks
...
+            }
+
+            Version version = GetAssemblyVersionFromFile(sourcePath);
+
+            if (version is not null)
+            {
+                s_versionCache[sourcePath] = (lastWriteTimeUtc, version);
+            }
+
+            // TODO: clear or prune cache sometimes?
```
rainersigwald, line 46: This is something I'd like feedback on.

I can think of a few approaches:

1. Do nothing, let it grow: the cached information is small and while it _can_ grow indefinitely in practice it won't: you're going to build the same few things over and over again within an MSBuild node lifetime.
2. Clear entirely if cache size reaches some threshold. Easy, keeps memory bounded, but perf is funky because it periodically "starts over". Also have to tune the size.
3. Keep last-used timestamp, periodically purge old entries. Complex-ish (are we starting a thread on a timer to do this?) and potentialy funky perf characteristics (if no thread, do this every Nth call?)

I think I'm talking myself into 1 here. @ladipro, thoughts?

ladipro: I agree that 1. is OK. It is a common helper so probably not safe to assume any calling pattern but the upper bound of "all assemblies on disk ever participating in any build" is reasonable given how little data is stored.

src/Tasks/Common/FileUtilities.MetadataReader.cs
```diff
@@ -21,31 +22,60 @@ namespace Microsoft.NET.Build.Tasks
...
+                && lastWriteTimeUtc == cacheEntry.LastKnownWriteTimeUtc)
             {
-                Version result = null;
-                try
+                return cacheEntry.Version;
+            }
+
+            Version version = GetAssemblyVersionFromFile(sourcePath);
+
+            if (version is not null)
```
dsplaisted, line 39: Is there any reason not to cache a `null` return value for the version?  I think that means it wasn't a valid .NET assembly, but sometimes we do attempt to get the assembly version of those files, and it's not like the second time we try to do so it's going to be any more successful.

src/Tasks/Common/FileUtilities.MetadataReader.cs
```diff
@@ -21,31 +22,60 @@ namespace Microsoft.NET.Build.Tasks
 {
     static partial class FileUtilities
     {
+        private static Dictionary<string, (DateTime LastKnownWriteTimeUtc, Version Version)> s_versionCache = new(StringComparer.OrdinalIgnoreCase /* Not strictly correct on *nix. Fix? */);
```
dsplaisted, line 25: Do we need to worry about being thread-safe here?  I'm guessing not because MSBuild doesn't run tasks in parallel in the same process.

dsplaisted (approved):

src/Tasks/Common/FileUtilities.MetadataReader.cs
rainersigwald: Not that I can think of.

src/Tasks/Common/FileUtilities.MetadataReader.cs
rainersigwald: Correct; MSBuild should never run the same task concurrently in a process.

marcpopMSFT:
@rainersigwald , is this ready to merge?

rainersigwald:
Yes, sorry, didn't realize I had sufficient sign off.

PR #21022, opened 2021-09-15
Update BlazorWebAssembly targets to work with updated wasm targets

radical:
[This PR](https://github.com/dotnet/runtime/pull/57556) adds support for
relinking the native wasm files during the build (instead of only during
publish).

This required changes to the targets available, and how they can be
used, and thus require changes in the blazor targets to work with that.

The new setup is explained in
https://github.com/radical/runtime/blob/wasm-build/src/mono/wasm/build/README.md .

Forward ports the changes from release/6.0.1xx


pranavkm:
Could I get @javiercn to review this? He's much more familiar with the WASM SDK than I am.

javiercn (approved):
Looks great!

radical:
@javiercn I cherrypicked the fix that was added in rc2, since it's all related.

radical:
@javiercn Could you please review this with the cherry picked commit?

javiercn (approved):
Apologies,

I thought I signed-off.

radical:
> Apologies,
> 
> I thought I signed-off.

That's on me. You did sign-off, but then I added the short fix that was pushed to rc2, and so I asked for the review again! Thanks :)

PR #21027, opened 2021-09-15
Update ShowMissingWorkloads for IPA VS top level workloads

sfoslund:


sfoslund:
@dsplaisted @lewing is this the correct behavior for IPA? We want both `Microsoft.VisualStudio.Workload.NetCrossPlat` and `Microsoft.VisualStudio.ComponentGroup.Maui.All` for maui and `Microsoft.VisualStudio.Workload.NetWeb` for wasm?

src/Tasks/Microsoft.NET.Build.Tasks/ShowMissingWorkloads.cs
```diff
@@ -84,8 +88,10 @@ internal static string ToSafeId(string id)
         private static IEnumerable<string> GetSuggestedWorkloadsList(WorkloadInfo workloadInfo)
         {
             return MauiWorkloadIds.Contains(workloadInfo.Id.ToString()) ?
-                new string[] { ToSafeId(workloadInfo.Id), MauiTopLevelVSWorkload } :
-                new string[] { ToSafeId(workloadInfo.Id)  };
+                MauiTopLevelVSWorkloads.Append(ToSafeId(workloadInfo.Id)) :
```
dsplaisted, line 91: Nit:  I think this would be more readable with if / else statements rather than two different ternary operators.

Also, it's probably not worth creating a HashSet with at most two items.  So I'd consider using `yield return`, something like this:

```c#
yield return ToSafeId(workloadInfo.Id);
if (MauiWorkloadIds.Contains(workloadInfo.Id.ToString()))
{
   yield return MauiTopLevelVSCrossPlatWorkload;
   yield return MauiVSComponentMauiAll;
}
if (WasmWorkloadIds.Contains(workloadInfo.Id.ToString()))
{
   yield return WasmTopLevelVSCrossPlatWorkload;
}
```

This will also handle the case correctly if a project needs both WASM and Maui workloads.

dsplaisted (commented):
I'm not sure about the exact VS component IDs.  We should verify with @marcpopMSFT or @joeloff.

src/Tasks/Microsoft.NET.Build.Tasks/ShowMissingWorkloads.cs
sfoslund: LGTM, just updated to this logic.

dsplaisted (approved):

PR #21084, opened 2021-09-17

v-wuzhai:


src/Tests/Microsoft.DotNet.CommandFactory.Tests/GivenAProjectDependencyCommandResolver.cs
```diff
@@ -49,7 +49,7 @@ public void ItReturnsACommandSpecWhenToolIsInAProjectRef()
                 CommandName = "dotnet-portable",
                 Configuration = _configuration,
                 ProjectDirectory = testAsset.Path,
-                Framework = NuGet.Frameworks.FrameworkConstants.CommonFrameworks.NetCoreApp30
+                Framework = NuGet.Frameworks.FrameworkConstants.CommonFrameworks.Net60
```
dsplaisted, line 52: This would need to be updated each time we changed what the CurrentTargetFramework was.  Try the following:

```suggestion
                Framework = NuGet.Frameworks.NuGetFramework.Parse(ToolsetInfo.CurrentTargetFramework)
```

Likewise for the other changes in this file.

sfoslund, line 52: Can we do something similar to the `$(CurrentTargetFramework)` property here too so it'll be easier to change in the future? Maybe make a `CurrentTargetFramework` constant in the TestContext (https://github.com/dotnet/sdk/blob/main/src/Tests/Microsoft.NET.TestFramework/TestContext.cs)?

sfoslund (approved):

src/Tests/Microsoft.DotNet.CommandFactory.Tests/GivenAProjectDependencyCommandResolver.cs
sfoslund: Just saw Daniel's review, I agree with his comment.

PR #21085, opened 2021-09-17
Use a BlockingCollection in the logger to reduce CPU usage

### LINKED ISSUE #21083: TimestampedFileLogger is eating a lot of CPU
dotnet sdk 6.0.100-rc.1.21458.32

I have a case where a test is getting stuck (I believe this is unrelated to the SDK). However, while waiting for it, the parent `dotnet.exe` process is using a full CPU core.

Digging further, it seems like there is a tight loop in `TimestampedFileLogger`:

```csharp
        private void WriteLog(object flushThreshold)
        {
            int writeCount = 0;
            int threshold = (int)flushThreshold;

            while (!Done)
            {
                if (_messageQueue.TryDequeue(out string message))
                {
                    _stream.WriteLine($"{TimeStamp} {message}");
                    writeCount = (writeCount + 1) % threshold;

                    if (writeCount == 0)
                    {
                        _stream.Flush();
                    }
                }
            }
```

Because of that, it will keep using a full core until the test run is finished. I believe some synchronization around the queue would be needed here (just switching to a `BlockingCollection` would help a lot).

kevingosse:
Fixes https://github.com/dotnet/sdk/issues/21083

### DIFF
--- a/src/Cli/dotnet/Installer/Windows/TimestampedFileLogger.cs
+++ b/src/Cli/dotnet/Installer/Windows/TimestampedFileLogger.cs
@@ -24,7 +24,7 @@ namespace Microsoft.DotNet.Installer.Windows
         /// <summary>
         /// Thread safe queue use to store incoming log request messages.
         /// </summary>
-        private ConcurrentQueue<string> _messageQueue = new ConcurrentQueue<string>();
+        private BlockingCollection<string> _messageQueue = new BlockingCollection<string>();
 
         private bool _disposed;
         private readonly StreamWriter _stream;
@@ -42,8 +42,6 @@ namespace Microsoft.DotNet.Installer.Windows
         {
             get;
         }
-
-        private bool Done = false;
 
         private Thread LogWriter;
 
@@ -114,13 +112,14 @@ namespace Microsoft.DotNet.Installer.Windows
         {
             if (!_disposed)
             {
-                Done = true;
+                _messageQueue.CompleteAdding();
                 LogWriter?.Join();
                 _stream.WriteLine($"{TimeStamp} {FormatMessage("=== Logging ended ===")}");
 
                 if (disposing)
                 {
                     _stream?.Dispose();
+                    _messageQueue.Dispose();
                 }
 
                 _disposed = true;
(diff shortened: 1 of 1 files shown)


joeloff:
I am seeing one issue when OnProcessExit is called because it runs after the collection is disposed because the OnExit handler in the logger is called before the command's handler (which is intended to deal with and log abnormal process termination).

`
Unhandled exception. System.ObjectDisposedException: The collection has been disposed.
Object name: 'BlockingCollection'.
`

https://github.com/dotnet/sdk/blob/97fa42744bf80db38f161b225fa82408cb487329/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs#L866

executes after https://github.com/dotnet/sdk/blob/97fa42744bf80db38f161b225fa82408cb487329/src/Cli/dotnet/Installer/Windows/TimestampedFileLogger.cs#L135

It could just be an ordering issue. Unfortunately, unit tests won't catch these because the events would trigger when the test host terminates I think. The full install client+server requires UAC to trigger which you can't do with unit tests. OnExit is necessary to ensure critical services that may have been stopped like WUA are restarted (or at least a best effort is made to do so).

I do like the BlockingCollection though. I'll dig around a bit further and see if the event handling can be cleaned up. I'd very much like to see the BlockingCollection for RC2.

kevingosse:
The fix could be as simple as not disposing the BlockingCollection, the process is tearing down anyway so that shouldn't be an issue.
That said, it means that whatever gets logged in the `Shutdown` event will be discarded (and that was the same before this PR). So there's definitely something to change there.

joeloff:
We can still easily take fixes in RC2 for the SDK, and those would flow back to main. I have change prepped that includes both the BlockingCollection change and deals with the dispose exception. I can prep a PR for that for RC2 later today.

src/Cli/dotnet/Installer/Windows/TimestampedFileLogger.cs
```diff
@@ -24,7 +24,7 @@ internal class TimestampedFileLogger : SetupLoggerBase, IDisposable, ISetupLogge
         /// <summary>
         /// Thread safe queue use to store incoming log request messages.
         /// </summary>
-        private ConcurrentQueue<string> _messageQueue = new ConcurrentQueue<string>();
+        private BlockingCollection<string> _messageQueue = new BlockingCollection<string>();
```
WeihanLi, line 27: Since no need to update, maybe it can be `readonly`

kevingosse: Done in https://github.com/dotnet/sdk/pull/21085/commits/15cabbb8ac38e066a5dd028ce289098d64e1a141

kevingosse:
@joeloff I added a check to discard new logs if the logger is disposed, and catch the possible `ObjectDisposedException` because there's still a race condition (I could add interlocked operations to fix the race condition but I'm not sure it's worth it).

It means that shutdown logs will be discarded, but that was already the case with the previous implementation.

joeloff:
Yeah, I have an RC2 change that fixes the issue with the logs. It's related to how the client/server process shuts down where the server acknowledged the request immediately causing the client to shut down the log while the server was still performing operations that tried to write to the log. 

Can you target this PR for release/6.0.1xx-rc2  instead of main, then I can fix up the log issue. The RC2 changes will flow back to release/6.0.1xx and main

kevingosse:
@joeloff done

joeloff (approved):

joeloff:
Thank you @kevingosse

Issue #19739, opened 2021-08-14
Maui: Workload installation failed: The transaction has aborted 


salimhali:
hello
I have problem with workload install maui
I've tried to fix it but all my attempts were failed
same problem appears with maui-check command or dotnet workload install maui command
I need your help

Installing pack Microsoft.NETCore.App.Runtime.Mono.android-arm64 version 6.0.0-preview.7.21377.19...
Workload installation failed: The transaction has aborted.

this what appears exactly 

! Attempt to fix? [y/n] (y): y
» Attempting to fix: .NET SDK - Workloads (6.0.100-preview.7.21379.14)
SHELL: C:\Program Files\dotnet\dotnet.exe workload repair --source "https://api.nuget.org/v3/index.json" --source "https://pkgs.dev.azure.com/dnceng/public/_packaging/6.0.100-preview.7.21379.14-shipping-1/nuget/v3/index.json"

Repairing workload installation for workloads:
Garbage collecting for SDK feature bands ...

Successfully repaired worklods:

SHELL: C:\Program Files\dotnet\dotnet.exe workload update --sdk-version 6.0.100-preview.7.21379.14 --no-cache --disable-parallel --from-rollback-file "C:\Users\halis\AppData\Local\Temp\maui-check-dotnet-work-dir\workload.json" --source "https://api.nuget.org/v3/index.json" --source "https://pkgs.dev.azure.com/dnceng/public/_packaging/6.0.100-preview.7.21379.14-shipping-1/nuget/v3/index.json"

No workloads installed for this feature band. To update from the previous SDK, include the --from-previous-sdk option.
Updated advertising manifest microsoft.net.sdk.android.
Updated advertising manifest microsoft.net.sdk.ios.
Updated advertising manifest microsoft.net.sdk.maccatalyst.
Updated advertising manifest microsoft.net.sdk.macos.
Updated advertising manifest microsoft.net.sdk.maui.
Updated advertising manifest microsoft.net.sdk.tvos.
Updated advertising manifest microsoft.net.workload.emscripten.
Updated advertising manifest microsoft.net.workload.mono.toolchain.
Installing workload manifest microsoft.net.sdk.android version 30.0.100-preview.7.110.
Installing workload manifest microsoft.net.sdk.ios version 15.0.100-preview.7.230.
Installing workload manifest microsoft.net.sdk.maccatalyst version 15.0.100-preview.7.230.
[...]

Issue #21101, opened 2021-09-17
Replace TransactionScope


wli3:
We don't really use _TransactionScope_ sematic to the fullest. The download is not atomic, the "commit" action (the right thing is to use atomic operation like File.Move) is missing. At the same time, [TransactionScope has a hard code 10 minutes timeout in dotnet core.
](https://github.com/dotnet/runtime/blob/c43facbb363ddf49ec0e2e677fe124bc5666a940/src/libraries/System.Transactions.Local/src/System/Transactions/Configuration/ConfigurationStrings.cs#L9-L10) We should replace the usage of TransactionScope. Maybe a try catch block is good enough for our usage. It is too late in the release cycle, a workaround is used. We should try to replace that in the next release.

dsplaisted:
The runtime added this API, so I think we could remove our private reflection code: https://github.com/dotnet/sdk/blob/d2972edbb4ef60b1c58ec1c4e9b696aa28064821/src/Cli/dotnet/TransactionalAction.cs#L104-L105

PR #21102, opened 2021-09-17
Fix 10 minutes time out of TransactionScope

wli3:
https://github.com/dotnet/sdk/issues/21101

Fix https://github.com/dotnet/sdk/issues/19739

### DIFF
--- a/src/Cli/dotnet/TransactionalAction.cs
+++ b/src/Cli/dotnet/TransactionalAction.cs
@@ -2,12 +2,19 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Reflection;
 using System.Transactions;
+using Microsoft.DotNet.Cli.Utils;
 
 namespace Microsoft.DotNet.Cli
 {
     public sealed class TransactionalAction
     {
+        static TransactionalAction()
+        {
+            DisableTransactionTimeoutUpperLimit();
+        }
+
         private class EnlistmentNotification : IEnlistmentNotification
         {
             private Action _commit;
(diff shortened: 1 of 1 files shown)


wli3:
@dsplaisted could you have a look?

src/Cli/dotnet/TransactionalAction.cs
```diff
@@ -65,19 +67,44 @@ public void Rollback(Enlistment enlistment)
...
+            typeof(TransactionManager).GetField(fieldName, BindingFlags.NonPublic | BindingFlags.Static)
+                .SetValue(null, value);
+        }
+
+        // HACK: https://github.com/dotnet/sdk/issues/21101
+        // we should replace TransactionScope later.
+        private static TransactionScope CreateTransactionScope(TimeSpan timeout)
+        {
+            SetTransactionManagerField("s_cachedMaxTimeout", true);
+            SetTransactionManagerField("s_maximumTimeout", timeout);
```
dsplaisted, line 106: These are static fields, so it probably makes sense to just set them once rather than each time we create a transaction.

dsplaisted (approved):

wli3:
@mconnew thanks for the suggestion! Please help me double check the fix

src/Cli/dotnet/TransactionalAction.cs
```diff
@@ -65,19 +72,44 @@ public void Rollback(Enlistment enlistment)
...
-                TransactionScopeOption.Required,
-                TimeSpan.Zero))
+            try
             {
-                Transaction.Current.EnlistVolatile(
-                    new EnlistmentNotification(commit, rollback),
-                    EnlistmentOptions.None);
+                using (var scope = new TransactionScope(
+                    TransactionScopeOption.Required,
+                    TimeSpan.Zero))
```
mconnew, line 79: Are you sure you want an infinite timeout? A really high timeout will still eventually timeout whereas something going wrong with an infinite timeout can cause things to block indefinitely.

dsplaisted: I think that the other operations happening within the transaction have their own timeouts, so it wouldn't end up being effectively infinite.

mattleibow:
How does one test this PR? I can test locally here and make sure we got it :)

dsplaisted:
> How does one test this PR? I can test locally here and make sure we got it :)

Build the repo locally, then run `eng\dogfood.cmd` to get into an environment which will use the just-built SDK.  Then you can try doing a workload install for that.

See also https://github.com/dotnet/sdk/blob/main/documentation/project-docs/developer-guide.md#using-the-built-dotnet

wli3:
@mattleibow I tested it, it does work.

That's how I test, after running dogfood build, get an VPN (I used my home machine). Find the furthest node so the download would need more than 10 minutes. Run `dotnet workload install maui`. Without this change, the installation would fail in 10 minutes, with it, the installation will not fail.

wli3:
@mattleibow let me know when it is good. I do need to get this in soon.

wli3:
Talked to Matthew offline. This is good to merge for now

PR #21103, opened 2021-09-17
Create UpdateToNewTargetFramework.md

marcpopMSFT:


dsplaisted (commented):
When viewed as markdown, the line wrapping isn't as you'd expect it.  Probably you want to make bulleted lists, or I think you can end a line with two spaces to force a line wrap.

PR #21104, opened 2021-09-17
[release/6.0.1xx-rc2] BlazorWasm: Fix including the target assembly

radical:
Use `$(IntermediateAssembly)` when gathering files for wasm build.

Fixes https://github.com/dotnet/runtime/issues/59255 .

### DIFF
--- a/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
+++ b/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
@@ -114,7 +114,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <Target Name="_GatherWasmFilesToBuild">
     <ItemGroup>
       <WasmAssembliesToBundle Remove="@(WasmAssembliesToBundle)" />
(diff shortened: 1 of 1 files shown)


SteveSandersonMS (approved):
Makes sense based on what I saw before. Will this make it able to pick up DllImport from ProjectReferences too (if it didn’t already)?

pranavkm (approved):

radical:
> Will this make it able to pick up DllImport from ProjectReferences too (if it didn’t already)?

Can you elaborate on this?

lewing (approved):

SteveSandersonMS:
I just wanted to check that the issue that prevented it seeing your main assembly didn’t also prevent it from seeing the assemblies of other projects in your solution. 

We know the feature works for PackageReference dependencies, but does it work for ProjectReference dependencies? Has that been tried out?

Hope that makes sense!

radical:
> I just wanted to check that the issue that prevented it seeing your main assembly didn’t also prevent it from seeing the assemblies of other projects in your solution.

Yes, dllimports in referenced projects would get scanned for pinvokes during the native build. Currently, the native library won't get automatically included in the app project though. IOW, you would need an explicit `<NativeFileReference Include="/path/to/native/lib" />` in the app project.
But we are trying to figure out how the experience could be different.

lewing:
@marcpopMSFT what is the process to get this in rc2 now?

Issue #21009, opened 2021-09-15
help usage text missing workload 


Varorbc:
https://github.com/dotnet/sdk/blob/release/6.0.1xx/src/Cli/dotnet/commands/dotnet-help/HelpUsageText.cs

sfoslund:
Good catch, thanks!

sfoslund:
Fixed by #21177

PR #21177, opened 2021-09-20
Update top level help with workload string

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/21009

### DIFF
--- a/src/Cli/dotnet/BuiltInCommandsCatalog.cs
+++ b/src/Cli/dotnet/BuiltInCommandsCatalog.cs
@@ -125,7 +125,7 @@ namespace Microsoft.DotNet.Cli
             ["sdk"] = new BuiltInCommandMetadata
             {
                 Command = SdkCommand.Run,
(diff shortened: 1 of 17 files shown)


src/Cli/dotnet/BuiltInCommandsCatalog.cs
```diff
@@ -125,6 +125,7 @@ public static class BuiltInCommandsCatalog
             ["sdk"] = new BuiltInCommandMetadata
             {
                 Command = SdkCommand.Run,
+                DocLink = "https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-sdk-check"
```
joeloff, line 128: Should these be aka.ms links in case pages get shuffled around?

joeloff (approved):

src/Cli/dotnet/BuiltInCommandsCatalog.cs
sfoslund: Good idea, just updated.

joeloff (approved):

PR #21236, opened 2021-09-21
Always reference .NET Framework reference assembly packages if using static GraphWithoutRuntimeDependencies

### LINKED ISSUE #19506: Microsoft.NETFramework.ReferenceAssemblies not added when using NuGet static restore graph
In Roslyn we still build some packages against net45. The .NET Framework 4.5 targeting pack is not longer installed by default (or even available) in the VS2022 installer. 

We would expect if all of the following occur, then include the package:
- AutomaticallyUseReferenceAssemblyPackages is true
- The targeting pack is not known to be available at time of restore
- NuGet static restore graph is used (currently this implies the previous item)

dsplaisted:
Fixes #19506

@jeffkl @cdmihai for review.  There's not a good way to test the end-to-end scenario here, as it involves creating a static graph on one machine, and building on another, where one of the machines has the targeting pack installed and the other doesn't.  So in the test I added, I just search for a version of .NET Framework where the targeting pack is installed, and verify that it's still referenced when doing a static graph build.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
@@ -396,12 +396,17 @@ Copyright (c) .NET Foundation. All rights reserved.
     -->
   <Target Name="IncludeTargetingPackReference" BeforeTargets="_CheckForInvalidConfigurationAndPlatform;CollectPackageReferences;CheckForImplicitPackageReferenceOverrides"
           Condition="'$(TargetFrameworkMoniker)' != '' and '$(TargetFrameworkIdentifier)' == '.NETFramework' and '$(AutomaticallyUseReferenceAssemblyPackages)' == 'true'">
+    
+    <!-- Skip this task if is using static graph, because in that case we don't want the NuGet packages referenced to change based on machine state, so we should always
+         reference the reference assembly package. -->
     <GetReferenceAssemblyPaths
         TargetFrameworkMoniker="$(TargetFrameworkMoniker)"
         RootPath="$(TargetFrameworkRootPath)"
         TargetFrameworkFallbackSearchPaths="$(TargetFrameworkFallbackSearchPaths)"
         BypassFrameworkInstallChecks="true"
-        SuppressNotFoundError="true">
+        SuppressNotFoundError="true"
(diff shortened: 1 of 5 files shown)


jeffkl (approved):
LGTM

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
```diff
@@ -396,12 +396,17 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <!-- Skip this task if is using static graph, because in that case we don't want the NuGet packages referenced to change based on machine state, so we should always
+         reference the reference assembly package. -->
     <GetReferenceAssemblyPaths
         TargetFrameworkMoniker="$(TargetFrameworkMoniker)"
         RootPath="$(TargetFrameworkRootPath)"
         TargetFrameworkFallbackSearchPaths="$(TargetFrameworkFallbackSearchPaths)"
         BypassFrameworkInstallChecks="true"
-        SuppressNotFoundError="true">
+        SuppressNotFoundError="true"
+        Condition="'$(IsGraphBuild)' != 'true'"
```
cdmihai, line 408: Isn't this hiding the real problem that for some reason static graph based restore leads to different reference assembly paths?

dsplaisted: Not as far as I understand it.  I think the different reference assembly paths were caused because the reference assemblies installed in Program Files would be used if they were installed, otherwise the NuGet packages with those reference assemblies would be referenced and they would be used from the package.

This change makes it so that for graph build, it will always use the reference assembly NuGet packages, whether or not the reference assemblies are installed in Program Files.

cdmihai: It could be that I also misunderstand the problem. Does the static graph based restore cause any difference in the ref assembly paths? If yes, then the root cause for that should be a bug right? Because using static graph should not lead to observable differences in build output.

dsplaisted:
Let's not take this for 6.0.1xx, as I [realized](https://github.com/dotnet/sdk/issues/19506#issuecomment-929536537] I was confusing static graph with NuGet lock files and don't actually know what the static graph issue is.

PR #21283, opened 2021-09-22
Add rollback option to install command and allow missing manifest ids

sfoslund:
- Add `--from-rollback-file` to install command
- Allow rollback definition to be missing manifest ids. This allows the rollback definition to include a subset of the currently installed manifests, instead of requiring all manifests be present in the rollback file. It still errors when an unrecognized manifest id exists in the rollback file. 
- Warn instead of error when a rollback definition has extra manifest ids that are not recognized by the SDK. With this change, any unrecognized ids will now cause a warning then be ignored.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
@@ -45,6 +45,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
         private readonly string _userProfileDir;
         private readonly string _tempDirPath;
         private readonly string _dotnetPath;
+        private readonly string _fromRollbackDefinition;
 
         public WorkloadInstallCommand(
             ParseResult parseResult,
@@ -75,6 +76,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             _tempDirPath = tempDirPath ?? (string.IsNullOrWhiteSpace(parseResult.ValueForOption<string>(WorkloadInstallCommandParser.TempDirOption)) ?
                 Path.GetTempPath() :
                 parseResult.ValueForOption<string>(WorkloadInstallCommandParser.TempDirOption));
+            _fromRollbackDefinition = parseResult.ValueForOption<string>(WorkloadInstallCommandParser.FromRollbackFileOption);
 
             var configOption = parseResult.ValueForOption<string>(WorkloadInstallCommandParser.ConfigOption);
             var sourceOption = parseResult.ValueForOption<string[]>(WorkloadInstallCommandParser.SourceOption);
@@ -176,7 +178,9 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                 workloadIds = workloadIds.Concat(installedWorkloads).Distinct();
 
                 _workloadManifestUpdater.UpdateAdvertisingManifestsAsync(includePreviews, offlineCache).Wait();
-                manifestsToUpdate = _workloadManifestUpdater.CalculateManifestUpdates().Select(m => (m.manifestId, m.existingVersion, m.newVersion));
+                manifestsToUpdate = string.IsNullOrWhiteSpace(_fromRollbackDefinition) ?
+                    _workloadManifestUpdater.CalculateManifestUpdates().Select(m => (m.manifestId, m.existingVersion, m.newVersion)) :
+                    _workloadManifestUpdater.CalculateManifestRollbacks(_fromRollbackDefinition);
             }
 
             InstallWorkloadsWithInstallRecord(workloadIds, _sdkFeatureBand, manifestsToUpdate, offlineCache);
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
@@ -61,6 +61,8 @@ namespace Microsoft.DotNet.Cli
 
         public static readonly Option<VerbosityOptions> VerbosityOption = CommonOptions.VerbosityOption();
 
+        public static readonly Option<string> FromRollbackFileOption = WorkloadUpdateCommandParser.FromRollbackFileOption;
+
         public static Command GetCommand()
         {
             var command = new Command("install", LocalizableStrings.CommandDescription);
@@ -84,6 +86,7 @@ namespace Microsoft.DotNet.Cli
             command.AddOption(TempDirOption);
             command.AddWorkloadCommandNuGetRestoreActionConfigOptions();
             command.AddOption(VerbosityOption);
+            command.AddOption(FromRollbackFileOption);
         }
     }
 }
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -210,7 +210,8 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
             if (!manifestRollbacks.All(rollbackManifest => currentManifestIds.Contains(rollbackManifest.Item1)))
             {
-                throw new Exception(string.Format(LocalizableStrings.RollbackDefinitionContainsExtraneousManifestIds, rollbackDefinitionFilePath));
+                _reporter.WriteLine(string.Format(LocalizableStrings.RollbackDefinitionContainsExtraneousManifestIds, rollbackDefinitionFilePath).Yellow());
+                manifestRollbacks = manifestRollbacks.Where(rollbackManifest => currentManifestIds.Contains(rollbackManifest.Item1));
             }
 
             var manifestUpdates = manifestRollbacks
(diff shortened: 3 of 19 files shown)


dsplaisted:
If we allow manifest IDs to be omitted from the rollback definition, then it would be a lot easier to get into a state where you have a mismatched set of workload versions.

sfoslund:
> If we allow manifest IDs to be omitted from the rollback definition, then it would be a lot easier to get into a state where you have a mismatched set of workload versions.

Definitely. Based on the meeting this afternoon, I believe the maui team plans to use this feature to avoid including the blazor manifest version in their rollback files, is that correct @Redth? Since there won't be any overlap between maui and blazor, this should be fine, but obviously adding this feature opens us up to others using it in ways that could be breaking to their installations.

Redth:
There's a couple scenarios with the rollback file we are hoping to ensure work:

1. Arollback file with workload ids in it which are not already installed should either skip, or maybe install the missing workload, but not result in an error 

2. A rollback file without a workload id / version in it for a workload which is already installed should not result in an error. 

For 2 I think we maybe could scope it to be just workloads that the baseline sdk doesn't know about.

This is mostly about the case where there is a workload we don't know about and therefore can't reasonably suggest a version for, which is really right now just Tizen.  This might be a bigger problem in the future but we can provide a version as of now for all of the workload ids the manifest is aware of in the rollback files we generate in the Maui build if that simplifies things.

Redth:
@dsplaisted also is this much different than providing a bad/mismatched version in the rollback file to begin with? I could imagine someone just picking the current or some arbitrary version of workloads they don't know enough about to provide a good version for to make this work. 

I guess it's a bit more intentional that way, but using a rollback file seems pretty intentional in and of itself despite the contents.

mattleibow:
Right now, there is an issue when I update to a preview build of dotnet using the dotnet6 feed (https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json) as it just pulls the latest and they never works all together with all the bits. And right now, RTM is in there so there is no way to test RC2 easily. And also, iOS was having some issues/delays pushing bits, so we got RTM .net, preview 9 android and preview 8 ios. And they are not happy campers.

The only way to use that feed is to use a rollback with the set of approved versions.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -176,7 +178,9 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
                 workloadIds = workloadIds.Concat(installedWorkloads).Distinct();
 
                 _workloadManifestUpdater.UpdateAdvertisingManifestsAsync(includePreviews, offlineCache).Wait();
-                manifestsToUpdate = _workloadManifestUpdater.CalculateManifestUpdates().Select(m => (m.manifestId, m.existingVersion, m.newVersion));
+                manifestsToUpdate = string.IsNullOrWhiteSpace(_fromRollbackDefinition) ?
+                    _workloadManifestUpdater.CalculateManifestUpdates().Select(m => (m.manifestId, m.existingVersion, m.newVersion)) :
+                    _workloadManifestUpdater.CalculateManifestRollbacks(_fromRollbackDefinition);
```
jonathanpeppers, line 183: Related to @dsplaisted's point: https://github.com/dotnet/sdk/pull/21283#issuecomment-925369614

This looks like it either updates *or* uses the rollback file.

Would this potentially break if:
1. You installed some .NET SDK, let's say it's RC2 on release day.
2. You give it a rollback file such as:
```json
{"microsoft.net.sdk.android":"31.0.101-preview.9.6"}
```
```
> dotnet workload install android --from-rollback-file foo.json
```
3. Would the install pretty much fail every time, when you don't list `microsoft.net.workload.mono.toolchain`? It is a dependency of the `android` workload, and the version of the various Mono baseline manifests won't be on NuGet.org?

sfoslund:
> There's a couple scenarios with the rollback file we are hoping to ensure work:
A rollback file with workload ids in it which are not already installed should either skip, or maybe install the missing workload, but not result in an error
A rollback file without a workload id / version in it for a workload which is already installed should not result in an error.
For 2 I think we maybe could scope it to be just workloads that the baseline sdk doesn't know about.

I have concerns about the security of allowing a rollback file to specify a manifest id that the base SDK doesn't know about. This essentially forces the SDK to download a potentially malicious nuget package. If the SDK is signed then we have the protection of sign checking this package, but with an unsigned SDK we could be installing a malicious set of files into our dotnet installation or, if it's an admin install, even running a malicious MSI.

Redth:
> > There's a couple scenarios with the rollback file we are hoping to ensure work:
> 
> A rollback file with workload ids in it which are not already installed should either skip, or maybe install the missing workload, but not result in an error
> 
> A rollback file without a workload id / version in it for a workload which is already installed should not result in an error.
> 
> For 2 I think we maybe could scope it to be just workloads that the baseline sdk doesn't know about.
> 
> 
> 
> I have concerns about the security of allowing a rollback file to specify a manifest id that the base SDK doesn't know about. This essentially forces the SDK to download a potentially malicious nuget package. If the SDK is signed then we have the protection of sign checking this package, but with an unsigned SDK we could be installing a malicious set of files into our dotnet installation or, if it's an admin install, even running a malicious MSI. 

Totally agree with those concerns. I think it would be fine to just 'ignore' unknown manifest ids completely in the rollback file. Don't try to install them, don't fail if they are installed but the rollback file doesn't list them.

sfoslund:
> I think it would be fine to just 'ignore' unknown manifest ids completely in the rollback file. Don't try to install them, don't fail if they are installed but the rollback file doesn't list them.

This sounds like a good approach to me. As long as @joeloff  and @dsplaisted don't have any concerns I can add that change to this PR.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommandParser.cs
```diff
@@ -61,6 +61,8 @@ internal static class WorkloadInstallCommandParser
 
         public static readonly Option<VerbosityOptions> VerbosityOption = CommonOptions.VerbosityOption();
 
+        public static readonly Option<string> FromRollbackFileOption = WorkloadUpdateCommandParser.FromRollbackFileOption;
```
marcpopMSFT, line 64: should it be hidden like it is in update?

sfoslund: We're just reusing the same option here so the isHidden value will be true for both commands.

joeloff:
> > I think it would be fine to just 'ignore' unknown manifest ids completely in the rollback file. Don't try to install them, don't fail if they are installed but the rollback file doesn't list them.
> 
> This sounds like a good approach to me. As long as @joeloff and @dsplaisted don't have any concerns I can add that change to this PR.

No concerns, I don't even know whether this works for admin installs. I'd be surprised if it does.

sfoslund:
My latest commit removes the error when an extra, unrecognized manifest id is included in the rollback definition and instead it warns then ignores the given manifest id(s). I've updated the PR description so it's up to date on all of the changes this PR includes, @Redth @mattleibow @jonathanpeppers are there any other feature requests I'm missing here?

@joeloff @dsplaisted @wli3 this PR should be ready for review.

Redth:
> > > I think it would be fine to just 'ignore' unknown manifest ids completely in the rollback file. Don't try to install them, don't fail if they are installed but the rollback file doesn't list them.
> 
> > 
> 
> > This sounds like a good approach to me. As long as @joeloff and @dsplaisted don't have any concerns I can add that change to this PR.
> 
> 
> 
> No concerns, I don't even know whether this works for admin installs. I'd be surprised if it does.

Do you just mean you need to run it elevated to work for a global/admin install?  

Otherwise this looks good, thanks!

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -208,9 +208,10 @@ public IEnumerable<(ManifestId manifestId, ManifestVersion existingVersion, Mani
             var currentManifestIds = GetInstalledManifestIds();
             var manifestRollbacks = ParseRollbackDefinitionFile(rollbackDefinitionFilePath);
 
-            if (!new HashSet<ManifestId>(currentManifestIds).SetEquals(manifestRollbacks.Select(manifest => manifest.Item1)))
+            if (!manifestRollbacks.All(rollbackManifest => currentManifestIds.Contains(rollbackManifest.Item1)))
             {
-                throw new Exception(string.Format(LocalizableStrings.RollbackDefinitionContainsExtraneousManifestIds, rollbackDefinitionFilePath));
+                _reporter.WriteLine(string.Format(LocalizableStrings.RollbackDefinitionContainsExtraneousManifestIds, rollbackDefinitionFilePath).Yellow());
```
dsplaisted, line 213: This message should include the manifest IDs that weren't found.  If you don't want to change the string I think you could just add them to the end.

dsplaisted (approved):

steveisok:
I feel like this type of change meets the bar for GA and we should try to get it in.

sfoslund:
Sounds good, I have a PR out and I'll send a request to get it into GA: https://github.com/dotnet/sdk/pull/22190

PR #21358, opened 2021-09-24
Fix failure when manifest that is expected to exist isn't found

dsplaisted:
Fixes the fallback code which is supposed to find workload manifests from previous versions.  It was not correctly handling the case where the manifest it expects to find isn't found.  It ended up returning an empty string and adding that to the list of manifest directories, which later on caused the manifest reader to look for WorkloadManifest.json in whatever the current directory is, causing the workload resolver to fail when it couldn't find that file.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1409464 (though separately it may still be a problem that the IncludedWorkloadManifests.txt file lists manifests that aren't installed by VS).

@sfoslund Can you advise on whether there's an easy way to add test coverage for this?


sfoslund:
> Can you advise on whether there's an easy way to add test coverage for this?

You can add a test modeled off this one: https://github.com/dotnet/sdk/pull/19545/files#diff-6960cb26569ca2b37bee7066bec4dc96de422d44f157d265e278e6eef7fc2c4aR286

sfoslund (approved):

dsplaisted:
@sfoslund Thanks, I've added test coverage for this

sfoslund (approved):

sfoslund:
@dsplaisted is this ready for a tell mode notification/ to merge?

PR #21390, opened 2021-09-24
Clean up client/server shutdown

joeloff:
The server previously responded immediately when it received a shutdown request, causing the client to terminate before the server can log all necessary shutdown steps, especially around WUA restarts.

The process exit event handler in the logger also caused issues because it would run the wrong order before the client's handler. Instead, the client will create the handler and call the logger's dispose to close out the log file and flush out pending requests from both the client/server./

This would also address the previous object disposed exception that could occur


joeloff:
Gentle ping if anyone has time to take a peek so this can get into RC2

sfoslund (approved):
Sorry for the delay!

PR #21429, opened 2021-09-27
Make aspnet-browser-refresh.js tolerate duplication insertions

SteveSandersonMS:
## Description

Allows `aspnet-browser-refresh.js` to be inserted multiple times, so that it doesn't matter if it's both auto-injected and a developer adds it manually.

## Customer Impact

There are cases where we can't auto-inject `aspnet-browser-refresh.js` (the script that handles the JS side of hot reload), and in those cases we tell developers to add it to their page manually. But this is problematic if, in some cases, we *can* auto-inject it, because then it ends up being referenced twice and this leads to errors.

This PR is a very trivial tweak that makes duplicate insertions no-op to simplify things for developers in this case.

## Regression?
- [ ] Yes
- [x] No

## Risk
- [ ] High
- [ ] Medium
- [x] Low

Code change is super simple, and has been manually verified.

## Verification
- [x] Manual (required)
- [ ] Automated

- [ ] Yes
- [ ] No
- [x] N/A

Addresses https://github.com/dotnet/aspnetcore/issues/35334

### DIFF
--- a/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
+++ b/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
@@ -76,7 +76,7 @@ setTimeout(async function () {
       document.querySelector(`link[href^="${document.baseURI}${path}"]`);
 
     // Receive a Clear-site-data header.
-    await fetch('_framework/clear-browser-cache');
+    await fetch('/_framework/clear-browser-cache');
 
     if (!styleElement || !styleElement.parentNode) {
       console.debug('Unable to find a stylesheet to update. Updating all local css files.');
@@ -140,7 +140,7 @@ setTimeout(async function () {
       }
     });
 
-    fetch('_framework/blazor-hotreload', { method: 'post', headers: { 'content-type': 'application/json' }, body: JSON.stringify(deltas) })
+    fetch('/_framework/blazor-hotreload', { method: 'post', headers: { 'content-type': 'application/json' }, body: JSON.stringify(deltas) })
       .then(response => {
         if (response.status == 200) {
           const etag = response.headers['etag'];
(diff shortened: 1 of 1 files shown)


src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
```diff
@@ -69,7 +76,7 @@ setTimeout(async function () {
       document.querySelector(`link[href^="${document.baseURI}${path}"]`);
 
     // Receive a Clear-site-data header.
-    await fetch('/_framework/clear-browser-cache');
+    await fetch('_framework/clear-browser-cache');
```
SteveSandersonMS, line 79: This is to make it consistent with how we fetch `_framework/blazor-hotreload` and support non-root-hosting. Please let me know if there was some specific reason why this particular URL should always be host-rooted.

pranavkm: Wouldn't this bork if your page is being served from `/SubDir/MyPage`? The middleware is only listening to `$host/_framework/clear-browser-cache`?

SteveSandersonMS: In that case yes, we need the leading slash. How come we don’t also need it for _framework/blazor-hotreload? I was trying to make them consistent.

pranavkm: That sounds like a bug. Could you update the `_framework/blazor-hotreload` to have the leading slash?

SteveSandersonMS: Done

SteveSandersonMS:
@pranavkm Ready for re-review.

pranavkm (approved):

pranavkm:
Thanks!

PR #21460, opened 2021-09-27
Prevent GlobalUsings file from being cleaned.

pranavkm:
Performing a clean removes all files that were recorded using FileWrites.
In VS performing a clean and doing nothing starts producing build warnings for missing namespaces since the GlobalUsings file
has gone amiss. This change ensures that the file is not recorded and consequently not removed as part of "Clean".

---

### Description

In 6.0, the SDK generates a file containing global usings for projects with ImplicitUsings=true configured. In VS, performing a Clean deletes the file and starts producing build warnings since all of the globally added namespaces are no longer available. We think a fix ultimately needs to come from VS (e.g. kicking off an additional design time build after a Clean), In the meanwhile this seems like an acceptable workaround.

### Testing

[x] Manual
[x] Automated

### Risk
Low.


pranavkm:
@DamianEdwards do we want to fix it for rc2?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateGlobalUsings.targets
```diff
@@ -50,7 +50,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <ItemGroup>
       <Compile Include="$(GeneratedGlobalUsingsFile)" />
-      <FileWrites Include="$(GeneratedGlobalUsingsFile)" />
+      <!-- Do not put in FileWrites: performing a clean and waiting without a build results in build warnings in VS due to all the missing namespaces -->
```
dsplaisted, line 53: I think at a minimum, we should only omit this from FileWrites if we're building inside VS.

Ideally, we should solve this a different way.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateGlobalUsings.targets
```diff
@@ -50,7 +50,8 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <ItemGroup>
       <Compile Include="$(GeneratedGlobalUsingsFile)" />
-      <FileWrites Include="$(GeneratedGlobalUsingsFile)" />
+      <!-- Performing a clean and waiting without a build results in build warnings in VS due to all the missing namespaces, so we'll avoid recording it in FileWrites when building in VS -->
+      <FileWrites Include="$(GeneratedGlobalUsingsFile)" Condition="'$(BuildingInsideVisualStudio)' != 'true'" />
```
pranavkm, line 54: @dsplaisted updated

dsplaisted (approved):
We are OK with this fix for this release.  Past that, we want this to be solved a different way.

Issue #21560, opened 2021-09-29
NETSDK1179 (for `-r`) over-reports


richlander:
The following call to `dotnet run` should not warn.

```bash
rich@MacBook-Air-M1-2020 app % dotnet run -r osx-arm64 --self-contained false
/usr/local/share/dotnet/x64/sdk/6.0.100-rc.2.21478.25/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1110,5): warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [/Users/rich/app/app.csproj]
Hello, Arm64!

sfoslund:
Fixed in RC2 by #21561

PR #21561, opened 2021-09-29
Fix over-warning bug when specifying runtime CLI option on dotnet run

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/21560


sfoslund:
@marcpopMSFT We're in M2 mode starting at 4, right? When does M2 mode end?

dsplaisted (approved):
So the issue wasn't so much that it was warning, but that the `--self-contained` parameter wasn't supported at all by `dotnet run`?

sfoslund:
> So the issue wasn't so much that it was warning, but that the --self-contained parameter wasn't supported at all by dotnet run?

Yep, at least not explicitly.

marcpopMSFT:
M2 mode ends Friday at 4pm

joeloff (approved):

sfoslund:
Approved by Don

PR #21654, opened 2021-10-01
Simplify hash implementation

martincostello:
Use `SHA256.HashData()` and `Convert.ToHexString()` to simplify the code that SHA-256 hashes strings, which also cleans up the `SHA256` instance that was being used but not being disposed of every time a string was hashed.


marcpopMSFT:
@martincostello did you test that the resulting hash didn't change with this?

martincostello:
There's some existing tests that went through the hashing code paths that I assumed would verify that, happy to add some specific tests for some known values and their corresponding hashes to verify further though if that's needed.

https://github.com/dotnet/sdk/blob/529c048b7d7d4b04974fa40ce4073693eb8a0916/src/Tests/dotnet.Tests/TelemetryFilterTest.cs#L43

joeloff (approved):

PR #21659, opened 2021-10-01
Hide RazorGeneral rule file properties

### LINKED ISSUE #31150: Purpose of RazorGeneral.xaml rule file
The .NET Project System team is producing a new Project Properties UI. It generates its property model from XAML rule files.

For ASP.NET Web Applications, a "Razor Properties" rule file is injecting properties at the top of this UI:

image

The values are coming from https://github.com/dotnet/sdk/blob/main/src/RazorSdk/Targets/Rules/RazorGeneral.xaml

Are these properties useful to the user?

If these properties are not useful to the user, let's work to remove this rule file from the Project Properties UI altogether.

If they are useful, let's tweak the experience:

- Order the page lower down in the UI
- Add descriptions to the properties
- Add `HelpUrl` values is appropriate, so that the user can easily navigate to more information about these properties
- Potentially change them from `StringProperty` to `EnumProperty` so that a drop-down list of valid values is provided, rather than an open text box.

cc: @captainsafia

drewnoakes:
Fixes dotnet/aspnetcore#31150

Follows on from #16623

These properties are only intended for programmatic access, and not for display to the user.

The current configuration of these rules means they appear in the updated Project Properties UI for ASP.NET projects.

Making these properties non-visible fixes the issue.

### DIFF
--- a/src/RazorSdk/Targets/Rules/RazorGeneral.xaml
+++ b/src/RazorSdk/Targets/Rules/RazorGeneral.xaml
@@ -23,7 +23,7 @@
     DisplayName="Razor Language Version"
     Name="RazorLangVersion"
     ReadOnly="True"
-    Visible="True" />
+    Visible="False" />
 
   <StringProperty
     Category="General"
@@ -31,6 +31,6 @@
     DisplayName="Razor Configuration Name"
     Name="RazorDefaultConfiguration"
     ReadOnly="True"
-    Visible="True" />
+    Visible="False" />
 
 </Rule>


captainsafia (approved):
Looks good for me.

cc: @pranavkm @NTaylorMullen There's no value in having these options visible to the user, AFAIK. Hiding them to avoid an issue where they show up untranslated in certain VS UIs.

pranavkm (approved):

pranavkm:
@drewnoakes is the plan to get this in for Dev17?

NTaylorMullen (approved):

drewnoakes:
> @drewnoakes is the plan to get this in for Dev17? 

I'm hoping to get https://github.com/dotnet/project-system/pull/7654 in, which will mean the timing of this SDK change is less important.

drewnoakes:
> I'm hoping to get dotnet/project-system#7654 in, which will mean the timing of this SDK change is less important.

We got that PR in for 17.0.

PR #21748, opened 2021-10-05
Use the portable rid for --use-current-runtime.

### LINKED ISSUE #14296: --use-current-runtime semantics
https://github.com/dotnet/sdk/pull/14093 adds a new option that makes the sdk use an appropriate rid for publishing.

There was an on-going discussion whether the appropriate rid is the portable rid, machine rid, or whether both options should be added, see https://github.com/dotnet/sdk/pull/14093#issuecomment-708440603.

cc @eerhardt @tmds, @dagood, @am11, @dsplaisted, @janvorli, @marcpopMSFT

tmds:
Fixes https://github.com/dotnet/sdk/issues/14296.

cc @dsplaisted @am11 @omajid @dseefeld @marcpopMSFT

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -62,7 +62,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == 'true'">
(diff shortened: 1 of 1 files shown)


am11:
Nice! Looks like we don't run tests in `Source-Build (Managed)` leg, otherwise it would require delete a couple of suppression. Perhaps we should run tests in that leg to catch regressions.

Should we target release/6.0.1xx? The flow in sdk repo is opposite to that of runtime, changes from release / preview branches flow back to main (but not the other way around).

dsplaisted (approved):
LGTM.  It's late enough in the release cycle that I would not put this in 6.0.1xx unless you think it's critical to get it in.  I think we can merge it into main now and then backport to 6.0.2xx when we have a branch for that.

PR #21894, opened 2021-10-08
Enable signing of tools host on mac

vitek-karas:
## Description
Mac currently enforces signing requirements on new hardware so when creating the apphost during dotnet tool installation, we need to sign that apphost.  The runtime added that functionality but the sdk was never opted into that behavior.

## Regression?
No

## Risk
Low

## Verification
[X] Manual (required)
[] Automated


mateoatr (approved):

agocke:
I can create the backport

PR #21895, opened 2021-10-08
Fix 1ES image name

lpatalas:
We've missed one image name when upgrading main to 1ES hosted pools.

Tracking issue: https://github.com/dotnet/core-eng/issues/14276


lpatalas:
This was already fixed here: https://github.com/dotnet/sdk/pull/21999.

PR #21919, opened 2021-10-09
Use NuGetPackageVersion in dotnet.*.js file name

TanayParikh:
## Description / Customer Impact

Users running a .NET 6.0 SDK with a .NET 5.0 TFM would get a `dotnet.{VERSION}.js` file annotated with the SDK version rather than the TFM/package versions. However the contents of the file would reflect the TFM/package versions and _NOT_ the SDK. To prevent confusion this'll cause, this PR ensures the VERSION used for the `dotnet.js` file matches the `NuGetPackageVersion` rather than the SDK. 

Ex. Say you run a 6.0 SDK against a 5.0 TFM, you'll see `dotnet.6.0.0-preview-xyz.js` file in the build/publish outputs, which may lead you to believe you're running a 6.0 app. However, it is still a 5.0 app with 5.0 content/packages.

<details>
<summary>Another example</summary>

> If you use the .NET 5 SDK to create a Blazor WebAssembly application, at runtime it will load a JS file called `dotnet.5.0.10.js`.
>
> If you then install the .NET 6 SDK (and assuming there's no `global.json` pinning you to the older SDK), then when you next build and run the application, the same JS file will now be called `dotnet.6.0.0-rtm.21478.11.js` or similar. That is,
>
> * The *name* of the file reflects the SDK you built your app with, regardless of the TFM or package versions you're referencing. So in this case it looks like it's a 6.0.x file.
> * The *contents* of the file match up with the TFM/package versions you're using, i.e., it's actually a 5.0.x file.

</details>

## Regression?
- [ ] Yes
- [x] No

## Risk
- [ ] High
- [ ] Medium
- [x] Low

When `MonoPlatform` actually finds & utilizes this file, it [globs based on the prefix `dotnet.` and suffix `.js`](https://github.com/dotnet/aspnetcore/blob/d02192f634038ff944d461dbddff28db80683ad4/src/Components/Web.JS/src/Platform/Mono/MonoPlatform.ts#L176-L182), hence the version / hash do not impact functionality.

## Verification
- [x] Manual (required)
- [ ] Automated

<details>
<summary>Manual Validation Details</summary>

### net5.0 TFM with 6.0 SDK

Build:
image

Publish:
image

---

### net6.0 TFM with 6.0 SDK

Build:
image

Publish:
image

</details>

- [ ] Yes
- [ ] No
- [x] N/A

Fixes: https://github.com/dotnet/aspnetcore/issues/37256

### DIFF
--- a/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.5_0.targets
+++ b/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.5_0.targets
@@ -90,7 +90,6 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <PropertyGroup>
       <_DotNetJsVersion>%(_DotNetJsItem.NuGetPackageVersion)</_DotNetJsVersion>
-      <_DotNetJsVersion Condition="'$(_DotNetJsVersion)' == ''">$(TargetFramework)</_DotNetJsVersion>
       <_BlazorDotnetJsFileName>dotnet.$(_DotNetJsVersion).js</_BlazorDotnetJsFileName>
       <_BlazorDotNetJsFilePath>$(IntermediateOutputPath)$(_BlazorDotnetJsFileName)</_BlazorDotNetJsFilePath>
     </PropertyGroup>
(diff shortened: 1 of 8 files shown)


javiercn (dismissed):
Looks promising however, I think the change scope is larger than necessary.

Static web assets don't have the concept of "NugetPackageVersion" and we shouldn't add it unless we have a good reason for it.

Unless I'm missing something, this change should be simplified to grabbing the NugetPackageVersion from the _dotNetJsItem and using that to define the relative path for dotnet.js.

pranavkm (approved):

PR #21954, opened 2021-10-11
Prevent browser refresh script from injecting into iframes

pranavkm:
The aspnet-browser-refresh.js script is injected in to documents that appear to destined to be displayed in the browser. As part of https://github.com/dotnet/aspnetcore/issues/37326, we realized the script was being added to an iframe that appears in the background. (We only found this because the particular page being served uses CSP that is designed to prevent modifications). Even if amending the script had succeeded in this case, VS which uses this code currently does not support having multiple instances of the script calling back.

The fix is to avoid injecting the scripts in responses that are clearly not destined for a visible browser document.

Partly fixes https://github.com/dotnet/aspnetcore/issues/37326

### Customer impact

Warnings in the browser console when running F5 / Ctrl-F5 on a React or Angular project template

### Testing
* [x] Added unit tests
* [x] Verified the E2E manually

### Risk
Low


SteveSandersonMS (approved):
Looks great!

Do you know if there's a specific plan to fix this for browser link too? That's equally problematic.

Issue #11008, opened 2020-03-26
Re-enable C++/CLI tests


wli3:
C++ does not support package restore. So, it is cannot download apphost pack from the internet. Since I made a change recently to use stage 2 bundled version props file instead of stage 0 (to avoid needing to update stage 0 when runtime insert), apphost will not be in the "dotnet" folder. The solution could be, in the test set up, download the missing apphost pack. But that will duplicate a lot of dotnet/installer code. Or move these tests to dotnet/installer.

wli3:
Or overriding the implicit versions during the test themselves

Nirmal4G:
Why not use `PackageDownload`?

wli3:
C++/CLI project system does not support PackageDownload. And PackageDownload uses nuget

marcpopMSFT:
Fixed using the new package reference support!

MiYanni:
There are 3 tests, [It_can_generate_correct_depsJson_file](https://github.com/dotnet/sdk/blob/941ec62ab28b01b4efcab6bcdaec37b2d319b965/test/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProjectWithTransitiveDeps.cs#L15-L16), [It_can_generate_all_runtimeconfig_files_to_output_folder](https://github.com/dotnet/sdk/blob/941ec62ab28b01b4efcab6bcdaec37b2d319b965/test/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProjectWithTransitiveDeps.cs#L26-L27), [It_can_generate_all_depsjson_files_to_output_folder](https://github.com/dotnet/sdk/blob/941ec62ab28b01b4efcab6bcdaec37b2d319b965/test/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProjectWithTransitiveDeps.cs#L38-L39), that were all skipped for the [predecessor](https://github.com/dotnet/sdk/issues/3785) to this issue. When running those tests, they all fail for the same reason:
```
[...]

PR #21971, opened 2021-10-12

v-wuzhai:


v-codyguan (changes requested):
@dsplaisted We can keep this PR until [11008](https://github.com/dotnet/sdk/issues/11008) is resolved. Does that make sense？

PR #21999, opened 2021-10-13
Fix demand for internal build on Windows

nohwnd:
Add demand for internal build to match what is being used on 6.0.1xx-rc2, because otherwise Ubuntu machine is selected to do the build.


nohwnd:
@sfoslund / @jonfortescue The internal build on main is broken since 8a1289b72e7b58bad0e555f03d4a175ae83f1ae8 because it uses Ubuntu machine to run the windows build. Since then there are no insertions into dotnet/installer. I compared with rc2 and this seems to be the culprit. The public build still works because it is using the correct demand. Please merge this, we have some insertions into sdk that we would like to see in the installer. :)

sfoslund (approved):
Thanks!

jonfortescue (approved):

PR #22045, opened 2021-10-14
[Razor] Fix up-to-date check on static web assets

javiercn:
When the up to date check is triggered for a project, we emit a file to disk with the list of assets from that project that other projects would reference.

On a referencing project we collect the list of files from referenced projects as up to date check inputs and use our static web assets build manifest as the up to date check built.

When the user makes an edit in a given project (like adding, renaming, or removing a file) the up to date check in that project is triggered. As a result, the marker file would change at that point.

When a build is triggered for the server project, one of the marker files will have changed, and that will cause the server project to also rebuild.

Here is the example of it working:
image

* We rebuild
* We build again (For some reason it thinks an assembly has changed and rebuilds two projects)
* After that second build, the up to date check returns everything is up to date
* We then make a change on a file in the Client project, which updates the marker
* After that, we trigger a build on the server project. It builds both the client and the server projects.
* We do another build and at this time it only builds the client project (the server is now up to date)
* We change a c# file on the client project and build the server project (only the client builds
* We trigger another build on the server project (now everything is up to date)

Fixes https://github.com/dotnet/aspnetcore/issues/37508

## Customer Impact

When a customer adds a static file to a referenced project, visual studio still believes the referencing project is up to date and won't update the static web assets manifest.

## Regression?

- [X] Yes
- [ ] No

This change would be detected in previous versions.

## Risk

- [ ] High
- [ ] Medium
- [X] Low

The introduced changes only affect the computation of the up-to-date check by visual studio. This change does not impact any of the core functionality of the application.

## Verification

- [X] Manual (required)
- [ ] Automated

The testing is completely manual because it can't be automated due to this being a service invoked by the Visual Studio project system.

- [ ] Yes
- [ ] No
- [X] N/A

### DIFF
--- a/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.StaticWebAssets.targets
+++ b/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.StaticWebAssets.targets
@@ -740,7 +740,6 @@ Copyright (c) .NET Foundation. All rights reserved.
 
       <ItemGroup>
         <FileWrites Include="$(StaticWebAssetsReferenceUpToDateMarkersListPath)" />
-        <FileWrites Include="$(StaticWebAssetsUpToDateMarker)" />
       </ItemGroup>
 
       <MergeConfigurationProperties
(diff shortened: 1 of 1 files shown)


pranavkm (approved):

javiercn:
Cleanup targets file

PR #22213, opened 2021-10-20
Update the Microsoft.CodeAnalysis.CSharp and Microsoft.CodeAnalysis.C…

pranavkm:
…Sharp.Workspaces versions in SDK


sfoslund (approved):
LGTM, you can also add this commit directly to the merge PR if that's easier

PR #22314, opened 2021-10-25
Add implicit RIDs when not specified on self contained option

### LINKED ISSUE #21405: Enable implicit RID with `--sc`
The new `--sc` syntax works for self-contained apps, per https://github.com/dotnet/sdk/pull/19681.

It still requires a RID. Let's enable it to work with the implicit SDK RID so that we don't create a breaking change in .NET 7. This argument is new. It also will enable users to get around the new warning with `-r`.

This change is some really nice fit-and-finish for .NET 6. This behavior is the original intent of https://github.com/dotnet/sdk/issues/19576.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/21405


dsplaisted (approved):
Looks good.

PR #22364, opened 2021-10-27
Fix version parsing that depended on current culture

### LINKED ISSUE #21518: FXVersion.TryParse fails to parse version on Thai culture
## Problem
Steps to reproduce, on UNIX machine:
```
LANG=th_TH.UTF-8  dotnet new
```
this will print:
https://gist.github.com/DavidKarlas/29a0749d371be2858ea629a3c66d26fb

Unfortunately on Windows its not so easy to reproduce, but I have this example:
```
CultureInfo.CurrentCulture = new CultureInfo("th-TH");
Console.WriteLine("55.66.77.88".IndexOf("."));
```
Which prints `0` instead of `2`.
And this is where FXVersion parsing fails: https://github.com/dotnet/sdk/blob/a5bda50/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/FXVersion.cs#L237 because it tries to parse empty string.

I just want to add this is not just `dotnet new` problem, similar issue can be reproduced with `LANG=th_TH.UTF-8  dotnet build` which will print https://gist.github.com/DavidKarlas/a16d14224ad993e71f2b45a0b2103015

And I believe it will also be problem on Windows when user has Thai(or some other similar culture set) as Windows language/culture.
## Solution
Use `.IndexOf('.', , StringComparison.Ordinal)`.

marcpopMSFT:
Fixes #21518

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/FXVersion.cs
+++ b/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/FXVersion.cs
@@ -4,6 +4,7 @@
 #nullable disable
 
 using System.Text;
+using System;
 
 namespace Microsoft.DotNet.MSBuildSdkResolver
 {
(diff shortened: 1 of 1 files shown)


Reapism (approved):
Nice!

KalleOlaviNiemitalo:
> Fix usage of IndexOf that didn't respect current culture

The usage did respect the current culture already and that's what caused the problem. How about: "Fix version parsing that depended on current culture"

PR #22446, opened 2021-11-02
Treat SDK previews as separate workload feature bands

sfoslund:


sfoslund:
@dsplaisted not sure if we've already discussed this in a workload meeting, but do we want to require manifest teams to produce manifests with ids `foo.manifest.6.0.100-preview.1` and install manifests to a 6.0.100-preview.1 directory? Or should we allow preview feature bands to use the manifests without the preview tags? This PR currently has the first implementation but it feels somewhat clunky and would require a change in manifest authoring as well.

dsplaisted:
> @dsplaisted not sure if we've already discussed this in a workload meeting, but do we want to require manifest teams to produce manifests with ids `foo.manifest.6.0.100-preview.1` and install manifests to a 6.0.100-preview.1 directory? Or should we allow preview feature bands to use the manifests without the preview tags? This PR currently has the first implementation but it feels somewhat clunky and would require a change in manifest authoring as well.

Good point, I hadn't considered this, which will make this more complicated than I was thinking.

Manifest teams would need to produce manifests using the preview bands in order for them to work side-by-side.  However, just like we can't expect them to immediately produce an updated manifest when we branch for a new feature band, they also can't produce an updated manifest immediately when we branch for a new preview.  So we would need to update the [manifest fallback logic](https://github.com/dotnet/sdk/blob/7ba8cd3f2807dcc825cbd2495db58fc9960818d9/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs#L159-L177) to work across previews.  To make it more complicated, manifest teams should also be able to produce manifests that are installed in the non-preview feature band folder.  IE near the release they should be able to switch to using 6.0.200 for the feature band instead of 6.0.200-rc.2 or whatever.  So in addition to falling back to a previous prerelease, it should also be able to use the non-prerelease version, even though semantically that would be greater than the current prerelease version.

I didn't review the PR in detail yet but one thing I'd like to see is unit tests for the way `SdkFeatureBand` converts from the version to the feature band.  That would give examples of how the mapping works.

sfoslund:
@dsplaisted I've updated this PR as you suggested, mind reviewing?

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -293,6 +279,82 @@ var sdkDirectoryWorkloadManifestProvider
...
+            CreateMockManifest(fakeDotnetRootDirectory, "5.0.100", "Android");
+
+            // Write 6.0.100-preview.2 manifests-> ios only
+            CreateMockManifest(fakeDotnetRootDirectory, "6.0.100-preview.2", "iOS");
+
+            // Write 7.0.100 manifests-> ios and android
+            CreateMockManifest(fakeDotnetRootDirectory, "7.0.100", "Android");
+            CreateMockManifest(fakeDotnetRootDirectory, "7.0.100", "iOS");
+
+            var knownWorkloadsFilePath = Path.Combine(fakeDotnetRootDirectory, "sdk", "6.0.100-preview.4.12345", "IncludedWorkloadManifests.txt");
```
dsplaisted, line 301: Nit: Avoid repeating `"6.0.100-preview.4.12345"` twice, break it out into a variable or constant.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -293,6 +279,82 @@ var sdkDirectoryWorkloadManifestProvider
...
+            Directory.CreateDirectory(Path.GetDirectoryName(knownWorkloadsFilePath)!);
+            File.WriteAllText(knownWorkloadsFilePath, "Android\niOS");
+
+            var sdkDirectoryWorkloadManifestProvider
+                = new SdkDirectoryWorkloadManifestProvider(sdkRootPath: fakeDotnetRootDirectory, sdkVersion: "6.0.100-preview.4.12345", userProfileDir: null);
+
+            GetManifestContents(sdkDirectoryWorkloadManifestProvider)
+                .Should()
+                .BeEquivalentTo("4.0.100", "5.0.100");
```
dsplaisted, line 310: How about including both the manifest ID and version band in the mock manifest contents so you can verify which manifest came from which version?  IE:

```suggestion
                .BeEquivalentTo("4.0.100/iOS", "5.0.100/Android");
```

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -293,6 +279,82 @@ var sdkDirectoryWorkloadManifestProvider
                 .BeEquivalentTo("6.0.100", "5.0.100");
         }
 
+        [Fact]
+        public void ItShouldFallbackWhenPreviewFeatureBandHasNoManifests()
```
dsplaisted, line 283: I'd suggest also having a test for rolling back between preview bands, IE where the SDK is `6.0.100-preview.7` but it can pick up a workload manifest from the `6.0.100-preview.6` band.

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -309,6 +309,52 @@ public void GivenWorkloadInstallItDoesNotRemoveOldInstallsOnRollback(bool userLo
...
+            Directory.GetFiles(rc1InstallRecordPath).Count().Should().Be(1);
+            File.Exists(Path.Combine(rc1InstallRecordPath, existingWorkload))
+                .Should().BeTrue();
+
+            var packRecordDirs = Directory.GetDirectories(Path.Combine(dotnetRoot, "metadata", "workloads", "InstalledPacks", "v1"));
+            packRecordDirs.Count().Should().Be(3);
+            var featureBandRecords = Directory.GetFiles(Directory.GetDirectories(packRecordDirs[0])[0]);
+            featureBandRecords.Count().Should().Be(2);
+            var installPacks = Directory.GetDirectories(Path.Combine(dotnetRoot, "packs"));
+            installPacks.Count().Should().Be(2);
```
dsplaisted, line 355: What are the different packs and directories that are being checked here?  Some comments or possibly some more assertions about the actual values would be nice.

dsplaisted (approved):

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkFeatureBand.cs
```diff
@@ -0,0 +1,68 @@
...
+            var fullVersion = version ?? throw new ArgumentNullException(nameof(version));
+            if (string.IsNullOrEmpty(version.Prerelease) || version.Prerelease.Contains("dev") || version.Prerelease.Contains("ci"))
+            {
+                _featureBand = new ReleaseVersion(fullVersion.Major, fullVersion.Minor, fullVersion.SdkFeatureBand);
+            }
+            else
+            {
+                // Treat preview versions as their own feature bands
+                var prereleaseComponents = fullVersion.Prerelease.Split('.');
+                var formattedPrerelease = prereleaseComponents.Length > 1 ?
```
joeloff, line 27: This assumes we'll never have more than 2 prerelease labels, e.g. -rc.2 or -preview.9

dsplaisted: @joeloff I'm not following.  The elements after the first two are the build numbers which we want to ignore for the purposes of feature bands.  We aren't expecting to have a Preview 3 and then a Preview 3.1, for example.

joeloff: Yup, that's fine, I just wanted to confirm the assumptions we're making because technically semver supports an infinite number of prerelease labels.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
sfoslund: @joeloff I'm noticing that ReleaseVersions treat different previews as equal (for example, 6.0.100-preview.4 == 6.0.100-preview.5). Is that expected? Do we have any libraries that know how to determine the precedence of all prereleases (ie rc > preview > alpha) or is that something I'll need to implement by hand here?

dsplaisted: `NuGetVersion` does a full semantic version comparison, I believe.

joeloff: The ReleaseVersion should comply with SemVer 2.0.

sfoslund: Looks like there's a bug in ReleaseVersion, @joeloff says he has a PR out to fix but for now I'm using NugetVersions for cross preview comparisons.

dsplaisted: Can one of you link the ReleaseVersion fix PR?

sfoslund: Sorry meant to say he's putting a PR out, there's not a link yet.

joeloff: https://github.com/dotnet/deployment-tools/pull/158

sfoslund:
@joeloff your fix to ReleaseVersion fixed things up without having to use NuGetVersion, thanks! I'll merge this once it goes green unless anyone has feedback.

joeloff:
> @joeloff your fix to ReleaseVersion fixed things up without having to use NuGetVersion, thanks! I'll merge this once it goes green unless anyone has feedback.

Sounds good. We have an internal build that's all signed for 1.0.0-preview2.6.21561.1  I'll see if we can get it published to NuGet as well.

PR #22477, opened 2021-11-04
Blazor AOT Publish Tests

### LINKED ISSUE #36481: [Blazor] Add AoT pipeline on the SDK repo to validate AoT scenarios with automation
Currently AoT validation is manual or based on CTI coverage. This makes it hard to update the build scenarios without risking a break and makes fixing issues in this area expensive.

The main issue with AoT builds is the time they take. To improve the situation we want to run a separate build pipeline to make sure it can run in parallel and not add time to the final build.

We will own this pipeline and be responsible for its health, being optional in the beginning.

TanayParikh:
Fixes: https://github.com/dotnet/aspnetcore/issues/36481


eng/build.yml
```diff
@@ -135,6 +137,12 @@ jobs:
         displayName: Build
         env:
           BuildConfig: $(_BuildConfig)
+
+    - ${{ if eq(parameters.agentOs, 'Darwin') }}:
```
TanayParikh, line 141: Ubuntu AOT testing isn't currently possible due to an out of date python in the CI image which doesn't support f-string. Consequently the `emcc.py` script is failing during AOT. This bit should be reverted when CI Ubuntu image is upgraded to 20.04 LTS.

eng/build.yml
```diff
@@ -187,3 +187,107 @@ jobs:
         publishLocation: Container	
       continueOnError: true	
       condition: always()
+
+- ${{ if in(parameters.agentOs, 'Windows_NT', 'Darwin') }}:
```
TanayParikh, line 191: We have to run the AOT tests in a separate `job` (however they can still be in the same pipeline), due to https://github.com/dotnet/sdk/issues/22655. We may be able to consolidate the AOT tests into the existing test job once that issue is resolved.

PR #22532, opened 2021-11-10
Update Test project x64SolutionBuild to current framework

v-wuzhai:


joeloff:
The change appears to be fine. @dsplaisted Given that the generated output has changed between 1.1 and the latest TFM, are we at risk of losing coverage around back compat if we convert everything to latest?

dsplaisted (approved):
We may lose some coverage, but we may have other coverage that explicitly tests these differences.  IMO it's better to keep most of our test coverage mostly current rather than holding it back for the low risk of breaking something for targets that aren't even supported anymore and whose behavior is likely covered under other tests.

PR #22557, opened 2021-11-11
Tell user when workload has already been installed

### LINKED ISSUE #20085: Consecutive installs should report installed workload
Install the same workload twice,

1. ```dotnet workload install wasm-tools```
2. ```dotnet workload install wasm-tools```

After the second command it tells you that the workload was installed. The typical behavior for installers would be to prompt for repair/uninstall, or we could simply detect the workload record and inform the user that it's already installed and no-op.

If files were simply copied again, then the behavior sounds similar to a repair. For MSI based installs there would be no effect, but the message is still confusing because it made no changes to the system.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/20085

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
@@ -169,7 +169,7 @@
     <value>Download of workload packs to offline cache failed: {0}</value>
   </data>
   <data name="WorkloadAlreadyInstalled" xml:space="preserve">
-    <value>Workload '{0}' is already installed.</value>
+    <value>Workload(s) '{0}' are already installed. Run `dotnet workload update` to update your workloads.</value>
   </data>
   <data name="InvalidNuGetVersionRange" xml:space="preserve">
     <value>Specified version '{0}' isn't a valid NuGet version range.</value>
--- a/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
@@ -169,7 +169,7 @@
     <value>Download of workload packs to offline cache failed: {0}</value>
   </data>
   <data name="WorkloadAlreadyInstalled" xml:space="preserve">
-    <value>Workload(s) '{0}' are already installed. Run `dotnet workload update` to update your workloads.</value>
+    <value>Workload(s) '{0}' are already installed.</value>
   </data>
   <data name="InvalidNuGetVersionRange" xml:space="preserve">
     <value>Specified version '{0}' isn't a valid NuGet version range.</value>
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
@@ -175,7 +175,12 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             {
                 // Update currently installed workloads
                 var installedWorkloads = _workloadInstaller.GetWorkloadInstallationRecordRepository().GetInstalledWorkloads(_sdkFeatureBand);
-                workloadIds = workloadIds.Concat(installedWorkloads).Distinct();
+                var previouslyInstalledWorkloads = installedWorkloads.Intersect(workloadIds);
+                if (previouslyInstalledWorkloads.Any())
+                {
+                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadAlreadyInstalled, string.Join(" ", previouslyInstalledWorkloads)));
+                }
+                workloadIds = workloadIds.Concat(installedWorkloads);
 
                 _workloadManifestUpdater.UpdateAdvertisingManifestsAsync(includePreviews, offlineCache).Wait();
                 manifestsToUpdate = string.IsNullOrWhiteSpace(_fromRollbackDefinition) ?
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
@@ -178,7 +178,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                 var previouslyInstalledWorkloads = installedWorkloads.Intersect(workloadIds);
                 if (previouslyInstalledWorkloads.Any())
                 {
-                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadAlreadyInstalled, string.Join(" ", previouslyInstalledWorkloads)));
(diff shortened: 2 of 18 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -175,7 +175,12 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
             {
                 // Update currently installed workloads
                 var installedWorkloads = _workloadInstaller.GetWorkloadInstallationRecordRepository().GetInstalledWorkloads(_sdkFeatureBand);
-                workloadIds = workloadIds.Concat(installedWorkloads).Distinct();
+                var previouslyInstalledWorkloads = installedWorkloads.Intersect(workloadIds);
+                if (previouslyInstalledWorkloads.Any())
+                {
+                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadAlreadyInstalled, string.Join(" ", previouslyInstalledWorkloads)));
```
joeloff, line 181: should we join the string on ", " instead so it reads easier?

dsplaisted:
I'm not sure if we should bail out entirely if any of the specified workloads have already been installed.  There might be scripts or people might be following tutorial steps that want to ensure that workloads are installed.  If we make this change there's no easy command to say "install these workloads if they are not already installed".

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
sfoslund: I think it's standard to use spaces since commas don't always translate well when these strings are localized

sfoslund:
@dsplaisted makes sense, I've converted it to a non-blocking warning instead of an error.

sfoslund:
@dsplaisted can I get a quick review here? It includes a commit to fix up 6.0.2xx builds

src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
```diff
@@ -169,7 +169,7 @@
     <value>Download of workload packs to offline cache failed: {0}</value>
   </data>
   <data name="WorkloadAlreadyInstalled" xml:space="preserve">
-    <value>Workload '{0}' is already installed.</value>
+    <value>Workload(s) '{0}' are already installed. Run `dotnet workload update` to update your workloads.</value>
```
dsplaisted, line 172: By default if you run `dotnet workload install workload1 workload2`, it will already update the workloads.  You would have to specify `--skip-manifest-update` to avoid the update.  So does it make sense to include the second sentence in this message here?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -175,7 +175,12 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
             {
                 // Update currently installed workloads
                 var installedWorkloads = _workloadInstaller.GetWorkloadInstallationRecordRepository().GetInstalledWorkloads(_sdkFeatureBand);
-                workloadIds = workloadIds.Concat(installedWorkloads).Distinct();
+                var previouslyInstalledWorkloads = installedWorkloads.Intersect(workloadIds);
+                if (previouslyInstalledWorkloads.Any())
+                {
+                    _reporter.WriteLine(string.Format(LocalizableStrings.WorkloadAlreadyInstalled, string.Join(" ", previouslyInstalledWorkloads)).Yellow());
```
dsplaisted, line 181: I think we had some encoding issues with Yellow messages not too long ago, is this doing things correctly?  Was the issue before that we were using `Console` instead of the reporter?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -175,7 +175,12 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
...
+                }
+                workloadIds = workloadIds.Concat(installedWorkloads);
```
dsplaisted, line 183: Does it matter if there are duplicates in the list here (which it looks like will be the case if you specify a workload that is already installed)?  Should we be calling `Distinct()`?
```suggestion
                workloadIds = workloadIds.Concat(installedWorkloads).Distinct();
```

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -368,6 +368,31 @@ public void GivenWorkloadInstallItErrorsOnInvalidWorkloadRollbackFile()
             string.Join(" ", _reporter.Lines).Should().Contain("Invalid rollback definition");
         }
 
+        [Fact]
+        public void GivenWorkloadInstallItErrorsWhenTheWorkloadIsAlreadyInstalled()
```
dsplaisted, line 372: Nit: Update test name, it doesn't error anymore.

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -368,6 +368,31 @@ public void GivenWorkloadInstallItErrorsOnInvalidWorkloadRollbackFile()
...
+
+            // Successfully install a workload
+            var installParseResult = Parser.Instance.Parse(new string[] { "dotnet", "workload", "install", workloadId });
+            var installCommand = new WorkloadInstallCommand(installParseResult, reporter: _reporter, workloadResolver: workloadResolver, nugetPackageDownloader: new MockNuGetPackageDownloader(tmpDir),
+                workloadManifestUpdater: manifestUpdater, userProfileDir: userProfileDir, version: sdkFeatureVersion, dotnetDir: dotnetRoot, tempDirPath: testDirectory);
+            installCommand.Execute();
+            _reporter.Clear();
+
+            // Install again, this time it should tell you that you already have the workload installed
+            installCommand.Execute();
```
dsplaisted, line 392: Should we be checking the return / exit code here (and earlier)?

dsplaisted, line 392: It might be a good idea for the second install command to specify a workload that was already installed, and one that should be installed, and to verify that the second one gets installed.

dsplaisted (approved):

src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
sfoslund: Good point, made more sense when this was an error. I'll remove.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
sfoslund: Yes, I believe that issue was console vs reporter, I'm not aware of any issues with .Yellow()

PR #22573, opened 2021-11-12
Do not advertise workload updates on run output

### LINKED ISSUE #22571: Suppress workload update notification from `dotnet run`
If workload updates are available, we print the message `Workload updates are available. Run `dotnet workload list` for more information.`

We should probably avoid doing this for the `dotnet run` command, as we typically try to avoid outputting anything that doesn't come from the project being run.

sfoslund:
Fixes https://github.com/dotnet/sdk/issues/22571

### DIFF
--- a/src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
+++ b/src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
@@ -305,9 +305,10 @@ namespace Microsoft.DotNet.Cli.Workload.Install.Tests
         }
 
         [Theory]
-        [InlineData("build")]
-        [InlineData("publish")]
-        public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(string commandName)
+        [InlineData("build", false)]
+        [InlineData("publish", false)]
+        [InlineData("run", true)]
+        public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(string commandName, bool commandOptedOut)
         {
             var testInstance = _testAssetsManager.CopyTestAsset("HelloWorld", identifier: commandName)
                 .WithSource()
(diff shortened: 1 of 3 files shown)


sfoslund:
@baronfel tagging you in case you have any other thoughts on when we should output this message. Right now it's outputted any time one of your installed workloads has an update  on any restoring command (with this change, excluding run). @marcpopMSFT suggested possibly only writing the message when the current project uses the workload, but I'm not sure that would be very easy to implement and I think it would have significant perf impacts.

baronfel:
That seems fine to me, @sfoslund. IF we had https://github.com/dotnet/msbuild/issues/3911 or similar we might be able to add information about the workloads to the invocation of the `SeparateRestoreCommand` as sort of an all-in-one approach, but in lieu of that this is a good change.

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -305,9 +305,10 @@ public void GivenWorkloadManifestUpdateItChoosesHighestManifestVersionInCache()
         }
 
         [Theory]
-        [InlineData("build")]
-        [InlineData("publish")]
-        public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(string commandName)
+        [InlineData("build", false)]
+        [InlineData("publish", false)]
+        [InlineData("run", true)]
+        public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(string commandName, bool commandOptedOut)
```
dsplaisted, line 311: `commandOptedOut` ends up being sort of a double negative.  I would invert the boolean values and call the variable something like `shouldShowUpdateAvailable`.

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -320,14 +321,27 @@ public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(st
...
+                    .HaveStdOutContaining(Workloads.Workload.Install.LocalizableStrings.WorkloadUpdatesAvailable);
+            }
+            else
+            {
+                commandResult
+                    .Should()
+                    .Pass()
+                    .And
+                    .NotHaveStdOutContaining(Workloads.Workload.Install.LocalizableStrings.WorkloadUpdatesAvailable);
+            }
```
dsplaisted, line 343: I think you can do the following if you want to deduplicate the code that checks that the command passed.

```suggestion
            commandResult
                .Should()
                .Pass()

            if (!commandOptedOut)
            {
                commandResult
                    .Should()
                    .HaveStdOutContaining(Workloads.Workload.Install.LocalizableStrings.WorkloadUpdatesAvailable);
            }
            else
            {
                commandResult
                    .Should()
                    .NotHaveStdOutContaining(Workloads.Workload.Install.LocalizableStrings.WorkloadUpdatesAvailable);
            }
```

dsplaisted (approved):
Looks good, though it looks like there may be some test failures.

sfoslund:
> Looks good, though it looks like there may be some test failures.

This PR needs eab7a38 too, CI should pass once that gets merged along with https://github.com/dotnet/sdk/pull/22557

PR #22576, opened 2021-11-13
GenerateRuntimeConfig now uses stream serialization instead of loading file in memory

Therzok:


src/Tasks/Microsoft.NET.Build.Tasks/GenerateRuntimeConfigurationFiles.cs
```diff
@@ -91,7 +91,7 @@ protected override void ExecuteCore()
 
             if (!string.IsNullOrEmpty(RollForward))
             {
-                if (!RollForwardValues.Any(v => string.Equals(RollForward, v, StringComparison.OrdinalIgnoreCase)))
+                if (!RollForwardValues.Contains(RollForward, StringComparer.OrdinalIgnoreCase))
```
marcpopMSFT, line 94: Why did you decide to change from any to contains here?

joeloff: They're both similar, but for large collections the underlying structure can impact performance I think. `Any` has to enumerate all the elements, but `Contains` can be faster, e.g. a list vs. hash set. I'm good with the switching to the `Contains ` here.

Therzok: I'm trying to remember this since PR is old. `RollForwardValues` is actually just an array `string[]`, so it's still resolving to a LINQ method. 

`Contains`/`Any` have no functional perf differences, so I think I just [wanted to avoid paying unconditionally for the lambda closure allocations](https://github.com/dotnet/roslyn/issues/20777). Obviously it won't be a huge benefit in the big picture, but it was easy to cleanup.

Therzok:
@joeloff @marcpopMSFT PTAL

marcpopMSFT:
Triggering a new PR build as one hasn't run in a long time and it'd be good to get more recent validation

Therzok:
Build looks green :D

marcpopMSFT (approved):

marcpopMSFT:
Thank you for your contribution.

PR #22636, opened 2021-11-16
Change messages for ILLink

### LINKED ISSUE #22518: Publishing trimmed console app prints a message which warns about trimming
Running publish with trimming on a empty hello world in .NET 6 produces output like this:
```console
  Restored ...\SampleTrim.csproj (in 16.91 sec).
  SampleTrim -> F:\AppModel\repro\SampleTrim\bin\Debug\net6.0\win-x64\SampleTrim.dll
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  SampleTrim -> ...\bin\Debug\net6.0\win-x64\publish\
```

Since this build mode has trim warnings enabled, there should be very little risk trimming the application. I think we should not print out this "warning" if trim warnings are enabled.

We may consider keeping a simple "Optimizing assemblies for size" message as trimming can be relatively slow. In which case it should be printed out before trimming starts (right now it's printed after).

tlakollo:
- Now "Running ILLink, this process might take a while" is printed every time we run the linker, but is not printed on incremental builds since we don't run the linker
- "Assemblies were optimized for size. See: https://aka.ms/dotnet-illink" is printed at the end, if the linker process passed or didn't run (incremental build) to let know the users that they have trimming enabled

Fixes #22518


src/Tasks/Common/Resources/Strings.resx
```diff
@@ -506,8 +506,11 @@ The following are names of parameters or literal values and should not be transl
     <value>NETSDK1100: Windows is required to build Windows desktop applications.</value>
     <comment>{StrBegin="NETSDK1100: "}</comment>
   </data>
-  <data name="ILLink_Info" xml:space="preserve">
-    <value>Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink</value>
+  <data name="ILLinkTimeAwareness" xml:space="preserve">
+    <value>Running ILLink, this process might take a while</value>
```
sbomer, line 510: I'd avoid mentioning ILLink here - maybe "Optimizing assemblies for size. This might take a while."?

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -506,8 +506,11 @@ The following are names of parameters or literal values and should not be transl
     <value>NETSDK1100: Windows is required to build Windows desktop applications.</value>
     <comment>{StrBegin="NETSDK1100: "}</comment>
   </data>
-  <data name="ILLink_Info" xml:space="preserve">
-    <value>Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink</value>
+  <data name="ILLinkTimeAwareness" xml:space="preserve">
```
sbomer, line 509: Nit: maybe call this ILLinkRunning

sbomer (commented):
I'm personally fine with this behavior - but I wonder if in https://github.com/dotnet/sdk/issues/22518 the intention was to continue producing a "warning" (This may change the behavior...) if trim analysis warnings are suppressed? I'll let @vitek-karas comment on that.

src/Tasks/Common/Resources/xlf/Strings.cs.xlf
```diff
@@ -449,9 +449,14 @@
         <target state="translated">NETSDK1102: Pro vybranou konfiguraci publikování se optimalizace velikosti sestavení nepodporuje. Ujistěte se, že publikujete samostatnou aplikaci.</target>
         <note>{StrBegin="NETSDK1102: "}</note>
       </trans-unit>
-      <trans-unit id="ILLink_Info">
-        <source>Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink</source>
-        <target state="translated">Optimalizuje se velikost sestavení, což může změnit chování aplikace. Nezapomeňte po vydání provést testy. Další informace: https://aka.ms/dotnet-illink</target>
+      <trans-unit id="ILLinkOptimizedAssemblies">
+        <source>Assemblies were optimized for size. See: https://aka.ms/dotnet-illink</source>
```
vitek-karas, line 453: Unrelated: The Czech translation is ... oh god... I always wondered why they have to translate terms like "Assembly" (which in this case is translated verbatim... like if you're assembling a LEGO set).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
```diff
@@ -104,8 +104,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                       '$(_TargetFrameworkVersionWithoutV)' >= '3.0' And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' "
           DependsOnTargets="_RunILLink">
 
-    <NETSdkInformation ResourceName="ILLink_Info" />
-
+    <NETSdkInformation Condition="'$(_ILLinkExitCode)' == '' Or '$(_ILLinkExitCode)' == '0'" ResourceName="ILLinkOptimizedAssemblies" />
```
vitek-karas, line 107: Personally I'm not a fan of writing this message. Honestly we shouldn't have any message really (Even before ILLink), but it's so slow that we kind of have to. But I don't see any reason to put one after it's done. There's little value to it. I would only keep the one before.

That also means moving the http link to the "start" message.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -506,8 +506,11 @@ The following are names of parameters or literal values and should not be transl
     <value>NETSDK1100: Windows is required to build Windows desktop applications.</value>
     <comment>{StrBegin="NETSDK1100: "}</comment>
   </data>
-  <data name="ILLink_Info" xml:space="preserve">
-    <value>Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink</value>
+  <data name="ILLinkRunning" xml:space="preserve">
+    <value>Optimizing assemblies for size. This process might take a while</value>
```
vitek-karas, line 510: I think we should print out something like the original message if trim warnings are suppressed. Reasoning:
* There's no message telling people that trimming is potentially problematic with warnings off.
* It's the default setting for Blazor/Xamarin and probably MAUI as well

I know that info message is not very strong, but I think we should not forgo a chance to "warn" people about potential issues.

vitek-karas (commented):
As for the "warn" term - sorry for using it in the issue. I don't think we should emit a real warning, just info message as we're already doing.

The purpose of the message is two-fold
  * Provide some feedback to the user about a long running operation
  * In some cases raise awareness that trimming may cause issues

So I think we should:
  * Move the message before trimming
  * If trim warnings are on (default for console), don't mention potential problems, that's already covered by the warnings - just provide feedback about long running operation. I would not include the URL in this case either
  * If trim warnings are off (Blazor/Xamarin, ...), basically keep the existing message, with the URL which tries to make people aware of potential problems.

sbomer:
As written now, the "before" message is shown only on incremental builds (we shouldn't say "this could take a while..." on incremental builds). @vitek-karas I think the two-fold purpose of the message you suggested makes sense, but I think they belong in separate messages - the one that raises awareness that trimming may cause issues should be shown on every build if warnings are suppressed (including incremental builds that don't actually run the linker). Thoughts?

vitek-karas:
By "incremental build" you mean the case where we publish the app but there were no changes which would cause linker to actually run, right?

If that is the case I agree:
* I do want to show a message about the long running operation - that should absolutely be showed only when we're about to run the linker.
* The "Warning" should be shown every time we know the app is trimmed (and warnings are suppressed). Regardless if linker actually ran or not (that's an optimization and should have no effect on the meaning of the message).

tlakollo:
With the latest commit we now:
- Print `Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink` before running the trimming operation, only in the case SuppressTrimAnalysisWarnings is set to true
- Print `Running Optimization. This process might take a while` if the ILLink process will actually be executed, but won't print anything if it's an incremental build (ILLink process is not executed).

I'm not fully sure about the case in which we have an incremental build (and ILLink will not be executed), and SuppressTrimAnalysisWarnings is false. That specific case will not print anything

vitek-karas:
I think we should change the second message `Running Optimizations` - if this applied to R2R and other similar things, then yes, but since it currently only applies to trimming I think it should read `Optimizing assemblies for size. This process might take a while.`.

If we used that wording then we should probably change the first message as well, maybe something like:
`Optimizing assemblies for size may change the behavior of the app` (very small change).

If trim warnings are on (so SuppressTrimAnalysisWarnings == false) and there's no linker running, we should not show anything (no need).

vitek-karas (approved):
Looks good to me - but I would like either @sbomer or @agocke to review the messages (as my English is not ideal)

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
```diff
@@ -206,6 +204,10 @@ Copyright (c) .NET Foundation. All rights reserved.
    <Target Name="PrepareForILLink"
            DependsOnTargets="_ComputeManagedAssemblyToLink">
 
+    <!-- We print a message to the user to explain that trimming is 
+         potentially problematic when warnings are suppressed. -->
+    <NETSdkInformation Condition="'$(PublishTrimmed)' == 'true' And '$(SuppressTrimAnalysisWarnings)' == 'true'" ResourceName="ILLinkOptimizedAssemblies" />
```
sbomer, line 209: I think you can remove the PublishTrimmed condition since this target only kicks in when trimming.

sbomer (commented):
The messages look good to me. It would be good to add a test for the new behavior.

tlakollo:
I was trying to investigate the failures in this PR, it fails reliably on amd64 machines on the ILLink_respects_feature_settings_from_host_config test. Having said that, I don't think my changes are related to the failure

vitek-karas:
As far as I can tell it's failing due to nuget restore behavior. I'm trying to fix it with this: https://github.com/dotnet/sdk/pull/22750.

PR #22648, opened 2021-11-17
Update Test project NuGetConfigDependentProject to current framework

v-wuzhai:


v-wuzhai:
@dsplaisted We expect that the [test](https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Tests/dotnet-store.Tests/GivenDotnetStoresAndPublishesProjects.cs#L74) contains an error message, but it passed without error.

dsplaisted:
> @dsplaisted We expect that the [test](https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/Tests/dotnet-store.Tests/GivenDotnetStoresAndPublishesProjects.cs#L74) contains an error message, but it passed without error.

The `dotnet store` tests are tricky and probably not worth updating to target new frameworks unless we have a specific reason to do so.

PR #22658, opened 2021-11-17
Exclude runtimeidentifiers for runtimepack

### LINKED ISSUE #19891: `dotnet new android` attempts to restore Microsoft.AspNetCore.App.Runtime.linux-arm64
If you install .NET 6 RC1 and the Android workload, we are now using `**FromWorkload**` so a `NuGet.config` should no longer be needed when building Android projects.

Unfortunately, you get a different error now when testing this:
```
error NU1102: Unable to find package Microsoft.AspNetCore.App.Runtime.linux-arm64 with version (= 6.0.0-rc.1.21417.2)
```

Build log: [msbuild.zip](https://github.com/dotnet/sdk/files/7016700/msbuild.zip)

I found I could workaround the problem with:
```xml
<Target Name="_RemoveLinuxFrameworkReferences"
    AfterTargets="ProcessFrameworkReferences">
  <ItemGroup>
    <_ProblematicRIDs Include="linux-arm;linux-arm64;linux-x86;linux-x64" />
    <PackageDownload Remove="Microsoft.AspNetCore.App.Runtime.%(_ProblematicRIDs.Identity)" />
    <PackageDownload Remove="Microsoft.NETCore.App.Host.%(_ProblematicRIDs.Identity)" />
  </ItemGroup>
</Target>
```
I will probably put this workaround in the Android workload *for now*.

This is probably related to the RID graph. I tried making an adjustment to the RID graph in the past, but they wouldn't accept this change:

https://github.com/dotnet/runtime/pull/45927

dsplaisted:
Allow RuntimeIdentifiers to be excluded from runtime packs.

Will fix #19891, once we add `RuntimePackExcludedRuntimeIdentifiers="android"` to the ASP.NET KnownFrameworkReference items.

Also fixes an unrelated issue where ResolveAppHosts would generate errors in debug mode because it was generating messages prefixed with NETSDK error codes.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs
@@ -83,19 +83,37 @@ namespace Microsoft.NET.Build.Tasks
         public static string GetBestMatchingRid(RuntimeGraph runtimeGraph, string runtimeIdentifier,
             IEnumerable<string> availableRuntimeIdentifiers, out bool wasInGraph)
         {
+            return GetBestMatchingRidWithExclusion(runtimeGraph, runtimeIdentifier,
+                runtimeIdentifiersToExclude: null,
+                availableRuntimeIdentifiers, out wasInGraph);
+        }
+
+        public static string GetBestMatchingRidWithExclusion(RuntimeGraph runtimeGraph, string runtimeIdentifier,
+            IEnumerable<string> runtimeIdentifiersToExclude,
+            IEnumerable<string> availableRuntimeIdentifiers, out bool wasInGraph)
+        {
             wasInGraph = runtimeGraph.Runtimes.ContainsKey(runtimeIdentifier);
 
+            string bestMatch = null;
+
             HashSet<string> availableRids = new HashSet<string>(availableRuntimeIdentifiers);
+            HashSet<string> excludedRids = runtimeIdentifiersToExclude switch { null => null, _ => new HashSet<string>(runtimeIdentifiersToExclude) };
             foreach (var candidateRuntimeIdentifier in runtimeGraph.ExpandRuntime(runtimeIdentifier))
             {
-                if (availableRids.Contains(candidateRuntimeIdentifier))
+                if (bestMatch == null && availableRids.Contains(candidateRuntimeIdentifier))
                 {
-                    return candidateRuntimeIdentifier;
+                    bestMatch = candidateRuntimeIdentifier;
+                }
+
+                if (excludedRids != null && excludedRids.Contains(candidateRuntimeIdentifier))
+                {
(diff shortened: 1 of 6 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs
```diff
@@ -82,20 +82,38 @@ bool FileMatchesProjectLanguage()
...
+        public static string GetBestMatchingRidWithExclusion(RuntimeGraph runtimeGraph, string runtimeIdentifier,
+            IEnumerable<string> runtimeIdentifiersToExclude,
+            IEnumerable<string> availableRuntimeIdentifiers, out bool wasInGraph)
         {
             wasInGraph = runtimeGraph.Runtimes.ContainsKey(runtimeIdentifier);
 
+            string bestMatch = null;
+
             HashSet<string> availableRids = new HashSet<string>(availableRuntimeIdentifiers);
+            HashSet<string> excludedRids = runtimeIdentifiersToExclude switch { null => null, _ => new HashSet<string>(runtimeIdentifiersToExclude) };
```
jonathanpeppers, line 100: Is it worth using the ctor for `HashSet<T>` where you pass a `StringComparer`? `StringComparer.Ordinal` might help performance a bit? (`Or OrdinalIgnoreCase` if it is correct)

jonathanpeppers (approved):
Otherwise, the test looks like this should fix https://github.com/dotnet/sdk/issues/19891.

src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs
```diff
@@ -82,20 +82,38 @@ bool FileMatchesProjectLanguage()
...
+
+                if (excludedRids != null && excludedRids.Contains(candidateRuntimeIdentifier))
                 {
-                    return candidateRuntimeIdentifier;
+                    //  Don't treat this as a match
+                    return null;
                 }
             }
 
             //  No compatible RID found in availableRuntimeIdentifiers
```
sfoslund, line 115: Looks like this comment is out of date now?

src/Tests/Microsoft.NET.Build.Tests/KnownRuntimePackTests.cs
```diff
@@ -106,6 +106,40 @@ public void RuntimePackWithLabelIsSelected()
 
         }
 
+        [Fact]
+        public void AspNetRuntimePackIsNotRestoredForAndroid()
+        {
+            var testProject = new TestProject()
+            {
+                TargetFrameworks = "net6.0",
```
sfoslund, line 114: nit: don't we have a current framework field now to make it easier to update test TFMs?

sfoslund (approved):

PR #22661, opened 2021-11-18
Update Test project WindowsRuntimeComponent to current framework

v-wuzhai:


dsplaisted (changes requested):
We don't actually want to update this.  The `winmdobj` output type is not supported on .NET 5 and higher, and the test verifies that this generates an error.  So we should leave this project targeting .NET 5.

PR #22666, opened 2021-11-18
Don't split collect command into multiple collect commands

fhnaseer:
Old implementation:
--collect:"Code Coverage;XPlat Code Coverage" ---> --collect:"Code Coverage" --collect:"XPlat Code Coverage"

New implementation:
--collect:"Code Coverage;arg1=val1" --> --collect:"Code Coverage;arg1=val1"

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
@@ -78,11 +78,7 @@ namespace Microsoft.DotNet.Cli
         public static readonly Option<IEnumerable<string>> CollectOption = new ForwardedOption<IEnumerable<string>>("--collect", LocalizableStrings.cmdCollectDescription)
         {
             ArgumentHelpName = LocalizableStrings.cmdCollectFriendlyName
-        }.ForwardAsSingle(o =>
-        {
-            var loggersString = string.Join(";", GetSemiColonEscapedArgs(o));
-            return $"-property:VSTestCollect=\"{loggersString}\"";
-        })
+        }.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", GetSemiColonEscapedArgs(o))}\"")
         .AllowSingleArgPerToken();
 
         public static readonly Option<bool> BlameOption = new ForwardedOption<bool>("--blame", LocalizableStrings.CmdBlameDescription)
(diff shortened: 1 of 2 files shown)


src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs
```diff
@@ -508,6 +502,42 @@ public void ItCreatesCoverageFileInResultsDirectory()
             result.ExitCode.Should().Be(1);
         }
 
+        [PlatformSpecificFact(TestPlatforms.Windows)]
```
jakubch1, line 505: please add test:

dotnet test --collect "Code Coverage;Format=Cobertura" --collect "XPlat Code Coverage"

it should generate 2 cobertura.xml files

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -78,7 +78,11 @@ internal static class TestCommandParser
         public static readonly Option<IEnumerable<string>> CollectOption = new ForwardedOption<IEnumerable<string>>("--collect", LocalizableStrings.cmdCollectDescription)
         {
             ArgumentHelpName = LocalizableStrings.cmdCollectFriendlyName
-        }.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", o)}\"")
+        }.ForwardAsSingle(o =>
+        {
```
jakubch1, line 82: can we just simply change it to:

}.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", GetSemiColonEscapedArgs(o))}\"")
.Allow....

if you want to keep it longer please rename loggersString to collectString

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs
jakubch1: can we run also this test on linux?

fhnaseer: Demo test project doesn't contain coverlet reference so XPlat Code Coverage will not work. Though I added a test with two collect and it will verify that both collectors are passed.
Updated code coverage related tests to run on linux as well.

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
fhnaseer: Updated.

jakubch1 (approved):

Issue #22639, opened 2021-11-16
Inconsistent error code for NETSDK1174


dsplaisted:
We are using the NETSDK1174 error code for two different things.

How this happened:

- In #18301 we added a new warning that comes from the CLI.  We wanted a warning code for it, but all of our other warning codes come from SDK tasks.  So we inserted a placeholder resource in the SDK tasks (`PlaceholderRunCommandProjectAbbreviationDeprecated`) to reserve the warning code (which ended up being NETSDK1174.
- Subsequently, we merged an earlier preview branch into the current 6.0 branch in #18519.  Other new error codes had been added in the earlier branch, so this pushed the `PlaceholderRunCommandProjectAbbreviationDeprecated` error code down to NETSDK1176.  However, we didn't notice this, and didn't update the resource in the CLI code that generates this (or in dotnet-watch, which does the same thing).

So now we use NETSDK1174 for two different errors:

- NETSDK1174: The abbreviation of -p for --project is deprecated. Please use --project.
- NETSDK1174: Compression in a single file bundle is only supported when publishing a self-contained application.

The first instance is [already documented](https://docs.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1174), while it looks like the second one may not be documented.  So probably we should change the error code for the second one (resource name `CompressionInSingleFileRequiresSelfContained`) to use a new error code.

FYI @AntonLapounov @vitek-karas @agocke

marcpopMSFT:
@dsplaisted Yes as 6.0.100 is an LTS so will be in support a long while. Error number confusion can use up a bunch of customer's time.

agocke:
@VSadov Can you bump the error code here for the compression error?

VSadov:
Sure!

VSadov:
I assume this would need to be fixed in 6.0 (and then matched in 7.0). Right?

agocke:
@marcpopMSFT Do you all do merge forward or port back for SDK?

dsplaisted:
> @marcpopMSFT Do you all do merge forward or port back for SDK?

Normally we check into the oldest branch a change is going into, and then merge forward.

marcpopMSFT:
Change is in 6.0.1xx-3xx so closing.

PR #22676, opened 2021-11-18
[release/6.0.1xx] bump the error number for CompressionInSingleFileRequiresSelfContained

VSadov:
Fixes: https://github.com/dotnet/sdk/issues/22639

As a result of a merge `CompressionInSingleFileRequiresSelfContained` error number conflicts with another SDK error.
Changing the error number to avoid the conflict.

## Customer Impact

This fixes a bad merge where we ended up with two error messages using the same number.

## Testing

Manual

## Regression

Yes.

## Risk

Low.

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -799,10 +799,6 @@ To install these workloads, run the following command: {1}</value>
     <value>NETSDK1173: The provided type library '{0}' is in an invalid format.</value>
     <comment>{StrBegin="NETSDK1173: "}</comment>
   </data>
-  <data name="CompressionInSingleFileRequiresSelfContained" xml:space="preserve">
-    <value>NETSDK1174: Compression in a single file bundle is only supported when publishing a self-contained application.</value>
-    <comment>{StrBegin="NETSDK1174: "}</comment>
-  </data>
   <data name="TrimmingWindowsFormsIsNotSupported" xml:space="preserve">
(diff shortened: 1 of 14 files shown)


dsplaisted (commented):
Instead of changing the error code to a new error code, can you swap the error codes for NETSDK1174 and NETSDK1176?  IE the `PlaceholderRunCommandProjectAbbreviationDeprecated` resource should use the NETSDK1174 error code, so you can change `CompressionInSingleFileRequiresSelfContained` to use the NETSDK1176 error code.

VSadov:
No problem to change. Just curious why swapping would be better. Would we get another conflict, just with `PlaceholderRunCommandProjectAbbreviationDeprecated`?

dsplaisted:
We're using NETSDK1174 elsewhere, and the placeholder was put in so that we have a record in our main error code file that that error code has been used.  So we should keep the placeholder in sync with that.

We also have tests which are supposed to validate that there are no gaps (or duplicates) in the error codes that we're using, which is why I suggested swapping the error codes instead of leaving a gap.  (I think those tests may not be running though)

VSadov:
@dsplaisted , @agocke  Is there something more that can be done for this fix?   (i.e. fill the bug template, etc..)

dsplaisted (approved):

dsplaisted:
@VSadov The PR looks good, can you fill out the bug template for servicing fixes?

Thanks,
Daniel

VSadov:
Thanks!!!

PR #22684, opened 2021-11-19
Fix version comparison for dotnet tool install

joeloff:
.version file can contain semantic versions causing Version.Parse to throw an exception. Since we're only comparing against the package framework versions, we'll parse and then discard the semantic elements.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
@@ -7,7 +7,6 @@ using System.CommandLine.Parsing;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


sfoslund (approved):
LGTM, but can we add any tests?

joeloff:
> LGTM, but can we add any tests?

I've been thinking about it. It seems none of the tests we have today cover that path because it relies on the .version file of the installed SDK. I'll see if we can add anything

joeloff:
@sfoslund I added a unit test (the change looks wors since I Ctrl-K+D, R+G the test code).

sfoslund (approved):
Test looks good to me!

joeloff:
@sfoslund I'm resetting to before trying to add tests. It appears that we had code in the test project for generating the .version file, but that got removed/commented out. It seems that once we generate the file, it breaks a number of msbuild tests and without it we can't test this fix. I'll follow up with Daniel to figure this out, but I need to unblock ASP.NET

PR #22700, opened 2021-11-20
Update developer-guide to instruct to source the dogfood bash script

4-rodrigo-salazar:
Running the script does not work since it runs it in a sub-shell. You need to source the file so that it exports the env variables in the current shell.


danmoseley (approved):

danmoseley:
thanks!

PR #22750, opened 2021-11-24
Fix ILLink test instalibility

vitek-karas:
The tests created a NuGet package and then refered to it from other test projects with package reference. But some tests recreated the nuget package with slightly different content. Nuget doesn't always "re-restore" the nuget package it seems, and in the local nuget cache some tests got stale version of the package, causing the tests to fail.

Depending on the order in which the tests ran this could lead to failure.

The fix is to use normal project reference instead of going through NuGet. I could not find a reason for the tests to use nuget references like this - it should not make a different for the ILLink coverage.

Also reenabled a test which was disabled due to an already fixed bug.


tlakollo (approved):
Looks good to me

vitek-karas:
Going to merge this to unblock @tlakollo work. Feel free to send feedback and I'll react in a followup change.

PR #22759, opened 2021-11-24
Use Environment.Process{Path,Id} & delete wrappers

am11:
Replaces:

* `Process.GetCurrentProcess().MainModule.FileName` in a few places with [`Environment.ProcessPath`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.processpath#System_Environment_ProcessPath) property, which was introduced in .NET 6.
* `Process.GetCurrentProcess().Id` in a few places with [`Environment.ProcessId`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.processid#System_Environment_ProcessPath) property, which was introduced in .NET 5.

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Internal/DotnetMuxer.cs
+++ b/src/BuiltInTools/dotnet-watch/Internal/DotnetMuxer.cs
@@ -1,6 +1,7 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Diagnostics;
 using System.IO;
 
@@ -10,7 +11,7 @@ namespace Microsoft.DotNet.Watcher.Tools
     {
         static DotnetMuxer()
         {
-            MuxerPath = Process.GetCurrentProcess().MainModule.FileName;
+            MuxerPath = Environment.ProcessPath;
             Debug.Assert(Path.GetFileNameWithoutExtension(MuxerPath) == "dotnet", $"Invalid muxer path {MuxerPath}");
         }
 
--- a/src/RazorSdk/Tool/ServerProtocol/ServerLogger.cs
+++ b/src/RazorSdk/Tool/ServerProtocol/ServerLogger.cs
@@ -111,7 +111,8 @@ namespace Microsoft.NET.Sdk.Razor.Tool
         {
             if (IsLoggingEnabled)
             {
-                var prefix = GetLoggingPrefix();
+                var prefix = string.Format(CultureInfo.InvariantCulture, "{0} PID={1} TID={2} Ticks={3}: ", s_prefix,
+                    GetCurrentProcessId(), Environment.CurrentManagedThreadId, Environment.TickCount);
 
                 var output = prefix + message + "\r\n";
                 var bytes = Encoding.UTF8.GetBytes(output);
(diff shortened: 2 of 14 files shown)


am11:
cc @eerhardt, this will use fast path on .NET 5 and 6, and avoid procfs lookup on linux in favor of `getauxval(AT_EXECFN)` for `ProcessPath` [when available](https://github.com/dotnet/runtime/blob/40dc8631c3a507370e501083e877759b8e45c3b8/src/native/minipal/getexepath.h#L77) (since glibc 2.19 and musl-libc 1.1.0).

src/RazorSdk/Tool/ServerProtocol/ServerLogger.cs
```diff
@@ -124,23 +125,11 @@ public static void Log(string message)
...
-
-        /// <summary>
-        /// Get the string that prefixes all log entries. Shows the process, thread, and time.
-        /// </summary>
-        private static string GetLoggingPrefix()
-        {
-            return string.Format(CultureInfo.InvariantCulture, "{0} PID={1} TID={2} Ticks={3}: ", s_prefix, GetCurrentProcessId(), GetCurrentThreadId(), Environment.TickCount);
-        }
+        private static int GetCurrentProcessId() =>
+#if NET5_0_OR_GREATER
```
eerhardt, line 129: Why is the `#if` needed in this file, but not for other files in this project? Doesn't this project only build for .NET 6+?

src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
```diff
@@ -72,7 +72,11 @@ public string GetDotnetExeDirectory()
 
             if (string.IsNullOrWhiteSpace(dotnetExe))
             {
+#if NET6_0_OR_GREATER
```
eerhardt, line 75: Sometimes you use `NET5_0_OR_GREATER`, and here you use `NET6_0_OR_GREATER`. We should be consistent.

src/Tests/dotnet-watch.Tests/MsBuildFileSetFactoryTest.cs
```diff
@@ -328,7 +328,7 @@ public async Task ProjectReferences_Graph()
 
             var output = new OutputSink();
             var options = GetWatchOptions();
-            var filesetFactory = new MsBuildFileSetFactory(options, DotNetHostPath, _reporter, projectA, output, waitOnError: false, trace: true);
+            var filesetFactory = new MsBuildFileSetFactory(options, _reporter, projectA, output, waitOnError: false, trace: true);
```
eerhardt, line 331: Is the static `DotNetHostPath` field still used? Can it be removed?

src/BuiltInTools/dotnet-watch/Internal/DotnetMuxer.cs
```diff
@@ -1,19 +0,0 @@
...
-using System.IO;
-
-namespace Microsoft.DotNet.Watcher.Tools
-{
-    internal static class DotnetMuxer
-    {
-        static DotnetMuxer()
-        {
-            MuxerPath = Process.GetCurrentProcess().MainModule.FileName;
-            Debug.Assert(Path.GetFileNameWithoutExtension(MuxerPath) == "dotnet", $"Invalid muxer path {MuxerPath}");
```
eerhardt, line 14: While I do like making things simpler, I feel we are missing out on this Debug.Assert with these changes. It makes anyone reading this code understand that we expect the current executable to be `dotnet`, and not some other host.

What do you think about keeping this class around and just changing:

```
MuxerPath = Environment.ProcessPath;
```

?

eerhardt (commented):
Looks good to me. It would be great to get the owners approval as well.

I left some comments for thoughts.

src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
am11: ProcessId was added in 5.0, ProcessPath in 6.0.

src/RazorSdk/Tool/ServerProtocol/ServerLogger.cs
am11: It is actually `Microsoft.NET.Sdk.Razor.Tasks.csproj ` which explicitly includes ServerProtocol and it builds for net472:
https://github.com/dotnet/sdk/blob/45e032c4d0f3625fb5bad6e50067a06b6b7fdc0a/src/RazorSdk/Tasks/Microsoft.NET.Sdk.Razor.Tasks.csproj#L35
https://github.com/dotnet/sdk/blob/45e032c4d0f3625fb5bad6e50067a06b6b7fdc0a/src/RazorSdk/Tasks/Microsoft.NET.Sdk.Razor.Tasks.csproj#L69

src/BuiltInTools/dotnet-watch/Internal/DotnetMuxer.cs
am11: I've brought back DotnetMuxer.

am11:
Thanks @eerhardt. I have addressed the feedback.
cc @dsplaisted, @marcpopMSFT

dsplaisted (approved):

PR #22760, opened 2021-11-24
Remove expected warning from linker test as it's no longer expected.

vitek-karas:
I assume this is a left-over from the very last eventsource trimming changes.

@LakshanF you probably have the most idea what this was about.
In any case, the warning is not generated anymore, so it should be removed regardless.


tlakollo (approved):

LakshanF:
Sorry for not getting to this sooner. I can't remember all the context but there were significant work in the area around and generating the warning around EventSource EnsureDescriptorsInitialized's and no longer generating the warning seems correct to me

PR #22861, opened 2021-12-01
Merge 6.0.1xx into the 6.0.2xx branch to catch it up to current

marcpopMSFT:


marcpopMSFT:
azp /run

dsplaisted (approved):

marcpopMSFT:
@jmarolf, @JoeRobich dotnet format had version number changes in this merge. Typically we don't see that going between branches since there is usually separate codeflow directly to 6.0.2xx for all components. a) is this expected and b) is this because of a missing codeflow from format to the 6.0.2xx branch?

JoeRobich:
> is this because of a missing codeflow from format to the 6.0.2xx branch?

That's the problem. I will update the subscription.

PR #22883, opened 2021-12-03
Support for DisableDiagnosticTracing to suppress TRACE

### LINKED ISSUE #11530: Provide a way to turn off Diagnostic Tracing in projects
Today, tracing support is always enabled by default, unless it is overridden externally (to the project), for e.g., at the msbuild command-line. There seems to be no way - at least no easy way - to control this in a project. 

https://github.com/dotnet/sdk/blob/bc5c514a45a1a900b71d39a3541196bc2cf1bb0d/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props#L21-L24

This could be made opt-in by supplying a property - for e.g., `EnableDiagnosticTracing` - that's `true` by default for compatibility purposes. 

```xml
  <PropertyGroup>
    <DefineConstants Condition=" '$(DefineConstants)' != '' ">$(DefineConstants);</DefineConstants>
    <DefineConstants Condition="'$(EnableDiagnosticTracing)'!='false'">$(DefineConstants)TRACE</DefineConstants>
  </PropertyGroup>
```

vatsan-madhavan:
Fixes #11530

DisableDiagnosticTracing=true will disable the definition of `TRACE` in `DefineConstants`


KalleOlaviNiemitalo:
This semicolon hack would still change e.g. `RATRACE` to `RA`. Wouldn't it be easiest to use `$([System.Text.RegularExpressions.Regex]::Replace(something))`?

vatsan-madhavan:
> This semicolon hack would still change e.g. `RATRACE` to `RA`. Wouldn't it be easiest to use `$([System.Text.RegularExpressions.Regex]::Replace(something))`?

The more I think about this and test a few things, it's getting clear that doing this purely with Property functions is going to be a bit tricky. 

But using a Target makes it much easier to do something like this:

```xml
    <ItemGroup>
      <_DefineConstantsWithoutTrace Include="$(DefineConstants)" />
      <_DefineConstantsWithoutTrace Remove="TRACE"/>
    </ItemGroup>

    <PropertyGroup>
      <DefineConstants>@(_DefineConstantsWithoutTrace)</DefineConstants>
    </PropertyGroup>
```

BeforeCommon.targets already has targets to update/fixup `DefineConstants`, and looks like a better place to fix this using a new target.
https://github.com/dotnet/sdk/blob/0af3943a9106cb63333da3bc7006731590dbd89c/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L223-L241

marcpopMSFT:
Coming back to old PRs. This looks like a good change but should have a test. @vatsan-madhavan any interest in adding that or should we look to find help adding that test?

vatsan-madhavan:
Can you point me at where to add a test? It's been so long I've lost track of where to look ;-)

dsplaisted:
> Can you point me at where to add a test? It's been so long I've lost track of where to look ;-)

@vatsan-madhavan Here is an example of a test that looks like it is very similar to what we would want for DisableDiagnosticTracing: https://github.com/dotnet/sdk/blob/bdf7b7a56d17b8f4ce12262a4ef8ab9717961412/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs#L490

Thanks!

marcpopMSFT:
@vatsan-madhavan ping if you wanted to add a test for this per Daniel's example

marcpopMSFT:
@dsplaisted I went ahead and added a test per your example as this seemed like a useful property to have.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildANetCoreApp.cs
```diff
@@ -1028,5 +1028,48 @@ public void It_warns_on_nonportable_rids(string targetFramework, string[] rids,
...
+}";
+            var testAsset = _testAssetsManager.CreateTestProject(testProj);
+
+            var buildCommand = new BuildCommand(Log, Path.Combine(testAsset.Path, testProj.Name));
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
+
+            var runCommand = new RunExeCommand(Log, Path.Combine(buildCommand.GetOutputDirectory(ToolsetInfo.CurrentTargetFramework).FullName, $"{testProj.Name}.exe"));
```
dsplaisted, line 1070: Hard-coding `.exe` means this will only work on Windows.

```suggestion
            var runCommand = new RunExeCommand(Log, Path.Combine(buildCommand.GetOutputDirectory(ToolsetInfo.CurrentTargetFramework).FullName, $"{testProj.Name}{EnvironmentInfo.ExecutableExtension}"));
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildANetCoreApp.cs
```diff
@@ -1028,5 +1028,48 @@ public void It_warns_on_nonportable_rids(string targetFramework, string[] rids,
...
+            var stdOut = runCommand.Execute().StdOut;
+            stdOut.Should().BeEquivalentTo(expectedOutput);
```
dsplaisted, line 1072: We have assertions specifically for checking the output of a command:

```suggestion
            runCommand
                .Execute().
                .Should().HaveStdOut(expectedOutput);
```

dsplaisted (approved):

vatsan-madhavan:
> @dsplaisted I went ahead and added a test per your example as this seemed like a useful property to have.

Hey @marcpopMSFT thanks for doing this. I've been pretty swamped (I know, lame excuse) and this keeps falling by the wayside. Apologies for giving up and not following through :-/

PR #22902, opened 2021-12-03
Update to use new Razor transport packages

pranavkm:


TanayParikh (approved):

TanayParikh:
Disabled auto-merge. @pranavkm is this something we still want to take in, package versions may be a bit stale now given this is 11 days old

pranavkm:
We spoke about the 11-day packages offline. We haven't produced new 6.0.2xx packages since this PR, so this seems fine.

Issue #22508, opened 2021-11-08
It was not possible to find any installed .NET SDKs.


arivoir:
After installing Visual Studio 2022 Preview Version 17.1.0 Preview 1.0, it installed Net6, but now the sdk's can not be found

> C:\Program Files\Microsoft Visual Studio\2022\Preview>dotnet --version
> Could not execute because the application was not found or a compatible .NET SDK is not installed.
> Possible reasons for this include:
>   * You intended to execute a .NET program:
>       The application '--version' does not exist.
>   * You intended to execute a .NET SDK command:
>       It was not possible to find any installed .NET SDKs.
>       Install a .NET SDK from:
>         https://aka.ms/dotnet-download
> 

But it is installed there

![image](https://user-images.githubusercontent.com/17496741/140793907-3c0fdd1d-0dd0-49f7-af78-b569da07c39a.png)

markhewett:
I had a similar "where did all my SDKs go" issue after installing VS2022 today.  It seems like my path is picking up the 32-bit version of dotnet.exe instead of the 64-bit version, and I only have 64-bit SDKs installed now.  Not sure what changed exactly, but try...

    "C:\Program Files\dotnet\dotnet" --version

And see what that shows.

arivoir:
That works. 

> "C:\Program Files\dotnet\dotnet" --version
> 6.0.100
> 

I wonder what is broken

InKahootz:
Was facing the same problem when I installed VS2022. @markhewett comment led me to a fix.

My fix. Make sure x86 dotnet is below x64 dotnet on System PATH

![image](https://user-images.githubusercontent.com/1909698/140804269-91fa3408-0e82-447d-87cb-8b43a852e1fa.png)

Silvenga:
We ran into it upgrading from VS2022 RC to VS2022 GA. This is going to trip up a lot of people.

marcpopMSFT:
https://github.com/dotnet/core/blob/main/release-notes/6.0/known-issues.md#2-upgrade-of-visual-studio-or-net-sdk-from-earlier-builds-can-result-in-a-bad-path-configuration-on-windows

Looks like there's a bug in the upgrade code that is uninstalling the old host first (removing the path) and then installing the new host. We fixed an iteration of this between preview 5 and GA but apparently missed a specific instance. We're investigating.

tbolon:
I can confirm that I was impacted by this.

[...]

PR #22919, opened 2021-12-06
Improve error message for path mismatch

marcpopMSFT:
Improvement created to help customers with #22508.  A separate change was added to VS to remove the x86 path to try to avoid this but updating the error message just in case anyway.

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
+++ b/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
@@ -130,7 +130,7 @@
     <value>Version {0} of the .NET SDK is smaller than the minimum version {1} required by Visual Studio. Check that a recent enough .NET SDK is installed or increase the version specified in global.json.</value>
   </data>
   <data name="UnableToLocateNETCoreSdk" xml:space="preserve">
-    <value>Unable to locate the .NET SDK. Check that it is installed and that the version specified in global.json (if any) matches the installed version.</value>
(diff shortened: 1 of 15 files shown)


src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
```diff
@@ -130,7 +130,7 @@
     <value>Version {0} of the .NET SDK is smaller than the minimum version {1} required by Visual Studio. Check that a recent enough .NET SDK is installed or increase the version specified in global.json.</value>
   </data>
   <data name="UnableToLocateNETCoreSdk" xml:space="preserve">
-    <value>Unable to locate the .NET SDK. Check that it is installed and that the version specified in global.json (if any) matches the installed version.</value>
+    <value>Unable to locate the .NET SDK. Check that it is installed, your PATH is configured for the correct architecture, and that the version specified in global.json (if any) matches the installed version.</value>
```
baronfel, line 133: thoughts on an aka.ms link to https://github.com/dotnet/core/blob/main/release-notes/6.0/known-issues.md#2-upgrade-of-visual-studio-or-net-sdk-from-earlier-builds-can-result-in-a-bad-path-configuration-on-windows for follow up?  This is a good hint, but I fear that given the number of issues we've seen around it more direct links may be required.

baronfel (approved):

src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
baronfel: Chatted offline - that page isn't stable, so if we have a more permanent link I can come back and add it to this message. This is already a good change though, so let's take it.

PR #22927, opened 2021-12-07
Handle the case when procfs is inaccessible

am11:
Access to proc filesystem is expected to be unavailable. In the runtime repo, we have most calls to files under `/proc` protected by unconditional `catch { }` and some with more specific handlers; either exception object or its inner exception object is `IOException`. This PR applies the latter pattern in two places where SDK reads files from under /proc.

### DIFF
--- a/src/Cli/dotnet/Telemetry/DockerContainerDetectorForTelemetry.cs
+++ b/src/Cli/dotnet/Telemetry/DockerContainerDetectorForTelemetry.cs
@@ -32,9 +32,23 @@ namespace Microsoft.DotNet.Cli.Telemetry
             }
             else if (OperatingSystem.IsLinux())
             {
-                return ReadProcToDetectDockerInLinux()
-                    ? Cli.Telemetry.IsDockerContainer.True
-                    : Cli.Telemetry.IsDockerContainer.False;
+                try
+                {
+                    bool isDocker = File
+                        .ReadAllText("/proc/1/cgroup")
+                        .Contains("/docker/");
+
+                    return isDocker
+                        ? Cli.Telemetry.IsDockerContainer.True
+                        : Cli.Telemetry.IsDockerContainer.False;
+                }
+                catch (Exception ex) when (ex is IOException || ex.InnerException is IOException)
+                {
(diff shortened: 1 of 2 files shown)


am11:
cc @adamsitnik, @dsplaisted, PTAL.
in the runtime repo, all reads from /proc in C# code is protected by try-catch (exception being files listed in `src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/NetworkFiles.cs` which looks to be an oversight).
Scanned with this command: `git -C ~/projects/runtime grep /proc ':/*.cs' ':/!*[tT]est*'`.

src/Cli/dotnet/Telemetry/DockerContainerDetectorForTelemetry.cs
```diff
@@ -48,9 +48,20 @@ public IsDockerContainer IsDockerContainer()
...
+            {
+                return File
+                    .ReadAllText("/proc/1/cgroup")
+                    .Contains("/docker/");
+            }
+            catch (Exception ex) when (ex is IOException || ex.InnerException is IOException)
+            {
+                // in some environments (restricted docker container, shared hosting etc.),
+                // procfs is not accessible and we get UnauthorizedAccessException while the
+                // inner exception is set to IOException. Ignore and continue when that happens.
```
dsplaisted, line 61: It sounds like if we get an exception here, we don't know if we're in a docker container or not.  So should we return `IsDockerContainer.Unkown` from `IsDockerContainer` in that case?

am11: Sure, updated. I didn't realize that there exist `IsDockerContainer.Unknown`. :)

danmoseley:
> exception being files listed in src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/NetworkFiles.cs which looks to be an oversight

maybe we should fix that..

am11:
@danmoseley, sure, I have opened https://github.com/dotnet/runtime/issues/62513 for discussion and gather inputs from area owners and others.

PR #22937, opened 2021-12-07
Update branding to 3.1.417

vseanreesermsft:


wtgodbe:
@marcpopMSFT looks like we need to update from Ubuntu16 here too

PR #22945, opened 2021-12-07
Unskip Blazor Integration Tests

### LINKED ISSUE #38844: Blazor SDK Targets Packaging Fixes
> It's the files package.json and dotnet.d.ts which I added to native of the Microsoft.NETCore.App.Runtime.Mono.browser-wasm.7.0.0-dev.nupkg in runtimes\browser-wasm\native. Each of them twice, once as is, and second as .gz, that's why 4 new items in the diff.

> The Blazor would put those files into wwwroot\_framework\
I think that they would do not harm in there, but I would prefer to keep them compile time files, not copied to wwwroot.
How is that achieved ?

This is a rough corner of how we build the runtime pack, I suspect the new files are probably just escaping a work around in the blazor sdk targets.  We should fix that but I'd like to solve the underlying problem as well.

cc @TanayParikh @javiercn

_Originally posted by @lewing in https://github.com/dotnet/sdk/pull/22892#issuecomment-985933133_

TanayParikh:
Tests were skipped to unblock https://github.com/dotnet/sdk/pull/22892. The `package.json` and `dotnet.d.ts` files were [recently added](https://github.com/dotnet/sdk/pull/22892#issuecomment-985925815) which was causing issues with out baselines. I've filtered those files out as @pavelsavara mentioned they aren't needed.

Fixes: https://github.com/dotnet/aspnetcore/issues/38844

### DIFF
--- a/src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
+++ b/src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
@@ -278,7 +278,8 @@ namespace Microsoft.NET.Sdk.BlazorWebAssembly
                 ".props" when fromMonoPackage => "extension is .props is not supported.",
                 ".blat" when !timezoneSupport => "timezone support is not enabled.",
                 ".dat" when invariantGlobalization && fileName.StartsWith("icudt") => "invariant globalization is enabled",
-                ".json" when fromMonoPackage && fileName == "emcc-props" => $"{fileName}{extension} is not used by Blazor",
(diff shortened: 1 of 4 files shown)


pavelsavara:
I confirm the files don't belong to `wwwroot`, so LGTM.
I don't know enough about RazorSDK yet.

pavelsavara (approved):

TanayParikh:
Going to merge this in so we regain the lost test coverage. I'll follow-up with Pranav or Javier offline once they return to ensure they're fine with the change.

PR #22682, opened 2021-11-18
Blazor AoT Publish Integration Tests

TanayParikh:
- Adds `WasmAOTPublishIntegrationTest` which are Blazor AoT Publish Integration tests
- Adds `Windows_NT_AoT_Tests` and `Darwin_AoT_Tests` (macos) jobs to the existing `dotnet-sdk-public-ci` pipeline.
  - This job builds the SDK and installs the `wasm-tools` workload
  - Only the AoT tests are executed.
  - We have to run the AoT tests in a separate job (however they can still be in the same pipeline), due to #22655. We may be able to consolidate the AoT tests into the existing test job once that issue is resolved.
  - Note, these jobs are run in parallel with the rest of the build/test jobs, however they would of course be requiring additional resources. We'll need to get appropriate clearance before merging.
    - ~~WIP: Get a better idea if this actually increases overall CI execution time, or not.~~

image

The graph depicts the time the Windows & MacOS (Darwin) AoT jobs took, in comparison to the longest job within that CI run. The AoT jobs were never the longest running jobs.

<details>
<summary>How it looks in the CI</summary>

image

image

image

</details>

Fixes: https://github.com/dotnet/aspnetcore/issues/36481

PR #22966, opened 2021-12-08
Merge 6.0.2xx to main

dsplaisted:


dsplaisted:
@captainsafia, @pranavkm, @mkArtakMSFT @TanayParikh It looks like this merge from 6.0.2xx is breaking the tests added in #22682.  They are failing saying the wasm-tools workload isn't installed.

I don't see why this would be happening with these changes.  Can you take a look?  Thanks.

TanayParikh:
> It looks like this merge from 6.0.2xx is breaking the tests added in #22682. They are failing saying the wasm-tools workload isn't installed.
> 
> I don't see why this would be happening with these changes. Can you take a look? Thanks.

Interestingly the Windows AoT jobs [successfully installs](https://dev.azure.com/dnceng/public/_build/results?buildId=1504129&view=logs&j=92885c4a-db2e-5086-f9ba-51524576e2ac&t=2d9cb2ec-6eb3-557c-77d6-4cf5e61b8429) the workload, but the Darwin job [failed to install the workload](https://dev.azure.com/dnceng/public/_build/results?buildId=1504129&view=logs&j=c87185f3-6209-5c78-5ae3-375ac5d75c64&t=bb045e65-84ac-5a8f-e767-c27deb9e7544&s=6884a131-87da-5381-61f3-d7acc3b91d76). 

I don't believe I've seen that particular step from Darwin fail before in that nature. Given the [automated retry succeeded](https://dev.azure.com/dnceng/public/_build/results?buildId=1504129&view=logs&j=6b1c5615-6e94-55cf-f960-932a885bcfb6&t=41358b1e-b44b-5df1-e976-67b05ffac601) I'm going to chalk this up to momentary flakiness. Will keep an eye on it though.

TanayParikh:
> I don't believe I've seen that particular step from Darwin fail before in that nature.

Given the manifest failures, we may wish to consider updating:

https://github.com/dotnet/sdk/blob/9b336e6216de6b88ae4df917166e048dee6748ad/eng/build.yml#L238-L239

with the `--skip-manifest-update --no-cache`, I believe this is what @radical does in dotnet/runtime. Will put up a PR with this change.

dsplaisted:
FYI @marcpopMSFT @joeloff I was getting consistent errors in one leg that I think were coming from the way Helix and non-Helix tests were running in parallel.  What I think was happening is that when one of the parallel processes (not sure if it was for the Helix tests or not) was killing running `dotnet` processes (with the message "Killing running build processes..."), which killed the other parallel process.

I added 97a2fb6 to try to add more logging output to help investigate, but now the leg is passing.  But anyway, if the same thing happens again this is probably what's going on.

PR #22974, opened 2021-12-08
Update application insights version

marcpopMSFT:


marcpopMSFT:
See https://github.com/microsoft/ApplicationInsights-dotnet/issues/1152 for more details on this change. Unfortunately, I'm missing context on telemetry to be sure the fix is correct.

dsplaisted (commented):
Per [this comment](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1152#issuecomment-518742922)

> Keep one `TelemetryConfiguration` per app lifetime (a few is also fine), but you may create as many clients as you want.
> Do not forget to dispose `TelemetryConfiguration` when gracefully shutting down the application.

It looks like we're not doing that here, and probably we should be.

dsplaisted (approved):

PR #23085, opened 2021-12-15
Update System.CommandLine

baronfel:
This updates our version of System.CommandLine to the latest available and adjusts to some API breaking changes around:
* Completions
* Binding Handlers to Commands
* Help Layout

The vast majority of the existing tests pass on my machine, I've created this to

* get CI verification
* get initial feedback on the approach
* determine where we should target this


src/Cli/dotnet/commands/CommandExtensions.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.CommandLine.Parsing;
+using System.CommandLine;
+
+namespace Microsoft.DotNet.Cli {
+
+    public static class CommandExtensions {
+        public static Command SetHandler(this Command command, Func<ParseResult, int> func) {
```
baronfel, line 8: This extension method follows some patterns in the S.CL codebase itself, and is typed specifically for the delegates that we have currently. The S.CL overloads of `SetHandler` take `Action<'t>` and `Func<_, Task>` overloads only, which didn't fit with our existing models, so rather than rewrite those entirely I made this wrapper shim to fix the shape mismatches.

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -129,52 +131,62 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
...
+                 shortProjectOption,
+                 listOption
             };
-
+            
             root.TreatUnmatchedTokensAsErrors = false;
-            root.Handler = CommandHandler.Create((CommandLineOptions options, ParseResult parseResults) =>
-            {
-                if (string.IsNullOrEmpty(options.Project))
+            root.SetHandler((ParseResult parseResults) => {
```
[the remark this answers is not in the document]
vlada-shubina: This is still possible via binder, example here: https://github.com/dotnet/templating/blob/5c68d5bd7cb4a8e4c836487d2472e9c67bca92b2/src/Microsoft.TemplateSearch.TemplateDiscovery/TemplateDiscoveryCommand.cs#L112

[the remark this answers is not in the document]
vlada-shubina: I usually add it together with command definition

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -129,52 +131,62 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
...
+                var projectValue = parseResults.GetValueForOption(longProjectOption);
```
vlada-shubina, line 155: nit: it would be good to extract to method probably

src/BuiltInTools/dotnet-watch/Program.cs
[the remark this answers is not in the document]
baronfel: Thanks, that's exactly what I was looking for!

src/Cli/dotnet/commands/CommandExtensions.cs
vlada-shubina: this doesn't look correct, I guess it is needed for commands like `command.SetHandler((parseResult) => new RemoveProjectToProjectReferenceCommand(parseResult).Execute());` 
why parse result is needed to create the command?

src/BuiltInTools/dotnet-watch/Program.cs
baronfel: With your comment above it can be extracted to the binder, it's not necessary for the actual handler logic.

src/Cli/dotnet/commands/CommandExtensions.cs
baronfel: Digging into the various commands, the parse results are used to fetch the results of various arguments and options.  This is something that in more recent versions of S.CL would be done at the binding layer prior to the command's handler being invoked. So this will likely be a larger effort involving defining 'options' structures for these commands and passing those into constructors, dropping the parse results requirement.

There are other things that happen with the parse results (like debugging support or triggering help usage + fast failure on parse errors) but those can be done via the middleware concept on the main builder now to ensure consistency.

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -129,52 +130,28 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
                 return null;
             });
 
-            var projectOption = new Option<string>("-p", "The project to watch") { IsHidden = true };
+            var listOption = new Option<bool>(
+                    "--list",
+                    "Lists all discovered files without starting the watcher");
```
pranavkm, line 135:
```suggestion
          "--list",
          "Lists all discovered files without starting the watcher");
```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -129,52 +130,28 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
...
+
+            var shortProjectOption = new Option<string>("-p", "The project to watch") { IsHidden = true };
+            var longProjectOption = new Option<string>("--project","The project to watch");
+            var noHotReloadOption = new Option<bool>(
+                    new[] { "--no-hot-reload" },
+                    "Suppress hot reload for supported apps.");
```
pranavkm, line 141:
```suggestion
           new[] { "--no-hot-reload" },
           "Suppress hot reload for supported apps.");
```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -129,52 +130,28 @@ internal static RootCommand CreateRootCommand(Func<CommandLineOptions, Task<int>
...
-                 new Option<bool>(
-                    "--list",
-                    "Lists all discovered files without starting the watcher"),
+                 noHotReloadOption,
+                 longProjectOption,
+                 shortProjectOption,
+                 listOption
             };
-
+
```
pranavkm, line 151:
```suggestion

```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -420,5 +397,63 @@ private static void RegisterAssemblyResolutionEvents(string sdkRootDirectory)
                 return null;
             };
         }
+        private class CommandLineOptionsBinder : BinderBase<CommandLineOptions> {
```
pranavkm, line 400:
```suggestion

        private sealed class CommandLineOptionsBinder : BinderBase<CommandLineOptions> {
```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -420,5 +397,63 @@ private static void RegisterAssemblyResolutionEvents(string sdkRootDirectory)
...
+                _longProjectOption = longProjectOption;
+                _shortProjectOption = shortProjectOption;
+                _quietOption = quietOption;
+                _listOption = listOption;
+                _noHotReloadOption = noHotReloadOption;
+                _verboseOption = verboseOption;
+                _reporter = reporter;
+            }
+
+            protected override CommandLineOptions GetBoundValue(BindingContext bindingContext) {
```
pranavkm, line 419:
```suggestion
            protected override CommandLineOptions GetBoundValue(BindingContext bindingContext)
            {
```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -420,5 +397,63 @@ private static void RegisterAssemblyResolutionEvents(string sdkRootDirectory)
...
+                    remainingArguments.Add("--");
+                    remainingArguments.AddRange(parseResults.UnparsedTokens);
+                }
+                else
+                {
+                    remainingArguments.AddRange(parseResults.UnmatchedTokens);
+                    remainingArguments.AddRange(parseResults.UnparsedTokens);
+                }
+
+                var options = new CommandLineOptions {
```
pranavkm, line 446:
```suggestion
                var options = new CommandLineOptions
                {
```

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -10,7 +10,7 @@
 using System.Linq;
 using Microsoft.DotNet.Cli.Utils;
 using System.CommandLine.Parsing;
-using System.CommandLine.Suggestions;
+using System.CommandLine.Completions;
```
pranavkm, line 13: Sort usings (.Completions would come before .Parsing)

src/Cli/dotnet/commands/CommandExtensions.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.CommandLine.Parsing;
+using System.CommandLine;
+
+namespace Microsoft.DotNet.Cli {
```
pranavkm, line 5:
```suggestion
namespace Microsoft.DotNet.Cli;
```

src/Cli/dotnet/commands/CommandExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+    public static class CommandExtensions {
```
pranavkm, line 7: Formatting - braces usually go on the next line

src/Cli/dotnet/commands/ParseResultCommandHandler.cs
```diff
@@ -0,0 +1,17 @@
+using System.CommandLine.Invocation;
+using System.Threading.Tasks;
+using System;
+using System.CommandLine.Parsing;
+
+namespace Microsoft.DotNet.Cli
+{
+    internal class ParseResultCommandHandler : ICommandHandler
+    {
+        private Func<ParseResult, int> _action;
```
pranavkm, line 10:
```suggestion
        private readonly Func<ParseResult, int> _action;
```

src/Cli/dotnet/commands/ParseResultCommandHandler.cs
```diff
@@ -0,0 +1,17 @@
...
+
+        internal ParseResultCommandHandler(Func<ParseResult, int> action) {
```
pranavkm, line 12:
```suggestion
        internal ParseResultCommandHandler(Func<ParseResult, int> action)
        {
```

src/Cli/dotnet/commands/ParseResultCommandHandler.cs
```diff
@@ -0,0 +1,17 @@
+using System.CommandLine.Invocation;
```
pranavkm, line 1: Missing license header. Also usings need to be sorted.

baronfel:
Thanks for the style/consistency review @pranavkm, I'll get right on those changes.

eng/Versions.props
```diff
@@ -38,7 +38,7 @@
     <SystemReflectionMetadataLoadContextVersion>7.0.0-alpha.1.21614.10</SystemReflectionMetadataLoadContextVersion>
     <DeploymentReleasesVersion>1.0.0-preview1.1.21112.1</DeploymentReleasesVersion>
     <SystemManagementPackageVersion>4.6.0</SystemManagementPackageVersion>
-    <SystemCommandLineVersion>2.0.0-beta1.21601.2</SystemCommandLineVersion>
+    <SystemCommandLineVersion>2.0.0-beta2.21616.4</SystemCommandLineVersion>
```
MichaelSimons, line 41: When updating this, the Versions.Details.xml should be updated as well (e.g. https://github.com/dotnet/sdk/pull/23032) otherwise there is a good changes source-build will be broken.

baronfel: Thanks @MichaelSimons! Follow up - are there docs I can read on this whole process (assume I know zero about source-build), and is there tooling we can put in the repo to ensure they are kept in sync (or better - automatically updated)?

pranavkm: Usually both these files are updated via automations (note that this thing is under an ItemGroup labeled "Automated"). But I think System.CommandLine is a bit of an exception given how dramatically their APIs change between releases.

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -420,5 +397,65 @@ private static void RegisterAssemblyResolutionEvents(string sdkRootDirectory)
                 return null;
             };
         }
+        private sealed class CommandLineOptionsBinder : BinderBase<CommandLineOptions> {
```
pranavkm, line 400:
```suggestion
        private sealed class CommandLineOptionsBinder : BinderBase<CommandLineOptions>
        {
```

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -420,5 +397,65 @@ private static void RegisterAssemblyResolutionEvents(string sdkRootDirectory)
...
+            private readonly Option<string> _longProjectOption;
+            private readonly Option<string> _shortProjectOption;
+            private readonly Option<bool> _quietOption;
+            private readonly Option<bool> _listOption;
+            private readonly Option<bool> _noHotReloadOption;
+            private readonly Option<bool> _verboseOption;
+            private readonly IReporter _reporter;
+
+            internal CommandLineOptionsBinder(Option<string> longProjectOption, Option<string> shortProjectOption, Option<bool> quietOption, Option<bool> listOption, Option<bool> noHotReloadOption, Option<bool> verboseOption, IReporter reporter) {
```
pranavkm, line 409:
```suggestion
            internal CommandLineOptionsBinder(Option<string> longProjectOption, Option<string> shortProjectOption, Option<bool> quietOption, Option<bool> listOption, Option<bool> noHotReloadOption, Option<bool> verboseOption, IReporter reporter)
            {
```

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -1,16 +1,16 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
+using Microsoft.DotNet.Cli.Utils;
```
pranavkm, line 4: Minor nit: Most of these repos sort usings where System namespaces appear first. There's a VS setting / .editorconfig that will do that for you. I'm surprised the .editorconfig didn't kick in: https://github.com/dotnet/sdk/blob/main/.editorconfig#L81

src/Cli/dotnet/commands/CommandExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.DotNet.Cli;
+
+public static class CommandExtensions
+{
+    public static Command SetHandler(this Command command, Func<ParseResult, int> func)
+    {
+        command.Handler = new ParseResultCommandHandler(func);
+        return command;
+    }
+}
```
pranavkm, line 17:
```suggestion
}

```

src/Cli/dotnet/commands/ParseResultCommandHandler.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    private readonly Func<ParseResult, int> _action;
+
+    internal ParseResultCommandHandler(Func<ParseResult, int> action)
+    {
+        _action = action;
+    }
+
+    public Task<int> InvokeAsync(InvocationContext context) => Task.FromResult(_action(context.ParseResult));
+}
```
pranavkm, line 21:
```suggestion
}

```

src/Tests/dotnet.Tests/CommandTests/CompleteCommandTests.cs
```diff
@@ -327,7 +327,6 @@ public void GivenDotnetToolInWithPosition()
             var expected = new[] {
                 "install",
                 "uninstall",
-                "--info"
```
pranavkm, line 330: I'll run the test change by @dsplaisted since he has the most context here.

pranavkm (approved):
dotnet-watch changes look good to me.

src/Cli/dotnet/CommonOptions.cs
baronfel: would dotnet-format apply these as well? if so it's probably a good practice for me to start running that over changed files before commit.

src/Tests/dotnet.Tests/CommandTests/CompleteCommandTests.cs
baronfel: I think this was just a bug in the old completions that was fixed in library updates. At the position the completion was requested (dotnet tool in|), where | is the cursor, `--info` isn't a valid option. `--info` is only valid on the root `dotnet` command.

src/Tests/dotnet-run.Tests/GivenDotnetRunRunsCsProj.cs
```diff
@@ -635,7 +635,7 @@ public void ItForwardsEmptyArgumentsToTheApp()
                 .Should()
                 .Pass()
                 .And
-                .HaveStdOutContaining($"0 = a{Environment.NewLine}1 = {Environment.NewLine}2 = c");
+                .HaveStdOutContaining($"0 = a{Environment.NewLine}1 = c");
```
baronfel, line 638: explicitly calling out this test change as well: S.CL strips empty arguments now, as I understand it this is in keeping with shell argument parsing semantics across multiple platforms: https://github.com/jonsequitur/command-line-api/commit/260a559f5a5f4ec8d075a7accec1f2a185528a43

eng/Version.Details.xml
```diff
@@ -256,9 +256,9 @@
       <Sha>e101c379c6e3c7011562ee3af3e3de066da0048f</Sha>
       <SourceBuild RepoName="roslyn-analyzers" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="System.CommandLine" Version="2.0.0-beta1.21601.2">
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta2.21623.1">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
-      <Sha>8f280ecd202ab3ce64af177a6e2aa18644a26f6f</Sha>
+      <Sha>5ca753b7e262de171234ade2936796a5f0822fcb</Sha>
     </Dependency>
     <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.257901">
```
MichaelSimons, line 263: This dependency needs to be updated as well.  The sha should be identical to the `System.CommandLine` dependency.  There is no mechanism to enforce this.  The appropriate way to ensure everything is updated correct is to use the darc command - `darc update-dependencies --id <buildId>`

baronfel: Thanks again @MichaelSimons, I've done that just now. The DARCs docs were super helpful to get started👍

MichaelSimons (approved):
LGTM from a source-build perspective.

pranavkm:
@baronfel could you cherry-pick this commit: https://github.com/dotnet/sdk/pull/23254/commits/1f47a3877a6358d3de2c4c31d2f9d4ded24f2de8? It fixes the failing blazor tests - see https://github.com/dotnet/sdk/pull/23254.

baronfel:
@pranavkm done, thank you!

baronfel:
Some of the completion tests are failing because the update introduced a regression around completions for global options: https://github.com/dotnet/command-line-api/issues/1563. There are two kinds of test failures remaining:
* there's some oddity around the name of the commandFile some project manipulation tests
* some msbuild tests are getting a distributed logger option forwarded unexpectedly
still digging in to both of those categories of error.

baronfel:
Feeling pretty happy about this now: the main tests failing in CI now are the completions ones, and S.CL already has an MR out to fix that.  Hopefully we can get green in the next day or so.

pranavkm:
@baronfel I noticed this when I was looking at the blazor test failures - it looks like the CommandHandler type exists in a separate assembly: https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine.NamingConventionBinder/CommandHandler.cs. Did you intentionally choose to avoid that reference? It might have been less churn.

baronfel:
@pranavkm I did intentionally avoid taking on that reference, mostly because it introduces a reflection component. The base library was mostly expressive enough IMO (at least for the core SDK commands), and being more linker-friendly seemed like a good secondary goal as well.  I'm not hard-against it, I just didn't see a compelling enough reason to require it.

baronfel:
Rebased and updated S.CL to fix the completions errors. Fingers crossed, this could be the one!

baronfel:
All green, this was a fun one :)

PR #23086, opened 2021-12-15
Files not ending in css are probably not css files

Bouke:
The log message incorrectly labels these files "css" files, but actual "css" files are handled in the other branch.


baronfel:
This looks sane to me, but I'd like @pranavkm's eyes on it real quick in case something else is happening here that I'm just not seeing.

pranavkm (approved):
Thanks!

mkArtakMSFT (approved):
Thanks @Bouke. Sorry for taking so long to get this merged.

PR #23097, opened 2021-12-17
Fix dotnet tool list command

joeloff:
The command handler was assigning the wrong command type (Restore) and not assigning the created handler to the command.


baronfel (approved):

dougbu:
Curious: Do we have tests of `dotnet tool list`❔

joeloff:
> Curious: Do we have tests of `dotnet tool list`❔

doesn't look like it.

PR #23177, opened 2021-12-26
Build Single-File for Windows 7

NN---:
After fixing Single-File support in .NET: https://github.com/dotnet/runtime/pull/63196 , the SDK should not issue an error.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -159,10 +159,6 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkError Condition="'$(PublishSingleFile)' == 'true' and '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0'"
                  ResourceName="PublishSingleFileRequiresVersion30" />
 
-    <NETSdkError Condition="'$(PublishSingleFile)' == 'true' and $(RuntimeIdentifier.StartsWith('win7-'))"
-                 ResourceName="SingleFileWin7Incompatible"
-                 FormatArguments="$(RuntimeIdentifier)" />
-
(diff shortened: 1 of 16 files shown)


shaulz1 (approved):

NN---:
@VSadov

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -159,10 +159,6 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkError Condition="'$(PublishSingleFile)' == 'true' and '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0'"
                  ResourceName="PublishSingleFileRequiresVersion30" />
 
-    <NETSdkError Condition="'$(PublishSingleFile)' == 'true' and $(RuntimeIdentifier.StartsWith('win7-'))"
-                 ResourceName="SingleFileWin7Incompatible"
-                 FormatArguments="$(RuntimeIdentifier)" />
-
```
vitek-karas, line 165: Since this is the only place where the string resource is used we should remove it as well (tests don't count, we can modify the test to check just for the `NETSDK1180` string).

NN---: I thought about it.
My assumption is that Windows 7 is going to be deprecated somewhere in a future, so it can be left.

I can remove all string resources and fix the tests if needed.

vitek-karas: I would prefer we remove it - if it's left I would not be surprised that in the future when we add it back (if) we will find the unused message and reuse it.

NN---: Removed

agocke (approved):

PR #23249, opened 2022-01-04
Opt into IEventSource4

### LINKED ISSUE #23248: Properties and items aren't logged at evaluation for `dotnet build`
Unlike MSBuild, when you run `dotnet build`, something is preventing the new behavior of logging properties and items after evaluation. 

As a result we get legacy broken behavior where some properties and items are logged for some projects at ProjectStarted, but not others, and none at all are logged at evaluation.

This is very bad and we should fix this soon.

rainersigwald:
This enables MSBuild to use more-efficient and higher-fidelity logging approaches that ensure capturing the post-evaluation state of items and properties, which is extremely helpful for debugging.

Fixes #23248.


rainersigwald:
@dsplaisted is there any test coverage of binlog outputs? I didn't see any so didn't see a good spot to add a test for this.

KirillOsenkov (approved):
:heart:

PR #23321, opened 2022-01-07
Pin Microsoft.CodeAnalysis version for API Compat to the 4.0.x stable package.

### LINKED ISSUE #22090: [API Compat] Pin Microsoft.CodeAnalysis dependency to 4.0.0 once it ships
See the conversation in https://github.com/dotnet/sdk/pull/22072#discussion_r729539313.

ApiCompatibility is currently taking a dependency on the latest `Microsoft.CodeAnalysis` package that is being used in the SDK. However, the Compatibility package also ships in an OOB NuGet package, which means it can run on older SDK versions than what it is currently targeted by the SDK.

Once Microsoft.CodeAnalysis v4.0.0 ships, we should pin the dependency to reference 4.0.0. That way when we ship a new version of the Compatibility package, it will still work with older versions (i.e. 4.0.0) of Roslyn.

safern:
Fixes: https://github.com/dotnet/sdk/issues/22090


eerhardt (approved):
LGTM.

FYI @MichaelSimons @crummel

PR #23358, opened 2022-01-10
Better caching for RazorSourceGenerator when SuppressRazorSourceGener…

### LINKED ISSUE #32867: Net Core 6 Blazor Server Hot Reload is Slow for Big Projects
Hi,

I have a project Blazor Server with ~ 180 .razor Files (186 Files, 55 Folder)

If all folder structure is in the same project the hot reload takes ~ 15 seconds , more than build time ( 10 seconds ) .

If we split the .razor files on Libraries , we get the next Results

Main App (Only Host.razor) : (Reference to Shared Library, Front Library,Admin Library )
– Hot Reload for all tests after 1 seconds

Shared Library (21 Files, 1 Folder) : Layout, Shared Components, etc
– Add p: Hot Reload after 2-3 seconds
– Add style to p : Hot Reload after after 2-3 seconds
– Counter Example with ++ : Hot Reload after 1 seconds

Front Library (54 Files, 15 Folders) (Reference to Shared Library)
– Add p: Hot Reload after 20 seconds
– Add style to p : Hot Reload after after 4 seconds
– Counter Example with ++ : Hot Reload after 3 seconds

Admin Library (106 Files, 36 Folders) (Reference to Shared Library)
– Add p: Hot Reload after 12 seconds
– Add style to p : Hot Reload after after 8 seconds
– Counter Example with ++ : Hot Reload after 8 seconds

I am suspecting  maybe we have something from old Projects because we have upgraded from 2.2 to 3.1 ( Blazor ) to 5,etc

We have tried to remove html from Some razor Files specially Layout and seems to work quickly if there is just some simple html lines.

How can we improve ? It depends on File nr and complexity ? or depends on references ? How can we structure the project ?

PS: The issue is posted based on disscussion : https://devblogs.microsoft.com/aspnet/asp-net-core-updates-in-net-6-preview-3/comment-page-3/#comment-3245

pranavkm:
…ator changes

In VisualStudio SuppressRazorSourceGenerator changes when it's being invoked by the EnC and tooling.
In our current implementation, while we no-op the operation, when SuppressRazorSourceGenerator=true,
the code also evicts the valid previously cached results when SuppressRazorSourceGenerator=false by returning null values,
causing us do work the next time its false.

Fixes https://github.com/dotnet/aspnetcore/issues/32867

### DIFF
--- a/src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
+++ b/src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
@@ -15,18 +15,9 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
         public RazorConfiguration Configuration { get; set; } = RazorConfiguration.Default;
 
         /// <summary>
-        /// Gets a flag that determines if the source generator waits for the debugger to attach.
-        /// <para>
-        /// To configure this using MSBuild, use the <c>_RazorSourceGeneratorDebug</c> property.
-        /// For instance <c>dotnet msbuild /p:_RazorSourceGeneratorDebug=true</c>
-        /// </para>
-        /// </summary>
-        public bool WaitForDebugger { get; set; } = false;
-
-        /// <summary>
         /// Gets a flag that determines if generated Razor views and Pages includes the <c>RazorSourceChecksumAttribute</c>.
         /// </summary>
-        public bool GenerateMetadataSourceChecksumAttributes { get; set; } = false;
+        public bool GenerateMetadataSourceChecksumAttributes { get; set; }
 
         /// <summary>
         /// Gets a flag that determines if the source generator should no-op.
@@ -36,7 +27,7 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
         /// The property is set by the SDK via an editor config.
         /// </para>
         /// </summary>
-        public bool SuppressRazorSourceGenerator { get; set; } = false;
+        public bool SuppressRazorSourceGenerator { get; set; }
 
         /// <summary>
         /// Gets the CSharp language version currently used by the compilation.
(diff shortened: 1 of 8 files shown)


src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -96,17 +109,27 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
             var tagHelpersFromCompilation = context.CompilationProvider
                 .Combine(generatedDeclarationSyntaxTrees.Collect())
                 .Combine(razorSourceGeneratorOptions)
+                .WithLambdaComparer(static (a, b) =>
```
chsienki, line 112: Consider extracting this logic out into either an `IValueComparer<Tuple<object, RazorSourceGenerationOptions>>` or another extension method, so you don't have to duplicate the logic.

chsienki (approved):

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
```diff
@@ -14,19 +14,10 @@ internal class RazorSourceGenerationOptions : IEquatable<RazorSourceGenerationOp
...
         public RazorConfiguration Configuration { get; set; } = RazorConfiguration.Default;
 
-        /// <summary>
-        /// Gets a flag that determines if the source generator waits for the debugger to attach.
-        /// <para>
-        /// To configure this using MSBuild, use the <c>_RazorSourceGeneratorDebug</c> property.
-        /// For instance <c>dotnet msbuild /p:_RazorSourceGeneratorDebug=true</c>
-        /// </para>
-        /// </summary>
-        public bool WaitForDebugger { get; set; } = false;
```
captainsafia, line 24: Any motivation for removing the debug flag?

captainsafia (commented):
Retroactive approve.

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
pranavkm: We stopped using it when we migrated to the incremental APIs. You kinda need to break inside one of the steps for this to work reliably and we never did that.

PR #23419, opened 2022-01-13
Better caching for RazorSourceGenerator when SuppressRazorSourceGener…

pranavkm:
…ator changes (#23358)

In VisualStudio SuppressRazorSourceGenerator changes when it's being invoked by the EnC and tooling.
In our current implementation, while we no-op the operation, when SuppressRazorSourceGenerator=true,
the code also evicts the valid previously cached results when SuppressRazorSourceGenerator=false by returning null values, causing us do work the next time its false.

Contributes to dotnet/aspnetcore#32867

Backport #23358 to 6.0.2xx

### DIFF
--- a/src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
+++ b/src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
@@ -1,4 +1,3 @@
-
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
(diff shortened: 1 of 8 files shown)


TanayParikh (approved):
Approving backport.

mikes-gh:
I see the do not merge has been added.  We are eagerly awaiting this fix so our large Blazor project can use hot reload in VS.  At the moment its too slow to use.

pranavkm:
@mikes-gh my plan is to get this merged in the 6.0.3xx release which aligns with VS 17.2. The code involves quite a bit of how the source generator caches and there's a very real risk that we introduce subtle caching bugs that weren't caught in local testing. Putting it in the 17.2 preview builds gives us a bit more time to address bugs while gathering feedback.

PR #23434, opened 2022-01-14
Fixed vstest version in release/6.0.2xx

Haplois:
[Test Platform 17.2](https://github.com/microsoft/vstest) versions leaked into this branch because of an incorrect config,
this commit inserts the correct version instead.


nohwnd (dismissed):

nohwnd:
@Haplois were the changes that @MarcoRossignoli did yesterday (changes to how we resolve the testhost in x86 runner) ported to 17.1.0? If not this would roll back that fix. Can you sync to get the fix into 6.0.2xx as well?

Haplois:
@nohwnd yes, they are included in this version

nohwnd (approved):

PR #23470, opened 2022-01-18
Allow skipping ILLink target

MichalStrehovsky:
For NativeAOT we need to be able to opt in to all `PublishTrimmed` behaviors except the part that runs ILLink. We currently do that in a hacky way (set `PublishTrimmed` to true and create ILLink semaphore file to trick ILLink into not running); it's not great.

I tried introducing a `EnableTrimmingBehaviors` property that becomes true when `PublishTrimmed` is true, but in the end, we would like to be able to support NativeAOT for projects that have `PublishTrimmed` in the project file without NativeAOT failing, so what we really need is a way to disable running ILLink even though `PublishTrimmed` is true.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets
@@ -101,6 +101,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     -->
   <Target Name="ILLink"
           Condition=" '$(PublishTrimmed)' == 'true' And
+                      '$(RunILLink)' != 'false' And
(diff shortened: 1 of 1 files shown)


sbomer (approved):
The code change looks good for now but I am wondering about the approach. I see that the NativeAOT build logic is adding new dependencies on some of the linker targets and fixing up the output of `_ComputeManagedAssemblyToLink`. I wonder if we should instead tweak the linker build logic to make the relevant parts reusable from NativeAOT without depending on implementation details? I'm not sure how easy that will be in MSBuild, curious what you think.

MichalStrehovsky:
I think we just need to override it because the rest of the ILCompiler hookup is a bit hacky. It's possible we can put it into a shape where the hookup is not needed. I don't like depending on private targets either.

PR #23502, opened 2022-01-19
Workloads: Enable support for SxS previews

joeloff:
Ensure that the sentinel file can be detected correctly. The sentinel file always resides in a folder that excludes any prerelease values, e.g. 7.0.104-preview.4 will still install the sentinel to 7.0.100. To support SxS installs of previews, the SdkFeatureBand version will include prerelease values, so we need to ensure that we use the short form that only includes the major.minor.featureband (without patch) to detect the sentinel file.


joeloff:
Fixes #23503

marcpopMSFT (approved):

PR #23519, opened 2022-01-20
Fix version.details entry for System.Commandline

wtgodbe:
Matches what we have in 6.0.1xx, and should unblock https://github.com/dotnet/installer/pull/12981

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -268,9 +268,9 @@
       <Sha>c750b5a665adb75b528a93d844f238bd1360a91a</Sha>
       <SourceBuild RepoName="roslyn-analyzers" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="System.CommandLine" Version="2.0.0-beta1.21601.2">
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta1.21473.1">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
-      <Sha>8f280ecd202ab3ce64af177a6e2aa18644a26f6f</Sha>
+      <Sha>82273cb56c83b589e8e5b63da0ac9745ffc6e105</Sha>
     </Dependency>
     <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.257901">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -272,9 +272,9 @@
       <Uri>https://github.com/dotnet/command-line-api</Uri>
       <Sha>82273cb56c83b589e8e5b63da0ac9745ffc6e105</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.257901">
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.247301">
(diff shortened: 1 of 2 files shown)


eng/Version.Details.xml
```diff
@@ -268,9 +268,9 @@
       <Sha>c750b5a665adb75b528a93d844f238bd1360a91a</Sha>
       <SourceBuild RepoName="roslyn-analyzers" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="System.CommandLine" Version="2.0.0-beta1.21601.2">
+    <Dependency Name="System.CommandLine" Version="2.0.0-beta1.21473.1">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
-      <Sha>8f280ecd202ab3ce64af177a6e2aa18644a26f6f</Sha>
+      <Sha>82273cb56c83b589e8e5b63da0ac9745ffc6e105</Sha>
     </Dependency>
     <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.257901">
```
MichaelSimons, line 275: This dependency need to be updated as well.  It should be point to the same sha.

MichaelSimons (approved):

marcpopMSFT:
@wtgodbe we saw this before randomly and then it went away on rerun. I'm not sure if it's something happening to the images used for building or a networking issue.  We do exclude those versions of the runtime for other build legs here where they aren't supported but they're supposed to be supported on the legs they are trying to install them on: https://github.com/dotnet/sdk/blob/0a093cf221875beb36d509d0b407c1b8f61bd1af/eng/restore-toolset.sh#L24

MichaelSimons:
> I'm not sure if it's something happening to the images used for building or a networking issue.

The images haven't changed so I suspect networking or something else.

This leg is using `mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-3e800f1-20190501005343`.

One thing I just noticed is this https://github.com/dotnet/sdk/blob/0a093cf221875beb36d509d0b407c1b8f61bd1af/eng/restore-toolset.sh#L10

```
  # The following frameworks and tools are used only for testing.
  # Do not attempt to install them in source build.
  if [[ "${DotNetBuildFromSource:-}" == "true" ]]; then
    return
  fi
```

Our builds aren't setting this and are run via `./build.sh --ci --configuration Release --restore --build --pack --publish -bl /p:SourceBuildNonPortable= /p:ArcadeBuildFromSource=true`

wtgodbe:
@MichaelSimons what's the difference between `DotNetBuildFromSource` and `ArcadeBuildFromSource`? Changing `restore-toolset.sh` to use the latter still tries to restore 1.0.5, and I see both are used in Arcade

marcpopMSFT:
@wtgodbe a new install script released today so it might be that. I pinged @bekir-ozturk and the first responders channel.

MichaelSimons:
> @MichaelSimons what's the difference between DotNetBuildFromSource and ArcadeBuildFromSource? Changing restore-toolset.sh to use the latter still tries to restore 1.0.5, and I see both are used in Arcade

ArcadeBuildFromSource is what is used to trigger a source-build.  DotNetBuildFromSource is used to detect if you are within the inner source-build.  I know it is confusing but it was necessary for legacy reasons.  That aside, I think the issue is `/p:ArcadeBuildFromSource=true` is getting passed as a msbuild argument to build.sh and is therefore not set as an ENV in the context of this script.

PR #23525, opened 2022-01-20
Remove 1.0.5 and 1.1.2 installs from our linux runtimes

marcpopMSFT:
There has been instability in the install scripts here so seeing what fails if I remove it.

### DIFF
--- a/eng/restore-toolset.sh
+++ b/eng/restore-toolset.sh
@@ -20,10 +20,6 @@ function InitializeCustomSDKToolset {
   fi
 
   InitializeDotNetCli true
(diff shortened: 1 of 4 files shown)


wtgodbe (approved):

marcpopMSFT:
@wtgodbe The new install script should be live in the next half hour per Vlada. I talked to Daniel and he said to hold off taking this PR for now since the script is getting fixed.  We have it if we need it but we should still have tests for 1.0 so no reason to take this now.

marcpopMSFT:
Closing for now as the downlevel install is now unblocked.

PR #23629, opened 2022-01-27
[wip]:  Include fsharp core

KevinRansom:
Update the Sdk to incorporate FSharp.Core as a library-pack.  Allowing fsharp to work offline.  And facility simple distribution of preview fsharp.cores.

Also notifies arcade, of the correct versions of fsharp.core and fsharp.compiler.service to publish to nuget.

//cc @mmitche , @dsplaisted

### DIFF
--- a/eng/restore-toolset.sh
+++ b/eng/restore-toolset.sh
@@ -7,7 +7,7 @@ function InitializeCustomSDKToolset {
 
   # The following frameworks and tools are used only for testing.
   # Do not attempt to install them in source build.
-  if [[ "${DotNetBuildFromSource:-}" == "true" ]]; then
+  if [[ $properties == *"ArcadeBuildFromSource=true"* ]]; then
     return
   fi
 


jmarolf (approved):

dsplaisted:
The logic to put the FSharp NuGet packages in the library-packs folder probably needs to go in the dotnet/installer repo.  The target in this PR is copying them to a layout folder which we use to run tests in this repo, but the real product is build in the installer repo.  The installer repo consumes the `dotnet-toolset-internal-<version>.zip` file that is created in `artifacts\packages\<Configuration>\NonShipping` when you run `build -pack` on this repo.  That zip file includes the contents of the SDK folder, so it doesn't pick up the library-packs folder that's outside of it.

We also need to figure out what MSI is going to install these .nupkg files.  The MSI authoring for that may need to be done in the same repo that produces the .nupkg files.  If it's another repo (such as dotnet/installer) which consumes the .nupkgs and tries to wrap them in an MSI, then different builds of that repo will build different versions of the MSI trying to wrap the same version of the nupkg.  That caused problems with workload packs, though we may or may not have the same issues here.  /cc @joeloff 

Once these packages are in the library-packs folder, how is the version number supposed to flow through to F# projects so that they are referencing the same version that can be found in the library-packs folder?

KevinRansom:
@dsplaisted --- I wondered where the installer was, since I couldn't find it.  I will take a look.

KevinRansom:
@dsplaisted I updated this pr to produce a non-shipping package containing fsharp.core ... 
````
sdk\artifacts\packages\Release\NonShipping\dotnet-librarypack-internal-6.0.200-dev.zip
````

PR #23696, opened 2022-02-01
Don't split collect command into multiple collect commands (#22666)

fhnaseer:
* Don't split collect command into multiple collect commands,

--collect:"Code Coverage;arg1=val1" will be considered as 1 collect,

* Adding test for multiple collect commands,
also enabling tests for linux, code coverage is supported on linux,

* disabling tests on linux as some dependent packages are missing,

Co-authored-by: faisal <faisalhafeez@microsoft.com>

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
@@ -77,7 +77,7 @@ namespace Microsoft.DotNet.Cli
         public static readonly Option<IEnumerable<string>> CollectOption = new ForwardedOption<IEnumerable<string>>("--collect", LocalizableStrings.cmdCollectDescription)
         {
             ArgumentHelpName = LocalizableStrings.cmdCollectFriendlyName
-        }.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", o)}\"")
+        }.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", GetSemiColonEscapedArgs(o))}\"")
         .AllowSingleArgPerToken();
 
         public static readonly Option<bool> BlameOption = new ForwardedOption<bool>("--blame", LocalizableStrings.CmdBlameDescription)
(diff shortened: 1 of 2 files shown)


jakubch1 (approved):

Haplois (approved):

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -77,7 +77,7 @@ internal static class TestCommandParser
         public static readonly Option<IEnumerable<string>> CollectOption = new ForwardedOption<IEnumerable<string>>("--collect", LocalizableStrings.cmdCollectDescription)
         {
             ArgumentHelpName = LocalizableStrings.cmdCollectFriendlyName
-        }.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", o)}\"")
+        }.ForwardAsSingle(o => $"-property:VSTestCollect=\"{string.Join(";", GetSemiColonEscapedArgs(o))}\"")
```
Evangelink, line 80: I am investigating a bug where having commas can break the command line, could we have commas here?

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs
```diff
@@ -431,7 +431,7 @@ public void ItAcceptsNoLogoAsCliArguments()
             }
         }
 
-        [WindowsOnlyFact]
```
Evangelink, line 434: Out of curiosity, is this one deprecated?

fhnaseer: No. I was trying to run tests for Windows and Linux. But there were were issues like sdk was using older version of code coverage libs. So removed linux tests.

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
fhnaseer: I believe you mean commas instead of semi colon here. In our workflow we cannot.
Example:
--collect "Code Coverage;arg1=something,anything;arg2=val2
We can provide multiple settings (arg1, arg2) separated by semi colons. And each setting can have multiple values (arg1=something,anything).

Evangelink: My point was more to say, is it possible to have one of the collector name with a comma in its name (e.g. if that's done with F# through ```My collector, bla bla```.

Evangelink: So I double checked and there is no problem because the whole property is escaped with double quotes.

PR #23720, opened 2022-02-02
Ensure properties are escaped with double quotes

Evangelink:
Fixes #10423 and [AB#1469755](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1469755)

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.Linq;
@@ -188,6 +189,14 @@ namespace Microsoft.DotNet.Cli
         /// Adding double quotes around the property helps MSBuild arguments parser and avoid incorrect splits on ',' or ';'.
         /// </summary>
         private static string SurroundWithDoubleQuotes(string s)
-            => string.Concat("\"", s, "\"");
+        {
+            if (s.StartsWith("\"", StringComparison.Ordinal)
+                && s.EndsWith("\"", StringComparison.Ordinal))
+            {
+                return s;
+            }
+
+            return string.Concat("\"", s, "\"");
+        }
     }
 }
--- a/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
@@ -3,8 +3,6 @@
 
 using System.Collections.Generic;
 using System.CommandLine;
-using System.CommandLine.Invocation;
-using System.CommandLine.Parsing;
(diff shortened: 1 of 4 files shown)


src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -185,5 +183,11 @@ private static string[] GetSemiColonEscapedArgs(IEnumerable<string> args)
 
             return array;
         }
+
+        /// <summary>
+        /// Adding double quotes around the property helps MSBuild arguments parser and avoid incorrect splits on ',' or ';'.
+        /// </summary>
+        private static string SurroundWithDoubleQuotes(string s)
```
Evangelink, line 190: Feel free to suggest a better name.

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -25,7 +23,7 @@ internal static class TestCommandParser
         public static readonly Option<string> SettingsOption = new ForwardedOption<string>(new string[] { "-s", "--settings" }, LocalizableStrings.CmdSettingsDescription)
         {
             ArgumentHelpName = LocalizableStrings.CmdSettingsFile
-        }.ForwardAsSingle(o => $"-property:VSTestSetting={CommandDirectoryContext.GetFullPath(o)}");
+        }.ForwardAsSingle(o => $"-property:VSTestSetting={SurroundWithDoubleQuotes(CommandDirectoryContext.GetFullPath(o))}");
```
Evangelink, line 26: I am wondering if we should make a small builder function that would accept the category and the value (e.g. `BuildProperty("VSTestSetting", CommandDirectoryContext.GetFullPath(o))`), in this helper we could build all the common parts like `-property:XXX=YYY` and always escape the entry.

Evangelink: Nice! Any preference in doing so in this PR or in a follow-up?

MarcoRossignoli: `ForwardAsSingleMSBuildProperty(string propertyName,string propertyValue`) or similar?

Evangelink: I did a follow-up PR as there are a couple of classes impacted by this refactoring, PR (draft) can be found here #23730.

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
MarcoRossignoli, line 190: https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs#L441

we have `AddDoubleQuotes` as extension on our repo...but I think it's ok atm...also because will be superseded by new api, right?

Evangelink: It will be yes.

baronfel (approved):
Thanks for doing this cleanup, as well as adding tests 👍

PR #23748, opened 2022-02-03
Reverse versions in error message

lewing:
based on https://github.com/dotnet/sdk/blob/4122a50afa09e012a6a4eb094c410afbf84dd909/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/FXVersion.cs#L37-L43 it looks like this condition is reversed but I could be reading it wrong.  This check is failing however in a case I would expect it to work

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
@@ -142,9 +142,9 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
                     {
                         if (_manifests.TryGetValue(dependency.Key, out var resolvedDependency))
                         {
-                            if (FXVersion.Compare(dependency.Value, resolvedDependency.ParsedVersion) < 0)
+                            if (FXVersion.Compare(dependency.Value, resolvedDependency.ParsedVersion) > 0)
                             {
-                                throw new WorkloadManifestCompositionException(Strings.ManifestDependencyVersionTooLow, dependency.Key, dependency.Value, resolvedDependency.Version, manifest.Id, manifest.InformationalPath);
+                                throw new WorkloadManifestCompositionException(Strings.ManifestDependencyVersionTooLow, dependency.Key, resolveDependency.Version, dependency.Value, manifest.Id, manifest.InformationalPath);
                             }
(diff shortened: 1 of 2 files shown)


lewing:
reading closer it appears the check is ok but the error message has the strings reversed

lewing:
see https://github.com/xamarin/xamarin-android/pull/6698

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
```diff
@@ -144,7 +144,7 @@ private void ComposeWorkloadManifests()
                         {
                             if (FXVersion.Compare(dependency.Value, resolvedDependency.ParsedVersion) > 0)
                             {
-                                throw new WorkloadManifestCompositionException(Strings.ManifestDependencyVersionTooLow, dependency.Key, dependency.Value, resolvedDependency.Version, manifest.Id, manifest.InformationalPath);
+                                throw new WorkloadManifestCompositionException(Strings.ManifestDependencyVersionTooLow, dependency.Key, resolveDependency.Version, dependency.Value, manifest.Id, manifest.InformationalPath);
```
joeloff, line 147: should be `resolved`, missing a "d"

joeloff (approved):

joeloff:
There might be tests lurking that evaluate the error messages - I don't know the resolver code that well

filipnavara (approved):

dsplaisted:
Looks like there's one test which needs to be updated: `ManifestReaderTests.ManifestTests.ItChecksDependencies`

dsplaisted (approved):

akoeplinger (approved):

lewing:
What's going on with the red lane? I don't see any failures

dsplaisted:
@marcpopMSFT This seems to be failing running tests in Helix in the same way that we thought #23284 fixed

akoeplinger:
Can we get this merged? I just hit this again today and it is pretty confusing :)

dsplaisted:
We should also check this in to 6.0.3xx.  We could either retarget this PR or merge this PR (once passing) and cherry-pick the changes to 6.0.3xx.

dsplaisted:
@gkulin Could you backport this to 6.0.3xx?  I can work with you on it if needed.

PR #23830, opened 2022-02-08
Enable launch profile arguments with dotnet run

### LINKED ISSUE #23565: dotnet cli doesn't seem to respect profile / launch file / command line arguments
I added launchSettings via VS2022 GUI

obraz

it looks like that:

	{
	  "profiles": {
		"LaunchSettingsQuirks": {
		  "commandName": "Project",
		  "commandLineArgs": "--test"
		}
	  }
	}

when I run program via VS with that profile then it prints

	args:
	--test

but when I run it via 

`dotnet run` or 

`dotnet run --launch-profile LaunchSettingsQuirks` or

` dotnet run --launch-profile "LaunchSettingsQuirks"`

 then it doesn't, it just prints 

	args:

when I use name of the profile that does not exist, then the error is displayed correctly

	dotnet run --launch-profile test
	The launch profile "test" could not be applied.
	A profile with the specified name could not be found or is not a valid JSON object.

Repo: https://github.com/Swiftly1/LaunchSettingsQuirks

	dotnet --version
	6.0.101

marcpopMSFT:
Fixes #23565
I specifically only set the arguments if there aren't arguments already so the user can override from the CLI

### DIFF
--- a/src/Assets/TestProjects/TestAppWithLaunchSettings/Program.cs
+++ b/src/Assets/TestProjects/TestAppWithLaunchSettings/Program.cs
@@ -11,6 +11,14 @@ namespace ConsoleApplication
         {
             Console.WriteLine("Hello world");
             Console.WriteLine($"MyCoolEnvironmentVariableKey={Environment.GetEnvironmentVariable("MyCoolEnvironmentVariableKey")}");
+            if (args.Length > 0)
+            {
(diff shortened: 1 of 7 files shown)


baronfel:
Tested locally, works great 👍

fizietechD (commented):
Test

PR #23882, opened 2022-02-11
Fix cli translation issue in Strings.zh-Hans.xlf

Cai1Hsu:
The Chinese translation of NETSDK1179:  `warning NETSDK1179: 使用“--runtime”时，必需选择“--自包含”或“--非自包含”选项。(NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.)`. It seems that the parameters `--self-contained` and `--no-self-contained` were translated into `--自包含` and `--非自包含` by mistake. This PR fixes the issue.


marcpopMSFT:
@richaverma1 @cristianosuzuki77 I recall that we are supposed to file a translation issue and get the underlying translation resolved rather than accept direct community updates. Is that still the process?

richaverma1:
@marcpopMSFT yes that is correct

marcpopMSFT:
@Cai1Hsu Thanks for reporting this. I've filed an internal ticket with our translation team to correct the translation based on your recommendation. There is an internal copy of all our localization so the strings need to be fixed there.

Including that link for internal folks who want to follow: https://ceapex.visualstudio.com/CEINTL/_workitems/edit/569469

PR #23934, opened 2022-02-15
reenable composite r2r singlefile test on OSX

VSadov:
The test was disabled on OSX since it caused random crashes with unactionable dumps. We kept the test enabled on Linux/Windows to see if we get dumps that are more debuggable. It looks like the test has been stable so the issues if they are still present are OSX-specific.  
It is also possible that the OSX instabilities were related to issues fixed as a part of cleanups in https://github.com/dotnet/runtime/pull/61938

Either way we should reenable the test.

Re: https://github.com/dotnet/runtime/issues/60308


VSadov:
I do not think the failure on Windows_NT_FullFramework is caused by this change, since the change is OSX-specific.

VSadov:
rebased onto recent main

VSadov:
This is a test-only change.

PR #23973, opened 2022-02-16
Annotate content items for web projects with a special attribute

pranavkm:
Contributes to https://github.com/dotnet/maui/issues/791

### DIFF
--- a/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
+++ b/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
@@ -108,7 +108,18 @@ Integration with static web assets:
 
   <ItemGroup>
     <Content Remove="@(ScopedCssInput)" />
-    <Content Include="@(ScopedCssInput)" Pack="false" CopyToPublishDirectory="Never" />
+    <!--
+      See https://github.com/dotnet/maui/issues/791 for details.
+      PublishFolderType="None" prevents content items frrom being bundled as part of MAUI iOS targets.
+      ExcludeFromContentCheck="true" prevents content items from being inspected by MAUI Android targets for warnings.
+    -->
+
+    <Content
+      Include="@(ScopedCssInput)"
+      Pack="false"
+      CopyToPublishDirectory="Never"
+      PublishFolderType="None"
+      ExcludeFromContentCheck="true" />
   </ItemGroup>
 </Target>
 
--- a/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
+++ b/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
@@ -103,12 +103,12 @@ Integration with static web assets:
   </DiscoverDefaultScopedCssItems>
 
   <ItemGroup Condition="'$(EnableDefaultScopedCssItems)' == 'true'">
-    <ScopedCssInput Include="@(_DiscoveredScopedCssInputs)" Pack="false" CopyToPublishDirectory="Never" />
+    <ScopedCssInput Include="@(_DiscoveredScopedCssInputs)" />
   </ItemGroup>
 
   <ItemGroup>
-    <!-- Configure Pack="false" and CopyToPublishDirectory="Never" for Content items that are ScopedCssInputs. -->
-    <Content Pack="%(ScopedCssInput.Pack)" CopyToPublishDirectory="%(ScopedCssInput.CopyToPublishDirectory)" />
+    <Content Remove="@(ScopedCssInput)" />
+    <Content Include="@(ScopedCssInput)" Pack="false" CopyToPublishDirectory="Never" />
   </ItemGroup>
 </Target>
 
--- a/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
+++ b/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
@@ -103,23 +103,12 @@ Integration with static web assets:
   </DiscoverDefaultScopedCssItems>
 
   <ItemGroup Condition="'$(EnableDefaultScopedCssItems)' == 'true'">
-    <ScopedCssInput Include="@(_DiscoveredScopedCssInputs)" />
+    <ScopedCssInput Include="@(_DiscoveredScopedCssInputs)" Pack="false" CopyToPublishDirectory="Never" />
   </ItemGroup>
 
   <ItemGroup>
-    <Content Remove="@(ScopedCssInput)" />
-    <!--
-      See https://github.com/dotnet/maui/issues/791 for details.
-      PublishFolderType="None" prevents content items frrom being bundled as part of MAUI iOS targets.
-      ExcludeFromContentCheck="true" prevents content items from being inspected by MAUI Android targets for warnings.
-    -->
-
-    <Content
-      Include="@(ScopedCssInput)"
-      Pack="false"
-      CopyToPublishDirectory="Never"
-      PublishFolderType="None"
-      ExcludeFromContentCheck="true" />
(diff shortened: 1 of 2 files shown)


src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
```diff
@@ -108,7 +108,18 @@ Integration with static web assets:
...
+      PublishFolderType="None" prevents content items frrom being bundled as part of MAUI iOS targets.
+      ExcludeFromContentCheck="true" prevents content items from being inspected by MAUI Android targets for warnings.
+    -->
+
+    <Content
+      Include="@(ScopedCssInput)"
+      Pack="false"
+      CopyToPublishDirectory="Never"
+      PublishFolderType="None"
+      ExcludeFromContentCheck="true" />
```
javiercn, line 122: I'm a bit uneasy about having this check here. I think the WebContentItem is ok, but I rather have something on the Maui targets than add something Maui specific to the Razor SDK.

javiercn:
I think some changes might also need to be applied to JS Modules similar to what we do for scoped CSS

src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
pranavkm: I can play around with this - what happens today is that .razor.css start off as being `None` but get transformed to `Content` because of this addition. I wanted to avoid  relying on the target order between Razor's and Xamarin's targets since it might change with either of us knowing and break in subtle ways. 

One option might be to update .razor.css during global evaluation if the feature is enabled, but let me try it out to see how it works.

javiercn: > I can play around with this - what happens today is that .razor.css start off as being None but get transformed to Content because of this addition. I wanted to avoid relying on the target order between Razor's and Xamarin's targets since it might change with either of us knowing and break in subtle ways.

If there is a concrete target on the Maui SDK we can hook into, we can have this logic contained within our targets in Maui.

src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
```diff
@@ -103,12 +103,12 @@ Integration with static web assets:
...
   <ItemGroup Condition="'$(EnableDefaultScopedCssItems)' == 'true'">
-    <ScopedCssInput Include="@(_DiscoveredScopedCssInputs)" />
+    <ScopedCssInput Include="@(_DiscoveredScopedCssInputs)" Pack="false" CopyToPublishDirectory="Never" />
   </ItemGroup>
 
   <ItemGroup>
-    <Content Remove="@(ScopedCssInput)" />
-    <Content Include="@(ScopedCssInput)" Pack="false" CopyToPublishDirectory="Never" />
+    <!-- Configure Pack="false" and CopyToPublishDirectory="Never" for Content items that are ScopedCssInputs. -->
+    <Content Pack="%(ScopedCssInput.Pack)" CopyToPublishDirectory="%(ScopedCssInput.CopyToPublishDirectory)" />
```
pranavkm, line 111: @javiercn this seemingly works: https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-items?view=vs-2022#updating-metadata-on-items-in-an-itemgroup-of-a-target. What do you think?

KalleOlaviNiemitalo: demo.msbuildproj

```XML
<Project>
  <ItemGroup>
    <Content Include="A.txt; B.css; C.css" />
    <ScopedCssInput Include="B.css; C.css" Pack="false" CopyToPublishDirectory="Never" />
  </ItemGroup>

  <Target Name="Build">
    <ItemGroup>
      <Content Pack="%(ScopedCssInput.Pack)" CopyToPublishDirectory="%(ScopedCssInput.CopyToPublishDirectory)" />
    </ItemGroup>

    <Message Text="%(Content.Identity): Pack='%(Content.Pack)' CopyToPublishDirectory='%(Content.CopyToPublishDirectory)'" />
  </Target>
</Project>
```

dotnet msbuild demo.msbuildproj -verbosity:detailed

```
Microsoft (R) Build Engine version 17.1.0+ae57d105c for .NET

Build started 23.2.2022 0.00.42.
     0>Process = "C:\Program Files\dotnet\dotnet.exe"
       MSBuild executable path = "C:\Program Files\dotnet\sdk\6.0.200\MSBuild.dll"
       Command line arguments = ""C:\Program Files\dotnet\sdk\6.0.200\dotnet.dll" msbuild demo.msbuildproj -verbosity:detailed"
       Current directory = "C:\Users\Kalle\Desktop"
       MSBuild version = "17.1.0+ae57d105c"
     1>Project "C:\Users\Kalle\Desktop\demo.msbuildproj" on node 1 (default targets).
     1>Target "Build" in project "C:\Users\Kalle\Desktop\demo.msbuildproj" (entry point):
     1>Using "Message" task from assembly "Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
       Task "Message"
         A.txt: Pack='false' CopyToPublishDirectory='Never'
       Done executing task "Message".
       Task "Message"
         B.css: Pack='false' CopyToPublishDirectory='Never'
       Done executing task "Message".
       Task "Message"
         C.css: Pack='false' CopyToPublishDirectory='Never'
       Done executing task "Message".
     1>Done building target "Build" in project "demo.msbuildproj".
     1>Done Building Project "C:\Users\Kalle\Desktop\demo.msbuildproj" (default targets).

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.15
```

It applies `Pack="%(ScopedCssInput.Pack)" CopyToPublishDirectory="%(ScopedCssInput.CopyToPublishDirectory)"` even to `A.txt` even though that does not match any `ScopedCssInput` item.

KalleOlaviNiemitalo: I thought that could be fixed by adding `Update="@(ScopedCssInput)"` or `Update="%(ScopedCssInput.Identity)"`, but no, MSBuild ignores that attribute. [This attribute is valid only if it's specified for an item in an `ItemGroup` that is not in a `Target`.](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-element-msbuild?view=vs-2019) Odd that there is not even a warning.

KalleOlaviNiemitalo: Possibly `Condition="'%(Content.Identity)' == '%(ScopedCssInput.Identity)'"` could work, but if it does, I think it will cost quadratic time.

KalleOlaviNiemitalo: If everything else fails, you can just define a custom task for this.

pranavkm: Thanks for catching this @KalleOlaviNiemitalo. I tried the positive cases, and was mislead by its apparent success. Adding a condition doesn't work and it'd be easier to resolve this in the MAUI targets where these items are consumed.

javiercn: @pranavkm the way I've done this in the past involves removing the items and re-adding them.

src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.ScopedCss.targets
javiercn: Should this be marked with `WebContentItem="true"` so that we can identify it on a Blazor Maui target?

pranavkm:
Superseded by https://github.com/dotnet/maui/pull/4861

PR #23977, opened 2022-02-16
Update 6.0.3xx with the 6.0.200 version of files

marcpopMSFT:


eerhardt:
Is this PR going forward? It looks like bugs that were fixed in dotnet/linker in 6.0.2xx SDK aren't in the latest 6.0.3xx SDK versions.

cc @jonathanpeppers @sbomer @vitek-karas

src/Tasks/Common/Resources/xlf/Strings.it.xlf
```diff
@@ -224,7 +224,7 @@
       </trans-unit>
       <trans-unit id="CompressionInSingleFileRequiresSelfContained">
         <source>NETSDK1176: Compression in a single file bundle is only supported when publishing a self-contained application.</source>
-        <target state="translated">NETSDK1176: la compressione in un unico bundle di file è supportata solo quando si esegue la pubblicazione di un'applicazione indipendente.</target>
+        <target state="new">NETSDK1176: Compression in a single file bundle is only supported when publishing a self-contained application.</target>
```
baronfel, line 227: I think we need to revert all of these translations changes before merge

baronfel (dismissed):

src/Tasks/Common/Resources/xlf/Strings.it.xlf
baronfel: reverted those changes.

baronfel:
@joeloff has concerns about the darc-int packages too that we'll need to talk about tomorrow.

v-wuzhai (approved):

PR #23988, opened 2022-02-17
Fix escaping function when argument ends with odd number of backslashes

### LINKED ISSUE #23970: Trailing \ or / will fail for path parameters for `dotnet test`
Failing sample
```cs
dotnet test C:\git\localPlayground\TestProject\ --results-directory:C:\git\localPlayground\TestProject\
...
The path 'C:\git\localPlayground\TestProject"' specified in the 'ResultsDirectory' is invalid. Error: The filename, directory name, or volume label syntax is incorrect. : 'C:\git\localPlayground\TestProject"'
```

Evangelink:
Fixes #23970


src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs
```diff
@@ -750,6 +750,51 @@ public void EnsureOutputPathEscaped(string flag)
...
+        [InlineData("--output", "/")]
+        [InlineData("--output", "///")]
+        [InlineData("--diag", "\\")]
+        [InlineData("--diag", "\\\\\\")]
+        [InlineData("--diag", "/")]
+        [InlineData("--diag", "///")]
+        [InlineData("--results-directory", "\\")]
+        [InlineData("--results-directory", "\\\\\\")]
+        [InlineData("--results-directory", "/")]
+        [InlineData("--results-directory", "///")]
```
MarcoRossignoli, line 779: nit: empty test and null test

MarcoRossignoli (approved):

MarcoRossignoli (commented):
some comments

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs
```diff
@@ -750,6 +750,51 @@ public void EnsureOutputPathEscaped(string flag)
...
+            CommandResult result = new DotnetTestCommand(Log)
+                .WithWorkingDirectory(testProjectDirectory)
+                .Execute(flag, Path.Combine(testProjectDirectory, Guid.NewGuid().ToString()) + slashOrBackslash);
+
+            // Verify
+            if (!TestContext.IsLocalized())
+            {
+                result.StdOut.Should().Contain("Total:     2");
+                result.StdOut.Should().Contain("Passed:     1");
+                result.StdOut.Should().Contain("Failed:     1");
```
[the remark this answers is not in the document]
Evangelink: I am now also testing that the specified directory exists and contains items.

src/Tests/dotnet-test.Tests/TestCommandParserTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.NET.TestFramework;
+using Microsoft.NET.TestFramework.Assertions;
+using Microsoft.NET.TestFramework.Commands;
+using Microsoft.NET.TestFramework.Utilities;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Cli.Test.Tests
+{
+    public class TestCommandParserTests
```
Evangelink, line 20: @MarcoRossignoli I have added specific unit tests in addition to integration tests.

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -188,15 +188,40 @@ private static string[] GetSemiColonEscapedArgs(IEnumerable<string> args)
         /// <summary>
         /// Adding double quotes around the property helps MSBuild arguments parser and avoid incorrect splits on ',' or ';'.
         /// </summary>
-        private static string SurroundWithDoubleQuotes(string s)
+        internal /* for testing purposes */ static string SurroundWithDoubleQuotes(string input)
         {
-            if (s.StartsWith("\"", StringComparison.Ordinal)
-                && s.EndsWith("\"", StringComparison.Ordinal))
+            if (input == null)
```
MarcoRossignoli, line 193:
```suggestion
            if (input is null)
```

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -188,15 +188,40 @@ private static string[] GetSemiColonEscapedArgs(IEnumerable<string> args)
...
             {
-                return s;
+                throw new ArgumentNullException(nameof(input));
```
MarcoRossignoli, line 195: throw or return null? 🤔

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs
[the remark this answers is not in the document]
MarcoRossignoli: great thx

MarcoRossignoli (approved):

MarcoRossignoli:
@Evangelink  it's failing in some distributions

MarcoRossignoli (changes requested):
some tests are failing

MarcoRossignoli:
@Evangelink  pls feel free to fix this one once we're good with the implementation https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs#L69

Evangelink:
@KalleOlaviNiemitalo Thank you for your feedback on the ticket and here!

I have discussed with the team and we believe it's better to keep the "add quotes" solution because of the various paths that could be taken (depending for example if you pass a project/sln or dll/exe - the latter doesn't get through msbuild) so some escape/unescape sequences might not happen. We know that the process is complex and there are lots of jumps between escape/unescape but sadly we are not able to simplify this much without introducing potential breaking changes.

> Does MSBuild itself parse quotation marks, or are those parsed only by the CLR when it translates the Windows command-line string to an array of strings? If MSBuild does not parse them, then perhaps TestCommand.SurroundWithDoubleQuotes should be replaced with something that percent-encodes commas, semicolons, and other problematic characters for MSBuild. And the quotation-mark complexity would be in ArgumentEscaper only.

MSBuild is also doing specific parsing of the quotes. On the original PR, when trying to fix an issue with a comma not escaped, we considered doing an encoding of all the special characters or to use the quotes solution. It felt that the quotes was the easiest/more global solution. Again as said in the previous paragraph, there are different paths taken with different escaping/unescaping mechanism in place which is not the easiest to follow.

MarcoRossignoli (approved):

Issue #21811, opened 2021-10-06
Support workloads installed via Visual Studio from CLI `dotnet workload` commands


dsplaisted:
We'd like (at a minimum) the following commands to have visibility into the workloads that were installed by Visual Studio and handle them appropriately.

- `dotnet workload update`
- `dotnet workload list`

Right now the CLI can't see which workloads were installed via VS.  We could:

- Read VS catalog (expensive and not recommended)
- Look at installed packs and infer installed workloads based on that
- Have VS write workload installation records that the CLI can read (probably via a custom MSI that is only installed by VS)

dsplaisted:
#24005 fixed this for `dotnet workload list`

For `dotnet workload update`, we want to get the list of workloads installed by VS before running the update, and then install the new versions of those workloads after updating the manifests.  This will fix the issue where workload update can update manifests and break workloads that were installed by Visual Studio.

dsplaisted:
Update: We want to try to do this in the next feature release, to fix the following:

- Install VS with workload
- Updated workload is released
- Run `dotnet workload update` from command-line
- Workload is no longer installed (and templates don't show up, for example)

This is because `dotnet workload update` will update the workload manifests, but it doesn't see the workloads installed by Visual Studio, so it doesn't install the corresponding packs.

We've already addressed this for `dotnet workload list` so applying the same logic to `dotnet workload update` should not be too hard.

baronfel:
We should also update `dotnet workload uninstall` to have an error message specifically for when a user uninstalles a VS-managed workload with no CLI install manifest. In this case, we should tell the user that they need to uninstall the workload via the VS Installer, and that the CLI command they run had no effect.

marcpopMSFT:
Besides the update command, we'll want to keep an eye out for any other command that results in updated manifests (I'm thinking of install and restore) as anything that updates manifests should update all installed workloads rather than letting them fail.

nagilson:
Should we close this now that everything here is done, or are there more interactions we'd like to support?

We havent added the uninstall message @baronfel mentioned eariler yet I think.

baronfel:
@nagilson let me extract that to a separate issue real quick then we can close this one.

nagilson:
Thank you!

PR #24005, opened 2022-02-18
Add Visual Studio workloads to dotnet workload list

joeloff:
Add support to include workloads installed by Visual Studio if the SDK's feature band matches the feature band of an SDK installed by Visual Studio.

Partially address #21811


joeloff:
@marcpopMSFT @dsplaisted The second part of #21811 relates to `dotnet workload update`. While we can find workloads installed by VS, we have no easy way to update them from the CLI since we can never touch the VS catalog and we don't really have any insights into the more complicated toplevel components that bring in emulators. The only way to know if there's an update available would be to check if a higher compatible version of VS for the given channel with higher versioned components has been published.

src/Cli/dotnet/commands/dotnet-workload/list/InstalledWorkloadsCollection.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.NET.Sdk.WorkloadManifestReader;
+
+namespace Microsoft.DotNet.Workloads.Workload.List
+{
+    internal class InstalledWorkloadsCollection : Dictionary<string, string>
```
dsplaisted, line 10: Can we encapsulate instead of inheriting here?  The override to `Add` makes this work differently than I would expect a Dictionary to work.

src/Cli/dotnet/dotnet.csproj
```diff
@@ -95,8 +95,17 @@
     <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
     <PackageReference Include="Microsoft.Deployment.DotNet.Releases" Version="$(MicrosoftDeploymentDotNetReleasesVersion)" />
     <PackageReference Include="System.ServiceProcess.ServiceController" Version="$(SystemServiceProcessServiceControllerVersion)" />
+    <PackageReference Include="Microsoft.VisualStudio.Setup.Configuration.Interop" Version="$(MicrosoftVisualStudioSetupConfigurationInteropVersion)" PrivateAssets="All" ExcludeAssets="Runtime"/>
```
dsplaisted, line 98: Looks like we embed the VS Interop assemblies, so this reference doesn't add any other new dependencies to the output?

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -115,17 +116,20 @@ public override int Execute()
             }
             else
             {
+                InstalledWorkloadsCollection www = new(installedList, $"SDK {_currentSdkFeatureBand}");
```
dsplaisted, line 119: Nit: Why `www` for a variable name?

src/Cli/dotnet/commands/dotnet-workload/list/VisualStudioWorkloads.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// Finds all workloads installed by all Visual Studio instances given that the
+        /// SDK installed by an instance matches the feature band of the currently executing SDK.
+        /// </summary>
+        /// <param name="workloadResolver">The workload resolver used to obtain available workloads.</param>
+        /// <param name="sdkFeatureBand">The feature band of the executing SDK.</param>
+        /// <param name="installedWorkloads">The collection of installed workloads to update.</param>
+        internal static void GetInstalledWorkloads(IWorkloadResolver workloadResolver, SdkFeatureBand sdkFeatureBand,
+            ref InstalledWorkloadsCollection installedWorkloads)
```
dsplaisted, line 61: As far as I can see, there's no need to pass in `installedWorkloads` via `ref`.  The method modifies the collection but doesn't change the reference passed in.

dsplaisted (approved):
Looks good.  I'd suggest changing the InstalledWorkloadsCollection data structure though.

dsplaisted:
As for `dotnet workload update`, what happens today if you install a workload via VS and then try to update to a newer version with `dotnet workload update`?  We should figure out what the behavior is and what it should be.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
joeloff: Thanks, I thought I renamed it. I used a temp name when I couldn't come up with something good, but thought I renamed everything. I'll fix this.

src/Cli/dotnet/dotnet.csproj
joeloff: It shouldn't AFAIK

joeloff:
> As for `dotnet workload update`, what happens today if you install a workload via VS and then try to update to a newer version with `dotnet workload update`? We should figure out what the behavior is and what it should be.

Today, even with my change it won't find the workloads, so it won't offer anything. We could expose the workload set returned by the new logic, but VS does have higher level components that include items like the emulators/designers that are not part of the CLI based installation. We should discuss this more - in some cases this may be fine, but I can't be 100% sure that it won't impact VS scenarios because we don't update the catalogs in VS

PR #24012, opened 2022-02-18
Add artifacts post processing

MarcoRossignoli:
* Add artifacts post processing(disabled by default until https://github.com/microsoft/vstest/pull/3384 will flow here)
* Add feature flag feat
* Add tracing feature

contributes to https://github.com/microsoft/vstest/issues/1811

### DIFF
--- a/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
+++ b/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
@@ -7,4 +7,10 @@
   <ItemGroup>
 	<PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="$(MicrosoftNETTestSdkPackageVersion)" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!-- Microsoft.NET.Test.Sdk package includes source files which shouldn't be automatically included. -->
+    <!-- Excluding those -->
+    <Compile Remove="pkgs\Microsoft.NET.Test.Sdk\**" />
+  </ItemGroup>
 </Project>
--- a/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
+++ b/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
@@ -4,6 +4,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-	<PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="17.2.0-preview-20220131-20" />
+        <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="17.2.0-preview-20220131-20" />
   </ItemGroup>
 </Project>
--- a/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
+++ b/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+	<PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="17.2.0-preview-20220131-20" />
+  </ItemGroup>
+</Project>
--- a/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
+++ b/src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
@@ -1,9 +1,10 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), testAsset.props))\testAsset.props" />
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
-        <PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="17.2.0-preview-20220131-20" />
+	<PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="$(MicrosoftNETTestSdkPackageVersion)" />
   </ItemGroup>
 </Project>
--- a/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
+++ b/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
@@ -1,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="2.2.7" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.2.7" />
+    <PackageReference Include="coverlet.collector" Version="3.1.0" />
+  </ItemGroup>
+
+</Project>
--- a/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
+++ b/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
@@ -2,9 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
-
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
--- a/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
+++ b/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
@@ -1,15 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), testAsset.props))\testAsset.props" />
 
   <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
-    <IsPackable>false</IsPackable>
+    <TargetFramework>$(CurrentTargetFramework)</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
-    <PackageReference Include="MSTest.TestAdapter" Version="2.2.7" />
-    <PackageReference Include="MSTest.TestFramework" Version="2.2.7" />
-    <PackageReference Include="coverlet.collector" Version="3.1.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="$(MSTestVersion)" />
+    <PackageReference Include="MSTest.TestAdapter" Version="$(MSTestVersion)" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPackageVersion)" />
   </ItemGroup>
-
 </Project>
--- a/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
+++ b/src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
@@ -10,4 +10,10 @@
     <PackageReference Include="MSTest.TestAdapter" Version="$(MSTestVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPackageVersion)" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!-- Microsoft.NET.Test.Sdk package includes source files which shouldn't be automatically included. -->
+    <!-- Excluding those -->
+    <Compile Remove="pkgs\Microsoft.NET.Test.Sdk\**" />
+  </ItemGroup>
 </Project>
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -194,7 +194,7 @@ namespace Microsoft.DotNet.Tools.Test
             {
                 if (Directory.Exists(expectedArtifactDirectory))
                 {
-                    VSTestTrace.WriteTrace($"Cleaning artifact directory '{expectedArtifactDirectory}'");
+                    VSTestTrace.WriteTrace($"Cleaning artifact directory '{expectedArtifactDirectory}'.");
                     try
                     {
                         Directory.Delete(expectedArtifactDirectory, true);
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -201,7 +201,7 @@ namespace Microsoft.DotNet.Tools.Test
                     }
                     catch (Exception ex)
                     {
-                        VSTestTrace.WriteTrace($"Exception during artifact cleanup:'{ex}'");
+                        VSTestTrace.WriteTrace($"Exception during artifact cleanup: {ex}");
                     }
                 }
             }
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -103,7 +103,7 @@ namespace Microsoft.DotNet.Tools.Test
                 string commandLineParameters = "";
                 if (args?.Length > 0)
                 {
-                    commandLineParameters = args.Aggregate((a, b) => $"{a}|{b}");
+                    commandLineParameters = args.Aggregate((a, b) => $"{a} | {b}");
                 }
                 VSTestTrace.WriteTrace($"Argument list: '{commandLineParameters}'");
             }
@@ -179,16 +179,16 @@ namespace Microsoft.DotNet.Tools.Test
 
             VSTestTrace.WriteTrace($"Artifacts found inside '{expectedArtifactDirectory}', running post-processing.");
 
-            var artifactsProcessingModePostprocess = new List<string> { "--artifactsProcessingMode-postprocess", $"--testSessionCorrelationId:{testSessionCorrelationId}" };
+            var artifactsPostProcessArgs = new List<string> { "--artifactsProcessingMode-postprocess", $"--testSessionCorrelationId:{testSessionCorrelationId}" };
 
             if (parseResult.HasOption(TestCommandParser.DiagOption))
             {
-                artifactsProcessingModePostprocess.Add($"--diag:{parseResult.GetValueForOption(TestCommandParser.DiagOption)}");
+                artifactsPostProcessArgs.Add($"--diag:{parseResult.GetValueForOption(TestCommandParser.DiagOption)}");
             }
 
             try
             {
-                return new VSTestForwardingApp(artifactsProcessingModePostprocess).Execute();
+                return new VSTestForwardingApp(artifactsPostProcessArgs).Execute();
             }
             finally
             {
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -22,7 +22,92 @@ namespace Microsoft.DotNet.Tools.Test
         {
         }
 
-        public static TestCommand FromParseResult(ParseResult result, string[] settings, string testSessionCorrelationId, string msbuildPath = null)
+        public static int Run(ParseResult parseResult)
+        {
+            parseResult.HandleDebugSwitch();
+
+            FeatureFlag.Default.PrintFlagFeatureState();
+
+            // We use also current process id for the correlation id for possible future usage in case we need to know the parent process
+            // from the VSTest side.
+            string testSessionCorrelationId = $"{Environment.ProcessId}_{Guid.NewGuid()}";
+
+            string[] args = parseResult.GetArguments();
+
+            if (VSTestTrace.TraceEnabled)
+            {
+                string commandLineParameters = "";
+                if (args?.Length > 0)
+                {
+                    commandLineParameters = args.Aggregate((a, b) => $"{a} | {b}");
+                }
+                VSTestTrace.SafeWriteTrace(() => $"Argument list: '{commandLineParameters}'");
+            }
+
+            // settings parameters are after -- (including --), these should not be considered by the parser
+            string[] settings = args.SkipWhile(a => a != "--").ToArray();
+            // all parameters before --
+            args = args.TakeWhile(a => a != "--").ToArray();
+
+            // Fix for https://github.com/Microsoft/vstest/issues/1453
+            // Run dll/exe directly using the VSTestForwardingApp
+            if (ContainsBuiltTestSources(args))
+            {
+                return ForwardToVSTestConsole(parseResult, args, settings, testSessionCorrelationId);
+            }
+
+            return ForwardToMsbuild(parseResult, settings, testSessionCorrelationId);
+        }
+
+        private static int ForwardToMsbuild(ParseResult parseResult, string[] settings, string testSessionCorrelationId)
+        {
+            // Workaround for https://github.com/Microsoft/vstest/issues/1503
+            const string NodeWindowEnvironmentName = "MSBUILDENSURESTDOUTFORTASKPROCESSES";
+            string previousNodeWindowSetting = Environment.GetEnvironmentVariable(NodeWindowEnvironmentName);
+            try
+            {
+                Environment.SetEnvironmentVariable(NodeWindowEnvironmentName, "1");
+                int exitCode = FromParseResult(parseResult, settings, testSessionCorrelationId).Execute();
+
+                // We run post processing also if execution is failed for possible partial successful result to post process.
+                exitCode |= RunArtifactPostProcessingIfNeeded(testSessionCorrelationId, parseResult, FeatureFlag.Default);
+
+                return exitCode;
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(NodeWindowEnvironmentName, previousNodeWindowSetting);
+            }
+        }
+
+        private static int ForwardToVSTestConsole(ParseResult parseResult, string[] args, string[] settings, string testSessionCorrelationId)
+        {
+            List<string> convertedArgs = new VSTestArgumentConverter().Convert(args, out List<string> ignoredArgs);
+            if (ignoredArgs.Any())
+            {
+                Reporter.Output.WriteLine(string.Format(LocalizableStrings.IgnoredArgumentsMessage, string.Join(" ", ignoredArgs)).Yellow());
+            }
+
+            // merge the args settings, we don't need to escape
+            // one more time, there is no extra hop via msbuild
+            convertedArgs.AddRange(settings);
+
+            if (FeatureFlag.Default.IsEnabled(FeatureFlag.ARTIFACTS_POSTPROCESSING))
+            {
+                // Add artifacts processing mode and test session id for the artifact post-processing
+                convertedArgs.Add("--artifactsProcessingMode-collect");
+                convertedArgs.Add($"--testSessionCorrelationId:{testSessionCorrelationId}");
+            }
+
+            int exitCode = new VSTestForwardingApp(convertedArgs).Execute();
+
+            // We run post processing also if execution is failed for possible partial successful result to post process.
+            exitCode |= RunArtifactPostProcessingIfNeeded(testSessionCorrelationId, parseResult, FeatureFlag.Default);
+
+            return exitCode;
+        }
+
+        private static TestCommand FromParseResult(ParseResult result, string[] settings, string testSessionCorrelationId, string msbuildPath = null)
         {
             result.ShowHelpOrErrorIfAppropriate();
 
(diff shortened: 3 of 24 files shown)


src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
```
Evangelink, line 4: Is there a need to test multiple tfm?

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -70,27 +79,45 @@ public static TestCommand FromParseResult(ParseResult result, string[] settings,
             if (!hasRootVariable)
             {
                 testCommand.EnvironmentVariable(rootVariableName, rootValue);
+                VSTestTrace.WriteTrace($"Root variable set {rootVariableName}:{rootValue}");
             }
 
+            VSTestTrace.SafeWriteTrace(() => $"Starting test using MSBuild with arguments '{testCommand.GetArgumentsToMSBuild()}' custom MSBuild path '{msbuildPath}' norestore '{noRestore}'");
```
Evangelink, line 85: Why do you use `SafeWriteTrace` but regular `WriteTrace` in the previous if?

Evangelink, line 85: Could it be worth to split the line? Something like:

```
Starting test using MSBuild with
- arguments: `testCommand.GetArgumentsToMSBuild()`
- MSBuild path: `msbuildPath`
- norestore: `noRestore`
```

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -100,7 +127,19 @@ public static int Run(ParseResult parseResult)
                 // one more time, there is no extra hop via msbuild
                 convertedArgs.AddRange(settings);
 
-                return new VSTestForwardingApp(convertedArgs).Execute();
+                if (FeatureFlag.Default.IsEnabled(FeatureFlag.ARTIFACTS_POSTPROCESSING))
```
Evangelink, line 130: I am not sure why we need to add twice these arguments? Here and into msbuild args.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -110,17 +149,67 @@ public static int Run(ParseResult parseResult)
...
+
+            // VSTest runner will save artifacts inside a temp folder if needed.
+            string expectedArtifactDirectory = Path.Combine(Path.GetTempPath(), testSessionCorrelationId);
+            if (!Directory.Exists(expectedArtifactDirectory))
+            {
+                VSTestTrace.WriteTrace("No artifact found, post-processing won't run.");
+                return 0;
+            }
+
+            VSTestTrace.WriteTrace($"Artifacts found inside '{expectedArtifactDirectory}', running post-processing.");
```
Evangelink, line 180: It feels weird to say we found artefacts in the in the folder while we only checked that folder exists. So if that's empty the message is a little weird.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -110,17 +149,67 @@ public static int Run(ParseResult parseResult)
...
+
+            try
+            {
+                return new VSTestForwardingApp(artifactsProcessingModePostprocess).Execute();
+            }
+            finally
+            {
+                if (Directory.Exists(expectedArtifactDirectory))
+                {
+                    VSTestTrace.WriteTrace($"Cleaning artifact directory '{expectedArtifactDirectory}'");
```
Evangelink, line 197: Nit:

```suggestion
                    VSTestTrace.WriteTrace($"Cleaning artifact directory '{expectedArtifactDirectory}'.");
```

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -38,25 +40,32 @@ public static TestCommand FromParseResult(ParseResult result, string[] settings,
             if (settings.Any())
             {
                 // skip '--' and escape every \ to be \\ and every " to be \" to survive the next hop
-                var escaped = settings.Skip(1).Select(s => s.Replace("\\", "\\\\").Replace("\"", "\\\"")).ToArray();
```
Evangelink, line 41: Have we changed the `.editorconfig` settings for this repo?

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
Evangelink, line 3: Nit: 

```suggestion
```

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -0,0 +1,46 @@
...
+        public static FeatureFlag Default { get; } = new FeatureFlag();
+
+        public FeatureFlag()
+        {
+            FeatureFlagsBag.Add(ARTIFACTS_POSTPROCESSING, false);
+        }
+
+        // Added for artifact porst-processing, it enable/disable the post processing.
+        // Added in 17.2-preview 7.0-preview
+        public const string ARTIFACTS_POSTPROCESSING = Prefix + "ARTIFACTS_POSTPROCESSING";
```
Evangelink, line 26: `ARTIFACTS_POSTPROCESSING_SDK_KEEP_OLD_UX` is missing

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.Tools.Test
+{
+    // !!! FEATURES MUST BE KEPT IN SYNC WITH https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs !!!
```
Evangelink, line 10: If we want the file to be exactly the same, can't we somehow link it? For example, with [Paket](https://fsprojects.github.io/Paket/github-dependencies.html) we can reference a file directly.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -0,0 +1,46 @@
...
+    internal class FeatureFlag
+    {
+        private const string Prefix = "VSTEST_FEATURE_";
+
+        public Dictionary<string, bool> FeatureFlagsBag { get; } = new();
```
Evangelink, line 15: This is named `FeatureFlags` in the other class, I think that's better than bag.

src/Cli/dotnet/commands/dotnet-test/VSTestForwardingApp.cs
```diff
@@ -19,7 +21,10 @@ public VSTestForwardingApp(IEnumerable<string> argsToForward)
             if (!hasRootVariable)
             {
                 WithEnvironmentVariable(rootVariableName, rootValue);
+                VSTestTrace.WriteTrace($"Root variable set {rootVariableName}:{rootValue}");
             }
+
+            VSTestTrace.SafeWriteTrace(() => $"Forwarding to '{GetVSTestExePath()}' with args \"{argsToForward?.Aggregate((a, b) => $"{a} | {b}")}\"");
```
Evangelink, line 27: Why `SafeWriteTrace` here and not at the other places?

Evangelink, line 27: I don't know if that matters but in the other similar trace you did `{a}|{b}` (no space around `|`).

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
Evangelink, line 3:
```suggestion
```

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
```diff
@@ -0,0 +1,61 @@
...
+            s_traceFilePath = Environment.GetEnvironmentVariable("DOTNET_CLI_VSTEST_TRACEFILE");
+            if (TraceEnabled)
+            {
+                Console.WriteLine($"[dotnet test - {DateTime.UtcNow}]Logging to {(!string.IsNullOrEmpty(s_traceFilePath) ? s_traceFilePath : "console")}");
+            }
+        }
+
+        public static void WriteTrace(string logText)
+        {
+            if (TraceEnabled)
```
Evangelink, line 27: Could we revert the condition to reduce nesting?

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
```diff
@@ -0,0 +1,61 @@
...
+//
+
+using System;
+using System.IO;
+
+namespace Microsoft.DotNet.Tools.Test
+{
+    internal class VSTestTrace
+    {
+        public static bool TraceEnabled { get; set; }
```
Evangelink, line 12: Do we want to allow changing the value? (public setter).

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
```diff
@@ -0,0 +1,61 @@
...
+                        logFile.WriteLine(message);
+                    }
+                    else
+                    {
+                        Console.WriteLine(message);
+                    }
+                }
+                catch
+                {
+                    // Avoid exception if we have issue with the log file.
```
Evangelink, line 44: Might be worth logging the exn?

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            TraceEnabled = int.TryParse(Environment.GetEnvironmentVariable("DOTNET_CLI_VSTEST_TRACE"), out int enabled) && enabled == 1;
+            s_traceFilePath = Environment.GetEnvironmentVariable("DOTNET_CLI_VSTEST_TRACEFILE");
+            if (TraceEnabled)
+            {
+                Console.WriteLine($"[dotnet test - {DateTime.UtcNow}]Logging to {(!string.IsNullOrEmpty(s_traceFilePath) ? s_traceFilePath : "console")}");
+            }
+        }
+
+        public static void WriteTrace(string logText)
```
Evangelink, line 25: What about overloads with params array? both writing function support them.

src/Cli/dotnet/commands/dotnet-vstest/Program.cs
```diff
@@ -1,12 +1,12 @@
 ﻿// Copyright(c) .NET Foundation and contributors.All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.Cli;
 using System.CommandLine.Parsing;
 using System.Collections.Generic;
 using System.Linq;
 using System;
+using Microsoft.DotNet.Tools.Test;
```
Evangelink, line 9: Looks like MS namespaces are defined before `System` here, could you please move this using?

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+            while (true)
+            {
+                string line = reader.ReadLine()?.Trim();
+                if (line is null) break;
+                output.Add(line);
+            }
+
+            if (merge)
+            {
+                Assert.Equal(string.Empty, output[output.Count - 3].Trim());
```
Evangelink, line 128: I don't know if repo supports it but could you use the `^X` operator?

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+        public GivenDotnetTestBuildsAndRunsArtifactPostProcessing(ITestOutputHelper log) : base(log)
+        {
+            BuildDataCollector();
+            BuildDataCollectorNoMerge();
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ArtifactPostProcessing_CsProjs(bool merge)
```
Evangelink, line 37: Could we try to have a bit more explanatory test names?

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+        {
+            TestAsset testInstance = _testAssetsManager.CopyTestAsset("VSTestMultiProjectSolution", Guid.NewGuid().ToString())
+                .WithSource();
+
+            string runsettings = GetRunsetting(testInstance.Path);
```
Evangelink, line 42: THis is similar in all your tests, might be worth having a method for it? Maybe only the `TestAsset` creation.

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            CommandResult result = new DotnetTestCommand(Log)
+                                    .WithWorkingDirectory(testInstance.Path)
+                                    .WithEnvironmentVariable(FeatureFlag.ARTIFACTS_POSTPROCESSING, "1")
+                                    .Execute(
+                                    "--configuration", "release",
+                                    "--collect", "SampleDataCollector",
+                                    "--test-adapter-path", merge ? Path.GetDirectoryName(s_dataCollectorDll) : Path.GetDirectoryName(s_dataCollectorNoMergeDll),
+                                    "--settings", runsettings,
+                                    "--diag", testInstance.Path + "/logs/");
```
Evangelink, line 52: Nit:

```suggestion
            CommandResult result = new DotnetTestCommand(Log)
                .WithWorkingDirectory(testInstance.Path)
                .WithEnvironmentVariable(FeatureFlag.ARTIFACTS_POSTPROCESSING, "1")
                .Execute(
                    "--configuration", "release",
                    "--collect", "SampleDataCollector",
                    "--test-adapter-path", merge ? Path.GetDirectoryName(s_dataCollectorDll) : Path.GetDirectoryName(s_dataCollectorNoMergeDll),
                    "--settings", runsettings,
                    "--diag", testInstance.Path + "/logs/");
```

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+                                    .WithEnvironmentVariable(FeatureFlag.ARTIFACTS_POSTPROCESSING, "1")
+                                    .WithEnvironmentVariable("DOTNET_CLI_VSTEST_TRACE", "1")
+                                    .Execute(
+                                    Directory.GetFiles(testInstance.Path, "test1.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
+                                    Directory.GetFiles(testInstance.Path, "test2.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
+                                    Directory.GetFiles(testInstance.Path, "test3.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
+                                    "--collect:SampleDataCollector",
+                                    $"--test-adapter-path:{(merge ? Path.GetDirectoryName(s_dataCollectorDll) : Path.GetDirectoryName(s_dataCollectorNoMergeDll))}",
+                                    $"--settings:{runsettings}",
+                                    "--diag:" + testInstance.Path + "/logs/");
```
Evangelink, line 81: Nit:

```suggestion
            CommandResult result = new DotnetTestCommand(Log)
                .WithWorkingDirectory(testInstance.Path)
                .WithEnvironmentVariable(FeatureFlag.ARTIFACTS_POSTPROCESSING, "1")
                .WithEnvironmentVariable("DOTNET_CLI_VSTEST_TRACE", "1")
                .Execute(
                    Directory.GetFiles(testInstance.Path, "test1.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
                    Directory.GetFiles(testInstance.Path, "test2.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
                    Directory.GetFiles(testInstance.Path, "test3.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
                    "--collect:SampleDataCollector",
                    $"--test-adapter-path:{(merge ? Path.GetDirectoryName(s_dataCollectorDll) : Path.GetDirectoryName(s_dataCollectorNoMergeDll))}",
                    $"--settings:{runsettings}",
                    $"--diag:{testInstance.Path}/logs/");
```

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+              .WithWorkingDirectory(testInstance.Path)
+                                    .WithEnvironmentVariable(FeatureFlag.ARTIFACTS_POSTPROCESSING, "1")
+                                    .Execute(
+                                    Directory.GetFiles(testInstance.Path, "test1.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
+                                    Directory.GetFiles(testInstance.Path, "test2.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
+                                    Directory.GetFiles(testInstance.Path, "test3.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
+                                    "--collect:SampleDataCollector",
+                                    $"--testAdapterPath:{(merge ? Path.GetDirectoryName(s_dataCollectorDll) : Path.GetDirectoryName(s_dataCollectorNoMergeDll))}",
+                                    $"--settings:{runsettings}",
+                                    "--diag:" + testInstance.Path + "/logs/");
```
Evangelink, line 109: Nit

```suggestion
            CommandResult result = new DotnetVSTestCommand(Log)
                .WithWorkingDirectory(testInstance.Path)
                .WithEnvironmentVariable(FeatureFlag.ARTIFACTS_POSTPROCESSING, "1")
                .Execute(
                    Directory.GetFiles(testInstance.Path, "test1.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
                    Directory.GetFiles(testInstance.Path, "test2.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
                    Directory.GetFiles(testInstance.Path, "test3.dll", SearchOption.AllDirectories).SingleOrDefault(x => x.Contains("publish")),
                    "--collect:SampleDataCollector",
                    $"--testAdapterPath:{(merge ? Path.GetDirectoryName(s_dataCollectorDll) : Path.GetDirectoryName(s_dataCollectorNoMergeDll))}",
                    $"--settings:{runsettings}",
                    $"--diag:{testInstance.Path}/logs/");
```

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
Evangelink, line 128: Could we favor `FluentAssertions` syntax?

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+                output.Add(line);
+            }
+
+            if (merge)
+            {
+                Assert.Equal(string.Empty, output[output.Count - 3].Trim());
+                Assert.Equal("Attachments:", output[output.Count - 2].Trim());
+                string mergedFile = output[output.Count - 1].Trim();
+
+                var fileContent = new List<string>();
```
Evangelink, line 132: Please use a `HashSet<string>` instead of the `List`, this way you don't need the `Distinct()`.

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
```diff
@@ -0,0 +1,249 @@
...
+
+                string testProjectDirectory = testInstance.Path;
+
+                new BuildCommand(testInstance)
+                            .Execute("/p:Configuration=Release")
+                            .Should()
+                            .Pass();
+
+                return Directory.GetFiles(testProjectDirectory, "AttachmentProcessorDataCollector.dll", SearchOption.AllDirectories).Single(x => x.Contains("bin"));
+            });
```
Evangelink, line 197: This looks so weird. Why do we need static members that holds something that is changed for each test?

src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+	<PackageReference Include="Microsoft.TestPlatform.ObjectModel" Version="17.2.0-preview-20220131-20" />
```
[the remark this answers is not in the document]
MarcoRossignoli: I don't know if it will be used...`MicrosoftNETTestSdkPackageVersion` is inside /eng/Version.props this directory will be created inside /artifacts/... I need only to be sure to add a version that contains the postprocessing feature and at the moment we have only public preview

src/Assets/TestProjects/VSTestMultiProjectSolution/test1/test1.csproj
MarcoRossignoli: nope

src/Cli/dotnet/commands/dotnet-test/Program.cs
MarcoRossignoli: `SafeWriteTrace` is used only where we don't simply generate a string but also where to do that we're running complex code(where parameters are not primitive). This code will run always so I don't want to fail normal run(where trace are disabled) because for instance in this case `testCommand` is in "weird" strate and `GetArgumentsToMSBuild` will fail.

MarcoRossignoli: It's log for "emergency" we don't need beautiful formatting. By design is disabled and I hope that we never need to use it 😄

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -70,27 +79,45 @@ public static TestCommand FromParseResult(ParseResult result, string[] settings,
...
         }
 
         public static int Run(ParseResult parseResult)
         {
             parseResult.HandleDebugSwitch();
 
-            var args = parseResult.GetArguments();
+            FeatureFlag.Default.PrintFlagFeatureState();
+
+            // We use also current process id for the correlation id for possible future usage in case we need to know the parent process
```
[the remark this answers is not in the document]
MarcoRossignoli: We need and id to correlate parallel run on vstest side. When we run 

```
dotnet test sln/csproj
```
we're running 1 dotnet.exe process for every project inside the solutions/list of proj specified.

If in future we need for some reason know who is the "parent" process we can parse the first fragment. By design msbuild start some process in the middle so we don't have any way from vstest runner side to know who is the "real" parent process.

src/Cli/dotnet/commands/dotnet-test/Program.cs
MarcoRossignoli: For msbuild we pass argument how our MSBuild task understand. 
But we can also run 
```cs
dotnet test container.dll
```
In the case above the parameter has got different name than msbuild one...so sdk is splitted in two main branch, run through Msbuild and forward to vstest.console.dll directly(using dotnet exe vstest.console.dll --param).

src/Cli/dotnet/commands/dotnet-test/Program.cs
MarcoRossignoli: The folder is created by vstest side and is created only if we have artifacts. Cannot be empty...or can be empty if there's a bug.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -110,17 +149,67 @@ public static int Run(ParseResult parseResult)
...
+            string expectedArtifactDirectory = Path.Combine(Path.GetTempPath(), testSessionCorrelationId);
+            if (!Directory.Exists(expectedArtifactDirectory))
+            {
+                VSTestTrace.WriteTrace("No artifact found, post-processing won't run.");
+                return 0;
+            }
+
+            VSTestTrace.WriteTrace($"Artifacts found inside '{expectedArtifactDirectory}', running post-processing.");
+
+            var artifactsProcessingModePostprocess = new List<string> { "--artifactsProcessingMode-postprocess", $"--testSessionCorrelationId:{testSessionCorrelationId}" };
```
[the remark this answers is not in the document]
MarcoRossignoli: I agree better name I'll fix thx

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
```diff
@@ -0,0 +1,61 @@
...
+        public static bool TraceEnabled { get; set; }
+        private static readonly string s_traceFilePath;
+
+        static VSTestTrace()
+        {
+            TraceEnabled = int.TryParse(Environment.GetEnvironmentVariable("DOTNET_CLI_VSTEST_TRACE"), out int enabled) && enabled == 1;
+            s_traceFilePath = Environment.GetEnvironmentVariable("DOTNET_CLI_VSTEST_TRACEFILE");
+            if (TraceEnabled)
+            {
+                Console.WriteLine($"[dotnet test - {DateTime.UtcNow}]Logging to {(!string.IsNullOrEmpty(s_traceFilePath) ? s_traceFilePath : "console")}");
```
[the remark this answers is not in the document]
MarcoRossignoli: We don't need a lot of precision here...it's only to have some time ref, we don't need offset or stuff like that. But I'll update it's ok.

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
MarcoRossignoli: will do

src/Cli/dotnet/commands/dotnet-test/Program.cs
MarcoRossignoli: I don't think we can do it...this repo is used by a lot of different team. I'm simply applying current style where suggested.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
MarcoRossignoli: It's weird...I've used auto format inside VS and it added this.

src/Cli/dotnet/commands/dotnet-test/VSTestForwardingApp.cs
MarcoRossignoli: I'll add space everywhere thx

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: I don't know...I'll check

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: The naming is the same of other tests...we defin csprojs when as argument for the `dotnet test` we have project files.

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: I don't know, it's the same to me...I've used the default VS formatting shortcut...

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: Not strong opinion...to me an assertion is an assertion and I expect that it's in place, I let the dev do what he likes or what he find more comfortable. 
I'll move to fluent if you prefer.

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: I'll do

src/Assets/TestProjects/VSTestDataCollectorSample/AttachmentProcessorDataCollector.csproj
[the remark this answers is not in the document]
MarcoRossignoli: You're right looks like it's possible we're doing the same in other tests...I'll fix.

src/Cli/dotnet/commands/dotnet-test/Program.cs
Evangelink: I also saw the suggestions but they were only set as suggestion so I didn't go with fixing them.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
MarcoRossignoli: Not needed here we're not using it inside the SDK. We need to keep in sync only used flag

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
MarcoRossignoli: we don't need the identical, the meaning is that if we touch this one or the other side we need to keep in sync what we're using(I'll add to vstest repo as soon we merge this one).

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
MarcoRossignoli: if we fail the trace we cannot log...or we can log only on console but I don't like the idea to see trace error, I prefer lose the error also because it's not so useful as message if it happens.

src/Cli/dotnet/commands/dotnet-test/VSTestForwardingApp.cs
MarcoRossignoli: Ditto explained on the other comment.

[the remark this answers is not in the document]
MarcoRossignoli: mmm why? It's a trace it should not fail...if we see that's null we know that something is strange...also I think that code will fail somewhere if for some reason this prop is null.

src/Cli/dotnet/commands/dotnet-test/VSTestTrace.cs
MarcoRossignoli: Added catch and log in console.

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: Works in local..let's see the CI

src/Tests/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsArtifactPostProcessing.cs
MarcoRossignoli: Updated with `SolutionProjects`

MarcoRossignoli:
@Evangelink refactored a bit the main Run() splitting between run for test containers and run with MSBuild, removed non safe write trace(I do the check inside there, we'll have more closure but it's ok).

PR #24028, opened 2022-02-19
[WorkloadInstall] Simplify code workload install command line 

mandel-macaque:
I am trying to locate a bug that is hitting me in the xamarin-macios ci. I was finding it hard to follow the code and found some small bugs/issues while looking at it. I have tried to make it simpler for people like me :) 

1. I have removed all the not needed enum boxing that was done via Equals. I cannot think of a reason why that boxing was preferred.
2. Negated a number of || to use && which short-circuits making it a little better in performance but more important, easier to understand.
3. Reduced the number of nesting levels.
4. Removed a number of allocations that were not needed but got added to avoid a NRE in a captured variable, maybe Enumerable.Emty was not available when this was landed?
5. Removed the used or the TaskAwaiter which is meant to be used by the compiler but not by us humans.
6. Enabled nullability in the file, which pointed out a number of small errors.

I believe that the reviews will be easier done commit by commit, each commit makes changes to a single function when possible (nullability is an exception).

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
@@ -86,7 +86,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
             var configOption = parseResult.GetValueForOption(WorkloadInstallCommandParser.ConfigOption);
             var sourceOption = parseResult.GetValueForOption(WorkloadInstallCommandParser.SourceOption);
-            if (string.IsNullOrEmpty(configOption) && sourceOption != null && sourceOption.Any())
+            if (string.IsNullOrEmpty(configOption) && !(sourceOption != null && sourceOption.Any()))
             {
                 _packageSourceLocation = null;
             }
@@ -130,7 +130,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             }
         }
 
-        int PrintDownloadLink() { 
+        private int PrintDownloadLink() { 
             _reporter.WriteLine(string.Format(LocalizableStrings.ResolvingPackageUrls, string.Join(", ", _workloadIds)));
             var packageUrls = GetPackageDownloadUrlsAsync(_workloadIds.Select(id => new WorkloadId(id)), _skipManifestUpdate, _includePreviews).GetAwaiter().GetResult();
 
@@ -141,7 +141,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             return _workloadInstaller.ExitCode;
         }
 
-        int DownloadToCache()
+        private int DownloadToCache()
         {
             try
             {
@@ -379,10 +379,10 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             }
         }
 
-        async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
+        private async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
         {
             var manifestPackagePaths = await _workloadManifestUpdater.DownloadManifestPackagesAsync(includePreview, tempPath);
-            if (manifestPackagePaths != null && manifestPackagePaths.Any())
+            if (!(manifestPackagePaths != null && manifestPackagePaths.Any()))
             {
                 _reporter.WriteLine(LocalizableStrings.SkippingManifestUpdate);
                 return;
@@ -392,7 +392,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             _workloadResolver = _workloadResolver.CreateOverlayResolver(overlayProvider);
         }
 
-        async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadIds, DirectoryPath offlineCache, bool skipManifestUpdate, bool includePreviews)
+        private async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadIds, DirectoryPath offlineCache, bool skipManifestUpdate, bool includePreviews)
         {
             string? tempManifestDir = null;
             if (!skipManifestUpdate)
@@ -417,7 +417,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             switch (_workloadInstaller.GetInstallationUnit())
             {
                 case InstallationUnit.Packs:
-                    var packInstaller= _workloadInstaller.GetPackInstaller();
+                    var packInstaller = _workloadInstaller.GetPackInstaller();
                     var workloadPacks = GetPacksToInstall(workloadIds);
                     foreach (var pack in workloadPacks)
                     {
@@ -439,7 +439,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             }
         }
 
-        IEnumerable<PackInfo?> GetPacksToInstall(IEnumerable<WorkloadId> workloadIds)
+        private IEnumerable<PackInfo?> GetPacksToInstall(IEnumerable<WorkloadId> workloadIds)
         {
             var installedPacks = _workloadInstaller.GetPackInstaller().GetInstalledPacks(_sdkFeatureBand);
             return workloadIds
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
@@ -18,15 +18,14 @@ using NuGet.Common;
 using static Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadResolver;
 using System.Threading.Tasks;
 
+#nullable enable
+
 namespace Microsoft.DotNet.Workloads.Workload.Install
 {
     internal class WorkloadInstallCommand : CommandBase
     {
         private readonly IReporter _reporter;
         private readonly bool _skipManifestUpdate;
-        private readonly string _fromCacheOption;
-        private readonly string _downloadToCacheOption;
-        private readonly PackageSourceLocation _packageSourceLocation;
         private readonly bool _printDownloadLinkOnly;
         private readonly bool _includePreviews;
         private readonly VerbosityOptions _verbosity;
@@ -40,20 +39,24 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
         private readonly string _userProfileDir;
         private readonly string _tempDirPath;
         private readonly string _dotnetPath;
-        private readonly string _fromRollbackDefinition;
+        // optional/nullable
+        private readonly PackageSourceLocation? _packageSourceLocation;
+        private readonly string? _fromCacheOption;
+        private readonly string? _downloadToCacheOption;
+        private readonly string? _fromRollbackDefinition;
 
         public WorkloadInstallCommand(
             ParseResult parseResult,
-            IReporter reporter = null,
-            IWorkloadResolver workloadResolver = null,
-            IInstaller workloadInstaller = null,
-            INuGetPackageDownloader nugetPackageDownloader = null,
-            IWorkloadManifestUpdater workloadManifestUpdater = null,
-            string dotnetDir = null,
-            string userProfileDir = null,
-            string tempDirPath = null,
-            string version = null,
-            IReadOnlyCollection<string> workloadIds = null)
+            IReporter? reporter = null,
+            IWorkloadResolver? workloadResolver = null,
+            IInstaller? workloadInstaller = null,
+            INuGetPackageDownloader? nugetPackageDownloader = null,
+            IWorkloadManifestUpdater? workloadManifestUpdater = null,
+            string? dotnetDir = null,
+            string? userProfileDir = null,
+            string? tempDirPath = null,
+            string? version = null,
+            IReadOnlyCollection<string>? workloadIds = null)
             : base(parseResult)
         {
             _reporter = reporter ?? Reporter.Output;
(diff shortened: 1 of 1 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -369,24 +414,23 @@ private async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadI
...
-            }
-            else
-            {
-                var installer = _workloadInstaller.GetWorkloadInstaller();
-                foreach (var workloadId in workloadIds)
-                {
-                    installer.DownloadToOfflineCache(workloadId, offlineCache, includePreviews);
-                }
+                case InstallationUnit.Packs:
+                    var packInstaller= _workloadInstaller.GetPackInstaller();
```
mandel-macaque, line 420: Comment to self: nit, missing space

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -395,15 +439,14 @@ private async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadI
             }
         }
 
-        private IEnumerable<PackInfo> GetPacksToInstall(IEnumerable<WorkloadId> workloadIds)
+        IEnumerable<PackInfo?> GetPacksToInstall(IEnumerable<WorkloadId> workloadIds)
```
premun, line 442: If you ever figure out how to make Roslyn tell this should be non-nullable becase you do `Where(not null)`, let me know.

I know about `Select(x => x!)` and `Cast<>()` but they seem like extra leg work..

mandel-macaque: I know how to, you can use the more SQL looking syntax, I used it here: https://github.com/dotnet/sdk/pull/24028/files#diff-498ad5982ef0f101943528170b9fb47715442c897bf94ef0c8c1f2438f4f8effR361

Not ideal, that's why I do it in the caller and not the callee.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -64,27 +67,41 @@ internal class WorkloadInstallCommand : CommandBase
...
-            _packageSourceLocation = string.IsNullOrEmpty(configOption) && (sourceOption == null || !sourceOption.Any()) ? null :
-                new PackageSourceLocation(string.IsNullOrEmpty(configOption) ? null : new FilePath(configOption), sourceFeedOverrides: sourceOption);
+            if (string.IsNullOrEmpty(configOption) && sourceOption != null && sourceOption.Any())
+            {
+                _packageSourceLocation = null;
+            }
+            else
+            {
+                _packageSourceLocation = new(string.IsNullOrEmpty(configOption) ? null : new FilePath(configOption), sourceFeedOverrides: sourceOption);
+            }
```
KalleOlaviNiemitalo, line 96: That is not equivalent to the old code.  If `string.IsNullOrEmpty(configOption) && sourceOption == null`, then the old code would assign `_packageSourceLocation = null`, but your new code goes to the other branch instead.

KalleOlaviNiemitalo: If you want to use `&&`, then `string.IsNullOrEmpty(configOption) && !(sourceOption != null && sourceOption.Any())` is equivalent to the old code, but I wouldn't say it's clearer than the old code. I feel there's a nice symmetry between string.IsNull***Or***Empty and `||`.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -358,10 +358,10 @@ async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<Workload
             }
         }
 
-        private async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
+        async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
```
KalleOlaviNiemitalo, line 382: This goes against coding style <https://github.com/dotnet/runtime/blob/a55d5300457004aff05a603ba9960290517f5681/docs/coding-guidelines/coding-style.md?plain=1#L14-L16>

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -358,10 +358,10 @@ async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<Workload
...
         {
             var manifestPackagePaths = await _workloadManifestUpdater.DownloadManifestPackagesAsync(includePreview, tempPath);
-            if (manifestPackagePaths == null || !manifestPackagePaths.Any())
+            if (manifestPackagePaths != null && manifestPackagePaths.Any())
             {
                 _reporter.WriteLine(LocalizableStrings.SkippingManifestUpdate);
```
KalleOlaviNiemitalo, line 387: This changes the behavior to the opposite.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -99,22 +99,19 @@ internal class WorkloadInstallCommand : CommandBase
...
-                    {
-                        throw new GracefulException(string.Format(LocalizableStrings.WorkloadNotSupportedOnPlatform, workloadId), isUserError: false);
-                    }
-                    else
-                    {
-                        throw new GracefulException(string.Format(LocalizableStrings.WorkloadNotRecognized, workloadId), isUserError: false);
-                    }
-                }
+                if (availableWorkloads.Any(workload => workload.Id.ToString() == workloadId))
+                    continue;
```
KalleOlaviNiemitalo, line 123: This allocates an extra closure and a non-static delegate, but perhaps fewer enumerators. I don't know whether efficiency is important here.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
mandel-macaque: yep, you are correct, I'm missing a ! around the &&, good catch.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
mandel-macaque: Sorry, i'm used to the mono coding style.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
mandel-macaque: Should be doing a lot less enumerations because Any will break ASAP.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -69,27 +67,41 @@ internal class WorkloadInstallCommand : CommandBase
...
+            {
+                _tempDirPath = tempDirPath;
+            }
+            
 
             var configOption = parseResult.GetValueForOption(WorkloadInstallCommandParser.ConfigOption);
             var sourceOption = parseResult.GetValueForOption(WorkloadInstallCommandParser.SourceOption);
-            _packageSourceLocation = string.IsNullOrEmpty(configOption) && (sourceOption == null || !sourceOption.Any()) ? null :
-                new PackageSourceLocation(string.IsNullOrEmpty(configOption) ? null : new FilePath(configOption), sourceFeedOverrides: sourceOption);
+            if (string.IsNullOrEmpty(configOption) && sourceOption != null && sourceOption.Any())
```
mandel-macaque, line 89:
```suggestion
            if (string.IsNullOrEmpty(configOption) && !(sourceOption != null && sourceOption.Any()))
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -305,39 +345,44 @@ private async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<
...
+                if (!string.IsNullOrWhiteSpace(tempPath?.Value) && Directory.Exists(tempPath?.Value)) 
                 {
-                    Directory.Delete(tempPath.Value.Value, true);
+                    Directory.Delete(tempPath?.Value!, true); // will not throw NRE thanks to Directory.Exists(null) => false
                 }
             }
         }
 
-        private async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
+        async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
```
mandel-macaque, line 382:
```suggestion
        private async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, bool includePreview)
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
KalleOlaviNiemitalo: `||` does short-circuit.  `sourceOption == null || !sourceOption.Any()` does not call `sourceOption.Any()` if `sourceOption == null` is true.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -305,39 +345,44 @@ private async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<
...
         {
             var manifestPackagePaths = await _workloadManifestUpdater.DownloadManifestPackagesAsync(includePreview, tempPath);
-            if (manifestPackagePaths == null || !manifestPackagePaths.Any())
+            if (manifestPackagePaths != null && manifestPackagePaths.Any())
```
mandel-macaque, line 385:
```suggestion
            if (!(manifestPackagePaths != null && manifestPackagePaths.Any()))
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
mandel-macaque, line 420:
```suggestion
                    var packInstaller = _workloadInstaller.GetPackInstaller();
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
mandel-macaque, line 442:
```suggestion
        private IEnumerable<PackInfo?> GetPacksToInstall(IEnumerable<WorkloadId> workloadIds)
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -347,9 +392,9 @@ private async Task UseTempManifestsToResolvePacksAsync(DirectoryPath tempPath, b
             _workloadResolver = _workloadResolver.CreateOverlayResolver(overlayProvider);
         }
 
-        private async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadIds, DirectoryPath offlineCache, bool skipManifestUpdate, bool includePreviews)
+        async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadIds, DirectoryPath offlineCache, bool skipManifestUpdate, bool includePreviews)
```
mandel-macaque, line 395:
```suggestion
        private async Task DownloadToOfflineCacheAsync(IEnumerable<WorkloadId> workloadIds, DirectoryPath offlineCache, bool skipManifestUpdate, bool includePreviews)
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
mandel-macaque: Good catch (should not doo boolean logic at 3 am :( )

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -198,99 +223,114 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
...
+                    InstallPacksWithInstallRecord(workloadIds, sdkFeatureBand, manifestsToUpdate, offlineCache);
+                    break;
+                default:    
+                    InstallWorkloadsWithInstallRecord(workloadIds);
+                    break;
             }
         }
 
-        private async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<WorkloadId> workloadIds, bool skipManifestUpdate, bool includePreview)
+        async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<WorkloadId> workloadIds, bool skipManifestUpdate, bool includePreview)
```
mandel-macaque, line 333:
```suggestion
        private async Task<IEnumerable<string>> GetPackageDownloadUrlsAsync(IEnumerable<WorkloadId> workloadIds, bool skipManifestUpdate, bool includePreview)
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -198,99 +223,114 @@ public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManif
...
+        {
+            var installer = _workloadInstaller.GetWorkloadInstaller();
+            foreach (var workloadId in workloadIds)
+            {
+                installer.InstallWorkload(workloadId);
             }
         }
 
-        private void InstallWorkloadsWithInstallRecord(
+        void InstallPacksWithInstallRecord(
```
mandel-macaque, line 249:
```suggestion
        private void InstallPacksWithInstallRecord(
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -102,64 +114,77 @@ internal class WorkloadInstallCommand : CommandBase
...
+            var packageUrls = GetPackageDownloadUrlsAsync(_workloadIds.Select(id => new WorkloadId(id)), _skipManifestUpdate, _includePreviews).GetAwaiter().GetResult();
+
+            _reporter.WriteLine("==allPackageLinksJsonOutputStart==");
+            _reporter.WriteLine(JsonSerializer.Serialize(packageUrls));
+            _reporter.WriteLine("==allPackageLinksJsonOutputEnd==");
+
+            return _workloadInstaller.ExitCode;
+        }
+
+        int DownloadToCache()
```
mandel-macaque, line 144:
```suggestion
        private int DownloadToCache()
```

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -102,64 +114,77 @@ internal class WorkloadInstallCommand : CommandBase
...
+
+                var errorMsg = _workloadResolver.IsPlatformIncompatibleWorkload(new (workloadId)) ? 
+                    string.Format(LocalizableStrings.WorkloadNotSupportedOnPlatform, workloadId) :
+                    string.Format(LocalizableStrings.WorkloadNotRecognized, workloadId);
+                
+                throw new GracefulException(errorMsg, isUserError: false);
+            }
+        }
+
+        int PrintDownloadLink() {
```
mandel-macaque, line 133:
```suggestion
        private int PrintDownloadLink() { 
```

marcpopMSFT:
Following up on old PRs. The merge conflicts are now fairly involved. If you want to resolve them we can take a look but otherwise I think we'll close this out. Thanks for the effort.

MiYanni:
I looked over the changes and much of this code was extracted into a base class, `InstallingWorkloadCommands`. I didn't see anything that stood out as a fundamental improvement; mostly formatting and general cleanliness.

I'd like to do improvements like these across the entire repo to improve readability/ease-of-use. I plan on adding editorconfig rules so that things remain tidy. Unfortunately, this particular set of changes is out-of-date, so I'm closing the PR. Feel free to reopen if you want some of these changes with the current state of the code.

PR #24063, opened 2022-02-23
Fix DefaultExcludesInProjectFolder value ignored

ViktorHofer:
The `DefaultItemExcludesInProjectFolder` property is never defined and was likely forgotten to be renamed. Because of that, the `DefaultExcludesInProjectFolder` property overwrites any user set value. Add `DefaultExcludesInProjectFolder` and keep `DefaultItemExcludesInProjectFolder` to not break existing consumers.


dsplaisted:
This probably was a mistake, but I'm not sure about taking the change now, as it looks like there are projects setting the property that isn't otherwise set or used in the SDK: https://github.com/search?q=DefaultItemExcludesInProjectFolder&type=code

ViktorHofer:
What about excluding both `DefaultItemExcludesInProjectFolder ` and `DefaultExcludesInProjectFolder` so that it matches the documentation but also doesn't break existing customers?

dsplaisted (approved):

PR #24070, opened 2022-02-23
Add design for cross-version workloads

dsplaisted:
This describes how we can consume "runtime-specific" workload assets in later versions of the .NET SDK.

### DIFF
--- a/documentation/general/workloads/cross-version-workloads.md
+++ b/documentation/general/workloads/cross-version-workloads.md
@@ -1,0 +1,138 @@
+# Handling workload assets across major .NET versions
+
+The latest version of the .NET SDK supports targeting previous versions of .NET.  To do so, it needs assets produced from the build of the previous version of the .NET runtime.  Prior to .NET 6, this included packs such as targeting packs and runtime packs, and the packs were acquired as needed during NuGet restore.  We will call packs that are specific to the version of the runtime being targeted "runtime-specific packs".  Runtime-specific packs are produced as part of the build of the corresponding runtime, but are used by later versions of the .NET SDK when targeting that version of the runtime.  For example, when the the .NET 7 SDK builds a project targeting .NET 6, it will need targeting packs and runtime packs for .NET 6 which were produced by the .NET 6 build.
+
+In .NET 6, some runtime-specific packs are now delivered as workload packs.  In the .NET 7 SDK, these packs need to be available for projects that are targeting .NET 6.  The packs that are traditional targeting and runtime packs can be downloaded using existing mechanisms without being installed as part of a workload, but there are other packs which do not support this.  These include the pack delivering the Mono AoT cross-compiler, and the Emscripten toolchain for Blazor WASM apps.  The Mono AoT cross-compiler is similar to CrossGen 2, which we already support downloading via NuGet restore.  Thus, we can and probably should update the .NET SDK and the workloads which use this AoT compiler to support downloading it in a similar fashion to CrossGen 2.
+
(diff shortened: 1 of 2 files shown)


documentation/general/workloads/cross-version-workloads.md
```diff
@@ -0,0 +1,138 @@
...
+
+We will need to keep the versions of the runtime-specific workload packs synced between the .NET 6 and .NET 7 versions of the runtime-specific workload manifests.  During previews for .NET 7, this will be a manual process, where we will update the versions in the .NET 7 manifest manually when a new servicing version of .NET 6 is released.  After .NET 7 is released, we hope to be able to automatically calculate the .NET 6 version numbers by adding a fixed amount to the .NET 7 patch version.  For example, if .NET 7.0.1 releases together with .NET 6.0.12, then we can add 11 to the patch version of .NET 7 to calculate the corresponding version for the .NET 6 runtime-specific packs.  This would let us avoid having to manually sync the version numbers for each servicing release.
```
baronfel, line 130: > After .NET 7 is released, we hope to be able to automatically calculate the .NET 6 version numbers by adding a fixed amount to the .NET 7 patch version.

Once the initial release is done, would this number be frozen? I naively expect this to be the case, because from the time that 7.0.1 is released there would likely be runtime releases for both 6.0.x and 7.0.x from that time onward.

How does EOL/servicing lifetimes impact this? If the number 11 is hard-coded but 6.0.x runtimes are published longer than 7.0.x runtimes for example, would we be forced to come back and do more complex logic here?

baronfel: We chatted about this today and the takeaway was we're fine with this amount of book-keeping. We will default to an offset-based version calculation, but use that offset to compute a default value for the specific version offsets per-runtime. Then, if we need to pin specific offsets we can simply set the matching properties to specific values.

Redth:
@mattleibow @jonathanpeppers

documentation/general/workloads/cross-version-workloads.md
```diff
@@ -0,0 +1,138 @@
...
+
+```xml
+  <ImportGroup Condition="'$(RunAOTCompilation)' == 'true' And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == '6.0'">
+    <Import Project="Sdk.props" Sdk="Emscripten.Sdk.net6" />
+  </ImportGroup>
+  <ImportGroup Condition="'$(RunAOTCompilation)' == 'true' And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == '7.0'">
+    <Import Project="Sdk.props" Sdk="Emscripten.Sdk.net7" />
+  </ImportGroup>
+```
```
jonathanpeppers, line 120: When .NET 7 is stable, do we anticipate that building a `net6.0-android` project would use the .NET 7 Android workload (SDK & .targets)? Or will it somehow need to use the *old* .NET 6 targets?

documentation/general/workloads/cross-version-workloads.md
```diff
@@ -0,0 +1,138 @@
...
+
+## .NET 6 manifests in .NET 7 SDK
```
jonathanpeppers, line 122: How far back does this go? Will .NET 8 support `net7.0-android` and `net6.0-android`? Or is it only have 1 LTS release behind?

documentation/general/workloads/cross-version-workloads.md
dsplaisted: It should use the .NET 7 MSBuild tasks and targets from the workload, and they should be written so that they support targeting .NET 6 and .NET 7.

(A workload could decide to use the .NET 6 targets, but I wouldn't recommend that because it could be hard to keep multiple SDK versions and multiple workload targets all compatible with each other).

documentation/general/workloads/cross-version-workloads.md
dsplaisted: That's up to the workload.  For .NET, we still support building projects all the way back to .NET Core 1.0.  I understand that for iOS and Android this is not as important, as you need to move forward to new tooling / targeting new versions of .NET in order to continue to publish to the app stores.  So supporting one previous version for iOS and Android is probably fine.

PR #24086, opened 2022-02-24
Remove override into preview version of System.Security.Cryptography.ProtectedData

baronfel:
We've had this in place for about a year, and our version of S.S.C.PD has stuck as a result.  We should not be pinned to a preview version.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -85,7 +85,6 @@
     <!-- Dependencies from https://github.com/dotnet/runtime -->
     <SystemCodeDomPackageVersion>7.0.0-preview.2.22116.9</SystemCodeDomPackageVersion>
     <SystemTextEncodingCodePagesPackageVersion>7.0.0-preview.2.22116.9</SystemTextEncodingCodePagesPackageVersion>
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -85,7 +85,6 @@
     <!-- Dependencies from https://github.com/dotnet/runtime -->
     <SystemCodeDomPackageVersion>7.0.0-preview.2.22116.9</SystemCodeDomPackageVersion>
     <SystemTextEncodingCodePagesPackageVersion>7.0.0-preview.2.22116.9</SystemTextEncodingCodePagesPackageVersion>
-    <SystemSecurityCryptographyProtectedDataPackageVersion>5.0.0-preview.7.20364.11</SystemSecurityCryptographyProtectedDataPackageVersion>
```
joeloff, line 88: Shouldn't we just update this version. Presumably this property is used somewhere else in the code, or not all?

baronfel: We're already doing that in the first version block at the top of this file - this property came from the runtime ages ago, and from what I could tell in their versions.props that don't even have this dependency anymore.

This being set was overriding our top-level 6.0 stable dependency.

baronfel: (See Line 33 of this same file)

joeloff (approved):

PR #24091, opened 2022-02-25
Add EnableAotAnalyzer support into SDK

tlakollo:
Add EnableAotAnalyzer support into SDK
 - Set EnableAotAnalyzer if PublishAot is set
 - Import Microsoft.NET.ILLink.Analyzers.props if EnableAotAnalyzer is set
 - Import ILLink Analyzers if EnableAotAnalyzer is set

Add tests in SDK

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
@@ -64,6 +64,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!-- Enable the trim analyzer when any trimming settings are enabled. Warnings may suppressed based on other settings. -->
     <EnableTrimAnalyzer Condition="'$(EnableTrimAnalyzer)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</EnableTrimAnalyzer>
+    
+    <!-- Enable the AOT analyzer when any AOT settings are enabled. Warnings may supressed based on other settings. -->
+    <EnableAotAnalyzer Condition="'$(EnableAotAnalyzer)' == '' And '$(PublishAot)' == 'true'">true</EnableAotAnalyzer>
 
     <!-- Compiler warning level, defaulted to 4. We promote it to 5 if the user has set analysis level to 5 or higher
          NOTE: at this time only the C# compiler supports warning waves -->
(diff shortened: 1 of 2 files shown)


tlakollo:
I'm creating this as a draft since when solving issues in the tests I realized that I needed to fix https://github.com/dotnet/linker/pull/2650 and https://github.com/dotnet/linker/pull/2655. Now I need to wait for those changes to flow into SDK otherwise the tests will fail. 
@LakshanF you could wait till this PR is merged to continue your work in the runtime or use the command-line instructions to build your own sdk (after building the repo the bits are in `artifacts\bin\redist\Debug\dotnet\dotnet.exe`) and use that to test

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -64,6 +64,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!-- Enable the trim analyzer when any trimming settings are enabled. Warnings may suppressed based on other settings. -->
     <EnableTrimAnalyzer Condition="'$(EnableTrimAnalyzer)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</EnableTrimAnalyzer>
+    
+    <!-- Enable the AOT analyzer when any AOT settings are enabled. Warnings may supressed based on other settings. -->
+    <EnableAotAnalyzer Condition="'$(EnableAotAnalyzer)' == '' And '$(PublishAot)' == 'true'">true</EnableAotAnalyzer>
```
sbomer, line 69: I think maybe we should avoid introducing `PublishAot` until it actually does something, otherwise it could be confusing if you set it and the output isn't AOT.

Maybe just add `EnableAotAnalyzer` in this PR?

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -0,0 +1,141 @@
...
+            var publishCommand = new PublishCommand(Log, Path.Combine(testAsset.TestRoot, testProject.Name));
+            publishCommand
+                .Execute(RuntimeIdentifier)
+                .Should().Pass()
+                .And.NotHaveStdOutContaining("IL2026");
+        }
+
+        [RequiresMSBuildVersionTheory("17.0.0.32901")]
+        [InlineData(ToolsetInfo.CurrentTargetFramework)]
+        public void ILLink_aot_analyzer_can_treat_warnings_as_errors(string targetFramework)
```
sbomer, line 69: I'm not sure it's worth testing warnaserror or nowarn (below) since those are handled the same for every analyzer by Roslyn.

sbomer (approved):

PR #24104, opened 2022-02-25
Workload update timer

### LINKED ISSUE #23760: Workload advertising update timer should be separate for each feature band
By default, when a `dotnet` CLI command that runs restore is run, the CLI will also check for updated workloads if it hasn't done so in the last 24 hours.  This can be disabled by setting a `WORKLOAD_UPDATE_NOTIFY_DISABLE` environment variable to `true`, and the interval can be specified via the `WORKLOAD_UPDATE_NOTIFY_INTERVAL_HOURS` environment variable.  The core code for this is in `WorkloadManifestUpdater.BackgroundUpdateAdvertisingManifestsWhenRequiredAsync`.

Currently, the sentinel file which is used to track the last time that the CLI checked for workload updates is not specific to the SDK feature band: https://github.com/dotnet/sdk/blob/40d6e475f48aa319c46905610ed1dae197608eb0/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs#L459

This means that if different SDK feature bands are being used, the update check for one band may prevent other bands from checking for updates.

We should probably include the feature band in the sentinel file path so that bands don't interfere with each other's update checks.

gkulin:
Fixes #23760

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and contributor`. All rights reserved.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -98,18 +98,18 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
         public async Task BackgroundUpdateAdvertisingManifestsWhenRequiredAsync()
         {
             if (!BackgroundUpdatesAreDisabled() &&
-                AdManifestSentinalIsDueForUpdate() &&
+                AdManifestSentinelIsDueForUpdate() &&
                 UpdatedAdManifestPackagesExistAsync().GetAwaiter().GetResult())
             {
                 await UpdateAdvertisingManifestsAsync(false);
-                var sentinalPath = GetAdvertisingManifestSentinalPath(_sdkFeatureBand);
-                if (File.Exists(sentinalPath))
+                var sentinelPath = GetAdvertisingManifestSentinelPath(_sdkFeatureBand);
+                if (File.Exists(sentinelPath))
                 {
-                    File.SetLastAccessTime(sentinalPath, DateTime.Now);
+                    File.SetLastAccessTime(sentinelPath, DateTime.Now);
                 }
                 else
                 {
-                    File.Create(sentinalPath).Close();
+                    File.Create(sentinelPath).Close();
                 }
             }
         }
@@ -118,7 +118,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
         {
             var installedWorkloads = _workloadRecordRepo.GetInstalledWorkloads(_sdkFeatureBand);
             var updatableWorkloads = GetUpdatableWorkloadsToAdvertise(installedWorkloads);
-            var filePath = GetAdvertisingWorkloadsFilePath();
+            var filePath = GetAdvertisingWorkloadsFilePath(_sdkFeatureBand);
             var jsonContent = JsonSerializer.Serialize(updatableWorkloads.Select(workload => workload.ToString()).ToArray());
             if (Directory.Exists(Path.GetDirectoryName(filePath)))
             {
@@ -129,7 +129,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
         public void DeleteUpdatableWorkloadsFile()
         {
-            var filePath = GetAdvertisingWorkloadsFilePath();
+            var filePath = GetAdvertisingWorkloadsFilePath(_sdkFeatureBand);
             if (File.Exists(filePath))
             {
                 File.Delete(filePath);
@@ -141,7 +141,8 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             try
             {
                 var backgroundUpdatesDisabled = bool.TryParse(Environment.GetEnvironmentVariable(EnvironmentVariableNames.WORKLOAD_UPDATE_NOTIFY_DISABLE), out var disableEnvVar) && disableEnvVar;
-                var adUpdatesFile = GetAdvertisingWorkloadsFilePath(CliFolderPathCalculator.DotnetUserProfileFolderPath);
+                SdkFeatureBand featureBand = new SdkFeatureBand(Product.Version);
+                var adUpdatesFile = GetAdvertisingWorkloadsFilePath(CliFolderPathCalculator.DotnetUserProfileFolderPath, featureBand);
                 if (!backgroundUpdatesDisabled && File.Exists(adUpdatesFile))
                 {
                     var updatableWorkloads = JsonSerializer.Deserialize<string[]>(File.ReadAllText(adUpdatesFile));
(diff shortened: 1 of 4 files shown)


src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -345,12 +339,56 @@ public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(st
...
+            File.GetLastAccessTime(sentinalPath1).Should().BeBefore(updateTime2);
+
+            downloader2.DownloadCallParams.Clear();
+            updater2.BackgroundUpdateAdvertisingManifestsWhenRequiredAsync().Wait();
+            var updateTime1 = DateTime.Now;
+            downloader2.DownloadCallParams.Should().BeEmpty();
+            File.GetLastAccessTime(sentinalPath2).Should().BeBefore(updateTime1);
+        }
+
+        private List<(PackageId, NuGetVersion, DirectoryPath?, PackageSourceLocation)> GetExpectedDownloadedPackages(string sdkFeatureBand)
```
dsplaisted, line 372: We can add a default value to this parameter, and then we don't have to pass it in for most of the call sites.

```suggestion
        private List<(PackageId, NuGetVersion, DirectoryPath?, PackageSourceLocation)> GetExpectedDownloadedPackages(string sdkFeatureBand = "6.0.100")
```

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -345,12 +339,56 @@ public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(st
 
         }
 
+        [Fact]
+        public void TestSideBySideUpdateChecks()
```
dsplaisted, line 343: It would probably be a good idea to add a comment describing what the test basically does, ie checking that different version bands don't interfere with each other's update check timers.

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -345,12 +339,56 @@ public void GivenWorkloadsAreOutOfDateUpdatesAreAdvertisedOnRestoringCommands(st
...
+            var updateTime2 = DateTime.Now;
+
+            downloader1.DownloadCallParams.Clear();
+            updater1.BackgroundUpdateAdvertisingManifestsWhenRequiredAsync().Wait();
+            downloader1.DownloadCallParams.Should().BeEmpty();
+            File.GetLastAccessTime(sentinalPath1).Should().BeBefore(updateTime2);
+
+            downloader2.DownloadCallParams.Clear();
+            updater2.BackgroundUpdateAdvertisingManifestsWhenRequiredAsync().Wait();
+            var updateTime1 = DateTime.Now;
```
dsplaisted, line 367: I think you can delete this line and just use `updateTime2` for the check later on.

dsplaisted (approved):
As we realized while working on this, we also need a similar fix to `GetAdvertisingWorkloadsFilePath`.  That can be part of this PR or separate.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -456,7 +456,7 @@ private IEnumerable<(ManifestId, ManifestVersion)> ParseRollbackDefinitionFile(s
         private bool BackgroundUpdatesAreDisabled() =>
             bool.TryParse(_getEnvironmentVariable(EnvironmentVariableNames.WORKLOAD_UPDATE_NOTIFY_DISABLE), out var disableEnvVar) && disableEnvVar;
 
-        private string GetAdvertisingManifestSentinalPath() => Path.Combine(_userProfileDir, ".workloadAdvertisingManifestSentinal");
+        private string GetAdvertisingManifestSentinalPath(SdkFeatureBand featureBand) => Path.Combine(_userProfileDir, ".workloadAdvertisingManifestSentinal" + featureBand.ToString());
```
joeloff, line 459: You can probably just use interpolation and avoid calling ToString explicitly, `$".workloadAvertisingManifestSentinel{featureband}"` Also, should there be a separator between the filename and version like "." or "-"?

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -43,48 +44,40 @@ public void GivenWorkloadManifestUpdateItCanUpdateAdvertisingManifests()
...
-                .Select(id => ((PackageId, NuGetVersion, DirectoryPath?, PackageSourceLocation))(new PackageId($"{id}.manifest-6.0.100"), null, null, null));
-            nugetDownloader.DownloadCallParams.Should().BeEquivalentTo(expectedDownloadedPackages);
+            nugetDownloader.DownloadCallParams.Should().BeEquivalentTo(GetExpectedDownloadedPackages("6.0.100"));
         }
 
         [Fact]
         public void GivenAdvertisingManifestUpdateItUpdatesWhenNoSentinalExists()
         {
-            (var manifestUpdater, var nugetDownloader, var userProfileDir) = GetTestUpdater();
+            (var manifestUpdater, var nugetDownloader, var sentinalPath) = GetTestUpdater();
```
joeloff, line 53: Should that be sentinel instead of sentinal?

gkulin: Good question. Sentinal was used throughout the files before, but I think it technically should be sentinel... should I go back through and change it everywhere?

joeloff: Maybe a separate PR. It feels wrong to have this glaring typo in the source :)

baronfel: noticed this too: the actual sentinel file name has this same typo, so we may want to consider fixing that as well.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Copyright (c) .NET Foundation and contributor`. All rights reserved.
```
joeloff, line 1: This should be contributors I think. You should be able to confirm looking at the other source files.

src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Microsoft.DotNet.MSBuildSdkResolver.csproj
```diff
@@ -37,9 +37,18 @@
...
+      <_TargetPathRelativePath>..\..\$(TargetFileName)</_TargetPathRelativePath>
+    </PropertyGroup>
+
+    <!-- Path.GetRelativePath doesn't exist on full framework. So we hard-code the relative path, and then if we're running on .NET, we check
+         to make sure the value we hard-coded is the same as we would get by calling Path.GetRelativePath -->
+    <PropertyGroup Condition=" '$(MSBuildRuntimeType)' == 'Core'">
+      <_TargetPathRelativePathCheck>$([System.IO.Path]::GetRelativePath('$(_ResolverManifestDir)','$(TargetPath)'))</_TargetPathRelativePathCheck>
     </PropertyGroup>
 
+    <Error Condition=" '$(MSBuildRuntimeType)' == 'Core' And '$(_TargetPathRelativePath)' != '$(_TargetPathRelativePathCheck)'"
```
joeloff, line 49: @dsplaisted should this error come from a resource so that it can be localized and do we need to assign an SDK error number to it?

joeloff: Ignore my previous comment, this is our project. For some reason I thought this was in the resolver targets.

joeloff (approved):
Apart from the licensing comment I flagged, I think this is good to go. Thank you for fixing the "sentinel" issue.

PR #22072, opened 2021-10-15
Eliminate pre-built assets during source-build

eerhardt:
This work eliminates using pre-built NuGet packages during source-build.

* Ensure Microsoft.CodeAnalysis participates in dependency flow
* Remove unnecessary reference to Microsoft.Win32.Registry, this library is referenced automatically in net6.0 and net472
* Remove hard-coded Roslyn versions in ApiCompatibility and use the Roslyn version used in the SDK
* Clean up the Razor source generator dependency and comment

Contributes to https://github.com/dotnet/source-build/issues/2422

PR #24130, opened 2022-02-28
Pin the compiler version in RSG to an exact version

pranavkm:
Addresses https://github.com/dotnet/sdk/issues/23972

### DIFF
--- a/src/RazorSdk/SourceGenerators/Microsoft.NET.Sdk.Razor.SourceGenerators.csproj
+++ b/src/RazorSdk/SourceGenerators/Microsoft.NET.Sdk.Razor.SourceGenerators.csproj
@@ -11,8 +11,14 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Use a pinned version of the compiler unless building from source. This avoids issues where the source generator is unable to launch in VS because the compiler versions is older than the one in the SDK. -->
+    <RSG_MicrosoftCodeAnalysisCSharpPackageVersion Condition="'$(DotNetBuildFromSource)' != 'true'">$(MicrosoftCodeAnalysisCSharpAnalyzerPinnedVersionPackageVersion)</RSG_MicrosoftCodeAnalysisCSharpPackageVersion>
+    <RSG_MicrosoftCodeAnalysisCSharpPackageVersion Condition="'$(DotNetBuildFromSource)' == 'true'">$(MicrosoftCodeAnalysisCSharpPackageVersion)</RSG_MicrosoftCodeAnalysisCSharpPackageVersion>
+  </PropertyGroup>
+
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(MicrosoftCodeAnalysisCSharpPackageVersion)" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(RSG_MicrosoftCodeAnalysisCSharpPackageVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Razor.SourceGenerator.Tooling.Internal" Version="$(MicrosoftAspNetCoreRazorSourceGeneratorToolingInternalPackageVersion)" GeneratePathProperty="true" />
   </ItemGroup>
 
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -149,6 +149,7 @@
   <PropertyGroup Label="Manually updated">
     <!-- Dependencies from https://github.com/microsoft/MSBuildLocator -->
     <MicrosoftBuildLocatorPackageVersion>1.4.1</MicrosoftBuildLocatorPackageVersion>
+    <MicrosoftCodeAnalysisCSharpAnalyzerPinnedVersionPackageVersion>4.0.1</MicrosoftCodeAnalysisCSharpAnalyzerPinnedVersionPackageVersion>
   </PropertyGroup>
   <PropertyGroup>
     <!-- Dependencies from https://github.com/dotnet/xliff-tasks -->


src/RazorSdk/SourceGenerators/Microsoft.NET.Sdk.Razor.SourceGenerators.csproj
```diff
@@ -11,8 +11,14 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <PropertyGroup>
```
pranavkm, line 14: @eerhardt \ @MichaelSimons wanted to get your thoughts on this, as an enhancement to https://github.com/dotnet/sdk/pull/22072/files. Users are often affected by this when running a fairly new SDK using MSBuild where the compiler hasn't been updated as yet.

eerhardt: I think this is a bigger issue / question for the SDK and Roslyn teams.  cc @jaredpar @chsienki @dsplaisted 

So the repro steps here are:

1. Install VS 2022, but not the latest version
2. Install .NET Core SDK 6.0.200

and now we don't have a coherent SDK + VS?  There are components in the 6.0.200 SDK that reference Roslyn `4.1`, but the VS 2022 installed only has Roslyn `4.0`?

Is that all true?

IMO - we shouldn't be using different version when building from source vs. building for Microsoft distribution. That's a recipe for disaster.

pranavkm: Interestingly, this appears to be the path the ApiCompatibility tool has taken - https://github.com/dotnet/sdk/pull/23321/files

eerhardt: Agreed - I wasn't excited about that change either 😄. The difference with the ApiCompatibility tool is that it also ships OOB from the SDK into it's own NuGet package:

https://www.nuget.org/packages/Microsoft.DotNet.Compatibility

So someone could be using an older SDK, and PackageReference the newer version of that NuGet package.

Correct me if I'm wrong, but I believe that the Razor Source Generator only ships as part of the SDK, right? Shouldn't we have a guaranteed version of Roslyn that the SDK supports?

pranavkm: > Correct me if I'm wrong, but I believe that the Razor Source Generator only ships as part of the SDK, right? 

That is right.

> Shouldn't we have a guaranteed version of Roslyn that the SDK supports?

👍🏽

jaredpar: Pinning the compiler means that you're not getting bug fixes, features, etc ... 

It's less important for a tool like API Compat which is using us for a limited use case. Mostly to grab a semantic model to inspect the code. Much more important for components that are actually participating in mainline compilation. 

In this case though this is not pinning the compiler, it's pinning the Compiler API version Razor is compiling against. At runtime we're going to use the MS.CA that ships with the compiler. Still though this feels very unnecessary.

pranavkm: @jaredpar do you have suggestions on how we should proceed here? Users can unblock themselves by updating their VS / pinning the SDK, but both of these seem like crummy option.

Would we be open to taking this PR for 6.0.x and consider something more lasting for 7.0?

eerhardt: Is this the only analyzer / source generator that actually ships in box with the SDK? I know that the roslyn-analyzer pin their version way back to something like `3.3`. 

We could follow that path here, and then for source-build, we need to put the `ref assemblies` for the version you pin in the "SBRP".

https://github.com/dotnet/roslyn-analyzers/blob/eb9a7a5d2e4dcbfd4db89c2ee48e79bca107484c/eng/Versions.props#L45-L57

pranavkm: > Is this the only analyzer / source generator that actually ships in box with the SDK?

ASP.NET Core builds some analyzers that ship as part of the WebSDK. Those are currently pinned to build against 3.3:

https://github.com/dotnet/aspnetcore/blob/main/eng/Versions.props#L198-L199

SteveSandersonMS:
@pranavkm Who would be most relevant to take on ownership of this PR (or the underlying issue) after this week? It sounds like @eerhardt or @jaredpar have the most context, and perhaps views on the correct way to proceed.

src/RazorSdk/SourceGenerators/Microsoft.NET.Sdk.Razor.SourceGenerators.csproj
chsienki: Opened https://github.com/dotnet/source-build-reference-packages/pull/357 to add the required bits to SBRP so we can target the same version in source build.

jaredpar (approved):

jaredpar:
@RikkiGibson, @jcouv FYI

RikkiGibson:
I didn't think this was a supported scenario--using a VS with an older version of the compiler than what's in the SDK. Did I misunderstand something?

chsienki:
@RikkiGibson We're still defining the exact support policy, but it is expected to work in some scenarios. In this case the generator isn't actually using any of the new features, it's just picking up a ref to latest MS.CA by default, so we want to change it so it will continue to load and work in downstream compilers.

RikkiGibson (approved):
Thanks for clarifying.

Issue #23795, opened 2022-02-07
Detect if the SDK is being run on a CI host machine


baronfel:
### Is your feature request related to a problem? Please describe.

Certain functionalities in the SDK could be toggled on or off more intelligently if we knew that the SDK was running on a Continuous Integration host, like a GitLab Runner agent, or a GitHub actions runner.  These systems typically expose environment variables that can be used to determine if a CI agent is active.

### Describe the solution you'd like
We should create a list of such variables, and synthesize a boolean `IsCI` (or similar) that is true if any of these environment variables are populated.  These variables are:

**Boolean values**
* TF_BUILD
* GITHUB_ACTIONS
* APPVEYOR
* CI
* TRAVIS
* CIRCLECI

**Non-null values**
* TEAMCITY_VERSION
* JB_SPACE_API_URL

**Complex checks**
* AWS CloudBuild - CODEBUILD_BUILD_ID != '' and AWS_REGION != ''
* GCP Cloud Build - BUILD_ID != '' and PROJECT_ID != ''

This is not an exhaustive list, but should be enough to get going.  We can/should also document the `CI` variable for user-hosted CI systems like an in-house build server farm.

KalleOlaviNiemitalo:
In one repo, I had to stop our CI system from setting the CI variable because it triggered a warnings-as-errors mode in a nodejs build system and we weren't ready for that. The same repo included .NET code as well. So the CI variable alone would not be reliable for telemetry.

baronfel:
Duly noted! This is definitely a heuristic, so we'll need that caveat when looking at the data. I do think it's a heuristic that's a reasonable approximation of the reality, though.

baronfel:
Ah, changing from a short circuiting model, I see. That would definitely be more correct 👍

baronfel:
Implemented by #24166

PR #24166, opened 2022-03-01
Detect if a particular execution of the SDK is on a CI system or not

baronfel:
This is an implementation of #23795 following the established patterns. I've added a test verifying the new field is populated, and attempted to fast-exit from the implementation. ~It _may_ also be useful to cache or lazily compute the value, since each telemetry event would add this new field. Repeated access should be zero-cost in the ideal case.~  Since this is only ever invoked once per invocation of the `dotnet` command, then cached into a reuseable Dictionary, no caching or laziness needs to be done in the implementation.

When this merges, we'll need to update [this list of collected data](https://docs.microsoft.com/en-us/dotnet/core/tools/telemetry#data-points) with a brief description of the new data point.

### DIFF
--- a/src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
+++ b/src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
@@ -9,7 +9,7 @@ namespace Microsoft.DotNet.Cli.Telemetry
 {
     internal class CIEnvironmentDetectorForTelemetry : ICIEnvironmentDetector
     {
-        // variables that will hold boolean values only
+        // Systems that provide boolean values only, so we can simply parse and check for true
         private static readonly string[] _booleanVariables = new string[] {
             // Azure Pipelines - https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables#system-variables-devops-services
             "TF_BUILD",
@@ -26,6 +26,7 @@ namespace Microsoft.DotNet.Cli.Telemetry
             "CIRCLECI",
         };
 
+        // Systems where every variable must be present and not-null before returning true
         private static readonly string[][] _allNotNullVariables = new string[][] {
             // AWS CodeBuild - https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
             new string[]{ "CODEBUILD_BUILD_ID", "AWS_REGION" },
@@ -35,6 +36,7 @@ namespace Microsoft.DotNet.Cli.Telemetry
             new string[]{ "BUILD_ID", "PROJECT_ID" }
         };
 
+        // Systems where the variable must be present and not-null
         private static readonly string[] _ifNonNullVariables = new string[] {
             // TeamCity - https://www.jetbrains.com/help/teamcity/predefined-build-parameters.html#Predefined+Server+Build+Parameters
             "TEAMCITY_VERSION",
@@ -62,7 +64,7 @@ namespace Microsoft.DotNet.Cli.Telemetry
 
             foreach (var variable in _ifNonNullVariables)
             {
-                if (Environment.GetEnvironmentVariable(variable) is {} v)
+                if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(variable)))
                 {
                     return true;
                 }
--- a/src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
+++ b/src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
@@ -1,0 +1,65 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Cli.Telemetry
+{
+    internal class CIEnvironmentDetectorForTelemetry : ICIEnvironmentDetector
+    {
+        // variables that will hold boolean values
+        private static readonly string[] _booleanVariables = new string[] {
+            "TF_BUILD", // Azure Pipelines
+            "GITHUB_ACTIONS", // GitHub Actions
+            "APPVEYOR", // AppVeyor
+            "CI", // a general-use flag
+            "TRAVIS", // Travis CI
+            "CIRCLECI", // CircleCI
+        };
+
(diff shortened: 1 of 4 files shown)


src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
```diff
@@ -0,0 +1,65 @@
...
+                if(Environment.GetEnvironmentVariable(variable) is {} v) {
+                    return Cli.Telemetry.IsCIEnvironment.True;
+                }
+            }
+
+            return Cli.Telemetry.IsCIEnvironment.False;
+        }
+    }
+
+    internal enum IsCIEnvironment
```
dsplaisted, line 60: Why define a new enum here instead of just using the built-in Boolean type?

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            foreach (var variables in _allNotNullVariables) {
+                if (variables.All((variable) => !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(variable))))
+                {
+                    return Cli.Telemetry.IsCIEnvironment.True;
+                }
+            }
+
+            foreach (var variable in _ifNonNullVariables) {
+                if(Environment.GetEnvironmentVariable(variable) is {} v) {
```
dsplaisted, line 51: I'm not super familiar with some of the new C# syntax, but I believe this is a pattern match to check that something's not null.  What's the reason to use this syntax here but to use `string.IsNullOrEmpty` for the check of `_allNotNullVariables`?

dsplaisted (approved):
Have we done any manual sanity testing that this works as expected in the different CI systems (and outside of them)?

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            foreach (var booleanVariable in _booleanVariables)
+            {
+                if (bool.TryParse(Environment.GetEnvironmentVariable(booleanVariable), out bool envVar) && envVar)
+                {
+                    return Cli.Telemetry.IsCIEnvironment.True;
+                }
+            }
+
+            foreach (var variables in _allNotNullVariables) {
```
joeloff, line 43: nit: open braces on new line

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
joeloff: +1, am also curious

src/Cli/dotnet/Telemetry/TelemetryCommonProperties.cs
```diff
@@ -54,6 +57,7 @@ internal class TelemetryCommonProperties
 
         private const string MachineIdCacheKey = "MachineId";
         private const string IsDockerContainerCacheKey = "IsDockerContainer";
+        private const string CI = "Continuous Integration";
```
joeloff, line 60: Should the space be removed? It seems to serve as a key and it appears we don't use spaces for these  (looking at MachineId and IsDockerContainer)

joeloff (approved):

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
baronfel: Initially I thought there might be a use case for Unknown, like in the docker detection, but in hindsight this is a more binary question. I'll simplify it.

src/Cli/dotnet/Telemetry/TelemetryCommonProperties.cs
dsplaisted: I had the same question but some of the other keys do have spaces.

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
baronfel: In my testing, empty string reports as null from this API so I felt ok with this choice:

```fsharp
11:41:05 ❯ $env:Empty=""; $env:Present="hi"; dotnet fsi

Microsoft (R) F# Interactive version 12.0.1.0 for F# 6.0
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> open type System.Environment;;
> GetEnvironmentVariable "Empty";;
val it: string = <null>

> GetEnvironmentVariable "Present";;
val it: string = "hi"

> GetEnvironmentVariable "Absent";;
val it: string = <null>
```

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Cli.Telemetry
+{
+    internal class CIEnvironmentDetectorForTelemetry : ICIEnvironmentDetector
```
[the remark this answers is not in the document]
baronfel: I'll add links to documentation for each provider, but my understanding is that they are uniform across platforms. I took these directly from DotNet.ReproducibleBuilds, so if I find that's not the case then I'll submit an issue on that repo to keep us uniform.

src/Cli/dotnet/Telemetry/TelemetryCommonProperties.cs
baronfel: This is a good suggestion, I'll do so.

baronfel: _Actually_, the telemetry property keys all use spaces. The keys without spaces are local cache keys only. I'll move this key up with the other telemetry property keys to reduce visual confusion.

src/Cli/dotnet/Telemetry/CIEnvironmentDetectorForTelemetry.cs
joeloff: I don't think so. I only use VS and Ctrl+K+D has been ingrained over the years. We should chat about this.

PR #24168, opened 2022-03-02
deploy common roslyn assemblies used by multiple dotnet tools

jmarolf:
consolidates deployment of roslyn assemblies for several global tools


pranavkm:
@jmarolf we could have watch re-use these assemblies too. It's already set up to re-use the binaries from the `Roslyn\bincore` path so this would be an extension of that: https://github.com/dotnet/sdk/blob/main/src/BuiltInTools/dotnet-watch/Program.cs#L408-L421

jmarolf:
@pranavkm thats my hope. We really should just be loading the same binaries that are versioned with the compiler. Was going to start with dotnet-format and verify that worked but once this gets merged would be happy to ensure this works for dotnet-watch.

PR #24237, opened 2022-03-05
Move WebAssembly ProjectCapability to props file

pranavkm:
Both VS and dotnet-watch have heuristics that assumes that an ASP.NET Core app referencing
a BlazorWebAssembly app is a Blazor hosted app for hot reload purposes. They then attempt
to send deltas to the browser to update the WASM app.

We've seen a few cases of users attempting to build dual-mode Blazor apps where the client is
a Blazor WebAssembly app, but during development the app references blazor.server.js. In this event,
applying the delta results in a warning with no simple way for a user to tell VS / watch to
not treat this as a Blazor hosted app.

Detecting the client as Blazor WebAssembly is based on ProjectCapability. Removing this capability
should solve the issue here and we think this was a sufficient solution at this time. However,
the ProjectCapability is defined within the targets file which makes it difficult to remove it
from within the project. This change moves it to the props making that easier.

Contributes to https://github.com/dotnet/aspnetcore/issues/40527


mkArtakMSFT (approved):
Thanks @pranavkm!

PR #24253, opened 2022-03-07
Update CODEOWNERS

TanayParikh:
No commas required: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -12,23 +12,23 @@
 /src/Tests/Microsoft.NET.Sdk.Publish.Tasks.Tests/ @vijayrkn
 
 # Area-AspNetCore
-/src/BlazorWasmSdk/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Assets/TestProjects/BlazorHosted/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Assets/TestProjects/BlazorHostedRID/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Assets/TestProjects/BlazorWasmMinimal/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Assets/TestProjects/BlazorWasmWithLibrary/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/RazorSdk/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.AoT.Tests/ @pranavkm, @mkArtakMSFT, @TanayParikh, @javiercn
-/src/Tests/Microsoft.NET.Sdk.Razor.Tests/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Tests/Microsoft.NET.Sdk.Razor.Tool.Tests/ @captainsafia, @pranavkm, @mkArtakMSFT
-/src/Assets/TestPackages/PackageLibraryDirectDependency/ @captainsafia, @pranavkm, @mkArtakMSFT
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -62,13 +62,13 @@
...
+/src/BuiltInTools/   @captainsafia @mkArtakMSFT
 
 # Compatibility tools owned by runtime team
 # Area-Compatibility
-/src/Compatibility/ @Anipik, @safern, @ericstj
-/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/ @Anipik, @safern, @ericstj
-/src/Tests/Microsoft.DotNet.PackageValidation.Tests/ @Anipik, @safern, @ericstj
+/src/Compatibility/ @Anipik @safern @ericstj
+/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/ @Anipik @safern @ericstj
+/src/Tests/Microsoft.DotNet.PackageValidation.Tests/ @Anipik @safern @ericstj
```
ericstj, line 74:
```suggestion
/src/Compatibility/ @ericstj @dotnet/area-infrastructure-libraries @joperezr 
/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/ @ericstj @dotnet/area-infrastructure-libraries @joperezr 
/src/Tests/Microsoft.DotNet.PackageValidation.Tests/ @ericstj  @dotnet/area-infrastructure-libraries @joperezr 
```
Can we use this?  cc @joperezr @jeffhandley

jeffhandley: Works for me. Alternatively, if @joperezr wants to be considered a "Consultant" for `area-infrastructure-libraries` and would be OK with receiving the notifications, we could add him to the team and use the team membership as the single source of truth. But that has the tradeoff of also getting notifications for issues/prs labeled with the area, which might not be desired.

ericstj: This is for our package validation tasks in dotnet/sdk only, whereas area-infrastructure-libraries covers a broader set of things the team owns.

TanayParikh: Thanks, committed this in to see if it resolves the errors. Feel free to add further suggestions if necessary.

TanayParikh: Looks like `dotnet/area-infrastructure-libraries` is missing access perms.

TanayParikh:
@AntonLapounov do you know why GitHub may be giving an error for clrappmodel@microsoft.com.

CODEOWNERS
joperezr: > Alternatively, if @joperezr wants to be considered a "Consultant" for area-infrastructure-libraries and would be OK with receiving the notifications,

I'm fine with whatever you think is the better approach, I don't mind getting the extra notifications and filtering down only to the ones where my input is relevant.

ericstj: Is there a way to retrigger that check?  I just changed the parent team of https://github.com/orgs/dotnet/teams/area-infrastructure-libraries to be npt which has write access to dotnet/sdk.  I think this should satisfy it.

TanayParikh: There are a bunch of failing checks (essentially every one involving a team). I'm not sure if this is actually indicative of missing perms or if the check simply doesn't take into account transitive access. I think the check is a fairly new feature, so it may be the latter.

AntonLapounov:
> Unknown owner on line 55: make sure a user with the email address clrappmodel@microsoft.com exists and has write access to the repository
…/Microsoft.NET.ILLink.targets clrappmodel@microsoft.com

clrappmodel@microsoft.com is a team handle, which has no write access to the repository. GitHub might need to relax this owner verification.

CODEOWNERS
ericstj: It's possible my change hasn't yet taken effect.  It's still "pending" in the github UI now that I check it.  IMO it's still better to use a team here so we don't run into cases where it gets stale like it did before.

ericstj: I got that team re-parenting approved and it seems to have fixed it.

TanayParikh:
@marcpopMSFT any objections to merging this in? Remaining errors are group related and seems to be due to GitHub just not tracking transitive access.

PR #24288, opened 2022-03-08
Error using skip-manifest-update and --from-rollback-file

### LINKED ISSUE #24023: Report an error when using --skip-manifest-update and --from-rollback-file
We added rollback files to the update command and skip manifest update to the install command. Later we add rollback to the install command but the two switches don't work well together.

We should trigger an error or at least a warning as --skip-manifest-update appear to take precedence.

gkulin:


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -145,6 +146,10 @@ public override int Execute()
                     throw new GracefulException(string.Format(LocalizableStrings.WorkloadCacheDownloadFailed, e.Message), e, isUserError: false);
                 }
             }
+            else if (_skipManifestUpdate && usedRollback)
+            {
+                throw new GracefulException(LocalizableStrings.CannotCombineSkipManifestAndRollback, isUserError: true);
```
joeloff, line 151: Can we pass the two options names and use `string.format` when constructing the error? That way, if there's ever a rename (probably unlikely), we don't have to update the error message. And it also avoids issues if the translators make a mistake with a resource string.

gkulin: Just updated that, thanks for pointing it out!

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -51,6 +51,21 @@ public void GivenWorkloadInstallItErrorsOnFakeWorkloadName()
...
+        {
+            var command = new DotnetCommand(Log);
+            command
+                .WithEnvironmentVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR", string.Empty)
+                .WithEnvironmentVariable("PATH", "fake")
+                .Execute("workload", "install", "maui", "--skip-manifest-update", "--from-rollback-file", "foo.txt")
+                .Should()
+                .Fail()
+                .And
+                .HaveStdErrContaining(Workloads.Workload.Install.LocalizableStrings.CannotCombineSkipManifestAndRollback);
```
dsplaisted, line 65: This probably needs to be updated to check for the formatted version of the string.

dsplaisted (approved):

joeloff (approved):

src/Tests/dotnet-sln.Tests/GivenDotnetSlnAdd.cs
```diff
@@ -1162,12 +1161,13 @@ public void WhenSolutionFolderAndInRootIsPassedItFails()
                 .WithWorkingDirectory(solutionDirectory)
                 .Execute($"sln", "App.sln", "add", "--solution-folder", "blah", "--in-root", projectToAdd);
             cmd.Should().Fail();
-            cmd.StdErr.Should().Be(Microsoft.DotNet.Tools.Sln.LocalizableStrings.SolutionFolderAndInRootMutuallyExclusive);
-            cmd.StdOut.Should().BeVisuallyEquivalentToIfNotLocalized(HelpText(solutionDirectory));
+            cmd.StdOut.Should().BeVisuallyEquivalentToIfNotLocalized("");
+            //cmd.StdErr.Should().Be(Microsoft.DotNet.Tools.Sln.LocalizableStrings.SolutionFolderAndInRootMutuallyExclusive);
```
dsplaisted, line 1165: This test looks like it still needs to be updated.

dsplaisted (approved):

PR #24333, opened 2022-03-11
Properly handle empty solution and project paths in SDK resolver

### LINKED ISSUE #24320: MSBuildSdkResolver throws ArgumentException when solution path is an empty string
### Describe the bug
Reported as part of https://github.com/dotnet/msbuild/issues/7453

When an MSBuild project SDK is being resolved in some cases (ie Visual Studio), the `SdkResolverContext` has an empty string for the solution path.  The MSBuildSdkResolver tries to call `Path.GetDirectoryName()` on an empty string and an `ArgumentException` is thrown.  The code is trying to fall back to using the project path but is only checking for `null`, not an empty string.

image

image

https://github.com/dotnet/sdk/blob/b69e8f53c6f65fc8f26fefc0cbff03725ee0f8ba/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/MSBuildSdkResolver.cs#L91

### To Reproduce

Evalute an MSBuild project file with a solution path as an empty string.

### Further technical details
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and its version

jeffkl:
Fixes #24320

In some cases, MSBuild project evaluations are in-memory and so no path context is passed to the SDK resolver.  Additionally, some evaluations in Visual Studio specify an empty string for the solution file path which causes the SDK resolver to throw an `ArgumentException` error when passing that to `Path.GetDirectoryName()`.

This change checks for `null` and empty/whitespace values to know that the solution file path shouldn't be used.  Additionally, if an in-memory project is being evaluated and there is no project file path to fall back to, the `Environment.CurrentDirectory` is used instead as a best effort starting directory for `global.json`.  This could result in the wrong `global.json` being used but its more likely that no `global.json` will be found and SDK resolution will work like normal.

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/MSBuildSdkResolver.cs
+++ b/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/MSBuildSdkResolver.cs
@@ -88,7 +88,7 @@ namespace Microsoft.DotNet.MSBuildSdkResolver
             if (msbuildSdksDir == null)
             {
                 dotnetRoot = EnvironmentProvider.GetDotnetExeDirectory(_getEnvironmentVariable);
-                string globalJsonStartDir = Path.GetDirectoryName(context.SolutionFilePath ?? context.ProjectFilePath);
+                string globalJsonStartDir = GetGlobalJsonStartDir(context);
                 var resolverResult = _netCoreSdkResolver.ResolveNETCoreSdkDirectory(globalJsonStartDir, context.MSBuildVersion, context.IsRunningInVisualStudio, dotnetRoot);
 
                 if (resolverResult.ResolvedSdkDirectory == null)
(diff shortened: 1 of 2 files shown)


dsplaisted (approved):
Looks good.  If you want this to ship in 17.2, then it needs to be retargeted to the release/6.0.3xx branch.

jeffkl:
@drewnoakes do you have a sense of how urgent this fix is?  It will pretty much fix the bug reported in https://github.com/dotnet/msbuild/issues/7453, not sure if you were seeing it a lot?

drewnoakes:
How hard would it be to retarget this to 17.2? 17.2 is LTS, so it would be good to fix this.

jeffkl:
I have retargeted to `release/6.0.3xx` for now to see if we can get this for 17.2/6.0.3xx

jeffkl:
@dsplaisted Let me know if you want to get this in 6.0.3xx

Issue #22353, opened 2021-10-26
Port PackageValidation flaky build failure fixes to 6.0


dagood:
* https://github.com/dotnet/runtime/pull/60847
* https://github.com/dotnet/sdk/pull/22277

This would help source-build reliability in particular (https://github.com/dotnet/source-build/issues/2501) and allow removing this workaround:

* https://github.com/dotnet/runtime/pull/60881

ericstj:
@safern @dagood can this be closed, or have these not yet been ported to 6.0?  Adjusting to 6.0.2xx

dagood:
Seems like it can be closed. @lbussell @MichaelSimons correct me if I'm wrong.

ericstj:
FWIW I don't see @safern's fixes in 6.0 codebase.

dagood:
Ah, yep, I must have misread some PRs.

joperezr:
This was fixed by https://github.com/dotnet/runtime/pull/66635

PR #24339, opened 2022-03-11
Improve loading of Roslyn assemblies on validate package task (#22277)

joperezr:
Contributes to https://github.com/dotnet/sdk/issues/22353

Porting #22277 to release/6.0

cc: @carlossanlop @ericstj @dagood 

After this change is in and we have a build for it, I will go ahead and create a PR in runtime to update the version of PackageValidation we use.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
+++ b/src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
@@ -6,7 +6,7 @@
     <IsPackable>true</IsPackable>
     <IsShippingPackage>true</IsShippingPackage>
     <StrongNameKeyId>Open</StrongNameKeyId>
-    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionPrefix>1.1.0</VersionPrefix>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <!-- This package doesnot contain any lib or ref assemblies because its a tooling package.-->
     <NoWarn>$(NoWarn);NU5128</NoWarn>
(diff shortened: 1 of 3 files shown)


carlossanlop (approved):
Direct backport. LGTM.

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
```diff
@@ -6,7 +6,7 @@
     <IsPackable>true</IsPackable>
     <IsShippingPackage>true</IsShippingPackage>
     <StrongNameKeyId>Open</StrongNameKeyId>
-    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionPrefix>1.1.0</VersionPrefix>
```
ericstj, line 9: Separately we were discussing just calculating this based on the repo branding, is that possible?  Maybe we can address the stabilization issue in this PR too.

dagood:
> cc: @dagood

I'm not on the .NET source-build team anymore: +cc @dotnet/source-build-internal

src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
joperezr: We could do that, but I thought that we said we were going to do that on main. The only problem about doing that here, is that main today generates builds that are 2.0.0 so making that change here will start generating versions that are greater (in NuGet's comparisons) to main branch. If that's ok with you, I'd rather make that version change first in main branch, and then cherry pick back if we think it is necessary.

PR #24362, opened 2022-03-14
Pin Microsoft.CodeAnalysis version for API Compat to the 4.0.x stable package

joperezr:
Porting #23321 to 6.0.3xx in order to have PackageValidation depend on a lower version of Roslyn in order to support being loaded by desktop MSBuild and older VS.

cc: @ericstj @carlossanlop

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
@@ -5,10 +5,14 @@
     <StrongNameKeyId>Open</StrongNameKeyId>
     <!-- We need to compare ISymbols in a special way (by name) and roslyn symbol comparers take more heuristics into consideration.-->
     <NoWarn>RS1024;$(NoWarn)</NoWarn>
+    <!-- We pin the code analysis version when not in source build as we need to support running on older
+    SDKs when the OOB package is used. -->
+    <_MicrosoftCodeAnalysisVersion>4.0.1</_MicrosoftCodeAnalysisVersion>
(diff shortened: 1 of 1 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -5,10 +5,14 @@
     <StrongNameKeyId>Open</StrongNameKeyId>
     <!-- We need to compare ISymbols in a special way (by name) and roslyn symbol comparers take more heuristics into consideration.-->
     <NoWarn>RS1024;$(NoWarn)</NoWarn>
+    <!-- We pin the code analysis version when not in source build as we need to support running on older
+    SDKs when the OOB package is used. -->
+    <_MicrosoftCodeAnalysisVersion>4.0.1</_MicrosoftCodeAnalysisVersion>
```
ericstj, line 10: Do we know if this is the right version to select?  How does it differ from 4.0.0 which seems like the version we used at 6.0 GA: https://github.com/dotnet/sdk/blob/2c798cbf6bf259ce49d94d91bc665a95d42d7a7f/eng/Versions.props#L125-L129

joperezr: Not sure if 4.0.0 existed, but if so it is delisted from NuGet: https://www.nuget.org/packages/Microsoft.CodeAnalysis/4.0.1

I picked 4.0.1 primarily for two reasons:
1. It is the version we use in main (since it is the change that matches the cherry-picked commit)
2. It is the latest (and only) stable version of the 4.0.x band.

ericstj: Ok, I took a look at 4.0.1 and found the assembly version is still 4.0.0.0 which matches what we built against at GA, so my concern is moot.

ericstj (approved):

PR #24426, opened 2022-03-17
Update feature flag logic to disable only semantics

MarcoRossignoli:
Update feature flag logic to disable only semantics

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -92,7 +92,7 @@ namespace Microsoft.DotNet.Tools.Test
             // one more time, there is no extra hop via msbuild
             convertedArgs.AddRange(settings);
 
-            if (!FeatureFlag.Instance.IsDisabled(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
+            if (!FeatureFlag.Instance.IsSet(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
             {
                 // Add artifacts processing mode and test session id for the artifact post-processing
                 convertedArgs.Add("--artifactsProcessingMode-collect");
@@ -141,7 +141,7 @@ namespace Microsoft.DotNet.Tools.Test
                 }
             }
 
-            if (!FeatureFlag.Instance.IsDisabled(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
+            if (!FeatureFlag.Instance.IsSet(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
             {
                 // Add artifacts processing mode and test session id for the artifact post-processing
                 msbuildArgs.Add("-property:VSTestArtifactsProcessingMode=collect");
@@ -173,7 +173,7 @@ namespace Microsoft.DotNet.Tools.Test
 
         internal static int RunArtifactPostProcessingIfNeeded(string testSessionCorrelationId, ParseResult parseResult, FeatureFlag disableFeatureFlag)
         {
-            if (disableFeatureFlag.IsDisabled(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
+            if (disableFeatureFlag.IsSet(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
             {
                 return 0;
             }
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -26,7 +26,7 @@ namespace Microsoft.DotNet.Tools.Test
         {
             parseResult.HandleDebugSwitch();
 
-            FeatureFlag.Default.PrintFlagFeatureState();
+            FeatureFlag.Instance.PrintFlagFeatureState();
 
             // We use also current process id for the correlation id for possible future usage in case we need to know the parent process
             // from the VSTest side.
@@ -70,7 +70,7 @@ namespace Microsoft.DotNet.Tools.Test
                 int exitCode = FromParseResult(parseResult, settings, testSessionCorrelationId).Execute();
 
                 // We run post processing also if execution is failed for possible partial successful result to post process.
-                exitCode |= RunArtifactPostProcessingIfNeeded(testSessionCorrelationId, parseResult, FeatureFlag.Default);
+                exitCode |= RunArtifactPostProcessingIfNeeded(testSessionCorrelationId, parseResult, FeatureFlag.Instance);
 
                 return exitCode;
             }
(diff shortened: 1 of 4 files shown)


src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -92,7 +92,7 @@ private static int ForwardToVSTestConsole(ParseResult parseResult, string[] args
             // one more time, there is no extra hop via msbuild
             convertedArgs.AddRange(settings);
 
-            if (FeatureFlag.Default.IsEnabled(FeatureFlag.ARTIFACTS_POSTPROCESSING))
+            if (!FeatureFlag.Instance.IsDisabled(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
```
nohwnd, line 95: I missed this in the other review. I would rename IsDisabled to IsSet, or IsTrue, that way it is not confusing whether or not it means DISABLE_ARTIFACTS_POSTPROCESSING is Disabled = isEnabled. Or DISABLE_ARTIFACTS_POSTPROCESSING isDisabled = isEnabled.

```suggestion
            if (!FeatureFlag.Instance.IsSet(FeatureFlag.DISABLE_ARTIFACTS_POSTPROCESSING))
```

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -9,35 +9,35 @@ namespace Microsoft.DotNet.Tools.Test
     // !!! FEATURES MUST BE KEPT IN SYNC WITH https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs !!!
     internal class FeatureFlag
     {
-        private const string Prefix = "VSTEST_FEATURE_";
+        private static readonly Dictionary<string, bool> FeatureFlags = new();
```
nohwnd, line 12: Make it dictionary of nullable bools, and only use it for lookups, so we query the environment only once per flag usage.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -9,35 +9,35 @@ namespace Microsoft.DotNet.Tools.Test
...
-        public FeatureFlag()
+        static FeatureFlag()
         {
-            FeatureFlags.Add(ARTIFACTS_POSTPROCESSING, true);
+            FeatureFlags.Add(DISABLE_ARTIFACTS_POSTPROCESSING, false);
         }
 
-        // Added for artifact porst-processing, it enable/disable the post processing.
+        // Added for artifact post-processing, it enable/disable the post processing.
         // Added in 17.2-preview 7.0-preview
```
nohwnd, line 24:
```suggestion
        // Added for artifact post-processing, it enable/disable the post processing.
        // Added in TP 17.2-preview, .NET 7.0-preview, disables additional artifact post-processing, such as combining code coverage files into one file.
```

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -9,35 +9,35 @@ namespace Microsoft.DotNet.Tools.Test
...
+        private const string VSTEST_ = nameof(VSTEST_);
 
-        public static FeatureFlag Default { get; } = new FeatureFlag();
+        public static FeatureFlag Instance { get; } = new FeatureFlag();
 
-        public FeatureFlag()
+        static FeatureFlag()
         {
-            FeatureFlags.Add(ARTIFACTS_POSTPROCESSING, true);
+            FeatureFlags.Add(DISABLE_ARTIFACTS_POSTPROCESSING, false);
```
nohwnd, line 20: Do not initialize here, we don't know what the user preference is and we have to look it up from the environment when the flag is used in code.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -9,35 +9,35 @@ namespace Microsoft.DotNet.Tools.Test
...
+        public const string DISABLE_ARTIFACTS_POSTPROCESSING = VSTEST_ + "_" + nameof(DISABLE_ARTIFACTS_POSTPROCESSING);
 
         // For now we're checking env var.
         // We could add it also to some section inside the runsettings.
-        public bool IsEnabled(string featureName) =>
-            int.TryParse(Environment.GetEnvironmentVariable(featureName), out int enabled)
-                ? enabled == 1
-                : FeatureFlags.TryGetValue(featureName, out bool isEnabled) && isEnabled;
+        public bool IsDisabled(string featureName) =>
+            int.TryParse(Environment.GetEnvironmentVariable(featureName), out int disabled) ?
```
nohwnd, line 30: Here, we should cache the value. 

Check if we have the value in FeatureFlags, and use the value from there. If not, look it up from environment. If the variable is not set, or if the value is trimmed and equal to "0", then isSet is false. Otherwise it is true.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
MarcoRossignoli: I think we can remove the dictionary, now that we have default always `false` semantics we can only check the env var if present and the value, if it's not set is false.

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -9,35 +9,35 @@ namespace Microsoft.DotNet.Tools.Test
     // !!! FEATURES MUST BE KEPT IN SYNC WITH https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs !!!
```
Evangelink, line 9: It says it must be kept in sync but do you mean like exact copy/paste or having same level of features? We are not implementing `IFeatureFlag` for example.

Also we are not porting the `DISABLE_ARTIFACTS_POSTPROCESSING_NEW_SDK_UX` feature.

MarcoRossignoli: I meant with flag names where pertinent...not all env vars are used here, we can copy past everything anyway to avoid confusion

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
MarcoRossignoli: I'll remove the dictionary

src/Cli/dotnet/commands/dotnet-test/VSTestFeatureFlag.cs
```diff
@@ -2,44 +2,38 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace Microsoft.DotNet.Tools.Test
 {
-    // !!! FEATURES MUST BE KEPT IN SYNC WITH https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs !!!
+    // !!! USED FEATURES NAME MUST BE KEPT IN SYNC WITH https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs !!!
```
nohwnd, line 10: The original comment was better. If you are set on this one, fix the grammar :)

```suggestion
    // !!! USED FEATURE NAMES MUST BE KEPT IN SYNC WITH https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs !!!
```

nohwnd (approved):

PR #24438, opened 2022-03-18
Update minimum vs version to 17.2

marcpopMSFT:
Per a [new policy](https://docs.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules) we're discussing, we will want .net 7 to load in 17.3 and newer.

This will move us to 17.2 and then we'll move to 17.3 once it ships so early adopters have at least a short window to move.


KalleOlaviNiemitalo:
> let's not lock them

Is there an extra "not" or what does this mean?

marcpopMSFT:
@baronfel , I think it's time to do this and then we'll update to 17.3 probably in September.

baronfel (approved):
I agree, let's do it.

Issue #23646, opened 2022-01-28
Package validation does not report any compatibility errors if package name != assembly name


Alex-Sob:
Hi team,

We're migrating our existing backward compatibility NuGet package to package validation feature introduced in .NET 6 SDK (baseline version validation).
It works fine in one repo, but we found out it didn't work in another repo - build is not reporting any compatibility errors. I had to do a pretty tough investigation that involved digging into SDK code and creating a test app referencing SDK assemblies, and finally I found out that ```RunPackageValidation``` target expects the same package and assembly names:
https://github.com/dotnet/sdk/blob/209d1385a6a7650510509413d9a42819651861c8/src/Compatibility/Microsoft.DotNet.PackageValidation/NupkgParser.cs#L47

So, if ```PackageId``` and ```AssemblyName``` are different in the project file, package validation will not report any compatibility errors and build will succeed.

safern:
Working on these issues now to get a fix for 6.0.3.

safern:
Working on these issues now to get a fix for 6.0.3.

iedeny:
> Working on these issues now to get a fix for 6.0.3.

@safern what is a release timeline for 6.0.3?

claudiamurialdo:
6.0.3 was released two days ago but this error seems to continue happening. Do you have any idea when it is going to be fixed?

safern:
@claudiamurialdo sorry for not giving an update, it slipped through. Unfortunately I was not able to make the fix on time for 6.0.3 and I have switched teams inside MSFT (not part of .NET anymore), so I don't think I will be able to take a look at this any time soon. So perhaps @joperezr or @ericstj can take over this one. 

The fix that I was planning is simple, it is basically flowing `AssemblyName` `MSBuild` property to [`ValidatePackage`](https://github.com/dotnet/sdk/blob/8cfa52e8b15db97e18c5b6e08036ac1e0b45b742/src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs#L18) task [here](https://github.com/dotnet/sdk/blob/8cfa52e8b15db97e18c5b6e08036ac1e0b45b742/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets#L33-L45).

Then update this line:
[...]

Issue #17364, opened 2021-05-03
[Package Validation] Add support for multiple assemblies in the package.


Anipik:
Currently package validation support only one assembly in the package, there could be multiple copies but it would be the same dll.

But some packages like Microsoft.ML contatins more than one assembly.

cc @safern @ViktorHofer @ericstj

ericstj:
What happens with this today?  Might make sense to take a brief look and put an error message in place if we can't implement it yet.

PR #24439, opened 2022-03-18
Adding Package Validation support for cases where PackageId is different than AssemblyName

joperezr:
Fixes https://github.com/dotnet/sdk/issues/23646

As reported in the issue, we currently only run validation on the assemblies inside the package that have the same AssemblyName than the PackageId. This is not ideal since it is common for PackageId to be overwritten. With this change, we instead will fallback to use AssemblyName as the key to be used for detecting which assemblies should be validated. This is the release/6.0.3xx PR, I will also have one for main branch in a bit.

cc: @ericstj @Alex-Sob @iedeny @claudiamurialdo

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
+++ b/src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
@@ -23,7 +23,6 @@ namespace Microsoft.DotNet.Compatibility
         [Required]
         public string RoslynAssembliesPath { get; set; }
 
-        [Required]
         public string AssemblyName { get; set; }
 
         public string RuntimeGraph { get; set; }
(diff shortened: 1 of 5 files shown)


src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
```diff
@@ -23,6 +23,9 @@ public class ValidatePackage : TaskBase
         [Required]
         public string RoslynAssembliesPath { get; set; }
 
+        [Required]
```
ericstj, line 26: Seems rude to add this as required.  Especially since in the future we expect to support multiple Assemblies in a single package. https://github.com/dotnet/sdk/issues/17364

joperezr: I marked it as required since we control the targets that call into the task, and we know that AssemblyName property is an MSBuild property that should always be defined, so I didn't see any harm on making it required. Anyway, I'm fine with marking it optional and default back to use PackageId if it wasn't passed in.

ericstj: Please make it optional.  We may want to remove this completely in the future.

ericstj (approved):
LGTM

Issue #23687, opened 2022-01-31
[6.0.200-preview.22055.15] `dotnet tool run ...` will "hang" if the requested tool is not installed


brettfo:
REPRO:
======
0. Open a command prompt at some empty directory.
1. `dotnet new tool-manifest` // success
2. `dotnet tool run dotnet-interactive -- --version`

EXPECTED:
======
There should be a single instance of the message "Cannot find a tool in the manifest file that has a command named 'dotnet-interactive'."

ACTUAL:
======
The error message "Cannot find a tool in the manifest file that has a command named 'dotnet-interactive'." is repeated indefinitely.  From an automation standpoint, it appears as if the command has hung and nothing can continue.

This only happens with SDK version 6.0.200-preview.22055.15 and not 6.0.101.

`dotnet --info`:

```
.NET SDK (reflecting any global.json):
 Version:   6.0.200-preview.22055.15
 Commit:    a3c3a2bf3b

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.200-preview.22055.15\

Host (useful for support):
  Version: 6.0.1
  Commit:  3a25a7f1cc

.NET SDKs installed:
  3.1.416 [C:\Program Files\dotnet\sdk]
  5.0.404 [C:\Program Files\dotnet\sdk]
  6.0.101 [C:\Program Files\dotnet\sdk]
  6.0.200-preview.22055.15 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.18 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.19 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #23688, opened 2022-01-31
fix dotnet run infinite loop by rendering help explicitly

baronfel:
Fixes #23687 and #23034 by directly returning a HelpContext InvocationResult in the help middleware pipeline, instead of trying to force the system to kinda-sorta invoke the HelpContext itself on a manually-altered parameter string.

Issue #23034, opened 2021-12-13
`dotnet run` with invalid framework and `--` goes into infinite loop


zivkan:
### Describe the bug

`dotnet run` allows us to use `--` to separate arguments that might be ambiguous to the app being run, or the `dotnet` CLI itself. For example, `dotnet run` has argument `-f`, but if I want my app to have an `-f` argument, I have to pass it using `dotnet run -- -f`.

When a project uses `<TargetFrameworks>` with more than one framework specified, it's also mandatory to tell `dotnet run` which framework to run, for example `dotnet run -f net6.0 -- my app arguments here` or `dotnet run -f net48`.

Normally if you tell `dotnet run` to use a framework that isn't specified in the project file, it generates an error message and the process exits. However, if the command line includes `--`, then `dotnet run` enters an infinite loop showing two errors over and over. If you press ctrl-c, then it starts repeating only one of those errors, and now it does so more quickly. The only way to break out is to kill the `dotnet` process, no amount of ctrl-c will stop it.

### To Reproduce

```text
dotnet new console
dotnet run -f net48 -- hello
```

### Further technical details
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and its version

```text
.NET SDK (reflecting any global.json):
 Version:   6.0.200-preview.21573.3
 Commit:    a2bb0dde82

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19043
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.200-preview.21573.3\

Host (useful for support):
  Version: 6.0.0
  Commit:  4822e3c3aa

.NET SDKs installed:
  3.1.415 [C:\Program Files\dotnet\sdk]
  5.0.202 [C:\Program Files\dotnet\sdk]
  5.0.209 [C:\Program Files\dotnet\sdk]
  5.0.302 [C:\Program Files\dotnet\sdk]
  5.0.403 [C:\Program Files\dotnet\sdk]
  6.0.100 [C:\Program Files\dotnet\sdk]
  6.0.200-preview.21573.3 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.28 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #24449, opened 2022-03-18
Fix help infinite loop.

baronfel:
https://github.com/dotnet/sdk/pull/23688 got way out of hand, so here is a more restrained fix.  The core issue is that we can't reliably trigger help output by adding `-h` to the end of the tokens, instead now we find the set tokens associated with the current command (if any) and place the `-h` token there.

More sophisticated help detection and output could be done, but this may require coordination with the System.CommandLine team and should be done for .NET 7.

Fixes https://github.com/dotnet/sdk/issues/23687 and https://github.com/dotnet/sdk/issues/23034.


src/Cli/dotnet/Program.cs
```diff
@@ -64,11 +64,6 @@ public static int Main(string[] args)
                 {
                     return ProcessArgs(args, startupTime);
                 }
-                catch (HelpException e)
```
baronfel, line 67: this clause, and the type caught in it, I removed because the HelpException type isn't used anywhere.

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -13,9 +13,20 @@ namespace Microsoft.DotNet.Cli
...
+        ///<summary>
+        ///<remarks>
+        /// This is accomplished by finding a set of tokens that should be valid and appending a help token
+        /// to that list, then re-parsing the list of tokens. This is not ideal - either we should have a direct way
+        /// of invoking help for a ParseResult.
+        ///</remarks>
         public static void ShowHelp(this ParseResult parseResult)
         {
-            Parser.Instance.Parse(parseResult.Tokens.Select(t => t.Value).Append("-h").ToArray()).Invoke();
+            var tokenList = parseResult.Tokens.TakeWhile(t => t.Type == TokenType.Command).Select(t => t.Value).ToList(); // this can be empty when there's no leading command token
```
baronfel, line 27: this works as long as all command names are front-loaded, which I believe they are currently. A 'full' fix will be more intrusive IMO.

baronfel: I've updated this already to handle commands that take arguments as well, so the majority of edge cases should be covered.

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -13,9 +13,22 @@ namespace Microsoft.DotNet.Cli
...
     public static class ParseResultExtensions
     {
+        ///<summary>
+        /// Finds the command of the parse result and invokes help for that command.
+        /// If no command is specified, invokes help for the application.
+        ///<summary>
+        ///<remarks>
+        /// This is accomplished by finding a set of tokens that should be valid and appending a help token
+        /// to that list, then re-parsing the list of tokens. This is not ideal - either we should have a direct way
+        /// of invoking help for a ParseResult.
```
dsplaisted, line 23: Missing the second part of either / or.  What's the other option?

dsplaisted (approved):

src/Cli/dotnet/ParseResultExtensions.cs
baronfel: Pushed up a new commit with that, thanks! It summarizes future work we could do to get more System.CommandLine-ey with the way we invoke help.

joeloff (approved):

PR #24450, opened 2022-03-18
Change the missing workload message to recommend doing a restore.

### LINKED ISSUE #23445: NETSDK1147 Change this error message to recommend a dotnet workload restore
Currently the workload not installed message tells you to install the workload.  The better option is to tell customers to run `dotnet workload restore`. We should not remove the list of missing workloads that that information is useful for debugging user issues.  However, restore is a better command to run especially if there are multiple workloads they will require.

marcpopMSFT:
Fixes #23445

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -693,7 +693,7 @@ The following are names of parameters or literal values and should not be transl
   <data name="WorkloadNotInstalled" xml:space="preserve">
     <value>NETSDK1147: To build this project, the following workloads must be installed: {0}
 To install these workloads, run the following command: dotnet workload restore</value>
-    <comment>{StrBegin="NETSDK1147: "}</comment>
+    <comment>{StrBegin="NETSDK1147: "} LOCALIZATION: Do not localize "dotnet workload restore"</comment>
   </data>
   <data name="WindowsSDKVersionConflicts" xml:space="preserve">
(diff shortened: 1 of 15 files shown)


src/Tasks/Common/Resources/Strings.resx
```diff
@@ -692,7 +692,7 @@ The following are names of parameters or literal values and should not be transl
   </data>
   <data name="WorkloadNotInstalled" xml:space="preserve">
     <value>NETSDK1147: To build this project, the following workloads must be installed: {0}
-To install these workloads, run the following command: {1}</value>
+To install these workloads, run the following command: dotnet workload restore</value>
```
baronfel, line 695: Is there any context we need to give to the Loc team that `dotnet workload restore` is a non-localizable value? Or do they have that context already from prior experience?

baronfel (approved):

src/Tasks/Common/Resources/Strings.resx
marcpopMSFT: That's a good question. I'm not sure if there is a way to mark these that way and I've definitely seen items like this get localized in the past. @cristianosuzuki77 what can we do to mark this so the localizers don't translate the new text "dotnet workload restore" as those are CLI commands?

marcpopMSFT: I didn't find any examples in the sdk repo but I did find some in other repos like msbuild and templating where messages were included in the comments to the loc team. As such, I've added a comment here about that. If we don't hear back from Cristiano, we can proceed with this and see if we get pushback/feedback on this approach.

PR #24451, opened 2022-03-18
Hide manifest updates during workload install

### LINKED ISSUE #23446: When installing workloads, hide manifest updates from customers
We've had multiple customers get confused when installing workloads that we also update the manifests of unrelated workloads.  "Why did it install tvos?" and similar questions are ones we get.

Instead, perhaps we should hide that information behind the verbosity switch and instead report something like "Checking for updated workloads...".  Check with Chet on the exact design.

image

gkulin:
fixes #23446


src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -42,7 +43,9 @@ internal class WorkloadManifestUpdater : IWorkloadManifestUpdater
             string tempDirPath,
             IWorkloadInstallationRecordRepository workloadRecordRepo,
             PackageSourceLocation packageSourceLocation = null,
-            Func<string, string> getEnvironmentVariable = null)
+            Func<string, string> getEnvironmentVariable = null,
+            bool displayManifestUpdates = true
+            )
```
dsplaisted, line 48: Small nit: I would put the parenthesis on the same line as the last parameter.
```suggestion
            bool displayManifestUpdates = true)
```

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -420,6 +420,20 @@ public void GivenWorkloadInstallItWarnsWhenTheWorkloadIsAlreadyInstalled()
...
+        [InlineData("--verbosity:normal")]
+        public void HideManifestUpdatesWhenVerbosityIsMinimalOrNormal(string verbosityFlag)
+        {
+            var command = new DotnetCommand(Log);
+            command
+                .WithEnvironmentVariable("DOTNET_MSBUILD_SDK_RESOLVER_CLI_DIR", string.Empty)
+                .WithEnvironmentVariable("PATH", "fake")
+                .Execute("workload", "install", verbosityFlag, "android")
+                .Should()
+                .HaveStdOutContaining(Workloads.Workload.Install.LocalizableStrings.CheckForUpdatedWorkloadManifests);
```
dsplaisted, line 434: This should also verify that the output doesn't have the `AdManifestUpdated` message.

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -420,6 +420,20 @@ public void GivenWorkloadInstallItWarnsWhenTheWorkloadIsAlreadyInstalled()
             Directory.GetFiles(installRecordPath).Count().Should().Be(2);
         }
 
+        [Theory]
+        [InlineData("--verbosity:minimal")]
+        [InlineData("--verbosity:normal")]
+        public void HideManifestUpdatesWhenVerbosityIsMinimalOrNormal(string verbosityFlag)
```
dsplaisted, line 426: Ideally we should also have a test that checks that the AdManifestUpdated message is displayed when the verbosity is detailed or diagnostic.  It may be hard to mock things up so that there are updates available though.

dsplaisted (approved):

joeloff (approved):
Have we looked at the output for admin installs? This should hide the message, but it will still show progress information when the actual install happens.

PR #24481, opened 2022-03-22
SDK support for NativeAot publish

LakshanF:
Currently, publishing an app for native aot form factor, requires adding a reference package, `Microsoft.DotNet.ILCompiler`, to the user project.
This is an exploratory PR to add support to the SDK to publish for this form factor similar to `trimming `and `readytorun` publish support. 
 - The property, `PublishNativeAot`, is used to publish for this mode.
 - An SDK test, `GivenThatWeWantToRunAnAotApp`, is used to track this work but currently doesn't complete
 - Given the Platform specific assets in the package and due to size concerns, the method here follows the `ReadyToRun` format which does not include adding most of the binary package assets in the SDK
 - Most of the SDK related assets are added to `src\Tasks\Microsoft.NET.Build.Tasks\targets` including the entry point asset `Microsoft.DotNet.ILCompiler.targets`
 - The entry point asset `Microsoft.DotNet.ILCompiler.targets` is imported from `Microsoft.NET.Sdk.targets`
 - The package details are currently not defined correctly in a way for the SDK to find these on its own. These properties (`PackageDefinitions `and `PackageDependencies`) are currently hardcoded in `Microsoft.DotNet.ILCompiler.targets`
 - The SDK is not able to current find `ILCompiler.Build.Tasks.dll`
 - The analyzer, `ILLink.RoslynAnalyzer.dll`, is causing issues with some test projects and requires the warnings to be suppressed (these are likely analyzer issues)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.DotNet.ILCompiler.targets
```diff
@@ -0,0 +1,85 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
agocke, line 1: Isn't this file from the package? We can't move the targets out of the package as long as we still have consumers of the package (which is everyone right now). I think we have to start by integrating with the package into the SDK as-is for one release, then moving it over.

agocke:
I think you might be able to integrate NativeAOT similar to how we download and integrate the package for crossgen2, with two caveats:

1. NativeAOT needs to download two packages, since it has both the meta package with the shared targets, and the runtime pack
2. The targets files need to stay in the meta package for now, but we can integrate them into the SDK later

I started mocking this up at https://github.com/dotnet/sdk/compare/main...agocke:publish-aot?expand=1

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.DotNet.ILCompiler.targets
LakshanF: I was thinking of copying the targets to the SDK and adding a guard to the SDK side and leave the targets in the package for backward compatibility. But your suggestion below sounds a reasonable one to take and move the targets to the SDK later when existing folks move.

LakshanF:
Incorporated this part into #24550 and will close this one

Issue #23403, opened 2022-01-12
Workload feature band fallback doesn't work with workload update


dsplaisted:
The .NET SDK will fall back to workload manifests from previous feature bands.  This ensures that when we create the first builds of a new .NET SDK, workloads will continue to work without having been updated to the new feature band.

However, if there are newer versions of the workload manifests from a previous feature band, they will not be picked up with `dotnet workload update`.  This is because the update process looks for the latest version of the workload manifest package, but the package ID is constructed using the current SDKs feature band version.  So if it's falling back to a previous band, and there's a newer version of that workload manifest package, it won't find it.

To fix this we should probably first check for the latest package using the current SDK's feature band.  If we don't find any packages, and the current manifest on disk is coming from falling back to a previous band's manifest, then we should also try to get the latest version of the package for that previous feature band.

We probably don't need to worry about possible feature bands in between the two.  IE if we had a 6.0.300 SDK using a 6.0.100 manifest, then we should check for the latest 6.0.300 manifest, and if that's not found check for the latest version of the 6.0.100 manifest, but never check for a 6.0.200 manifest.

https://github.com/dotnet/sdk/blob/22c4860dcb2cf6b123dd641cc4a87a50380759d5/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs#L165

EDIT: We will also need to figure out how to differentiate between the feature band an advertising manifest is branded as and the feature band of the SDK it's advertised to, for example in `WorkloadManifestUpdater.GetAdvertisingManifestVersionAndWorkloads`.  Right now we put advertising manifests in separate folders per feature band, but this doesn't let us model a 6.0.200 manifest being advertised to a 6.0.300 SDK, for example.

dsplaisted:
FYI @jonathanpeppers @Redth

Redth:
> We probably don't need to worry about possible feature bands in between the two. IE if we had a 6.0.300 SDK using a 6.0.100 manifest, then we should check for the latest 6.0.300 manifest, and if that's not found check for the latest version of the 6.0.100 manifest, but never check for a 6.0.200 manifest.

I know we talked about this probably being ok, but I'm doubting that now.  I guess there's added cost in this check if it needs to query nuget?  But maybe it's still worth that cost to avoid unexpected behaviour here?  I think even with our current baseline manifests in the installer today might be at 6.0.100 and we don't have 6.0.300 yet, so really 6.0.200 _is_ what we want but would be missed in this case.

[...]

Issue #23402, opened 2022-01-12
Workload feature band fallback doesn't work with rollback files


dsplaisted:
When updating workload versions using a [rollback file](https://github.com/dotnet/sdk/blob/main/documentation/general/workload-rollback.md), it will not work correctly if any of the workload manifests are using a different feature band (via the fallback feature).

This is because the rollback file specifies the manifest ID and version, but to find the package that includes the manifest, the feature band is added to the manifest ID to create the manifest package ID.  When applying the rollback file, the SDK's current feature band will be used when trying to download the manifest package, even if the original manifest used a different feature band.

To fix this, we should update the format of the rollback file to include the feature band that the manifest is branded for, and use that when constructing the package ID to download.

- [ ] Update --print-rollback to include the feature band
- [ ] Update the rollback reader to include parse the manifest version and use during update of the manifest
- [ ] When applying rollback, uninstall existing manifest before installing new one as necessary (otherwise a newer band manifest could be left on disk and the rollback wouldn't actually take effect)
- [ ] If no band found in the rollback, default to either the current SDK band or the current manifest band (not sure which is preferred)
- [ ] Inform our partners who may have hardcoded rollback files that they could be impacted

https://github.com/dotnet/sdk/blob/22c4860dcb2cf6b123dd641cc4a87a50380759d5/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs#L206

Redth:
So would this allow us to do something like:

```
{
  "microsoft.net.sdk.android" : { 
    "version" : "31.0.200-preview.14.106",
    "sdkBand" : "6.0.200"
  },
  "microsoft.net.sdk.ios" : {
    "version" : "15.2.302-preview.14.91",
    "sdkBand" : "6.0.200"
  },
  "microsoft.net.sdk.maccatalyst" : {
    "version" : "15.2.302-preview.14.91",
    "sdkBand" : "6.0.200"
  },
  "microsoft.net.sdk.macos" : {
    "12.1.302-preview.14.91",
    "sdkBand" : "6.0.200"
  },
  "microsoft.net.sdk.maui" : {
    "6.0.200-preview.14.5055",
    "sdkBand" : "6.0.200"
  },
  "microsoft.net.sdk.tvos" : {
    "version" : "15.2.302-preview.14.91",
    "sdkBand" : "6.0.200"
  }
  "microsoft.net.workload.mono.toolchain" : {
     "version" : "6.0.3"
  },
[...]

Issue #21741, opened 2021-10-05
Support grouping multiple workload packs in a single MSI


dsplaisted:
Currently, we create an MSI for each workload pack.  However, workloads can be made up of many workload packs.  Maui for example has something like 40 workload packs, many of them because they depend on various Microsoft.Extensions NuGet packages which they would like to include as library packs.  Installing all of the workload packs as MSIs can have a high overhead.

We should support grouping multiple workload packs into a single MSI.  The MSI workload manifest could include an extra file that could map from workload packs to an MSI that includes them.  The MSI-based installer would then use that mapping to find the MSI to install.

The set of workload packs that should be grouped together in an MSI would be defined manually in the repo that created the workload manifest MSI.  The MSI creation tools would create the mapping file and the grouped MSIs based on that.

It's possible that the same workload pack would be installed by multiple different MSIs (for example if two different workloads in different repos depended on the same pack).  We should generally try to avoid this because it will increase the overall installer size, but the installation system should allow this.  If there were multiple MSIs that installed a given workload pack, the installer could install both of them if needed and they shouldn't conflict with each other (because the workload packs are all side-by-side).

For reference, this is the same as library packs that we've discussed before.

marcpopMSFT:
https://github.com/dotnet/sdk/issues/21227

marcpopMSFT:
This is also workload pack groups.

PR #24488, opened 2022-03-22
Refactor some workload APIs

dsplaisted:
Refactoring to support various things:

- Pass list of packs to pack installer, in order to support grouping workload packs (#21741)
- Save and provide path to workload manifest directory.  This will also support grouping workload packs, so that we can add additional information to the MSI-based workload manifest install about which packs are grouped together
- Add property to retrieve workload manifest feature band.  This will help support #23402 and #23403

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadManifestInfo.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadManifestInfo.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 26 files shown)


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadManifestInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+            Id = id;
+            Version = version;
+            ManifestDirectory = manifestDirectory;
+            ManifestFeatureBand = Path.GetFileName(Path.GetDirectoryName(manifestDirectory))!;
+        }
+
+        public string Id { get; }
+        public string Version { get; }
+        public string ManifestDirectory { get; }
+        public string ManifestFeatureBand { get; }
```
joeloff, line 27: Do we ever need to use this in the context of the SdkFeatureBand struct (e.g. version comparisons, etc.) or we only need the string representation?

joeloff (approved):

PR #24512, opened 2022-03-23
Detect arm64 hostfxr.dll

### LINKED ISSUE #24023: Report an error when using --skip-manifest-update and --from-rollback-file

benvillalobos:
Help msbuild understand when it should load an arm64 hostfxr. I verified this works locally on an arm64 machine.

The check arm64 folder check should only run when finding `hostfxr`, otherwise we crash later in the build.

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.NativeWrapper/Interop.cs
+++ b/src/Resolvers/Microsoft.DotNet.NativeWrapper/Interop.cs
@@ -40,7 +40,7 @@ namespace Microsoft.DotNet.NativeWrapper
         {
             string basePath = Path.GetDirectoryName(typeof(Interop).Assembly.Location);
             string architecture = IntPtr.Size == 8 ? "x64" : "x86";
-            architecture = dllFileName.Equals("hostfxr", StringComparison.OrdinalIgnoreCase) && RuntimeInformation.ProcessArchitecture == Architecture.Arm64 ? "arm64" : architecture;
+            architecture = RuntimeInformation.ProcessArchitecture == Architecture.Arm64 ? "arm64" : architecture;
             string dllPath = Path.Combine(basePath, architecture, $"{dllFileName}.dll");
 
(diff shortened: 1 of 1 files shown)


BenVillalobos:
cc @marcpopMSFT

dsplaisted (approved):
Why are we special casing this to only hostfxr?  Are there other DLLs we load that we don't have an Arm64 version of, and do the Amd64 versions of those work correctly on Arm64?

BenVillalobos:
> Why are we special casing this to only hostfxr? Are there other DLLs we load that we don't have an Arm64 version of, and do the Amd64 versions of those work correctly on Arm64?

I retried my local build with this check fully expecting this to fail so I could share the log, but I retried without the hostfxr-specificity and it built properly. Pushed up that change.

src/Resolvers/Microsoft.DotNet.NativeWrapper/Interop.cs
```diff
@@ -40,6 +40,7 @@ private static void PreloadWindowsLibrary(string dllFileName)
         {
             string basePath = Path.GetDirectoryName(typeof(Interop).Assembly.Location);
             string architecture = IntPtr.Size == 8 ? "x64" : "x86";
+            architecture = RuntimeInformation.ProcessArchitecture == Architecture.Arm64 ? "arm64" : architecture;
```
kasperk81, line 43: why not:
```suggestion
            string architecture = Enum.GetName(RuntimeInformation.ProcessArchitecture).ToLowerInvariant();
```

BenVillalobos: For the sake of time, I decided not to consider the other code paths it could potentially go down if it were on a different arch. Though I suspect this to be the better way to go about it

PR #24538, opened 2022-03-24
Revert #23358 

chsienki:
Revert "Better caching for RazorSourceGenerator when SuppressRazorSourceGenerator changes (#23358)"

This reverts commit a9b8bed2cace989c938fd783f0460105d017328a.


jaredpar (approved):

sharwell:
For people confused about the above, this revert occurred because the change in question caused a _severe_ performance regression inside the IDE (increased memory allocation load by up to 100GB/minute). The primary problem stemmed from the source generator now running certain paths that were previously bypassed.

PR #24545, opened 2022-03-24
Workload feature band fallback with rollback files

### LINKED ISSUE #23402: Workload feature band fallback doesn't work with rollback files

gkulin:
fixes #23402

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
@@ -17,12 +17,14 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
         IEnumerable<(
             ManifestId manifestId, 
-            ManifestVersion existingVersion, 
+            ManifestVersion existingVersion,
+            SdkFeatureBand existingFeatureBand,
             ManifestVersion newVersion,
-            Dictionary<WorkloadId, WorkloadDefinition> Workloads,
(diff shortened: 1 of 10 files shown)


dsplaisted (approved):
LGTM.

We also need to handle uninstalling the old versions of manifests (especially when we downgrade).  I've filed #24628 for that.

Also, I believe we are now in M2 approval mode, so don't merge this yet until we get signoff.

PR #24550, opened 2022-03-25
SDK integration via BundledNetCoreApp

LakshanF:
Adds support to publish NativeAot applications via the SDK with 3 parts:

1. Support to download the required packages

- Add `Microsoft.DotNet.ILCompiler` package details to [OverrideAndCreateBundledNETCoreAppPackageVersion.cs](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Layout/toolset-tasks/OverrideAndCreateBundledNETCoreAppPackageVersion.cs#L126-L134), specifically by adding a new element node and by passing `Stage0MicrosoftNETCoreAppRefPackageVersionPath` 
- Add code to download the package to [ProcessFrameworkReferences.cs](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs)
- Pass in publish property to [Microsoft.NET.Sdk.FrameworkReferenceResolution.targets](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets#L102) to invoke the `ProcessFrameworkReferences `task and get back package details, `ResolvedILCompilerPack`

2. Import Native AOT targets and properties.
- Copy these from the runtime repo [location](https://github.com/dotnet/runtime/tree/main/src/coreclr/nativeaot/BuildIntegration), 
- Modify the entry point target, `ImportRuntimeIlcPackageTarget`, in [Microsoft.DotNet.ILCompiler.targets](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.DotNet.ILCompiler.targets#L43) to guard it against the publish property
- Change the `ImportRuntimeIlcPackageTargetDependsOn `property in [Microsoft.DotNet.ILCompiler.targets](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.DotNet.ILCompiler.targets#L36) `_ComputeManagedRuntimePackAssemblies `from the existing value, `RunResolvePackageDependencies`, to set the required `PackageDefinitions` items
- Change `IlcBuildTasksPath `in Microsoft.NETCore.Native.targets to the package directory

3. Add a test, [GivenThatWeWantToRunAnAotApp.cs](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunAnAotApp.cs), as a smoke test

Current Issues
- Had to hard code the [ILCompilerPackVersion ](https://github.com/dotnet/sdk/blob/6c77bad07ed3c18900a1c51cddb7d223d9d21ad8/src/Layout/toolset-tasks/OverrideAndCreateBundledNETCoreAppPackageVersion.cs#L131)since I couldn't get a "7.0.0-*" wild card to be accepted. Tried some combinations as defined in this [doc](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning) without much success.
- Can't find where IlcPath is set in the existing configuration, which is required to find the ILCompiler.Build.Tasks.dll. This is due to import being higher in evaluation order.
- Some test projects, `Microsoft.DotNet.Cli.Utils.csproj, Microsoft.NET.TestFramework.csproj, Microsoft.NET.Publish.Tests.csproj,` failing with the trimming code analyzers


agocke:
Ended up looking into this more and I think we'll have to copy all the targets/tasks into the SDK -- I forgot that the MSBuild import logic is evaluated before any task or itemgroup resolution, so it's impossible to import the targets in a straightforward way from an itemgroup analysis

LakshanF:
> Ended up looking into this more and I think we'll have to copy all the targets/tasks into the SDK -- I forgot that the MSBuild import logic is evaluated before any task or itemgroup resolution, so it's impossible to import the targets in a straightforward way from an itemgroup analysis

Incorporated your feedback but `ILCompilerDir `does not get resolved correctly yet.

MichalStrehovsky:
> Ended up looking into this more and I think we'll have to copy all the targets/tasks into the SDK -- I forgot that the MSBuild import logic is evaluated before any task or itemgroup resolution, so it's impossible to import the targets in a straightforward way from an itemgroup analysis

Can we make it so that the files aren't physically commited in this repo? We need them in the runtime repo for a couple reasons:

* They're used for building crossgen2 and to build all the tests.
* We need to ship them as part of the ILCompiler package to support our existing customers that use NativeAOT with .NET 6 (this is a point in time issue).
* Compiler changes often need matching .targets changes

We could have the files commited to both repos but it makes the logistics of updating them complex (compiler changes sometimes need targets changes and vice versa).

Could we pull the ILCompiler Nuget package as part of the SDK repo build and simply package up files from the package we just pulled down? We could make a new transport package out of runtime repo that has just the .targets files if it helps anything.

agocke:
@MichalStrehovsky This makes sense, although I was surprised to figure out that R2R just checked everything into the SDK.

We should be able to copy the files from the package into the SDK layout -- the only thing we can't do is import from the NuGet package dynamically.

LakshanF:
Created a new PR, #24693, that combines downloading both the packages as discussed above (SDK assets via bundling, and build assets dynamically)

MichalStrehovsky:
> @MichalStrehovsky This makes sense, although I was surprised to figure out that R2R just checked everything into the SDK.

R2R duplicates the crossgen2 command line building logic in the runtime repo in multiple places. crossgen2 has a simpler command line interface so it's more tolerable, although it wouldn't hurt to centralize it. NativeAOT has a lot more complex command line interface (interactions with feature switches, trimming settings, runtime configuration, etc, along with having to construct the command line for the platform linker that multiplexes the complexity because it has Windows/Linux/macOS specifics).

PR #24563, opened 2022-03-25
Update to the 6.0.3 runtime for preview 3

marcpopMSFT:


mmitche (approved):

marcpopMSFT:
@dotnet/domestic-cat any idea why this test keeps failing? Seems unrelated to the runtime version:
Given_multi_target_It_should_get_suggested_workload_by_GetRequiredWorkloads_target
```
Expected collection {"macos", "android"} to contain {"macos", "android-aot"}, but could not find {"android-aot"}.
```

joeloff:
Nope, but I'll spend some time digging into it

marcpopMSFT:
@joeloff I don't see any reason why that would be caused by this PR. Other 3xx PRs are passing. I wonder if we picked up a bad change? I'll try forcing a remerge.

PR #24630, opened 2022-03-30
Support installing workload pack groups

dsplaisted:
- Supports installing workload pack group MSIs (#21741).  Creating the MSIs will be a separate update to the workload creation infrastructure in dotnet/arcade.
- Creates a `ManifestVersionUpdate` class to replace passing around tuples of 5 elements in a lot of places
- Switches to a new transaction infrastructure for workloads, to allow for more control of the ordering of messages and rollback actions, as well as making it explicit which installation methods should participate in the transaction (by taking an `ITransactionContext` parameter)

(This is rebased on top of #24545, so it also currently includes commits from that PR.)

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.PackGroup.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.PackGroup.cs
@@ -1,0 +1,117 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using static Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadResolver;
+
+namespace Microsoft.DotNet.Workloads.Workload.Install
+{
+    internal partial class NetSdkMsiInstallerClient
+    {
+        class WorkloadPackGroupJson
+        {
+            public string GroupPackageId { get; set; }
+            public string GroupPackageVersion { get; set; }
+
+            public List<WorkloadPackJson> Packs { get; set; } = new List<WorkloadPackJson>();
+        }
+
+        class WorkloadPackJson
+        {
+            public string PackId { get; set; }
+
+            public string PackVersion { get; set; }
+        }
+
+        Dictionary<(string packId, string packVersion), List<WorkloadPackGroupJson>> GetWorkloadPackGroups()
+        {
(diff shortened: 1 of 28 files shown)


gkulin (approved):

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.PackGroup.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        Dictionary<(string packId, string packVersion), List<WorkloadPackGroupJson>> GetWorkloadPackGroups()
+        {
+            Dictionary<(string packId, string packVersion), List<WorkloadPackGroupJson>> ret = new();
+
+            var manifests = _workloadResolver.GetInstalledManifests();
+            foreach (var manifest in manifests)
+            {
+                var packGroupFile = Path.Combine(manifest.ManifestDirectory, "WorkloadPackGroups.json");
```
joeloff, line 40: This is the new mapping file we're going to add to the manifest installers?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/ManifestVersionUpdate.cs
```diff
@@ -0,0 +1,94 @@
...
+        //  Returns an object representing an undo of this manifest update
+        public ManifestVersionUpdate Reverse()
+        {
+            return new ManifestVersionUpdate(ManifestId, NewVersion, NewFeatureBand, ExistingVersion, ExistingFeatureBand);
+        }
+
+        public int CompareTo(ManifestVersionUpdate? other)
+        {
+            if (other == null) return 1;
+            int ret = ManifestId.CompareTo(other.ManifestId);
```
joeloff, line 39: For a version update of a manifest, shouldn't the IDs always be the same and the underlying versions be different. Should we instead throw something like `InvalidOperationException` if the IDs are different?

joeloff (commented):
I think I'm good with this. The tricky bit is going to be computing the group IDs/names, but that needs to happen in Arcade

joeloff (approved):

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/ManifestVersionUpdate.cs
dsplaisted: This is just implemented so that equality works correctly, so that in tests you can assert that the manifest updates were what was expected.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.PackGroup.cs
dsplaisted: Correct.

joeloff (approved):

PR #24678, opened 2022-04-01
Feature band fallback with workload update

### LINKED ISSUE #23403: Workload feature band fallback doesn't work with workload update

gkulin:
fixes #23403


src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -317,6 +316,9 @@ public void ItCanFallbackWithNoUpdates()
             nugetDownloader.DownloadCallParams[0].version.ShouldBeEquivalentTo(null);
             nugetDownloader.DownloadCallParams[1].id.ToString().Should().Be($"{testManifestName}.manifest-6.0.200");
             nugetDownloader.DownloadCallParams[1].version.ShouldBeEquivalentTo(null);
+
+            // check that output lines from _reporter do not say failed
+            _reporter.Lines.Should().NotContain("fail");
```
dsplaisted, line 321: This should probably be a case insensitive comparison.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -317,7 +317,7 @@ private async Task UpdateAdvertisingManifestAsync(WorkloadManifestInfo manifest,
                         // then try downloading the manifest's feature band
                         if (!(manifest.ManifestFeatureBand).Equals(_sdkFeatureBand))
                         {
-                            packagePath = await GetOnlinePackagePath(new SdkFeatureBand(manifest.ManifestFeatureBand), manifestId, includePreviews);
+                            packagePath = await GetOnlinePackagePath(new SdkFeatureBand(manifest.ManifestFeatureBand), manifestId, includePreviews);        // goes to failure message from here
```
dsplaisted, line 320: Instead of letting the exception fall through here, I think it would be better to catch it immediately (as is done on line 313).

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -291,40 +293,33 @@ private IEnumerable<ManifestId> GetInstalledManifestIds()
...
-                    catch (NuGetPackageNotFoundException)
+                    if (!(manifest.ManifestFeatureBand).Equals(_sdkFeatureBand))
                     {
-                        _reporter.WriteLine(string.Format(LocalizableStrings.AdManifestPackageDoesNotExist, manifestId));
+                        (success, packagePath) = await GetManifestPackageUpdate(new SdkFeatureBand(manifest.ManifestFeatureBand), manifestId, includePreviews, offlineCache);
+                        currentFeatureBand = manifest.ManifestFeatureBand.ToString();
                     }
                 }
-                else
+                if (!success)
```
joeloff, line 317: Is the second `if` needed? It appears to be inside the same block as the `if (!success)` on line 309

src/Tests/dotnet-workload-update.Tests/GivenDotnetWorkloadUpdate.cs
```diff
@@ -366,6 +366,8 @@ public void ApplyRollbackAcrossFeatureBand(string existingSdkFeatureBand, string
             packInstaller.InstalledManifests[0].manifestUpdate.ManifestId.Should().Be(manifestsToUpdate[0].manifestUpdate.ManifestId);
             packInstaller.InstalledManifests[0].manifestUpdate.NewVersion.Should().Be(manifestsToUpdate[0].manifestUpdate.NewVersion);
             packInstaller.InstalledManifests[0].manifestUpdate.NewFeatureBand.Should().Be(manifestsToUpdate[0].manifestUpdate.NewFeatureBand);
+            packInstaller.InstalledManifests[0].manifestUpdate.ExistingVersion.Should().Be(manifestsToUpdate[0].manifestUpdate.ExistingVersion);
```
joeloff, line 369: Question: Before the change, were these values not set, or did they have different values?

gkulin: These values were introduced in this PR from Daniel: https://github.com/dotnet/sdk/commit/6ba9c46bbdf6989d83a558535487f8ef087dc281
If I remember correctly, Daniel asked me to add these checks because his change was made after the PR where I originally wrote this test

joeloff (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -336,6 +331,8 @@ private async Task UpdateAdvertisingManifestAsync(ManifestId manifestId, bool in
                 Directory.CreateDirectory(Path.GetDirectoryName(adManifestPath));
                 FileAccessRetrier.RetryOnMoveAccessFailure(() => DirectoryPath.MoveDirectory(Path.Combine(extractionPath, "data"), adManifestPath));
 
+                File.WriteAllText(Path.Combine(adManifestPath, "AdvertisedManifestFeatureBand.txt"), currentFeatureBand);
```
dsplaisted, line 334: It would be good to add a comment here about what this file is for.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -357,15 +354,18 @@ private async Task UpdateAdvertisingManifestAsync(ManifestId manifestId, bool in
                 {
                     File.Delete(packagePath);
                 }
-
-                var versionDir = Path.GetDirectoryName(packagePath);
-                if (Directory.Exists(versionDir) && !Directory.GetFileSystemEntries(versionDir).Any())
+                if (!string.IsNullOrEmpty(packagePath))
```
dsplaisted, line 357: I think we only want to do this Directory cleanup if we were *not* using the offline cache.

src/Tests/dotnet-workload-install.Tests/MockNuGetPackageInstaller.cs
```diff
@@ -69,5 +92,8 @@ public Task<NuGetVersion> GetLatestPackageVerion(PackageId packageId, PackageSou
         {
             return Task.FromResult("mock-url-" + packageId.ToString());
         }
+
+        // when it's called from the new test 
+            // if the feature band is a part of a certain list that should have exceptions thrown then it should do that and then fallback
```
dsplaisted, line 97: Looks like these were notes that can probably be deleted.

dsplaisted (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
gkulin: @dsplaisted so should this not check anything regarding packagePath and instead only check if we are using the offline cache? Or should the offline cache check be added?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -357,15 +356,18 @@ private async Task UpdateAdvertisingManifestAsync(ManifestId manifestId, bool in
                 {
                     File.Delete(packagePath);
                 }
-
-                var versionDir = Path.GetDirectoryName(packagePath);
-                if (Directory.Exists(versionDir) && !Directory.GetFileSystemEntries(versionDir).Any())
+                if (offlineCache == null || !offlineCache.HasValue)
```
dsplaisted, line 359: This would be better to guard against cases where `packagePath` is null:

```suggestion
                if (!string.IsNullOrEmpty(packagePath) && (offlineCache == null || !offlineCache.HasValue))
```

dsplaisted (approved):

PR #24770, opened 2022-04-08
Don't retarget testhost

### LINKED ISSUE #24769: Testhost runtimeconfig is rewritten
### Describe the bug
Testhost uses a collection of testhost-*runtimeconfig.json files that are shipped with SDK, and that help projects that are not using Microsoft.NET.Test.SDK, or that don't generate their own runtimeconfig to run. They are used as fallback and should help the application target similar runtime as what it defines in its attributes. 

An example file (testhost-3.1.runtimeconfig.json) looks like this: 

```json
{
  "runtimeOptions": {
    "tfm": "netcoreapp3.1",
    "framework": {
      "version": "3.1.0-preview.0"
    }
  }
}
```

This file is rewritten by sdk build to always contain the current SDK version, which is not what the test app should target. 

It happens here: https://github.com/dotnet/sdk/blob/main/src/Layout/redist/targets/GenerateLayout.targets#L403-L428

### To Reproduce
Look into `"C:\Program Files\dotnet\sdk\6.0.201\testhost-3.1.runtimeconfig.json"`

```json
{
  "runtimeOptions": {
    "tfm": "netcoreapp3.1",
    "framework": {
      "version": "6.0.3"
    }
  }
}
```

### Exceptions (if any)
None.

nohwnd:
Testhost is shipped together with multiple runtimeconfigs to be used as fallback when user does not provide runtimeconfig of their own, but tfm can be detected from the assembly. Rewriting them makes the tested assembly use the current runtime version, which is wrong.

Fix #24769

### DIFF
--- a/src/Layout/redist/targets/GenerateLayout.targets
+++ b/src/Layout/redist/targets/GenerateLayout.targets
@@ -409,8 +409,9 @@
     </PropertyGroup>
     <ItemGroup>
       <!-- Exclude F# from retargeting: https://github.com/dotnet/toolset/issues/2866 -->
+      <!-- Exclude testhost from retargeting: https://github.com/dotnet/sdk/issues/24769 -->
       <ToolRuntimeConfigPath Include="$(OutputPath)/**/*.runtimeconfig.json"
-                             Exclude="$(OutputPath)/**/fsc.runtimeconfig.json;$(OutputPath)/**/fsi.runtimeconfig.json"/>
+                             Exclude="$(OutputPath)/**/fsc.runtimeconfig.json;$(OutputPath)/**/fsi.runtimeconfig.json;$(OutputPath)/**/testhost-*.runtimeconfig.json"/>
 
       <MSBuild15Items Include="$(OutputPath)/15.0/**/*" />
     </ItemGroup>


MarcoRossignoli (approved):

src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -409,8 +409,9 @@
     </PropertyGroup>
     <ItemGroup>
       <!-- Exclude F# from retargeting: https://github.com/dotnet/toolset/issues/2866 -->
```
kasperk81, line 411: this issue was fixed in runtime in year 2019

nohwnd: @kasperk81 thx for letting me know. I don't want to mix removing the current exclusions into this PR.

@vzarytovskii Fsharp tooling is excluded from retargetting (the task does text replace to change the target framework to match the version that the SDK is shipping with). According to the comment above, there is a fix in place since 2019. Please have a look into removing this exclusion.

nohwnd:
@dsplaisted Please review. These testhost runtimconfigjsons should be kept untouched.

dsplaisted (approved):
How long has this been broken?  Should this be fixed in 6.0.4xx or in servicing for previous versions?

nohwnd:
@dsplaisted Probably since the beginning, this is a fallback of a fallback, and it is rarely used. I will make another PR for 6.0.4xx, it should be okay to fix it only there.

PR #24826, opened 2022-04-12
bump WarningLevel to 7 for .NET 7

### LINKED ISSUE #21599: WarningLevel is no longer respected in .NET 5.0
It appears that in .NET 5.0, the `<WarningLevel>` project property is no longer respected, and the project is always compiled at warning level 5 even if it is changed to something else. Changing back to .NET Core 3.1 fixes the issue.

Steps to reproduce:

1. Create a new .NET 5.0 project. (My Visual Studio doesn't seem to yet have an option to create a .NET 5 project, so I created a .NET Core 3.1 project and changed to to .NET 5.0.)

2. Replace the C# source code with the following:

```
using System;

namespace CSTest
{
	class Program
	{
		private static int Foo;

		static void Main()
		{
			Console.WriteLine(Foo);
		}
	}
}
```

3. Compile it. As expected, you will receive the warning `warning CS0649: Field 'Program.Foo' is never assigned to, and will always have its default value 0`

4. Try changing the warning level of the project to 3 (CS0649 is level 4, so 3 should hide the warning). My project file now looks like this:

```

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
    <WarningLevel>3</WarningLevel>
  </PropertyGroup>

</Project>
```

5. Compile again. Note that the warning is still shown. (This is the bug.)

6. Change to .NET Core 3.1 and compile again. Note that the warning is no longer displayed, as was the original desire.

This behavior occurs regardless of whether I'm building using Visual Studio or running `dotnet build` from the command line. (Interestingly, when setting the warning level using Visual Studio via the Build tab in the project properties, *sometimes* the dropdown will immediately revert back to level 5, though Visual Studio does at least seem to consistently update the .csproj file itself.)

baronfel:
The C# compiler supports a warning wave of 7 now, and we should float that on like was done for .NET 6. This PR does that, in a slightly more maintainable way IMO.

The rules are:
* AnalysisLevel should track the current TFM if not specifed
* AnalysisLevel should be set to `latest` if the current TFM is the 'latest' TFM (as defined by a property that we need to bump)
* WarningLevel should track the current TFM if not specified
* WarningLevel shouldn't override the user-provided value
* WarningLevel should be set to 4 if the project is a .NET Framework project

Fixes https://github.com/dotnet/sdk/issues/21599 as well

cc @Youssef1313 via https://github.com/dotnet/templating/issues/4580

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
@@ -18,7 +18,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- If not specified and greater than or equal to .NET 5, default the AnalysisLevel to the the MAJ.MIN version -->
     <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
                               '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">$(_TargetFrameworkVersionWithoutV)</AnalysisLevel>
+                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">latest</AnalysisLevel>
 
 
     <!-- AnalysisLevel can also contain compound values with a prefix and suffix separated by a '-' character.
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
@@ -20,12 +20,16 @@ Copyright (c) .NET Foundation. All rights reserved.
                               '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
                               $([MSBuild]::VersionEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">5.0</AnalysisLevel>
 
-    <!-- Automatically set AnalysisLevel if not specified -->
+    <!-- Automatically set AnalysisLevel if not specified based on TFM major version -->
     <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
                               (('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
                                $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))) Or
-                               '$(AnalysisMode)' != '')">latest</AnalysisLevel>
+                               '$(AnalysisMode)' != '')">6.0</AnalysisLevel>
 
+    <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
+                               (('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
+                                $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '7.0'))) Or
+                                '$(AnalysisMode)' != '')">7.0</AnalysisLevel>
 
     <!-- AnalysisLevel can also contain compound values with a prefix and suffix separated by a '-' character.
          The prefix indicates the core AnalysisLevel and the suffix indicates the bucket of
@@ -65,26 +69,23 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Enable the trim analyzer when any trimming settings are enabled. Warnings may suppressed based on other settings. -->
     <EnableTrimAnalyzer Condition="'$(EnableTrimAnalyzer)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</EnableTrimAnalyzer>
 
-    <!-- Enable the AOT analyzer when AOT is enabled. Warnings may suppressed based on other settings. -->
-    <EnableAotAnalyzer Condition="'$(EnableAotAnalyzer)' == '' And '$(PublishAot)' == 'true'">true</EnableAotAnalyzer>
-
-    <!-- Compiler warning level, defaulted to 4. We promote it to 5 if the user has set analysis level to 5 or higher
+    <!-- The compiler warning level is defaulted to 4. We promote it to 5 if the user has set analysis level to 5 or higher
          NOTE: at this time only the C# compiler supports warning waves -->
-    <WarningLevel Condition="'$(Language)' == 'C#'">5</WarningLevel>
+    <WarningLevel Condition="'$(Language)' == 'C#' And '$(WarningLevel)' == ''">5</WarningLevel>
   </PropertyGroup>
 
   <!-- Set the default WarningLevel based on EffectiveAnalysisLevel
        For .NET 6 we want the Warning level to be 6 -->
   <PropertyGroup Condition="'$(EffectiveAnalysisLevel)' != '' And
                              $([MSBuild]::VersionGreaterThanOrEquals($(EffectiveAnalysisLevel), '6.0'))">
-    <WarningLevel Condition="'$(Language)' == 'C#'">6</WarningLevel>
+    <WarningLevel Condition="'$(Language)' == 'C#' And '$(WarningLevel)' == ''">6</WarningLevel>
   </PropertyGroup>
 
   <!-- Set the default WarningLevel based on EffectiveAnalysisLevel
        For .NET 7 we want the Warning level to be 7 -->
   <PropertyGroup Condition="'$(EffectiveAnalysisLevel)' != '' And
                              $([MSBuild]::VersionGreaterThanOrEquals($(EffectiveAnalysisLevel), '7.0'))">
-    <WarningLevel Condition="'$(Language)' == 'C#'">7</WarningLevel>
+    <WarningLevel Condition="'$(Language)' == 'C#' And '$(WarningLevel)' == ''">7</WarningLevel>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
@@ -15,21 +15,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Analysis level is a single property that can be used to control both the compiler warning waves
          and enable .NET analyzers. Valid values are 'none', 'latest', 'preview', or a version number  -->
 
-    <!-- Automatically set AnalysisLevel to 5 if targeting .net 5 and nothing else specified -->
+    <!-- If not specified and greater than or equal to .NET 5, default the AnalysisLevel to the the MAJ.MIN version -->
     <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
                               '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                              $([MSBuild]::VersionEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">5.0</AnalysisLevel>
+                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">$(_TargetFrameworkVersionWithoutV)</AnalysisLevel>
 
-    <!-- Automatically set AnalysisLevel if not specified based on TFM major version -->
-    <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
-                              (('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                               $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))) Or
-                               '$(AnalysisMode)' != '')">6.0</AnalysisLevel>
-
-    <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
-                               (('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                                $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '7.0'))) Or
-                                '$(AnalysisMode)' != '')">7.0</AnalysisLevel>
 
     <!-- AnalysisLevel can also contain compound values with a prefix and suffix separated by a '-' character.
          The prefix indicates the core AnalysisLevel and the suffix indicates the bucket of
@@ -45,8 +35,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- EffectiveAnalysisLevel is used to differentiate from user specified strings (such as 'none')
          and an implied numerical option (such as '4')-->
     <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'none' or '$(AnalysisLevelPrefix)' == 'none'">4.0</EffectiveAnalysisLevel>
-    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'latest' or '$(AnalysisLevelPrefix)' == 'latest'">6.0</EffectiveAnalysisLevel>
-    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'preview' or '$(AnalysisLevelPrefix)' == 'preview'">7.0</EffectiveAnalysisLevel>
+    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'latest' or '$(AnalysisLevelPrefix)' == 'latest'">7.0</EffectiveAnalysisLevel>
+    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'preview' or '$(AnalysisLevelPrefix)' == 'preview'">8.0</EffectiveAnalysisLevel>
 
     <!-- Set EffectiveAnalysisLevel to the value of AnalysisLevel if it is a version number -->
     <EffectiveAnalysisLevel Condition="'$(EffectiveAnalysisLevel)' == '' And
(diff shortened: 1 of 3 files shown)


Youssef1313 (commented):
While the change on SDK side is needed, it doesn't fully resolve the suggestion for templates.

A user might not specify `AnalysisLevel` or might specify low value for AnalysisLevel to get a smaller set of CAxxxx analyzers to run (due to FPs, rules that user think are noisy, too much new warnings after update, etc).

For compiler warnings, they're almost always correct warnings that users are encouraged to fix.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -77,6 +77,13 @@ Copyright (c) .NET Foundation. All rights reserved.
     <WarningLevel Condition="'$(Language)' == 'C#'">6</WarningLevel>
   </PropertyGroup>
 
+  <!-- Set the default WarningLevel based on EffectiveAnalysisLevel
+       For .NET 7 we want the Warning level to be 7 -->
+  <PropertyGroup Condition="'$(EffectiveAnalysisLevel)' != '' And
+                             $([MSBuild]::VersionGreaterThanOrEquals($(EffectiveAnalysisLevel), '7.0'))">
+    <WarningLevel Condition="'$(Language)' == 'C#'">7</WarningLevel>
```
Youssef1313, line 84: I'm curious if this can override a value the user has explicitly set for `<WarningLevel>`?

Should the condition be `'$(Language)' == 'C#' And '$(WarningLevel)' == ''`?

baronfel: It seems that `WarningLevel` isn't intended to be the user-facing field, instead `AnalysisLevel` is that field. We do similar defaulting for that property [here](https://github.com/dotnet/sdk/blob/bc24f09e31688ae3013a3150d8cbdb2c120bb236/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L15-L51), but it too is missing defaults/values for 7.0 (which I can add here).

The intent of the two different-but-related properties isn't entirely clear to me, I'll check our docs on docs.ms.com and see what we recommend to try and glean more context.

Youssef1313: @baronfel I think it *is* user-facing. It's documented [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#warninglevel)

I think most properties that are not intended to be user-facing are prefixed with an underscore.

baronfel: Of course you're correct on both counts - I had only found the AnalysisLevel documentation (which exists at the SDK docs, rather than the compiler docs)

Youssef1313: > new WarningLevel for .Net 7

I'm not sure what you mean by this.

---

> separately investigating the way we've tied WarningLevel and AnalysisLevel together

It's okay to have WarningLevel tied with AnalysisLevel. Only have a question: if the user sets both AnalysisLevel and WarningLevel, which value of WarningLevel will take effect (ie, passed to the compiler)? The one the user set in his csproj? or the one from the SDK [here](https://github.com/dotnet/sdk/blob/bc24f09e31688ae3013a3150d8cbdb2c120bb236/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L82-L85)? If the one from the SDK can take effect over the user's defined one, then we need to adjust the condition to add `And '$(WarningLevel)' != ''`

---

> because you assert that we basically always want the latest WarningLevel regardless of user-specified AnalysisLevel

This will break backward compatibility badly, hence, beats the purpose of warning waves feature.

What I meant is adding `<WarningLevel>7</WarningLevel>` in .NET 7 templates.

---

Adding @jmarolf @mavasani for thoughts too.

Youssef1313: To summarize, there are two points here:


1. I confirmed that AnalysisLevel modifies WarningLevel in a very bad way. Given the following csproj:

    ```xml

      <PropertyGroup>
        <OutputType>Exe</OutputType>
        <Nullable>enable</Nullable>
        <WarningLevel>7</WarningLevel>
        <AnalysisLevel>5</AnalysisLevel>
      </PropertyGroup>

    </Project>
    ```

    I see the following in binlog:

    image

    I'm **not** opposed to the SDK setting WarningLevel, but it should do that only if the user doesn't explicitly provide a value for it.

    I'll consider this as a bug.

---

2. This one is a feature request, to let the default template have all the new compiler warnings that are introduced in a warning wave. It's simply adding `<WarningLevel>` in csproj in the templates.

baronfel: I completely agree with (1) - we should only ever conditionally-set WarningLevel in the SDK.
I think we can implement (2) purely by the mechanism already here in the SDK - if no user-set WarningLevel is present, default to the level for the TFM chosen (so net6.0 -> 6, net7.0 -> 7, and so on).

I've pushed up a version of this with tests to illustrate the behavior now, can you see if you agree?

My main concern is crowding the project templates, we pretty aggressively try to limit the content generated in them. I think we can get the auto-incrementing behavior here and get to the more-safe-by-default you're asking for.

Youssef1313: > if no user-set WarningLevel is present, default to the level for the TFM chosen (so net6.0 -> 6, net7.0 -> 7, and so on).

I don't disagree here. Seems fine, but needs attention/approval from Roslyn team I think.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -15,16 +15,10 @@ Copyright (c) .NET Foundation. All rights reserved.
...
     <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
                               '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                              $([MSBuild]::VersionEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">5.0</AnalysisLevel>
-
-    <!-- Automatically set AnalysisLevel if not specified -->
-    <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
-                              (('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                               $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))) Or
-                               '$(AnalysisMode)' != '')">latest</AnalysisLevel>
+                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">$(_TargetFrameworkVersionWithoutV)</AnalysisLevel>
```
Youssef1313, line 21:
```suggestion
                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">latest</AnalysisLevel>
```

This will match the documentation in https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview#latest-updates, and also fix https://github.com/dotnet/docs/issues/29054 :)

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -41,8 +35,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- EffectiveAnalysisLevel is used to differentiate from user specified strings (such as 'none')
          and an implied numerical option (such as '4')-->
     <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'none' or '$(AnalysisLevelPrefix)' == 'none'">4.0</EffectiveAnalysisLevel>
-    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'latest' or '$(AnalysisLevelPrefix)' == 'latest'">6.0</EffectiveAnalysisLevel>
-    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'preview' or '$(AnalysisLevelPrefix)' == 'preview'">7.0</EffectiveAnalysisLevel>
+    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'latest' or '$(AnalysisLevelPrefix)' == 'latest'">7.0</EffectiveAnalysisLevel>
```
baronfel, line 38: I don't like these explicit numbers, but now that they are the only place where we do number-y things I'm less bothered by it. They become once-a-release updates now I suppose.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -41,8 +35,8 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <EffectiveAnalysisLevel Condition="'$(AnalysisLevel)' == 'preview' or '$(AnalysisLevelPrefix)' == 'preview'">8.0</EffectiveAnalysisLevel>
```
Youssef1313, line 39: 7.0 is still preview, and there is no 8.0 yet. Anything I'm missing?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -65,25 +59,15 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-  </PropertyGroup>
-
-  <!-- Set the default WarningLevel based on EffectiveAnalysisLevel
-       For .NET 6 we want the Warning level to be 6 -->
-  <PropertyGroup Condition="'$(EffectiveAnalysisLevel)' != '' And
-                             $([MSBuild]::VersionGreaterThanOrEquals($(EffectiveAnalysisLevel), '6.0'))">
-    <WarningLevel Condition="'$(Language)' == 'C#'">6</WarningLevel>
-  </PropertyGroup>
+    <!-- If the user specified 'preview' we want to pick a very high warning level to opt into the highest possible warning wave -->
+    <WarningLevel Condition="'$(Language)' == 'C#' And '$(AnalysisLevel)' == 'preview'">9999</WarningLevel>
```
Youssef1313, line 63: Do we want to check WarningLevel here too as part of the condition?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props
```diff
@@ -12,7 +12,6 @@ Copyright (c) .NET Foundation. All rights reserved.
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <WarningLevel Condition=" '$(WarningLevel)' == '' ">4</WarningLevel>
```
baronfel, line 15: This is potentially controversial, but I think necessary because of the way props/projects/targets properties are ordered during import. Due to this ordering, `WarningLevel` is virtually guaranteed to be `''` at this point, so we'll set it here, which means if a user explicitly sets a WarningLevel in their project file we won't have a way of distinguishing that. This layering is part of the reason why we've moved more towards settings properties in `targets` files over the years.

I've checked and nothing seems to modify WarningLevel in this way aside from the Analyzers targets, so this seems pretty safe. Compilers and the ILLink targets consume it, but that's well after the Analyzers targets run.

Nonetheless I've reached out to some Roslyn folks to confirm.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
baronfel: Mostly that this work is targeted at the 7.0 work stream (through `main` branch) and so I wanted to go ahead and bump this. That can also be done at another time, I'd just need to update the tests to account for `latest` being `6.0` on `net7.0` TFM at this exact moment.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
baronfel: Good spot, fixing that now.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
baronfel: I've roped them in on an email for other parts of this (see my review comments below) so we should get alignment on it one way or the other.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props
Youssef1313: @baronfel Can we reach a point where WarningLevel is never set? For .NET Framework for example where we might not do anything for it in Microsoft.NET.Sdk.Analyzers.targets?

Is it safer to move the deleted line into a PropertyGroup after:

https://github.com/dotnet/sdk/blob/4a495a322321e95c0c8da15cff6c857aacf9d00a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L1158

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -65,25 +59,15 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-  <!-- Set the default WarningLevel based on EffectiveAnalysisLevel
-       For .NET 6 we want the Warning level to be 6 -->
-  <PropertyGroup Condition="'$(EffectiveAnalysisLevel)' != '' And
-                             $([MSBuild]::VersionGreaterThanOrEquals($(EffectiveAnalysisLevel), '6.0'))">
-    <WarningLevel Condition="'$(Language)' == 'C#'">6</WarningLevel>
-  </PropertyGroup>
+    <!-- If the user specified 'preview' we want to pick a very high warning level to opt into the highest possible warning wave -->
+    <WarningLevel Condition="'$(Language)' == 'C#' And '$(WarningLevel)' == '' And '$(AnalysisLevel)' == 'preview'">9999</WarningLevel>
+    <!-- NOTE: at this time only the C# compiler supports warning waves like this. -->
+    <WarningLevel Condition="'$(Language)' == 'C#' And '$(WarningLevel)' == ''">$(_TargetFrameworkVersionWithoutV.Substring(0, 1))</WarningLevel>
```
Youssef1313, line 65: Would the second `WarningLevel` always overwrite the first one?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props
baronfel: Another way to do this would be to put the default set in `CSharp.targets` instead of `CSharp.props`, and in `CSharp.targets`

* set the warning level if not set at this point (post-project-file-import occurring),
* also set a `_WarningLevelWasSet` boolean property that we could use instead of `WarningLevel == ''` in the Analyzers target

I believe I've seen this pattern elsewhere in the SDK.

This illustrates that there should probably be a .NET Framework test as well for this behavior!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
baronfel: This second one shouldn't be hit because the WarningLevel should be 9999 at this point, so the second condition should fail. I can definitely add test cases to cover this scenario, though. "When analysis level is preview warning level is 9999".

Youssef1313: > This second one shouldn't be hit because the WarningLevel should be 9999 at this point, so the second condition should fail.

Oh yes. My bad!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -15,16 +15,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Analysis level is a single property that can be used to control both the compiler warning waves
          and enable .NET analyzers. Valid values are 'none', 'latest', 'preview', or a version number  -->
 
-    <!-- Automatically set AnalysisLevel to 5 if targeting .net 5 and nothing else specified -->
+    <!-- If not specified and greater than or equal to .NET 5, default the AnalysisLevel to the the MAJ.MIN version -->
```
Youssef1313, line 18:
```suggestion
    <!-- If not specified and greater than or equal to .NET 5, default the AnalysisLevel to latest -->
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props
baronfel: Of course, we could also just set it to 4 here in the Analyzers.targets if the project is a .NETFramework project, which I've done in the interim here as an example.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -15,16 +15,10 @@ Copyright (c) .NET Foundation. All rights reserved.
...
     <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
                               '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                              $([MSBuild]::VersionEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">5.0</AnalysisLevel>
-
-    <!-- Automatically set AnalysisLevel if not specified -->
-    <AnalysisLevel Condition="'$(AnalysisLevel)' == '' And
-                              (('$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
-                               $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))) Or
-                               '$(AnalysisMode)' != '')">latest</AnalysisLevel>
+                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '5.0'))">latest</AnalysisLevel>
```
baronfel, line 21: TODO: add explicit handling for 5/6/7 based on TFM so that the TFM caps the AnalysisLevel to match

baronfel: This is now done.

baronfel:
I would love to add a test that computes the uses TFM of the test project to assert that the `latest` and `preview` AnalysisLevels (and the warning level) float to the TFMs as we update them. Ideally the test would fail as soon as the TFM is changed from net7.0 to net8.0, so that we know that some update here has to be done.

jmarolf (approved):

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
```diff
@@ -0,0 +1,170 @@
...
+                }
+            };
+
+            var testAsset = _testAssetsManager
+                .CreateTestProject(testProject, identifier: "warningLevelConsoleApp"+tfm, targetExtension: ".csproj");
+
+            var buildCommand = new GetValuesCommand(
+                Log,
+                Path.Combine(testAsset.TestRoot, testProject.Name),
+                tfm, "WarningLevel")
```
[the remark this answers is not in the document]
baronfel: Only one property/Item, but it can get the values of a property or the items of an itemgroup with metadata. We do a build and inject a target that computes this [here](https://github.com/dotnet/sdk/blob/5d2faf1bf54d4aa3fb81537c1359d1df2f6a0b8a/src/Tests/Microsoft.NET.TestFramework/Commands/GetValuesCommand.cs#L64-L126)

cremor:
@baronfel 
> WarningLevel should be set to 4 if the project is a .NET Framework project

Why is this rule/condition on the framework type and not on the C# language version?

If I remember correctly, my .NET Framework project which uses an SDK-style csproj file with `<LangVersion>10.0</LangVersion>` and `<AnalysisLevel>latest</AnalysisLevel>` used warning level 6 with the .NET 6 SDK. Now with the .NET 7 SDK it only uses warning level 4 again.

Yes, I can explicitly specify the `<WarningLevel>`, but shouldn't setting `<LangVersion>` and `<AnalysisLevel>` be enough to say "I want the latest compiler features and warnings"?

Btw, this also affects documentation, see dotnet/docs#33559

sliekens:
@baronfel I think this PR introduced a regression.

Before this PR, the `WarningLevel` could be set to match the version number of the `EffectiveAnalysisLevel`, but this code was removed. Now, setting `AnalysisLevel=preview` still results in `WarnLevel=9999` but any other value for `AnalysisLevel` will be ignored and the effective `WarnLevel` is derived from the TFM.

First I thought this was a documentation issue.
https://github.com/dotnet/docs/issues/41295

Now I believe it is a regression.

PR #24836, opened 2022-04-13
Update the hardcoded net6.0 to the current framework

v-wuzhai:


dsplaisted (approved):

dsplaisted:
@marcpopMSFT Should these be targeting 6.0.1xx, 6.0.4xx, or main?

marcpopMSFT:
@v-wuzhai can you target test changes as 6.0.4xx from now on? While 6.0.1xx is still supported for another year, I think we have updated enough tests in .1xx to have enough coverage to last us. .4xx will be supported for another 2.5 years and should be our ongoing focus.

Issue #15149, opened 2021-01-03
Need better logging for the native SDK resolver


KirillOsenkov:
I don't see a way to debug/investigate the decision process of the SDK resolver native .dll.

I've gotten as far as 
https://github.com/dotnet/sdk/blob/917b98c1388c3fd7cc8fa5d2ee7f22ba01575747/src/Resolvers/Microsoft.DotNet.SdkResolver/NETCoreSdkResolverNativeWrapper.cs#L21-L23

but from then on it's a black box and I don't see anything that would explain the resulting SDK that comes out of it. I want to know what factors influenced the decision:

 * did it use an environment variable?
 * did it use PATH?
 * did it use global.json?
 * which global.json?
 * was IsRunningInVisualStudio specified?

The resolver should explain its choice and clearly communicate which factors made a difference and which didn't. Otherwise I have to resort to cargo culting it and so far unsuccessfully.

KirillOsenkov:
As an example, https://github.com/dotnet/msbuild/issues/5999 is an issue that took me a whole day to figure out, and as a result of that investigation I have emerged with the knowledge that I should have set these two environment variables to influence the SDK resolution process:

```
    DOTNET_INSTALL_DIR = C:\msbuild\.dotnet
    DOTNET_MULTILEVEL_LOOKUP = 0
```

Had there been a log entry like "choosing 3.1.100 because the DOTNET_INSTALL_DIR was set explicitly and pointed to a 3.1.100 SDK at C:\msbuild\.dotnet" I could have binged an entire season of something wholesome today instead of debugging deep into the SDK resolution.

Currently the only logging I see is 
![image](https://user-images.githubusercontent.com/679326/103470642-8d075d00-4d29-11eb-9aae-aa396f33b2e7.png)

KirillOsenkov:
Correction, I **also** had to set `PATH=C:\msbuild\.dotnet;%PATH%`

rainersigwald:
I don't know if the logging interface currently exposed by MSBuild makes this possible--it's fairly minimal. We might need to augment that first before improving the dotnet resolver.

dsplaisted:
@KirillOsenkov Are you saying you need more visibility into what the native hostfxr_resolve_sdk2 method is doing, or what the managed code calling it does with the result? (Or are you not sure?)

KirillOsenkov:
[...]

PR #24870, opened 2022-04-14
simplify arch for preload library

kasperk81:


kasperk81:
@marcpopMSFT tiny change removing a hack

marcpopMSFT:
Enum.GetName may not be the ideal solution so I'd prefer to avoid that (Daniel believes that'll use reflection). We don't want every value from RuntimeEnformation but is there a simpler way of mapping this onto the three values we do want?

src/Resolvers/Microsoft.DotNet.NativeWrapper/Interop.cs
```diff
@@ -39,8 +39,7 @@ static Interop()
         private static void PreloadWindowsLibrary(string dllFileName)
         {
             string basePath = Path.GetDirectoryName(typeof(Interop).Assembly.Location);
-            string architecture = IntPtr.Size == 8 ? "x64" : "x86";
-            architecture = RuntimeInformation.ProcessArchitecture == Architecture.Arm64 ? "arm64" : architecture;
+            string architecture = Enum.GetName(typeof(Architecture), RuntimeInformation.ProcessArchitecture).ToLowerInvariant();
             string dllPath = Path.Combine(basePath, architecture, $"{dllFileName}.dll");
```
kasperk81, line 43:
```suggestion
            string architecture = RuntimeInformation.ProcessArchitecture.ToString().ToLowerInvariant();
            string dllPath = Path.Combine(basePath, architecture, $"{dllFileName}.dll");
```

kasperk81:
removed `Enum.GetName`. it came up in https://github.com/dotnet/sdk/issues/15149, for each architecutre, we should use separate path. rather than hardcoding the architecture, we can use architecture provided by core library to make it future proof.

skundlony (approved):

marcpopMSFT (approved):

PR #24873, opened 2022-04-14
Support multiple runtime packs per FrameworkReference

### LINKED ISSUE #24077: Question about supporting runtime packs for "parent" RIDs
The Android (32) workload currently ships four runtime packs, one for each Android architecture:

  * `Microsoft.Android.Runtime.32.android-arm`
  * `Microsoft.Android.Runtime.32.android-arm64`
  * `Microsoft.Android.Runtime.32.android-x86`
  * `Microsoft.Android.Runtime.32.android-x64`

In order to reduce our installation footprint we have a desire to introduce a new runtime pack for managed assemblies that are not architecture specific, but are still only intended to be used with the parent `android` RID. Right now these managed assemblies are duplicated in all four packs mentioned above. Ideally, we would move the non-architecture specific assemblies into a new pack:

  * `Microsoft.Android.Runtime.32.android`
  
I tried to get this to work by adding a second `RuntimePackNamePatterns` value to the `KnownFrameworkReference` we declare:

```diff
<KnownFrameworkReference
    Include="Microsoft.Android"
    TargetFramework="net6.0"
    RuntimeFrameworkName="Microsoft.Android"
    LatestRuntimeFrameworkVersion="**FromWorkload**"
    TargetingPackName="Microsoft.Android.Ref.32"
    TargetingPackVersion="**FromWorkload**"
-   RuntimePackNamePatterns="Microsoft.Android.Runtime.32.**RID**"
+   RuntimePackNamePatterns="Microsoft.Android.Runtime.32.android;Microsoft.Android.Runtime.32.**RID**"
    RuntimePackRuntimeIdentifiers="android-arm;android-arm64;android-x86;android-x64"
    Profile="Android"
/>
```

The `ProcessFrameworkReferences` task handles this approach, but the [ResolveFrameworkReferences](https://github.com/dotnet/sdk/blob/6706529258c3b6beac968ed4a0d9ddef30f1c73a/src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs#L29) task breaks as it only allows for a 1:1 relationship between runtime pack and framework reference:

```
Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(326,5): error MSB4018: System.ArgumentException: An item with the same key has already been added. Key: Microsoft.Android
```

I'm wondering how risky or big of a breaking change it would be to change the `RuntimePack*` metadata on `ResolvedFrameworkReferences`, and do something like this instead:
```diff
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
@@ -26,7 +26,6 @@ protected override void ExecuteCore()
             }

             var resolvedTargetingPacks = ResolvedTargetingPacks.ToDictionary(tp => tp.ItemSpec, StringComparer.OrdinalIgnoreCase);
-            var resolvedRuntimePacks = ResolvedRuntimePacks.ToDictionary(rp => rp.GetMetadata(MetadataKeys.FrameworkName), StringComparer.OrdinalIgnoreCase);

             var resolvedFrameworkReferences = new List<TaskItem>(FrameworkReferences.Length);

@@ -48,12 +47,11 @@ protected override void ExecuteCore()
                 resolvedFrameworkReference.SetMetadata("TargetingPackVersion", targetingPack.GetMetadata(MetadataKeys.NuGetPackageVersion));
                 resolvedFrameworkReference.SetMetadata("Profile", targetingPack.GetMetadata("Profile"));

-                ITaskItem runtimePack;
-                if (resolvedRuntimePacks.TryGetValue(frameworkReference.ItemSpec, out runtimePack))
-                {
-                    resolvedFrameworkReference.SetMetadata("RuntimePackPath", runtimePack.GetMetadata(MetadataKeys.PackageDirectory));
-                    resolvedFrameworkReference.SetMetadata("RuntimePackName", runtimePack.GetMetadata(MetadataKeys.NuGetPackageId));
-                    resolvedFrameworkReference.SetMetadata("RuntimePackVersion", runtimePack.GetMetadata(MetadataKeys.NuGetPackageVersion));
+                var matchingPacks = ResolvedRuntimePacks.Where(rp => rp.GetMetadata(MetadataKeys.FrameworkName).Equals(frameworkReference.ItemSpec, StringComparison.OrdinalIgnoreCase));
+                if (matchingPacks.Any()) {
+                    resolvedFrameworkReference.SetMetadata("RuntimePackPath", string.Join (";", matchingPacks.Select (mp => mp.GetMetadata(MetadataKeys.PackageDirectory))));
+                    resolvedFrameworkReference.SetMetadata("RuntimePackName", string.Join (";", matchingPacks.Select (mp => mp.GetMetadata(MetadataKeys.NuGetPackageId))));
+                    resolvedFrameworkReference.SetMetadata("RuntimePackVersion", string.Join (";", matchingPacks.Select (mp => mp.GetMetadata(MetadataKeys.NuGetPackageVersion))));
                 }

                 resolvedFrameworkReferences.Add(resolvedFrameworkReference);
```
We may also want to update the metadata names, but that might cause more breakage. These changes would produce `;` separated `RuntimePack*` metadata values, for example:

```
RuntimePackName = Microsoft.Android.Runtime.32.android;Microsoft.Android.Runtime.32.android-x64
```
These changes to `ResolveFrameworkReferences` appear to work as needed for the Android use case. I've also got something working on the workload side by [introducing a new `FrameworkReference` for Android](https://github.com/xamarin/xamarin-android/pull/6785/files), but I am not convinced that this `FrameworkReference` addition is the most sane or desirable approach.

Maybe there are other approaches that would support a "generic RID" runtime pack that I am not considering?

pjcollins:
Fixes: https://github.com/dotnet/sdk/issues/24077

The `<ProcessFrameworkReferences/>` task is already able to process a
`@(KnownFrameworkReference)` with multiple `%(RuntimePackNamePatterns)`
defined in a semi-colon delimited list.  However, if multiple runtime
pack patterns are used the `<ResolveFrameworkReferences/>` will fail
soon after with the following error:

    C:\Program Files\dotnet\sdk\6.0.300-preview.22173.2\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(326,5): error MSB4018: The "ResolveFrameworkReferences" task failed unexpectedly.
    C:\Program Files\dotnet\sdk\6.0.300-preview.22173.2\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(326,5): error MSB4018: System.ArgumentException: An item with the same key has already been added. Key: Microsoft.Android

Fixes `<ResolveFrameworkReferences/>` to allow multiple runtime packs to
be associated with a single `@(FrameworkReference)`.  The following
`@(ResolvedFrameworkReferences)` metadata will now contain semi-colon
delimited lists of runtime pack information if multiple patterns are
provided:

    %(ResolvedFrameworkReferences.RuntimePackPath) = path1;path2
    %(ResolvedFrameworkReferences.RuntimePackName) = pack1;pack2.rid
    %(ResolvedFrameworkReferences.RuntimePackVersion) = 7.0.0;7.0.0

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
@@ -26,7 +26,6 @@ namespace Microsoft.NET.Build.Tasks
             }
 
             var resolvedTargetingPacks = ResolvedTargetingPacks.ToDictionary(tp => tp.ItemSpec, StringComparer.OrdinalIgnoreCase);
-            var resolvedRuntimePacks = ResolvedRuntimePacks.ToDictionary(rp => rp.GetMetadata(MetadataKeys.FrameworkName), StringComparer.OrdinalIgnoreCase);
 
             var resolvedFrameworkReferences = new List<TaskItem>(FrameworkReferences.Length);
 
@@ -48,13 +47,14 @@ namespace Microsoft.NET.Build.Tasks
                 resolvedFrameworkReference.SetMetadata("TargetingPackVersion", targetingPack.GetMetadata(MetadataKeys.NuGetPackageVersion));
                 resolvedFrameworkReference.SetMetadata("Profile", targetingPack.GetMetadata("Profile"));
 
-                ITaskItem runtimePack;
-                if (resolvedRuntimePacks.TryGetValue(frameworkReference.ItemSpec, out runtimePack))
+                // Allow more than one runtime pack to be associated with this FrameworkReference
+                var matchingRuntimePacks = ResolvedRuntimePacks.Where(rp => rp.GetMetadata(MetadataKeys.FrameworkName).Equals(frameworkReference.ItemSpec, StringComparison.OrdinalIgnoreCase));
+                if (matchingRuntimePacks.Any())
                 {
-                    resolvedFrameworkReference.SetMetadata("RuntimePackPath", runtimePack.GetMetadata(MetadataKeys.PackageDirectory));
-                    resolvedFrameworkReference.SetMetadata("RuntimePackName", runtimePack.GetMetadata(MetadataKeys.NuGetPackageId));
-                    resolvedFrameworkReference.SetMetadata("RuntimePackVersion", runtimePack.GetMetadata(MetadataKeys.NuGetPackageVersion));
-                }
+                    resolvedFrameworkReference.SetMetadata("RuntimePackPath", string.Join (";", matchingRuntimePacks.Select (mrp => mrp.GetMetadata(MetadataKeys.PackageDirectory))));
+                    resolvedFrameworkReference.SetMetadata("RuntimePackName", string.Join (";", matchingRuntimePacks.Select (mrp => mrp.GetMetadata(MetadataKeys.NuGetPackageId))));
+                    resolvedFrameworkReference.SetMetadata("RuntimePackVersion", string.Join (";", matchingRuntimePacks.Select (mrp => mrp.GetMetadata(MetadataKeys.NuGetPackageVersion))));
+                 }
 
                 resolvedFrameworkReferences.Add(resolvedFrameworkReference);
             }
--- a/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveFrameworkReferencesTask.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveFrameworkReferencesTask.cs
@@ -1,0 +1,88 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
+using Microsoft.Build.Framework;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
+namespace Microsoft.NET.Build.Tasks.UnitTests
+{
+    public class GivenAResolveFrameworkReferences
+    {
+        [Fact]
+        public void It_resolves_with_multiple_runtime_packs()
+        {
+            var task = new ResolveFrameworkReferences
+            {
+                FrameworkReferences = new[]
+                {
+                    new MockTaskItem("Microsoft.NETCore.App", new Dictionary<string, string>()),
+                    new MockTaskItem("Microsoft.Android", new Dictionary<string, string>()),
+                },
+
+                ResolvedTargetingPacks = new[]
+                {
+                    new MockTaskItem("Microsoft.NETCore.App",
+                        new Dictionary<string, string>()
+                        {
+                            {"NuGetPackageId", "Microsoft.NETCore.App.Ref"},
+                            {"NuGetPackageVersion", "6.0.2"},
+                            {"Path", "empty"},
+                        }),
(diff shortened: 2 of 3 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
```diff
@@ -48,13 +47,14 @@ protected override void ExecuteCore()
...
+                if (matchingRuntimePacks.Any())
                 {
-                    resolvedFrameworkReference.SetMetadata("RuntimePackPath", runtimePack.GetMetadata(MetadataKeys.PackageDirectory));
-                    resolvedFrameworkReference.SetMetadata("RuntimePackName", runtimePack.GetMetadata(MetadataKeys.NuGetPackageId));
-                    resolvedFrameworkReference.SetMetadata("RuntimePackVersion", runtimePack.GetMetadata(MetadataKeys.NuGetPackageVersion));
-                }
+                    resolvedFrameworkReference.SetMetadata("RuntimePackPath", string.Join (";", matchingRuntimePacks.Select (mrp => mrp.GetMetadata(MetadataKeys.PackageDirectory))));
+                    resolvedFrameworkReference.SetMetadata("RuntimePackName", string.Join (";", matchingRuntimePacks.Select (mrp => mrp.GetMetadata(MetadataKeys.NuGetPackageId))));
+                    resolvedFrameworkReference.SetMetadata("RuntimePackVersion", string.Join (";", matchingRuntimePacks.Select (mrp => mrp.GetMetadata(MetadataKeys.NuGetPackageVersion))));
```
dsplaisted, line 56: I don't see any changes in code that consumes these metadata values.  Is that code going to be able to handle the multiple runtime packs?

dsplaisted (approved):
Do you only need this for 7.0?  That's what main is targeting.

Were you able to test this end-to-end yet?  It's OK if not, but I wonder if there are other parts of the code that will need to be updated to handle this.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
pjcollins: I couldn't find any references in dotnet/sdk, aside from the tests that only make sure that this metadata is set. Maybe there are external usages that could break? Are you familiar with anything else that depends on these?

pjcollins:
As far as Android is concerned we would only want this for 7.0.100. This is working end to end for the couple of cases that I tested (single and multi rid build+run attempts on device). However, I only tested this patch after applying it to a .NET 6.0.100 SDK, as we aren't compatible with 7.0.100 previews yet.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveFrameworkReferences.cs
dsplaisted: I think it's used somewhere but I'm not sure.  If it's really not used, then we don't even need to set the metadata at all (though it could be useful for debugging).

pjcollins: @dsplaisted are there any other folks we should loop in to determine whether we should remove this metadata or not? I'm not sure how to proceed.

PR #24928, opened 2022-04-19
Cache pipeline inputs when suppressed:

chsienki:
- When razor suppression is true, treat all pipeline inputs as cached
- Treat razor options as cached if only the supression flag has changed
- Update tests to include output code for regression testing


chsienki:
@RikkiGibson @cston for review please

src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
```diff
@@ -8,8 +8,43 @@
...
+        {
+            _equals = equals ?? EqualityComparer<T>.Default.Equals;
+        }
+
+        public bool Equals((T Left, RazorSourceGenerationOptions Right) x, (T Left, RazorSourceGenerationOptions Right) y)
+        {
+            if (y.Right.SuppressRazorSourceGenerator)
+            {
+                // If source generation is suppressed, we can always use previously cached results.
+                return true;
```
cston, line 24: >return true; [](http://example.com/codeflow?start=16&length=12)

Does this mean `RazorSourceGeneratorComparer<T>` is not symmetric? That is `Equals(x, y)` does not necessarily imply `Equals(y, x)`?

chsienki: Yes, its actually designed for that to specifically be the case. Any time the options flip from unsuppressed -> suppressed we want to override and pretend everything is cached and do no work. When we switch back to unsuppressed we want to do the actual check.  It's taking advantage of knowing how the generator internals work (https://sourceroslyn.io/#Microsoft.CodeAnalysis/SourceGeneration/Nodes/NodeStateTable.cs,386), so a a horrible hack frankly but necessary to support razors odd behavior caused by the design time / runtime split (something we're working on removing which would remove all these required hacks). 

> And does Equals(x, y) guarantee GetHashCode(x) == GetHashCode(y)

It does not, but again we know this will never actually be used in a hash table, so it's never called. I think I'll add some comments on this to make it clear its a very specialized piece of code that shouldn't be re-used generally.

src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
```diff
@@ -8,8 +8,43 @@
...
     internal static class IncrementalValuesProviderExtensions
     {
+        internal static IncrementalValueProvider<T> AsCachedIfSuppressed<T>(this IncrementalValueProvider<T> provider, IncrementalValueProvider<RazorSourceGenerationOptions> options)
+            where T : notnull
+        {
+            return provider.Combine(options).WithComparer(new RazorSourceGeneratorComparer<T>()).Select((pair, _) => pair.Left);
+        }
+
+        internal static IncrementalValuesProvider<T> AsCachedIfSuppressed<T>(this IncrementalValuesProvider<T> provider, IncrementalValueProvider<RazorSourceGenerationOptions> options)
+    where T : notnull
```
cston, line 42: >where [](http://example.com/codeflow?start=4&length=5)

Indent?

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
```diff
@@ -49,11 +49,14 @@ internal class RazorSourceGenerationOptions : IEquatable<RazorSourceGenerationOp
         public bool SupportLocalizedComponentNames { get; set; } = false;
 
         public bool Equals(RazorSourceGenerationOptions other)
+                   => SuppressRazorSourceGenerator == other.SuppressRazorSourceGenerator && EqualsIgnoringSupression(other);
+
+        public bool EqualsIgnoringSupression(RazorSourceGenerationOptions other)
```
cston, line 54: >EqualsIgnoringSupression [](http://example.com/codeflow?start=20&length=24)

Should this consider `WaitForDebugger` as well?

cston:
>     internal class RazorSourceGenerationOptions : IEquatable<RazorSourceGenerationOptions>

Please consider marking `sealed` to simplify reasoning about `Equals()`.

---
Refers to: src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs:11 in 2d6359e. [](commit_id = 2d6359e05f259ad68c63718bdfb98acc2c8cfcfa, deletion_comment = False)

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -22,11 +22,17 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
                 .Select(ComputeRazorSourceGeneratorOptions);
-            var razorSourceGeneratorOptions = razorSourceGeneratorOptionsWithDiagnostics.ReportDiagnostics(context);
+            var liveRazorSourceGeneratorOptions = razorSourceGeneratorOptionsWithDiagnostics.ReportDiagnostics(context);
 
-            var sourceItemsWithDiagnostics = context.AdditionalTextsProvider
+            var analyzerConfigOptions = context.AnalyzerConfigOptionsProvider.AsCachedIfSuppressed(liveRazorSourceGeneratorOptions);
+            var additionalTexts = context.AdditionalTextsProvider.AsCachedIfSuppressed(liveRazorSourceGeneratorOptions);
+            var parseOptions = context.ParseOptionsProvider.AsCachedIfSuppressed(liveRazorSourceGeneratorOptions);
+            var compilation = context.CompilationProvider.AsCachedIfSuppressed(liveRazorSourceGeneratorOptions);
+            var razorSourceGeneratorOptions = liveRazorSourceGeneratorOptions.WithLambdaComparer((l, r) => l.EqualsIgnoringSupression(r), l => l.GetHashCode());
```
cston, line 31: >l.EqualsIgnoringSupression(r) [](http://example.com/codeflow?start=107&length=29)

Why are we ignoring suppression here rather than changing the `RazorSourceGeneratorOptions`? When do the options change?

cston:
>     internal class LambdaComparer<T> : IEqualityComparer<T>

`sealed`

---
Refers to: src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs:89 in 2d6359e. [](commit_id = 2d6359e05f259ad68c63718bdfb98acc2c8cfcfa, deletion_comment = False)

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
chsienki: It was no longer used, so have removed.

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
chsienki: When we go from suppressed -> unsuppressed, we don't want to note the options as having been changed unless any of the other options actually have changed. 

Thinking about it, I wonder if it might be more obvious if we break out the suppression from the rest of the options. I'll take a quick look what that might look like

chsienki: Ok, have broken the suppression out into its own input which simplifies the options significantly

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -37,12 +39,51 @@ public async Task SourceGenerator_RazorFiles_Works()
             var compilation = await project.GetCompilationAsync();
             var driver = await GetDriverAsync(project);
 
-            var result = RunGenerator(compilation!, ref driver);
+            var result = RunGenerator(compilation!, ref driver)
+                            .VerifyPageOutput(
+@"#pragma checksum ""Pages/Index.razor"" ""{ff1816ec-aa5e-4d10-87f7-6f4963833460}"" ""6b5db227a6aa2228c777b0771108b184b1fc5df3""
```
pranavkm, line 44: FYI - Historically this was bothersome because the codegen would change for existing code and it would block insertion PRs and it was non trivial to get this string. The razor-compiler had an option to regen it but it’s a bit of a pain to use and makes diffs harder to work with.

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -19,17 +19,23 @@ public partial class RazorSourceGenerator : IIncrementalGenerator
...
-                .Select(ComputeRazorSourceGeneratorOptions);
-            var razorSourceGeneratorOptions = razorSourceGeneratorOptionsWithDiagnostics.ReportDiagnostics(context);
+            // determine if we should suppress this run and set all the inputs to cached if so
+            var isGeneratorSuppressed = context.AnalyzerConfigOptionsProvider.Select(GetSupressionStatus);
 
-            var sourceItemsWithDiagnostics = context.AdditionalTextsProvider
-                .Where(static (file) => file.Path.EndsWith(".razor", StringComparison.OrdinalIgnoreCase) || file.Path.EndsWith(".cshtml", StringComparison.OrdinalIgnoreCase))
-                .Combine(context.AnalyzerConfigOptionsProvider)
-                .Select(ComputeProjectItems);
+            var analyzerConfigOptions = context.AnalyzerConfigOptionsProvider.AsCachedIfSuppressed(isGeneratorSuppressed);
```
pranavkm, line 25: Very nicely done! 👍🏽

src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
```diff
@@ -8,8 +8,55 @@
 
 namespace Microsoft.NET.Sdk.Razor.SourceGenerators
 {
+    /// <summary>
+    /// A highly specialized comparer that allows for treating an event source as cached if the razor options are set to suppress generation.
+    /// </summary>
+    /// <remarks>
+    /// This should not be used outside of <see cref="IncrementalValuesProviderExtensions.AsCachedIfSuppressed{T}(IncrementalValueProvider{T}, IncrementalValueProvider{RazorSourceGenerationOptions})"/>
```
cston, line 15: >IncrementalValuesProviderExtensions [](http://example.com/codeflow?start=54&length=35)

Consider making `RazorSourceGeneratorComparer<T>` a private nested class in `IncrementalValuesProviderExtensions`, to reduce the chance this will be used elsewhere.

src/RazorSdk/SourceGenerators/IncrementalValueProviderExtensions.cs
```diff
@@ -8,8 +8,55 @@
...
+            return _equals(x.Left, y.Left);
+        }
+
+        public int GetHashCode((T Left, bool IsSuppressed) obj) => throw new NotImplementedException("GetHashCode is never expected to be called");
+    }
+
     internal static class IncrementalValuesProviderExtensions
     {
+        /// <summary>
+        /// Adds a comparer that will force the provider to be considered as cached if the razor options call for supression
```
cston, line 42: >supression [](http://example.com/codeflow?start=114&length=10)

Typo, here and below.

src/RazorSdk/SourceGenerators/RazorSourceGenerator.RazorProviders.cs
```diff
@@ -14,6 +14,20 @@ namespace Microsoft.NET.Sdk.Razor.SourceGenerators
...
     {
+        /// <summary>
+        /// Gets a flag that determines if the source generator should no-op.
+        /// <para>
+        /// This flag exists to support scenarios in VS where design-time and EnC builds need
+        /// to run without invoking the source generator to avoid duplicate types being produced.
+        /// The property is set by the SDK via an editor config.
+        /// </para>
+        /// </summary>
+        private static bool GetSupressionStatus(AnalyzerConfigOptionsProvider optionsProvider, CancellationToken _)
```
cston, line 25: >Supression [](http://example.com/codeflow?start=31&length=10)

Typo.

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
```diff
@@ -8,34 +8,17 @@
...
 
-        /// <summary>
-        /// Gets a flag that determines if the source generator should no-op.
-        /// <para>
-        /// This flag exists to support scenarios in VS where design-time and EnC builds need
-        /// to run without invoking the source generator to avoid duplicate types being produced.
-        /// The property is set by the SDK via an editor config.
-        /// </para>
-        /// </summary>
         public bool SuppressRazorSourceGenerator { get; set; } = false;
```
cston, line 22: >SuppressRazorSourceGenerator [](http://example.com/codeflow?start=20&length=28)

Is this `SuppressRazorSourceGenerator` field still used? If so, why are we able to ignore the value in `Equals()` below? (Consider adding a comment, either here or in `Equals()`.)

cston:
>             var tagHelpersFromCompilation = context.CompilationProvider

`compilation`?

---
Refers to: src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs:91 in 28b694b. [](commit_id = 28b694bfe0e7d0d17a3f7da1daf441cea748c5a5, deletion_comment = False)

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -223,19 +211,12 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                 .Combine(importFiles.Collect())
                 .Combine(allTagHelpers)
                 .Combine(razorSourceGeneratorOptions)
-                .Combine(context.ParseOptionsProvider)
```
cston, line 226: >.Combine(context.ParseOptionsProvider) [](http://example.com/codeflow?start=16&length=38)

Was this unnecessary?

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -905,113 +2288,210 @@ <h1>Malformed h1
...
-            }
+            var updatedText = new TestAdditionalText("Pages/Index.razor", SourceText.From(@"<h1>Hello world 1</h1>", Encoding.UTF8));
+            driver = driver.ReplaceAdditionalText(additionalTexts.First(f => f.Path == updatedText.Path), updatedText);
 
-            foreach (var option in testOptionsProvider!.TestGlobalOptions.Options)
-            {
-                updatedOptionsProvider.TestGlobalOptions[option.Key] = option.Value;
-            }
+            // now run the generator with suppression
+            var supressedOptions = optionsProvider.Clone();
```
cston, line 2416: >supressed [](http://example.com/codeflow?start=16&length=9)

Typo.

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -905,113 +2288,210 @@ <h1>Malformed h1
...
+            driver = driver.WithUpdatedAnalyzerConfigOptions(supressedOptions);
 
-            updatedOptionsProvider.TestGlobalOptions["build_property.SuppressRazorSourceGenerator"] = "false";
+            // results should be the same (even though we changed text)
+            result = RunGenerator(compilation!, ref driver)
+                    .VerifyOutputsMatch(result);
 
-            driver = driver.WithUpdatedAnalyzerConfigOptions(updatedOptionsProvider);
-            result = RunGenerator(compilation!, ref driver);
+            // now unsupress and re-run
```
cston, line 2424: >unsupress [](http://example.com/codeflow?start=19&length=9)

Typo.

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -1164,4 +2664,68 @@ public override bool TryGetValue(string key, [NotNullWhen(true)] out string? val
...
+                if (diff is null)
+                {
+                    try
+                    {
+                        Assert.Equal(expected.GeneratedSources[i].SourceText.ToString(), actual.GeneratedSources[i].SourceText.ToString());
+                    }
+                    catch (EqualException e)
+                    {
+                        Assert.False(true, $"No diff supplied. But index {i} was:\r\n\r\n{e.Actual.Replace("\"", "\"\"")}");
+                    }
```
cston, line 2718: What is the `try/catch` for? Could we use `Assert.True(expectedText == actualText, $"... index {i} was ... {actual}");` instead?

src/RazorSdk/SourceGenerators/RazorSourceGenerationOptions.cs
chsienki: Removed. just an oversight.

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
chsienki: Yes, it wasn't actually being used.

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
chsienki: Ah yeah, that was just from where I was iterating to get a decent mechanism for comparison. That path is only used if you don't supply a diff at all, so we can just assert if they don't match.

cston (approved):

RikkiGibson (approved):

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -223,19 +205,12 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
                 .Select(static (pair, _) =>
                 {
-                    var ((((sourceItem, imports), allTagHelpers), razorSourceGeneratorOptions), parserOptions) = pair;
+                    var (((sourceItem, imports), allTagHelpers), razorSourceGeneratorOptions) = pair;
```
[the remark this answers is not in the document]
chsienki: Yeah, its something we should think about. 

One API we considered was `TOut Combine<TOut, TLeft, TRight>(IncrementalValueProvider<TLeft>, IncrementalValuesProvider<TRight>, Func<TOut, TLeft, TRight>);`

Essentially allow you to select at the same time as combine, so you can build up a datamodel as you go. You can do it today via chaining, but it would be nice to make it easier.

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -2723,7 +2723,7 @@ public static GeneratorRunResult VerifyOutputsMatch(this GeneratorRunResult actu
 
         private static string TrimChecksum(string text)
         {
-            var trimmed = text.Trim('\r', '\n').Replace("\r\n", "\n").Replace('\n', '\r');
+            var trimmed = text.Trim('\r', '\n').Replace("\r\n", "\r").Replace('\r', '\n');
```
RikkiGibson, line 2726: don't we have a .NormalizeWhitespace() or similar helper that can remove the guesswork here?

chsienki: We're in the SDK repo, so no, this is all new custom test stuff. We're going to move this over to the razor compiler repo, and I imagine i'll pull in some of the roslyn testing bits when we do.

PR #22977, opened 2021-12-09
Skip Manifest Update When Installing Wasm Tools for AoT Testing

TanayParikh:
https://github.com/dotnet/sdk/pull/22966#issuecomment-989471898

PR #24958, opened 2022-04-20
Update the wasm install to skip manifest updates

marcpopMSFT:
When installing workloads using our build feeds, we have to either use a rollback file or skip manifest updates. Otherwise, we'll be installing live builds of those manifests which may not work.

### DIFF
--- a/eng/build.yml
+++ b/eng/build.yml
@@ -238,7 +238,7 @@ jobs:
             BuildConfig: $(_BuildConfig)
             SYSTEM_ACCESSTOKEN: $(System.AccessToken)
             HelixAccessToken: ${{ parameters.HelixAccessToken }}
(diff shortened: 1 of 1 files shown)


TanayParikh (approved):
LGTM if CI passes. I'd tried this earlier (https://github.com/dotnet/sdk/pull/22977), however it led to a bunch of stale packages being used. 

Edit: Maybe that was because of `--no-cache`.

marcpopMSFT:
@TanayParikh It'll use whatever versions of the workloads are included in the SDK that's in global.json I believe. Do you need more up to date versions? If so, have you looked into using `dotnet workload update --from-rollback-file` (try `--print-rollback` to see the format)?  The problem is that the feeds we have configured have all builds on it so any workload commands that aren't careful could end up picking up breaking changes (we saw that last week when there was a breaking manifest from maui for about a day)

PR #24971, opened 2022-04-20
Allow file changes from tar as it's just a warning

marcpopMSFT:
I kept seeing tar errors in our builds. it turns out that tar returning 1 means the files changed during execution. Let's allow that to be more reliable in our builds as I think this is from helix and non-helix together.

The version of tar we use doesn't support the --warning flag for this so map a returncode of 1 back to zero. In parallel, I'm digging into how our version of tar is determined.

### DIFF
--- a/src/Tests/HelixTasks/TarGzFileCreateFromDirectory.cs
+++ b/src/Tests/HelixTasks/TarGzFileCreateFromDirectory.cs
@@ -110,7 +110,7 @@ namespace Microsoft.DotNet.SdkCustomHelix.Sdk
 
         protected override string GenerateCommandLineCommands()
         {
-            return $"--warning:no-file-changed {GetDestinationArchive()} {GetSourceSpecification()}";
+            return $"--warning=no-file-changed {GetDestinationArchive()} {GetSourceSpecification()}";
         }
 
         private string GetSourceSpecification()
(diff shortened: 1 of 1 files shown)


marcpopMSFT:
Per dngeng, tar comes from the OS we use so this is the best solution we have for the moment. Also, the tar on OSX may not support the warning suppression we'd want to do anyway.

Forgind (approved):

src/Tests/HelixTasks/TarGzFileCreateFromDirectory.cs
```diff
@@ -43,8 +43,9 @@ public sealed class TarGzFileCreateFromDirectory : ToolTask
         protected override int ExecuteTool(string pathToTool, string responseFileCommands, string commandLineCommands)
         {
             int returnCode = base.ExecuteTool(pathToTool, responseFileCommands, commandLineCommands);
-
-            if (IgnoreExitCode)
+            
+            // returnCode 1 indicates files changed during tar which is just a warning
```
Forgind, line 47:
```suggestion
            // returnCode 1 indicates that files changed during tar. In our case, it is likely being overwritten by another copy of the same file, so we can safely ignore this warning.
```

kasperk81:
looks good. does HelixTask.csproj require multiple frameworks or can it use a single `$(SdkTargetFramework)` which is currently net7.0? https://github.com/dotnet/sdk/blob/fa9625d61a62aaea2550f89977c7357758dd438b/src/Tests/HelixTasks/HelixTasks.csproj#L3

PR #24973, opened 2022-04-20
update exception message for invalid rollback file

### LINKED ISSUE #24680: Bad exception when rollback file has invalid version
Run `dotnet workload update --from-rollback-file rollback.json` where rollback.json has the following contents:

json
```
{
    "microsoft.net.sdk.android": "31.0.300-rc.1.35/6.0.300",
    "microsoft.net.sdk.ios": "15.4.100-rc.1.104/6.0.300",
    "microsoft.net.sdk.maccatalyst": "15.4.100-rc.1.104/6.0.300",
    "microsoft.net.sdk.macos": "12.3.100-rc.1.104/6.0.300",
    "microsoft.net.sdk.maui": "6.0.300-rc.1.5299/6.0.300",
    "microsoft.net.sdk.tvos": "15.4.100-rc.1.104/6.0.300",
    "microsoft.net.workload.mono.toolchain": "6.0.0.15/6.0.100",
    "microsoft.net.workload.emscripten": "6.0.0/6.0.100"
}
```

Expected: Message that 6.0.0.15 isn't a valid version number
Actual:

> Workload update failed: Value cannot be null. (Parameter 'version')

The problem is the error handling in the `ManifestVersion` constructor.

gkulin:
Fixes #24680

Run `dotnet workload update --from-rollback-file rollback.json` where rollback.json has the following contents:

```
{
    "microsoft.net.sdk.android": "31.0.300-rc.1.35/6.0.300",
    "microsoft.net.sdk.ios": "15.4.100-rc.1.104/6.0.300",
    "microsoft.net.sdk.maccatalyst": "15.4.100-rc.1.104/6.0.300",
    "microsoft.net.sdk.macos": "12.3.100-rc.1.104/6.0.300",
    "microsoft.net.sdk.maui": "6.0.300-rc.1.5299/6.0.300",
    "microsoft.net.sdk.tvos": "15.4.100-rc.1.104/6.0.300",
    "microsoft.net.workload.mono.toolchain": "6.0.0.15/6.0.100",
    "microsoft.net.workload.emscripten": "6.0.0/6.0.100"
}
```

Original Message:

`Workload update failed: Value cannot be null. (Parameter 'version')`

New Message:
`Workload update failed: Value cannot be null. (Parameter 'Invalid version 6.0.0.15.')`

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
@@ -331,6 +331,6 @@
     <value>Checking for updated workload manifests.</value>
   </data>
   <data name="InvalidVersionForWorkload" xml:space="preserve">
-    <value>Error parsing version for workload {0}: Invalid version {1}.</value>
+    <value>Error parsing version '{1}' for workload manifest ID '{0}'</value>
   </data>
 </root>
--- a/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
@@ -330,4 +330,7 @@
   <data name="CheckForUpdatedWorkloadManifests" xml:space="preserve">
     <value>Checking for updated workload manifests.</value>
   </data>
+  <data name="InvalidVersionForWorkload" xml:space="preserve">
+    <value>Error parsing version for workload {0}: Invalid version {1}.</value>
+  </data>
 </root>
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -3,24 +3,23 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.DotNet.Cli.Utils;
-using Microsoft.NET.Sdk.WorkloadManifestReader;
 using System.IO;
 using System.Linq;
-using Microsoft.DotNet.Cli.NuGetPackageDownloader;
-using Microsoft.DotNet.ToolPackage;
+using System.Net.Http;
+using System.Runtime.InteropServices;
+using System.Text.Json;
 using System.Threading.Tasks;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.NuGetPackageDownloader;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Configurer;
+using Microsoft.DotNet.MSBuildSdkResolver;
+using Microsoft.DotNet.ToolPackage;
 using Microsoft.DotNet.Workloads.Workload.Install.InstallRecord;
 using Microsoft.Extensions.EnvironmentAbstractions;
-using NuGet.Versioning;
-using Microsoft.DotNet.Configurer;
+using Microsoft.NET.Sdk.WorkloadManifestReader;
 using NuGet.Common;
-using System.Text.Json;
-using System.Runtime.InteropServices;
-using Microsoft.DotNet.Cli;
-using System.Net;
-using System.Net.Http;
-using Microsoft.DotNet.MSBuildSdkResolver;
+using NuGet.Versioning;
 
 namespace Microsoft.DotNet.Workloads.Workload.Install
 {
@@ -473,7 +472,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                     SdkFeatureBand manifestFeatureBand;
                     var parts = manifest.Value.Split('/');
                     
-                    string manifestVersionString = (parts[0]).ToString();
+                    string manifestVersionString = (parts[0]);
                     if (!FXVersion.TryParse(manifestVersionString, out FXVersion version))
                     {
                         if (string.IsNullOrWhiteSpace(manifestVersionString))
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -480,7 +480,6 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                         {
                             throw new ArgumentNullException(nameof(manifestVersionString));
                         }
-                        //throw new FormatException($"Error parsing version for workload {manifest.Key}: Invalid version {manifestVersionString}.");
                         throw new FormatException(String.Format(LocalizableStrings.InvalidVersionForWorkload, manifest.Key, manifestVersionString));
                     }
                     manifestVersion = new ManifestVersion(parts[0]);
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -474,14 +474,15 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                     var parts = manifest.Value.Split('/');
                     
                     string manifestVersionString = (parts[0]).ToString();
-                    if (!FXVersion.TryParse(manifestVersionString, out FXVersion _version))
+                    if (!FXVersion.TryParse(manifestVersionString, out FXVersion version))
                     {
                         if (string.IsNullOrWhiteSpace(manifestVersionString))
                         {
                             throw new ArgumentNullException(nameof(manifestVersionString));
                         }
                         throw new FormatException(String.Format(LocalizableStrings.InvalidVersionForWorkload, manifest.Key, manifestVersionString));
-                    }
+                    } 
+
                     manifestVersion = new ManifestVersion(parts[0]);
                     if (parts.Length == 1)
                     {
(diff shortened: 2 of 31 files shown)


gkulin:
@joeloff @dsplaisted I thought about throwing a different/new type of exception in this case because the "Value cannot be null." part of the error message may not be correct. But I wasn't sure if that was necessary / how I could go about doing that. 
Providing the version that is causing the issue seems like an improvement from how it currently works.

joeloff:
> @joeloff @dsplaisted I thought about throwing a different/new type of exception in this case because the "Value cannot be null." part of the error message may not be correct. But I wasn't sure if that was necessary / how I could go about doing that. Providing the version that is causing the issue seems like an improvement from how it currently works.

I think just throwing a general `new Exception(.....)` should be fine. We already do that for other failures in the resolver

KalleOlaviNiemitalo:
Is it possible to actually get null there? This is a public constructor of a public class. Perhaps it should check for null after FXVersion.TryParse has failed, and then throw either ArgumentNullException or ArgumentException accordingly.

joeloff:
Ideally FXVersion should have a `Parse` method that throws different exceptions based on what failed. If you look at something like `System.Version`, the `Parse` method can throw `ArgumentNullException`, `ArgumentOutOfRangeException` and `FormatException`. I agree with @KalleOlaviNiemitalo that it would be better to handle both cases and use the `ArgumentException` for the non-null case.

gkulin:
@joeloff I made the change in order to handle both when it should throw a null exception and when it should form a format exception. I also updated the message to include the key that failed parsing as @baronfel mentioned. 

Is this something I should write a new test for?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -471,6 +472,17 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
                     ManifestVersion manifestVersion;
                     SdkFeatureBand manifestFeatureBand;
                     var parts = manifest.Value.Split('/');
+                    
+                    string manifestVersionString = (parts[0]).ToString();
+                    FXVersion _version;
```
joeloff, line 477: You can probably inline the declaration here so you have `TryParse(manifestVersionString, out FXVersion _version)`

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -471,6 +472,17 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
...
+                    if (!FXVersion.TryParse(manifestVersionString, out _version))
+                    {
+                        if (string.IsNullOrWhiteSpace(manifestVersionString))
+                        {
+                            throw new ArgumentNullException(nameof(manifestVersionString));
+                        }
+                        throw new FormatException($"Error parsing version for workload {manifest.Key}: Invalid version {manifestVersionString}.");
```
joeloff, line 484: The string should probably be in a resource so it can get localized as it would get surfaced to the user on the commandline when there's a problem with the rollback file.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -471,6 +472,17 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
...
                     var parts = manifest.Value.Split('/');
+                    
+                    string manifestVersionString = (parts[0]).ToString();
+                    if (!FXVersion.TryParse(manifestVersionString, out FXVersion _version))
+                    {
+                        if (string.IsNullOrWhiteSpace(manifestVersionString))
+                        {
+                            throw new ArgumentNullException(nameof(manifestVersionString));
+                        }
+                        //throw new FormatException($"Error parsing version for workload {manifest.Key}: Invalid version {manifestVersionString}.");
```
joeloff, line 483: You can probably delete this comment

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -471,6 +472,16 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
...
+                    string manifestVersionString = (parts[0]).ToString();
+                    if (!FXVersion.TryParse(manifestVersionString, out FXVersion _version))
+                    {
+                        if (string.IsNullOrWhiteSpace(manifestVersionString))
+                        {
+                            throw new ArgumentNullException(nameof(manifestVersionString));
+                        }
+                        throw new FormatException(String.Format(LocalizableStrings.InvalidVersionForWorkload, manifest.Key, manifestVersionString));
+                    }
                     manifestVersion = new ManifestVersion(parts[0]);
```
dsplaisted, line 485: The `ManifestVersion` constructor should be fixed too.  We can also check in this method as we can generate a better error message which incudes the manifest ID.

It's probably a good idea to add a constructor overload to `ManifestVersion` which takes an `FXVersion` as a parameter so we don't have to parse it twice.

src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
```diff
@@ -330,4 +330,7 @@
   <data name="CheckForUpdatedWorkloadManifests" xml:space="preserve">
     <value>Checking for updated workload manifests.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="InvalidVersionForWorkload" xml:space="preserve">
+    <value>Error parsing version for workload {0}: Invalid version {1}.</value>
```
dsplaisted, line 334:
```suggestion
    <value>Error parsing version `{1}` for workload manifest ID '{0}'</value>
```

@baronfel Any suggestions for the wording / formatting of this message?

baronfel: In lieu of any failure reasons for the parsing failure, I think your suggestion is good. It read more naturally to me. The biggest concern I had was making sure the workload Id was available in the error message, so users had better guidance to the root cause. I would suggest using single-quotes instead of backticks in the message, though.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -20,6 +20,7 @@
 using Microsoft.DotNet.Cli;
 using System.Net;
 using System.Net.Http;
+using Microsoft.DotNet.MSBuildSdkResolver;
```
joeloff, line 23: Nit, can you sort the usings (Ctrl+R+G inside VS)

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -471,6 +472,17 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
                     ManifestVersion manifestVersion;
                     SdkFeatureBand manifestFeatureBand;
                     var parts = manifest.Value.Split('/');
+                    
+                    string manifestVersionString = (parts[0]).ToString();
```
joeloff, line 476: Is `ToString` needed? Split should be returning a string[]

gkulin: Yeah, it's not needed I don't know why I had it there. Thanks!

joeloff (approved):

src/Tests/dotnet-workload-update.Tests/GivenDotnetWorkloadUpdate.cs
```diff
@@ -397,6 +398,34 @@ public void ApplyRollbackWithMultipleManifestsAcrossFeatureBand()
...
+            _reporter.Clear();
+            
+            var testDirectory = _testAssetsManager.CreateTestDirectory().Path;
+            var dotnetRoot = Path.Combine(testDirectory, "dotnet");
+            var userProfileDir = Path.Combine(testDirectory, "user-profile");
+            Directory.CreateDirectory(userProfileDir);
+
+            var sdkFeatureVersion = "6.0.200";
+            var workloadResolver = WorkloadResolver.CreateForTests(new MockManifestProvider(new[] { _manifestPath }), dotnetRoot, userLocal: false, userProfileDir);
+            var nugetDownloader = new MockNuGetPackageDownloader(dotnetRoot);
```
dsplaisted, line 413: I don't think you need these two lines anymore, it looks like those variables aren't being used.

src/Tests/dotnet-workload-update.Tests/GivenDotnetWorkloadUpdate.cs
```diff
@@ -397,6 +398,34 @@ public void ApplyRollbackWithMultipleManifestsAcrossFeatureBand()
...
+            
+            var mockRollbackFileContent = @"{""mock.workload"":""6.0.0.15/6.0.100""}";
+            var rollbackFilePath = Path.Combine(testDirectory, "rollback.json");
+            var manifestUpdater = new MockWorkloadManifestUpdater();
```
dsplaisted, line 417: This line also looks like it's not necessary.

dsplaisted (approved):

PR #25001, opened 2022-04-22
Assign names to arguments for nuget commands

baronfel:
The arguments for the NuGet subcommands don't have names, which means help output shows unhelpful names like `<ienumerable'1>` for them instead of something descriptive.

To fix this, I've gone through each command in its raw form from the NuGet.CommandLine.XPlat.dll and duplicated what it shows for these arguments as much as possible.

Longer term this should go away with the transition to System.CommandLine from NuGet, but for now small tweaks like this seem acceptable.


nkolev92 (commented):
One question/suggestions. 

Looks great though. 

We recently had this issue filled in our repo, and weren't sure if this fix needed to be the dotnet or nuget side, https://github.com/NuGet/Home/issues/11714.

nkolev92 (approved):

PR #25011, opened 2022-04-22
[ASP.NET] Fix pack for static web assets in cross-targeting scenarios

javiercn:
This change updates the pack support in Static Web Assets to better
account for cross targeting scenarios.

This change doesn't add support for having different assets based on
the target framework, it just ensures that the build pipeline behaves
correctly in such situations.

Previously we would "shim" support for cross-targeting builds by emitting
a file to disk on each of the inner builds indicating the static web assets
version.

This is problematic because the builds run in parallel and we would see intermittent
issues when the the multiple inner builds tried to write to the file at the same time.

This change updates the pack support in static web assets to correctly account for
multitargeting scenarios. The main changes are as follows:

* We will produce a pack manifest as part of the build process that is used to capture all
  the static web assets + supporting targets to pack and a list of items that need to be
  removed from content groups when packing (this is used for compatibility with static web assets V1)
* During pack, we will read the manifest, remove the files (if any) marked for packing that shouldn't
  be packed, and add the files for packing with their appropriate package paths as content items.
* On cross targeting builds there is an outer build and an inner build for each individual target framework.
  We have special targets that we import inside the outer build and that take care of selecting the assets
  we are going to use for packing. The steps to do this are as follow:
  * The outer build starts and dispatches inner builds (one per each target framework), which computes the
    list of pack items as well as the files to include for each target framework.
  * During the inner builds we avoid doing any packing work. We can detect this by checking for IsInnerBuild.
  * After the inner builds have completed, we run a target that takes care of performing the packing in the
    cross targeting environment.
    * We look at the target frameworks available during the build and compute a candidate target framework that has
      the highest static web asset version possible. We always want to pack using the highest static web assets
      version as they are "backwards" compatible, meaning a V1 target can consume V2 assets.
    * When computing the version we only care about the target framework identifier and version, we don't account
      for the platform nor the platform version as the assets should be generated in the same way for those.
    * We have the following relative precendence net6.0+, net5.0, netstandard2.0+ based on .net6.0+ being static
      web assets V2, net5.0 being static web assets V1 but including scoped css support and netstandard2.0/1 and
      netcoreapp3.1 being static web assets V1.
    * Any other target framework that we don't know about gets ignored (like .NET full framework) and if we don't
      find a suitable target framework, the rest of the steps no-op.
    * Once we've selected a target framework, we invoke a target to retrieve the paths to the build and pack manifests.
    * We read the build manifest if available and remove every static webasset that could be considered for packing.
    * We then read the pack manifest and remove any additional file to remove from the content items and then
      we add the files to pack with their respective package paths as Content.
* On single targeting builds we just generate the manifest and during the pack process we read it and perform the
  same steps described for the packing in the cross targeting build.

Pack has several variants that we need to account for packing. The matrix is as follows:
* GeneratePackageOnBuild
* pack --no-build
* cross targeting
* single target framework

In addition to those we need to consider the following variants for static web assets based on the version:
* netstandard2.0
* net5.0
* net6.0+
* netstandard2.0;net5.0
* net5.0;net6.0


src/RazorSdk/Tasks/StaticWebAssets/ComputeReferenceStaticWebAssetItems.cs
```diff
@@ -26,6 +26,8 @@ public class ComputeReferenceStaticWebAssetItems : Task
         [Required]
         public string Source { get; set; }
 
+        public bool UpdateSourceType { get; set; } = true;
```
javiercn, line 29: We now use the same logic for determining the assets to pack that we use when determining the assets that a referencing project can consume. In other words, the logic for determining the assets via ProjectReference and PackageReference is now the same.

src/RazorSdk/Tasks/StaticWebAssets/StaticWebAssetsGeneratePackManifest.cs
```diff
@@ -0,0 +1,125 @@
...
+            WriteIndented = true
+        };
+
+        [Required]
+        public ITaskItem[] Assets { get; set; }
+
+        [Required]
+        public ITaskItem[] AdditionalPackageFiles { get; set; }
+
+        public ITaskItem[] AdditionalElementsToRemoveFromPacking { get; set; } = Array.Empty<ITaskItem>();
```
javiercn, line 34: This is used by static web assets V1 since they don't have a build manifest. When we compute the assets to pack in a cross-targeting environment, we add them here to make sure we remove them as content items from the actual packing so that they don't appear on `content` nor `contentFiles`

Issue #24592, opened 2022-03-28
Fix unstable test build


marcpopMSFT:
I've seen a few times in test leg failures a build issue with a locked files. Below is one recent example. This is causing PR instability.

https://dev.azure.com/dnceng/public/_build/results?buildId=1682488&view=logs&j=fdda0353-544b-5e04-999c-0d26850a7bd6&t=a233f37a-3b31-5df5-e868-2cdcbfab3a02&s=ff05ad62-bb9a-53b6-ce9f-72f329a63e7c

/Users/runner/work/1/s/.dotnet/sdk/6.0.100/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(221,5): error : The process cannot access the file '/Users/runner/work/1/s/artifacts/bin/Microsoft.DotNet.PackageInstall.Tests/Release/TestAssetLocalNugetFeed/global.tool.console.demo.1.0.4.nupkg' because it is being used by another process. [/Users/runner/work/1/s/src/Tests/Microsoft.DotNet.PackageInstall.Tests/SampleGlobalTool/consoledemo.csproj]

##[error].dotnet/sdk/6.0.100/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(221,5): error : (NETCORE_ENGINEERING_TELEMETRY=Build) The process cannot access the file '/Users/runner/work/1/s/artifacts/bin/Microsoft.DotNet.PackageInstall.Tests/Release/TestAssetLocalNugetFeed/global.tool.console.demo.1.0.4.nupkg' because it is being used by another process.

marcpopMSFT:
A slightly different one:
https://dev.azure.com/dnceng/public/_build/results?buildId=1682639&view=logs&j=c766a980-3f0e-5da5-03a0-7235c55e2c80&t=a8b8a440-4d80-5066-5110-817d2848b6de
[...]

PR #25054, opened 2022-04-25
Reduce parallel copies of the test .zip file that were overlapping

marcpopMSFT:


marcpopMSFT:
This will help with the second example in #24592

joeloff (approved):

PR #25055, opened 2022-04-25
Try out pre-downloading the test packages on the helix machines

marcpopMSFT:


joeloff (approved):

marcpopMSFT:
last build was a while ago so rerunning just to make sure it still passes.

PR #25117, opened 2022-04-28
Use System.Environment to obtain user HOME dir

am11:
Using `Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)` to obtain user's home or profile directory is preferred over manually probing the environment variable.

Under linux base installation, e.g. container, if HOME is not set, `Environment.GetFolderPath(UserProfile)` will not give up right away. It fallback to POSIX `getpwuid(3)` API to obtain this path.

corehost was updated in https://github.com/dotnet/runtime/commit/cef245d28ae8d50f3bf0ad7adedc461cd45b085c (targeting .NET 7 release), which puts host on the same fallback plan as `Environment.GetFolderPath` and apphost/singlefilehost etc. do not complain if HOME is not set.

For SDK, before this change:

```sh
$ unset HOME
$ dotnet --help
The user's home directory could not be determined. Set the 'DOTNET_CLI_HOME' environment variable to specify the directory to use.
```

and with this change, it does not produce the error.

### DIFF
--- a/src/Common/CliFolderPathCalculatorCore.cs
+++ b/src/Common/CliFolderPathCalculatorCore.cs
@@ -13,8 +13,7 @@ namespace Microsoft.DotNet.Configurer
     {
         public const string DotnetHomeVariableName = "DOTNET_CLI_HOME";
         public const string DotnetProfileDirectoryName = ".dotnet";
-
-        public static string PlatformHomeVariableName =>
+        public static readonly string PlatformHomeVariableName =
             RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "USERPROFILE" : "HOME";
 
         public static string? GetDotnetUserProfileFolderPath()
@@ -28,17 +27,19 @@ namespace Microsoft.DotNet.Configurer
             return Path.Combine(homePath, DotnetProfileDirectoryName);
         }
 
-        public static string? GetDotnetHomePath(Func<string, string?>? getEnvironmentVariable = null)
+        public static string? GetDotnetHomePath()
         {
-            getEnvironmentVariable ??= key => Environment.GetEnvironmentVariable(key);
-
-            var home = getEnvironmentVariable(DotnetHomeVariableName);
+            var home = Environment.GetEnvironmentVariable(DotnetHomeVariableName);
             if (string.IsNullOrEmpty(home))
             {
-                home = getEnvironmentVariable(PlatformHomeVariableName);
+                home = Environment.GetEnvironmentVariable(PlatformHomeVariableName);
                 if (string.IsNullOrEmpty(home))
                 {
-                    return null;
+                    home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+                    if (string.IsNullOrEmpty(home))
+                    {
+                        return null;
+                    }
                 }
             }
 
--- a/src/Tests/Microsoft.NET.Sdk.Razor.Tool.Tests/ServerCommandTest.cs
+++ b/src/Tests/Microsoft.NET.Sdk.Razor.Tool.Tests/ServerCommandTest.cs
@@ -61,22 +61,13 @@ namespace Microsoft.NET.Sdk.Razor.Tool
         public void GetPidFilePath_ReturnsCorrectDefaultPath()
         {
             // Arrange
-            var expectedPath = Path.Combine("homeDir", ".dotnet", "pids", "build");
-            var server = GetServerCommand();
+            var expectedPath = Path.Combine(".dotnet", "pids", "build");
 
             // Act
-            var directoryPath = server.GetPidFilePath(getEnvironmentVariable: env =>
-            {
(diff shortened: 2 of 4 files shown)


am11:
cc @vitek-karas

am11:
@eerhardt, @vitek-karas, would appreciate your review. :)

With additional fallback, the implementation is more robust against missing `HOME` (or `UserProfile` for Windows), so I removed / adjusted the tests which were expecting the implementation to be sensitive to empty values for those environment variables.

src/Tests/dotnet.Tests/GivenThatDotNetRunsCommands.cs
```diff
@@ -40,21 +40,6 @@ public void UnresolvedPlatformReferencesFailAsExpected()
                      .And.HaveStdErrContaining(string.Format(LocalizableStrings.NoExecutableFoundMatchingCommand, "dotnet-crash"));
         }
 
-        [Theory]
-        [InlineData("")]
-        [InlineData(null)]
-        public void GivenAMissingHomeVariableItPrintsErrorMessage(string value)
```
vitek-karas, line 46: If it doesn't fail anymore, then please test the new behavior.

src/Common/CliFolderPathCalculatorCore.cs
```diff
@@ -28,17 +27,19 @@ static class CliFolderPathCalculatorCore
             return Path.Combine(homePath, DotnetProfileDirectoryName);
         }
 
-        public static string? GetDotnetHomePath(Func<string, string?>? getEnvironmentVariable = null)
```
vitek-karas, line 31: I think the owners of the code should comment on the fact that this removes the "mocking" of env. variables capability (I know that Environment.GetFolderPath would circumvent it anyway, but still). I don't know this code to tell if this is something we want, or just a one-off.

am11: We can keep it if necessary. I just noticed that in the entire repo, we have one place where we were passing `getEnvironmentVariable` and that place was also using the default value (on line 33). From that outlook, it seemed unused.

src/Tests/dotnet.Tests/GivenThatDotNetRunsCommands.cs
am11: Will do. Thank you.

src/Tests/Microsoft.NET.Sdk.Razor.Tool.Tests/ServerCommandTest.cs
```diff
@@ -61,49 +61,29 @@ public void WritePidFile_WorksAsExpected()
...
-        {
-            // Arrange
-            var server = GetServerCommand();
-
-            // Act
-            var directoryPath = server.GetPidFilePath(getEnvironmentVariable: env => null);
-
-            // Assert
-            Assert.Null(directoryPath);
+            Environment.SetEnvironmentVariable("DOTNET_BUILD_PIDFILE_DIRECTORY", "");
```
vitek-karas, line 86: Nit - could you please wrap the rest of the test in try/finally and reset the env. variable in the finally block... so that it failing doesn't potentially cause cascade failures.

vitek-karas (approved):
@captainsafia could you please find an owner who could take another look a this PR?

joeloff:
I'll bring up this change in the next SDK triage sessions.

marcpopMSFT:
Adding Jacques to review. @am11 this should target main rather than preview 4 as preview 4 is locked and we typically won't take commits directly to the preview branches.

am11:
I have switched the target branch to `main` and rebased. I had to squash my commits since intermediate commits were getting conflicts. :(

joeloff:
Sorry for the delay. Just one question, what happens with symbolic links - those all get resolved by the underlying native APIs? I don't think it's a concern, more just curiosity

Otherwise I think we're good to merge

joeloff (approved):

captainsafia (approved):

am11:
@joeloff, thanks. I have tested it on linux-x64, and `SpecialFolder.UserProfile` seems to resolve symlink'd HOME as expected.

Code:
```c#
using System;
Console.WriteLine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile));
```
Publish and test NativeAOT:
```sh
$ alias dotnet7=~/.dotnet7/dotnet
$ dotnet7 --version
7.0.100-preview.5.22261.8

$ dotnet7 publish -p:PublishAot=true --use-current-runtime
$ bin/Debug/net7.0/linux-x64/publish/naot1 

$ bin/Debug/net7.0/linux-x64/publish/naot1 
/home/am11

$ HOME=/tmp2 bin/Debug/net7.0/linux-x64/publish/naot1 
# {empty} because /tmp2 doesn't exist

# create symlink HOME -> /tmp2 and try again
$ sudo ln -s ~/ /tmp2
$ HOME=/tmp2 bin/Debug/net7.0/linux-x64/publish/naot1 
/tmp2
```

PR #25132, opened 2022-04-29
Update Microsoft.NET.Compatibility.Common.targets

### LINKED ISSUE #24943: PackageValidation with --no-build when project has ProjectReference causes NETSDK1085
### Describe the bug

I have two projects, A and B, where B depends on A using a `<ProjectReference Include="A.csproj"/>`.

Project B has `<EnablePackageValidation>true</EnablePackageValidation>`.

If I run `dotnet pack --no-build B.csproj`, I get the error:

> C:\Program Files\dotnet\sdk\6.0.200\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(136,5): error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked. [C:\Path\To\A.csproj]

It seems that package validation is trying to build the target's dependencies, which then conflicts with the `--no-build` passed to `dotnet pack`.

### To Reproduce

`A/A.csproj`:

```xml

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>

</Project>
```

`B/B.csproj`:

```xml

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <EnablePackageValidation>true</EnablePackageValidation>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="../A/A.csproj"/>
  </ItemGroup>

</Project>
```

Run:

```
dotnet build B
dotnet pack --no-build B
```

### Further technical details

`dotnet --info`:

```
.NET SDK (reflecting any global.json):
 Version:   6.0.200
 Commit:    4c30de7899

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.200\

Host (useful for support):
  Version: 6.0.2
  Commit:  839cdfb0ec

.NET SDKs installed:
  6.0.200 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.14 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.14 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/24943

Make sure that the `BuildProjectReferences` property  is set to false when the `NoBuild=true` property is set, to avoid the ResolveReferences target to build project references when it shouldn't.

Before this change, a non cross-targeting build avoided to use an MSBuild task to gather the project's references but as NuGet's Pack target itself already doesn't differentiate between if it's running as part of a cross targeting or a non-cross targeting build and always calls into an MSBuild task, this optimization isn't necessary in package validation. Instead, this change brings package validation in sync with how NuGet's Pack target invokes inner build projects to make sure that additional project evaluations are avoided.

As an alternative, an additional target could have been added that sets `BuildProjectReferences` to false [like GenAPI is doing](https://github.com/dotnet/arcade/blob/769455cb75153c025502d9800ea5f25afe15ea54/src/Microsoft.DotNet.GenAPI/build/Microsoft.DotNet.GenAPI.targets#L17-L22) but it feels wrong to me to diverge in behavior between NuGet and package validation as the latter already depends on the former in execution.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
@@ -3,7 +3,7 @@
   <UsingTask TaskName="Microsoft.DotNet.Compatibility.ValidatePackage" AssemblyFile="$(DotNetCompatibilityAssembly)" />
 
   <PropertyGroup>
-    <!--Add any custom targets that need to run before package validation to the following property-->
+    <!-- Add any custom targets that need to run before package validation to the following property. -->
     <RunPackageValidationDependsOn>_GetReferencePathFromInnerProjects;$(RunPackageValidationDependsOn)</RunPackageValidationDependsOn>
   </PropertyGroup>
 
@@ -45,32 +45,23 @@
       ReferencePaths="@(PackageValidationReferencePath)" />
   </Target>
 
-  <PropertyGroup>
-    <_GetReferencePathFromInnerProjectsDependsOn Condition="'$(IsCrossTargetingBuild)' != 'true'">_GetReferencePathForPackageValidation</_GetReferencePathFromInnerProjectsDependsOn>
-    <_GetReferencePathFromInnerProjectsDependsOn Condition="'$(IsCrossTargetingBuild)' == 'true'">_ComputeTargetFrameworkItems</_GetReferencePathFromInnerProjectsDependsOn>
-  </PropertyGroup>
-
   <Target Name="_GetReferencePathForPackageValidation"
           DependsOnTargets="ResolveReferences"
           Returns="@(_ReferencePathWithTargetFramework)">
     <ItemGroup>
-      <_ReferencePathWithTargetFramework Include="@(ReferencePath)"  TargetFramework="$(TargetFramework)" />
+      <_ReferencePathWithTargetFramework Include="@(ReferencePath)" TargetFramework="$(TargetFramework)" />
     </ItemGroup>
   </Target>
 
+  <!-- Depends on NuGet's _GetTargetFrameworksOutput target to calculate inner target frameworks. -->
   <Target Name="_GetReferencePathFromInnerProjects"
-          DependsOnTargets="$(_GetReferencePathFromInnerProjectsDependsOn)"
+          DependsOnTargets="_GetTargetFrameworksOutput"
           Condition="'$(RunPackageValidationWithoutReferences)' != 'true'">
-
-    <MSBuild Projects="@(_InnerBuildProjects)"
-             Condition="'$(IsCrossTargetingBuild)' == 'true'"
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="_GetReferencePathForPackageValidation"
-             Properties="BuildProjectReferences=false">
+             Properties="TargetFramework=%(_TargetFrameworks.Identity);
+                         BuildProjectReferences=false">
       <Output ItemName="PackageValidationReferencePath" TaskParameter="TargetOutputs" />
     </MSBuild>
-
-    <ItemGroup Condition="'$(IsCrossTargetingBuild)' != 'true'">
-      <PackageValidationReferencePath Include="@(_ReferencePathWithTargetFramework)" />
-    </ItemGroup>
   </Target>
 </Project>


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -45,32 +45,23 @@
...
     <ItemGroup>
-      <_ReferencePathWithTargetFramework Include="@(ReferencePath)"  TargetFramework="$(TargetFramework)" />
+      <_ReferencePathWithTargetFramework Include="@(ReferencePath)" TargetFramework="$(TargetFramework)" />
     </ItemGroup>
   </Target>
 
+  <!-- Depends on NuGet's _GetTargetFrameworksOutput target to calculate inner target frameworks. -->
   <Target Name="_GetReferencePathFromInnerProjects"
-          DependsOnTargets="$(_GetReferencePathFromInnerProjectsDependsOn)"
+          DependsOnTargets="_GetTargetFrameworksOutput"
```
ericstj, line 58: Previously we used `_ComputeTargetFrameworkItems` which, though private, has a lot of other dependencies in this repo.  Why did we need to change to `_GetTargetFrameworksOutput` and is the NuGet team OK with that?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
```diff
@@ -45,32 +45,23 @@
       ReferencePaths="@(PackageValidationReferencePath)" />
   </Target>
 
-  <PropertyGroup>
-    <_GetReferencePathFromInnerProjectsDependsOn Condition="'$(IsCrossTargetingBuild)' != 'true'">_GetReferencePathForPackageValidation</_GetReferencePathFromInnerProjectsDependsOn>
```
ericstj, line 49: By making this an `MSBuild` call below will that result in an extra evaluation for a non-cross-targeting project?

ViktorHofer: I provided the reasoning behind this change in the top post 🔝

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
ViktorHofer: Even though we depend on a NuGet internal here, we should be able to catch any change to it by adding a test that utilizes this code path. NuGet directly inserts into dotnet/sdk so that would be caught as part of the insertion. Of course we want to keep the noise low but I wouldn't expect this to change often.

> Why did we need to change to _GetTargetFrameworksOutput

As discussed offline, we intentionally want to walk the same code path as NuGet does during packaging to keep our logic in sync. The change that requires this target is the MSBuild task that now batches on the `TargetFrameworks` property.

> and is the NuGet team OK with that?

@dotnet/nuget-team are you ok with us relying on this private `_GetTargetFrameworksOutput` target?

joperezr (approved):
The changes look good to me as long as NuGet team is ok with us depending on this private target. Of course, ideally we would want to have them promote this to a non-private target so that we have a higher confidence of not getting randomly broken by this.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets
nkolev92: This functionality ships with every new SDK version right? There's a world where someone installs a higher version of the nuget.build.tasks.pack package that has tweaked that target and it might break you.

We haven't changed that target in 5 years, so I don't expect a lot of movement there either. 
The standard `"internal"` target disclosure applies, but it should be reasonably easy react to any changes.

joperezr: > This functionality ships with every new SDK version right?

Yes and no. This does ship inside the SDK, but it also ships as a standalone package that someone can reference in order to pick a different version of this functionality (similar to how the Roslyn compiler ships in the SDK but you can also opt to use a different version). I'm assuming that fact doesn't change much your stance given that as you point out, this target hasn't been changed in 5 years?

ViktorHofer: IMO we don't want to advertise being able to reference a new version of the Compatibility tooling via a Package Reference. Roslyn or the linker who ship these toolset packages don't advertise them either as they are only meant for internal consumption of our core stack repositories.

joperezr: I wasn't suggesting advertising them. My point was more that because we have that shipping vehicle, it does mean that we could run into additional cases of SDK and PackageValidation comming from two different builds/branches/releases, so I was wondering if that might aggravate @nkolev92 concern above since he specifically asked if the shipping vehicle for this was only the SDK.

ViktorHofer: What I find interesting is that the `Microsoft.DotNet.Compatibility` package ships publicly. I thought we wanted to treat it as a transport only package but it shipped/ships to nuget.org. We could change that if we would like to.

nkolev92: > , given that referencing the package and not using the feature as part of the SDK isn't a supported scenario (beyond experimentation).

Make sure you document that :)

PR #25141, opened 2022-04-29
Add back build retry

marcpopMSFT:
I turned on test retry but that won't retry non-helix tests. Since we still have a lot of helix tests, adding back build retry (which will run after test retry). This will hopefully increase our PR success rate until I can get more tests move to helix.


danmoseley (approved):
appears to match doc https://github.com/dotnet/arcade/blob/2a91e85b18e375ccb015f52ca3fe01e07601998e/Documentation/Projects/Build%20Analysis/BuildRetryOnboard.md and makes sense until tests can move out.

PR #25195, opened 2022-05-03
Update build.yml to target loc build at 6.0.4xx

marcpopMSFT:
This should flow to 6.0.4xx and require a loc request to switch branches on their side.


marcpopMSFT:
This is a change we'll want to flow to 4xx and will require creating a loc branch change request work item after I merge it.

joeloff (approved):

PR #25225, opened 2022-05-05
Make the ImplicitUsings feature minimally define each imported namespace

### LINKED ISSUE #24909: Make ImplicitUsings does not work out of the box with .NET Framework project
*Apologies if this is not the right repo! Reporting here because there are at least some issues in this repo referencing `<ImplicitUsings>`*

Whenever I create a new class library project, it defaults to setting `<ImplicitUsings>enable</ImplicitUsings>`. This is good. However, when I multi-target or re-target the project to hit .NET Framework, I get the following compiler error:

> The type or namespace name 'Http' does not exist in the namespace 'System.Net' (are you missing an assembly reference?)

This can be resolved by adding a framework reference, but it would be nice if it "just worked".

I'd like for this to "just work" out of the box. One easy way to do this would be to change the generated usings file (ProjectName.GlobalUsings.g.cs) to look like the following:

```cs
// <auto-generated/>
global using global::System;

// new addition
#if NETFRAMEWORK
namespace System.Net.Http { }
#endif
```

Potentially we could define all the referenced namespaces to ensure that the auto-generated code never gives an error regardless of what framework is targeted and what references are imported.

### Additional context

Example .csproj to reproduce the issue:
```

  <PropertyGroup>
    <TargetFrameworks>net472;netstandard2.0;net6.0</TargetFrameworks>
    <LangVersion>10</LangVersion>
  </PropertyGroup>

</Project>
```

madelson:
This avoids compilation errors based on target framework.

Fix #24909

Right now, there are several cases where ImplicitUsings doesn't work "out of the box" if you are not targeting the latest framework version available in the current SDK. With this approach, we ensure that the default set of implicit usings should not cause compilation errors because each namespace is at least minimally defined.

The biggest downside I see to this approach is that if someone customizes the ImplicitUsings list and they misspell the namespace, they would no longer get a compilation error alerting them to that fact.

An alternative solution to this would be carefully selecting the set of default namespaces based on the lowest-common-denominator set available in the referenced target frameworks.


marcpopMSFT:

baronfel:
I think we should close this - we've taken the approach of conditioning the Usings to appropriate TFMs instead, which is slightly more tricky but reduces the churn in the generated files (which users sometimes look at and are confused about).

PR #25258, opened 2022-05-06
Update the RID target from lower win to the latest one

v-wuzhai:


marcpopMSFT (approved):

marcpopMSFT:
@v-wuzhai Thanks for putting this together. I don't expect the windows rids to change as quickly in the future as the ubuntu and osx ones

PR #25273, opened 2022-05-06
Add support for opting out of ANSI coloring via NO_COLOR env. variable

### LINKED ISSUE #25264: AnsiConsole should support NO_COLOR
The AnsiConsole used by the CLI doesn't seem to support the  [NO_CONSOLE](https://no-color.org/) mechanism for opting out of shell coloration. We should do so. 

Longer term this is something that a shared IConsole implementation should handle, but that's still a while out, so it makes sense for us to make this small change in the interim.

Marusyk:
Or even put it into `Write` method?

Fix #25264


marcpopMSFT (approved):

marcpopMSFT:

PR #25301, opened 2022-05-09
Enable Aot, Trim and single file analyzers if PublishAot is set

tlakollo:
Enable Aot, Trim and single file analyzers if PublishAot is set
Add more tests and a warning being generated by single file

### DIFF
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
@@ -54,7 +54,7 @@ namespace Microsoft.NET.Publish.Tests
 
         [RequiresMSBuildVersionTheory("17.0.0.32901")]
         [InlineData(LatestTfm)]
-        public void Run_requires_analyzers_without_PublishAot(string targetFramework)
+        public void Requires_analyzers_produce_warnings_without_PublishAot_being_set(string targetFramework)
         {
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -32,60 +32,118 @@ public GivenThatWeWantToPublishAnAotApp(ITestOutputHelper log) : base(log)
 
         [RequiresMSBuildVersionTheory("17.0.0.32901")]
         [InlineData(LatestTfm)]
-        public void ILLink_aot_analyzer_warnings_are_produced(string targetFramework)
+        public void Trimming_warnings_not_produced_if_only_EnableAotAnalyzer_is_set(string targetFramework)
```
LakshanF, line 35: It will be great to have more comments on this - it looks like aot warnings are generated and it might be a little confusing for folks who don't know the distinction between aot and trim warnings ranges

LakshanF (approved):

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
tlakollo: Let me know if the new comments/test renaming clarifies the goal of the tests :D

sbomer (approved):
Thank you!

vitek-karas (approved):

PR #25337, opened 2022-05-10
Add a helper script for identifying the SDK release containing a SHA

baronfel:
While researching commit history for https://github.com/dotnet/docs/pull/29354 I created a script that will scrape the tag list and walk it in chronological order, stopping when it finds the first tag that contains the provided commit-ish reference.  

I think it's a difficult enough task that keeping this script would be useful.

### DIFF
--- a/scripts/find-earliest-tag.ps1
+++ b/scripts/find-earliest-tag.ps1
@@ -1,0 +1,53 @@
+<#
+.SYNOPSIS
+Used to walk the history of tags in a repo and find the first tag containing a given commit.
+#>
+param (
+    # the SHA whose parent we're looking for
+    [Parameter(Mandatory = $true, Position=0)][string] $targetSHA
+)
+
+function Check-Ref {
+    param (
+        [string] $childRef,
+        [string] $parentRef
+    )
+    Write-Host "Checking if $childRef is contained inside $parentRef" -ForegroundColor Yellow
+    git merge-base --is-ancestor $childRef $parentRef
+    return $?
+}
+
+function Get-Tags {
+    param (
+        [string] $remote
+    )
+    $tags=@{}
+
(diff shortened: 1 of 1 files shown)


joeloff (approved):
Sounds great. We'll probably need a .sh copy too.

KalleOlaviNiemitalo:
`git describe --contains` is similar.

scripts/find-earliest-tag.ps1
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+    return $tags.Keys.GetEnumerator()
+}
+
+$allTags=(Get-Tags "upstream")
+$found=$false
+foreach ($tag in $allTags) {
+    if (Check-Ref $targetSHA $tag) {
+        Write-Host "The earliest tag that is an ancestor of $targetSHA is $tag" -ForegroundColor Green
```
KalleOlaviNiemitalo, line 46: Don't you mean "descendant"? As in, the commit was made first, and the tag later.

baronfel: yeah, that's incorrect on my part

baronfel:
The pointer to `git describe` was a good one, @KalleOlaviNiemitalo - the invocation we want would be something like `git describe --tags --contains <target SHA> --exclude "v*-*"` to get similar behavior, _and_ it's way faster that walking each SHA.

KalleOlaviNiemitalo:
There's also `git tag --contains` if you don't want decorations, but then it would list all matching tags rather than just the earliest one.

Do you run your script manually or as part of some automation?

baronfel:
This is purely for manual fact finding. The motivation in this case was finding out which SDK releases started tracking certain telemetry data points that slipped through our documentation (which have now been documented :) ).

baronfel:
Closing because it's a one-liner!

Issue #24205, opened 2022-03-03
Support flowing workload dependencies through project and package dependencies


dsplaisted:
Workload dependencies should be able to flow through project and package dependencies.  IE, if a project depends on a workload, projects which reference the project may need to also depend on that workload.

For iOS and Android workloads, the TargetFramework is what lights up the workload, so in that case the dependencies automatically depend on the same workload because they need to use a compatible TargetFramework.  However, projects using .NET Maui can put shared logic in a project with a TargetFramework of `net6.0`, and set the `UseMauiEssentials` property to `true` to opt in to the Maui workload.  We currently don't have a good way of flowing that workload dependency to other projects.

Some possible solutions:

# Do nothing

Projects which consume NuGet packages which use NET Maui will have to explicitly set the `UseMauiEssentials` property.  This is a poor experience, especially because there won't be anything to help developers know they need to do this.

# Use `FrameworkReference` to drive explicit workload opt-in

Maui essentials does use a `FrameworkReference` to bring in the Maui library, and this will already flow through packages and (I think) project references.  We could add logic to the SDK (possibly coming from workload manifests) that recognizes that this `FrameworkReference` comes from a workload, and generates an error about how to opt in to the workload.

# Move `KnownFrameworkReference` for Maui essentials into .NET SDK

The `KnownFrameworkReference` for Maui essentials is in one of the Maui workloads.  We could move it to the .NET SDK itself, so you don't need to use the workload in order to consume the `FrameworkReference`.  However, this would make it harder to version consistently with the rest of the Maui workloads.  Also, while it would work for Maui Essentials, which only adds some APIs, it would not work for a workload which needs to include build logic (such as Maui.Core, I think).

# New NuGet feature to support flowing workloads

NuGet currently flows `FrameworkReference` items to consuming projects.  However, we can't use MSBuild items to enable workloads, we need to use properties (this is due to the multi-pass way MSBuild evaluation works, where items aren't evaluated until all project imports have been resolved).

NuGet could be updated so that if a project depends on a workload, the information about what property needs to be set would be included in the NuGet package, and NuGet restore would set that property in the .nuget.g.props file it generates.

/cc @dotnet/dotnet-cli @Redth @jonathanpeppers @eerhardt @davidortinau @mhutch

eerhardt:
[...]

Issue #23306, opened 2022-01-06
NETSDK1100 when building projects with multiple TargetFrameworks


Hendi48:
### Describe the bug
The docs state:
> To make your app portable across different platforms but still have access to OS-specific APIs, you can target multiple OS-specific TFMs and add platform guards around OS-specific API calls using #if preprocessor directives.

I did this and most of the projects in the solution are targeting `net6.0-windows;net6.0` if they have optional GUI components, and only `net6.0` otherwise.

Now, when I ask dotnet to specifically target `net6.0` when building in CI (so, Linux), it fails and complains about Windows regardless, even though it had no reason to even touch that. The same solution publishes flawlessly to Linux targets when building on Windows. Note that `Sdk` is **not** set to `Microsoft.NET.Sdk.WindowsDesktop`.

```
$ dotnet build TheSolution.sln -c Release -f net6.0
Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.
  Determining projects to restore...
/usr/share/dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(90,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/builds/foo/Project/Project.csproj]
```

### Expected behavior

When I specify `-f net6.0`, it assumes `TargetFramework = 'net6.0'` for all projects and acts as if `net6.0-windows` weren't there.

### To Reproduce
```xml
<Project Sdk="Microsoft.NET.Sdk">
	<PropertyGroup>
		<TargetFrameworks>net6.0-windows;net6.0</TargetFrameworks>
		<Platform Condition=" '$(Platform)' == '' ">x86</Platform>
		<OutputType>Exe</OutputType>
		<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
	</PropertyGroup>
	<PropertyGroup Condition="$(TargetFramework.EndsWith('windows'))">
		<UseWindowsForms>true</UseWindowsForms>
		<ImportWindowsDesktopTargets>true</ImportWindowsDesktopTargets>
		<DefineConstants>TARGET_WINDOWS</DefineConstants>
	</PropertyGroup>
	<ItemGroup Condition="!$(TargetFramework.EndsWith('windows'))">
		<Compile Remove="MainForm.cs" />
		<Compile Remove="MainForm.Designer.cs" />
[...]

PR #25358, opened 2022-05-11
Support suppressing transitive framework reference downloads, and other changes

dsplaisted:
- Adds support for a `DisableTransitiveFrameworkReferenceDownloads` property.  If set to true, targeting and runtime packs will not be downloaded unless there is a corresponding FrameworkReference in the project being built (instead of the default behavior, which is to download all of them because they might be needed for a transitive framework reference).  This should make testing internal builds easier, as we plan to put the base runtime packs in a workload, and setting this property will mean that the ASP.NET and WindowsDesktop runtime packs won't always need to be downloaded from special build-specific feeds.
- Adds a specific error message when the Microsoft.Maui.Essentials framework reference is not recognized: "NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference.  To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary)."  This was the [outcome](https://github.com/dotnet/sdk/issues/24205#issuecomment-1101668900) of our discussion on how to handle the Maui transitive framework references.  However, note that the Maui KnownFrameworkReferences are currently set to `Pack="false" PrivateAssets="All"` (see [here](https://github.com/dotnet/maui/blob/744966449e4bd6c9935442984069a71aff1586cc/src/Workload/Microsoft.Maui.Sdk/Sdk/Sdk.targets#L22-L23)), so these framework reference don't flow to referencing projects at all.
- Adds support for building Windows projects on other operating systems if an `EnableWindowsTargeting` property is set to true.  This should help address dotnet/sdk#3592 and dotnet/sdk#23306.  The reason this was disabled in the first place was because we download all targeting packs (and runtime packs for self-contained builds) for the current target framework whether they are needed or not, because they might be brought in by a transitive framework reference.  We didn't want to ship the Windows targeting packs with the non-Windows SDK builds, but we also didn't want a vanilla Console or ASP.NET app to automatically download these targeting and runtime packs the first time you build.  The new property enables them to only be downloaded if you opt in.
- Supports RID-specific "RuntimePackAlwaysCopyLocal" runtime packs.  This should help support framework-dependent Maui apps targeting Windows: dotnet/maui#7170

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -841,7 +841,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1184: "}</comment>
   </data>
   <data name="UnknownFrameworkReference_MauiEssentials" xml:space="preserve">
-    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
+    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference, but doesn't declare that dependency explicitly. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
     <comment>{StrBegin="NETSDK1185: "}</comment>
   </data>
 </root>
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -833,15 +833,15 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1182: "}</comment>
   </data>
   <data name="TargetingPackNotRestored_TransitiveDisabled" xml:space="preserve">
-    <value>NETSDK1183: The Targeting Pack for FrameworkReference '{0}' was not available.  This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
+    <value>NETSDK1183: The Targeting Pack for FrameworkReference '{0}' was not available. This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
     <comment>{StrBegin="NETSDK1183: "}</comment>
   </data>
   <data name="RuntimePackNotRestored_TransitiveDisabled" xml:space="preserve">
-    <value>NETSDK1184: The Runtime Pack for FrameworkReference '{0}' was not available.  This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
+    <value>NETSDK1184: The Runtime Pack for FrameworkReference '{0}' was not available. This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
     <comment>{StrBegin="NETSDK1184: "}</comment>
   </data>
   <data name="UnknownFrameworkReference_MauiEssentials" xml:space="preserve">
-    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference.  To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
+    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
     <comment>{StrBegin="NETSDK1185: "}</comment>
   </data>
 </root>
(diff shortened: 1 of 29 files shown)


dsplaisted:
@baronfel To review the new / updated error messages

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -832,4 +832,16 @@ You may need to build the project on another operating system or architecture, o
...
+  <data name="TargetingPackNotRestored_TransitiveDisabled" xml:space="preserve">
+    <value>NETSDK1183: The Targeting Pack for FrameworkReference '{0}' was not available.  This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
+    <comment>{StrBegin="NETSDK1183: "}</comment>
+  </data>
+  <data name="RuntimePackNotRestored_TransitiveDisabled" xml:space="preserve">
+    <value>NETSDK1184: The Runtime Pack for FrameworkReference '{0}' was not available.  This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
+    <comment>{StrBegin="NETSDK1184: "}</comment>
+  </data>
+  <data name="UnknownFrameworkReference_MauiEssentials" xml:space="preserve">
+    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference.  To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
```
joeloff, line 844: Looks like there's a double space after "reference.  To build..."

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenThatWeHaveErrorCodes.cs
```diff
@@ -37,6 +37,7 @@ public class GivenThatWeHaveErrorCodes
             1066,
             1101,
             1108,
+            1180,
```
joeloff, line 40: Is the a corresponding code change for the deletion?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
```diff
@@ -78,14 +80,19 @@ protected override void ExecuteCore()
...
 
                 if (s_allowCacheLookup)
                 {
                     BuildEngine4?.RegisterTaskObject(cacheKey, results, RegisteredTaskObjectLifetime.AppDomain, allowEarlyCollection: true);
                 }
             }
 
+            foreach (var error in results.Errors)
+            {
+                Log.LogError(error);
```
joeloff, line 93: Is it worth continuing once errors have been logged?

joeloff (approved):
LGTM - no concerns other than a some questions I had

gkulin (approved):

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenThatWeHaveErrorCodes.cs
dsplaisted: No, I think that none of these unit tests are running in CI, so this happened earlier and I just noticed when I ran the test manually.

src/Tasks/Microsoft.NET.Build.Tasks/ResolveTargetingPackAssets.cs
dsplaisted: I don't think it hurts anything and it's the way it behaved previously.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -832,4 +832,16 @@ You may need to build the project on another operating system or architecture, o
...
+  <data name="TargetingPackNotRestored_TransitiveDisabled" xml:space="preserve">
+    <value>NETSDK1183: The Targeting Pack for FrameworkReference '{0}' was not available. This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
+    <comment>{StrBegin="NETSDK1183: "}</comment>
+  </data>
+  <data name="RuntimePackNotRestored_TransitiveDisabled" xml:space="preserve">
+    <value>NETSDK1184: The Runtime Pack for FrameworkReference '{0}' was not available. This may be because DisableTransitiveFrameworkReferenceDownloads was set to true.</value>
+    <comment>{StrBegin="NETSDK1184: "}</comment>
+  </data>
+  <data name="UnknownFrameworkReference_MauiEssentials" xml:space="preserve">
+    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
```
baronfel, line 844: This message doesn't say why there's a problem, and the other ones do (which is a good thing). I think we can unify the tone by adding a brief section on why this is a problem. 

```suggestion
    <value>NETSDK1185: This project depends on Maui Essentials through a project or NuGet package reference, but doesn't declare that dependency explicitly. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
```

baronfel (approved):

Issue #24269, opened 2022-03-08
With .NET SDK 6.0.300-preview.22154.4, warning NETSDK1179 still shows when running "dotnet publish -r win-x64 --self-contained" in CLI


WenJunJi01:
**Environment:** .NET SDK 6.0.300-preview.22154.4 (runtime-6.0.2)

**Repro steps:**
1. Create an empty solution in CLI
2. Add a new .NET 6.0 console application with reference to a .NET 6.0 class library
3. Run "dotnet publish -r win10-x64“ 
4. The app can publish successfully and a warning occurs: warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.
5. Run "dotnet publish -r win10-x64 --self-contained"

**Expected Result:**
Warning NETSDK1179 should not display.
![image](https://user-images.githubusercontent.com/87680983/157203389-27a29399-e179-4427-b728-5324107002b3.png)

**Actual Result:**
warning NETSDK1179 still shows after adding '--self-contained' or '--no-self-contained'.
![image](https://user-images.githubusercontent.com/87680983/157202548-dd7b80fc-31fa-4f10-ae39-1c60f4d97e25.png)

**Dotnet info:**
.NET SDK (reflecting any global.json):
Version: 6.0.300-preview.22151.9
Commit: c5b0a4598d

Runtime Environment:
OS Name: Windows
OS Version: 10.0.22000
OS Platform: Windows
RID: win10-x64
Base Path: C:\Program Files\dotnet\sdk\6.0.300-preview.22151.9\

Host (useful for support):
Version: 6.0.2
Commit: 839cdfb0ec

.NET SDKs installed:
6.0.300-preview.22151.9 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
Microsoft.AspNetCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
Microsoft.NETCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.WindowsDesktop.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

To install additional .NET runtimes or SDKs:
https://aka.ms/dotnet-download

waynebrantley:
[...]

PR #25367, opened 2022-05-11
Always detect if SelfContained was specified by the user

baronfel:
Fixes https://github.com/dotnet/sdk/issues/24269 by moving the _SelfContainedWasSpecified property detection out of a conditioned PropertyGroup and into an always-evaluated PropertyGroup.

## Description

Users cannot correctly specify that they want a runtime-specific, self-contained application without warnings when building applications that reference other libraries. This is because in this scenario a sentinel property isn't set, and later that sentinel property is used to drive a build diagnostic. This appears to be a gap that we missed when tackling a bugfix that a few months ago: https://github.com/dotnet/sdk/pull/23999. For that fix, we moved to conditionally setting this flag via the CLI to inferring if the property was set through any means (which is good), but not in a way that was resilient to all kinds of property evaluations.  Now we unconditionally do this detection.

## Customer Impact

This particular kind of warning is often treated as an error by users in build scenarios, and so their builds fail. There's no user-accessible workaround that doesn't also surface undesirable side effects.

## Testing

Additional unit testing was added around the erroring reference pattern to safeguard against future regresssions.

## Risk

This is safe for two reasons: 

a) `_SelfContainedWasSpecified` is a private property that we don't have any contracts for, and
b) it's only used in one place - the RuntimeIdentifierInference targets where it is consumed in order to log an error message.

## Regression

Yes, it started appearing with 6.0.300 previews and slipped through the cracks.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -109,8 +109,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     Default SelfContained based on the RuntimeIdentifier, so projects don't have to explicitly set SelfContained.
     This avoids a breaking change from 1.0 behavior.
     -->
-  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
+  <PropertyGroup>
+    <!-- Detecting property presence is not harmful and can be done in an unconditioned way -->
     <_SelfContainedWasSpecified Condition="'$(SelfContained)' != ''">true</_SelfContainedWasSpecified>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
     <SelfContained Condition="'$(SelfContained)' == '' and '$(RuntimeIdentifier)' != ''">true</SelfContained>
     <SelfContained Condition="'$(SelfContained)' == ''">false</SelfContained>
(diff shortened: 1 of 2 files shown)


baronfel:
Needs a test to verify the behavior going forward, the existing in-memory project APIs should support this.

baronfel:
@joeloff / @gkulin would one or both of you mind reviewing the change and test?

gkulin (approved):
You should probably wait for Jacques to review it too, but this looks good to me

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -109,8 +109,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     Default SelfContained based on the RuntimeIdentifier, so projects don't have to explicitly set SelfContained.
     This avoids a breaking change from 1.0 behavior.
     -->
-  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
+  <PropertyGroup>
+    <!-- Detecting property presence is not harmful and can be done in an unconditioned way -->
     <_SelfContainedWasSpecified Condition="'$(SelfContained)' != ''">true</_SelfContainedWasSpecified>
```
joeloff, line 114: What happens if you do `SelfContained=false`?

baronfel: That's perfectly fine - we just want to enforce that if a user passed a runtime identifier to the publish command that they use one of --self-contained/--no-self-contained to be explicit (or pass the property directly, of course).  The code that actually cares about the value of SelfContained does the explicit true/false checks.

joeloff (approved):

PR #25395, opened 2022-05-13
Enable signature checks in the package downloader

### LINKED ISSUE #25377: Fix package signature verification when installing workloads.
## Description
The new `--skip-sign-check` flag for workloads is bypassing some package validation. MSI signature verification is fine. The information is not being passed to the downloader. 

## Customer Impact

When performing a workload command, NuGet package signatures are not verified. This only impacts Windows. MSI signature verification is not impacted.

## Regression

Yes

## Testing

Manual. It requires dotnet.dll to be signed with proper certificate. The test plan for workloads will be updated to include a scenario that covers signature verification for workloads.

## Risk

The fix is relatively low risk as an additional parameter just need to be passed through from the various commands. It will require some refactoring to ensure that duplicated code in various workload commands are moved to a common base class to handle package downloads/verification in a consistent fashion.

joeloff:
Fixes #25377 

The simple fix is to just pass `VerifySignatures` whenever a workload command instantiates `NuGetPackageDownloader`. That still leaves the risk of adding a new command and not passing it through in the future. Instead, all of that code was moved to the `WorkloadCommandBase` so that commands don't duplicate this.

In doing so I uncovered a number of other inconsistencies around the package download folders that vary significantly between commands and because the downloader depends on these values, it also needed to be refactored.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -170,7 +170,7 @@
     <FluentAssertionsJsonVersion>4.19.0</FluentAssertionsJsonVersion>
     <MicrosoftDotNetXUnitExtensionsVersion>6.0.0-beta.22212.5</MicrosoftDotNetXUnitExtensionsVersion>
     <MoqPackageVersion>4.8.2</MoqPackageVersion>
-    <MicrosoftDotNetInstallerWindowsSecurityTestDataPackageVersion>6.0.0-beta.21376.2</MicrosoftDotNetInstallerWindowsSecurityTestDataPackageVersion>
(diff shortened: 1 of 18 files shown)


joeloff:
Some additional tests are pending until https://github.com/dotnet/runtime-assets/pull/242 makes its way through the official CI build and we have a new test asset package available. Most of the testing remains manual because it requires a properly signed copy of dotnet.

marcpopMSFT (approved):

PR #25435, opened 2022-05-14
Set the Framework and Runtime options as required for 'dotnet store'

### LINKED ISSUE #23682: `dotnet store` doesn't set the Framework and Runtime options as required
### Describe the bug

Per https://github.com/dotnet/docs/issues/27775, as well as the [rendered docs](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-store), the `dotnet store` command requires the Framework and Runtime options in order to succeed, but this is not enforced by the option configuration for the command. 

We should change the configuration of the parser [here](https://github.com/dotnet/sdk/blob/47c3dd45c70391d98068087ce73df6f8c9d419e2/src/Cli/dotnet/commands/dotnet-store/StoreCommandParser.cs#L88-L91) to mark these two are required, which will need a bit of work. Right now everything that uses the 'shared' definitions of Framework/Runtime, but we need to ability to share _most_ of that definition with tiny changes, like marking `IsRequired = true` in this command.

Marusyk:
Fixes #23682

Please review

### DIFF
--- a/src/Cli/dotnet/OptionForwardingExtensions.cs
+++ b/src/Cli/dotnet/OptionForwardingExtensions.cs
@@ -51,6 +51,12 @@ namespace Microsoft.DotNet.Cli
             return option;
         }
 
+        public static Option<T> Required<T>(this Option<T> option)
+        {
+            option.IsRequired = true;
+            return option;
+        }
+
         public static Option<T> WithHelpDescription<T>(this Option<T> option, Command command, string helpText)
         {
             if (Parser.HelpDescriptionCustomizations.ContainsKey(option))
--- a/src/Cli/dotnet/commands/dotnet-store/StoreCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-store/StoreCommandParser.cs
@@ -85,8 +85,8 @@ namespace Microsoft.DotNet.Cli
             command.AddOption(WorkingDirOption);
             command.AddOption(SkipOptimizationOption);
             command.AddOption(SkipSymbolsOption);
-            command.AddOption(CommonOptions.FrameworkOption(LocalizableStrings.FrameworkOptionDescription));
-            command.AddOption(CommonOptions.RuntimeOption.WithHelpDescription(command, LocalizableStrings.RuntimeOptionDescription));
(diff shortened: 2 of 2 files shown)


baronfel:
Hey @Marusyk, thanks for this! 

This change broke a lot of tests because of the mutability of System.Commandline - the shared Runtime option in `CommonOptions.cs` are mutated with your new method call.  I think to do this in a way that doesn't impact other commands, we'll need to change the definition of the Runtime options in `CommonOptions.cs` to be members that return a new instance of an `Option`, instead of one-time members. This should look a lot like the way the Framework option is currently defined, and for many of the same reasons. This will have effects on all the callsites that use Framework/Runtime currently. What I'd expect you to need to do would be:

* Make the changes to CommonOptions.cs for the Runtime option to a) take an explicit description and b) return a new instance of an Option for each call
* In all of the command-specific parsers that use the Runtime option, adjust callsites to the new API

baronfel:
Hey @Marusyk, thanks for this! 

This change broke a lot of tests because of the mutability of System.Commandline - the shared Runtime option in `CommonOptions.cs` are mutated with your new method call.  I think to do this in a way that doesn't impact other commands, we'll need to change the definition of the Runtime options in `CommonOptions.cs` to be members that return a new instance of an `Option`, instead of one-time members. This should look a lot like the way the Framework option is currently defined, and for many of the same reasons. This will have effects on all the callsites that use Framework/Runtime currently. What I'd expect you to need to do would be:

* Make the changes to CommonOptions.cs for the Runtime option to a) take an explicit description and b) return a new instance of an Option for each call
* In all of the command-specific parsers that use the Runtime option, adjust callsites to the new API

Marusyk:
Hi @baronfel 
thanks for your feedback.
could you please suggest, you mean smth like this
```
public static Option<string> RuntimeOption(string description) =>
    new Option<string>(
    new string[] { "-r", "--runtime" },
    description);
```
Is there a way to apply `ForwardAsMany(RuntimeArgFunc)` and `AddCompletions` then?

another problem is that the `RuntimeOption` uses here
image

Should the `description` be optional (use `LocalizableStrings.RuntimeOptionDescription` as a default value)?

Marusyk:
any updates?

PR #25449, opened 2022-05-16
Fix EffectiveAnalysisLevel referring to unset prop

### LINKED ISSUE #69392: "[MSBuild]::VersionGreaterThanOrEquals(preview, 5.0)" failures on local build/test/VS
With a recent VS dogfood build (I don't know if that matters) and an up-to-date and cleanly built clone of main, when I then try to cd into a src or test directory and build, I'm inundated with errors of the form:
"The expression "[MSBuild]::VersionGreaterThanOrEquals(preview, 5.0)" cannot be evaluated. Version string was not in a correct format."

I've temporarily and locally worked around it by adding `<AnalysisLevel>6.0</AnalysisLevel>` in my Directory.Build.targets file.

ViktorHofer:
Fixes https://github.com/dotnet/runtime/issues/69392

Regressed as part of https://github.com/dotnet/sdk/pull/24826.

@baronfel apparently this setting isn't protected by tests, so I filed https://github.com/dotnet/sdk/issues/25450 for that.

cc @stephentoub


baronfel (approved):
I'll make a follow up issue to guard this with a test after this merges.

Thank you for the fix!

PR #25454, opened 2022-05-16
add tests to cover the resolution of AnalysisLevel known-strings

### LINKED ISSUE #25450: AnalysisLevel preview setting is missing tests
https://github.com/dotnet/sdk/pull/24826 recently broke project evaluations when the `AnalysisLevel` property is set to `preview`. Apparently the is a test gap that should be fixed so that this setting doesn't regress again.

cc @baronfel

baronfel:
Fixes #25450 by

* adding a test that covers the promotion of the `preview` term to a target TFM
* adding tests that verify that for all of the known AnalysisLevel terms, we resolve those terms to an actual version number before moving on

Potential enhancements I'd like some feedback on:

* [x] right now I'm hard-coding the 3 known string AnalysisLevel values. Is there an enum of these available somewhere to reference? Guessing not, since this is entirely an MSBuild-level construct. Resolution: just hard-coded them for now.
* [x] the 'preview floats to the next TFM' test currently hard-codes 8.0. I would love to be able to get the TFM we're built for at runtime somehow. anyone know how to do this? maybe a custom AssemblyAttribute? Resolution: Used the ToolSetInfo to ensure that when we bump our concept of 'next' and 'current' TFMs the sentinel test will break, so we will have to update this implementation.
* [ ] currently it's hard to tell the user the valid values of the AnalysisLevel property via MSBuild warnings or errors, because all of this calculation happens as part of property evaluation. Other .targets files do this property evaluation inside of a target to allow for creating NETSDK errors. Is it worth trying to move in that direction? What I'm trying to fix is unknown AnalysisLevels getting passed into a TFM version evaluation because of this condition:
  ```xml
      <EffectiveAnalysisLevel Condition="'$(EffectiveAnalysisLevel)' == '' And
                                       '$(AnalysisLevel)' != ''">$(AnalysisLevel)</EffectiveAnalysisLevel>
    ```
    
    Resolution: too hard and error prone to do - will discuss with SDK/MSBuild team for a later date

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
@@ -17,19 +17,19 @@ namespace Microsoft.NET.Build.Tests
     public class GivenThatWeWantToFloatWarningLevels : SdkTest
     {
         private const string targetFrameworkNet6 = "net6.0";
-        private const string targetFrameworkNet7 = "net7.0";
         private const string targetFrameworkNetFramework472 = "net472";
 
         public GivenThatWeWantToFloatWarningLevels(ITestOutputHelper log) : base(log)
         {
         }
 
-        [InlineData(targetFrameworkNet6, 6)]
-        [InlineData(targetFrameworkNet7, 7)]
-        [InlineData(targetFrameworkNetFramework472, 4)]
+        [InlineData(targetFrameworkNet6, "6")]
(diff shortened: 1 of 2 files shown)


jmarolf (approved):

jmarolf:
>  right now I'm hard-coding the 3 known string AnalysisLevel values. Is there an enum of these available somewhere to reference? Guessing not, since this is entirely an MSBuild-level construct.

There is no better way today as far as I know. We could have an enum and then generate the MSBuild definitions from there.

jmarolf:
>  Other .targets files do this property evaluation inside of a target to allow for creating NETSDK errors. Is it worth trying to move in that direction?

I think this is a good idea. If the developer provides an invalid value we should error.

baronfel:
Merging since these are test additions only.

PR #25490, opened 2022-05-17
document long paths-related error message when building on windows

smasher164:
Even though this error could technically show up on any repo, we still mention it here, since newcomers might hit this when building it for the first time.


joperezr (approved):
Thanks for adding this @smasher164 LGTM

joperezr:
@dsplaisted I suppose you guys are fine with this doc change?

PR #25494, opened 2022-05-18
Remove dependency on Assembly.GetCallingAssembly

steveharter:
This PR was used to prep expected changes based on a new breaking change caused by the expected behavior of the JIT inlining a target method due to the new support for IL-Emit'd invokers where IL is used to do the invoke and not native\interpreted logic. This caused `Assembly.GetCallingAssembly()` to not return the expected assembly, which caused SDK integration tests to fail.

In general, using `GetCallingAssembly()` is not recommended in general since it doesn't work on all platforms\environments and it is fragile essentially requiring `NoInlining` attribute.

However, there is now a work-around in place in https://github.com/dotnet/runtime/pull/69575 to avoid that breaking change plus plans to remove that work-around in favor of a better approach to using `calli` that does not re-introduce the breaking change.

Closing the PR since it is no longer required, although this PR still has merit as mentioned above due to using the fragile `GetCallingAssembly()`; i.e. if the tests were invoked without reflection they would fail today.

### DIFF
--- a/src/Tests/Microsoft.NET.TestFramework/AspNetSdkTest.cs
+++ b/src/Tests/Microsoft.NET.TestFramework/AspNetSdkTest.cs
@@ -18,10 +18,9 @@ namespace Microsoft.NET.TestFramework
     {
         public readonly string DefaultTfm;
 
-        protected AspNetSdkTest(ITestOutputHelper log) : base(log)
+        protected AspNetSdkTest(ITestOutputHelper log, Assembly testAssembly) : base(log)
         {
-            var assembly = Assembly.GetCallingAssembly();
-            var testAssemblyMetadata = assembly.GetCustomAttributes<AssemblyMetadataAttribute>();
+            var testAssemblyMetadata = testAssembly.GetCustomAttributes<AssemblyMetadataAttribute>();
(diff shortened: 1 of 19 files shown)


src/Tests/Microsoft.NET.TestFramework/AspNetSdkTest.cs
```diff
@@ -18,10 +18,9 @@ public abstract class AspNetSdkTest : SdkTest
     {
         public readonly string DefaultTfm;
 
-        protected AspNetSdkTest(ITestOutputHelper log) : base(log)
+        protected AspNetSdkTest(ITestOutputHelper log, Assembly testAssembly) : base(log)
```
ericstj, line 21: Can we use GetType() here instead?  We do that in some other types in the framework like [MemoryStream](https://github.com/dotnet/runtime/blob/6ca8c9bc0c4a5fc1082c690b6768ab3be8761b11/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L352).

steveharter: Since this class is not concrete, GetType() wouldn't work here.

PR #25541, opened 2022-05-19
Remove dotnet-format option parsing and have them pass-through.

JoeRobich:
This resolves several issues where the dotnet CLI is parsing arguments differently than dotnet-format.


baronfel:
While this will address the out-of-sync issue reported in dotnet/format#1479, you'll lose the built-in tab-completion support that having the parser inside the SDK CLI provides.  You can still get there if users have configured the `dotnet-suggest` tool, but the rate of users that have done that is probably quite low (based on similar reports of users with .NET CLI tab-completion enabled), and it's one more hurdle.  We do hope to centralize the completion story longer term, so eventually you would probably start lighting up support again as we figure that out.

An alternative - if you export a System.CommandLine command then we could simply mount that command in the SDK, like we do for the template engine (which provides the entire `dotnet new` command and infrastructure). That might have complications with your other desire to allow for dotnet-tool-based usage, though.

baronfel:
If you do move forward with removing the native parser, you'll also want to add a switch case to [this Help writer method](https://github.com/dotnet/sdk/pull/25541/files#diff-75f24d684a51fe02f83d64c70e9cd87276f55132b61facd9fd41bb0b5d1da847L209-L248) to delegate to your help command, so that users get useful help output.

jmarolf (approved):

PR #25542, opened 2022-05-19
add rules to make sure enums match

### LINKED ISSUE #23902: [API Compat] Add rule to make sure that enum underlying type matches
This rule should make sure that the underlying type of an enum is the same on both sides of the comparison.

### LINKED ISSUE #23903: [API Compat] add rule to make sure enum values match
We already cover that the members of an enum must exist via the [MembersMustExist rule](https://github.com/dotnet/sdk/blob/921dbcaa04265f86b4feefdffced9cfffc30378e/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/MembersMustExist.cs) however we must make sure that the constant values are the same on both sides as it is a breaking change to change the underlying value of an enum member.

This rule needs that at least we are running with a reference to the core library so that the constant type and constant value is retrieved by Roslyn.

smasher164:
Fixes #23903 
Fixes #23902


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=25542) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=25542) before we can accept your contribution.<br/><br/>:x: smasher164 [sign now](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=25542)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/sdk?pullRequest=25542) it.</sub>

src/Compatibility/compatibility.slnf
```diff
@@ -5,8 +5,8 @@
       "src\\Cli\\Microsoft.DotNet.Cli.Utils\\Microsoft.DotNet.Cli.Utils.csproj",
       "src\\Cli\\Microsoft.DotNet.InternalAbstractions\\Microsoft.DotNet.InternalAbstractions.csproj",
       "src\\Compatibility\\Microsoft.DotNet.ApiCompatibility\\Microsoft.DotNet.ApiCompatibility.csproj",
-      "src\\Compatibility\\Microsoft.DotNet.PackageValidation\\Microsoft.DotNet.PackageValidation.csproj",
       "src\\Compatibility\\Microsoft.DotNet.Compatibility\\Microsoft.DotNet.Compatibility.csproj",
```
joperezr, line 8: Did you manually make this change? We should probably undo it if so.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticIds.cs
```diff
@@ -17,6 +17,7 @@ public static class DiagnosticIds
         public const string CannotRemoveBaseType = "CP0007";
         public const string CannotRemoveBaseInterface = "CP0008";
         public const string CannotSealType = "CP0009";
+        public const string EnumValuesMustMatch = "CP0010";
```
joperezr, line 20: You can do this after the PR goes in, but eventually you'll have to add documentation for this diagnostic as well as how to address it. This is where it should go: [https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids)

src/Compatibility/compatibility.slnf
smasher164: VS asked me if I wanted to do that, and I blindly accepted. Undoing.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticIds.cs
smasher164: I'm guessing these are in the dotnet/docs repo?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -207,4 +207,7 @@
   <data name="TypeIsEffectivelySealed" xml:space="preserve">
     <value>Type '{0}' is sealed because it has no visible constructor on {1} but it does on {2}</value>
   </data>
-</root>
+  <data name="EnumValuesMustMatch" xml:space="preserve">
+    <value>Field '{0}.{1}' changed from '{2}' to '{3}'.</value>
```
joperezr, line 211:
```suggestion
    <value>Field '{0}.{1}' changed its value from '{2}' to '{3}'.</value>
```
Perhaps we can be more specific on the error? I also see that we kind of have different style of error messages across the diagnostics, some seem to state a fact like "Type bla does not inherit from bla, but it did before", while others use a more explisive description like: "Cannot add interface member bla because it did not exist before". I wonder if we should try to unify them.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiCompatibility.Abstractions;
+
+namespace Microsoft.DotNet.ApiCompatibility.Rules
+{
+    public class EnumValuesMustMatch : Rule
```
joperezr, line 14: I know perhaps old rules didn't do this, but it would be good if we add tripple slash comments on all of these rules with a description and remarks on what each is checking.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        public override void Initialize(RuleRunnerContext context)
+        {
+            context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
+        }
```
joperezr, line 19:
```suggestion
        public override void Initialize(RuleRunnerContext context) => context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
```
NIT: One statement methods can be simplified.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private static bool isEnum(ITypeSymbol type)
+        {
+            return type != null && type.IsType && type.TypeKind == TypeKind.Enum;
+        }
+        private void RunOnTypeSymbol(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
```
joperezr, line 25:
```suggestion
        private static bool isEnum(ITypeSymbol type)
            => type != null && 
                  type.IsType && 
                  type.TypeKind == TypeKind.Enum;

        private void RunOnTypeSymbol(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
```
Same here. Do we actually have to check for `type.IsType`? Is there ever a scenario where ITypeSymbol won't be a type?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
smasher164: Hmm yeah, I was trying to find some commonality across the other messages, but they were all different. I can file an issue about this, so that I can offer my suggestions on how to unify them.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            if (!isEnum(left) || !isEnum(right))
```
joperezr, line 27: Why do you have to check if both are Enum? I would expect that it would never be the case where the Mapper would give you an Enum on the left but not an enum on the right. Perhaps what you want to check here is only if one of them is an enum, and that should be enough.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return;
```
joperezr, line 29: NIT: also add comments here to say that we just want to run this rule for Enums

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+            var leftMembers = left.GetMembers().Where(a => a.Kind == SymbolKind.Field).Select(a => ((IFieldSymbol)a)).ToDictionary(a => a.Name);
```
joperezr, line 31: NIT: we only use `var` when the type can be easily inferred from the right side of the assignment like `var regex = new Regex()`, but in all of this cases please use explicit types instead for readability

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
smasher164: I think you're right. I might have gotten confused because I saw that it inherits `INamespaceOrTypeSymbol`.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,44 @@
...
+            var rightMembers = right.GetMembers().Where(a => a.Kind == SymbolKind.Field).Select(a => ((IFieldSymbol)a)).ToDictionary(a => a.Name);
+            foreach (var entry in leftMembers)
```
joperezr, line 33:
```suggestion
            var rightMembers = right.GetMembers().Where(a => a.Kind == SymbolKind.Field).Select(a => ((IFieldSymbol)a)).ToDictionary(a => a.Name);

            foreach (var entry in leftMembers)
```
NIT: we usually leave empty lines before code blocks like foreach's and if's

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
smasher164: I believe I saw this method being run where `left` was `null`. I'll double-check to make sure that `right` was an `Enum` in that circumstance.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
smasher164: Filed https://github.com/dotnet/sdk/issues/25565.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
smasher164: Filed https://github.com/dotnet/sdk/issues/25565.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/EnumValuesMustMatchTests.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
ViktorHofer, line 3:
```suggestion
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
ViktorHofer, line 3:
```suggestion
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// This class implements a rule to check that the constant values for an enum's fields don't change.
+    /// </summary>
+    public class EnumValuesMustMatch : Rule
+    {
+        public override void Initialize(RuleRunnerContext context) => context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
+
+        private void RunOnTypeSymbol(ITypeSymbol left, ITypeSymbol right, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            // Ensure that this rule only runs on enums.
+            if (left is null || left.TypeKind != TypeKind.Enum)
```
bartonjs, line 24: Seems like right being null or not an enum should also be considered; just to prevent the rule from crashing later.

joperezr: right, that was my bad, just checking if it is null should be enough, I don't think the mapper will map them if a previously-enum is now a class, but it should be good to check.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/EnumValuesMustMatchTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        public static void EnumValuesNotMatchingReported()
+        {
+            string leftSyntax = @"
+namespace CompatTests
+{
+  public enum First {
+    A = 0,
+    B = 1,
+    C = 2,
+    D = 3,
```
ViktorHofer, line 28: Are there other scenarios that would make sense to test for, i.e. a different order of enum members?

bartonjs: The declared order doesn't matter, except for when they're using automatic numbering. (And then the value would change, which this rule is already looking for)

smasher164:
I added a test for the backing store changing as well.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumValuesMustMatch.cs
```diff
@@ -18,32 +17,36 @@ public class EnumValuesMustMatch : Rule
     {
         public override void Initialize(RuleRunnerContext context) => context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
 
+        private bool isEnum(ITypeSymbol sym) => sym is not null && sym.TypeKind == TypeKind.Enum;
```
ViktorHofer, line 20: According to our capitalization conventions, you want to PascalCase the method name: `IsEnum(...)`

smasher164:
> > @bartonjs do you know if it is a breaking change to change the backing store of an enum? so for example, if it used to be public enum MyEnum and you change it to public enum MyEnum: byte?
> 
> There's at least one way that it breaks...
> 
> ```cs
> public enum MyEnum
> {
>    SomeValue = 5;
> }
> 
> // calling assembly
> const MyEnum SpecialValue = (MyEnum)1024;
> ```
> 
> Change it to be backed by `byte` and now that const decl won't compile.
> 
> There are probably runtime breaks, too, but I apparently didn't think of testing that when doing my breaking changes experiments for FDGv3's appendix on breaking changes.

Would this be considered a breaking change the other way around?

```C#
if (~MyEnum.SomeValue == 0) {
    System.Console.WriteLine("here");
}

public enum MyEnum: byte // change to int to see breakage
{
   SomeValue = 255,
}
```

bartonjs:
> Would this be considered a breaking change the other way around?

* Changing the size (e.g. byte \<-> {int,short,long}) could break anything using that enum in a `[StructLayout(LayoutKind.Sequential)]` struct used in a P/Invoke
* Changing the size to something smaller can induce compile-breaks from literals that no longer cast.
  * At runtime it interprets it as the unchecked downcast, even if in a checked block... which is awfully surprising.
* Changing the size to something larger can break assumptions about casting to a numeric type (specifically, the previously used backing type)... either OverflowException (in a checked context) or wrong behavior (unchecked)... assuming something actually took advantage of the new value space.
* Changing between signed and unsigned can introduce a compile break for constants that use the most significant bit (large unsigned or negative signed).

The P/invoke one is probably the worst; but they're all bad, and so probably any backing type change should count as a reportable break.

smasher164:
https://docs.microsoft.com/en-us/dotnet/core/compatibility/

> **DISALLOWED**: Changing the underlying type of an enumeration
>
> This is a compile-time and behavioral breaking change as well as a binary breaking change that can make attribute arguments unparsable.

Looks like this is codified in the compatibility guidelines as well, and there's an issue to detect for this: https://github.com/dotnet/sdk/issues/23902.

I'll make a separate rule for that.

smasher164:
It makes sense to implement a rule for #23902 in the same `Rule` class, because we anyways have to compare types to know if the constant values are comparable.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/EnumsMustMatchTests.cs
```diff
@@ -0,0 +1,160 @@
...
+    C = 2,
+    D = 3,
+  }
+}
+";
+
+            string rightSyntax = @"
+namespace CompatTests
+{
+  public enum First: int {
```
ViktorHofer, line 141: nit: the inheritance operator is usually separated by a whitespace.

```suggestion
  public enum First : int {
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/EnumsMustMatchTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            Assert.Empty(differences);
+        }
+
+        [Fact]
+        public static void BackingStoreChanged()
+        {
+            string leftSyntax = @"
+namespace CompatTests
+{
+  public enum First: short {
```
ViktorHofer, line 129:
```suggestion
  public enum First : short {
```

ericstj:
> cc'ing @ericstj for his thoughts on if these should be considered breaking.

Agree with @bartonjs.  Changing base type of an enum is breaking.  Changing the values of an enum member are also breaking.  We should have separate diagnostics for those.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/EnumsMustMatch.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            // Enum must be a named type to access its underlying type.
+            if (left is not INamedTypeSymbol l || right is not INamedTypeSymbol r)
+            {
+                return;
+            }
+
+            // Get enum's underlying type.
+            if (l.EnumUnderlyingType is not INamedTypeSymbol leftType || r.EnumUnderlyingType is not INamedTypeSymbol rightType)
```
ViktorHofer, line 38: When would the EnumUnderlyingType not be a INamedTypeSymbol?

smasher164: `EnumUnderlyingType` is of type `INamedTypeSymbol?`. When nullable reference types are enabled, it gets flagged if I try to use it as an `INamedTypeSymbol`. Using `is not` lets me match against it and early return when it's anything other than `INamedTypeSymbol`.

joperezr:
> but I wonder if we should think about [Flags](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-6.0) any differently?

While I understand that this will be the most common case, theoretically you still broke compatibility due to the exact same reason, so it would feel to me a bit odd to have two diagnostics pointing to the exact same issue, with the only difference being if your enum is Flags or not. I'm kind of assuming that whenever this is the case, the diagnostic would probably just have to be suppressed as once you have shipped a value like 'All' and want to grow the enum there is really not many options other than breaking that. Also, I think flags enums is honestly one of the most important aspects of this rule, as they are the types that care the most and are the most sensitive to value changes.

joperezr:
Imagine this enum shipping in version 1.0 of an assembly:
```c#
public enum Toppings : Flags
{
  Tomato = 0x01,
  Onion = 0x02,
  Ketchup = 0x04,
  All = 0x07
}
```

And then 2.0 comes around, and needs to add a topping:

```diff
public enum Toppings : Flags
{
  Tomato = 0x01,
  Onion = 0x02,
  Ketchup = 0x04,
+  Mayo = 0x08,
-  All = 0x07
+  All = 0x0F
}
```

That is why, as @ericstj points out, usually having an `All` on your flags enum is a bad pattern as you basically are stuck with not being able to add to it without breaking change, or having `All` no longer mean all.

ericstj:
Perhaps we could just cover this in documentation.  The more I think about it the more I think it doesn't need a different rule.  If the problem is just to educate folks that *yes, indeed, this rule still applies to Flags* then docs are probably better for that.

smasher164:
Could we also configure the EnumValuesMustMatch rule to run in strict mode? Or, at least enabled for _Flags_ in strict mode.

smasher164:
Added a strict mode test on the MembersMustExist rule for enums.

ViktorHofer (approved):
Looks good to me. Thanks!

ViktorHofer:
The failures are unrelated so I suggest to either rerun the entire CI or just the affected legs. The latter can be done in the AzDO UI when you click on the legs details. The former is done by adding the following comment to the PR: `/azp run dotnet-sdk-public-ci`.

smasher164:
/cc @marcpopMSFT 
This unrelated CI failure has shown up both times I've re-run it. Is this OK to merge?
Thanks

PR #25597, opened 2022-05-23
[6.0.3xx] Disable Microsoft.DN.Compatibility pkg

ViktorHofer:
The Microsoft.DotNet.Compatibility package gets published during servicing but hardcodes the VersionPrefix. Because of that, during a servicing event with stabilized versions, it fails to be pubished as the package with same version already exists upstream.

Don't publish the package unless there's a servicing need for it instead.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
+++ b/src/Compatibility/Microsoft.DotNet.Compatibility/Microsoft.DotNet.Compatibility.csproj
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <TargetFrameworks>net6.0;net472</TargetFrameworks>
-    <IsPackable>true</IsPackable>
+    <!-- Re-enable and update the VersionPrefix in case of a servicing change. -->
+    <IsPackable>false</IsPackable>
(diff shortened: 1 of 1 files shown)


mmitche (approved):

marcpopMSFT:
@ViktorHofer I assume your other change in main means this shouldn't go to main. What should we do with 6.0.4xx? Will we want to put this in after 400 ships but before 401?

ViktorHofer:
> I assume your other change in main means this shouldn't go to main. What should we do with 6.0.4xx? Will we want to put this in after 400 ships but before 401?

Right, we don't want this change in main as there this is handled differently, but we want this change in any release/6.0.xxx branch.

marcpopMSFT:
> Right, we don't want this change in main as there this is handled differently, but we want this change in any r**elease/6.0.xxx branch.**
@ViktorHofer to clarify, 6.0.4xx is still in preview and has not shipped yet. Is this safe to flow or do we need this only after it's shipped once?

marcpopMSFT:
CC @v-wuzhai as we'll want to remove this from the codeflow to main when that comes up.

ViktorHofer:
This change should be safe to flow.

PR #25615, opened 2022-05-24
Fix F# pattern matching codegen regression

vzarytovskii:
## Description

A very common feature of F# - pattern matching - has a extremely bad regression in 6.0.300. This regression results in a pattern silently runs the wrong code branch at runtime.

[PR](https://github.com/dotnet/fsharp/pull/13178)
[Issue](https://github.com/dotnet/fsharp/issues/13175)

## Customer Impact

The effect on users is that applications will behave incorrectly at runtime - _incorrect logic is executed_. This will be particularly difficult for users because the resulting behavior is unusual - a derived class may match a base class. We expect this to impact a significant number of users and it has already been reported independently by two different users.

We build the compiler with an earlier version of the compiler so the current shipping compiler is not affected by this regression.

One way this problem may occur is:

```fsharp
type A() = class end
type B() = inherit A()
type C() = inherit A()

let toName (x: obj) =
match x with
| :? A when false -> "not possible"
| :? B -> "B"
| :? C -> "C" // there will be a warning here in the IDE/build output
| _ -> "unknown"

toName (C()) // This will print "B", which is clearly wrong
```

## Regression

Yes. Introduced in 6.0.300.

## Testing

The error was in the optimizer.

234 new tests were added to cover this scenario and many adjacent scenarios in pattern matching. Tests now cover:

* Subtype analysis and relationships
* Guard clauses and how they interact with sub-typing checks
* Sealed types and how they interact with inheritance hierarchy checks
* Multi-column (multi-value) matches and how those interact with exhaustivity checks

## Risk

The risk of taking this fix is isolated to F#. Within that context of F#, the fix is straightforward and limited to the pattern matching optimizer and a single kind of pattern. As such, the risk to F# is small.


kasperk81:
@vzarytovskii, change section is empty. what is the `<null>` commit 4948e2a?

baronfel:
Closing in favor of the maestro update PR #25617, which does what this PR intended to do.

PR #25628, opened 2022-05-25
Update the RID target from lower win to the latest one

v-wuzhai:


marcpopMSFT:
@v-wuzhai let's hold off on any more win-* rid updates. win-x86 and win-x64 are fairly common rids so it's ok to leave tests targeting those versus updating them to win10

PR #25647, opened 2022-05-25
Implement new flag for ignoreing apphosts in dotnet run

### LINKED ISSUE #22544: `dotnet run` of a newly created ASP.NET Core project on MacOS prompts for password to allow keychain access
Feedback from https://twitter.com/leastprivilege/status/1458505534093643776

## Problem
When using `dotnet run` on a newly created ASP.NET Core project on MacOS, the user is prompted by the OS to allow access to the Keychain so that the app can access the HTTPS developer certificate. After entering their password and clicking "Allow" or "Always Allow" the app starts. This process must be repeated for every new ASP.NET Core project created.

*MacOS prompt showing message "Host6 wants to access key "tmpE4viKq" in your keychain."
A textbox for the password and three buttons are also shown: Always Allow, Deny, Allow.*

image

## Details
In .NET 6 `dotnet run` was changed to launch the app via the AppHost executable rather than via `dotnet MyApp.dll`. While this has advantages (e.g. support for cross-architecture app launching, the process name being the project name, etc.) it results in every new project being a new executable from the OS's security point-of-view, and thus requires each new web app to explicitly be granted access to the Keychain.

Note the issue doesn't occur when launching via Visual Studio for Mac 2022 as it doesn't use `dotnet run` to launch the app.

## Workaround
To workaround this issue the [`<UseAppHost>` MSBuild property](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#useapphost) can be set in the project file to `false`, e.g.:

```xml
<PropertyGroup>
</PropertyGroup>
```

You can also pass this property as part of the `dotnet run` command itself:

```command
> dotnet run -p:UseAppHost=false
```

Note that this property completely disables a native executable being created for the project during build, not just in the context of launching the project via `dotnet run`.

## Proposed Changes
We should ensure users are not faced with this prompt as part of the default experience for creating and running ASP.NET Core apps on MacOS.

To that end, we could consider changing the default value for the `<UseAppHost>` MSBuild property to `false` when a project uses the `Microsoft.NET.Sdk.Web` SDK. As noted above however this would actually disable the production of a native executable during build, and so has side-effects beyond the scope of launching the project via `dotnet run`. For that reason we should likely consider alternate approaches:
- We could consider adding support to `dotnet run` for controlling this behavior via a new environment variable, e.g. `DOTNET_USE_APP_HOST`, such that it could be set in the scope of the machine, profile, session, or process launch.
- We could also consider adding a new property to `launchSettings.json` under the `"profiles"` section, placed on a profile, that controls whether the project is launched via the AppHost or not when launched from `dotnet run` or an IDE (i.e. this property should also be honored in environments that launch the project using the AppHost today, like Visual Studio).
  - Note that while `launchSettings.json` supports setting environment variables for the launched app process, attempting to use that to set the variable proposed above (`DOTNET_USE_APP_HOST`) would not work as that variable needs to be set in the context of `dotnet run` itself, not the app subsequently being launched).

baronfel:
This is a draft PR to fix #22544.

It introduces a new flag `UseAppHostIfAvailable` that defaults to `true`. If this flag is ever false, then we evaluate the project with `UseAppHost=false`, which forces the RunCommand and RunArguments properties to invoke the dll directly.

TODO:
* [ ] should we change the default behavior on macos specifically? As it is, we will require users to create and maintain launch profiles.

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Internal/MsBuildFileSetFactory.cs
+++ b/src/BuiltInTools/dotnet-watch/Internal/MsBuildFileSetFactory.cs
@@ -27,14 +27,16 @@ namespace Microsoft.DotNet.Watcher.Internal
         private readonly ProcessRunner _processRunner;
         private readonly bool _waitOnError;
         private readonly IReadOnlyList<string> _buildFlags;
+        private readonly bool _useAppHostIfAvailable;
 
         public MsBuildFileSetFactory(
             IReporter reporter,
             DotNetWatchOptions dotNetWatchOptions,
             string projectFile,
             bool waitOnError,
-            bool trace)
-            : this(dotNetWatchOptions, reporter, projectFile, new OutputSink(), waitOnError, trace)
+            bool trace,
+            bool useAppHostIfAvailable)
+            : this(dotNetWatchOptions, reporter, projectFile, new OutputSink(), waitOnError, trace, useAppHostIfAvailable)
         {
         }
 
@@ -45,7 +47,8 @@ namespace Microsoft.DotNet.Watcher.Internal
             string projectFile,
             OutputSink outputSink,
             bool waitOnError,
-            bool trace)
+            bool trace,
+            bool useAppHostIfAvailable)
         {
             Ensure.NotNull(reporter, nameof(reporter));
             Ensure.NotNullOrEmpty(projectFile, nameof(projectFile));
@@ -57,6 +60,7 @@ namespace Microsoft.DotNet.Watcher.Internal
             _outputSink = outputSink;
             _processRunner = new ProcessRunner(reporter);
             _buildFlags = InitializeArgs(FindTargetsFile(), trace);
+            _useAppHostIfAvailable = useAppHostIfAvailable;
 
             _waitOnError = waitOnError;
         }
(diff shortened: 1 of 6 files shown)


vitek-karas:
I find the naming confusing. Specifically there's a weird contradiction between the project file and the launch settings. If I specify `UseAppHost` (whichever value) explicitly in my project file, this new setting feels really confusing. It's like "I already told you to use app host (or not) and now I'm telling you again?"
Maybe a better name would be to make the name more specific to "Run", so `RunAppHostIfAvailable` or `LaunchAppHostIfAvailable`.

As for the TODO: most non-web-apps will work just fine on macOS with apphost. I would still prefer if we were to change the default only for web apps where we know it's the other way round (most of the time they won't work).

baronfel:
@vitek-karas interesting point about webapps specifically being hit by this - I'd expect that any macos apphost-built app that uses usersecrets would hit this prompt on startup, right?

vitek-karas:
@baronfel you're right that it affects apps which access the secrets, but based on feedback (outside of this issue I haven't seen anything), it's probably not a common thing other than web apps.

agocke:
If this is mostly a web-app problem, is there any other way to mitigate the problem? What do other programming languages do?

baronfel:
Closing this because we're going a different direction to solve the root issue.

PR #25713, opened 2022-05-31
Delete unused resource files in ApiCompatibility

ViktorHofer:
These files were brought in via cb8c977b9c0e6a32e0e34ed3698ae11dbbf8b4c1 but aren't in use.


danmoseley (approved):
If it builds, presumably it's good.

Issue #24480, opened 2022-03-22
More detailed error message when SDK is not found


KirillOsenkov:
Currently when global.json specifies an SDK that's not installed, here's the error you get:

```
Unable to locate the .NET SDK. Check that it is installed and that the version specified in global.json (if any) matches the installed version.
```

This error is not nearly detailed enough. It is also very important as it's a gateway to our entire ecosystem (think someone new to .NET clones a repo, tries to build and sees this). We need to make this first run experience as smooth as possible, and now it's anything but.

Ways I see that we can improve this message:

 1. be explicit whether there was a global.json or not, and if yes, what was the exact location
 2. which version is expected by global.json?
 3. which versions are currently installed and were reachable by the resolver?
 4. what to do next? Provide a URL that is as precise as possible to download (perhaps get the exact URL to the blob)

Long term we need to think about potentially downloading and installing the SDK automatically and on demand if it wasn't found. Perhaps ask "would you like to download and install it now?" and have a -force flag or something that does it silently (for CI scenarios). This way we can avoid having a separate Use .NET SDK CI step, and we could just do msbuild /r /ensuresdk or something to make sure we can build the repo in a single automatable step.

richlander:
Here is my related prior art to align with:

- https://github.com/dotnet/runtime/pull/67564
- https://github.com/dotnet/runtime/pull/67022
- https://github.com/dotnet/sdk/pull/24541

PR #25733, opened 2022-05-31
Include requested .NET SDK version when known

sharwell:
Please let me know if/where tests should be for this. Relies on https://github.com/dotnet/runtime/pull/68355 to enable the new error message path.

Related to #24480

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
+++ b/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
@@ -135,4 +135,7 @@
   <data name="GlobalJsonResolutionFailed" xml:space="preserve">
     <value>Unable to locate the .NET SDK as specified by global.json, please check that the specified version is installed.</value>
   </data>
+  <data name="GlobalJsonResolutionFailedSpecificVersion" xml:space="preserve">
+    <value>Unable to locate the .NET SDK version '{0}' as specified by global.json, please check that the specified version is installed.</value>
+  </data>
 </root>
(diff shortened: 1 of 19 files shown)


src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
```diff
@@ -135,4 +135,7 @@
   <data name="GlobalJsonResolutionFailed" xml:space="preserve">
     <value>Unable to locate the .NET SDK as specified by global.json, please check that the specified version is installed.</value>
   </data>
+  <data name="GlobalJsonResolutionFailedSpecificVersion" xml:space="preserve">
+    <value>Unable to locate the .NET SDK version '{0}' as specified by global.json, please check that the specified version is installed.</value>
```
baronfel, line 139: Looks great, thanks for doing this. If I can be greedy for a moment since you're in here already - can you incorporate the path to the global.json file that caused this resolution result (which should be present in `resolverResult.GlobalJsonPath`) into the error message as well?

```
Unable to locate the .NET SDK version '{0}' as specified by the global.json at {1}, please check that the specified version is installed.
```

And similar for the `GlobalJsonResolutionFailed` error message directly above this?

baronfel (approved):

src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Strings.resx
sharwell: With the second variable, there are now four cases to consider:

1. Unknown global.json path, unknown version
2. Known path, unknown version
3. Unknown path, known version
4. Known path, known version

The easiest option is to use the new message only for case (4), and the only case which is penalized from such a change is case (3) since it would no longer show the expected version. Will that be a problem?

PR #25734, opened 2022-05-31
add rules to catch addition or removal of 'virtual' modifier

### LINKED ISSUE #18682: [API Compat] Add rule to make sure virtual modifier is not removed
This rule checks that a member should be virtual on both the implementation and in the contract when the member is overridable:

1. Is virtual.
2. Is not sealed.
3. Type is not sealed.

This rule should support StrictMode and make sure both sides are virtual when in strict mode.

cc: @ericstj @Anipik

### LINKED ISSUE #18683: [API Compat] Add rule to prohibit making a member virtual
Adding the virtual keyword to a member is a breaking change because the consumer code would often end up calling it non-virtually when the member is now virtual.

This rule should just run when not in strict mode as when running on strict mode that will be covered by: https://github.com/dotnet/sdk/issues/18682

cc: @ericstj @Anipik

smasher164:
There are a couple modifications to the requirements in #18682 and #18683.

1. Since it is not possible for a member to simultaneously have the 'virtual' keyword, and also be sealed or in a sealed class, the removal diagnostic only needs to check that the left member is virtual.
2. Since it is legal to change an abstract member to a virtual one, we do not emit an addition diagnostic when the left member is abstract.

Fixes #18682
Fixes #18683


src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticIds.cs
```diff
@@ -19,6 +19,8 @@ public static class DiagnosticIds
         public const string CannotSealType = "CP0009";
         public const string EnumTypesMustMatch = "CP0010";
         public const string EnumValuesMustMatch = "CP0011";
+        public const string CannotRemoveVirtualModifier = "CP0012";
```
joperezr, line 22: NIT: In order to be more consistent to other DiagnosticIds, can we name them something along the lines of `CannotMakeMemberVirtual` and `CannotRemoveVirtualFromMember`?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -213,4 +213,10 @@
   <data name="EnumTypesMustMatch" xml:space="preserve">
     <value>Underlying type of enum '{0}' changed from '{1}' to '{2}'.</value>
   </data>
+  <data name="CannotAddVirtualModifier" xml:space="preserve">
+    <value>Cannot add virtual modifier to member '{0}'.</value>
```
joperezr, line 217:
```suggestion
    <value>Cannot make member '{0}' virtual.</value>
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
```diff
@@ -213,4 +213,10 @@
...
+  </data>
+  <data name="CannotRemoveVirtualModifier" xml:space="preserve">
+    <value>Cannot remove virtual modifier from member '{0}'.</value>
```
joperezr, line 220:
```suggestion
    <value>Cannot make member '{0}' non-virtual.</value>
```

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
```diff
@@ -0,0 +1,58 @@
...
+        public override void Initialize(RuleRunnerContext context) => context.RegisterOnMemberSymbolAction(RunOnMemberSymbol);
+
+        private void RunOnMemberSymbol(ISymbol left, ISymbol right, ITypeSymbol leftContainingType, ITypeSymbol rightContainingType, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            // Members must exist
+            if (left is null || right is null)
+            {
+                return;
+            }
+            // If the left member is virtual, it must not be sealed or inside a sealed type.
```
joperezr, line 30:
```suggestion
            }

            // If the left member is virtual, it must not be sealed or inside a sealed type.
```
Style-NIT: Empty line after closing block. Same feedback on the rest of the cases.

joperezr: Also, I'm not sure I follow this part of the comment: `it must not be sealed or inside a sealed type.` I don't see how that relates to the code you write below.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                // If the right member is not virtual, emit a diagnostic
+                // that the virtual modifier cannot be removed.
+                if (!right.IsVirtual)
+                {
+                    differences.Add(new CompatDifference(
+                    DiagnosticIds.CannotRemoveVirtualModifier, string.Format(
+                        Resources.CannotRemoveVirtualModifier, left), DifferenceType.Removed, right));
+                }
+            }
```
joperezr, line 41:
```suggestion
            if (left.IsVirtual && !right.IsVirtual)
            {
                // If the right member is not virtual, emit a diagnostic
                // that the virtual modifier cannot be removed.
                differences.Add(new CompatDifference(
                    DiagnosticIds.CannotRemoveVirtualModifier, string.Format(
                    Resources.CannotRemoveVirtualModifier, left), DifferenceType.Removed, right));
            }
```
NIT: no need to have separate ifchecks

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
```diff
@@ -0,0 +1,58 @@
...
+            // Otherwise if the left member is not virtual, ensure that
+            // we're either in strict mode, or the left member is not
+            // abstract (since it is legal to change abstract to virtual).
+            else if (Settings.StrictMode || !left.IsAbstract)
```
joperezr, line 45: mm I would have expected that an abstract member is ALSO virtual. Is there a case where `left.IsAbstract` is `true` but `left.IsVirtual` is false?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
smasher164: I mention the sealed thing because https://github.com/dotnet/sdk/issues/18682 originally mentions that the left member is overridable if it's virtual, not sealed, and its containing type is not sealed. However, it's not possible to declare a member with the virtual modifier that is also sealed or inside a sealed type, so the condition just reduces to `if (left.IsVirtual)`.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
smasher164: In this case, the `else if` branch assumes that `left.IsVirtual` is `false`, so combining these two conditions would break the `else`. I *could* do
```C#
if (left.IsVirtual && !right.IsVirtual) {
...
}

if (!left.IsVirtual && (Settings.StrictMode || !left.IsAbstract)) {
...
}
```
but the nested if statements meant I only had to check `left.IsVirtual` once.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
smasher164: I based this off of
https://docs.microsoft.com/en-us/dotnet/core/compatibility/
where it mentions that
✔️ ALLOWED: Changing a member from abstract to virtual

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
joperezr: right, I guess it's up to you, but I would just remove the seal comment since it isn't even possible to begin with. People won't look into the issue and instead look into the code only, and from the way the comment reads, I kind of understood from it that you were saying that if the ifcheck was true, then inside you would ensure the type was not sealed but that is not possible from the compiler.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
joperezr: oh, I missed that.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
smasher164: Ah I see. Roslyn returns `false` for `IsVirtual` on an abstract member. I think it's just checking for the presence of the keyword, not if they use virtual dispatch underneath the hood.

joperezr: oh interesting, nevermind then.

smasher164: But this begs the question -- when the original issue mentions `virtual`, is it talking about the keyword or if the member is *actually* virtual? Hopefully I didn't misinterpret it. I'll check with Santiago.

smasher164: So it *is* just about the keyword -- so this logic is sound.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Resources.resx
smasher164: While this *is* shorter, I worry that "cannot" and "non-virtual" is bit hard to parse, since it's a double negative.

joperezr: My feedback is not so much about the wording, so feel free to change that as you wish, but more around having the message express the breaking change as opposed to the action that caused the breaking change. For example, the error message above that you added on the previous PR says something around "Underlying type of enum changed from foo to bar" instead of saying: "Cannot add `: short` to enum Foo"

joperezr: Actually I just went to our [official docs for breaking changes](https://docs.microsoft.com/en-us/dotnet/core/compatibility/) to see how we word it over there, and they seem to agree with what you had originally:

image

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
```diff
@@ -0,0 +1,58 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiCompatibility.Abstractions;
+
+#nullable enable
```
ViktorHofer, line 11: Thanks for that @smasher164. I wanted to recommend to you using the compiler's nullability feature anyway. Great that you already did it yourself. We might want to also look at the other existing rules and see if we can easily turn nullability on for them as well. Especially as I found some issues when I turned it on in the other Compatibility components.

ViktorHofer, line 11: Thanks for that @smasher164. I wanted to recommend to you using the compiler's nullability feature anyway. Great that you already did it yourself. We might want to also look at the other existing rules and see if we can easily turn nullability on for them as well. Especially as I found some issues when I turned it on in the other Compatibility components.

smasher164: I opened https://github.com/dotnet/sdk/issues/25760.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualModifier.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            // Members must exist
+            if (left is null || right is null)
+            {
+                return;
+            }
+
+            if (left.IsVirtual)
+            {
+                // If left is virtual right is not, emit a diagnostic
```
joperezr, line 33:
```suggestion
                // If left is virtual and right is not, then emit a diagnostic specifying
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualModifierTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            CompatDifference[] expected = new[]
+            {
+                new CompatDifference(DiagnosticIds.CannotRemoveVirtualModifier, string.Empty, DifferenceType.Removed, "M:CompatTests.First.F"),
+                new CompatDifference(DiagnosticIds.CannotAddVirtualModifier, string.Empty, DifferenceType.Added, "M:CompatTests.First.G"),
+            };
+            Assert.Equal(expected, differences);
+        }
+
+        [Fact]
+        public static void Null()
```
joperezr, line 53: Can we change the test names to names that are more descriptive of what is being tested? For example, for this one, I would write something like: `EnsureNoDiagnosticsLoggedForAddedOrRemovedVirtualMembers()`

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualModifierTests.cs
```diff
@@ -0,0 +1,115 @@
...
+  public class First {
+    public virtual void G() {}
+  }
+}
+";
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new();
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(new[] { left }, new[] { right });
+            Assert.NotEmpty(differences);
```
joperezr, line 75: I don't think this test is testing anything really valuable. What you want to make sure here, is to test that the differences reported do NOT contain the diagnostics that are thrown from your new rule, and instead only MembersMustExist diagnostics.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualModifierTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        [Fact]
+        public static void StrictMode()
```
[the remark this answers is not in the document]
joperezr: Even when the names are long, you don't care about the length of the member because no one is going to call it, what you want is for it to be descriptive enough so that when it fails, you immediately know what the unit test is checking and even probably have a good guess of why it broke.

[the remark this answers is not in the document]
smasher164: I assumed the test runner would namespace these test names under `CannotAddOrRemoveVirtualModifierTests` inside `Microsoft.DotNet.ApiCompatibility.Tests`, so that I wouldn't have to guess. But I agree that this name could be more specific.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualModifierTests.cs
smasher164: My initial goal was to make sure that null members on the left or right do not cause the test to fail with a `NullReferenceException`.

joperezr: right, and I think that is fine, but this Assert is not doing that. The fact that an exception is not thrown is simply by calling `differ.GetDifferences` and having that not throw an exception. What I'm saying, is that it would be good to also ensure not only that the code didn't hit any exceptions, but also that the rule didn't emit any diagnostics incorrectly.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualModifierTests.cs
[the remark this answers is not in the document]
joperezr: It depends on which view you use, some do have the fully qualified name, but not all. Even then, I think it is fine to Omit from the name some things, like the name of the rule for example, since as you point out, this is already on the test class name, but having a more descriptive name of what the test is actually testing is useful.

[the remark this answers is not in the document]
ViktorHofer: I usually follow our own documentation on unit test naming conventions: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests.

A method should include:
- The name of the method that is being tested
- The name of the scenario
- And the expected return/output

There's no need to encode the class name in it as it's already part of the method's signature.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualKeyword.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        public override void Initialize(RuleRunnerContext context) => context.RegisterOnMemberSymbolAction(RunOnMemberSymbol);
+
+        private void RunOnMemberSymbol(ISymbol left, ISymbol right, ITypeSymbol leftContainingType, ITypeSymbol rightContainingType, string leftName, string rightName, IList<CompatDifference> differences)
+        {
+            // Members must exist
+            if (left is null || right is null)
+            {
+                return;
+            }
```
ViktorHofer, line 29: I don't think that this branch can ever be reached but I understand that you need to add this to suppress compiler nullable errors. Is there a way to declare that either left or right must be not null  via nullable parameter attributes?

smasher164: The second unit test (`EnsureNoCrashWhenMembersDoNotExist`) reaches this branch. If you have members with two different names, this branch gets hit.

ViktorHofer: Oh I did misread that as an AND instead of an OR. I think there's a rule that will have such an issue but we can defer that discussion to when we enable nullable in the other rules.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualKeywordTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            for (int i = 0; i < args.Length; i++)
+            {
+                if (args[i].dt == DifferenceType.Removed)
+                {
+                    differences[i] = new CompatDifference(DiagnosticIds.CannotRemoveVirtualFromMember, string.Empty, DifferenceType.Removed, args[i].memberId);
+                }
+                else if (args[i].dt == DifferenceType.Added)
+                {
+                    differences[i] = new CompatDifference(DiagnosticIds.CannotAddVirtualToMember, string.Empty, DifferenceType.Added, args[i].memberId);
+                }
```
joperezr, line 40:
```suggestion
                differences[i] = new CompatDifference(
                    args[i].dt == DifferenceType.Removed ? 
                        new CompatDifference(DiagnosticIds.CannotRemoveVirtualFromMember, string.Empty, DifferenceType.Removed, args[i].memberId) :
                        new CompatDifference(DiagnosticIds.CannotAddVirtualToMember, string.Empty, DifferenceType.Added, args[i].memberId);
```
NIT: this could be simplified

joperezr (approved):
Left some minor comments, but looks good otherwise.

ericstj (approved):

Issue #22901, opened 2021-12-03
[Package Validation] We should support excluding tfms when running the validation


safern:
In dotnet/runtime we have been following a convention with our packages where we drop TFMs that are out of support from our packages. This results on a lot of errors that need to be suppressed because we are intentionally dropping these TFMs.

Also, when this happens, in baseline validation,  if an asset is drop it will take another asset that is compatible for that TFM in the new package, i.e, if we had netcoreapp3.1 and netstandard2.0 asset, then we drop netcoreapp3.1, and now it will compare the netcoreapp3.1 asset from the baseline package vs the netstandard2.0 from the new version, resulting in a lot of errors as the API surface might be fairly different, so that comparison is not valuable in a lot of cases and it can become very noisy.

I suggest we add some support like an ItemGroup to specify ignored tfms for baseline validation, i.e:

```xml
<PackageValidationBaselineIgnoreTfm Include="netcoreapp3.1" />
```

Or we could introduce a flag to ignore out of support tfms and then we can use the list from the SDK for which tfms are out of support, i.e:
```xml
<PackageValidationBaselineIgnoreEolTfms>true</PackageValidationBaselineIgnoreEolTfms>
```

cc: @ericstj @joperezr

joperezr:
While I agree this will become more common as time goes on and TFMs continue to go out of support, I have a few thoughts in terms of the design of the feature. Here are some of the things in my mind:
 - Technically we already have support for this by just baselining the error for that specific suppression, which you can easily just add a comment on top of the specific suppression and comment that it is due to the TFM going out of support. That said, I also realize that due to the fact that the suppression file is auto-generated, you might lose the comment if the file is re-generated later.
 - The problem with having this suppression via Item or property, means that the consumer project might have suppressions for a lot of TFMs that will no longer warn after the baseline version is updated and no longer has the error.
 - Regarding your comment that: `resulting in a lot of errors as the API surface might be fairly different, so that comparison is not valuable` I actually think it is great that when doing this, you can basically get the preview of the impact that your user's targeting that dropped TFM will get after they upgrade. For instance, we just dropped some TFMs from our dotnet/iot packages, and in our case PackageValidation helped a lot measuring the kind of impact that consumers would see once they upgraded their package.

[...]

PR #25765, opened 2022-06-01
Compatibility tooling refactoring

ViktorHofer:
I recommend to review by commit even though it wasn't able to isolate all the change into buildable separate commits.

Here's a summary of the changes:
1. Update the solution files as the PackageValidation test project was missing and add a new non-shipping Compatibility.Console project which makes debugging much easier.
2. Move the logger into ApiCompatibility and extend it with new features so that a future ApiCompatibility task can leverage it.
3. Fix the MetdataInformation type which only overrode the IEquatable<T>.Equals overload but not the object.Equals one (they should be in sync).
4. Move the SuppressionEngine and Suppression types into Logging namespace and make the SuppressionId a required field.
5. Nit, comment fix.
6. Simplify Package type, merge with NuPkgParser and avoid unnecessary LINQ iterations.
7. Nit, msbuild code clean-up
8. Update other files for nullability and optimize a few code paths.
9. Update validators for nullability and optimize code paths.
10. Nit, test namespace fixes and using statement clean-up
11. Test file updates because of the above refactoring.


src/Compatibility/Microsoft.DotNet.Compatibility.Console/Program.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Compatibility.Console;
+using Microsoft.DotNet.PackageValidation;
+
+if (args.Length != 11)
+{
+    throw new ArgumentException("Pass in the set of required parameters...");
```
smasher164, line 9: It's unclear to me what the usage of this CLI tool looks like based on just the args below.
Could we add a usage message?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/MetadataInformation.cs
```diff
@@ -25,6 +25,9 @@ public MetadataInformation(string name, string targetFramework, string assemblyI
             DisplayString = displayString ?? assemblyId;
         }
 
+        public override bool Equals(object obj) =>
+            Equals(obj is MetadataInformation) && Equals((MetadataInformation)obj);
```
smasher164, line 29: I thought the `is` operator returns a boolean?
This calls `Equals(boolean)`?
Should it be
```C#
obj is MetadataInformation && Equals((MetadataInformation)obj)
```
or perhaps
```C#
obj is MetadataInformation md && Equals(md)
```

ViktorHofer: Oh right, thanks.

src/Compatibility/Microsoft.DotNet.Compatibility.Console/Program.cs
ViktorHofer: Fixed. The console app now uses System.CommandLine for parsing the args.

smasher164 (approved):
LGTM!

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/CompatibilityLoggerBase.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+using System;
+using System.IO;
+
+namespace Microsoft.DotNet.ApiCompatibility.Logging
+{
+    public abstract class CompatibilityLoggerBase
```
joperezr, line 11: Please add tripple slash comments on the new class.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
```diff
@@ -16,7 +15,7 @@ public class Suppression : IEquatable<Suppression>
         /// <summary>
         /// The DiagnosticId representing the error to be suppressed.
         /// </summary>
-        public string? DiagnosticId { get; set; }
+        public string DiagnosticId { get; set; }
```
joperezr, line 18: I may be miss-remembering, but I believe the original reason why DiagnosticId was nullable is because we wanted to be able to have global suppressions in the future where you could suppress all errors for a particular pair of left and right, without the need to call out each individual DiagnosticId being suppressed.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -22,16 +22,19 @@ public class SuppressionEngine
...
-        private readonly HashSet<string?> _noWarn;
+        private readonly HashSet<string> _noWarn;
 
-        protected SuppressionEngine(string? suppressionFile, string noWarn)
+        /// <summary>
+        /// Creates a new instance of <see cref="SuppressionEngine"/>.
+        /// </summary>
+        /// <param name="noWarn"></param>
+        /// <param name="suppressionFile">The path to the suppressions file to be used for initialization.</param>
+        public SuppressionEngine(string? noWarn = null, string? suppressionFile = null)
```
joperezr, line 32: Seems like you are reversing the order of the params and params are both strings so we should be very careful on all of the callsites and make sure they all got updated correctly. Additionally, if there is no need to update the ordering, leaving it how it was is preferrable.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -22,16 +22,19 @@ public class SuppressionEngine
...
         private readonly ReaderWriterLockSlim _readerWriterLock = new();
         private readonly XmlSerializer _serializer = new(typeof(Suppression[]), new XmlRootAttribute("Suppressions"));
-        private readonly HashSet<string?> _noWarn;
+        private readonly HashSet<string> _noWarn;
 
-        protected SuppressionEngine(string? suppressionFile, string noWarn)
+        /// <summary>
+        /// Creates a new instance of <see cref="SuppressionEngine"/>.
+        /// </summary>
+        /// <param name="noWarn"></param>
```
joperezr, line 30: NIT: Empty description

src/Compatibility/Microsoft.DotNet.Compatibility/MSBuildCompatibilityLogger.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+using Microsoft.DotNet.ApiCompatibility.Logging;
+using Microsoft.NET.Build.Tasks;
+
+namespace Microsoft.DotNet.Compatibility
+{
+    internal class MSBuildCompatibilityLogger : CompatibilityLoggerBase
```
joperezr, line 11: Please add triple slash docs to the new type.

joperezr: NIT: inherit doc is fine for the members after you document the base, but it would be good to add a description for the class itself, so we easily understand when it is used.

src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
```diff
@@ -18,35 +18,35 @@ namespace Microsoft.DotNet.Compatibility
     public class ValidatePackage : TaskBase
     {
         [Required]
-        public string PackageTargetPath { get; set; }
+        public string? PackageTargetPath { get; set; }
```
joperezr, line 21: These Required parameters should not be marked as nullable

src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
```diff
@@ -96,45 +90,40 @@ protected override void ExecuteCore()
...
-        private Assembly LoadRoslyn(AssemblyName name, Func<string, Assembly> loadFromPath)
+        private Assembly? LoadRoslyn(AssemblyName name, Func<string, Assembly> loadFromPath)
         {
             const string codeAnalysisName = "Microsoft.CodeAnalysis";
             const string codeAnalysisCsharpName = "Microsoft.CodeAnalysis.CSharp";
             if (name.Name == codeAnalysisName || name.Name == codeAnalysisCsharpName)
             {
-                Assembly asm = loadFromPath(Path.Combine(RoslynAssembliesPath, $"{name.Name}.dll"));
-                Version resolvedVersion = asm.GetName().Version;
+                Assembly asm = loadFromPath(Path.Combine(RoslynAssembliesPath!, $"{name.Name}.dll"));
```
joperezr, line 125: This is an input that you are now marking as nullable, so you can't assert it is not null with an exclamation point.

joperezr: Either mark the parameter as non-nullable, or check for null and throw an exception.

joperezr:
> <Project Sdk="Microsoft.NET.Sdk">

What is the benefit of this over a unit test project that runs the validators? I'd rather us not to checkin code that is effectively dead. For internal development, testing and debugging, the unit tests projects should be used instead.

---
Refers to: src/Compatibility/Microsoft.DotNet.Compatibility.Console/Microsoft.DotNet.Compatibility.Console.csproj:1 in a28b887. [](commit_id = a28b887240e68fd969e98731f5ba51a0453d4cb2, deletion_comment = False)

src/Compatibility/Microsoft.DotNet.Compatibility.Shared/ValidatorBootstrap.cs
```diff
@@ -0,0 +1,60 @@
...
+
+#nullable enable
+using System.Collections.Generic;
+using Microsoft.DotNet.ApiCompatibility.Logging;
+using Microsoft.DotNet.PackageValidation;
+using Microsoft.DotNet.PackageValidation.Validators;
+
+namespace Microsoft.DotNet.Compatibility
+{
+    internal static class ValidatorBootstrap
```
joperezr, line 12: I'm honestly not seeing the benefit of having this shared project over what we used to have. This code belongs to the layer that understands how to get a package and queue ApiCompat runs for each pair that should be compared (which means Microsoft.DotNet.PackageValidation) so  I don't think adding this extra layer is adding any value, and instead just adding complexity IMO

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -20,18 +18,30 @@ namespace Microsoft.DotNet.PackageValidation
...
+                StrictMode = enableStrictMode
+            };
             _log = log;
-            _referencePaths = referencePaths ?? new();
+            _referencePaths = referencePaths;
+
+            _leftPackagePath = leftPackagePath;
+            _rightPackagePath = rightPackagePath ?? leftPackagePath;
+            // If assets from different packages are compared, mark the underlying suppression as baseline.
+            _isBaselineSuppression = !_leftPackagePath.Equals(_rightPackagePath, StringComparison.InvariantCultureIgnoreCase);
```
joperezr, line 44: Why not just make _rightPackagePath nullable and check here if it was null?

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
```diff
@@ -1,110 +0,0 @@
...
-using NuGet.ContentModel;
-using NuGet.Frameworks;
-
-namespace Microsoft.DotNet.PackageValidation
-{
-    /// <summary>
-    /// Validates that no target framework / rid support is dropped in the latest package.
-    /// Reports all the breaking changes in the latest package.
-    /// </summary>
-    public class BaselinePackageValidator
```
joperezr, line 18: Did the code on these files change at all? or was this just a file move?

src/Compatibility/Microsoft.DotNet.PackageValidation/CompatibleFrameworkInPackageValidator.cs
```diff
@@ -1,65 +0,0 @@
...
-using NuGet.ContentModel;
-using NuGet.Frameworks;
-
-namespace Microsoft.DotNet.PackageValidation
-{
-    /// <summary>
-    /// Validates that the api surface of the compatible frameworks.
-    /// </summary>
-    public class CompatibleFrameworkInPackageValidator
-    {
```
joperezr, line 17: Same question around code changes here

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
```diff
@@ -17,4 +17,8 @@
   <ItemGroup>
     <EmbeddedResource Update="Resources.resx" GenerateSource="true" />
   </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.DotNet.PackageValidation.Tests" Key="0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9" />
```
joperezr, line 22: We usually never do this. For testing internals, we almost always prefer compiling implementation files on the test project instead.

src/Compatibility/Microsoft.DotNet.PackageValidation/Package.cs
```diff
@@ -19,126 +20,137 @@ namespace Microsoft.DotNet.PackageValidation
...
 
         /// <summary>
-        /// List of the frameworks in the package.
+        /// Creates the package object from package path.
         /// </summary>
-        public IEnumerable<NuGetFramework> FrameworksInPackage { get; private set; }
+        /// <param name="packagePath">The path to the package path.</param>
+        /// <param name="runtimeGraph">The path to the the runtime graph.</param>
+        /// <returns>The package object.</returns>
+        public static Package Create(string? packagePath, string? runtimeGraph)
```
joperezr, line 115: Is the code here a direct copy-paste from what we had in NupkgParser? If so, any reason why we moved it? If we just want it to be a static member on the same class we could have just kept the code separate and make this package type partial.

joperezr: It is usually good to keep files as a boundary for logic, it makes the code easier to find and more readable.

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
joperezr: Oh, I see that we were already doing this. You don't have to fix this now, but we should definitely get rid of the InternalsVisibleTo as this is just messy.

src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
```diff
@@ -1,14 +1,14 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#nullable enable
```
joperezr, line 4: Why not just enable nullability on the whole project?

src/Compatibility/Microsoft.DotNet.PackageValidation/Validators/IPackageValidator.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.PackageValidation.Validators
+{
+    public interface IPackageValidator
```
joperezr, line 6: Please add triple slash comments here too.

src/Compatibility/Microsoft.DotNet.PackageValidation/Validators/PackageValidatorOption.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.PackageValidation.Validators
+{
+    public struct PackageValidatorOption
```
joperezr, line 8: Same here, let's add trippe slash comments. For this one in particular, why making this a struct instead of a read-only struct? There are big performance differences when dealing with structs vs classes, so structs should usually be readonly when possible.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Logging/SuppressionTests.cs
```diff
@@ -0,0 +1,62 @@
...
+            yield return new object[] { new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll" }, new Suppression(" pk004 ") { Target = "A.B", Left = "ref/net6.0/mylib.dll" } };
+            yield return new object[] { new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll", Right = "lib/net6.0/myLib.dll" }, new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll", Right = "lib/net6.0/myLib.dll", IsBaselineSuppression = false } };
+            yield return new object[] { new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll", Right = "lib/net6.0/myLib.dll", IsBaselineSuppression = false }, new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll", Right = "lib/net6.0/myLib.dll", IsBaselineSuppression = false } };
+            yield return new object[] { new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll", Right = "lib/net6.0/myLib.dll", IsBaselineSuppression = true }, new Suppression("PK004") { Target = "A.B", Left = "ref/net6.0/myLib.dll", Right = "lib/net6.0/myLib.dll", IsBaselineSuppression = true } };
```
joperezr, line 21: We should add tests that check that you can omit ceirtain aspects of the suppression in order to more globally suppress items. For example, suppress all errors from a left and a right, or suppress all errors for a certain diagnostic without specifying a target, etc.

joperezr (commented):
Left some comments, please address in a follow up when you get a chance 😄

joperezr:
Did you run this against dotnet/runtime to see if it produced any new warnings? I think we are using global suppressions in some of our suppression files in dotnet/runtime, so if that's the case I think this will break that build.

src/Compatibility/Microsoft.DotNet.Compatibility/MSBuildCompatibilityLogger.cs
joperezr: Also, one more NIT: all of these new types should be `sealed`

src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
ViktorHofer: Because the project imports tons of files from a Common folder that aren't yet marked as nullable and I didn't want to bloat the PR by marking those as nullable disable. Given that I will submit a follow-up PR, I will probably do that now.

src/Compatibility/Microsoft.DotNet.PackageValidation/Package.cs
ViktorHofer: Absolutely agree with keeping files as a boundary for logic. That's why I moved the Create method into Package as the creation of the package belongs into the Package itself. The Package type already heavily relies on NuGet apis, there's no separate that.

> Is the code here a direct copy-paste from what we had in NupkgParser?

Not exactly. It merges the previously existing methods together and leverages the compiler's nullability feature now: https://github.com/dotnet/sdk/blob/4fdc385bc22e9237fef3301777ed71dc4dcff714/src/Compatibility/Microsoft.DotNet.PackageValidation/NupkgParser.cs#L24-L48.

src/Compatibility/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj
ViktorHofer: I don't like InternalsVisibleTo either but that's unrelated to this change. If we feel strongly about it, let's open an issue to track the work.

src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
ViktorHofer: The parameter is already marked as required, that's how msbuild makes sure that a parameter must be set and isn't null. The parameter itself must be marked as nullable as otherwise the compiler complains that the value isn't set inside the constructor. Nullability doesn't work well with msbuild required properties and using the bang operator here is the best way that I could find to suppress these false positives.

src/Compatibility/Microsoft.DotNet.Compatibility/ValidatePackage.cs
ViktorHofer: As mentioned below, it must be marked as nullable as the value doesn't get set in the constructor (it's an msbuild task).

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Logging/SuppressionTests.cs
ViktorHofer: Agreed. Can you please file an issue so that we can track the work?

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
ViktorHofer: The ApiCompatRunner relies on the `_rightPackagePath` variable in other places. Setting the right package path to the left package path makes it easier to later work with the strings. Also I didn't change that, it was already like this before.

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
ViktorHofer: Noted in the commit and PR description but to provide more detail, all validators changed in the following way:
- They now accept an PackageValidatorOption object in the Validate method so that they can operate on multiple packages in the future. The PackageValidatorOption type contains all the information that is specific to a package validation run. The PackageValidator type itself can be cached and reused.
- The ApiCompatRunner is now instantiated during validation as with its current design, it's specific to a package validation run.
- Fix: The ApiCompatRunner's RunApiCompat is now only called when the `RunApiCompat` is true to avoid unnecessary work.

An additional fix landed in the `CompatibleTFMValidator` which compared the same assemblies twice when the runtimeAsset was the same as the runtimeRidSpecificAsset.

And last but not least as mentioned in the Team chat, I noticed that the `BaselinePackageValidator` compares baseline rid specific runtime asset against latest runtime asset which is unintentional. I didn't fix that yet though.

src/Compatibility/Microsoft.DotNet.PackageValidation/CompatibleFrameworkInPackageValidator.cs
ViktorHofer: see my response above

ViktorHofer:
> What is the benefit of this over a unit test project that runs the validators? I'd rather us not to checkin code that is effectively dead. For internal development, testing and debugging, the unit tests projects should be used instead.

There's a ton of benefit of having a Console in addition to unit tests:
1. @safern mentioned in the past that even though initially only an MSBuild Task is provided, a global tool is planned so that usages of the old ApiCompat tooling can't be migrated over when customers intentionally can't or don't want to rely on MSBuild.
2. The unit tests in this repository depend on VSTest which adds a lot of traces and makes it really hard to debug and diagnose issues when you just wanted to execute the tool. Just to run the tests either inside VS or via the CLI you need to have a prepared environment, documented in https://github.com/dotnet/sdk/blob/main/documentation/project-docs/developer-guide.md#run-tests-from-the-command-line.

The Console app is packaged as a global tool and non shipping which means it can be installed from the dotnet7-transport feeds and experimented with, by invoking the `dotnet tool packagevalidation` command.

src/Compatibility/Microsoft.DotNet.Compatibility.Shared/ValidatorBootstrap.cs
ViktorHofer: I have to disagree here. While the ApiCompatibility layer offers the necessary API to compare left with right(s), the PackageValidation layer exposes the API to read from a package, create an internal representation of it (`Package` type) and to validate it (via the Validators). The frontend (console, msbuild, whatever) is responsible for parsing passed in arguments and passing them in to the exposed APIs. The frontend implements a logger (based on the platform) and decides which validators to use (based on the passed in arguments).

The `ValidatorBootstrap` class contains the shareable frontend code and does exactly the above:
1. Leverages the PackageValidation.Package API to create an internal representation of the passed in packages
2. Instantiates and invokes the desired validators based on the user input via the PackageValidation.Validators API.
3. Invokes the exposed ApiCompatibility.Logging API to generate the suppression file.

ViktorHofer:
> 
Did you run this against dotnet/runtime to see if it produced any new warnings? I think we are using global suppressions in some of our suppression files in dotnet/runtime, so if that's the case I think this will break that build.

Yes I did generate the package locally and then use it in dotnet/runtime's build. I did generate a binlog to be sure that the dev version of the PackageValidation package is used:

image

The repository builds successfully: 

image

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
ViktorHofer: I nearly pulled my head out to understand how this could ever possible :D. From what I can see, a DiagnosticId is always required, in any case. A suppression created by PackageValidation, ApiCompatibility or any future component must always contain a DiagnosticId.

The suppression is passed to the [`IsErrorSuppressed` method](https://github.com/dotnet/sdk/blob/c8cd2cfbbe7e2977d20177c03472add5871c4c1a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/ErrorSuppression/SuppressionEngine.cs#L62). From what I noticed, the `error.DiagnosticId == null` check was dead code.

Based on the above, a global suppression is currently defined either as:
- DiagnosticId + Target + IsBaselineSuppression
- DiagnosticId + Left + Right + IsBaselineSuppression

Based on that, a global suppression is either a suppression of a specific target applying to all lefts and rights or a suppression of all targets applying to a specific left and right.

@safern hope it's okay to ask you if this matches your memory?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -67,11 +69,11 @@ public bool IsErrorSuppressed(Suppression error)
                 {
                     return true;
                 }
-                else if (error.DiagnosticId == null || error.DiagnosticId.StartsWith("cp", StringComparison.InvariantCultureIgnoreCase))
+                else if (error.DiagnosticId.StartsWith("cp", StringComparison.InvariantCultureIgnoreCase))
```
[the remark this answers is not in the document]
ViktorHofer: see my response above.

ViktorHofer:
@joperezr thanks for the great feedback. I tried to answer all of your questions and submitted https://github.com/dotnet/sdk/pull/25816 as a follow-up.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
joperezr: If having a global suppression (without diagnostic ID) wasn't supported before, that was not by design, and was likely a bug. As you can see on the discussion thread for this issue: https://github.com/dotnet/sdk/issues/22901 Having the ability to do global suppression for a left + right pair is something that people will want to have, so I would still push on the fact of reverting your change and have DiagnosticId be nullable again.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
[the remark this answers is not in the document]
joperezr: I've replied above as well, I think this should be reverted since, independently of the fact that this worked or not, having a global suppression for a tfm pair is desired.

src/Compatibility/Microsoft.DotNet.Compatibility.Shared/ValidatorBootstrap.cs
joperezr: Right, what you are saying is that it would be desired to be able to test one layer above and debug, so that could be done by adding scenario tests. Having a console project that isn't used at all (other than locally by someone) a) is prone to get broken as no one will execute it as part of the build, b) In this case it also complicates layering, c) it doesn't really add value for regression testing. If we instead were to add a scenario tests project that tested the whole layer and ran for each build, you would be adding extra protection to the code, while not having the protection be fragile and prone to break since it would be covered by CI.

joperezr:
Thanks for the replies and for pushing the follow-up PR, I will review that later today. Here are some replies from your global comments:

> @safern mentioned in the past that even though initially only an MSBuild Task is provided, a global tool is planned so that usages of the old ApiCompat tooling can't be migrated over when customers intentionally can't or don't want to rely on MSBuild.

This is true, but the tool was going to be ApiCompat tool specifically, as opposed to what you are adding which is a tool for the package validation layer. I think we still want the package validation layer to only be accessible via MSBuild.

> Yes I did generate the package locally and then use it in dotnet/runtime's build. 

Thanks for validating that. I do have to point out that it makes me nervous to read in your comments that we did change some crucial logic in this major refactoring which is removing some of the comparisons being made as we are now marking them as dupes. I would strongly suggest you to revert those changes back, and then making them again in standalone PRs so that a) they are easier to closely review, but most importantly b) they are easier to revert if they were to introduce a regression that we catch later. In general, I think it is a better pattern to avoid mixing functional changes with refactoring changes that span so many files, because, if the functional changes were to cause a regression, then reverting the whole PR would be nearly impossible and instead we would have to find a way of only reverting part of the PR. One of the reasons I'm particularly concerned about this happening, is that I know that we don't have enough test coverage here to catch potential problems with the logic that validates breaking changes with RID specific assets, and having a green build in dotnet/runtime doesn't necessarily mean that the functional changes were correct. In general, for functional changes, it is almost always better to have smaller-unit PRs, than bundling them with additional changes that are not closely related.

src/Compatibility/Microsoft.DotNet.PackageValidation/BaselinePackageValidator.cs
joperezr: I added a comment in the end touching this point, but I would prefer if the functional changes fixing these bugs happen in separate PRs, that could even optionally be tied to a specific issue calling out the problem. This makes it easier to track the fixes, review the changes, and gives you the option of reverting them easily in case those changes caused a regression.

src/Compatibility/Microsoft.DotNet.PackageValidation/Validators/CompatibleTFMValidator.cs
```diff
@@ -51,67 +48,67 @@ public void Validate(Package package)
...
                             Resources.NoCompatibleRidSpecificRuntimeAsset,
                             framework.ToString(),
                             rid);
                     }
-                    else
+                    // Invoke ApiCompat to compare the compile time asset with the runtime specific asset if they are not the same and
+                    // if the comparison hasn't already happened (when the runtime asset is the same as the runtime specific asset).
+                    else if (option.RunApiCompat &&
+                        compileTimeAsset.Path != runtimeRidSpecificAsset.Path &&
+                        (runtimeAsset == null || runtimeAsset.Path != runtimeRidSpecificAsset.Path))
```
ViktorHofer, line 94: @joperezr this is the only change in the PR that removed an apicompat runner work item because it was already queued (in L75) to avoid unnecessary work. I admit that this change should have been made in a separate PR and I will keep that in mind for future changes. That said, I don't think that this change should be reverted now and be brought back in a separate PR. If we would trace back a regression because of this change we can simply remove this line.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
ViktorHofer: My point is that we want to make sure that diagnostic id is always provided by compatibility tooling. I agree that a customer might want to globally suppress a left and a right (even though that raises interesting questions like why even compare something that is already suppressed).

I submitted https://github.com/dotnet/sdk/pull/25938 that updates the suppression engine to support this new type of global suppression. Support for what you are asking is added while still maintaining the desire to make sure that components need to provide a suppression id (so that suppressions are uniquely identifiable and suppressable => i.e. via NoWarn property).

src/Compatibility/Microsoft.DotNet.PackageValidation/Validators/CompatibleTFMValidator.cs
joperezr: Ok, as long as you are sure this is the only change I guess that is fine. My main point here is: what happens if there is a regression with this line that we discover in the future, we need to fix it, and you are not around? Tracking which line to revert would be a lot harder. Keeping functional changes in separate PRs with a lengthly description of the changes and dedicated review helps with this, since if a regression is found in the future and you (or whoever authors the change) are not around, then finding the culprit and knowing what to revert will be much easier.

Anyway, I'm fine with this instance being kept, but it would be good if for future changes we try to have several atomic and smaller PRs that are more specific to some change, than globing many changes on one single one 😉

PR #25816, opened 2022-06-03
Follow-Up on compatibility tooling refactoring

ViktorHofer:
Please review by commits. Every commit corresponds to one more PR feedbacks given in https://github.com/dotnet/sdk/pull/25765

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
@@ -49,7 +49,9 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
         }
 
         /// <summary>
-        /// It only makes sense to serialize IsBaselineSuppression when is true, if it is off, no need to have it on the file.
+        /// Only serialize the IsBaselineSuppression property when this is a baseline suppression. If it is not,
+        /// the property won't be serialized to keep the baseline file minimal. The method's name is important as
+        /// XmlSerializer will look for methods called ShouldSerializeX to determine if properties should be serialized.
         /// </summary>
         /// <returns>Returns true if IsBaselineSuppression should be serialized</returns>
         public bool ShouldSerializeIsBaselineSuppression() => IsBaselineSuppression;
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
@@ -32,6 +32,11 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
         /// </summary>
         public string? Right { get; set; }
 
+        /// <summary>
+        /// <see langword="true"/> if the suppression is to be applied to a baseline validation. <see langword="false"/> otherwise.
+        /// </summary>
+        public bool IsBaselineSuppression { get; set; }
+
         // Neccessary for XmlSerializer to instantiate an object of this class.
         private Suppression()
         {
@@ -44,11 +49,9 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
         }
 
         /// <summary>
-        /// <see langword="true"/> if the suppression is to be applied to a baseline validation. <see langword="false"/> otherwise.
+        /// It only makes sense to serialize IsBaselineSuppression when is true, if it is off, no need to have it on the file.
         /// </summary>
-        public bool IsBaselineSuppression { get; set; }
-
-        // It only makes sense to serialize IsBaselineSuppression when is true, if it is off, no need to have it on the file.
+        /// <returns>Returns true if IsBaselineSuppression should be serialized</returns>
         public bool ShouldSerializeIsBaselineSuppression() => IsBaselineSuppression;
 
         /// <inheritdoc/>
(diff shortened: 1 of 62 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
```diff
@@ -44,11 +49,9 @@ public Suppression(string diagnosticId)
         }
 
         /// <summary>
-        /// <see langword="true"/> if the suppression is to be applied to a baseline validation. <see langword="false"/> otherwise.
+        /// It only makes sense to serialize IsBaselineSuppression when is true, if it is off, no need to have it on the file.
```
joperezr, line 52: NIT: I would just add a quick blurb here explaining why is this method used and where, potentially pointing to xml docs. 

In short, this method is defined here because when the xml serializer is serializing this type, it uses reflection to call into this method in order to figure out whether or not the value of IsBaselineSuppression should be serialized. The naming here is also important, because the serializer will look for a method called: `ShouldSerialize<nameOfTheProperty>` and if it exists, it will call it and figure out if it has to serialize it or not. Anyway, just adding a small blurb explaining would be good so that it doesn't get removed or renamed by mistake in the future thinking that we aren't actually using it.

joperezr: I believe this is where the code is performing that check: https://github.com/dotnet/runtime/blob/c80a49d151b5e642be419067adf401a365b624b7/src/libraries/System.Private.Xml/src/System/Xml/Serialization/ReflectionXmlSerializationWriter.cs#L594-L599

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -91,10 +91,12 @@ public void RunApiCompat()
                             IsBaselineSuppression = _isBaselineSuppression
                         };
 
+                        // If the error is suppressed, don't log anything.
+                        if (_log.SuppressionEngine.IsErrorSuppressed(suppression))
```
joperezr, line 95: Does this work in the incremental case? AFAIR, we are rewriting the whole suppression file if it already existed everytime that you run this, so the reason why we would still log an error even when the error was already suppressed, was so that when the suppression file got overridden it would contain the error that was previously suppressed.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
```diff
@@ -25,8 +25,8 @@ public class ApiCompatRunner
         private readonly string _rightPackagePath;
         private readonly bool _isBaselineSuppression;
 
-        public ApiCompatRunner(bool enableStrictMode,
-            CompatibilityLoggerBase log,
+        public ApiCompatRunner(CompatibilityLoggerBase log,
+            bool enableStrictMode,
```
joperezr, line 29: NIT: Any particular reason for the param reordering?

src/Compatibility/Microsoft.DotNet.PackageValidation/Validators/PackageValidatorOption.cs
```diff
@@ -5,16 +5,50 @@
 
 namespace Microsoft.DotNet.PackageValidation.Validators
 {
-    public struct PackageValidatorOption
+    /// <summary>
+    /// A package validator option bag that is passed into package validators for configuration.
+    /// </summary>
+    public readonly struct PackageValidatorOption
```
joperezr, line 11: NIT: when you make the struct readonly, then I believe you don't need to have members explicitly say they are readonly, as this is done by the compiler.

src/Tasks/Common/BuildErrorException.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#nullable disable
```
joperezr, line 4: Do we have an issue tracking removing all of these?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Logging/SuppressionEngineTests.cs
```diff
@@ -51,14 +51,14 @@ public void SuppressionEngineCanParseInputSuppressionFile()
...
+            Assert.Throws<FileNotFoundException>(() => new SuppressionEngine("AFileThatDoesNotExist.xml"));
         }
 
         [Fact]
         public void SuppressionEngineDoesNotThrowOnEmptyFile()
         {
-            SuppressionEngine _ = new(suppressionFile: string.Empty);
-            _ = new SuppressionEngine(suppressionFile: "      ");
+            SuppressionEngine _ = new(suppressionsFile: string.Empty);
+            _ = new SuppressionEngine("      ");
```
joperezr, line 61: NIT: not a big deal, but why not also keep a named parameter here to match the above?

src/Tests/Microsoft.DotNet.PackageValidation.Tests/CompatibleFrameworksInPackageTests.cs
```diff
@@ -86,7 +86,7 @@ public class First
             var result = packCommand.Execute();
             Assert.Equal(string.Empty, result.StdErr);
             Package package = Package.Create(packCommand.GetNuGetPackage(), null);
-            new CompatibleFrameworkInPackageValidator(_log).Validate(new() { Package = package });
+            new CompatibleFrameworkInPackageValidator(_log).Validate(new(package));
```
joperezr, line 89: NIT: I know you didn't add these here, but since you are already editing these lines, can we type explicitly the type for which we are invoking the constructor? We usually are fine with implicitly just using `new()` whenever you can infer the type from the left operand, like `Dictionary<int, int> d = new();` but in cases like this it makes the code harder to read as one would need to know what Validate() takes as a parameter.

joperezr (approved):
Left a few Nits to be addressed but looks good otherwise. Please also consider reverting in this PR the functional changes made on the previous PR on the validators as per my comment in your other PR.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
ViktorHofer: Yes it does. The logger implementations immediately return if the error is suppressed: https://github.com/dotnet/sdk/blob/cf5ff15fd295f02cef652cf5e9a1c9879a0615bb/src/Compatibility/Microsoft.DotNet.Compatibility.Console/ConsoleCompatibilityLogger.cs#L26-L27. I didn't change that with this or my previous change.

Correct. The suppression file is (re-)written if the `GenerateCompatibilitySuppressionFile` setting is true and if a suppression file path is passed in. Its previous content is ignored as it's always regenerated.

If `IsErrorSuppressed` returns true, then that means that the error/suppression is either already part of the suppression backing storage or that a `NoWarn` setting was passed in to ignore the error. The `WriteSuppressionsToFile` serializes the backing storage to an xml file.

I do think this works as intended and as mentioned, this change shouldn't mutate the existing behavior. This makes sure that. This change avoids the LogError call by frontloading the IsErrorSuppressed check that already exists for the header logging condition.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunner.cs
ViktorHofer: Yes, I'm locally experimenting with leveraging an IoC container to better test components like the ApiCompatRunner without relying on full integration testing and specifying the type dependencies first, aids to that scenario. I will talk more about this in the meeting that I sent out for next Monday.

Issue #25770, opened 2022-06-01
Add an `-lp` option alias to `dotnet run` for `--launch-profile`


DamianEdwards:
As part of [EPIC] Revisiting HTTPS defaults in ASP.NET Core

ASP.NET Core projects are being updated to have multiple launch profiles by default. As such, we'd like it to be easier to select a launch profile when running the project using `dotnet run`.

We propose supporting `-lp` as an option alias for `--launch-profile`, e.g.:

```
$ dotnet run -lp https
```

PR #25832, opened 2022-06-03
Add support for specifying launch profile to `dotnet watch run`

### LINKED ISSUE #41988: Add support for specifying a launch profile to `dotnet watch run`
As part of [[EPIC] Revisiting HTTPS defaults in ASP.NET Core](https://github.com/dotnet/aspnetcore/issues/41990)

ASP.NET Core projects are being updated to have multiple launch profiles by default. As such, we'd like it to be easier to select a launch profile when running the project using `dotnet watch`/`dotnet watch run`.

We propose supporting `-lp` as an option alias for `--launch-profile`, e.g.:

```
$ dotnet watch -lp https
```

Matching issue for `dotnet run`: dotnet/sdk#25770

DamianEdwards:
Fixes dotnet/aspnetcore#41988

As part of [[EPIC] Revisiting HTTPS defaults in ASP.NET Core](https://github.com/dotnet/aspnetcore/issues/41990)

ASP.NET Core projects are being updated to have multiple launch profiles by default. As such, we'd like it to be easier to select a launch profile when running the project using `dotnet watch`/`dotnet watch run`.

Add support for `-lp` as an option alias for `--launch-profile`, e.g.:

```
$ dotnet watch -lp https
```

or

```
$ dotnet watch --launch-profile https
```

Matching issue for `dotnet run`: dotnet/sdk#25770

### DIFF
--- a/src/BuiltInTools/dotnet-watch/DotNetWatchContext.cs
+++ b/src/BuiltInTools/dotnet-watch/DotNetWatchContext.cs
@@ -26,7 +26,7 @@ namespace Microsoft.DotNet.Watcher.Tools
 
         public BrowserRefreshServer? BrowserRefreshServer { get; set; }
 
-        public LaunchSettingsProfile DefaultLaunchSettingsProfile { get; init; } = default!;
+        public LaunchSettingsProfile LaunchSettingsProfile { get; init; } = default!;
 
         public ProjectGraph? ProjectGraph { get; set; }
     }
--- a/src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
+++ b/src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
@@ -18,13 +18,15 @@ namespace Microsoft.DotNet.Watcher.Tools
 
         public string? CommandName { get; set; }
 
+        // TODO: Add support for commandLineArgs defined in the launch profile
+
         public bool LaunchBrowser { get; set; }
 
         public string? LaunchUrl { get; set; }
 
         public IDictionary<string, string>? EnvironmentVariables { get; set; }
 
-        internal static LaunchSettingsProfile? ReadDefaultProfile(string projectDirectory, IReporter reporter)
+        internal static LaunchSettingsProfile? ReadLaunchProfile(string projectDirectory, string? launchProfileName, IReporter reporter)
         {
             var launchSettingsPath = Path.Combine(projectDirectory, "Properties", "launchSettings.json");
             if (!File.Exists(launchSettingsPath))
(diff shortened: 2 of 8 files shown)


src/BuiltInTools/dotnet-watch/DotNetWatchContext.cs
```diff
@@ -26,7 +26,7 @@ public class DotNetWatchContext
 
         public BrowserRefreshServer? BrowserRefreshServer { get; set; }
 
-        public LaunchSettingsProfile DefaultLaunchSettingsProfile { get; init; } = default!;
```
timheuer, line 29: Is this API callable other ways and would be considered a BC for anyone else?

DamianEdwards: I don't believe we consider changes to public APIs in tools like this a breaking change. Certainly we don't have any of the usual public API tracking infrastructure in place for the tools that I know of. The biggest concern would be if something in VS is loading the tool assembly directly but I'd hope we're not doing that as it makes the dependency extremely unclear (from this repo's POV).

src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
```diff
@@ -45,11 +47,32 @@ public class LaunchSettingsProfile
...
             }
 
+            if (string.IsNullOrEmpty(launchProfileName))
+            {
+                // Load the default (first) launch profile
+                return ReadDefaultLaunchProfile(launchSettings, reporter);
+            }
+
+            // Load the specified launch profile
+            var namedProfile = launchSettings?.Profiles?.FirstOrDefault(f => string.Equals(f.Key , launchProfileName, StringComparison.OrdinalIgnoreCase)).Value;
```
DamianEdwards, line 57:
```suggestion
            var namedProfile = launchSettings?.Profiles?.FirstOrDefault(f => string.Equals(f.Key , launchProfileName, StringComparison.Ordinal)).Value;
```
Turns out this is case-sensitive in `dotnet run` so it should be here too.

src/BuiltInTools/dotnet-watch/DotNetWatchContext.cs
javiercn: This is only used by the command-line. We don't publish this as a package, so we should be good. In general, anything that we share with Visual Studio has an explicit agreement that defines it.

@BillHiebert can confirm here.

src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
```diff
@@ -18,13 +18,15 @@ public class LaunchSettingsProfile
 
         public string? CommandName { get; set; }
 
+        // TODO: Add support for commandLineArgs defined in the launch profile
```
javiercn, line 21: You can do this today?

src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
```diff
@@ -45,11 +47,32 @@ public class LaunchSettingsProfile
...
             }
 
+            if (string.IsNullOrEmpty(launchProfileName))
+            {
+                // Load the default (first) launch profile
+                return ReadDefaultLaunchProfile(launchSettings, reporter);
+            }
+
+            // Load the specified launch profile
+            var namedProfile = launchSettings?.Profiles?.FirstOrDefault(f => string.Equals(f.Key , launchProfileName, StringComparison.Ordinal)).Value;
```
javiercn, line 57:
```suggestion
            var namedProfile = launchSettings?.Profiles?.FirstOrDefault(f => string.Equals(f.Key , launchProfileName, StringComparison.OrdinalIgnoreCase)).Value;
```
As this is something that the user types and I don't believe there's going to be many people (if any) with profiles differing only on casing.

javiercn (approved):
Looks pretty solid, just one small suggestion.

Thanks @DamianEdwards!

src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
DamianEdwards: Actually, after testing it locally it seems only Visual Studio honors it and not `dotnet run`.

src/BuiltInTools/dotnet-watch/LaunchSettingsProfile.cs
DamianEdwards: I had that originally but it turns out `dotnet run` is case sensitive, so we'd have to change it both places, likely to first do an ordinal search, then fallback to case-insensitive if not found.

javiercn: Given that dotnet run is case sensitive and we have a warning below I'm ok if we leave it as is. Might be worth pointing about the casing difference in the warning. It can be as simple as saying "Did you mean (suggestion)?" when we find a case insensitive match

DamianEdwards: Yeah I like that idea.

DamianEdwards: image

PR #25834, opened 2022-06-03
Add --skip-sign-check option to repair, uninstall, and update

joeloff:
Currently, only the install command supports `--skip-sign-check` option. Other commands can also interact with downloading packages and should be able to turn off the check consistently


nagilson (approved):
Looks good to me, would appreciate feedback from someone else with further context who can confirm.

marcpopMSFT:
Check if this should go into .4xx instead (ie has maui moved forward in their SDK usage).

Issue #23136, opened 2021-12-22
'dotnet run' command should fail with unknow arguments


jmarolf:
##  Description

If someone forgets to put `--project` in front of a path to a csproj file when executing the `dotnet run` command this file path is added as a third argument to the program.

## Repro steps:
1. `dotnet new console --name consoleApp`
1. add the following to `Program.cs`
```csharp
using static System.Console;

WriteLine($"'{args.Length}' arguments passed to program");

for (int i = 0; i < args.Length; i++) {
    WriteLine($"[{i}]: '{args[i]}'");
}
```
1. execute `dotnet run .\consoleApp.csproj -- --list-files nddemo.txt`

## Expected:

I think there are two reasonable things that `dotnet run` could do.

### 1. Implicitly swallow the csproj argument

See that an arg that is given before `--` maps to a project file whose path is identical to the one found in the directory and elide this argument.  Meaning that all three of these commands are equivalent if there is a `consoleApp.csproj` in the working directory:
    - `dotnet run --project .\consoleApp.csproj -- --list-files nddemo.txt`
    - `dotnet run .\consoleApp.csproj -- --list-files nddemo.txt`
    - `dotnet run -- --list-files nddemo.txt`

Running either of these three commands would produce this output for the program given in the repo steps.

```cmd
> dotnet run .\consoleApp.csproj -- --list-files nddemo.txt
'2' arguments passed to program
[0]: '--list-files'
[1]: 'nddemo.txt'
```

### 2. Error stating that there is an unknown argument

Any arguments that are not know and come before a `--` argument should be errors and not silently passed on as additional arguments. So the output in the repo steps should be:

```cmd
> dotnet run .\consoleApp.csproj -- --list-files nddemo.txt
Error: unknown argument '.\consoleApp.csproj '
if you intend to pass this as an additional argument to the program being run place it after '--'
```

## Actual:
```cmd
> dotnet run .\consoleApp.csproj -- --list-files nddemo.txt
'3' arguments passed to program
[0]: '.\consoleApp.csproj'
[1]: '--list-files'
[2]: 'nddemo.txt'
```

baronfel:
Completely agree @jmarolf, my preference would be towards eliding the project file if present. This might require some massaging of UnparsedTokens in System.CommandLine parlance, but the UX is worth it, and passing any parameters on the left of the `--` to the invoked program is just wrong.

KalleOlaviNiemitalo:
[...]

PR #25854, opened 2022-06-05
[24068] Remove Legacy -- behavior and use of UnparsedTokens

Tunduk:
@baronfel It was a little bit tricky. 
Honestly, I dont't like how `System.CommandLine` works without double dash legacy behavior. It treats  double dash arguments as command arguments but it's not true. Double dash arguments should be separated from command arguments. It's impossible to create something like `dotnet vstest` with `System.CommandLine` now. 
I wanted to remove `ParseResultExtensions.GetArguments` because I don't think that we should read\modify raw values from command line. We have a whole library for that. But it can't be implemented now. 
So, I think this PR will resolve just a part of issue. It'll allow `System.CommandLine` to move forward, but it won't fix all problems in SDK 
What do you think about it?

Also fixed  some unrelated staff: 

- Fixed`dotnet test` argument arity. It should be ZeroOrOne by documentation.
- Fixed `dotnet format`  help output with subcommands. 

Thanks in advance for your review.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -26,41 +26,41 @@
       <Uri>https://github.com/dotnet/templating</Uri>
       <Sha>c988d74e251a9e04272b671aedaccd12eb527291</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NETCore.App.Ref" Version="7.0.0-preview.6.22303.5">
+    <Dependency Name="Microsoft.NETCore.App.Ref" Version="7.0.0-preview.6.22304.5">
       <Uri>https://github.com/dotnet/runtime</Uri>
-      <Sha>c80a49d151b5e642be419067adf401a365b624b7</Sha>
+      <Sha>bad2ab053c53f68a9b7b8ed7d0867fc58261bc9a</Sha>
     </Dependency>
-    <Dependency Name="VS.Redist.Common.NetCore.SharedFramework.x64.7.0" Version="7.0.0-preview.6.22303.5">
+    <Dependency Name="VS.Redist.Common.NetCore.SharedFramework.x64.7.0" Version="7.0.0-preview.6.22304.5">
       <Uri>https://github.com/dotnet/runtime</Uri>
-      <Sha>c80a49d151b5e642be419067adf401a365b624b7</Sha>
+      <Sha>bad2ab053c53f68a9b7b8ed7d0867fc58261bc9a</Sha>
     </Dependency>
-    <Dependency Name="VS.Redist.Common.NetCore.TargetingPack.x64.7.0" Version="7.0.0-preview.6.22303.5">
+    <Dependency Name="VS.Redist.Common.NetCore.TargetingPack.x64.7.0" Version="7.0.0-preview.6.22304.5">
       <Uri>https://github.com/dotnet/runtime</Uri>
-      <Sha>c80a49d151b5e642be419067adf401a365b624b7</Sha>
+      <Sha>bad2ab053c53f68a9b7b8ed7d0867fc58261bc9a</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NETCore.App.Runtime.win-x64" Version="7.0.0-preview.6.22303.5">
+    <Dependency Name="Microsoft.NETCore.App.Runtime.win-x64" Version="7.0.0-preview.6.22304.5">
       <Uri>https://github.com/dotnet/runtime</Uri>
-      <Sha>c80a49d151b5e642be419067adf401a365b624b7</Sha>
+      <Sha>bad2ab053c53f68a9b7b8ed7d0867fc58261bc9a</Sha>
     </Dependency>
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=25854) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=25854) before we can accept your contribution.<br/><br/>:x: Tunduk [sign now](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=25854)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/sdk?pullRequest=25854) it.</sub>

jmarolf (commented):
dotnet-format section looks good

baronfel:
This looks great! I'll dig into test specifically tomorrow. Is there anything specific you saw that can't be solved by repeatedly gathering args via `--` and re-parsing that subset of args?

Tunduk:
> Is there anything specific you saw that can't be solved by repeatedly gathering args via `--` and re-parsing that subset of args?

I'm not sure that I understand your question. Do you mean hack like in `src/Cli/dotnet/commands/dotnet-test/Program.cs:127` for separation command argument from `--` arguments?

src/Cli/dotnet/Parser.cs
```diff
@@ -269,6 +268,15 @@ public override void Write(HelpContext context)
...
                 }
+                else if (command.Name.Equals(FormatCommandParser.GetCommand().Name))
+                {
+                    var argumetns = context.ParseResult.GetValueForArgument(FormatCommandParser.Arguments);
+                    new DotnetFormatForwardingApp(argumetns.Concat(helpArgs).ToArray()).Execute();
+                }
+                else if (command.Name.Equals(FsiCommandParser.GetCommand().Name))
+                {
+                    new FsiForwardingApp(helpArgs).Execute();
+                }
```
baronfel, line 279: This is great for this PR (and nothing that you;ve done incorrectly), but the more I see this the more I want to extract all of these to explicit subcommands of `System.CommandLine.Command`, so that they could override their individual help implementations/potentially do completions/etc. Making a note to myself here :)

baronfel:
Ok I spent a good chunk of time on this yesterday and I'm really happy with it, excellent work! The problem of the test command specifically stuck with me, though, and I chatted with @jonsequitur about how we should model structures like the test command.

A refresher:

`dotnet test  [zero or one sln or project argument] [<tons of options>] -- [zero or more test application arguments] [zero or more test runner arguments]`

We can make use of custom parsing delegates to handle these three distinct arguments in a more general way, without really running afoul of `--` parsing semantics.  I've started a branch based off of this PR at [my fork](https://github.com/tunduk/sdk/compare/main...baronfel:test-parser-tweaks?expand=1) where I dug into what this looks like, and I think it has a lot of promise.  While prototyping this, however, I ran into https://github.com/dotnet/command-line-api/issues/1759. Once that's fixed I think we can actually do this!

The up-side of making custom parsers for these three arguments is
* codifying the requirements on the parsing side, which means we can get nicer parsing errors for users
* reducing the amount of custom logic during command execution
* getting more accurate help output for these kinds of arguments

baronfel (changes requested):
Putting 'request changes' on this just to prevent accidental merges.

@Tunduk I am very happy with this overall, I just think we have to hold it until the linked System.CommandLine issue is resolved so that my changes to dotnet-test can be part of it as well.  I've talked to the maintainers there and they're looking into it.

Thank you so much for this - I'm very impressed by the work, and I'm very glad that you took the time to get into the details!

src/Cli/dotnet/Parser.cs
Tunduk: I agree with you. 
But don't you think that it's hidden dependency? I mean a situation when DotnetFormat team add new subcommand and SDK team should add this subcommand too for correct help/completition. 
If we could deliver DotnetFormat via git submodule/nuget I think it will solve this situation.

Tunduk:
@baronfel Sorry for long responce. Thanks a lot for your review! It was very interesting to take part in SDK project. I think I'll continue contribute in Micsrosoft projects. 

As I can see System.CommandLine issue was resolved should I do anything else with this PR?

baronfel:
@Tunduk can you rebase your branch so that it's more up to date with `dotnet/sdk:main`? I'm ready to send a PR to your branch but because main here has advanced it looks like more changes than it really is.

Tunduk:
@baronfel Hi!
There is still the [issue](https://github.com/dotnet/command-line-api/issues/1779) in `System.CommandLine` that blocks using custom parser feature. 
Could we postpone custom parser changes? It'll allow System.CommandLine team start work on removing legacy double dash behavior. Also, I have seen some issues that await this PR. Such as [this](https://github.com/dotnet/sdk/issues/23136)
What do you think about it?

baronfel:
So @Tunduk at this point I think you can revert/reset to remove my commits entirely, and then we should be able to take this.

PR #25905, opened 2022-06-07
Explain the Purpose of Dogfood in The Dev Guide

nagilson:
I was initially confused because I did not read the main page when trying to initially build the SDK. This provides context for new developers as to why they are running dogfood and would've saved a lot of time for @joeloff (asking him about my initial confusions) and me if it were initially included.

### DIFF
--- a/documentation/project-docs/developer-guide.md
+++ b/documentation/project-docs/developer-guide.md
@@ -65,7 +65,14 @@ Run the following command from the root of the repository to run all the .NET Co
 
 The `dotnet` executable in the artifacts directory can be run directly.
 
-However, it's easier to configure a test environment to run the built `dotnet`.
+However, it's easier to configure a test environment to run the built `dotnet`. This test environment is managed by dogfood. 
+The dogfood script starts a new Powershell with the environment configured to redirect SDK resolution to your build.
+
+From that shell your SDK will be available in:
+
+- any Visual Studio instance launched (via `& devenv.exe`)
+- `dotnet build`
+- `msbuild`
 
 ### Windows
 


joeloff (approved):
Does the rest of the doc mention the script's location under `eng\dogfood,cmd`? If not, maybe include that. I don't recall if we have anything similar for non-Windows platforms, if not, it might be worth mentioning that too.

nagilson:
> Does the rest of the doc mention the script's location under `eng\dogfood,cmd`? If not, maybe include that. I don't recall if we have anything similar for non-Windows platforms, if not, it might be worth mentioning that too.

It does, it also provides instructions for Linux and MacOS platforms, so we're good there. Merging; thank you for your review. 👍

documentation/project-docs/developer-guide.md
```diff
@@ -65,7 +65,14 @@ Run the following command from the root of the repository to run all the .NET Co
 
 The `dotnet` executable in the artifacts directory can be run directly.
 
-However, it's easier to configure a test environment to run the built `dotnet`.
+However, it's easier to configure a test environment to run the built `dotnet`. This test environment is managed by dogfood. 
+The dogfood script starts a new Powershell with the environment configured to redirect SDK resolution to your build.
+
+From that shell your SDK will be available in:
```
gkulin, line 71: it is also available in the command prompt environment where you ran the dogfood script, right?

nagilson: Perhaps, though I didn't see it when looking through the commands. the artifacts build and dogfood are both silent, though I think that is a good thing

gkulin: I just mean that after running the dogfood command you can run things from the command line (example : `dotnet workload update` ) using whatever version of the sdk you have built. Maybe that's obvious to some people but if we are explaining the purpose of dogfood, it's probably worth calling out

PR #25938, opened 2022-06-09
Support suppressions without a diagnostic id

ViktorHofer:
Fixes [what @joperezr identified as a bug in the suppression engine]( https://github.com/dotnet/sdk/pull/25765#discussion_r889313428).

Any component in the Compatibility tooling that creates/logs a
suppression, needs to provide a suppression id, in order for suppressions
to be uniquely identifyable and suppressable (ie via NoWarn). Therefore
the suppression id is non nullable in the Suppression type.

A customer can provide a suppression file that contains a suppression
without a diagnostic id. When that is being deserialized, a Suppression
object with an empty diagnostic id is created. This change updates the
suppression engine to support this new form of global suppression, for
which only a left and right (+ IsBaselineSuppression) is passed in.

Adding tests that verify that this added global suppression type is
respected and correctly roundtrips.

### DIFF
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
@@ -53,8 +53,14 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
         /// the property won't be serialized to keep the baseline file minimal. The method's name is important as
         /// XmlSerializer will look for methods called ShouldSerializeX to determine if properties should be serialized.
         /// </summary>
-        /// <returns>Returns true if IsBaselineSuppression should be serialized</returns>
+        /// <returns>Returns true if IsBaselineSuppression should be serialized.</returns>
         public bool ShouldSerializeIsBaselineSuppression() => IsBaselineSuppression;
+
+        /// <summary>
+        /// Same as the above, this method is used to determine if the DiagnosticId property should be serialized.
+        /// </summary>
+        /// <returns>Returns true if DiagnosticId shoudl be serialized.</returns>
+        public bool ShouldSerializeDiagnosticId() => DiagnosticId != string.Empty;
 
         /// <inheritdoc/>
         public bool Equals(Suppression? other)
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
@@ -61,24 +61,25 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
         /// <returns><see langword="true"/> if the error is already suppressed. <see langword="false"/> otherwise.</returns>
         public bool IsErrorSuppressed(Suppression error)
         {
+            // An error is suppressed if
             _readerWriterLock.EnterReadLock();
             try
             {
-                if (_noWarn.Contains(error.DiagnosticId))
+                // An error is suppressed if there's a direct match in the suppression backing store or in the NoWarn property
+                if (_noWarn.Contains(error.DiagnosticId) || _validationSuppressions.Contains(error))
                 {
                     return true;
                 }
-                else if (_validationSuppressions.Count > 0)
+                // If the suppression backing store doesn't contain any items or the error is a package validation error for which a left and right doesn't exist,
+                // don't attempt to find a global suppression to avoid unnecessary Suppression allocations.
+                else if (_validationSuppressions.Count > 0 && !error.DiagnosticId.StartsWith("pkv", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    // An error is suppressed if
-                    // 1. There's a direct match in the suppression backing store
-                    return _validationSuppressions.Contains(error) ||
-                    // 2. The DiagnosticId, Target and IsBaselineSuppression values match
-                        _validationSuppressions.Contains(new Suppression(error.DiagnosticId) { Target = error.Target, IsBaselineSuppression = error.IsBaselineSuppression }) ||
-                    // 3. The DiagnosticId, Left, Right and IsBaselineSupression values match
+                    // The DiagnosticId, Target and IsBaselineSuppression values match
+                    return _validationSuppressions.Contains(new Suppression(error.DiagnosticId) { Target = error.Target, IsBaselineSuppression = error.IsBaselineSuppression }) ||
+                    // The DiagnosticId, Left, Right and IsBaselineSupression values match
                         _validationSuppressions.Contains(new Suppression(error.DiagnosticId) { Left = error.Left, Right = error.Right, IsBaselineSuppression = error.IsBaselineSuppression }) ||
-                    // 4. The Left, Right and IsBaselineSuppression values match. Any component that logs a suppressable warning or error must provide
-                    //    a DiagnosticId in order for NoWarn to work. Only suppression files can contain suppressions without diagnostic ids.
+                    // The Left, Right and IsBaselineSuppression values match. Any component that logs a suppressable warning or error must provide
+                    // a DiagnosticId in order for NoWarn to work. Only suppression files can contain suppressions without diagnostic ids.
                         _validationSuppressions.Contains(new Suppression(string.Empty) { Left = error.Left, Right = error.Right, IsBaselineSuppression = error.IsBaselineSuppression });
                 }
 
--- a/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
+++ b/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
@@ -62,7 +62,6 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
         /// <returns><see langword="true"/> if the error is already suppressed. <see langword="false"/> otherwise.</returns>
         public bool IsErrorSuppressed(Suppression error)
         {
-            // An error is suppressed if
             _readerWriterLock.EnterReadLock();
             try
             {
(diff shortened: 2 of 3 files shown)


src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -43,7 +44,7 @@ public SuppressionEngine(string? suppressionsFile = null, string ? noWarn = null
         /// <param name="left">Optional. The left operand in a APICompat error.</param>
         /// <param name="right">Optional. The right operand in a APICompat error.</param>
         /// <returns><see langword="true"/> if the error is already suppressed. <see langword="false"/> otherwise.</returns>
-        public bool IsErrorSuppressed(string diagnosticId, string? target, string? left = null, string? right = null, bool isBaselineSuppression = false)
+        public bool IsErrorSuppressed(string diagnosticId, string? target = null, string? left = null, string? right = null, bool isBaselineSuppression = false)
```
joperezr, line 47: I think we actually want diagnosticId to be nullable in order to fully support a roundtrip. Here are my thoughts on this: What happens if a customer decides to to a global suppression between a left and a right pair (meaning he doesn't use diagnosticID) but then wants to run the engine again and regenerate the suppression file? If I'm understanding correctly, with these changes a diagnosticId property will be created with string.Empty which is not desired.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -55,24 +56,32 @@ public bool IsErrorSuppressed(string diagnosticId, string? target, string? left
...
 
         /// <summary>
-        /// Checks if the passed in error is suppressed or not.
+        /// Checks if the passed in error is suppressed.
         /// </summary>
         /// <param name="error">The <see cref="Suppression"/> error to check.</param>
         /// <returns><see langword="true"/> if the error is already suppressed. <see langword="false"/> otherwise.</returns>
         public bool IsErrorSuppressed(Suppression error)
         {
+            // An error is suppressed if
```
joperezr, line 65: I suppose this incomplete comment should be removed, as I'm guessing the comment bellow this replaced it

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Logging/SuppressionEngineTests.cs
```diff
@@ -110,6 +110,36 @@ public void SuppressionEngineSupportsGlobalCompare()
...
+        [Fact]
+        public void CtorSuppressionWithoutDiagnosticIdRoundtrip()
+        {
+            SuppressionEngine engine = new();
+            // Explicitly add a suppression without a diagnostic id for testing.
+            Suppression suppression = new(string.Empty) { Left = "ref/net6.0/A.dll", Right = "lib/net6.0/A.dll", IsBaselineSuppression = true };
+            string filePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName());
+
+            engine.AddSuppression(suppression);
+            Assert.True(engine.WriteSuppressionsToFile(filePath));
```
joperezr, line 138: Another interesting bit to test here with roundtripping, is if there was a suppression file already which was globally suppressed (no diagnostic id) and then we re-write the suppressions file, did  it roundtrip exactly? or did it add an empty property?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs
```diff
@@ -53,9 +53,15 @@ public Suppression(string diagnosticId)
...
         /// </summary>
-        /// <returns>Returns true if IsBaselineSuppression should be serialized</returns>
+        /// <returns>Returns true if IsBaselineSuppression should be serialized.</returns>
         public bool ShouldSerializeIsBaselineSuppression() => IsBaselineSuppression;
 
+        /// <summary>
+        /// Same as the above, this method is used to determine if the DiagnosticId property should be serialized.
+        /// </summary>
+        /// <returns>Returns true if DiagnosticId shoudl be serialized.</returns>
+        public bool ShouldSerializeDiagnosticId() => DiagnosticId != string.Empty;
```
joperezr, line 63: Although this would work, an easier way to do it would be to just mark DiagnosticId as nullable. In IsBaselineSuppression we had to take this approach because the value is a boolean (value type) which cannot be marked as null, but for nullable types like string this is not required and that is why we don't need this for the rest of the string properties like left, right, or target. If we have found a valid scenario where DiagnosticId might be missing, why not just mark that property as nullable as well? It feels weird to have different behavior between properties like target and diagnosticId if technically both could be missing.

joperezr: Note: XmlSerializer already will not serialize properties which value is null. In your case it was doing it and required this ShouldSerialize method because you were setting it to string.Empty.

ViktorHofer: IMHO DiagnosticId should never be null, absolutely never. Doing so is a very special scenario as it disables all existing and future api comparison checks. If a customer explicitly decides to ignore all errors for a left right tuple, then that's their decision but as already mentioned it raises the question why we then even perform api comparisons that are completely ignored.

As expressed in other discussions, I'm ok with supporting this special scenario as long as it doesn't impact the common logic. I intentionally chose an implementation that satisfies the desire to ignore a left right tuple without giving up treating DiagnosticId as non nullable.

It would be interesting to know what others think about this? @ericstj @smasher164

joperezr: > IMHO DiagnosticId should never be null, absolutely never. 

I don't understand how this is different to TargetId. Just as there are global suppressions for a diagnostic id, it should be possible to do it for just a left and a right for all diagnostic ids.

> Doing so is a very special scenario as it disables all existing and future api comparison checks. If a customer explicitly decides to ignore all errors for a left right tuple, then that's their decision but as already mentioned it raises the question why we then even perform api comparisons that are completely ignored.

So here is a (common) scenario. People ship a package version 1.0 with a ns2.0 and nca2.0 assemblies. They then ship a  2.0.0 version that has a nca2.0 and net7.0 assemblies. This means they dropped ns2.0 support, which will cause them to get some pkv errors, but most importantly, when performing baseline validation it also means that there will be a lot of comparisons made between the 1.0.0 ns2.0 asset, and the nca2.0 and net7.0 assets of the new package. If they don't care about any breaks from that ns2.0 asset, It should be possible for them to create a global suppression that just lists (left=1.0.0 ns2.0 asset,  right= 2.0.0 net7.0) and have this suppress all diagnostics for differences found there, but they might still care about the other pairs. To me DiagnosticId is just another field, that in terms of global suppression, it should be able to apply for all diagnostic ids just as the rest of the properties.
When we internally create a Suppression, then I agree with you, DiagnosticId should never be null, but that is exactly the same case for Target, when we create one we would want that to not be null, but we do mark it as nullable today in order to allow global suppressions. In short, I don't see what is the difference between diagnosticid and target.

ViktorHofer: > So here is a (common) scenario

In your scenario above, rather than silencing baseline errors, I would instead want to tell package validation to not select this asset at all so that unnecessary comparisons aren't performed.

What was originally being asked by @dotMorten in https://github.com/dotnet/sdk/issues/22901#issuecomment-1139114195 was if it's possible to override the chosen baseline. While a suppression has the same end-result, the comparisons are still performed. Just in the example of a repo with hundreds of projects like dotnet/runtime, unnecessary comparisons are slowing down the overall build noticeably, especially with multiple lefts and multiple rights.

> When we internally create a Suppression, then I agree with you, DiagnosticId should never be null, but that is exactly the same case for Target

Then Target shouldn't be nullable either. This very much sounds like we are conflating two concepts of suppressions: exact match vs global. I want to be able to utilize the nullability compiler feature to express that our internal logic requires the minimum set of suppression data (DiagnosticId and presumably Target) when emitting a suppress able error, while customers can still hand-edit their suppression files and create global suppressions by omitting these fields. The PR makes that possible.

If we would change how global suppressions are encoded in suppression files, then I would rather introduce an additional type `GlobalSuppression` in order to correctly annotate the internal representation of the two different kinds of suppressions.

joperezr: >  I want to be able to utilize the nullability compiler feature to express that our internal logic requires the minimum set of suppression data (DiagnosticId and presumably Target) when emitting a suppress able error, while customers can still hand-edit their suppression files and create global suppressions by omitting these fields. The PR makes that possible.

That is not what nullable feature is for. The point of nullable annotations, is to try to prevent NRE from happening in your code. If there is a scenario where a particular property can return null (even if that scenario is global suppressions as opposed to our creation of them) and we mark that property as non-nullable, then you have a potential point in your code where NRE could happen. If what you want is to ensure that our uses in the code don't create these things with their values as null, you can do that by adding `Debug.Assert(targetId != null)` in the call sites, but marking a property non-nullable when it technically is possible to be null, then you will have a problem, as there are callsites that might not check if this property is null (as the compiler won't create a warning) and if that was a global suppression then you will get a runtime NRE. Just as an example of this, when we first introduced nullable annotations and begin adding them into APIs, if an API had a edge case (no matter how likely that was) where it could return null, then we would mark the return type as nullable, even when the 90% case would actually return non null.

ViktorHofer: In this case it's the opposite. DiagnosticId can never be null as if it isn't provided via the suppression file by the user, [`XmlSerializer` will initialize it to an empty string](https://github.com/dotnet/sdk/blob/main/src/Compatibility/Microsoft.DotNet.ApiCompatibility/Logging/Suppression.cs#L41-L44).

Aside from that, what I mentioned above might make also make sense: differentiating the different types of suppressions in the suppression file (global vs exact). The problem with that is that it would customers that already depend on global suppressions in their compatibility files.

PR #25940, opened 2022-06-09
[release/6.0.4xx] Don't retarget testhost 

nohwnd:
Testhost is shipped together with multiple runtimeconfigs to be used as fallback when user does not provide runtimeconfig of their own, but tfm can be detected from the assembly. Rewriting them makes the tested assembly use the current runtime version, which is wrong.

Fix https://github.com/dotnet/sdk/issues/24769

Cherry picked from https://github.com/dotnet/sdk/pull/24770

### DIFF
--- a/src/Layout/redist/targets/GenerateLayout.targets
+++ b/src/Layout/redist/targets/GenerateLayout.targets
@@ -409,8 +409,9 @@
     </PropertyGroup>
     <ItemGroup>
       <!-- Exclude F# from retargeting: https://github.com/dotnet/toolset/issues/2866 -->
+      <!-- Exclude testhost from retargeting: https://github.com/dotnet/sdk/issues/24769 -->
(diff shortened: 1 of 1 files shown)


Evangelink (approved):

MarcoRossignoli (approved):

marcpopMSFT:
@nohwnd is this change needed for 6.0.4xx as we enter M2 mode this afternoon.

marcpopMSFT:
Also, will https://github.com/dotnet/installer/blob/a37a7a7d064815641a17f87a80a371cc566a8fdb/src/redist/targets/GenerateLayout.targets#L459 also need to be updated with an exclusion?

nohwnd:
@marcpopMSFT  thanks for the tip. Will close this PR, and fix it in net7, it is a fallback of a fallback and it was broken for a long time and there are 0 issues reported.

PR #25962, opened 2022-06-10
Check capitalization for RID

Forgind:


joeloff:
I think we can close this PR. the proper fix is in the WAP publishing wizard in Visual Studio. The RID values are used throughout the SDK and will potentially create other issues. All the publishing wizards in VS construct the RIDs according to the known catalog names + casing and on case sensitive file systems, allowing something like win-ARM64 and win-arm64 could create problems if those appear in output folders as you'd have side-by-side directories. WAP wizard currently picks the platform from the solution configuration when triggering a solution build so that's where the change needs to be.

PR #25926, opened 2022-06-08
Update the default to Release for Pack/Pub in command building

nagilson:
Fix for https://github.com/dotnet/sdk/issues/23551#issuecomment-1150204351. Drafting until unit tests are added to assure this works. Note that adding a default through System.Commandline adds [Default: Release] to the description and help text. So we are cutting the unnecessary duplicated sentences

Issue #23551, opened 2022-01-23
Enable publishing as `Release` with `PublishRelease`


richlander:
# Enable publishing as `Release` with `PublishRelease`

> The following proposal has been replaced with this new (non-breaking) plan: https://github.com/dotnet/sdk/issues/23551#issuecomment-1176765107

# [breaking] dotnet publish/pack produce `Release` assets by default

Proposal: Default to `Release` builds for CLI commands that are strongly aligned with production deployment. Release builds are faster, and debug builds are not supported in production.

That means:

- Fewer `Debug` artifacts would be deployed into production. Developers would not have specify as much via the CLI to do the right thing.
- Some developers would be surprised during debugging (builds would be optimized and less debuggable) but there would be an easy workaround to resolve that (opt into `debug` mode).

## Context

Our basic guidance to developers since .NET Core 1.0 has been "use `build` for development and `publish` for prod". Given that, it would make a lot more sense if `publish` defaulted to a release build. Debug builds run observably slow (sometimes you can see this with just your eyes; no stopwatch required). It is near certain that plenty of .NET apps are deployed as debug due to the current defaults. The CLI would be much better if it offered more differentiated options. There should probably be a more broad re-assessment of `build` and `publish` but I'm not digging into that here. I'm proposing a much simpler change (that doesn't preclude broader changes later; in fact, this change would encourage them).

`pack` is a bit more nuanced since there is no duality set of commands for it, no `build` and `publish` analogues. There is just `pack`. In fact, the same could be said of libraries in general. I assert that most developers go through their debug cycle with their NuGet library as just a library, then `pack` it and upload it to their NuGet server/feed of choice.  In that workflow, `pack` defaulting to `Release` would be preferred. I haven't done the exercise, but it would interesting to see the split between `Release` and `Debug` libraries on NuGet.org. There should be none (except for some very specific scenarios).

There absolutely is a `Debug` scenario with NuGet packages and with `pack`. I'm guessing that it is <10% case. The scenario would be that developers build their package as `Debug`, deploy to a feed (which could just be a disk location), then run tests (in one form or another), and use the debugger in library code to determine library behavior. For that to work really well, they may have to use SourceLink. I haven't tested the non-SourceLink flow for a `Debug` NuGet package (where I have matching source) in quite some time. That's a detail to explore. The primary point is that debugging NuGet packages is a bit more involved than debugging ProjectReference libraries, so it probably happens a lot less often.

[...]

PR #25991, opened 2022-06-13
Allow Opt-In to change the default to Release for Pack/Pub

nagilson:
Fix for https://github.com/dotnet/sdk/issues/23551#issuecomment-1150204351. Drafting until unit tests are added to assure this works. Fixes the issue, but unlike the fix in this PR https://github.com/dotnet/sdk/pull/25926 it adds a property through MSBuild. 

A concern:
- [x] Do we want a separate property to provide more granular control over Release for pack and publish? I used one for now, I think that use case is quite rare. (YES)


src/Cli/dotnet/commands/dotnet-tool/update/xlf/LocalizableStrings.cs.xlf
```diff
@@ -52,6 +52,11 @@
         <target state="translated">Nástroj {0} byl úspěšně aktualizován z verze {1} na verzi {2} (soubor manifestu {3}).</target>
         <note />
       </trans-unit>
+      <trans-unit id="UpdateSucceededVersionNoChange">
```
joeloff, line 55: Is this from your previous PR?

nagilson: No, this is not up to date and merged with main yet, this is how it was before that PR was merged in 👍

nagilson:
These unrelated tests, and some centered around it, fail here, but they are also failing on main. Discussed that it's OK to review this nonetheless. 
`NativeAotStaticLib_only_runs_when_switch_is_enabled` 
`It_publishes_the_project_with_a_refs_folder_and_correct_deps_file` 
`ItReturnsACommandSpecWhenToolIsInAProjectRef` 
`ItRestoresWithTheSpecifiedVerbosity`

nagilson:
@joeloff If there is a concern with this breaking VS Publish, please reach out to me to discuss.

joeloff:
> @joeloff If there is a concern with this breaking VS Publish, please reach out to me to discuss.

I'll let @vijayrkn weigh in on this, but I think VS should be fine. In VS, the publishing wizard usually lets you choose the configurations and I believe it is also stored in the profile. The wizards usually trigger a solution build programmatically and should be passing a value for the `Configuration` property.

nagilson:
Ok, the feature is opt-in now instead of by default. If the build succeeds, would appreciate a final review before merging. Shall I also raise a PR reflecting this change here: [docs/msbuild-props.md at main · dotnet/docs (github.com)](https://github.com/dotnet/docs/blob/main/docs/core/project-sdk/msbuild-props.md) ?

richlander (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props
```diff
@@ -15,12 +15,14 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- This property disables the conflict resolution logic from the Microsoft.Packaging.Tools package, which is superceded by the logic here in the SDK -->
     <DisableHandlePackageFileConflicts>true</DisableHandlePackageFileConflicts>
   </PropertyGroup>
-
+
```
joeloff, line 18: nit: can you remove the extra space?

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -40,7 +43,7 @@ public void ItPublishesARunnablePortableApp()
                 .Execute("--framework", ToolsetInfo.CurrentTargetFramework)
                 .Should().Pass();
 
-            var configuration = Environment.GetEnvironmentVariable("CONFIGURATION") ?? "Debug";
+            var configuration = Environment.GetEnvironmentVariable("CONFIGURATION") ?? _expectedConfigurationDefault;
```
joeloff, line 46: Maybe rename to `defaultConfiguration`? Calling it `_expected` makes it seem like it's intended to be a test result, but it's used as a default for when the environment variable is not set.

baronfel (approved):
There are a few files where the only changes are whitespace, like the Publish.targets and WebSdk Publish props. Can you revert those whitespace-only changes to minimize diffs/reduce visual clutter?

dsplaisted (changes requested):
I haven't been involved in the recent conversations around this, but I don't get what's going on here.  With this PR, it looks like if you set `PackRelease` or `PublishRelease` to true, then all builds (whether doing a pack, publish, or a normal build) will be set to release by default.

Is this the intention?  I would expect that if you set `PublishRelease` to true, that would change the default configuration for publish, but non-publish builds would still use Debug.

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
nagilson: Good call, this name used to be more relevant but that's better now.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props
nagilson: Yes, thanks for the call out. 

I installed this extension now https://marketplace.visualstudio.com/items?itemName=mynkow.FormatdocumentonSave&ssr=false#overview so these artifacts of removed changes shouldn't reoccur. Sadly there were some problems elsewhere in the file that also got changed. Not sure if that's going to be annoying in the future if my auto-formatter is fixing the format of code that's in the same file but not relevant to my changes? (IMO it's better to fix the formatting everywhere.)

nagilson:
@dsplaisted 
> Is this the intention? I would expect that if you set `PublishRelease` to true, that would change the default configuration for publish, but non-publish builds would still use Debug.

Yikes, with these new changes that is a problem and I should've considered the implications more. Thank you, will work on that! But I don't think we can tell which command is going to be done early enough for this change to be made here. But then there are concerns of setting the Configuration later down in the chain. Will talk to @joeloff about this.

richlander:
> I would expect that if you set `PublishRelease` to `true`, that would change the default configuration for `publish`, but non-`publish` builds would still use `Debug`.

@dsplaisted -- That is indeed the intention.

src/Cli/dotnet/commands/dotnet-pack/LocalizableStrings.resx
```diff
@@ -148,9 +148,9 @@
     <value>The project to pack, defaults to the project file in the current directory. Can be a path to any project file</value>
   </data>
   <data name="ConfigurationOptionDescription" xml:space="preserve">
-    <value>The configuration to use for building the package. The default for most projects is 'Debug'.</value>
+    <value>The configuration to use for building the package. The default is 'Debug'. Use the `PackRelease` property to make 'Release' the default.</value>
```
dsplaisted, line 151: Suggestion:

```suggestion
    <value>The configuration to use for building the package. The default is 'Debug'. Use the `PackRelease` property to make 'Release' the default for this command.</value>
```

src/Cli/dotnet/commands/dotnet-publish/LocalizableStrings.resx
```diff
@@ -146,7 +146,7 @@
 The default is to publish a framework-dependent application.</value>
   </data>
   <data name="ConfigurationOptionDescription" xml:space="preserve">
-    <value>The configuration to publish for. The default for most projects is 'Debug'.</value>
+    <value>The configuration to publish for. The default is 'Debug'. Use the `PublishRelease` property to make 'Release' the default.</value>
```
dsplaisted, line 149:
```suggestion
    <value>The configuration to publish for. The default is 'Debug'. Use the `PublishRelease` property to make 'Release' the default for this command.</value>
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -84,5 +84,54 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
+               .Execute()
+               .Should()
+               .Pass();
+
+            var packCommand = new DotnetPackCommand(Log, helloWorldAsset.TestRoot);
+
+            packCommand
+                .Execute()
+                .Should()
+                .HaveStdOutContaining("Release")
```
dsplaisted, line 107: It might be better to check that the package is created in the `bin\Release` output folder.

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -84,5 +84,54 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
+               .Should()
+               .Pass();
+
+            // Another command, which should not be affected by PackRelease
+            var publishCommand = new DotnetPublishCommand(Log, helloWorldAsset.TestRoot);
+
+            publishCommand
+                .Execute()
+                .Should()
+                .NotHaveStdOutContaining("Release")
```
dsplaisted, line 132: I'd suggest checking that the `bin\Release` folder doesn't exist at all here.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -423,6 +423,30 @@ public void It_fails_for_unsupported_rid()
...
+           .Execute()
+           .Should()
+           .Pass();
+
+            var publishCommand = new DotnetPublishCommand(Log, helloWorldAsset.TestRoot);
+
+            publishCommand
+            .Execute()
+            .Should()
+            .HaveStdOutContaining("Release")
```
dsplaisted, line 445: Same here, I'd suggest checking for the package in the `bin\Release` folder.

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -340,5 +343,31 @@ public void DotnetPublishAllowsPublishOutputDir()
...
+               .Should()
+               .Pass();
+
+            // Another command, which should not be affected by PublishRelease
+            var packCommand = new DotnetPackCommand(Log, helloWorldAsset.TestRoot);
+
+            packCommand
+                .Execute()
+                .Should()
+                .NotHaveStdOutContaining("Release")
```
dsplaisted, line 368: Same here, I'd suggest checking that the `bin\Release` folder doesn't exist (and the `bin\Debug` one does).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -477,37 +477,37 @@ Copyright (c) .NET Foundation. All rights reserved.
                             ResolveRuntimePackAssets;
                             _ComputePackageReferencePublish">
 
-      <!-- For future: Delete ResolveCopyLocalAssets task.  Need to figure out how to get correct DestinationSubPath for
+    <!-- For future: Delete ResolveCopyLocalAssets task.  Need to figure out how to get correct DestinationSubPath for
```
dsplaisted, line 480: Please revert the whitespace changes to this file and any other whitespace-only changes.

dsplaisted (approved):

dsplaisted:
@richlander Note that as it currently stands, `PublishRelease` will affect `dotnet publish`, but not something like `dotnet build /t:Publish`.  That's the best we think we can do, because if you're just calling msbuild targets we don't know during evaluation which ones are going to be called.

Same thing goes for `PackRelease` and `dotnet pack`.

nagilson:
> Note that as it currently stands, `PublishRelease` will affect `dotnet publish`, but not something like `dotnet build /t:Publish`. That's the best we think we can do, because if you're just calling msbuild targets we don't know during evaluation which ones are going to be called.

Yep, called that out here: https://github.com/dotnet/sdk/issues/23551#issuecomment-1164795419

EskeRahn:
> @richlander Note that as it currently stands, `PublishRelease` will affect `dotnet publish`, but not something like `dotnet build /t:Publish`. That's the best we think we can do, because if you're just calling msbuild targets we don't know during evaluation which ones are going to be called.
> 
> Same thing goes for `PackRelease` and `dotnet pack`.

Any way to set this would certainly be an improvement, but will it not be a cause of confusion having them acting differently?  
But why not extend the idea and add a _new_ option: `dotnet build /t:PublishRelease` and perhaps even `dotnet publishrelease`?

PR #25993, opened 2022-06-14
Move ILLink targets to the linker repo

agocke:
This will keep the ILLink task and targets in-sync.


agocke:
I think there's some potential simplification here by getting rid of the "SDK" stuff entirely. Just `<Import Include="../../Microsoft.NET.Sdk.ILLink/Microsoft.NET.Sdk.ILLink.Tasks.props" etc`

agocke:
@dsplaisted Is there any benefit to the "Sdk" login in MSBuild for these built-in components? It seems like it's just a space/time saver for people typing project files manually.

agocke:
@MichalStrehovsky FYI, in case this conflicts with anything you're thinking. I was planning on extracting trimming pieces that are shared between NativeAOT and ILLink into the central publish targets (like the feature switches).

MichalStrehovsky:
> @MichalStrehovsky FYI, in case this conflicts with anything you're thinking. I was planning on extracting trimming pieces that are shared between NativeAOT and ILLink into the central publish targets (like the feature switches).

Looks good to me. We should do the ILLinkTargetPath for NativeAOT targets as well. (Not really a comment for this pull request, just an observation.)

LakshanF (approved):
I like the effort to keep in sync as well streamline common actions between `NativeAOT `and `Trimming`. I'm assuming that the existing tests we have for both `NativeAOT `and `Trimming` in the SDK will guard against regression here.

PR #25995, opened 2022-06-14
Fix the path detection logic for MSDeploy executable

vijayrkn:


src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
```diff
@@ -263,13 +264,24 @@ public string ExePath
...
+                        /// On 32-bit Operating Systems, this will return C:\Program Files
+                        /// On 64-bit Operating Systems - regardless of process bitness, this will return C:\Program Files
+                        if (!Environment.Is64BitOperatingSystem || Environment.Is64BitProcess)
+                        {
+                            programFiles = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
+                            m_exePath = IO.Path.Combine(programFiles, msdeployExePath);
+                        }
+                        else
+                        {
+                            // 32 bit process on a 64 bit OS can't use SpecialFolder.ProgramFiles to get the 64-bit program files folder
```
KalleOlaviNiemitalo, line 281: I thought the KnownFolder API might be more reliable than an environment variable, and there is FOLDERID_ProgramFilesX64 dedicated to this folder, but the docs say it is not supported in 32-bit processes. ☹️

src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
```diff
@@ -263,13 +264,24 @@ public string ExePath
 #else
                 if (string.IsNullOrEmpty(m_exePath))
                 {
-                    string programFilesX86 = System.Environment.ExpandEnvironmentVariables("%ProgramFiles(x86)%");
-                    string msdeployExePath = System.IO.Path.Combine("IIS", "Microsoft Web Deploy V3");
-                    m_exePath = IO.Path.Combine(programFilesX86, msdeployExePath);
+                    string programFiles = Environment.ExpandEnvironmentVariables("%ProgramFiles(x86)%");
+                    string msdeployExePath = IO.Path.Combine("IIS", "Microsoft Web Deploy V3");
```
KalleOlaviNiemitalo, line 268: IIRC the directory name does not always include "V3". Perhaps that was with an older version that you don't support here.

src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
```diff
@@ -263,13 +264,24 @@ public string ExePath
 #else
                 if (string.IsNullOrEmpty(m_exePath))
                 {
-                    string programFilesX86 = System.Environment.ExpandEnvironmentVariables("%ProgramFiles(x86)%");
-                    string msdeployExePath = System.IO.Path.Combine("IIS", "Microsoft Web Deploy V3");
-                    m_exePath = IO.Path.Combine(programFilesX86, msdeployExePath);
+                    string programFiles = Environment.ExpandEnvironmentVariables("%ProgramFiles(x86)%");
```
KalleOlaviNiemitalo, line 267: I'd have first checked the directory that matches the current process bitness, then the other one. This instead checks the x86 directory first and then x64. But I suppose the directories that msdeploy reads or writes are unlikely to be redirected by wow64 so it won't really matter.

src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
```diff
@@ -263,13 +264,24 @@ public string ExePath
...
+                    string msdeployExePath = IO.Path.Combine("IIS", "Microsoft Web Deploy V3");
+                    m_exePath = IO.Path.Combine(programFiles, msdeployExePath);
```
KalleOlaviNiemitalo, line 269: If an exception is thrown later in this property, then this m_exePath assignment will remain in effect and shortcut any subsequent reads of the property. But perhaps there won't be any such reads after the exception.

src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
vijayrkn: Went with the existing env variable approach because it can help fix issues like this #25887 where users can set the env variable themselves to bypass the issue.

src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
vijayrkn: If you look at the existing code, it does the x86 check first. Also, this reduces the risk of anyone who might have set the env variable themselves to bypass this issue. 

That is the reason, I left the first portion of the check unchanged. If the file is not present, then it goes through the right checks based on machine and process bitness to find the right exe.

src/WebSdk/Publish/Tasks/Tasks/MsDeploy/MSDeploy.cs
vijayrkn: I would expect it to fail as it fails today as mentioned here #25887. This tells the user where msbuild is checking for the specific file and the user can take action accordingly. 

@KalleOlaviNiemitalo - I really appreciate your thoughtful and thorough code review feedbacks for the PR. Thanks!

abpiskunov (approved):

anvillan (approved):

PR #26007, opened 2022-06-14
Define SystemCollectionsImmutableVersion Versions.props

MichaelSimons:
Remove hard-coded System.Collections.Immutable package version reference in Microsoft.NET.Build.Tasks.csproj and replace with Versions.props property.  This avoids package version downgrade issues in source-build.

This is related to the changes made with https://github.com/dotnet/sdk/commit/13a506b44ac6b0daa8ba2e77eb1342518f7e81a4


MichaelSimons:
Would be nice to get this in soon so we can remove a contentious source-build patch - see https://github.com/dotnet/installer/pull/13961

dsplaisted (approved):

PR #26080, opened 2022-06-17
Reference older MSBuild for .NET Framework tasks

rainersigwald:
Updates of MSBuild's transitive references have caused insertion breaks twice now, in #25507 and #25897. Both happened because, in older-but-still-supported Visual Studio environments, MSBuild's binding redirects mismatched what the task's compilation environment provided, causing some types to be resolved from an older version of an assembly and some types to be resolved from a newer, creating mismatches at runtime.

I initially "fixed" this by manually downgrading some of the transitive references at SDK compilation time in 66826cc. But that is confusing, labor-intensive, and hard to maintain.

This is a different solution: explicitly target the version of MSBuild that the tasks are expected to run on. MSBuild's compatibility guarantees are such that tasks compiled for an older MSBuild should always work, so this should be safe, and it more accurately reflects the intent of the SDK tasks: they should work on older MSBuild (and thus Visual Studio).

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj
@@ -65,17 +65,9 @@
         MSBuild depends on a new version of SRM/SCI; HostModel a very old one.
         In order to keep working with MSBuild 17.2, which had binding redirects
         for SCI 0-5.0.0.0 but not 6, this task must compile against 5, even
-        though that's not transitively correct. Nothing uses updated API
-        surface so it all works ok.
-
-        To accomplish this, explicitly reference the current version of the
-        packages, but exclude assets so it doesn't get passed to the compiler.
-        Also download an old version and (later in the target ReferenceOlderSCIandSRM)
-        pass them to the compiler. -->
-    <PackageReference Include="System.Reflection.Metadata" Version="$(SystemReflectionMetadataVersion)" IncludeAssets="none" />
-    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" IncludeAssets="none" />
-    <PackageDownload Include="System.Collections.Immutable" Version="[5.0.0]" GeneratePathProperty="true" />
-    <PackageDownload Include="System.Reflection.Metadata" Version="[5.0.0]" GeneratePathProperty="true" />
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj
```diff
@@ -61,21 +61,8 @@
...
-        surface so it all works ok.
-
-        To accomplish this, explicitly reference the current version of the
-        packages, but exclude assets so it doesn't get passed to the compiler.
-        Also download an old version and (later in the target ReferenceOlderSCIandSRM)
-        pass them to the compiler. -->
-    <PackageReference Include="System.Reflection.Metadata" Version="$(SystemReflectionMetadataVersion)" IncludeAssets="none" />
-    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" IncludeAssets="none" />
-    <PackageDownload Include="System.Collections.Immutable" Version="[5.0.0]" GeneratePathProperty="true" />
-    <PackageDownload Include="System.Reflection.Metadata" Version="[5.0.0]" GeneratePathProperty="true" />
```
rainersigwald, line 78: Oh whoops, this is too aggressive. Not needed for the new thing but needed for SRM/SCI :(

joeloff:
Do we need more changes? I'm seeing different missing method exceptions now in the Helix logs

```
C:\h\w\C36F0AA6\p\d\sdk\6.0.400-ci\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(628,5): error MSB4018: System.MissingMethodException: Method not found: 'System.Collections.Immutable.ImmutableArray`1<System.Reflection.Metadata.CustomAttributeTypedArgument`1<!0>> System.Reflection.Metadata.CustomAttributeValue`1.get_FixedArguments()
```

joeloff:
@dtivel Any idea why the PEM tests are failing on Linux. Has anything changed you're aware of? It's finding the file, just not the certificate, but finding the certificate on Windows.

dtivel:
@joeloff, I'm not aware of any changes.  How can I debug this failure?

joeloff:
> @joeloff, I'm not aware of any changes. How can I debug this failure?

WSL is the only option I know of, so I'm going to try and repro this.

benvillalobos (approved):

PR #26085, opened 2022-06-17
Remove Microsoft.AspNetCore.Razor.SourceGenerator.Tooling.Internal from version.details.xml

chsienki:


wtgodbe:
@MichaelSimons is this the right fix for https://github.com/dotnet/installer/pull/13961#issuecomment-1159237117?

chsienki:
Happy to do that if someone can point me in the right direction for how to?

MichaelSimons:
> @MichaelSimons so roll all razor-compiler dependencies back to 4cf24694112b70c35895666f159e25b07e60fb58, and turn off flow until sdk can ingest the new package?

Correct.

MichaelSimons:
> Happy to do that if someone can point me in the right direction for how to?

If you are familiar with the `darc` cli, you can do the following:

1. Run `darc get-asset --name Microsoft.AspNetCore.Razor.SourceGenerator.Tooling.Internal` and search the output to find the BAR Build Id for the commit to roll back to.  I see `137250`
2. From the SDK repo run `darc update-dependencies --id 137250` to update the dependencies from the specified build.

PR #26104, opened 2022-06-20
MSBuild server V2

AR-May:
Feature to improve inner loop experience of building small project. This would make building faster.
Previous attempts: #17916, #18263. 

The change makes sense after merging [MSBuild server PR](https://github.com/dotnet/msbuild/pull/7634) and its insertion to SDK.

Implementation:

Modify SDK to set MSBUILDUSESERVER = 1
Implement escape hatches DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER= 1


src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -100,8 +99,6 @@ public void EnvironmentVariable(string name, string value)
 
             if (value == string.Empty || value == "\0")
             {
-                // Do not allow MSBuild NOIPROCNODE as null env vars are not properly transferred to build nodes
```
rokonec, line 103: We shall decide if same logic apply to msbuild server v2. See the comment at line 105.
Since we can't set env to null in reaction to Server build request message, we can't guarantee that build which runs in server mode will run exactly same as build which runs in out-of-proc as that build will have null env var thanks to `ProcessStartInfo.EnvironmentVariables`.
It is worth to notice that if success of build relay on null env var, such build can't be parallelized to MSBuild nodes as those have same restrictions and limitations as MSBuild server.

rokonec (commented):
Lets discus my comment offline.

rokonec (approved):

baronfel:
Should MSBuild Server participate in `dotnet build-server shutdown`? I mentioned this off-hand to Rainer and he said there might not be an similar concept readily available for MSBuild to use.

rokonec:
Hi Chet,

I am not sure I understand  "Should MSBuild Server participate in dotnet build-server shutdown?" question.

Do you mean if msbuild.exe should/could somehow kill msbuild server created by dotnet build?

If so, why?

Best,
   roman
________________________________

baronfel:
The intent of the shutdown command is to give users a way to 'reset' any state that may have been persistent in long-running processes like the roslyn/razor compilation servers. It seems a natural fit for us to add ourselves to that list too for completeness' sake.

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -54,12 +54,7 @@ public MSBuildForwardingAppWithoutLogging(IEnumerable<string> argsToForward, str
             _argsToForward = argsToForward;
             MSBuildPath = msbuildPath ?? defaultMSBuildPath;
 
-            if (UseMSBUILDNOINPROCNODE)
-            {
-                // Force MSBuild to use external working node long living process for building projects
-                // We also refers to this as MSBuild Server V1 as entry process forwards most of the work to it.
-                EnvironmentVariable("MSBUILDNOINPROCNODE", "1");
-            }
+            EnvironmentVariable("MSBUILDUSESERVER", UseMSBuildServer ? "1" : "0");
```
Forgind, line 57:
```suggestion
            EnvironmentVariable("MSBUILDUSESERVER", UseMSBuildServer ? "1" : "0");
            
```

rokonec:
@baronfel @marcpopMSFT We, MSBuild perf team, did lots of testing and have high enough confidence to consider this PR ready to merge.

baronfel:
merging! Great work landing this, folks - let's get dogfooding going to get early feedback.

sebastienros:
image

PR #26105, opened 2022-06-20
Rollback razor tooling version

chsienki:


chsienki:
@MichaelSimons @wtgodbe This rolls back the version of the razor tooling to the last version that included the now deleted package. I also set the dependencies to `Pinned="true"` which I *think* should stop them updating?

wtgodbe (approved):
Nice! This means we won't have to disable the darc subscription either - but we'll need to remember to un-pin these once we ingest the "new" thing from razor-compiler.

chsienki:
I've added the unpin to the umbrella merge issue so we don't forget, thanks

MichaelSimons (approved):
LGTM - Thanks for fixing this.

wtgodbe:
This is the pipeline for this run but for whatever reason it's now showing up in the UI: https://dev.azure.com/dnceng/public/_build/results?buildId=1834837&view=results

marcpopMSFT:
Darwin leg test failure may need this commit: https://github.com/dotnet/sdk/pull/26100/commits/83cc471c41263778c6980f660ab02f179c2e8459

wtgodbe:
> Darwin leg test failure may need this commit: https://github.com/dotnet/sdk/commit/83cc471c41263778c6980f660ab02f179c2e8459

Done

Issue #26021, opened 2022-06-14
#18903 spelling is still not fixed


Code-DJ:
#18903 is still not fixed, I checked the main branch.
See: https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-workload/list/LocalizableStrings.resx

nagilson:
Closing this as it was resolved! Might take a bit for it show up in public stable releases though.

Issue #18903, opened 2021-07-14
38 misspellings of avaliable vs available 


nosami:
https://github.com/dotnet/sdk/search?q=avaliable&type=code :)

Normally wouldn't bother reporting, but at least one of the misspellings appears in user documentation

```
jasonimison@Jasons-MacBook-Pro-2 ~/dotnet-core-uninstall $ dotnet workload
Required command was not provided.
workload
  Install or work with workloads that extend the .NET experience.

Usage:
  dotnet [options] workload [command]

Options:
  -?, -h, --help  Show help and usage information

Commands:
  install <WORKLOAD_ID>    Install a workload.
  update                   Update all installed optional workloads.
  list                     List workloads available.
  search <WORKLOAD_ID>     Search for avaliable workloads.
  uninstall <WORKLOAD_ID>  Uninstall a workload.
  repair                   Install a workload.
  ```

marcpopMSFT:
Thanks.  Workloads is a new features so we should definitely get that one fixed.

sfoslund:
This is fixed in @mhutch's PR.

sfoslund:
This has been fixed in preview 7 and main.

PR #26139, opened 2022-06-21
Fix available typo and XML typo

nagilson:
Fix for #18903 https://github.com/dotnet/sdk/issues/26021 
Also noticed there are 2 typos in all of the Resx file comments! (Not super important) @marcpopMSFT do you have any contacts on that team?

### DIFF
--- a/src/Assets/TestProjects/AllResourcesInSatellite/Strings.en.resx
+++ b/src/Assets/TestProjects/AllResourcesInSatellite/Strings.en.resx
@@ -5,7 +5,7 @@
 
 		Version 1.3
 
-		The primary goals of this format is to allow a simple XML format 
+		The primary goal of this format is to allow a simple XML format 
 		that is mostly human readable. The generation and parsing of the 
 		various data types are done through the TypeConverter classes 
 		associated with the data types.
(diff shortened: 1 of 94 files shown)


danmoseley:
Yes it's probably not worth fighting the resx comment. Ideally the template wouldn't have these comments at all. Last time I asked about that, some time ago, they pointed out it would have to be done in such a way that it wasn't removed from existing files.

marcpopMSFT:
@nagilson for translation/resource questions, reach out to @cristianosuzuki77

danmoseley:
@nagilson I think removing the resx comment by default would require something like

1. find all the .NET SDK/VS code that edits resx files and ensure that it doesn't add the resx comment back if it is not already present.
2. remove the comment from some or all of the resx files included in project templates (these are in https://github.com/dotnet/templating and some other repos) plus the individual template for "add new resx file" (not sure where this is located, it's likely in VS sources)

for 1, ResxResourceWriter was already fixed in .NET Core it seems -- https://github.com/dotnet/winforms/pull/452. But if I make a new Winforms project, targeting .NET 7, remove the comment, and then edit the resources through the graphical designer, it puts it back. So a change would be needed in that designer -- I assume it is effectively part of the Winforms designer. Perhaps @DustinCampbell knows.

for 2, this can only be done for templates that will only be manipulated by tooling that has already been adjusted to be tolerant. Eg., we could remove the comment from templates for .NET 8, once the lowest version of Visual Studio supporting those templates had been changed. We would certainly not want to ship templates that VS or some other tool immediately dirties.

It would be nice to remove the schema at the same time - ResxResourceWriter still writes that. 

Having said all this, although I've long wished this be cleaned up, I can't claim it's any kind of priority.

nagilson:
Thank you for the thorough explanation, there are more working parts to it than I thought. I see why it hasn't been done now. May discuss this with Rainer, though I don't think it will be on our priority list anytime soon. Perhaps if some free time emerges in some future sprint, I will take a look at it.

PR #26143, opened 2022-06-21
Use the Implicit SDK RID for RID-Specific Apps by Default

### LINKED ISSUE #23539: dotnet build/publish uses the implicit SDK RID for RID-specific apps by default
# dotnet build/publish uses the implicit SDK RID for RID-specific apps by default

**Proposal:** .NET SDK uses the implicit Runtime Identifier (RID) in each case when a RID is required but isn't specified by the user. It should affect (at least) the following `dotnet` commands:

* `build`
* `publish`
* `run`

We should validate where `--self-contained` is supported/valid. Wherever it is valid, this behavior change should be made.

**Example:** The following example (console app template) should not fail, but succeed (using the implicit SDK RID). This failure isn't useful (it doesn't protect the user from anything scary).

```bash
PS C:\Users\rich\app3> dotnet build --self-contained
Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET

  All projects are up-to-date for restore.
C:\Program Files\dotnet\sdk\6.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(150,5): error NETSDK1031: It is not supported to build or publish a self-contained application without specifying a RuntimeIdentifier. You must either specify a RuntimeIdentifier or set SelfContained to false. [C:\Users\rich\app3\app3.csproj]

Build FAILED.
```

Note: Part of this feature was implemented in https://github.com/dotnet/sdk/pull/22314.

### LINKED ISSUE #26028: Use implicit RID for  `Publish` Properties
# Use implicit RID for  `Publish` Properties

Based on https://github.com/dotnet/sdk/issues/23539, we have the opportunities to make additional improvements fit-and-finish improvements. The referenced change enables an implicit RID to be used when a RID is needed, specifically via the CLI like with `--self-contained`. There several properties where that is equally valuable.

We should extend the same change to:

- `SelfContained`
- `PublishAot`
- `PublishReadyToRun`
- `PublishSingleFile`
- `PublishTrimmed`

These proposed changes make it obvious that we're missing `PublishSelfContained` as an option. It's odd that this has never come up. It would enable producing self contained apps, but only for the `publish` verb, just like the other `Publish*` properties.

We should add:

- `PublishSelfContained`

In retrospect, I've struggled with the lack of `PublishSelfContained`, but it never occurred to me that we could have such a straightforward and symmetric solution to solve that.

## --use-current-runtime

The new implicit RID feature replaces `--use-current-runtime`. The following two lines are now equivalent.

```bash=
dotnet build --self-contained
dotnet build --self-contained --use-current-runtime
```

We should deprecate (and undocument) `--use-current-runtime` going foward. It no longer has a purpose.

@baronfel @agocke @nagilson @marcpopMSFT

### LINKED ISSUE #26445: Enable publishing a self-contained app with `PublishSelfContained`
We call self-contained apps as "SCD" or "self-contained deployment". It's clearly primarily intended as a publishing tasks not inner-loop, yet we don't have a `Publish*` property for it. We should create `PublishSelfContained` property.

This is also mentioned in https://github.com/dotnet/sdk/issues/26028 and is logically the same as #26444.

nagilson:
Part 1 of 4 for https://github.com/dotnet/sdk/issues/26503 
Resolves https://github.com/dotnet/sdk/issues/23539 
Resolves https://github.com/dotnet/sdk/issues/26028

## The implicit RID is now inferred under the following options:

SelfContained=true,
PublishReadyToRun=true,
PublishSingleFile=true
PublishAot=true
~~PublishSelfContained=true.~~

 ~~PublishSelfContained can be added to make a project publish self-contained by default.~~

Removed code from before where RID is injected in the CLI and moved it into the MS Build target. 
Some changes to how self-contained is inferred as well. 

## To observe changes:

The following worked before w/ Sarah's change but doesn't require extra computation in the CLI anymore:

-       dotnet publish --self-contained
- 	dotnet build --self-contained

The following would cause an error before: 
-       Also observe the property if set in .csproj. 

- 	dotnet publish -p:PublishSingleFile=true
- 	dotnet publish -p:PublishReadyToRun=true
- 	dotnet publish -p:SelfContained=true
- 	dotnet publish -p:PublishAot=true
- 	dotnet build -p:SelfContained=true (Note this also works if set in csproj unlike before, like others) 

~~The following did not mean anything before and now causes the correct behavior, validating it works it on publish scenarios through the CLI (and the CLI only ftr):~~

- Also check in the .csproj:
~~- 	dotnet publish -p:PublishSelfContained=true~~

 ~~Broken by this PR Until Daniel's PR is merged into MSBuild fixing the bug which causes this to break:~~

 ~~Publishing a rid agnostic library.~~

What didn't change:

- 	dotnet publish -p:PublishTrimmed

## TODO: 
- [x] Verify PublishReadyToRun and such actually use implicit RID, improve tests.
- [x] Handle PublishAot
- [x] Fix the NetSDK error in case something goes wrong 
- [x] verify publish self contained


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -61,7 +61,16 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RuntimeIdentifier Condition="'$(PlatformTarget)' == 'x86' or '$(PlatformTarget)' == ''">win7-x86</RuntimeIdentifier>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == 'true'">
+  <PropertyGroup Condition="'$(RuntimeIdentifier)' == '' and
```
nagilson, line 64: Probably change this next week to also just check use current runtime identifier

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -154,6 +163,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == ''"
                  ResourceName="CannotHaveSelfContainedWithoutRuntimeIdentifier" />
 
+    <NETSdkError Condition="('$(PublishAot)' == 'true' or '$(PublishSingleFile)' == 'true' or '$(PublishTrimmed)' == 'true' or '$(PublishReadyToRun)' == 'true') and '$(RuntimeIdentifier)' == ''"
```
nagilson, line 166: add publish self contained and make this less ugly

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -373,5 +373,6 @@ static int Last2DigitsTo0(int versionBuild)
             var publishCommand = new PublishCommand(createdAppProject);
             publishCommand.Execute(new [] {"-property:SelfContained=true", "-property:_CommandLineDefinedSelfContained=true", $"-property:RuntimeIdentifier={rid}", "-property:_CommandLineDefinedRuntimeIdentifier=true" }).Should().Pass().And.NotHaveStdOutContaining("warning");
         }
+
```
nagilson, line 376: get rid of this

nagilson:
So Rid Agnostic projects are a whole other thing to consider. Working on that right now

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASelfContainedApp.cs
```diff
@@ -129,7 +111,9 @@ public void It_can_make_a_Windows_GUI_exe()
                 .Be(2);
         }
 
-        [Fact]
+        // TODO: This test should be fixed by the fixes in MSBuild here:
+        // Won't merge this code in until that has flown in, elsewise libraries which are rid agnostic will get rids.
+        [Fact(Skip = "https://github.com/dotnet/msbuild/pull/6924")]
```
nagilson, line 116: Code from another PR linked above needs to be merged in for this to work because SelfContained is being pushed into library code that isn't actually self-contained. This change breaks that until that other change is merged in and fixes the issue.

nagilson:
Note on that one failing test: `SelfContained=false` is not supposed to be used with `PublishTrimmed` and it becomes a bit circular. Already having a discussion about that one.

richlander:
Can you document a set of CLI commands that would have had an error previously that now work because of this work? That would help a lot in reviewing your work. I'm not really the one to review all the code.

nagilson:
> Can you document a set of CLI commands that would have had an error previously that now work because of this work? That would help a lot in reviewing your work. I'm not really the one to review all the code.

@richlander Note this was tested exclusively in windows environments. 

The following worked before w/ Sarah's change but doesn't require extra computation in the CLI anymore:

-       dotnet publish --self-contained
- 	dotnet build --self-contained

The following would cause an error before: 
-       Also observe the property if set in .csproj. 

- 	dotnet publish -p:PublishSingleFile=true
- 	dotnet publish -p:PublishReadyToRun=true
- 	dotnet publish -p:SelfContained=true
- 	dotnet publish -p:PublishAot=true
- 	dotnet build -p:SelfContained=true (Note this also works if set in csproj unlike before, like others) 

The following did not mean anything before and now causes the correct behavior:

- Also check in the .csproj:
- 	dotnet publish -p:PublishSelfContained=true

Broken by this PR Until Daniel's PR is merged into MSBuild fixing the bug which causes this to break:

- 	Publishing a rid agnostic library. 

What didn't change:

- 	dotnet publish -p:PublishTrimmed

nagilson:
Retargeting to 7.0.100.

nagilson:
Marking as `draft` again as Daniel's change has flown into MSBuild and the SDK. Once I verify this change works with that and agnostic libraries aren't broken anymore, I will remark as open and ping.

nagilson:
I am going to try to get the implicit RID in for RC2. For `PublishSelfContained`, I don't think is going to happen for RC2 as it looks like its going to require another change to msbuild. Removing it from this PR and putting it elsewhere once it's ready.

nagilson:
@dotnet/domestic-cat May someone please take a quick look at this? Hoping to get it in for RC2.

joeloff (approved):
I'm happy with the change.

My only concern would be to maybe reword some of the errors. Right now it reads as `A runtime identifier for the property 'PublishAot' couldn't be inferred.` and I wonder whether `'PublishAot' property` reads better.

nagilson:
Awesome, thanks @joeloff for your review. Remaining work is looking into one test which appears to fail flakily on helix.

PR #26175, opened 2022-06-22
Not generate deps and config file for native aot apps

### LINKED ISSUE #63265: Publishing a NativeAOT app generates JSON files
### Description

Publishing a console app as AOT through VS still generates the following JSON files:

```
AppName.runtimeconfig.json
AppName.deps.json
```

If you look at these files it tells something about the .NET version. The .NET references etc. All of this seems pretty much obsolete for a native executable. Is there any reason why these files are generated or is this a bug?

### Reproduction Steps

1. Create console .NET 6 app
2. Add ILCompiler nuget package
3. In VS 2022 create a publish to folder profile with the following settings: Self-contained, win-x64
4. Publish through VS

### Expected behavior

Only executable in publish folder

### Actual behavior

Executables + JSON files in publish folder

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

LakshanF:
Fixes https://github.com/dotnet/runtime/issues/63265

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
@@ -381,7 +381,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
       <!-- Copy the deps file if using the build deps file. -->
       <ResolvedFileToPublish Include="$(ProjectDepsFilePath)"
-                             Condition="'$(GenerateDependencyFile)' == 'true' and '$(_UseBuildDependencyFile)' == 'true' and '$(PublishAot)' != 'true'">
+                             Condition="'$(GenerateDependencyFile)' == 'true' and '$(_UseBuildDependencyFile)' == 'true'">
         <RelativePath>$(ProjectDepsFileName)</RelativePath>
         <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
       </ResolvedFileToPublish>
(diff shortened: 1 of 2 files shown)


AntonLapounov (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -381,14 +381,14 @@ Copyright (c) .NET Foundation. All rights reserved.
 
       <!-- Copy the deps file if using the build deps file. -->
       <ResolvedFileToPublish Include="$(ProjectDepsFilePath)"
-                             Condition="'$(GenerateDependencyFile)' == 'true' and '$(_UseBuildDependencyFile)' == 'true'">
+                             Condition="'$(GenerateDependencyFile)' == 'true' and '$(_UseBuildDependencyFile)' == 'true' and '$(PublishAot)' != 'true'">
```
MichalStrehovsky, line 384: Do we need this part? `_UseBuildDependencyFile` should be false if we're publishing trimmed:

https://github.com/dotnet/sdk/blob/aa74dca0a303b40dff86939517eac3044847cc35/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets#L829-L837

Are we setting PublishTrimmed too late?

(Do not consider this blocking if this is an ordering issue again - just want to understand why we need to change both places in this file.)

vitek-karas:
The problem is that PublishAOT is publish-only... So when you `dotnet build` such an app it still needs to generate both of those files (so that F5 works). But when you `dotnet publish` those files must not be in the final output.
Since `dotnet publish` effectively calls `dotnet build` as a substep - we can't disable the generation - at least I think we can't.

LakshanF:
Yes, that is the intention. There is a test that explicitly checks that the build step keeps these files since not generating at build step will result in a poor user experience

MichalStrehovsky (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
LakshanF: Thanks!

Issue #26026, opened 2022-06-15
Breaking change in response file parsing 


matt-richardson:
### Describe the bug

Since upgrading from SDK `6.0.300` to `6.0.301`, the response file parsing logic has changed, where

```
-p:VSTestTestAdapterPath=".;C:\BuildAgent\plugins\dotnet\tools\vstest15"
```

used to work, but now shows an error:

```
dotnet build .\IPFX.UC.PresenceServer.sln @C:\BuildAgent\temp\agentTmp\1.rsp
Microsoft (R) Build Engine version 17.2.0+41abc5629 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.
MSBUILD : error MSB1006: Property is not valid.
Switch: C:\BuildAgent\plugins\dotnet\tools\vstest15
For switch syntax, type "MSBuild -help"
```

### To Reproduce

```
docker run -it --rm mcr.microsoft.com/dotnet/sdk:6.0.301 bash
mkdir repro
cd repro
dotnet new console
echo "-p:VSTestTestAdapterPath=\".;/opt/buildagent/plugins/dotnet/tools/vstest15\"" > params.rsp
dotnet restore repro.csproj @params.rsp
```

### Exceptions (if any)
```
MSBUILD : error MSB1006: Property is not valid.
Switch: /opt/buildagent/plugins/dotnet/tools/vstest15

For switch syntax, type "MSBuild -help"
```

### Expected
```
  Determining projects to restore...
  All projects are up-to-date for restore.
```

### Further technical details
- Include the output of `dotnet --info`
```
# dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.301
 Commit:    43f9b18481

Runtime Environment:
 OS Name:     debian
 OS Version:  11
 OS Platform: Linux
 RID:         debian.11-arm64
 Base Path:   /usr/share/dotnet/sdk/6.0.301/

Host (useful for support):
  Version: 6.0.6
  Commit:  7cca709db2

.NET SDKs installed:
  6.0.301 [/usr/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.6 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.6 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```

[...]

PR #26204, opened 2022-06-23
Revert to token-per-line response file handling

baronfel:
An update to System.CommandLine changed the response file handling support to a different, more cross-platform-consistent logic. However, we cannot take this change because users expect a token-per-line style of processing. As a result, we change the dotnet CLI to use token-per-line handling for response files.

## Description

There are two issues resolved here:
* A change to the defaults in System.CommandLine resulted in response-file parsing breaking across versions. This was resolved by using the S.CL extension points to implement the old, one-token-per-line behavior.
* A user scenario used a version of the MSBuild `--property` option (`/p`) that the .NET SDK didn't have handling for. This was resolved by adding in handling for all unhandled forms of this property.

Fixes https://github.com/dotnet/sdk/issues/26026.

## Customer Impact

Customers that relied on the old response-file behavior typically ran into it in one of two scenarios - the first is related to an internal Brotli compression tool, the workaround here is to publish your application in a directory that contains no spaces. The second involved response files with the `/p` flag - the workaround here is to modify the response file to convert `/p` into one of the recognized forms (`-p`, `--p`, `-property`).

## Regression?

- [X] Yes
- [ ] No

## Risk

- [ ] High
- [X] Medium
- [ ] Low

The change touches response file handling, which can be very broadly used. It does however have testing and reimplements prior behavior.

## Verification

- [X] Manual (required)
- [X] Automated

Manually tested the provided user scenarios, and also added automated testing around the parsing changes.

- [ ] Yes
- [ ] No
- [X] N/A

### DIFF
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -17,7 +17,8 @@ namespace Microsoft.DotNet.Cli
     internal static class CommonOptions
     {
         public static Option<string[]> PropertiesOption =
-            new ForwardedOption<string[]>(new string[] { "-property", "/p", "-p" })
+            // these are all of the forms that the property switch can be understood by in MSBuild
+            new ForwardedOption<string[]>(new string[] { "--property", "-property", "/property", "/p", "-p", "--p" })
             {
                 IsHidden = true
             }.ForwardAsProperty()
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.DotNet.Cli
     internal static class CommonOptions
     {
         public static Option<string[]> PropertiesOption =
-            new ForwardedOption<string[]>(new string[] { "-property", "/p" })
+            new ForwardedOption<string[]>(new string[] { "-property", "/p", "-p" })
             {
                 IsHidden = true
             }.ForwardAsProperty()
--- a/src/Cli/dotnet/OptionForwardingExtensions.cs
+++ b/src/Cli/dotnet/OptionForwardingExtensions.cs
@@ -19,7 +19,11 @@ namespace Microsoft.DotNet.Cli
         public static ForwardedOption<T> ForwardAsSingle<T>(this ForwardedOption<T> option, Func<T, string> format) => option.SetForwardingFunction(format);
 
         public static ForwardedOption<string[]> ForwardAsProperty(this ForwardedOption<string[]> option) => option
-            .SetForwardingFunction((optionVals) => optionVals.SelectMany(optionVal => new string[] { $"{option.Aliases.FirstOrDefault()}:{optionVal.Replace("roperty:", string.Empty)}" }));
+            .SetForwardingFunction((optionVals) =>
+                optionVals
+                    .Select(optionVal => optionVal.Replace(";", "%3B")) // must escape semicolon-delimited property values when forwarding them to MSBuild
+                    .SelectMany(optionVal => new string[] { $"{option.Aliases.FirstOrDefault()}:{optionVal.Replace("roperty:", string.Empty)}" })
+                );
 
         public static Option<T> ForwardAsMany<T>(this ForwardedOption<T> option, Func<T, IEnumerable<string>> format) => option.SetForwardingFunction(format);
 
--- a/src/Cli/dotnet/OptionForwardingExtensions.cs
+++ b/src/Cli/dotnet/OptionForwardingExtensions.cs
@@ -22,7 +22,7 @@ namespace Microsoft.DotNet.Cli
             .SetForwardingFunction((optionVals) =>
                 optionVals
                     .Select(optionVal => optionVal.Replace(";", "%3B")) // must escape semicolon-delimited property values when forwarding them to MSBuild
-                    .SelectMany(optionVal => new string[] { $"{option.Aliases.FirstOrDefault()}:{optionVal.Replace("roperty:", string.Empty)}" })
+                    .Select(optionVal => $"{option.Aliases.FirstOrDefault()}:{optionVal}")
                 );
 
         public static Option<T> ForwardAsMany<T>(this ForwardedOption<T> option, Func<T, IEnumerable<string>> format) => option.SetForwardingFunction(format);
--- a/src/Cli/dotnet/Parser.cs
+++ b/src/Cli/dotnet/Parser.cs
@@ -109,6 +109,24 @@ namespace Microsoft.DotNet.Cli
                 .FirstOrDefault(c => c.Name.Equals(commandName, StringComparison.OrdinalIgnoreCase));
         }
 
+        /// <summary>
+        /// Implements token-per-line response file handling for the CLI. We use this instead of the built-in S.CL handling
+        /// to ensure backwards-compatibility with MSBuild.
+        /// </summary>
+        public static bool TokenPerLine(string tokenToReplace, out IReadOnlyList<string> replacementTokens, out string errorMessage) {
+            var filePath = Path.GetFullPath(tokenToReplace);
+            if (File.Exists(filePath)) {
+                replacementTokens = File.ReadAllLines(filePath).Where(line => !line.StartsWith("#")).ToArray();
+                errorMessage = null;
+                return true;
+            } else {
+                // don't report an error here, since MSBuild doesn't in the case of nonexistent response files
+                replacementTokens = null;
+                errorMessage = null;
+                return false;
+            }
+        }
+
         public static System.CommandLine.Parsing.Parser Instance { get; } = new CommandLineBuilder(ConfigureCommandLine(RootCommand))
             .UseExceptionHandler(ExceptionHandler)
             .UseHelp()
@@ -118,6 +136,7 @@ namespace Microsoft.DotNet.Cli
             .UseSuggestDirective()
             .DisablePosixBinding()
             .EnableLegacyDoubleDashBehavior()
+            .UseTokenReplacer(TokenPerLine)
             .Build();
 
         private static void ExceptionHandler(Exception exception, InvocationContext context)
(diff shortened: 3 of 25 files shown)


src/Cli/dotnet/Parser.cs
```diff
@@ -109,6 +109,24 @@ public static Command GetBuiltInCommand(string commandName)
...
         }
 
+        /// <summary>
+        /// Implements token-per-line response file handling for the CLI. We use this instead of the built-in S.CL handling
+        /// to ensure backwards-compatibility with MSBuild.
+        /// </summary>
+        public static bool TokenPerLine(string tokenToReplace, out IReadOnlyList<string> replacementTokens, out string errorMessage) {
+            var filePath = Path.GetFullPath(tokenToReplace);
+            if (File.Exists(filePath)) {
+                replacementTokens = File.ReadAllLines(filePath).Where(line => !line.StartsWith("#")).Select(line => $"\"{line}\"").ToArray();
```
dsplaisted, line 119: How sure are you that this exactly matches the previous behavior?

dsplaisted (approved):

src/Cli/dotnet/Parser.cs
baronfel: working through some of the details of that with @jonsequitur right now, will report back

src/Tests/dotnet.Tests/ParserTests/ResponseFileTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Tests.ParserTests
+{
+    public class ResponseFileTests
+    {
+        [Fact]
+        public void Can_safely_expand_response_file_lines() {
+            var tempFilePath = Path.GetTempFileName();
```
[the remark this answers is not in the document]
baronfel: thanks, will do

src/Cli/dotnet/OptionForwardingExtensions.cs
```diff
@@ -19,7 +19,11 @@ public static class OptionForwardingExtensions
         public static ForwardedOption<T> ForwardAsSingle<T>(this ForwardedOption<T> option, Func<T, string> format) => option.SetForwardingFunction(format);
 
         public static ForwardedOption<string[]> ForwardAsProperty(this ForwardedOption<string[]> option) => option
-            .SetForwardingFunction((optionVals) => optionVals.SelectMany(optionVal => new string[] { $"{option.Aliases.FirstOrDefault()}:{optionVal.Replace("roperty:", string.Empty)}" }));
+            .SetForwardingFunction((optionVals) =>
+                optionVals
+                    .Select(optionVal => optionVal.Replace(";", "%3B")) // must escape semicolon-delimited property values when forwarding them to MSBuild
```
baronfel, line 24: this is critical - without escaping (which we do in several other places in the codebase we probably should try to unify), msbuild will eat the `;` character. This is a longstanding issue that was helpfully pointed out by @janv8000 [here](https://github.com/dotnet/sdk/issues/26026#issuecomment-1161431464) that I boorishly disregarded. many thanks, friend!

as a result now, we consistently handle all msbuild-known property option forms.

src/Cli/dotnet/Parser.cs
```diff
@@ -109,6 +109,39 @@ public static Command GetBuiltInCommand(string commandName)
                 .FirstOrDefault(c => c.Name.Equals(commandName, StringComparison.OrdinalIgnoreCase));
         }
 
+        /// <summary>
+        /// Implements token-per-line response file handling for the CLI. We use this instead of the built-in S.CL handling
+        /// to ensure backwards-compatibility with MSBuild.
+        /// </summary>
+        public static bool TokenPerLine(string tokenToReplace, out IReadOnlyList<string> replacementTokens, out string errorMessage) {
```
baronfel, line 116: this is the core change that gets us back to parity with the response file parsing from the older S.CL version.  We get a presumed file path, and if that file path isn't an actual path error out to the user (S.CL did this previously). Then, we read the lines, strip out data from any # tokens onward, and drop all subsequently-empty lines.

src/Cli/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs
```diff
@@ -57,6 +57,8 @@ public MSBuildForwardingApp(IEnumerable<string> argsToForward, string msbuildPat
             }
         }
 
+        public IEnumerable<string> MSBuildArguments { get { return _forwardingAppWithoutLogging.GetAllArguments(); } }
```
baronfel, line 60: test helper that helped me see the property-forwarding problem.

src/Tests/dotnet.Tests/ParserTests/MSBuildArgumentCommandLineParserTests.cs
```diff
@@ -31,9 +31,10 @@ public void MSBuildArgumentsAreForwardedCorrectly(string[] arguments, bool build
             RestoringCommand command = buildCommand ? 
                 BuildCommand.FromArgs(arguments) : 
                 PublishCommand.FromArgs(arguments);
+            var expectedArguments = arguments.Select(a => a.Replace("-p:", "-property:"));
```
baronfel, line 34: we expand the `-p` form to `-property` now for consistency, so the expectations needed updating

src/Tests/dotnet.Tests/ParserTests/ResponseFileTests.cs
```diff
@@ -0,0 +1,52 @@
...
+                "build",
+                "a b",
+                "-p",
+                "VSTestTestAdapterPath=\".;/opt/buildagent/plugins/dotnet/tools/vstest15\""
+            };
+
+            Log.WriteLine($"MSbuild Args are {string.Join(" ", bc.MSBuildArguments)}");
+            Log.WriteLine($"Parse Diagram is {parseResult.Diagram()}");
+            Log.WriteLine($"Token string is {tokenString}");
+            tokens.Skip(1).Should().BeEquivalentTo(tokenized);
```
baronfel, line 49: note - probably need a test with # and empty strings still, but this validates the user-provided use case tokenizes correctly.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetBuildInvocation.cs
```diff
@@ -56,7 +56,7 @@ public void MsbuildInvocationIsCorrect(string[] args, string expectedAdditionalA
 
         [Theory]
         [InlineData(new string[] { "-f", "tfm" }, "-target:Restore", "-property:TargetFramework=tfm")]
-        [InlineData(new string[] { "-p:TargetFramework=tfm" }, "-target:Restore", "-p:TargetFramework=tfm")]
+        [InlineData(new string[] { "-p:TargetFramework=tfm" }, "-target:Restore", "-property:TargetFramework=tfm")]
```
baronfel, line 59: since all `-p` forms are normalized to `-property`, this needed updating as well.

baronfel (commented):
A few notes on the updates from today/consulting with @jonsequitur.

src/Cli/dotnet/Parser.cs
baronfel: wrapping in quotes is no longer necessary - this was an artifacts of some msbuild ;-gobbling that I've now addressed.

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.DotNet.Cli
     internal static class CommonOptions
     {
         public static Option<string[]> PropertiesOption =
-            new ForwardedOption<string[]>(new string[] { "-property", "/p" })
+            new ForwardedOption<string[]>(new string[] { "-property", "/p", "-p" })
```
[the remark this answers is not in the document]
dsplaisted: What about `/property`?  I think that would work too.

src/Cli/dotnet/OptionForwardingExtensions.cs
```diff
@@ -19,7 +19,11 @@ public static class OptionForwardingExtensions
...
+                    .SelectMany(optionVal => new string[] { $"{option.Aliases.FirstOrDefault()}:{optionVal.Replace("roperty:", string.Empty)}" })
```
dsplaisted, line 25: Will this work correctly if you do something like `-p:MyCrazyProperty=property:value`, or will it remove the `roperty:` portion of the property value?

dsplaisted (approved):

src/Cli/dotnet/OptionForwardingExtensions.cs
baronfel: good catch - I'll add some tests around this forwarding function specifically (by extracting it to a named function and slamming it with different inputs), but IMO we should be canonicalizing all of the different input shapes into `-property:<user input>` for MSBuild (meaning that we use the most verbose form available).

baronfel:
I could add `/property` as well, I thought about that last night - ideally we'd accept all forms of the property option that MSbuild accepts, _as well_ as the more posix-standard `--property` long form and `-p` short form.

src/Tests/dotnet.Tests/ParserTests/ResponseFileTests.cs
baronfel: added this test now.

baronfel:
@dsplaisted if you wanna take another look that'd be cool :) I added full support for the different permutations of `property` that MSBuild supports (`--`, `-`, `/`), and fleshed out testing for the token replacement for response files as well as the MSBuild property forwarding since the last time you looked.

dsplaisted (approved):

BradBarnich:
will this change fix to allow handling a comma in a quoted string in an RSP file? I didn't see this case covered in the tests

such as:
```
/p:Copyright="Copyright (C) CareEvolution, Inc. 2022"
```
another example in this comment: https://github.com/dotnet/sdk/issues/26026#issuecomment-1157522720

baronfel:
@BradBarnich thanks for checking! the response file part is parsed correctly, but the value gets mutilated during the forwarding to MSBuild. I will add additional test cases and fixes for this now.

baronfel:
Actually this _did_ fix parameters with spaces and quotes, and I was just comparing to the wrong collection in the tests. Updated and we should be good to go!

Issue #22148, opened 2021-10-18
Add additional details on the workload version in dotnet workload --list


marcpopMSFT:
### Is your feature request related to a problem? Please describe.
Customers want to know the version of what they have installed. Unfortunately, multiple packs live within a single workload and even that is tied to a manifest version as the workload itself doesn't have a version.  Likely needs PM guidance.  Should this be a separate parameter? Is manifest version enough?  Should it list out the version of every pack installed related to that workload?

Hypothesis: The manifest version is enough as that's used for rollback updates.

### Describe the solution you'd like
Add a column that shows the manifest version when adding an additional parameter to the list command (--verbosity).  Could include pack versions on verbosity diagnostics but that may not be a priority.

baronfel:
Counter-proposal: make this `dotnet workload --info`, and emit a structured representation of installed workloads, workload install type (MSI or File-based), key workload locations, manifest versions, etc.

Reasoning:
We need most/all of this information for issue triage, and so having a simple gesture to get all of this information easily would help smooth the triage process.

davidortinau:
@baronfel @marcpopMSFT now that .NET MAUI is GA, the "what version am I on" has only gotten louder. Are we on track for the 6.0.4xx milestone?

marcpopMSFT:
Assigning to Noah to take a look. I think including at least the manifest version (example 32.0.415 if I installed [android](https://www.nuget.org/packages/Microsoft.NET.Sdk.Android.Manifest-6.0.300)) in list at a minimum. Chet may have some ideas of how to add this to the --info details as well but it should be in both, not one or the other (and we don't need to block one on the other).

I don't know whether to include the manifest feature band but that might be useful too (like 6.0.300 for Android in my example above).

PR #26326, opened 2022-06-28
Display manifest version inside of workload list

nagilson:
Part 1 of 2 for resolving the following: https://github.com/dotnet/sdk/issues/22148#issuecomment-1168005944 
Will open a new PR to include this in info if this is approved. 

See change:
image

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
@@ -600,6 +600,8 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
         /// <remarks>
         /// Will fail if the workloadId provided is invalid.
         /// </remarks>
+        /// <exception>KeyNotFoundException</exception>
+        /// <exception>ArgumentNullException</exception>
         public WorkloadManifest GetManifestFromWorkload(WorkloadId workloadId)
         {
             return _workloads[workloadId].manifest;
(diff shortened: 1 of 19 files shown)


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs
```diff
@@ -592,6 +592,19 @@ private bool PackHasChanged(WorkloadPack oldPack, WorkloadPack newPack)
...
+        /// Finds the manifest for a specified workload.
+        /// </summary>
+        /// <param name="workloadId">The workload Id for which we want the corresponding manifest.</param>
+        /// <returns>The manifest for a corresponding workload.</returns>
+        /// <remarks>
+        /// Will fail if the workloadId provided is invalid.
+        /// </remarks>
+        public WorkloadManifest GetManifestFromWorkload(WorkloadId workloadId)
+        {
+            return _workloads[workloadId].manifest;
```
joeloff, line 605: I think I'm okay having this throw an exception. Perhaps just add the exceptions to the doc comments

```cs
<exception>KeyNotFoundException</exception>
<exception>ArgumentNullException</exception>
```

So it's clear this behaves like a dictionary. The other Option would be to expose the `_workloads` as an IReadOnlyDictionary property, that way the user knows its accessing an actual dictionary

nagilson: Adding those exceptions in, thanks for pointing that out. I like the idea of exposing `_workloads` to make it more clear. But since it is an implementation detail of `WorkloadResolver` and not `IWorkloadResolver` I think it would require changes to the interface which I'm not sure we want.

marcpopMSFT:
Do we want to include the feature band of the manifest (which can be different than the installation source) or is that too much info? I'm inclined to think it's too much.

dsplaisted:
You could include the manifest feature band as part of the manifest version in the same format as we do for rollback files, for example `32.0.300-rc.1.4/6.0.300`

nagilson:
This was pushed in 6.0.4xx and merged into main

PR #26367, opened 2022-06-29
Simplify caching logic to not try and retain caches between supression

chsienki:
This is a little weird, but our previous attempts to hold two caches in memory at once don't work due to the way the IDE works.

@jasonmalinowski is working on an IDE change that will share the cache between EnC invocations (https://github.com/dotnet/roslyn/pull/62219/) so essentially there will be two generator caches: one for which `IsSuppressed` is always true and one for which `IsSuppressed` is always false. As such we no longer need to try and maintain the caches between this flag switching: it never will and we can just set them to empty in the suppressed state. 

This will also fix the Speedometer regression (MSFT internal only: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1554147) because we'll no longer be doing any work on the affected paths. 

We don't need to wait for @jasonmalinowski's change to actually go in, because we effectively just go back to the original behavior with this PR. Hot reload will suffer perf wise, but the regular IDE paths will do no work. When / if the IDE side goes in, the hot reload perf will just start getting better.


chsienki:
@RikkiGibson and @333fred for review please.

333fred (approved):
I love grim reaper PRs :shipit:

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -19,13 +19,16 @@ public partial class RazorSourceGenerator : IIncrementalGenerator
 
         public void Initialize(IncrementalGeneratorInitializationContext context)
         {
-            // determine if we should suppress this run and set all the inputs to cached if so
-            var isGeneratorSuppressed = context.AnalyzerConfigOptionsProvider.Select(GetSuppressionStatus);
+            var analyzerConfigOptions = context.AnalyzerConfigOptionsProvider;
+            var parseOptions = context.ParseOptionsProvider;
+            var compilation = context.CompilationProvider;
```
jasonmalinowski, line 24: Do we do compute any caches for compilations that will run even if we don't have any other input files? I see the tag helpers from references and such. Doing that up front isn't necessarily bad, but just wanted to confirm it's running when you think it's running.

jasonmalinowski (approved):

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
chsienki: Everything else has a 'skip if there are no razor files' so it means we'll bail straight away for everything else too.

src/RazorSdk/SourceGenerators/RazorSourceGenerator.cs
```diff
@@ -19,13 +19,16 @@ public partial class RazorSourceGenerator : IIncrementalGenerator
...
-            var analyzerConfigOptions = context.AnalyzerConfigOptionsProvider.AsCachedIfSuppressed(isGeneratorSuppressed);
-            var additionalTexts = context.AdditionalTextsProvider.AsCachedIfSuppressed(isGeneratorSuppressed);
-            var parseOptions = context.ParseOptionsProvider.AsCachedIfSuppressed(isGeneratorSuppressed);
-            var compilation = context.CompilationProvider.AsCachedIfSuppressed(isGeneratorSuppressed);
+            // determine if we should suppress this run and filter out all the additional files if so
+            var isGeneratorSuppressed = context.AnalyzerConfigOptionsProvider.Select(GetSuppressionStatus);
+            var additionalTexts = context.AdditionalTextsProvider
+                 .Combine(isGeneratorSuppressed)
+                 .Where(pair => !pair.Right)
+                 .Select((pair, _) => pair.Left);
```
RikkiGibson, line 31: apologies for the basic questions incoming.. what does the RHS represent in this lambda? And does 'pair' here refer to the same thing as 'pair' on the previous line?

src/Tests/Microsoft.NET.Sdk.Razor.SourceGenerators.Tests/RazorSourceGeneratorTests.cs
```diff
@@ -2324,70 +2324,20 @@ protected override void BuildRenderTree(global::Microsoft.AspNetCore.Components.
...
-        {
-            __builder.AddMarkupContent(0, ""<h1>Counter</h1>"");
-        }
-        #pragma warning restore 1998
-    }
-}
-#pragma warning restore 1591
-"
-);
+                .VerifyOutputsMatch(result);
```
RikkiGibson, line 2340: nit: I find it a tiny bit confusing to read and assign 'result' here, particularly since it doesn't look like 'result' is read again after this.

RikkiGibson (approved):

RikkiGibson:
anything blocking this?

PR #26369, opened 2022-06-29
Allow dotnet workload update print-download-link-only to use fallback and work with prior SDK versions

### LINKED ISSUE #24335: Allow "dotnet workload update --print-download-link-only --sdk-version" to work with prior SDK versions
### Describe the bug
User has 6.0.100 installed. 
Updater offers 6.0.201 update.
User do not have 6.0.201  installed. 
Updater executes "dotnet workload update --print-download-link-only --sdk-version 6.0.201.

Expected Behaviour:
dotnet command from prior SDK version should able to read the workload manifest offered with  6.0.201 and list the nuget package urls from latest SDK. This allows updater to parse the urls and download the packages to cache along with 6.0.201 SDK update. 

Actual Behaviour:
"No manifests exist for SDK version 6.0.201" message is returned.

Why is this needed?
This feature is required for VS For Mac Updater to do a cross band .NET6 Workload Update without a two step process. This support avoids, updater downloading and installing the SDK 6.0.201 and then requesting for workload updates again.

gkulin:
fixes #24335

## Description
The VSMac updater calls the prior band SDK and asks it which workloads to install with the current feature band. The feature was originally designed to work with only the current band and not cross-band and this change fixes it to work that way.

## Customer Impact

Without this change, VSMac customers who upgrade will not be told on their initial upgrade that there are workloads that require updates. They will be forced to either drop to the command line and manually update workloads or wait for a future update that may or may not work depending on if it's a patch update or not.

## Regression?

- [ ] Yes
- [X] No

## Risk

- [ ] High
- [ ] Medium
- [X] Low

The change itself is a bit bigger but it's affect is almost entirely in the --print-downlaod-linke-only feature of the workloads CLI. This command is _only_ used by VSMac and it's completely broken today.

## Verification

- [x] Manual (required)
- [X] Automated
We have not tested with VSMac yet as we need either an official build with newer workloads or to backport this to 6.0.3xx.


src/Cli/dotnet/commands/dotnet-workload/install/IInstaller.cs
```diff
@@ -60,11 +60,18 @@ public class WorkloadDownload
 
         public string NuGetPackageVersion { get; }
 
+        /***public WorkloadDownload(string id, string nuGetPackageId, string nuGetPackageVersion)
```
joeloff, line 63: Can probably remove commented out code

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -249,9 +250,31 @@ public async Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsyn
...
+                        (success, latestVersion) = await GetPackageVersion(newPackageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreviews);
+                        
+                        if (success)
+                        {
+                            downloads.Add(new WorkloadDownload(manifest.Id, newPackageId.ToString(), latestVersion.ToString()));
+                        }
+                    }
+                    if (!success)
+                    {
+                        _reporter.WriteLine(string.Format(LocalizableStrings.FailedToGetPackageManifestUrl, packageId));
```
joeloff, line 276: nit: check formatting here. Can't tell if this is just the diff trying to fit the text properly

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -272,6 +295,9 @@ private async Task UpdateAdvertisingManifestAsync(WorkloadManifestInfo manifest,
             var manifestId = new ManifestId(manifest.Id);
             string currentFeatureBand = _sdkFeatureBand.ToString();
 
+
+            // TODO: abstract the fallback logic so it can be used in the get manifest package downloads async
```
joeloff, line 299: Is the TODO necessary or related to a separate issue?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Net.Http;
 using System.Runtime.InteropServices;
+using System.Security.AccessControl;
```
joeloff, line 10: Is this `using` needed?

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -75,7 +75,7 @@ public override int Execute()
             }
             else if (_printDownloadLinkOnly)
             {
-                var packageUrls = GetUpdatablePackageUrlsAsync(_includePreviews).GetAwaiter().GetResult();
+                var packageUrls = GetUpdatablePackageUrlsAsync(_includePreviews).GetAwaiter().GetResult();
```
joeloff, line 78: nit: extra whitespace

src/Layout/toolset-tasks/OverrideAndCreateBundledNETCoreAppPackageVersion.cs
```diff
@@ -62,6 +62,8 @@ public override bool Execute()
 
             var isSDKServicing = IsSDKServicing(propertyGroup.Element(ns + "NETCoreSdkVersion").Value);
 
+            isSDKServicing = true;
```
joeloff, line 65: this assignment seems to replace previous value

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -43,19 +43,23 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
...
             var sdkVersionBand =
                 $"{sdkVersionParsed.Major}.{sdkVersionParsed.Minor}.{Last2DigitsTo0(sdkVersionParsed.Build)}";
-
+           
+            
             _sdkRootPath = sdkRootPath;
             _sdkVersionBand = sdkVersionBand;
-
-            var knownManifestIdsFilePath = Path.Combine(_sdkRootPath, "sdk", sdkVersion, "IncludedWorkloadManifests.txt");
+            //sdkVersion = "6.0.301";
```
joeloff, line 58: Looks like some test/debug code that's setting the version to 6.0.400

src/Tests/dotnet-workload-install.Tests/GivenWorkloadManifestUpdater.cs
```diff
@@ -674,7 +674,48 @@ public void TestSideBySideUpdateChecks()
             File.GetLastAccessTime(sentinelPath2).Should().BeCloseTo(updateTime2);
         }
 
-       
+        /**
```
joeloff, line 677: Is this test needed? If you want to keep the code and just ignore the test you can use `[Theory(Skip="some reason ")]`

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -423,7 +427,7 @@ private IEnumerable<PackageSource> LoadNuGetSources(PackageSourceLocation packag
                     .SelectMany(result => result.foundPackages.Select(package => (result.source, package)));
 
             if (!accumulativeSearchResults.Any())
-            {
+            {   // this is the error that's thrown
```
nagilson, line 430: I would remove this comment, seems like it was used in debugging

src/Cli/dotnet/commands/dotnet-sdk/check/SdkOutputWriter.cs
```diff
@@ -36,7 +36,7 @@ public void PrintSdkInfo()
             if (NewFeatureBandAvailable())
             {
                 _reporter.WriteLine();
-                _reporter.WriteLine(string.Format(LocalizableStrings.NewFeatureBandMessage, NewestFeatureBandAvailable()));
+                _reporter.WriteLine(string.Format(LocalizableStrings.NewFeatureBandMessage, NewestFeatureBandAvailable()));         // this is where a new feature band is advertised
```
nagilson, line 39: I would suggest moving this comment above, and maybe simplify the comment to something like "Advertise the new feature band". IMO comments should prescriptive and not descriptive as the former may cue someone debugging the code into incorrect assumptions

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -249,9 +250,31 @@ public async Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsyn
...
-                    var latestVersion = await _nugetPackageDownloader.GetLatestPackageVerion(packageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreviews);
-                    downloads.Add(new WorkloadDownload(manifest.Id, packageId.ToString(), latestVersion.ToString()));
+
+                    bool success;
+                    (success, var latestVersion) = await GetPackageVersion(packageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreviews);
+                    if (success)
+                    {
+                        downloads.Add(new WorkloadDownload(manifest.Id, packageId.ToString(), latestVersion.ToString()));
+                    }
+                    if (!success)
```
nagilson, line 262: This logic would read easier if it were an else imo. Same with the check below

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadOptionsExtensions.cs
```diff
@@ -27,9 +27,9 @@ internal static ReleaseVersion GetValidatedSdkVersion(string versionOption, stri
             else
             {
                 var manifests = new SdkDirectoryWorkloadManifestProvider(dotnetPath, versionOption, userProfileDir).GetManifests();
-                if (!manifests.Any())
+                if (!manifests.Any() && checkIfFeatureBandManifestsExist)
                 {
-                    throw new GracefulException(string.Format(LocalizableStrings.NoManifestsExistForFeatureBand, versionOption), isUserError: false);
+                    throw new GracefulException(string.Format(LocalizableStrings.NoManifestsExistForFeatureBand, versionOption), isUserError: false);
```
nagilson, line 32: Nit: remove whitespace

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -512,7 +538,22 @@ private async Task<(bool, string)> GetManifestPackageUpdate(SdkFeatureBand sdkFe
             }
         }
 
-        private string GetAdvertisingManifestPath(SdkFeatureBand featureBand, ManifestId manifestId) =>
+        private async Task<(bool, NuGetVersion)> GetPackageVersion(PackageId packageId, PackageSourceLocation packageSourceLocation = null, bool includePreview = false)
+        {
+            try
+            {
+                var latestVersion = await _nugetPackageDownloader.GetLatestPackageVersion(packageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreview);
+                return (true, latestVersion);
```
nagilson, line 546: @joeloff Do we use this pattern of returning (Success + Value) frequently, I would rather return just null and check that instead of having a separate boolean to check. Might want to add some exception or warning docstring that this can return null?

nagilson (changes requested):
In response to earlier comment, provided more targeted feedback

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
gkulin: this TODO is left over from earlier in my work. It's relevant anymore

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
gkulin: @nagilson  this is related to a change that Daniel and I made together when implementing fallback in a PR from before you started. we are continuing that same pattern of behavior here.

nagilson (approved):

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -512,7 +534,22 @@ private async Task<(bool, string)> GetManifestPackageUpdate(SdkFeatureBand sdkFe
...
+            }
+            catch
```
dsplaisted, line 544: I think we don't want to catch all exceptions here, just `NuGetPackageNotFoundException`

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -249,9 +249,31 @@ public async Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsyn
             {
                 try
                 {
+
```
dsplaisted, line 252: We should update `UpdateAdvertisingManifestAsync` so that it calls `GetManifestPackageDownloadsAsync` instead of having the logic duplicated.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -249,9 +249,31 @@ public async Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsyn
...
+                        (success, latestVersion) = await GetPackageVersion(newPackageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreviews);
+                        
+                        if (success)
+                        {
+                            downloads.Add(new WorkloadDownload(manifest.Id, newPackageId.ToString(), latestVersion.ToString()));
+                        }
+                    }
+                    if (!success)
+                    {
+                        _reporter.WriteLine(string.Format(LocalizableStrings.FailedToGetPackageManifestUrl, packageId));
```
dsplaisted, line 275: Similarly to how we switched to the "AdManifestPackageDoesNotExist" string that doesn't say something failed when updating the advertising manifests, I think we should reword this string so it doesn't say "Failed".

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -43,17 +43,20 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
...
 
             var sdkVersionBand =
                 $"{sdkVersionParsed.Major}.{sdkVersionParsed.Minor}.{Last2DigitsTo0(sdkVersionParsed.Build)}";
-
+           
+            
             _sdkRootPath = sdkRootPath;
             _sdkVersionBand = sdkVersionBand;
+            
+            //var knownManifestIdsFilePath = Path.Combine(_sdkRootPath, "sdk", "6.0.400-dev", "IncludedWorkloadManifests.txt");
```
dsplaisted, line 59: Looks like this can be deleted.

src/Tests/dotnet-workload-update.Tests/GivenDotnetWorkloadUpdate.cs
```diff
@@ -266,6 +266,21 @@ public void GivenWorkloadUpdateItPrintsDownloadUrls()
...
+        {
+            var mockWorkloadIds = new WorkloadId[] { new WorkloadId("wasm-tools") };
+            var parseResult = Parser.Instance.Parse(new string[] { "dotnet", "workload", "update", "--source", @"C:\src\nugetPackages", "--print-download-link-only", "--sdk-version", updateVersion, "--from-previous-sdk"});
+            (_, var command, _, _, _, _) = GetTestInstallers(parseResult, installedWorkloads: mockWorkloadIds, includeInstalledPacks: false);
+
+            command.Execute();
+
+            _reporter.Lines.Should().Contain("==allPackageLinksJsonOutputStart==");
+            _reporter.Lines.Should().Contain(@"C:\\src\\nugetPackages\\microsoft.net.workload.mono.toolchain.manifest-6.0.400.6.0.5.nupkg");
+            _reporter.Lines.Should().Contain(@"C:\\src\\nugetPackages\\microsoft.net.runtime.monotargets.sdk.7.0.0.nupkg");
```
dsplaisted, line 281: This isn't going to work in CI.  Did you mean to check this test in?

src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
```diff
@@ -153,7 +161,7 @@ protected async Task<List<WorkloadDownload>> GetDownloads(IEnumerable<WorkloadId
 
         protected IEnumerable<WorkloadId> GetInstalledWorkloads(bool fromPreviousSdk)
         {
-            var currentFeatureBand = new SdkFeatureBand(_sdkVersion);
+            var currentFeatureBand = new SdkFeatureBand(_installedFeatureBand.ToString());
```
joeloff, line 164: I wonder if you need the `_currentFeatureBand` here, you can probably just use the `_installedFeatureBand` instead

joeloff (approved):

marcpopMSFT:
Approved offline, fingers crossed.

PR #26427, opened 2022-07-04
Support for msbuild/build event

rokonec:
Support for telemetry emitted by https://github.com/dotnet/msbuild/pull/7778

This telemetry is supposed to be sent with every build, including base measurements and properties usable to understand usage of MSBuild.

This PR also fixes bug causing not sending telemetry from msbuild - introduced during changes of SDK calling msbuild.dll inproc.

NOTE: Although it is related to MSBuild server it could be merged into SDK before MSBuild server is released.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
+++ b/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
@@ -8,7 +8,9 @@ using Microsoft.DotNet.Cli;
 using Microsoft.DotNet.Cli.Telemetry;
 using Microsoft.DotNet.Configurer;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
 
 namespace Microsoft.DotNet.Tools.MSBuild
 {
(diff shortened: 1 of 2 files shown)


baronfel (approved):
Seems pretty straightforward - glad that it was easy to plumb through here.

src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
```diff
@@ -107,15 +111,47 @@ internal static void FormatAndSend(ITelemetry telemetry, TelemetryEventArgs args
 
                 telemetry.TrackEvent(newEventName, maskedProperties, measurements: null);
             }
+            else if (args.EventName == BuildTelemetryEventName)
+            {
+                var newEventName = $"msbuild/{BuildTelemetryEventName}";
+                Dictionary<string, string> properties = new Dictionary<string, string>(args.Properties);
+                Dictionary<string, double> measurements = new Dictionary<string, double>();
+
+                string[] toBeHashed = new[] { "ProjectPath" };
```
baronfel, line 120: the Target property should be hashed as well - users can and do define custom MSBuild targets that could reveal PII

rokonec, line 120:
```suggestion
                string[] toBeHashed = new[] { "ProjectPath", "BuildTarget" };
```

baronfel:
Anything blocking us from merging this @rokonec?

marcpopMSFT (approved):

PR #26439, opened 2022-07-05
Remoev the build retry and increase the test retry

marcpopMSFT:


baronfel:
To be honest that face was probably me being angry at my laptop for freezing mid-call :D

baronfel (approved):

PR #26455, opened 2022-07-06
MSBuild 17.4.0-preview-22356-01

rainersigwald:
Manual update because flow is blocked on Arcade updates dotnet/arcade#9881.

    darc update-dependencies --id 140643


rainersigwald:
Ah, never mind, the packages didn't get pushed so this will fail.

PR #26458, opened 2022-07-06
Add --info Option to Workload Base Command For Future Triage

### LINKED ISSUE #22148: Add additional details on the workload version in dotnet workload --list

nagilson:
Part 2/2 for adding version info for workloads. (Other part was merged in to add this info to list.) 
Resolves https://github.com/dotnet/sdk/issues/22148#issuecomment-1168005944  See sample output here: 

Current Style after feedback:

image

OLD Style:

image

It also shows that using --info on sub commands doesn't work and the old error message still applies if --info is not provided.

Regarding the tests: Talking to @joeloff and I'll add a manual test because testing the actual output would require us to engineer many mock classes for something that is relatively static and unlikely to frequently change or be influenced by other code changes.


src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
```diff
@@ -15,14 +20,68 @@ internal static class WorkloadCommandParser
...
         }
 
+        internal static void ShowWorkloadsInfo(IWorkloadInfoHelper workloadInfoHelper = null, IReporter reporter = null)
+        {
+            workloadInfoHelper ??= new WorkloadInfoHelper();
+            IEnumerable<WorkloadId> installedList = workloadInfoHelper.InstalledSdkWorkloadIds;
+            InstalledWorkloadsCollection installedWorkloads = workloadInfoHelper.AddInstalledVsWorkloads(installedList);
+            reporter ??= Cli.Utils.Reporter.Output;
+
+            if(!installedList.Any())
```
joeloff, line 38: missing space

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
```diff
@@ -15,14 +20,68 @@ internal static class WorkloadCommandParser
...
+            {
+                reporter.WriteLine(CommonStrings.NoWorkloadsInstalledInfoWarning);
+                return;
+            }
+
+
+            foreach (var workload in installedWorkloads.AsEnumerable())
+            {
+                reporter.WriteLine("\n");
+                reporter.WriteLine(CommonStrings.WorkloadIdColumn + " : [" + workload.Key + "]");
```
joeloff, line 48: Extra space between the `:` and value on the left. 

image

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
```diff
@@ -15,14 +20,68 @@ internal static class WorkloadCommandParser
...
+                var workloadManifest = workloadInfoHelper.WorkloadResolver.GetManifestFromWorkload(new WorkloadId(workload.Key));
+                var workloadFeatureBand = new WorkloadManifestInfo(
+                    workloadManifest.Id,
+                    workloadManifest.Version,
+                    Path.GetDirectoryName(workloadManifest.ManifestPath)!).ManifestFeatureBand;
+
+                reporter.Write(CommonStrings.WorkloadManfiestVersionColumn + ":");
+                reporter.WriteLine("\t" + workloadManifest.Version + "/" + workloadFeatureBand);
+
+                reporter.Write(CommonStrings.WorkloadManifestPathColumn + ":");
```
joeloff, line 62: I'm a bit on the fence with the inline tabs. Can we use interpolated strings with column widths or maybe a `PrintableTable` without column headings, that way the formatting remains more consistent, especially if the actual spaces in tabs varies between platforms

nagilson: I don't think we should use a `printabletable`, I used that initially, but:
a) it doesn't match `dotnet --info` even if it matches `dotnet workload list`
b) the manifest path is so long that the table becomes unreadable unless the command line is full screen and on a huge monitor

Interpolated strings though I advocate for!

nagilson: Please see committed fixes.

PR #26508, opened 2022-07-11
Merge pull request #22974 from dotnet/marcpopMSFT-updateappinsights

marcpopMSFT:
Update application insights version


joeloff (approved):
AppInsights is moving away from pure instrumentation keys to connection strings. We should look at updating that at some point

marcpopMSFT:
> AppInsights is moving away from pure instrumentation keys to connection strings. We should look at updating that at some point

Perhaps that's a .net 8 change. I'm doing this in servicing so that we can ensure our 6.0 releases are relatively new so we continue to get support.

PR #26556, opened 2022-07-13
Exclude VS references from source build

### LINKED ISSUE #25616: VisualStudioWorkloads is not source-build compatible
VisualStudioWorkloads requires a reference to Microsoft.VisualStudio.Setup.Configuration.Interop which is not something currently source-built therefore it is considered a prebuilt.  Looking at the [PR that introduce the reference](https://github.com/dotnet/sdk/commit/7e0c4280a712ce68ab2105c07a0ceb6a05609725), it looks like this code path is only used on Windows.  Is is possible to conditionally compile this code for Windows?

cc @joeloff

joeloff:
Fixes #25616


crummel (approved):
Looks like this removes the bits in question.  Thanks!

MichaelSimons:
@joeloff, I see this went into [release/7.0.1xx-preview7](https://github.com/dotnet/sdk/tree/release/7.0.1xx-preview7), will this automatically flow into main?

joeloff:
It should, but I'll double check

joeloff:
ok, I need to port this

Issue #23905, opened 2022-02-12
[API Compat] add rule that ensures there is not attribute differences in between sides


safern:
This rule checks attributes on public members and gives an error if there is a miss match on the attributes. This rule checks attributes on:
- Types (Class, Struct, Enum, Interface, Record, Delegate)
- Fields
- Properties
- Methods
- Event
- ReturnValue
- Constructor
- Generic Parameters

Should we check for assembly attributes maybe as opt-in?

It should check for the following:
1. Attribute does not exist in the contract or implementation.
2. Attribute arguments does not match in both the contract and implementation.
    2.1 For flag arguments (i.e AtrributeTargets) we should decide whether expanding or removing a flag is breaking or not.
    2.2 Also there could be arguments that are string arguments, that if changed shouldn't matter w.r.t compatibility. We should allow setting this via a rule setting somehow. 

We should have a list of attributes that are ignored by default but also we should allow a Rule setting to disable attribute diffing for a list of attributes where libraries can add attributes they don't care about compat, i.e `EditorBrowsableAttribute`. 

Also, we should be smart about ignoring compiler generated attributes that don't matter in compatibility.

Issue #25760, opened 2022-06-01
[ApiCompat] Enable nullable reference types


smasher164:
Enabling it library-wide currently shows some errors, so we should work towards this as a goal.
We can at least start with enabling it for the rules.

ViktorHofer:
https://github.com/dotnet/sdk/pull/25765 enables nullability for the other layers (PackageValidation and Compatibility). I agree that we should start with the ApiCompatibility rules and then enable it for the other types as well.

PR #26616, opened 2022-07-15
ApiCompat: Add ValidateAssemblies msbuild task and add global tool

### LINKED ISSUE #18677: [API Compat] Add MSBuild task to run API Compat standalone
We should add the MSBuild Task to the `Microsoft.DotNet.Compatibility` package.

I have a prototype here which includes targets and props, it also accepts c# source as an input for the left hand side of the comparison: https://github.com/safern/apicompat/tree/main/src/Microsoft.DotNet.ApiCompatibility.Tasks

cc: @ericst @terrajobst @Anipik

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/18677

Please review by commits.

This change adds a global tool frontend that makes invoking ApiCompat's ValidateAssemblies and ValidatePackage functionality outside of msbuild possible. In addition to that, the ValidateAssemblies feature is now also exposed via an msbuild task.

The commits are grouped by usefulness to review:
- Commit 1 and 2 are just updating resx and sln(f) files.
- Commit 3 adds the transport package only files for ValidateAssemblies for a ref<-->impl tuple and exposes the ValidateAssemblies task in the sdk.
- Commit 4 just updates the project files
- **Commit 5 adds the frontends (task & console + shared)**
- **Commit 6 adds the ApiCompatRunner and related files to the ApiCompatibility layer**
- **Commit 7 makes the required changes in the PackageValidation layer**
- **Commit 8 adds tests and updates the existing ones.**
- Commit 9 updates the CODEOWNERS file

https://github.com/dotnet/runtime/pull/72464 consumes this change and demonstrates that there aren't any regressions in dotnet/runtime.

### ValidateAssemblies

Features:
- Compare a set of given left assemblies with a set of given right assemblies
- MSBuild task and global tool command
- The task is exposed directly in the SDK
- Additional targets in the transport package to fully replace arcade's ApiCompat
- Able to run in an inner or outer build for max perf
- Fully incremental, the comparison is only triggered if the inputs or outputs changed.
- Uses the same CompatibilitySuppression file format as PackageValidation
- Optionally transforms the left and right input assembly paths that are emitted into the suppression file to avoid environment specific paths.

In discussions with @ericstj we agreed on that it doesn't make sense to expose the ValidateAssemblies targets in the SDK and only include them in the transport package for core stack consumption (i.e. dotnet/runtime, wcf, aspnetcore, etc.).

Differently than arcade's apicompat msbuild targets, these targets allow to run ApiCompat either in an inner or outer build, based on if the project multi-targets. This speeds up the overall time spent in ApiCompat as there's only a single invocation instead of n invocations for each inner build.

image

We don't want to enable the `ValidateAssemblies` task in dotnet/runtime until https://github.com/dotnet/sdk/issues/23905 is implemented which @smasher164 and I are currently working on.


joperezr:
Hey @ViktorHofer sorry I didn't get a chance to look at this today. Will review tomorrow.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/IMetadataStreamProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
```
smasher164, line 4: so are we only doing this per-file for the time being?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Runner/ApiCompatWorkItemTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        public void Ctor_ValidArguments_PropertiesSet()
+        {
+            ApiCompatRunnerOptions apiCompatOptions = new(enableStrictMode: true, isBaselineComparison: true);
+            MetadataInformation left = new("A.dll", @"ref\netstandard2.0\A.dll");
+            MetadataInformation right = new("A.dll", @"lib\netstandard2.0\A.dll");
+
+            ApiCompatRunnerWorkItem workItem = new(left, apiCompatOptions, right);
+
+            Assert.Equal(apiCompatOptions, workItem.Options);
+            //Assert.Equal(metadata, workItem.Left);
```
smasher164, line 23: extra comment

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Abstractions/IMetadataStreamProvider.cs
ViktorHofer: Only for `ApiCompatibility.csproj` which doesn't yet set `<Nullable>enable</Nullable>`. https://github.com/dotnet/sdk/issues/25760 tracks enabling that for the entire project.

smasher164 (approved):

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/Microsoft.DotNet.ApiCompat.Task.csproj
```diff
@@ -9,47 +9,38 @@
...
-      <PackagePath>%(RecursiveDir)%(Filename)%(Extension)</PackagePath>
-    </None>
-    <None Include="..\..\Tasks\Microsoft.NET.Build.Tasks\targets\Microsoft.NET.Compatibility.Common.targets" Pack="true">
-      <Link>build/Microsoft.NET.Compatibility.Common.targets</Link>
-      <PackagePath>build/%(Filename)%(Extension)</PackagePath>
-    </None>
   </ItemGroup>
 
   <ItemGroup>
+    <None Include="**/*.props;**/*.targets" Pack="true" PackagePath="%(RecursiveDir)%(Filename)%(Extension)" />
```
joperezr, line 38: NIT: we should unify to use Windows-style directory separators.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
```diff
@@ -0,0 +1,126 @@
...
+
+using System;
+using System.IO;
+using Microsoft.Build.Framework;
+using Microsoft.NET.Build.Tasks;
+using Microsoft.DotNet.ApiCompatibility.Logging;
+
+namespace Microsoft.DotNet.ApiCompat.Task
+{
+    public class ValidateAssembliesTask : TaskBase
```
joperezr, line 12: Can we please add triple slash docs for these two tasks?

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        [Required]
+        public string? RoslynAssembliesPath { get; set; }
+
+        [Required]
+        public string[]? LeftAssemblies { get; set; }
```
joperezr, line 18: Should we allow null for left, right and RoslynAssemblies sets?

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        protected override void ExecuteCore()
+        {
+            Func<ISuppressionEngine, MSBuildCompatibilityLogger> logFactory = (suppressionEngine) => new(Log, suppressionEngine);
+            ValidateAssemblies.Run(logFactory,
+                GenerateCompatibilitySuppressionFile,
+                CompatibilitySuppressionFilePath,
+                NoWarn,
+                LeftAssemblies!,
```
joperezr, line 64: This is a parameter that gets passed in, if the calling API doesn't take null then should we be checking for null before calling Run()?

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
```diff
@@ -0,0 +1,126 @@
...
+        private static (string CaptureGroupPattern, string ReplacementString)[]? ParseTransformationPattern(ITaskItem[]? transformationPatterns)
+        {
+            if (transformationPatterns == null)
+                return null;
+
+            var patterns = new (string CaptureGroupPattern, string ReplacementPattern)[transformationPatterns.Length];
+            for (int i = 0; i < transformationPatterns.Length; i++)
+            {
+                string captureGroupPattern = transformationPatterns[i].ItemSpec;
+                string replacementString = transformationPatterns[i].GetMetadata("ReplacementString");
```
joperezr, line 113: NIT: This should probably be a const

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
```diff
@@ -0,0 +1,106 @@
...
+        public string? PackageTargetPath { get; set; }
+
+        [Required]
+        public string? RoslynAssembliesPath { get; set; }
+
+        public string? RuntimeGraph { get; set; }
+
+        public string? NoWarn { get; set; }
+
+        public bool RunApiCompat { get; set; }
```
joperezr, line 24: When would you set this to false? Should we default it to `true`?

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        public string? CompatibilitySuppressionFilePath { get; set; }
+
+        public ITaskItem[]? PackageAssemblyReferences { get; set; }
+
+        public ITaskItem[]? BaselinePackageAssemblyReferences { get; set; }
+
+        public override bool Execute()
+        {
+            RoslynResolver.Register(RoslynAssembliesPath);
```
joperezr, line 42: NIT: this common registration could be abstracted to a new base class if we wanted to.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
joperezr: Oh, I think I recall we discussed this in the past and you said MSBuild would ensure it is not null.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
```diff
@@ -0,0 +1,108 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Jab;
```
joperezr, line 7: Can you explain this new dependency? Is this what is getting us Dependency Injection? If so, do you think it is worth it to take on this new dependency on the sdk for an abstraction that strictly speaking doesn't need to be there? I agree it makes testing simpler and it decouples your APIs, but I wonder if it's worth the risk of adding an external dependency for this.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/Helper.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.DotNet.ApiCompatibility.Logging;
+
+namespace Microsoft.DotNet.ApiCompat
+{
+    internal static class Helper
```
joperezr, line 9: NIT: Should this be more like `SuppressionFileHelper`

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/RegexStringTransformer.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        /// <param name="input">The input string to transform.</param>
+        /// <returns>Returns the transformed input string. If the matches weren't successful, returns the untouched input.</returns>
+        public string Transform(string input)
+        {
+            string current = input;
+            foreach ((Regex Regex, string ReplacementString) pattern in _patterns)
+            {
+                current = pattern.Regex.Replace(current, pattern.ReplacementString);
+            }
```
joperezr, line 49:
```suggestion
            foreach ((Regex Regex, string ReplacementString) in _patterns)
            {
                current = Regex.Replace(current, ReplacementString);
            }
```
If you are doing deconstruction you don't need to reference via `pattern`

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/RegexStringTransformer.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.ApiCompat
+{
+    /// <summary>
+    /// A regex string transformer that transforms an input string via a regex pattern to a replacement string.
```
joperezr, line 9: This isn't technically accurate, in reality what this is doing is to transform an input string against _**a set**_ of regex patterns and replacement strings, as opposed to just a single one.

joperezr: I wonder if we should instead have the pattern and replacement string be part of each left and right item themselves, in case you want to special case and only apply a single regex transform for a very specific left, but not the rest.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
```diff
@@ -0,0 +1,144 @@
...
+            bool enableStrictMode,
+            string[][]? leftAssembliesReferences,
+            string[][]? rightAssembliesReferences,
+            bool createWorkItemPerAssembly,
+            (string CaptureGroupPattern, string ReplacementString)[]? leftAssembliesTransformationPatterns,
+            (string CaptureGroupPattern, string ReplacementString)[]? rightAssembliesTransformationPatterns)
+        {
+            // Configure the suppression engine. Ignore the passed in suppression file if it should be generated and doesn't yet exist.
+            string? suppressionFileForEngine = generateSuppressionFile && !File.Exists(suppressionFile) ? null : suppressionFile;
+            Func<ISuppressionEngine> suppressionEngineFactory = () => new SuppressionEngine(suppressionFileForEngine, noWarn, generateSuppressionFile);
```
joperezr, line 54: >suppressionEngineFactory [](http://example.com/codeflow?start=37&length=24)

I'm not sure I understand why this isn't fetched via the serviceProvider like the rest.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            // Initialize the service provider
+            ValidateAssembliesServiceProvider serviceProvider = new(logFactory, suppressionEngineFactory);
+
+            IApiCompatRunner apiCompatRunner = serviceProvider.GetService<IApiCompatRunner>();
+            ApiCompatRunnerOptions apiCompatOptions = new(enableStrictMode);
+
+            // Optionally provide a string transformer if a transformation pattern is passed in.
```
joperezr, line 62: As suggested on the offline code review, I think we should have a fallback either here or in the targets to use a regex pattern that matches a vanilla project's output location.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        public bool EnableStrictMode { get; set; }
+
+        public string[]? LeftAssembliesReferences { get; set; }
+
+        public string[]? RightAssembliesReferences { get; set; }
+
+        public string? SemaphoreFile { get; set; }
+
+        public bool CreateWorkItemPerAssembly { get; set; }
```
joperezr, line 37: I forgot what this one is supposed to mean.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
```diff
@@ -2,20 +2,54 @@
...
+using Microsoft.DotNet.ApiCompatibility.Runner;
 using NuGet.ContentModel;
 using NuGet.Frameworks;
 
 namespace Microsoft.DotNet.PackageValidation
 {
     internal static class ApiCompatRunnerExtensions
     {
-        public static void QueueApiCompatFromContentItem(this ApiCompatRunner apiCompatRunner, string packageId, ContentItem leftItem, ContentItem rightItem, string header, bool isBaseline = false)
+        public static void QueueApiCompatFromContentItem(this IApiCompatRunner apiCompatRunner,
```
joperezr, line 14: Can we add tripple slash docs for this type? For example, it would be good to mention that the reason why these methods are defined here as opposed to on ApiCompatRunner directly is that this includes package-specific logic that we don't want to push to the commonly used `Microsoft.DotNet.ApiCompatibility`

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
joperezr: Oh got it, I see where this is used.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
ViktorHofer: This was implemented as is by Anirudh and the target that invokes this task already defaults to true. If we would like to change this behavior, let's do that in a separate change.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
ViktorHofer: Right. I considered that but I chose this model over a new base class.

src/Compatibility/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
ViktorHofer: This is an internal only type which we don't triple slash document as they don't appear in the compiler produced xml file. I think it makes sense to add a comment indicating why this type exists but I wouldn't want to comment every possible method parameter. I will document the type itself.

joperezr: That's fine by me. I usually go with triple slash comments as that also helps with intellisense for the callers (even though they are internal) all relevant information is right there for them. But the primarily ask is for our code to be documented in a way that it facilitates reading the code from an external's perspective, and for that purpose, I think adding any sort of documentation to most of the new code is always valuable.

ViktorHofer: Absolutely agree on that. It wasn't obvious to me when I first looked into this codebase why this extension class existed. A code comment would have helped. In general there were many places that would have benefited from more inline comments.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
ViktorHofer: RoslynAssembliesPath doesn't need to be provided, it's just a mechanism to load a different roslyn version, hence its null today. LeftAssemblies and RightAssemblies are marked as required, hence they must be provided - msbuild checks for that.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/RegexStringTransformer.cs
ViktorHofer: I considered that but I chose the existing implementation for couple of reasons:
1. Passing in and parsing multiple transformation patterns for every left and right assembly is hard (think about the non msbuild usage).
2. In the msbuild world, in most cases this runs as part of a project's outer build which has consistent output paths for inner builds. In the console case, you most likely want to pass in a directory that contains a list of assemblies for which you wouldn't need a per assembly transformation pattern. I don't think that you would ever want to pass in assemblies that reside under completely different output paths and then check-in the generated suppression file.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
ViktorHofer: That makes sense for the future but at this point we don't want to expose the ValidateAssemblies targets publicly, they will only reside in the transport package. We should see if it's possible to add a default in the future when this logic will be exposed more publicly.

What makes this problematic is that the counterpart frontend (console / global tool) doesn't have any information of a project's output path and hence will always require a manual user input.

I filed https://github.com/dotnet/sdk/issues/26860 to track that.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
ViktorHofer: Actually this was a one line change so I included it in this PR. This should have been true to begin with.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
ViktorHofer: Sure. [Jab](https://github.com/pakrym/jab) is a compile time source generator (AOT) which doesn't use any reflection and hence is a build time only dependency (it's marked as PrivateAssets="all"). Jab is already part of our source build and because of that doesn't introduce a prebuilt. The author is a previous .NET team member.

The source generator creates a composition file that just wraps the instantiation of the various types that are required by other types.

> If so, do you think it is worth it to take on this new dependency on the sdk for an abstraction that strictly speaking doesn't need to be there? I agree it makes testing simpler and it decouples your APIs, but I wonder if it's worth the risk of adding an external dependency for this.

Yes, I think this component helps tremendously with managing dependencies inside ApiCompat without sacrificing runtime performance. I reviewed the [generated code](https://github.com/pakrym/jab/blob/main/README.md#jab-compile-time-dependency-injection) and there's nothing that stands out in terms of risk or perf penalty.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
joperezr: >RoslynAssembliesPath doesn't need to be provided

Then why it's marked as Required?

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
joperezr: I guess my question was more, why is it there to begin with? what is the use case scenario where you want to queue all api compat runs but don't actually run?

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
joperezr: > I think this component helps tremendously with managing dependencies inside ApiCompat

I'd be interested in getting more details around this.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
```diff
@@ -0,0 +1,167 @@
...
+        public string[]? LeftAssemblies { get; set; }
+
+        /// <summary>
+        /// The assemblies that act as the implementation.
+        /// </summary>
+        [Required]
+        public string[]? RightAssemblies { get; set; }
+
+        /// <summary>
+        /// The path to the roslyn assemblies that should be loaded.
```
joperezr, line 27: This kind of hints me to believe that this should be a required argument, yet it is in reality optional. Either remove the Required attribute, specify more info on the comment saying we allow null, or both.

src/Compatibility/Microsoft.DotNet.ApiCompat.Tool/Program.cs
```diff
@@ -0,0 +1,286 @@
...
+using Microsoft.DotNet.ApiCompatibility.Logging;
+
+namespace Microsoft.DotNet.ApiCompat.Tool
+{
+    class Program
+    {
+        static int Main(string[] args)
+        {
+            // Global options
+            Option<string?> suppressionFileOption = new("--suppression-file",
```
joperezr, line 18: NIT: perhaps add a comment that mentions that this and the task-params should be kept in sync. Just for future reference in case someone is updating one but forgets to update the other

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparerFactory.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
```
joperezr, line 4: Not exactly related to your change, but we should aim to remove this and instead annotate the whole library sooner rather than later. Having it partially like we do with this assembly is potentially dangerous due to the assumptions we make.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticIds.cs
```diff
@@ -25,6 +27,5 @@ public static class DiagnosticIds
         // Assembly loading ids
         public const string AssemblyNotFound = "CP1001";
         public const string AssemblyReferenceNotFound = "CP1002";
-        public const string SearchDirectoriesNotFoundForTfm = "CP1003";
```
joperezr, line 28: Why was this removed?

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Runner/ApiCompatRunner.cs
```diff
@@ -0,0 +1,173 @@
...
+
+#nullable enable
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiCompatibility.Abstractions;
+using Microsoft.DotNet.ApiCompatibility.Logging;
+
+namespace Microsoft.DotNet.ApiCompatibility.Runner
```
joperezr, line 12: Was anything changed in these files other than the interface abstraction? It's hard when the diff shows the whole files as this essentially means everything in compat must be re-reviewed.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/DiagnosticIds.cs
joperezr: Oh, it was moved to PackageValidation.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/RegexStringTransformer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Initializes the regex string transformer with the given capture group and replacement patterns.
+        /// </summary>
+        public RegexStringTransformer((string CaptureGroupPattern, string ReplacementString)[] rawPatterns)
+        {
+            _patterns = new (Regex Regex, string ReplacementString)[rawPatterns.Length];
+            for (int i = 0; i < rawPatterns.Length; i++)
+            {
+                _patterns[i] = (new Regex(rawPatterns[i].CaptureGroupPattern, RegexOptions.Compiled), rawPatterns[i].ReplacementString);
```
joperezr, line 34: Keep in mind you are taking in user-provided strings as patterns, that are going to work over user-provided inputs. This in Regex is usually an anti-pattern if not guarded well, as it could easily result in a ReDOS from a malicious user. At the very least, I would suggest passing in a matchTimeout to the Regex engines as a way to prevent at least some of this.

src/Tests/Microsoft.DotNet.ApiCompat.Tests/RegexStringTransformerTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            const string ReplacementPattern = "lib/$1/$2";
+            const string Input = @"C:\git\runtime\artifacts\bin\System.Linq\Debug\net7.0-android\System.Linq.dll";
+
+            string output = new RegexStringTransformer(CaptureGroupPattern, ReplacementPattern).Transform(Input);
+
+            Assert.Equal("lib/net7.0-android/System.Linq.dll", output);
+        }
+
+        [Fact]
+        public void Transform_MultiplePatterns_ReturnsExpected()
```
joperezr, line 61: I'd also add a test that forces catastrophic backtracking just to make sure that the engine succesfully times out. You can try pattern `^(\w+\s?)*$` with an expression like `An input string that takes a very very very very very very very very very very very long time!`

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Microsoft.DotNet.ApiCompatibility.Tests.csproj
```diff
@@ -1,15 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
     <TargetFrameworks>net472;$(ToolsetTargetFramework)</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">$(ToolsetTargetFramework)</TargetFrameworks>
     <OutputType Condition="'$(TargetFramework)' == '$(ToolsetTargetFramework)'">Exe</OutputType>
-
-    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
joperezr, line 9: Where is this set now?

joperezr:
I obviously didn't review line by line, but this looks good other than the comments I added.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Microsoft.DotNet.ApiCompatibility.Tests.csproj
ViktorHofer: This was required when all projects were built as part of the repo's main sln file. These days, source build uses a different solution file for source build and doesn't attempt to build these projects anymore. The other test projects don't have that property anymore so I cleaned this up here.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/ApiComparerFactory.cs
ViktorHofer: Agreed. We have https://github.com/dotnet/sdk/issues/25760 to track that.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs
ViktorHofer: Agree. The existing targets always provide the path to the roslyn assemblies but as you mentioned, in reality it isn't required. Applied your feedback.

src/Compatibility/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
ViktorHofer: That was added as part of https://github.com/dotnet/sdk/commit/d22bb74e1b73006135c000323fdab8225834b133. Unfortunately neither the PR description nor the commit message explains that. @ericstj would you know why the `RunApiCompat` was added to PackageValidation?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Microsoft.DotNet.ApiCompatibility.Tests.csproj
joperezr: Sounds good, might be worth just checking with @MichaelSimons and other source build folks just to ensure this won't break them if we aren't sure.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Runner/ApiCompatRunner.cs
ViktorHofer: The ones that I already touched on in previous meetings:
- Removed package specific log to make it usable outside of PackageValidation.
- Dependencies are moved into the constructor so that the type becomes unit testable (by mocking the external dependencies).

In addition to that:
- More diagnostics with low importance logged so that they appear in a binlog / high verbosity text log
- Strongly type work items and store them in a `HashSet<T>` instead of a `Dictionary<T,U>`.

src/Compatibility/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
ViktorHofer: > I'd be interested in getting more details around this.

Basically the general advantages of using dependency injection. https://betterprogramming.pub/the-6-benefits-of-dependency-injection-7802b207ec69 summarizes them quite nicely.

Most importantly, the following paragraph:

> Probably the main benefit of dependency injection is maintainability. If your classes are loosely coupled and follow the single responsibility principle — the natural result of using DI — then your code will be easier to maintain.

And this section nicely translates to the challenges with the previous implementation of the ApiCompatRunner which created dependencies just in time, which made it impossible to unit test the runner:

> If you pass dependencies to classes, it’s quite simple to pass in a test double implementation. If dependencies are hard-coded, it’s impossible to create test doubles for those dependencies.

PR #26627, opened 2022-07-15
Trim response file lines to prevent forwarding along whitespace-only tokens

baronfel:
After talking with @jonathanpeppers, we discovered that Xamarin-Android creates response files that have leading spaces on the lines. Previous response file handling understood this and did a `String.Trim()` call on the lines post-comment-stripping, which I lost when I implemented this handling using the new System.CommandLine mechanism for token replacement.

This re-introduces the trimming that should have never left!

### DIFF
--- a/src/Cli/dotnet/Parser.cs
+++ b/src/Cli/dotnet/Parser.cs
@@ -128,6 +128,8 @@ namespace Microsoft.DotNet.Cli
                                 return line.Substring(0, hashPos).Trim();
                             }
                         })
+                        // trim leading/trailing whitespace to not pass along dead spaces
+                        .Select(x => x.Trim())
                         // Remove empty lines
                         .Where(line => line.Length > 0);
                 replacementTokens = trimmedLines.ToArray();


jonathanpeppers:
Somehow our tests were passing in a response file with a leading & trailing space character such as:
```
 /p:AndroidSdkDirectory="/Users/runner/Library/Android/sdk" 
```
The leading space character caused:
```
MSBUILD : error MSB1008: Only one project can be specified.
```

The trailing space character puts a space in the MSBuild property, and we get other test failures from that.

For now I've removed all the weird spaces from our tests, we'll see how it goes afterward.

marcpopMSFT:
We're still working through getting the final build prepped for 6.0 and we're trying not to let more changes in (ie limit risk). If this can wait, that would be preferred.

jonathanpeppers:
Our test run looks green now, thanks.

We have a lot of integration tests that generate projects and build them. It was concerning because *all of them* were failing before.

jonathanpeppers:
We started seeing this one in .NET 7 builds here: https://github.com/xamarin/xamarin-android/pull/7217

baronfel:
6.0.4xx is open for merges now, right? @dotnet/dotnet-cli can I get a review on this to get it prepped?

nagilson (approved):
LGTM

PR #26652, opened 2022-07-17
Native AOT publish changes to be active after runtime changes

LakshanF:
Draft PR until the runtime changes arrive


LakshanF:
This was merged to #26715

PR #26687, opened 2022-07-19
Don't set ObjCopyName in testing

MichalStrehovsky:
This should not be needed after https://github.com/dotnet/runtime/pull/72208. Let's test the mainstream scenario.

Also flipping the if check to prepare for macOS testing at some point.


LakshanF (approved):
Thanks!

PR #26729, opened 2022-07-20
Detect invalid package locales and warn the user

### LINKED ISSUE #26711: Could not copy Error in docker linux container
---
**Issue moved from dotnet/sdk#12611**
- Please respond to @aortiz-msft.

---

_From @aortiz-msft on Thursday, July 23, 2020 5:50:04 PM_

---
**Issue moved from NuGet/Home#9787**
- Please respond to @ZergRushJoe.

---

_From @ZergRushJoe on Monday, July 13, 2020 12:56:23 AM_

## Details about Problem
I keep getting this error from my 'docker build .'. I'm getting the error with this .net core microsoft.analysisservices.adomdclient.netcore.retail.amd64. The Error that comes from it:

/usr/share/dotnet/sdk/3.1.301/Microsoft.Common.CurrentVersion.targets(4360,5): error MSB3021: Unable to copy file "/root/.nuget/packages/microsoft.analysisservices.adomdclient.netcore.retail.amd64/19.4.0.2-preview/lib/netcoreapp3.0/zh-HANS/Microsoft.AnalysisServices.AdomdClient.resources.dll" to "bin/Debug/netcoreapp3.1/zh-HANS/Microsoft.AnalysisServices.AdomdClient.resources.dll". Could not find a part of the path '/app/Processor/bin/Debug/netcoreapp3.1/zh-HANS/Microsoft.AnalysisServices.AdomdClient.resources.dll'

Don't know why it can't find that package. it restores just fine. I know it builds on windows which more then likly means I can never get it to work. The project it is in is an F# project.

Docker File
From node:12 AS web-build

COPY ./Web.UI ./

RUN npm i
RUN npm run test
RUN npm run buildProd

FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS backend-build

WORKDIR /app
COPY . ./
COPY --from=web-build /Processor/wwwroot ./Processor/wwwroot
RUN dotnet test
RUN dotnet publish -c Release -o publish

FROM mcr.microsoft.com/dotnet/core/aspnet:3.1
WORKDIR /app
COPY --from=backend-build /app/publish .
EXPOSE 80

baronfel:
Fixes #26711 

There's a long discussion but the gist is that we can detect situations where locale codes are mismatched and tell the user about it. We can even correct those that can be normalized via the BCL culture lookups.

Open questions:
* [x] what should we do when an invalid culture is found? We have two options I see immediately: include it with a warning, or exclude it. Is there a use case for invalid culture identifiers?
  *  decided to allow invalid locale identifiers - ICU data can change, and may not be consistent between build machine and execution machine. 
* [x] what code should we choose? 
  * took the next two open numbers
* [x] need to think about the messages a bit more - is the packageName@packageId format useful?
  * spoke with Jon Douglas and got the normal nuget output format for package names and versions.


baronfel:
I'm only marking this for review to see if the tests run - there are more open questions

baronfel:
Alright - there's only one major question we should answer - **what do we do with assets that are found using completely unknown locale identifiers**? We can drop them (currently implemented) or include them. Should we be forwards-compatible for locale identifiers? How often does this data change?

baronfel:
Country codes change based on an IANA registry: https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -851,4 +851,12 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1186: This project depends on Maui Essentials through a project or NuGet package reference, but doesn't declare that dependency explicitly. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
     <comment>{StrBegin="NETSDK1186: "}</comment>
   </data>
+  <data name="PackageContainsIncorrectlyCasedLocale" xml:space="preserve">
+    <value>Package {0} {1} has a resource with the locale '{2}'. This locale has been normalized to the standard format '{3}' to prevent casing issues in the build. Consider notifying the package author about this casing issue.</value>
+    <comment>Error code is NETSDK1187. 0 is a package name, 1 is a package version, 2 is the incorrect locale string, and 3 is the correct locale string.</comment>
+  </data>
+  <data name="PackageContainsUnknownLocale" xml:space="preserve">
+    <value>Package {0} {1} has a resource with the locale '{2}'. This locale is not known by .NET. Consider notifying the package author about this casing issue.</value>
```
dsplaisted, line 859:
```suggestion
    <value>NETSDK1188: Package {0} {1} has a resource with the locale '{2}'. This locale is not recognized by .NET. Consider notifying the package author that it appears to be using an invalid locale.</value>
```

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -851,4 +851,12 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1186: This project depends on Maui Essentials through a project or NuGet package reference, but doesn't declare that dependency explicitly. To build this project, you must set the UseMauiEssentials property to true (and install the Maui workload if necessary).</value>
     <comment>{StrBegin="NETSDK1186: "}</comment>
   </data>
+  <data name="PackageContainsIncorrectlyCasedLocale" xml:space="preserve">
+    <value>Package {0} {1} has a resource with the locale '{2}'. This locale has been normalized to the standard format '{3}' to prevent casing issues in the build. Consider notifying the package author about this casing issue.</value>
```
dsplaisted, line 855:
```suggestion
    <value>NETSDK1187: Package {0} {1} has a resource with the locale '{2}'. This locale has been normalized to the standard format '{3}' to prevent casing issues in the build. Consider notifying the package author about this casing issue.</value>
```

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
```diff
@@ -117,6 +117,80 @@ public void It_does_not_error_on_duplicate_package_names()
...
+  }
+}".Replace("`", "\"");
+            File.WriteAllText(projectAssetsJsonPath, assetsContent);
+            var task = InitializeTask(out _);
+            task.ProjectAssetsFile = projectAssetsJsonPath;
+            task.TargetFramework = "net7.0";
+            var writer = new CacheWriter(task);
+            writer.WriteToCacheFile();
+            var logger = (task.BuildEngine as Msbuild.Tests.Utilities.MockEngine).MockLogger;
+            var invalidContextMessages = logger.BuildMessageEvents.Where(msg => msg.Code == "NETSDK12346");
```
dsplaisted, line 153: Update to correct error code.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
```diff
@@ -117,6 +117,80 @@ public void It_does_not_error_on_duplicate_package_names()
...
+        `type`: `package`,
+        `compile`: {
+          `lib/netstandard2.0/JavaScriptEngineSwitcher.Core.dll`: {}
+        },
+        `runtime`: {
+          `lib/netstandard2.0/JavaScriptEngineSwitcher.Core.dll`: {}
+        },
+        `resource`: {
+          `lib/netstandard2.0/ru-ru/JavaScriptEngineSwitcher.Core.resources.dll`: {
+            `locale`: `what is this even`
```
dsplaisted, line 139: Looks like the test names and data are backwards, the test name here seems to indicate that the locale for this test should be `ru-ru`, and the current value swapped to the subsequent test.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
```diff
@@ -117,6 +117,80 @@ public void It_does_not_error_on_duplicate_package_names()
...
+  }
+}".Replace("`", "\"");
+            File.WriteAllText(projectAssetsJsonPath, assetsContent);
+            var task = InitializeTask(out _);
+            task.ProjectAssetsFile = projectAssetsJsonPath;
+            task.TargetFramework = "net7.0";
+            var writer = new CacheWriter(task);
+            writer.WriteToCacheFile();
+            var logger = (task.BuildEngine as Msbuild.Tests.Utilities.MockEngine).MockLogger;
+            var invalidContextMessages = logger.BuildMessageEvents.Where(msg => msg.Code == "NETSDK12345");
```
dsplaisted, line 190: Update to correct error code.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -686,10 +686,10 @@ public CacheWriter(ResolvePackageAssets task)
                 }
                 else
                 {
-                    _compileTimeTarget = _lockFile.GetTargetAndThrowIfNotFound(_targetFramework, runtimeIdentifier: null); 
+                    _compileTimeTarget = _lockFile.GetTargetAndThrowIfNotFound(_targetFramework, runtimeIdentifier: null);
```
dsplaisted, line 689: Try to revert whitespace changes in this file (ideally with a rebase).

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1334,6 +1334,32 @@ private void WriteResourceAssemblies()
...
+                        // forms here. If the locale is mixed-case, that can cause issues on case-sensitive
+                        // file systems when the locale-specific assets are copied.
+                        try
+                        {
+                            var normalizedLocale = System.Globalization.CultureInfo.GetCultureInfo(locale).Name;
+                            _task.Log.Log(new (MessageLevel.HighImportance, 
+                                string.Format(Strings.PackageContainsIncorrectlyCasedLocale,
+                                    package.Name, package.Version.ToNormalizedString(), locale, normalizedLocale
+                                ),
+                                "NETSDK1187"));
```
dsplaisted, line 1347: It doesn't look like this checks to see whether the locale is actually incorrectly cased before emitting the warning.

Also, we should prefer calling LogWarning, which parses out the error code for us.

```suggestion
                            _task.Log.LogWarning(Strings.PackageContainsIncorrectlyCasedLocale,
                                    package.Name, package.Version.ToNormalizedString(), locale, normalizedLocale);
```

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1334,6 +1334,32 @@ private void WriteResourceAssemblies()
...
+                            locale = normalizedLocale;
+                        } catch (System.Globalization.CultureNotFoundException cnf)
+                        {
+                            _task.Log.Log(new(MessageLevel.HighImportance,
```
dsplaisted, line 1351: Also change this to a `LogWarning` or `LogError` call.

src/Tests/Msbuild.Tests.Utilities/MockEngine.cs
```diff
@@ -0,0 +1,494 @@
...
+     * task can log events.  It doesn't have to be the real Engine object, just
+     * something that implements the IBuildEngine4 interface.  So, we mock up
+     * a fake engine object here, so we're able to execute tasks from the unit tests.
+     * 
+     * The unit tests could have instantiated the real Engine object, but then
+     * we would have had to take a reference onto the Microsoft.Build.Engine assembly, which
+     * is somewhat of a no-no for task assemblies.
+     * 
+     **************************************************************************/
+    public sealed class MockEngine : IBuildEngine7
```
dsplaisted, line 35: Would the existing [`MockBuildEngine`](https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/Mocks/MockBuildEngine.cs) class work here?  If not, we should probably update that class rather than introducing an entirely new mock engine and logger class.

dsplaisted (approved):
Overall looks good.  Please make sure to do some end to end testing before merging (Ie building projects that reference packages that we know have non-normalized locales, as well as packages that have correct locales).

src/Tests/Msbuild.Tests.Utilities/MockEngine.cs
baronfel: The built-in was missing implementations for IBuildEngine4 for general use (and 5 in a few test cases), so I did that and extracted the various subs in a couple other tests into this centralized version.

baronfel:
Manual testing shows it working as expected:

> C:\Users\chethusk\OSS\dotnet-sdk\artifacts\bin\redist\Debug\dotnet\sdk\7.0.100-dev\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(267,5): warning NETSDK1187: Package JavaScriptEngineSwitcher.Core 3.3.0 has a resource with the locale 'ru-ru'. This locale has been normalized to the standard format 'ru-RU' to prevent casing issues in the build. Consider notifying the package author about this casing issue. [C:\Users\chethusk\OSS\dotnet-sdk\blah\blah.csproj]

baronfel:
One thing I'd like to call out - we've been fairly conservative about introducing new warnings/errors purely on updating the SDK. It typically requires updating a Warning or Analysis level (which by default is tied to TFM), which is a user-driven action.  That falls over a bit here because this bug can impact anyone on any TFM, but I just wanted to call out that we are breaking our own rule a bit here.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1334,6 +1334,33 @@ private void WriteResourceAssemblies()
...
+                        // Locales from packages can be free-form, so we normalize them to the standard
+                        // forms here. If the locale is mixed-case, that can cause issues on case-sensitive
+                        // file systems when the locale-specific assets are copied.
+                        try
+                        {
+                            var normalizedLocale = System.Globalization.CultureInfo.GetCultureInfo(locale).Name;
+                            if (normalizedLocale != locale)
+                            {
+                                _task.Log.LogWarning(
+                                    string.Format(Strings.PackageContainsIncorrectlyCasedLocale,
```
dsplaisted, line 1346: You don't need the call to `string.Format` here, `LogWarning` takes a list of args that it will pass to `string.Format`

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1334,6 +1334,33 @@ private void WriteResourceAssemblies()
...
+                                        package.Name, package.Version.ToNormalizedString(), locale, normalizedLocale
+                                ));
+                            }
+                            locale = normalizedLocale;
+                        } catch (System.Globalization.CultureNotFoundException cnf)
+                        {
+                            _task.Log.LogWarning(
+                                string.Format(
```
dsplaisted, line 1354: Same here, no need for `string.Format`.

dsplaisted (approved):

PR #26808, opened 2022-07-26
Get PublishRelease and PackRelease working in .*proj, Solutions 

### LINKED ISSUE #26732: PublishRelease=true in project file doesn't actually change the configuration to "Release" when running "dotnet publish"
RE #23551

Trying out `<PublishRelease>true</PublishRelease>` in a console project and it seems if I also set `<PublishAot>true</PublishAot>` then a debug build is performed rather than a release build.

HelloWorld.Console.csproj
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
    <PublishRelease>true</PublishRelease>
    <PublishAot>true</PublishAot>
  </PropertyGroup>
</Project>
```

Run `dotnet publish HelloWorld.Console -r win-x64 --self-contained -v -o .artifacts\HelloWorld.Console`

Output:
```
Publishing HelloWorld.Console: dotnet publish -r win10-x64 --self-contained
  Restored ~\HelloWorld.Console\HelloWorld.Console.csproj (in 72 ms).
C:\Program Files\dotnet\sdk\7.0.100-rc.1.22368.2\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(219
,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy [~\HelloWorld.Console\HelloWorld.Console.csproj]
  HelloWorld.Console -> ~\HelloWorld.Console\bin\Debug\net7.0\win10-x64\HelloWorld.Console
  .dll
  Generating compatible native code. To optimize for size or speed, visit https://aka.ms/OptimizeNativeAOT
     Creating library bin\Debug\net7.0\win10-x64\native\HelloWorld.Console.lib and object bin\Debug\net7.0\win10-x64\native\HelloWorld.Console.exp
  HelloWorld.Console -> ~\.artifacts\HelloWorld.Console\
```

`dotnet --info`
```
.NET SDK:
 Version:   7.0.100-rc.1.22368.2
 Commit:    a9c056cd39

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.100-rc.1.22368.2\

Host:
  Version:      7.0.0-rc.1.22366.5
  Architecture: x64
  Commit:       072eda8d6b

.NET SDKs installed:
  6.0.302 [C:\Program Files\dotnet\sdk]
  6.0.400-preview.22330.6 [C:\Program Files\dotnet\sdk]
  7.0.100-rc.1.22368.2 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.27 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.0-rc.1.22367.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.27 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0-preview.5.22301.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0-rc.1.22366.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.27 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.7 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.0-preview.5.22302.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.0-rc.1.22366.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:

Environment variables:
  Not set

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
```

@nagilson @richlander

nagilson:
Resolves https://github.com/dotnet/sdk/issues/26732 
Resolves part of https://github.com/dotnet/sdk/issues/27066
Note this doesn't solve the /t:Publish scenario.

Doc change: https://github.com/dotnet/docs/pull/30747

# Projects
Try:

Running `dotnet publish` `dotnet pack` and `dotnet publish foo.csproj` `dotnet pack foo.csproj` (or foo. any type of project besides a shproj). With `PackRelease` defined in the project or `PublishRelease` as `true`. Now, Release assets are produced instead of Debug assets. 

# Solutions
Try:
Running `dotnet publish` `dotnet pack` and `dotnet publish foo.sln` `dotnet pack foo.sln` for a solution which contains a project file with `PackRelease` defined in the project or `PublishRelease`. Try doing this without the environment variable set and with it set. It should produce release assets.

The code for solution files presents a 7% to 25% regression on publish performance on a solution for a large number of projects but only if opt-in. This enables `PublishRelease` and `PackRelease` to work when doing `dotnet publish` or `dotnet publish foo.sln` when a solution is targeted. The environment variable must be set to opt into this feature. If a user tries to use a solution with `PublishRelease` or `PackRelease` without the environment variable set, they will be warned that they need to set the environment variable but accept a performance overhead of doing so. This warning does not appear unless a user tries this scenario (publishing a solution with the properties set), and it does not occur with projects. 

Top level projects that conflict in `PublishRelease` or `PackRelease` inside of a solution should fail. The configuration value set in `PublishRelease` or `PackRelease` will propagate to any library or subproject based on the configuration set by a top level project property. 

We will need to document these details. 

# For Me
To Test:
- [x] The change picks up publish release and pack release when you put it in a .csproj and invoke `dotnet publish` or `dotnet publish foo.*proj`
- [x] If you have a folder with multiple projects or no projects and try to run, this doesn't change the MSBuild error message by catching it earlier on and silently continues even though it knows its wrong 

~~- [x] Solution files work and the change propagates to all of the subprojects if the top level project specifies `PublishRelease` or `PackRelease`~~
~~- [ ] If you publish or pack a solution with 3 class libraries each with their custom configuration it uses the library configuration based on depedency graph?~~
~~- [ ] Publishing a solution with hundreds of projects doesn't become significantly slower~~


nagilson:
> Will --configuration override this?

@KalleOlaviNiemitalo That's a good point -- it should override this. Thanks for asking that question, we shouldn't propagate the publishRelease `--configuration` if one is provided.

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -55,6 +64,57 @@ public static PublishCommand FromParseResult(ParseResult parseResult, string msb
...
+                    .Where(arg => File.Exists(arg) &&
+                        LikeOperator.LikeString(arg, "*.*proj", VisualBasic.CompareMethod.Text))
+                    .ToList()
+                    .FirstOrDefault();
+
+            if (string.IsNullOrWhiteSpace(potentialProject))
+            {
+                try
+                {
+                    potentialProject = MsbuildProject.GetProjectFileFromDirectory(Directory.GetCurrentDirectory()).Name;
```
nagilson, line 102: dotnet run would ideally use this code because there is actually code duplication between this and the msbuildproject and msbuildprojectlocator. but we won't make that change here.

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -55,6 +64,57 @@ public static PublishCommand FromParseResult(ParseResult parseResult, string msb
                 msbuildPath);
         }
 
+        public static string GetAutomaticConfigurationIfSpecified(
+            ParseResult parseResult,
+            string defaultedConfigurationProperty,
+            IEnumerable<string> slnOrProjectArgs,
+            Option<string> configOption)
+        {
+            ProjectInstance project = GetTargetedProject(parseResult, slnOrProjectArgs);
```
nagilson, line 73: all of this makes it easy to do for pack too, which i will demonstrate tmrw

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+            IEnumerable<string> calledArguments = parseResult.Tokens.Select(x => x.ToString());
+            IEnumerable<string> slnProjectAndCommandArgs = slnOrProjectArgs.Concat(calledArguments);
+            project = GetTargetedProject(slnProjectAndCommandArgs, defaultedConfigurationProperty);
+
+            if (project != null)
+            {
+                string releaseMode = "";
+                string releasePropertyFlag = project.GetPropertyValue(defaultedConfigurationProperty);
+                if (!string.IsNullOrEmpty(releasePropertyFlag))
+                    releaseMode = releasePropertyFlag == "true" ? "Release" : "Debug";
```
dsplaisted, line 139: We should do a case insensitive comparison here.  `TRUE` or `True` should work for the value.

Also, if the value is set but is false, we should not pass in a specific configuration.

```suggestion
                    releaseMode = releasePropertyFlag.Equals("true", StringComparison.OrdinalIgnoreCase) ? "Release" : "";
```

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            ProjectInstance project = null;
+
+            IEnumerable<string> calledArguments = parseResult.Tokens.Select(x => x.ToString());
+            IEnumerable<string> slnProjectAndCommandArgs = slnOrProjectArgs.Concat(calledArguments);
+            project = GetTargetedProject(slnProjectAndCommandArgs, defaultedConfigurationProperty);
+
+            if (project != null)
+            {
+                string releaseMode = "";
```
dsplaisted, line 136: I'd suggest renaming this to something like `configuration` or `configurationToUse`.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+                string releasePropertyFlag = project.GetPropertyValue(defaultedConfigurationProperty);
+                if (!string.IsNullOrEmpty(releasePropertyFlag))
+                    releaseMode = releasePropertyFlag == "true" ? "Release" : "Debug";
+
+                if (!ConfigurationAlreadySpecified(parseResult, project, configOption) && !string.IsNullOrEmpty(releaseMode))
+                    return new List<string> { $"-property:configuration={releaseMode}" };
+            }
+            return Array.Empty<string>();
```
dsplaisted, line 144: Nit: I think this is slightly better:

```suggestion
            return Enumerable.Empty<string>();
```

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -39,12 +44,16 @@ public static PublishCommand FromParseResult(ParseResult parseResult, string msb
                 "-property:_IsPublishing=true"
```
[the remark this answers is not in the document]
dsplaisted: I don't think we need `_IsPublishing` for those cases.  When we do the project evaluation, that should pick up the value for `PublishRelease` even if it's set in a Directory.Build.props file.

There might be other reasons we still need `_IsPublishing` though, I'm not sure.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackTool.targets
```diff
@@ -59,6 +59,10 @@ Copyright (c) .NET Foundation. All rights reserved.
         <PackagePath>%(_ResolvedFileToPublishWithPackagePath.PackagePath)</PackagePath>
       </TfmSpecificPackageFile>
     </ItemGroup>
+
+    <!-- Enable warning for trying to use PackRelease with a solution if env-var is not set.-->
+    <NETSdkWarning Condition="'$(PackRelease)' != '' and '$(SolutionExt)' == '.sln' and '$(ENABLE_P_RELEASE_SLN)' == ''"
```
dsplaisted, line 64: This is in the `PackTool` target, which I believe is only run when packing projects where `PackAsTool` is true.  So this won't apply to most cases of `dotnet pack`.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+                var projectData = TryGetProjectInstance(project.FilePath);
+                if (projectData == null)
+                    return;
+
+                if (projectData.GetPropertyValue("OutputType") == topLevelProjectOutputType)
+                {
+                    if (ProjectHasUserCustomizedConfiguration(projectData))
+                    {
+                        shouldReturnNull = true;
+                        state.Stop(); // We don't want to override Configuration if ANY project in a sln uses a custom configuration
```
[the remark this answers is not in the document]
nagilson: `state.Stop()` will make all threads stop running. This is because we found a project in the solution with a custom configuration that the user hardcoded. In that case, we do not want to respect `PublishRelease` or pack so we signal to the other threads to throw away their work as soon as possible, to exit and return to publishing.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+                sln = SlnFileFactory.CreateFromFileOrDirectory(slnPath);
+            }
+            catch (GracefulException)
+            {
+                return null; // This can be called if a solution doesn't exist. MSBuild will catch that for us.
+            }
+
+            List<ProjectInstance> configuredProjects = new List<ProjectInstance>();
+            HashSet<string> configValues = new HashSet<string>();
+            bool shouldReturnNull = false;
```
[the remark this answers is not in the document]
nagilson: I thought `Add` was an `atomic` operation and the order did not matter here, so didn't lock as it would decrease performance. Looks like I was wrong and `System.Collections.Concurrent` is the only thread-safe-lib. Oops, thank you!

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -523,6 +526,9 @@
   <data name="ToolSettingsUnsupportedRunner" xml:space="preserve">
     <value>Command '{0}' uses unsupported runner '{1}'."</value>
   </data>
+  <data name="TopLevelPublishConfigurationMismatchError" xml:space="preserve">
+    <value>Multiple top-level projects in the given solution contain conflicting PublishRelease or PackRelease values. To use conflicting configurations, leverage a Directory.build.props file instead of setting flags that don't match.</value>
```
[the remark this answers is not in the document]
nagilson: "Top-level" --> Project that has an entry point / main (though as of C#9 main isn't required, same idea)

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            return string.IsNullOrEmpty(potentialProject) ? null : TryGetProjectInstance(potentialProject);
+        }
+
+        /// <returns>The top-level project (first if multiple exist) in a SLN. Returns null if no top level project. Throws exception if two top level projects disagree
+        /// in the configuration property to check.</returns>
+        public override ProjectInstance GetSlnProject(string slnPath, string slnProjectConfigPropertytoCheck = "")
+        {
+            if (Environment.GetEnvironmentVariable("ENABLE_P_RELEASE_SLN") == null) // This has a performance overhead so don't do this unless opted in.
```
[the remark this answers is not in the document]
nagilson: This can also take effect for pack and not just publish, what is the best naming convention for that? Or do we want two separate ones? 
`DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS` 
`DOTNET_CLI_ENABLE_PACK_OR_PUBLISH_RELEASE_FOR_SOLUTIONS`

src/Cli/dotnet/commands/dotnet-publish/Program.cs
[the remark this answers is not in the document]
nagilson: It is the only way we have atm to make this only take effect for publish until we architect a solution for that. So going to keep for now.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+                    string configuration = projectData.GetPropertyValue(slnProjectConfigPropertytoCheck);
+                    if (!string.IsNullOrEmpty(configuration))
+                    {
+                        configuredProjects.Add(projectData); // we don't care about race conditions here
+                        configValues.Add(configuration);
+                    }
+                }
+            });
+
+            if (configuredProjects.Any() && configValues.Count > 1)
```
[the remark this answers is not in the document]
nagilson: Good point, let's not error in that case because it would be annoying.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                var projectData = TryGetProjectInstance(project.FilePath);
+                if (projectData == null)
+                    return;
+
+                if (projectData.GetPropertyValue("OutputType") == topLevelProjectOutputType)
+                {
+                    if (ProjectHasUserCustomizedConfiguration(projectData))
+                    {
+                        shouldReturnNull = true;
```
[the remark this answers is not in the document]
nagilson: Not currently. I don't think that is the desired behavior. If a user provides `-p:Configuration=Blah` but also opted in to `PublishRelease` by default, I don't think it would be a great experience for them to recieve a warning for providing `Blah` after opting into that default.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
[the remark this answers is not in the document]
nagilson: Also, thank goodness for RAII locks

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+                }
+                else if (Directory.Exists(arg)) // We should get here if the user did not provide a .proj or a .sln
+                {
+                    try
+                    {
+                        return TryGetProjectInstance(MsbuildProject.GetProjectFileFromDirectory(arg).FullName);
+                    }
+                    catch (GracefulException)
+                    {
+                        string potentialSln = Directory.GetFiles(arg, "*.sln", SearchOption.TopDirectoryOnly).FirstOrDefault();
```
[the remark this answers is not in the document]
nagilson: Good point, we should call out that nuance

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,55 @@
...
+            catch (Exception) // Catch failed file access, or invalid project files that cause errors when read into memory,
+            {
+                Reporter.Output.WriteLine(CommonLocalizableStrings.ProjectDeductionFailure.Yellow() + " " + projectPath.Yellow());
+            }
+            return null;
+        }
+
+        protected static bool IsValidProjectFilePath(string path)
+        {
+            return File.Exists(path) && LikeOperator.LikeString(path, "*.*proj", VisualBasic.CompareMethod.Text);
```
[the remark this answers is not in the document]
nagilson: There is `Directory.EnumerateFiles` which could do that but that has disk overhead over searching a directory, while this is just string manipulation logic. We could do a `regex` instead but I think `*.*proj` is much prettier.

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <returns>Creates a ProjectInstance if the project is valid, elsewise, fails..</returns>
+        protected static ProjectInstance TryGetProjectInstance(string projectPath)
+        {
+            try
+            {
+                return new ProjectInstance(projectPath);
+            }
+            catch (Exception) // Catch failed file access, or invalid project files that cause errors when read into memory,
+            {
+                Reporter.Output.WriteLine(CommonLocalizableStrings.ProjectDeductionFailure.Yellow() + " " + projectPath.Yellow());
```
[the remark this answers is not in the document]
nagilson: Gotcha!

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -420,6 +420,9 @@
   <data name="SpecifyAtLeastOneProjectToRemove" xml:space="preserve">
     <value>You must specify at least one project to remove.</value>
   </data>
+  <data name="ProjectDeductionFailure" xml:space="preserve">
+    <value>A problem with the project or solution prevented checking properties for Release.</value>
```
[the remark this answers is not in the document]
nagilson: Yep, this error sucks and was more of a debugging thing for me than anything. I'll remove it. I will add the actual exception details to any failures that call this.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,153 @@
...
+                return null; // This can be called if a solution doesn't exist. MSBuild will catch that for us.
+            }
+
+            List<ProjectInstance> configuredProjects = new List<ProjectInstance>();
+            HashSet<string> configValues = new HashSet<string>();
+            bool shouldReturnNull = false;
+
+            Parallel.ForEach(sln.Projects.AsEnumerable(), (project, state) =>
+            {
+                const string topLevelProjectOutputType = "Exe"; // Note that even on Unix when we don't produce exe this is still an exe, same for ASP
```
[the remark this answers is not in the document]
nagilson: Renaming to `executableProjectOutputType`

src/Assets/TestProjects/TestAppWithSlnUsingConflictingPublishReleaseConfigurations/Multiple/First.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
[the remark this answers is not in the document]
nagilson: The error for conflicting `PublishRelease` values only occurs if there's a conflict between two `top-level` projects. The lib is there to be evaluated first and show it doesn't trigger the error but this does. Could be better named.

src/Assets/TestProjects/TestAppWithSlnUsingPublishRelease/App.sln
```diff
@@ -0,0 +1,34 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26006.2
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "App", "App\App.csproj", "{7072A694-548F-4CAE-A58F-12D257D5F486}"
```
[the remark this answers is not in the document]
nagilson: Yes, the point was to make the test simpler but really it is silly to have the lib at all. Removing.

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        protected static bool IsValidProjectFilePath(string path)
+        {
+            return File.Exists(path) && LikeOperator.LikeString(path, "*.*proj", VisualBasic.CompareMethod.Text);
+        }
+
+        protected static bool ProjectHasUserCustomizedConfiguration(ProjectInstance project)
+        {
+            return project.GlobalProperties.ContainsKey("Configuration");
```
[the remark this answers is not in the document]
nagilson: No. It doesn't; I thought it covered that scenario. Thanks! I'm going to add a check for that higher up in the function calls because the project code is doing the same thing.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.PackTool.targets
nagilson: (All) of the `pack` targets seem to be in `NuGet.build.tasks`. So I'd need to raise a PR in NuGet. I'll wait for your response on the env name above.

nagilson:
@dsplaisted Thank you for reviewing this large thing. I responded to all of your comments and resolved everything I fixed. The remaining ~3 could be followed up. Then lmk if good to merge.

richlander:
Can you change the name of the ENV? It is pretty non-standard. Also, long ENVs are fine.

`ENABLE_P_RELEASE_SLN` -> `DOTNET_PUBLISH_RELEASE_SLN`

src/Cli/dotnet/commands/dotnet-publish/Program.cs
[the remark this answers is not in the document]
dsplaisted: Let's chat about this, I'm not sure I understand it all the way right now.

src/Cli/dotnet/ProjectLocator.cs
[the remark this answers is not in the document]
nil4: Would extracting the file extension for `path` and checking if it ends with `proj` achieve the desired result?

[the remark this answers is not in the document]
nagilson: Yes it would, thank you kind open source contributor 🙂 The simplest solution is always the best solution.

nagilson:
> Can you change the name of the ENV? It is pretty non-standard. Also, long ENVs are fine.
> 
> `ENABLE_P_RELEASE_SLN` -> `DOTNET_PUBLISH_RELEASE_SLN`

That is also one of Daniel's remaining feedbacks. We considered:
`DOTNET_CLI_ENABLE_PUBLISH_RELEASE_SOLUTIONS`
`DOTNET_CLI_ENABLE_PACK_RELEASE_SOLUTIONS`

or one for both like `DOTNET_CLI_ENABLE_PUBLISH_AND_PACK_RELEASE_SOLUTIONS`.  Either way we will definitely make it more standard 👍

nagilson:
OK to merge once approved ✔️

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,160 @@
...
+        public override ProjectInstance GetSlnProject(string slnPath, string slnProjectConfigPropertytoCheck = "")
+        {
+            // This has a performance overhead so don't do this unless opted in.
+            if (
+                // PublishRelease v
+                (slnProjectConfigPropertytoCheck == PublishCommandParser.customDefaultConfigurationProperty && Environment.GetEnvironmentVariable(PublishCommandParser.enablePublishReleaseSolutionEnvironmentVar) == null)
+                || // PackRelease v
+                (slnProjectConfigPropertytoCheck == PackCommandParser.customDefaultConfigurationProperty && Environment.GetEnvironmentVariable(PackCommandParser.enablePackReleaseSolutionEnvironmentVar) == null)
+                )
+                return null; // The user will be warned if they do not have this set and try this scenario with one of the properties set.
```
dsplaisted, line 65: We're only ever expecting `slnProjectConfigPropertytoCheck` to equal `PackRelease` or `PublishRelease`.  It would be good to write this code so that if it gets any other values, it errors out so that we fix the code sending unexpected values, or to change it so it's not possible to pass an invalid value.

I would suggest checking for the environment variables in the `PublishCommand` and `PackCommand` classes, and then based on that passing a `bool` parameter through (or setting a property on the locator, or something) which controls whether it should look into solution files.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,160 @@
...
+            HashSet<string> configValues = new HashSet<string>();
+            object projectDataLock = new object();
+
+            bool shouldReturnNull = false;
+
+            Parallel.ForEach(sln.Projects.AsEnumerable(), (project, state) =>
+            {
+                const string executableProjectOutputType = "Exe"; // Note that even on Unix when we don't produce exe this is still an exe, same for ASP
+                const string solutionFolderGuid = "{2150E333-8FDC-42A3-9474-1A3956D46DE8}";
+                const string sharedProjectGuid = "{D954291E-2A0B-460D-934E-DC6B0785DB48}";
```
dsplaisted, line 87: These can go out of the parallel ForEach (I don't think it actually makes a difference since they're `const`, but it feels cleaner to me).

src/Cli/dotnet/ProjectLocator.cs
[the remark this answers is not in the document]
dsplaisted: I think we can delete `ProjectHasUserCustomizedConfiguration` entirely then, right, since the check that actually works is higher up?

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -449,6 +450,87 @@ public void It_publishes_on_release_if_PublishRelease_property_set()
             Assert.True(File.Exists(expectedAssetPath));
         }
 
+
+        [Fact]
+        public void It_publishes_on_release_if_PublishRelease_property_set_in_sln()
```
dsplaisted, line 455: We should add tests that do `dotnet publish --configuration Debug` on a project / solution with `PublishRelease` set to true.  In that case it should publish with the Debug configuration.

Also we should test `dotnet publish -p:Configuration=Debug`, that should also use the Debug configuration but I don't think it currently will.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,160 @@
...
+                        return;
+                    }
+
+                    string configuration = projectData.GetPropertyValue(slnProjectConfigPropertytoCheck);
+                    if (!string.IsNullOrEmpty(configuration))
+                    {
+                        lock (projectDataLock)
+                        {
+                            configuredProjects.Add(projectData); // we don't care about race conditions here
+                            configValues.Add(configuration);
```
dsplaisted, line 112: What if one project has PublishRelease set to `true` and another has it set to `True` (casing difference).  Currently it looks like that would be an error.  We should make the hash set case insensitive, or try parsing the values into a `bool` to fix that.

We should also have a test covering the casing difference.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,160 @@
...
+                if (projectData.GetPropertyValue("OutputType") == executableProjectOutputType)
+                {
+                    if (ProjectHasUserCustomizedConfiguration(projectData))
+                    {
+                        shouldReturnNull = true;
+                        state.Stop(); // We don't want to override Configuration if ANY project in a sln uses a custom configuration
+                        return;
+                    }
+
+                    string configuration = projectData.GetPropertyValue(slnProjectConfigPropertytoCheck);
```
dsplaisted, line 106: I may have been confused before.  The "Configuration" would be something like "Debug" or "Release".  It looks like what we are getting here is the *value* of either the `PackRelease` or `PublishRelease` property, which we expect to be true or false.  Perhaps `useReleaseConfiguration` or something would be a better name here.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+            });
+
+            if (configuredProjects.Any() && configValues.Count > 1 && !shouldReturnNull)
+            {
+                throw new GracefulException(CommonLocalizableStrings.TopLevelPublishConfigurationMismatchError);
+            }
+
+            return shouldReturnNull || configuredProjects.Count == 0 ? null : configuredProjects.First();
```
dsplaisted, line 123: Nit: I think this is a bit simpler and clearer:

```suggestion
            return shouldReturnNull ? null : configuredProjects.FirstOrDefault();
```

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        /// <summary>
+        /// Provide a CLI input to change configuration based on
```
dsplaisted, line 127: Nit: I think this is a bit clearer

```suggestion
        /// Returns dotnet CLI command-line parameters (or an empty list) to change configuration based on 
```

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <returns>The top-level project (first if multiple exist) in a SLN. Returns null if no top level project. Throws exception if two top level projects disagree
+        /// in the configuration property to check.</returns>
+        public abstract ProjectInstance GetSlnProject(string potentialSlnPath, string slnProjectConfigPropertytoCheck = "");
+
+        /// <returns>Creates a ProjectInstance if the project is valid, elsewise, fails..</returns>
+        protected static ProjectInstance TryGetProjectInstance(string projectPath)
+        {
+            try
+            {
+                return new ProjectInstance(projectPath);
```
dsplaisted, line 27: We should probably pass through any global properties that are going to be passed to MSBuild.  That will help us correctly handle things like `--property:Configuration=Release`, but also something like `--property:PublishRelease=true` or other weirder things.

This is another scenario to add tests for, too.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -859,4 +859,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1188: Package {0} {1} has a resource with the locale '{2}'. This locale is not recognized by .NET. Consider notifying the package author that it appears to be using an invalid locale.</value>
     <comment>Error code is NETSDK1188. 0 is a package name, 1 is a package version, and 2 is the incorrect locale string</comment>
   </data>
+  <data name="PReleaseRequiresEnvVarOnSln" xml:space="preserve">
+    <value>NETSDK1189: To use PackRelease or PublishRelease in solution projects, you must set the environment variable DOTNET_CLI_ENABLE_X_RELEASE_FOR_SOLUTIONS to true where X is PUBLISH or PACK. This will increase the time to publish a sln.</value>
```
dsplaisted, line 863: We should parameterize this method so it includes the actual names / values:

```suggestion
    <value>NETSDK1189: To use {0} in solution projects, you must set the environment variable {0} to true. This will increase the time to complete the operation.</value>
```

dsplaisted (commented):
Sorry for another flood of feedback, but this is actually looking good. 👍

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Ok, I agree 👍 It was actually outside initially but then the debugger doesn't resolve outside variables in the thread, so I put them in the parallel code

nagilson:
> Sorry for another flood of feedback, but this is actually looking good. 👍

Thanks 🙂  Actually, I love getting feedback like this because it's the best way to learn and improve, so I appreciate the thoroughness.

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Build.Execution;
+using Microsoft.DotNet.Cli.Utils;
+
+namespace Microsoft.DotNet.Cli
+{
+    abstract class ProjectLocator
```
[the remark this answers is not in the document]
nagilson: Good question. The code that `dotnet run` is doing is very similar. So this allows us to reduce some of the duplication if we want to do that in future. Didn't do that here because would like to keep this as 'simple' as possible

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <returns>A project instance that will be targeted to publish/pack, etc. null if one does not exist.</returns>
+        public abstract ProjectInstance GetTargetedProject(IEnumerable<string> slnOrProjectArgs, string slnProjectPropertytoCheck = "");
```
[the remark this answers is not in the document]
nagilson: Made this public for future stuff like if we make `dotnet run` use this pattern. Thoughts?

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <returns>The top-level project (first if multiple exist) in a SLN. Returns null if no top level project. Throws exception if two top level projects disagree
```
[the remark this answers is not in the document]
nagilson: Going to use "executable"

src/Tasks/Common/Resources/Strings.resx
nagilson: Done, though I don't see why `complete the operation` is better because it requires more characters and also is less specific.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -523,6 +523,9 @@
   <data name="ToolSettingsUnsupportedRunner" xml:space="preserve">
     <value>Command '{0}' uses unsupported runner '{1}'."</value>
   </data>
+  <data name="TopLevelPublishConfigurationMismatchError" xml:space="preserve">
+    <value>Multiple top-level projects in the given solution contain conflicting PublishRelease or PackRelease values. To use conflicting configurations, leverage a Directory.build.props file instead of setting flags that don't match.</value>
```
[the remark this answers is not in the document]
nagilson: I don't think so since `Directory.build.props` will apply to all projects. We would still generate the error if the values in the projects themselves still mismatch. I provided a fix.

src/Cli/dotnet/ProjectLocator.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <returns>Creates a ProjectInstance if the project is valid, elsewise, fails..</returns>
+        protected static ProjectInstance TryGetProjectInstance(string projectPath)
+        {
+            try
+            {
+                return new ProjectInstance(projectPath);
+            }
+            catch (Exception e) // Catch failed file access, or invalid project files that cause errors when read into memory,
+            {
+                Reporter.Output.WriteLine(e.Message.Yellow());
```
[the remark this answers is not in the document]
nagilson: MSBuild will error if the project file is invalid. This is more so if `Projectinstance` lib doesn't work for some reason, like for example it doesn't correctly work for `.shproj`. But I made it so we don't even check publishrelease on those. We shouldn't fail anywhere els.e

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Done, I added a class member and added an assert

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: That is tricky, did not think of that. Good catch.

src/Cli/dotnet/ProjectLocator.cs
nagilson: Didn't consider what to do if the project defines `PublishRelease` `true` but its given `PublishRelease` `false` through `property:`. I would rather it respect the CLI over the project personally. But maybe we should just error.

nagilson: Looks like for all other properties CLI takes precedence. So gonna do that.

src/Cli/dotnet/ProjectLocator.cs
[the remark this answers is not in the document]
nagilson: We cant, because this is if someone puts in a project in a solution with <Configuration>, and not if someone does p:Configuration=Blah

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
nagilson: Added tests for both of those situations :)

nagilson:
All of the requested changes are in @dsplaisted though we want this to go into 7.0.100 as well, it seems this is going to pick up other depended upon changes to merge it into there. Ill get your confirmation for main first.

dsplaisted:
> All of the requested changes are in @dsplaisted though we want this to go into 7.0.100 as well, it seems this is going to pick up other depended upon changes to merge it into there. Ill get your confirmation for main first.

I think we should retarget this PR to `release/7.0.1xx`.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,175 @@
...
+            IEnumerable<string> slnOrProjectArgs,
+            Option<string> configOption,
+            IEnumerable<string> userPropertyArgs
+            )
+        {
+            ProjectInstance project = null;
+
+            IEnumerable<string> calledArguments = parseResult.Tokens.Select(x => x.ToString());
+            IEnumerable<string> slnProjectAndCommandArgs = slnOrProjectArgs.Concat(calledArguments);
+            if (!parseResult.HasOption(configOption) && !ArgsContainsProperty(userPropertyArgs, "Configuration") && !ArgsContainsProperty(userPropertyArgs, defaultedConfigurationProperty))
```
joeloff, line 146: Should we consider placing the Configuration and Release values in a const or readonly string in case they are needed elsewhere to avoid any typos in the future?

src/Cli/dotnet/commands/dotnet-pack/PackCommandParser.cs
```diff
@@ -14,6 +14,8 @@ namespace Microsoft.DotNet.Cli
     internal static class PackCommandParser
     {
         public static readonly string DocsLink = "https://aka.ms/dotnet-pack";
+        public static readonly string CustomDefaultConfigurationProperty = "PackRelease";
+        public static string enablePackReleaseSolutionEnvironmentVar = "DOTNET_CLI_ENABLE_PACK_RELEASE_FOR_SOLUTIONS";
```
joeloff, line 18: Are we tracking documentation changes for the new variables?

nagilson: I had a PR for doc changes before these worked in `.csproj` and `sln`s but haven't made one for these changes. I'll do that now and link it here when I'm done, thank ya!

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Thats a good idea. I could move the `PackRelease` and `PublishRelease` const there too. Where would be a good place, shall we make a `\src\Common\PropertyNames.cs?`

src/Cli/dotnet/commands/dotnet-pack/PackCommandParser.cs
nagilson: Added some docs here :) https://github.com/dotnet/docs/pull/30747

nagilson:
@dsplaisted removed the commits that came in after 7.0.100 retargeting.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -523,6 +523,9 @@
   <data name="ToolSettingsUnsupportedRunner" xml:space="preserve">
     <value>Command '{0}' uses unsupported runner '{1}'."</value>
   </data>
+  <data name="SolutionExecutableConfigurationMismatchError" xml:space="preserve">
+    <value>Multiple executable projects in the solution contain conflicting '{0}' values. See '{1}'. Make the values match. Consider using a Directory.build.props file to set all project configurations.</value>
```
dsplaisted, line 527: Since the list of projects is going to have newlines in it, I would move it to the end:
```suggestion
    <value>Multiple executable projects in the solution contain conflicting '{0}' values. Make the values match. Consider using a Directory.build.props file to set all project configurations.  Conflicting projects:
{1}</value>
```

src/Cli/dotnet/ProjectLocator.cs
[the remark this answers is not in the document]
dsplaisted: I would keep it simple for now (ie all one class, with the only public APIs being the ones used outside the class).  When we want to re-use it elsewhere for `dotnet run` or something else, then we can refactor it as needed.

It's easier to have a better design for re-use when you actually have multiple places to use something from.  We can change our APIs whenever we want, so there's no need to try to design for re-use up-front before we need it.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Suggestion applied

nagilson:
@dsplaisted Removed the project locator class 👍

nagilson:
This is ready, but we are working on higher priority items atm.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -523,6 +523,10 @@
   <data name="ToolSettingsUnsupportedRunner" xml:space="preserve">
     <value>Command '{0}' uses unsupported runner '{1}'."</value>
   </data>
+  <data name="SolutionExecutableConfigurationMismatchError" xml:space="preserve">
+    <value>Multiple executable projects in the solution contain conflicting '{0}' values. Make the values match. Consider using a Directory.build.props file to set all project configurations.  Conflicting projects:
```
joeloff, line 527: Maybe `Ensure the values match` instead of `Make`. Also a double space at the end before `Conflicting projects`

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+                    try
+                    {
+                        return TryGetProjectInstance(MsbuildProject.GetProjectFileFromDirectory(arg).FullName, globalProps);
+                    }
+                    catch (GracefulException)
+                    {
+                        // Fall back to looking for a solution if multiple project files are found.
+                        string potentialSln = Directory.GetFiles(arg, "*.sln", SearchOption.TopDirectoryOnly).FirstOrDefault();
+
+                        if (!string.IsNullOrEmpty(potentialSln))
```
joeloff, line 92: nit: missing braces around the statement on line 93

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+            if (!checkSolutions)
+                return null;
+            Debug.Assert(slnProjectConfigPropertytoCheck == MSBuildPropertyNames.PUBLISH_RELEASE || slnProjectConfigPropertytoCheck == MSBuildPropertyNames.PACK_RELEASE, "Only PackRelease or PublishRelease are currently expected");
+
+            SlnFile sln;
+            try
+            {
+                sln = SlnFileFactory.CreateFromFileOrDirectory(slnPath);
+            }
+            catch (GracefulException)
```
joeloff, line 115: Any chance the exception might hide important information when caught and not reported?

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            if (configuredProjects.Any() && configValues.Count > 1 && !shouldReturnNull)
+            {
+                throw new GracefulException(CommonLocalizableStrings.SolutionExecutableConfigurationMismatchError, slnProjectConfigPropertytoCheck, String.Join("\n", (configuredProjects).Select(x => x.FullPath)));
+            }
+
+            return shouldReturnNull ? null : configuredProjects.FirstOrDefault();
+        }
+
+        /// <returns>Creates a ProjectInstance if the project is valid, elsewise, fails..</returns>
```
joeloff, line 167: nit: extra `.` in comment

src/Cli/dotnet/commands/dotnet-pack/PackCommand.cs
```diff
@@ -8,6 +8,8 @@
 using System;
 using System.CommandLine;
 using System.CommandLine.Parsing;
+using Microsoft.DotNet.Tools.Publish;
```
joeloff, line 11: Sort using statements

src/Cli/dotnet/CommonLocalizableStrings.resx
nagilson: Also added a period at the end!

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+            IEnumerable<string> slnOrProjectArgs,
+            Option<string> configOption,
+            IEnumerable<string> userPropertyArgs
+            )
+        {
+            ProjectInstance project = null;
+            var globalProperties = GetGlobalPropertiesFromUserArgs(userPropertyArgs);
+
+            if (parseResult.HasOption(configOption) || globalProperties.ContainsKey(MSBuildPropertyNames.CONFIGURATION))
+                return Enumerable.Empty<string>();
```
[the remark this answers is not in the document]
nagilson: Oh! I didn't not know IEnumerables are like Python generator functions. Cool, thanks!

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: I don't think so, because MSBuild will still be called, and it should be what handles the errors here.

nagilson:
@joeloff Thanks for taking another look, please approve if OK. FYI this is what the err looks like in terminal 
image

joeloff (approved):

PR #26811, opened 2022-07-26
Stop double copying all of the watch and format files.

marcpopMSFT:
I found that Target PublishBuiltInTools was already copying files around so we were doing this weird thing where we would copy to another location and then had code that copied from both the original and old location.

For the below, the top copy come from the PBIT step and you can see in the double write, we're copying from the new and old location into the final location.  This removes those extra copies as I don't think they are needed at all.

DoubleWrite example
Copying file from "C:\repos\sdk\artifacts\bin\DotNetWatchTasks\Debug\netstandard2.0\DotNetWatchTasks.dll" to "C:\repos\sdk\artifacts\bin\redist\Debug\net7.0\DotnetTools\dotnet-watch\7.0.100-dev\tools\net7.0\any\DotNetWatchTasks.dll".

DoubleWrites
    C:\repos\sdk\artifacts\bin\redist\Debug\dotnet\sdk\7.0.100-dev\DotnetTools\dotnet-watch\7.0.100-dev\tools\net7.0\any\DotNetWatchTasks.dll
        C:\repos\sdk\artifacts\bin\redist\Debug\net7.0\DotnetTools\dotnet-watch\7.0.100-dev\tools\net7.0\any\DotNetWatchTasks.dll
        C:\repos\sdk\artifacts\bin\DotNetWatchTasks\Debug\netstandard2.0\DotNetWatchTasks.dll


dsplaisted (approved):
Awesome, great to fix and simplify our build system like this.

PR #26862, opened 2022-07-28
Avoid parallel workload MSI extraction

dsplaisted:
## Description

Fix issues updating advertising manifests where manifest packages haven't been updated to include `extractedManifest` folder.  The admin install extractions were running in parallel, which isn't supported by the MSI APIs.  Also add better logging.

Also fix issue where IncludedWorkloadManifests.txt wasn't being found if SDK version wasn't the exact feature band.

## Customer Impact

When attempting to update to the latest workloads using the dotnet CLI for an MSI-based installation, most updates will fail.  It appears that at most one manifest will be updated each time `dotnet workload update` is run.  Workarounds are to keep running the update command until it completes without displaying any manifest update errors, or to use a [rollback file](https://github.com/dotnet/sdk/blob/main/documentation/general/workloads/workload-rollback.md) to specify the versions of the manifests to install.

Separately, the issue with IncludedWorkloadManifests.txt had no effect on 6.0.400, but would prevent workload fallback from working in 6.0.401.  This would currently break all workloads as none of them have switched from 6.0.300 to 6.0.400 branding.

## Regression

Yes

## Risk

Low.  The fix is to avoid running the MSI operations concurrently, and to switch back to previous logic for how the SDK version was passed to the workload resolver.

## Testing

Manual testing of impacted scenarios.

### DIFF
--- a/src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
+++ b/src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
@@ -22,6 +22,7 @@ using Microsoft.Extensions.EnvironmentAbstractions;
 using Microsoft.NET.Sdk.WorkloadManifestReader;
 using NuGet.Versioning;
 using Command = System.CommandLine.Command;
+using Product = Microsoft.DotNet.Cli.Utils.Product;
 using Strings = Microsoft.DotNet.Workloads.Workload.Install.LocalizableStrings;
 
 namespace Microsoft.DotNet.Workloads.Workload
@@ -70,7 +71,7 @@ namespace Microsoft.DotNet.Workloads.Workload
             _sdkVersion = WorkloadOptionsExtensions.GetValidatedSdkVersion(parseResult.GetValueForOption(InstallingWorkloadCommandParser.VersionOption), version, _dotnetPath, _userProfileDir, _checkIfManifestExist);
             _sdkFeatureBand = new SdkFeatureBand(_sdkVersion);
 
-            _installedFeatureBand = installedFeatureBand == null ? new SdkFeatureBand(DotnetFiles.VersionFileObject.BuildNumber) : new SdkFeatureBand(installedFeatureBand);
+            _installedFeatureBand = installedFeatureBand == null ? new SdkFeatureBand(Product.Version) : new SdkFeatureBand(installedFeatureBand);
 
             _fromRollbackDefinition = parseResult.GetValueForOption(InstallingWorkloadCommandParser.FromRollbackFileOption);
             var configOption = parseResult.GetValueForOption(InstallingWorkloadCommandParser.ConfigOption);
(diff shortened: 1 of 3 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -523,12 +527,21 @@ public async Task ExtractManifestAsync(string nupkgPath, string targetPath)
...
 
 
-                    _ = WindowsInstaller.SetInternalUI(InstallUILevel.None);
-                    var result = WindowsInstaller.InstallProduct(msiPath, $"TARGETDIR={msiExtractionPath} ACTION=ADMIN");
-
-                    if (result != Error.SUCCESS)
+                    lock (_msiAdminInstallLock)
                     {
-                        throw new GracefulException(String.Format(LocalizableStrings.FailedToExtractMsi, msiPath));
+                        string adminInstallLog = Path.Combine(Path.GetTempPath(), $"Microsoft.NET.Workload_{DateTime.Now:yyyyMMdd_HHmmss}_AdminInstall_{Path.GetFileNameWithoutExtension(msiPath)}.log");
```
joeloff, line 532: Can you call into GetMsiLogName? (it's in MsiInstallerBase) You may need to add a new overload, but that will ensure that we have consistent names which helps to identify all the logs associated with a specific command execution.

joeloff (approved):

src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
```diff
@@ -70,15 +71,15 @@ internal abstract class InstallingWorkloadCommand : WorkloadCommandBase
...
+            _installedFeatureBand = installedFeatureBand == null ? new SdkFeatureBand(Product.Version) : new SdkFeatureBand(installedFeatureBand);
 
             _fromRollbackDefinition = parseResult.GetValueForOption(InstallingWorkloadCommandParser.FromRollbackFileOption);
             var configOption = parseResult.GetValueForOption(InstallingWorkloadCommandParser.ConfigOption);
             var sourceOption = parseResult.GetValueForOption(InstallingWorkloadCommandParser.SourceOption);
             _packageSourceLocation = string.IsNullOrEmpty(configOption) && (sourceOption == null || !sourceOption.Any()) ? null :
                 new PackageSourceLocation(string.IsNullOrEmpty(configOption) ? null : new FilePath(configOption), sourceFeedOverrides: sourceOption);
                        
-            var sdkWorkloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(_dotnetPath, _installedFeatureBand.ToString(), userProfileDir);
+            var sdkWorkloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(_dotnetPath, _sdkVersion.ToString(), userProfileDir);
```
marcpopMSFT, line 82: Note that if we want to fix `--print-download-link` with `--sdk-version`, this needs to be `_installedFeatureVersion`. I believe that @gkulin is working on this separately and this change is ok for fixing everything except for --sdk-version usage.

marcpopMSFT:
@dsplaisted branch is open for merges. Ok merging this or want to talk to Grace first about --sdk-version impact first?

PR #26872, opened 2022-07-28
Update minimum msbuild for RC2

marcpopMSFT:
.net 7 will require at least 17.3 to target downlevel runtimes and 17.4 to target net7.0.


v-wuzhai:
What puzzles me is that these failed tests are successful in local testing.

AraHaan:
What puzzles me, is that the ci still fails.

marcpopMSFT:
The CI failing is likely because the helix images don't have 17.3 yet. Let me run again to get a clean run and then ping the engineering system folks.

marcpopMSFT:
@baronfel I think we're ready to merge this if you don't mind reviewing.

baronfel (approved):
All aboard the bump train 🚂

src/Layout/redist/minimumMSBuildVersion
```diff
@@ -1 +1 @@
-17.2.0
+17.3.0
```
AraHaan, line 1: It would be nice if 17.4.0 was minimum msbuild.

marcpopMSFT: Unfortunately, we can't do that as we have a significant number of customers that want to update their SDK but not their Visual Studio version and still target downlevel. There is also sometimes a delay in pipelines on how quickly they can get Visual Studio updated on release date (for 5.0 for example, actions was delayed by ~3 weeks as I recall where anyone trying to use the 6.0 sdk was blocked).

See more details on this model here: https://docs.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules  We'll update to 17.4 with 7.0.200 giving customers one quarter to update their Visual Studio version.

PR #26891, opened 2022-07-29
add rule for detecting attribute differences

### LINKED ISSUE #23905: [API Compat] add rule that ensures there is not attribute differences in between sides

smasher164:
This PR adds a rule to detect attribute changes between two public members.
Merging of this PR is pending inclusion of an exclusion filter.

Fixes #23905


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,263 @@
...
+namespace Microsoft.DotNet.ApiCompatibility.Rules
+{
+    /// <summary>
+    /// This class implements a rule to check that the attributes between public members do not change.
+    /// </summary>
+    public class AttributesMustMatch : IRule
+    {
+        private readonly RuleSettings _settings;
+
+        private class AttributeGroup : IEqualityComparer<AttributeGroup>
```
ViktorHofer, line 26: nit: Either move this and the other private class to the bottom of the parent class or move it up to appear before the instance field `_settings`. Usually I prefer these to be at the bottom of the parent class (we do the same in the BCL for enumerators) but whatever you like.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,263 @@
...
+            {
+                _set = new HashSet<AttributeGroup>(new AttributeGroup(Settings));
+                _settings = Settings;
+                for (int i = 0; i < attributes.Count; i++)
+                {
+                    add(attributes[i]);
+                }
+            }
+
+            public void add(AttributeData attr)
```
ViktorHofer, line 67: Public methods should start with an upper case letter in C#. Same for the other methods below.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,263 @@
...
+    public class AttributesMustMatch : IRule
+    {
+        private readonly RuleSettings _settings;
+
+        private class AttributeGroup : IEqualityComparer<AttributeGroup>
+        {
+            public AttributeData? _repr;
+            public List<AttributeData?>? _attributes;
+            public List<bool>? _seen;
+            private RuleSettings _Settings;
```
ViktorHofer, line 31: This should be readonly and lower cased: `_settings`.
```suggestion
            private readonly RuleSettings _settings;
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,263 @@
...
+    /// </summary>
+    public class AttributesMustMatch : IRule
+    {
+        private readonly RuleSettings _settings;
+
+        private class AttributeGroup : IEqualityComparer<AttributeGroup>
+        {
+            public AttributeData? _repr;
+            public List<AttributeData?>? _attributes;
+            public List<bool>? _seen;
```
ViktorHofer, line 30: You can either initialize these in the ctor (as you currently do) or just directly here. That said, both fields should be readonly.

ViktorHofer: And as they are public, they should be capital case and not start with an underscore. You might want to make these private and add the following:

```
public IReadOnlyList<AttributeData?>? Attributes => _attributes;

public IReadOnlyList<bool>? Seen => _seen;
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,263 @@
...
+                if (!Enumerable.SequenceEqual(left.ConstructorArguments, right.ConstructorArguments))
+                {
+                    return false;
+                }
+                return Enumerable.SequenceEqual(left.NamedArguments, right.NamedArguments);
+            }
+            return left == right;
+        }
+
+        private void reportAttributeDifferences(ISymbol containing,
```
ViktorHofer, line 158: In general, all methods are capital cased in C#, including private ones, i.e. https://github.com/dotnet/runtime/blob/e71a9583b4d6c9bd97edd87cda7f98f232f63530/src/libraries/System.Private.CoreLib/src/System/String.cs#L66.

ViktorHofer:
@smasher164 can you please update the changes in the newly added rule according to our official [C# Coding Conventions](https://github.com/dotnet/sdk/pull/26891)? (method casing, naming conventions, ...)

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,263 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
```
ViktorHofer, line 3:
```suggestion
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+        }
+
+        private void readExclusions(IEnumerable<string> excludeAttributesFiles)
+        {
+            foreach (string filePath in excludeAttributesFiles)
+            {
+                if (!File.Exists(filePath))
+                {
+                    throw new FileNotFoundException($"File {filePath} was not found.", filePath);
+                }
```
ViktorHofer, line 41: This is a non localized exception. I think it's OK to just let the `File.ReadAllLines` call throw the FileNotFoundException if the file doesn't exist.

```suggestion
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+            _settings = settings;
+            if (excludeAttributesFiles != null)
+            {
+                readExclusions(excludeAttributesFiles);
+            }
+            context.RegisterOnMemberSymbolAction(RunOnMemberSymbol);
+            context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
+        }
+
+        private void readExclusions(IEnumerable<string> excludeAttributesFiles)
```
ViktorHofer, line 34: nit: we don't use lowercase names in C#

```suggestion
        private void ReadExclusions(IEnumerable<string> excludeAttributesFiles)
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+    {
+        private readonly RuleSettings _settings;
+        private readonly HashSet<string> _attributesToExclude = new();
+
+        public AttributesMustMatch(RuleSettings settings, IRuleRegistrationContext context, IEnumerable<string>? excludeAttributesFiles)
+        {
+            _settings = settings;
+            if (excludeAttributesFiles != null)
+            {
+                readExclusions(excludeAttributesFiles);
```
ViktorHofer, line 28:
```suggestion
                ReadExclusions(excludeAttributesFiles);
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+            }
+            context.RegisterOnMemberSymbolAction(RunOnMemberSymbol);
+            context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
```
ViktorHofer, line 31: I think we only want to run the rule right now if the exclusion file is provided. Therefore the registration should also be moved up into the if.

```suggestion
                context.RegisterOnMemberSymbolAction(RunOnMemberSymbol);
                context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
            }
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+                return;
+            }
+
+            CompatDifference difference = dt switch
+            {
+                DifferenceType.Changed => CompatDifference.CreateWithDefaultMetadata(
+                    DiagnosticIds.CannotChangeAttribute,
+                    string.Format(Resources.CannotChangeAttribute, attr.AttributeClass, containing),
+                    DifferenceType.Changed,
+                    itemRef + ":[" + attr.AttributeClass?.GetDocumentationCommentId() + "]"),
```
ViktorHofer, line 105: Sorry, I know this was a recent change and I didn't explain it fully. The `CreateWithDefaultMetadata` call is only for tests.

Instead you want to pass the metadata that you receive in the `RunOnMemberSymbol` and `RunOnTypeSymbol` methods down to this methods so that the differences belong to given assemblies.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+                {
+                    Add(attributes[i]);
+                }
+            }
+
+            public void Add(AttributeData attr)
+            {
+                foreach (AttributeGroup group in _set)
+                {
+                    if (_settings.SymbolComparer.Equals(group.Representative.AttributeClass!, attr.AttributeClass!))
```
ViktorHofer, line 327: I might be wrong but I don't think you need the `!` band operator, as the EqualityComparer allows nullable parameters: https://github.com/dotnet/sdk/blob/d9d39eba26dce167c60de8b47316734311e590c0/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs#L13

```suggestion
                    if (_settings.SymbolComparer.Equals(group.Representative.AttributeClass, attr.AttributeClass))
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -0,0 +1,357 @@
...
+
+            public void Add(AttributeData attr)
+            {
+                Attributes.Add(attr);
+                Seen.Add(false);
+            }
+        }
+        private class AttributeSet : IEnumerable<AttributeGroup>
+        {
+            private readonly List<AttributeGroup> _set;
```
ViktorHofer, line 310: Maybe add a comment explaining why this is a List<T> instead of a Hashset<T>.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
smasher164: Why not run if there are no exclusions?
Also, doing this would require passing in an empty file to the ruleFactory in the tests.

ViktorHofer:
@smasher164 are you currently filtering out attributes that aren't visible outside? See https://github.com/dotnet/arcade/blob/3dd26504b456b68c2e3edd18ebf159d0b5486135/src/Microsoft.Cci.Extensions/Filters/PublicOnlyCciFilter.cs#L72-L94.

ViktorHofer:
@ericstj can you please chime in on what I posted above:

> It ignores attributes that contain a typeof(T) parameter in their constructor when T is not visible outside the assembly.

Do you have context on that behavior? I wonder why we would only check visibility for `typeof(T)` parameters but not for others. Also, now that generic attributes exist, I wonder if we would need to consider those as well?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
```diff
@@ -666,12 +666,112 @@ public class First {
...
+";
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory);
+            IEnumerable<CompatDifference> got = differ.GetDifferences(new[] { left }, new[] { right });
+            Assert.Empty(got);
+        }
     }
 
+    internal class TempDirectory : IDisposable
```
ViktorHofer, line 728: Please put that class into its own file and put it into src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/TempDirectory.cs so that its clear that other tests can use it as well when needed.

ViktorHofer: Also is that file coped from somewhere? I think we already have that one in runtime. Did you check if this file already exists in the sdk repo?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
```diff
@@ -666,12 +666,112 @@ public class First {
...
+
+  [Serializable]
+  [Foo(""S"", A = true, B = 3)]
+  public class First {}
+}
+";
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory);
+            IEnumerable<CompatDifference> got = differ.GetDifferences(new[] { left }, new[] { right });
```
ViktorHofer, line 723: Use the single left and single right method overload instead:

```suggestion
            IEnumerable<CompatDifference> got = differ.GetDifferences(left, right);
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/NullableAttributes.cs
```diff
@@ -0,0 +1,91 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
ViktorHofer, line 1: Why do you need this file? Is it just for the `MaybeNullWhenAttribute` case?

smasher164: Yeah. It used to also be for `NotNull`, but once I started using `List`, that went away.
On net472, we don't have access to these attributes, so this file is a fallback.
I can see these attributes being used again in the future, so I think it's worth keeping.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
```diff
@@ -0,0 +1,727 @@
...
+        {
+            using TempDirectory root = new();
+            string filePath = Path.Combine(root.DirPath, "exclusions.txt");
+            File.Create(filePath).Dispose();
+            TestRuleFactory s_ruleFactory = new((settings, context) => new AttributesMustMatch(settings, context, new[] { filePath }));
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory);
+            IEnumerable<CompatDifference> got = differ.GetDifferences(left, right);
+            Assert.Equal(want, got);
```
ViktorHofer, line 677: nit: usually we call these `expected` and `actual` or `result`.

ViktorHofer (approved):
LGTM. Just left one more nit and the remaining patch that I shared on Teams. Note that you currently have merge conflicts and builds and tests didn't run.

ViktorHofer (approved):

PR #26905, opened 2022-08-01
Update branding to 6.0.401

v-wuzhai:


marcpopMSFT:
@v-wuzhai I think Sean has a script to do branding updates in all branches that he was going to run tomorrow that all the 6.0.400 PRs are waiting on. We also need to switch it from rtm to servicing.

PR #26926, opened 2022-08-02
Add baseline package validation strict mode option

### LINKED ISSUE #25302: [API Compat] Provide a strict option for package-validation baseline checks

We are hitting issues using Microsoft.CodeAnalysis.PublicApiAnalyzers to verify **no** public API changes in our servicing branches. This feature would enable us to use `package-validation` instead.

Quoting me:
> 2.	It checks for breaking changes, not meeting our “no public API changes” requirement at all.

Quoting @ericstj:
> 2.	We do have a “strict” mode that means no changes – this is how we make sure all API is exposed in our reference assemblies.  Probably we don’t expose this option for baseline comparisons but we could.  That’s an easy feature to add. 

### Additional context

If it matters, we already enable package validation in dotnet/aspnetcore and we leave `$(DisablePackageBaselineValidation)` alone in servicing branches release/6.0. We'd use package validation in release/5.0 as well with this feature in place.

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/25302

Adds a strict mode option for baseline validation to the apicompat
frontends. This mode is intended to be used during servicing, when in a
patch version update, no new public api additions are allowed.

Adds two tests that verify that the switch works when enabled and
disabled.

cc @dougbu

### DIFF
--- a/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
+++ b/src/Assets/TestProjects/PackageValidationTestProject/PackageValidationTestProject.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <DefineConstants Condition="'$(ForceValidationProblem)' == 'true'">$(DefineConstants);ForceValidationProblem</DefineConstants>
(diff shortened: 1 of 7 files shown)


smasher164 (approved):
LGTM!

Issue #23893, opened 2022-02-11
[Package Validation] investigate performance impact to the developer innerloop when creating a package


safern:
Currently we just added the package validation tool to run after the `Pack` task, however since we made the tool `opt-in` rather than `opt-out` we didn't focus much on perf numbers, meaning that we didn't establish a goal nor investigated how much time this increases a developer innerloop when creating a nuget package. 

We should investigate and establish a goal and try to identify hot paths and improvements that can be made, maybe when loading the assemblies, when running rules, traversing the mapping tree, etc.

safern:
Related: https://github.com/dotnet/sdk/issues/23896

PR #26935, opened 2022-08-02
Invoke PackageValidation only when inputs have changed

### LINKED ISSUE #23517: PackageValidation runs in no-op incremental builds
The PackageValidation SDK feature currently runs always, even if the package hasn't changed: https://github.com/dotnet/sdk/blob/69a8180583569bc64d7c5357ca64d13fc547ba77/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Compatibility.Common.targets#L10-L12

This is the case because the `RunPackageValidation` target is hooked onto the wrong target (maybe there's one that doesn't run in no-op incremental packaging builds) and because it doesn't define any Inputs and Outputs.

When incrementally building the libraries subset of dotnet/runtime for allconfigurations (`build.cmd libs -allconfigurations && build.cmd libs -allconfigurations /bl`), this adds **111 seconds** to the overall build.

image

cc @safern @joperezr @ericstj

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/23517
Contributes to https://github.com/dotnet/sdk/issues/23893

By defining inputs and a semaphore file as the output, PackageValidation doesn't run anymore in an no-op incremental build. In the example of dotnet/runtime, this saves nearly two minutes of unnecessary work.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets
@@ -11,6 +11,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 <Project>
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


smasher164 (approved):
LGTM

PR #26936, opened 2022-08-02
Enable nullability checks in rules and avoid reflection during start-up

ViktorHofer:
Contributes to https://github.com/dotnet/sdk/issues/25760, https://github.com/dotnet/sdk/issues/23893

Enabling nullability checks in all existing rules, in the rule conext and in the
runner. Passing in the `RuleRunnerSettings` and `RuleRunnerContext` objects
via the constructor instead of in an `Initialize` method to avoid nullability
errors. Without the `Initialize` method, the Rule base class can be converted
to an interface.

Also using Jab to retrieve the registered rules at compile time instead
of relying on run time discovery via reflection.


smasher164 (approved):
LGTM

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
```diff
@@ -31,8 +35,9 @@ private void RunOnAssemblySymbol(IAssemblySymbol left, IAssemblySymbol right, st
                 return;
             }
 
-            AssemblyIdentity leftIdentity = left.Identity;
-            AssemblyIdentity rightIdentity = right.Identity;
+            // At this point, left and right are both not null.
```
joperezr, line 38: NIT: Not technically true, you could have an edge case where both are null

ViktorHofer: Interesting. That would result in a NullReferenceException today. Do you have an example of such an edge case?

joperezr:
Not really pushing back on this, but just note that the reason why it was originally done via Reflection (and automatically detecting Rules at runtime) was for extensibility. We wanted to allow customers that have special compat rules to be able to just plug-in to our model and get called just as the rest of our rules automatically. Not sure if we have an issue for that, but I know Santi's plan here was to eventually have an optional task parameter `AdditionalApiCompatRulePaths` or something like that, which would take in a list of assemblies where we would also run reflection on to look for more rules.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
joperezr: AFAIK this shouldn't happen unless there is a bug with one of the mappers, but IMO it is better to have a Debug.Assert with a searchable and diagnosable error message than just letting it null-ref by adding a `!` in order to get rid of the nullability warning.

ViktorHofer:
Thanks for the insights @joperezr. I remembered that you already mentioned that in the initial hand-off meeting. Several weeks ago I spoke with @ericstj about this in our 1:1 and he mentioned that we should just get rid of identifying rules at run time to avoid the use of reflection here.

We don't yet have a customer scenario written down for such a feature and the previous implementation didn't make this possible anyway (as the flag that you mentioned didn't exist). That said, removing the reflection code path doesn't mean that we can't make this work. I imagine if we would want to support this, we could implement this via a convention, i.e. loading the passed in assembly path and instantiating an expected to exist type that returns an `IEnumerable<IRule>` without reflection over all exposed types.

src/Compatibility/Microsoft.DotNet.ApiCompatibility/Rules/AssemblyIdentityMustMatch.cs
ViktorHofer: > than just letting it null-ref by adding a !

FWIW, this change didn't modify the existing behavior. The `!` bang operator just silences the now emitted nullability error. I agree that a Debug Assert would make sense, but IMHO it would make more sense to not have that in every rule but at the place where the mapped values are passed into the rule (RuleRunner). For the rule itself ideally we would be able to express that both parameters can't be null via i.e. the following (which AFAIK doesn't exist in the BCL today):

```c# 
Rule([NotNullWhenParameterIsNull(nameof(right)) IAssemblySymbol? left, [NotNullWhenParameterIsNull(nameof(left)) IAssemblySymbol right, ...)
```

PR #26963, opened 2022-08-03
Fix dotnet test --blame parameters to validate incoming values, and update help

nohwnd:


src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -89,20 +89,24 @@ internal static class TestCommandParser
...
         public static readonly Option<string> BlameCrashDumpOption = new ForwardedOption<string>("--blame-crash-dump-type", LocalizableStrings.CmdBlameCrashDumpTypeDescription)
-            .ForwardAsMany(o => new[] { "-property:VSTestBlameCrash=true", $"-property:VSTestBlameCrashDumpType={o}" });
+        {
+            ArgumentHelpName = LocalizableStrings.CrashDumpTypeArgumentName,
+        }
+        .FromAmong(new string[] { "full", "mini" })
+        .ForwardAsMany(o => new[] { "-property:VSTestBlameCrash=true", $"-property:VSTestBlameCrashDumpType={o}" });
 
-        public static readonly Option<string> BlameCrashAlwaysOption = new ForwardedOption<string>("--blame-crash-collect-always", LocalizableStrings.CmdBlameCrashCollectAlwaysDescription)
+        public static readonly Option<bool> BlameCrashAlwaysOption = new ForwardedOption<bool>("--blame-crash-collect-always", LocalizableStrings.CmdBlameCrashCollectAlwaysDescription)
```
baronfel, line 99: it looks like the intent of this property is to never let the user specify a `false` value. for these kinds of flag-style properties, consider setting `{ Arity = ArgumentArity.Zero }` so that a user passing in an explicit true/false value gets a parse error instead of the boolean value getting silently parsed and then discarded by the forwarding.

baronfel (approved):
Love the see the nicer argument validation!

nohwnd:
@baronfel  I've applied your suggestion to all of our bool parameters, but saw that some common parameters that have the same shape don't use it (--no-restore would be one example).

Maybe it is not a good idea to do this, if we consider that people might be using them for scripting. For example in powershell they would do:

```powershell 
function RunBuild {
  param ([switch] $NoRestore) 

  dotnet build --no-restore:$NoRestore
}

RunBuild -NoRestore -> dotnet build --no-restore:True
RunBuild -> dotnet build --no-restore:False
```

Making `--no-restore` zero arity leads to parsing error, and the value for that option being always True. 

I think think is very likely to happen in CI scripts and so I am going to revert. 

(the [Switch] parameter type above in powershell code is using totally the same idea, it accepts zero or one argument, and implicitly casts from and to bool, to allow easy chaining of parameters from parent to child, or splatting parameters).

PR #26966, opened 2022-08-03
[Blazor] Update SDK to account for the crypto worker file

javiercn:
* Updates the Blazor SDK to account for the JS worker script used for crypto operations.
* The change updates blazor.boot.json to include a new section `runtimeAssets` that contains additional assets that are passed directly to the runtime for consumption.


pavelsavara (approved):

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -157,6 +157,9 @@ Copyright (c) .NET Foundation. All rights reserved.
       <!-- Remove dotnet.js/wasm from runtime pack, in favor of the relinked ones in @(WasmNativeAsset) -->
       <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)"
                                Condition="@(WasmNativeAsset->Count()) > 0 and '%(FileName)' == 'dotnet' and ('%(Extension)' == '.wasm' or '%(Extension)' == '.js')" />
+
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)"
+                               Condition="@(WasmNativeAsset->Count()) > 0 and '%(FileName)' == 'dotnet-crypto-worker' and '%(Extension)' == '.js'" />
```
SteveSandersonMS, line 162: I'm not sure that `@(WasmNativeAsset->Count())` is the right condition, either here or above. Shouldn't it be checking for `WasmBuildNative`? That automatically becomes true if there are any WasmNativeAsset entries, but can also be true even if there are none (e.g., because the developer is just trying to change some config like `EmccInitialHeapSize`, which I have been doing recently)

SteveSandersonMS: Also why are we changing this in the 6.0 targets? Does the new feature impact 6.0?

src/BlazorWasmSdk/Tasks/BootJsonData.cs
```diff
@@ -98,6 +98,14 @@ public class ResourcesData
         /// </summary>
         [DataMember(EmitDefaultValue = false)]
         public Dictionary<string, ResourceHashesByNameDictionary> extensions { get; set; }
+
+        /// <summary>
+        /// Extensions created by users customizing the initialization process. The format of the file(s)
+        /// is up to the user.
```
SteveSandersonMS, line 104: What do you mean by "users"? This isn't public API, is it?

Could we have a more specific message, for example stating that this is used for injecting assets from the runtime?

src/BlazorWasmSdk/Tasks/BootJsonData.cs
```diff
@@ -98,6 +98,14 @@ public class ResourcesData
...
+        /// </summary>
+        [DataMember(EmitDefaultValue = false)]
+        public Dictionary<string,AdditionalAsset> runtimeAssets { get; set; }
```
SteveSandersonMS, line 107:
```suggestion
        public Dictionary<string, AdditionalAsset> RuntimeAssets { get; set; }
```

src/BlazorWasmSdk/Tasks/ComputeBlazorBuildAssets.cs
```diff
@@ -118,6 +118,28 @@ public override bool Execute()
                         assetCandidates.Add(newDotNetJs);
                         continue;
                     }
+                    else if(candidate.GetMetadata("FileName") == "dotnet-crypto-worker" && candidate.GetMetadata("Extension") == ".js")
```
SteveSandersonMS, line 121:
```suggestion
                    else if (candidate.GetMetadata("FileName") == "dotnet-crypto-worker" && candidate.GetMetadata("Extension") == ".js")
```

Also are these definitely case-insensitive?

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -214,6 +216,34 @@ public override bool Execute()
                     continue;
                 }
 
+                if (isDotNetCryptoJs)
+                {
+                    var aotDotNetJs = WasmAotAssets.SingleOrDefault(a => $"{a.GetMetadata("FileName")}{a.GetMetadata("Extension")}" == "dotnet-crypto-worker.js");
```
SteveSandersonMS, line 221: This variable name looks wrong

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -214,6 +216,34 @@ public override bool Execute()
...
+                    ITaskItem newDotNetJs = null;
```
SteveSandersonMS, line 222: Variable name looks wrong

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -252,9 +282,13 @@ public override bool Execute()
             static bool IsDotNetJs(string key)
             {
                 var fileName = Path.GetFileName(key);
-                return fileName.StartsWith("dotnet.", StringComparison.Ordinal) && fileName.EndsWith(".js", StringComparison.Ordinal);
+                return fileName.StartsWith("dotnet.", StringComparison.Ordinal) && fileName.EndsWith(".js", StringComparison.Ordinal) && !fileName.Contains("worker");
+            }
+            static bool IsDotNetCryptoJs(string key)
```
SteveSandersonMS, line 287:
```suggestion

            static bool IsDotNetCryptoJs(string key)
```

src/BlazorWasmSdk/Tasks/ComputeBlazorPublishAssets.cs
```diff
@@ -603,10 +637,10 @@ private void ApplyPublishProperties(ITaskItem newAsemblyAsset)
             }
         }
 
-        private static bool IsNativeAsset(string traitValue) => string.Equals(traitValue, "native", StringComparison.Ordinal);
+        private static bool IsNativeAsset(string traitValue) => string.Equals(traitValue, "native", StringComparison.Ordinal) || string.Equals(traitValue, "js-module-crypto", StringComparison.Ordinal);
 
         private static bool IsRuntimeAsset(string traitValue) => string.Equals(traitValue, "runtime", StringComparison.Ordinal);
-
+
```
SteveSandersonMS, line 643:
```suggestion

```

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -149,6 +152,12 @@ public void WriteBootJson(Stream output, string entryAssemblyName)
                             string.Equals(assetTraitValue, "native", StringComparison.OrdinalIgnoreCase))
                     {
                         Log.LogMessage(MessageImportance.Low, "Candidate '{0}' is defined as a native application resource.", resource.ItemSpec);
+                        if (string.Equals(fileName, "dotnet", StringComparison.OrdinalIgnoreCase) &&
+                            string.Equals(fileExtension, ".wasm", StringComparison.OrdinalIgnoreCase))
```
SteveSandersonMS, line 156: Any reason we're case-insensitive here when most of the logic above isn't?

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -161,6 +170,10 @@ public void WriteBootJson(Stream output, string entryAssemblyName)
                         Debug.Assert(!string.IsNullOrEmpty(targetPath), "Target path for '{0}' must exist.", resource.ItemSpec);
                         AddResourceToList(resource, resourceList, targetPath);
                         continue;
+                    }else if(string.Equals(assetTraitName, "BlazorWebAssemblyResource", StringComparison.OrdinalIgnoreCase) &&
```
SteveSandersonMS, line 173:
```suggestion
                    } else if(string.Equals(assetTraitName, "BlazorWebAssemblyResource", StringComparison.OrdinalIgnoreCase) &&
```

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/StaticWebAssetsBaselines/Publish_DoesNotGenerateManifestJson_IncludesJSModulesOnBlazorBootJsonManifest.Publish.files.json
```diff
@@ -128,9 +128,6 @@
   "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\System.Private.CoreLib.dll",
   "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\System.Private.CoreLib.dll.br",
   "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\System.Private.CoreLib.dll.gz",
-  "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\System.Runtime.InteropServices.JavaScript.dll",
-  "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\System.Runtime.InteropServices.JavaScript.dll.br",
-  "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\System.Runtime.InteropServices.JavaScript.dll.gz",
```
SteveSandersonMS, line 133: Why has this changed? The 6.0 output shouldn't be different now should it?

SteveSandersonMS: Oh never mind, I see it's just the ordering that's changed. I'm sure that's not breaking.

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
javiercn: We do case insensitive comparisons for files to keep things consistent between windows and other OSes

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/StaticWebAssetsBaselines/Publish_DoesNotGenerateManifestJson_IncludesJSModulesOnBlazorBootJsonManifest.Publish.files.json
```diff
@@ -195,6 +195,9 @@
   "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\blazorwasm-minimal.dll.br",
   "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\blazorwasm-minimal.dll.gz",
   "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\blazorwasm-minimal.pdb.gz",
+  "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\dotnet-crypto-worker.${RuntimeVersion}.[[hash]].js",
+  "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\dotnet-crypto-worker.${RuntimeVersion}.[[hash]].js.br",
+  "${ProjectPath}\\bin\\Debug\\${Tfm}\\publish\\wwwroot\\_framework\\dotnet-crypto-worker.${RuntimeVersion}.[[hash]].js.gz",
```
SteveSandersonMS, line 200: So it looks like we are producing new published content for 6.0 apps. Is the intention that people can just ignore it and we know for definite that it will never be used at runtime?

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
SteveSandersonMS: I'm more asking about why the case sensitivity is different from the logic above.

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/StaticWebAssetsBaselines/Publish_DoesNotGenerateManifestJson_IncludesJSModulesOnBlazorBootJsonManifest.Publish.files.json
javiercn: Yeah, this is an artifact of the way we order things in the baseline

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
javiercn: You mean compared to https://github.com/dotnet/sdk/pull/26966/files/a995950af430f86cfc1c55c16239a6e588549ece#diff-18a7293544774a63e6250aa358d64358829902f2f3d8b30dbe352b8c2178b63aL116? I think that is just a small bug from what I have seen in the rest of the file.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: Not really since the 6.0 runtime pack does not contain these files. The only change that .NET 6.0 will see is that we add an additional section to the manifest when we build with .NET 7.0 SDK (because we put the dotnet.wasm file there). But anything that .NET 6.0 reads should be the same.

We had 5.0 and 6.0 targets because we didn't want to change how 5.0 apps built with the .NET 6.0 SDK at the time, but 6.0 means `current`. We can rename the file in a follow up change afterwards if we need to.

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/StaticWebAssetsBaselines/Publish_DoesNotGenerateManifestJson_IncludesJSModulesOnBlazorBootJsonManifest.Publish.files.json
javiercn: We know it won't be used at runtime.

These files are only present in 7.0.

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -161,6 +170,11 @@ public void WriteBootJson(Stream output, string entryAssemblyName)
                         Debug.Assert(!string.IsNullOrEmpty(targetPath), "Target path for '{0}' must exist.", resource.ItemSpec);
                         AddResourceToList(resource, resourceList, targetPath);
                         continue;
+                    } 
+                    else if(string.Equals(assetTraitName, "BlazorWebAssemblyResource", StringComparison.OrdinalIgnoreCase) &&
```
SteveSandersonMS, line 174:
```suggestion
                    else if (string.Equals(assetTraitName, "BlazorWebAssemblyResource", StringComparison.OrdinalIgnoreCase) &&
```

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: > I'm not sure that @(WasmNativeAsset->Count()) is the right condition, either here or above. Shouldn't it be checking for WasmBuildNative? That automatically becomes true if there are any WasmNativeAsset entries, but can also be true even if there are none (e.g., because the developer is just trying to change some config like EmccInitialHeapSize, which I have been doing recently)

The logic here predates me (this is something that Pranav worked out as part of 5.0) but essentially what this is doing is removing the file from the runtime pack if a different file was generated as part of AoT compilation so that the native version is picked up.

SteveSandersonMS: I know, but it's using the wrong condition to know whether a native compilation occurred. Native compilations occur if:

 * `$(WasmBuildNative)` was explicitly set to true
 * Or if `@(WasmNativeAsset)` is nonempty (which sets `$(WasmBuildNative)` to true automatically)
 * Or if `$(RunAotCompilation)` is true (which sets `$(WasmBuildNative)` to true automatically, and might also add entries to `@(WasmNativeAsset)` but I don't know)

So it's not just about AOT, nor just about having native dependencies. It's either of those, **or** the case when a developer explicitly turns on native compilation just because they want to tweak the Emscripten args.

As far as I can tell, the logic that predates you is incorrect, and probably behaves weirdly when developers just set `$(WasmBuildNative)` to true.

I appreciate you won't want to interfere with build logic unrelated to your PR here, so I totally accept if you want to just inherit the existing probably-buggy behavior. If this does turn out to lead to actual problems for customers, we can give the workaround of "just add an empty file as a native dependency". Longer term of course, we don't want to have any of this native build logic in the BlazorWebAssembly targets as it's all stuff that should be baked into the runtime.

javiercn: @SteveSandersonMS I think I see what you mean and it definitely seems that the condition here is wrong. I'm trying to think what the best way to solve this is, since the goal of this is something like:
* If there is a WasmNativeAsset file, prefer that.

src/BlazorWasmSdk/Tasks/BootJsonData.cs
```diff
@@ -103,7 +103,7 @@ public class ResourcesData
         /// Additional assets that the runtime consumes as part of the boot process.
         /// </summary>
         [DataMember(EmitDefaultValue = false)]
-        public Dictionary<string, AdditionalAsset> RuntimeAssets { get; set; }
+        public Dictionary<string, AdditionalAsset> runtimeAssets { get; set; }
```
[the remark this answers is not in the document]
SteveSandersonMS: Oops no, now I realise this weird casing is deliberate. Marking as resolved.

PR #26972, opened 2022-08-03
use $HOME in zsh tool path instead of ~

baronfel:
Fixes #22214 by using $HOME in the generated `/etc/path.d` shim for ZSH.  I need to add more testing, and I'd like to make sure that existing bad PATHs are detected - if the file exists and has bad content we should feel free to overwrite it to correct the mistake.

Some of the issues I'll link here suggest that path_helper doesn't expand variables, so testing for that will be required as well.


tullo-x86:
This doesn't work because `path_helper` escapes the `$` character.

image

baronfel:
You're exactly right, I meant to close this PR a couple days ago after I discovered that

PR #26989, opened 2022-08-04
Enable nullable for the mapper types and code clean-up

### LINKED ISSUE #25760: [ApiCompat] Enable nullable reference types

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/25760

I left out the mapper types in my previous changes as I haven't yet looked into them more closely. I just spent several hours drawing a call stack and trying to understand the mapping infrastructure and I fell much more enlightened now 😁.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/ElementContainer.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/ElementContainer.cs
@@ -9,17 +9,17 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// Class to wrap an Element of T with it's <see cref="MetadataInformation"/>.
     /// </summary>
     /// <typeparam name="T">The type of the Element that is holded</typeparam>
-    public class ElementContainer<T> where T : ISymbol
+    public readonly struct ElementContainer<T> where T : ISymbol
     {
         /// <summary>
         /// The element that the container is holding.
         /// </summary>
-        public T Element { get; private set; }
+        public T Element { get; }
 
         /// <summary>
         /// The metadata associated to the element.
         /// </summary>
-        public MetadataInformation MetadataInformation { get; private set; }
+        public MetadataInformation MetadataInformation { get; }
 
         /// <summary>
         /// Instantiates a new object with the <paramref name="element"/> and <paramref name="metadataInformation"/> used.
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#nullable disable
-
-using Microsoft.CodeAnalysis;
 using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
 
 namespace Microsoft.DotNet.ApiCompatibility.Abstractions
 {
@@ -15,7 +13,13 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// </summary>
     public class AssemblyMapper : ElementMapper<IAssemblySymbol>
     {
-        private Dictionary<INamespaceSymbol, NamespaceMapper> _namespaces;
+        private Dictionary<INamespaceSymbol, NamespaceMapper>? _namespaces;
+        private readonly List<CompatDifference>[] _assemblyLoadErrors;
+
+        /// <summary>
+        /// Gets the assembly load errors that happened when trying to follow type forwards.
+        /// </summary>
+        public IReadOnlyList<IReadOnlyList<CompatDifference>> AssemblyLoadErrors => _assemblyLoadErrors;
 
         /// <summary>
         /// Instantiates an object with the provided <see cref="ComparingSettings"/>.
@@ -23,7 +27,12 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
         /// <param name="settings">The settings used to diff the elements in the mapper.</param>
         /// <param name="rightSetSize">The number of elements in the right set to compare.</param>
         public AssemblyMapper(ComparingSettings settings, int rightSetSize = 1)
-            : base(settings, rightSetSize) { }
+            : base(settings, rightSetSize)
+        {
+            _assemblyLoadErrors = new List<CompatDifference>[rightSetSize];
+            for (int i = 0; i < rightSetSize; i++)
+                _assemblyLoadErrors[i] = new List<CompatDifference>();
+        }
 
         /// <summary>
         /// Gets the mappers for the namespaces contained in <see cref="ElementMapper{T}.Left"/> and <see cref="ElementMapper{T}.Right"/>
@@ -36,20 +45,14 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
                 _namespaces = new Dictionary<INamespaceSymbol, NamespaceMapper>(Settings.EqualityComparer);
                 AddOrCreateMappers(Left, ElementSide.Left);
 
-                if (Right.Length == 1)
+                for (int i = 0; i < Right.Length; i++)
                 {
-                    AddOrCreateMappers(Right[0], ElementSide.Right);
-                }
-                else
-                {
-                    for (int i = 0; i < Right.Length; i++)
-                    {
-                        AddOrCreateMappers(Right[i], ElementSide.Right, i);
-                    }
+                    AddOrCreateMappers(Right[i], ElementSide.Right, i);
                 }
 
-                void AddOrCreateMappers(IAssemblySymbol symbol, ElementSide side, int setIndex = 0)
+                void AddOrCreateMappers(IAssemblySymbol? symbol, ElementSide side, int setIndex = 0)
                 {
+                    // Silently return if the element hasn't been added yet.
                     if (symbol == null)
                     {
                         return;
@@ -62,7 +65,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
                     while (stack.Count > 0)
                     {
                         INamespaceSymbol nsSymbol = stack.Pop();
-                        bool hasTypeForwards = typeForwards.TryGetValue(nsSymbol, out List<INamedTypeSymbol> types);
+                        bool hasTypeForwards = typeForwards.TryGetValue(nsSymbol, out List<INamedTypeSymbol>? types);
                         if (hasTypeForwards || nsSymbol.GetTypeMembers().Length > 0)
                         {
                             NamespaceMapper mapper = AddMapper(nsSymbol);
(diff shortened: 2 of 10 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/ElementContainer.cs
```diff
@@ -9,17 +9,17 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// Class to wrap an Element of T with it's <see cref="MetadataInformation"/>.
     /// </summary>
     /// <typeparam name="T">The type of the Element that is holded</typeparam>
-    public class ElementContainer<T> where T : ISymbol
+    public readonly struct ElementContainer<T> where T : ISymbol
```
ViktorHofer, line 12: Changes in this file are unrelated but are an optimization (using a `readonly struct` over `class` reduces run time costs) that I felt isn't worth putting into a separate PR.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
```diff
@@ -15,15 +13,26 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
...
     public class AssemblyMapper : ElementMapper<IAssemblySymbol>
     {
-        private Dictionary<INamespaceSymbol, NamespaceMapper> _namespaces;
+        private Dictionary<INamespaceSymbol, NamespaceMapper>? _namespaces;
+        private readonly List<CompatDifference>[] _assemblyLoadErrors;
+
+        /// <summary>
+        /// Gets the assembly load errors that happened when trying to follow type forwards.
+        /// </summary>
+        public IReadOnlyList<IReadOnlyList<CompatDifference>> AssemblyLoadErrors => _assemblyLoadErrors;
```
ViktorHofer, line 22: Moving the AssemblyLoadErrors from the the base ElementMapper<T> type into the AssemblyMapper as that's the only mapper that deals with resolving external assemblies (via type forwards). This reduces the instantiation costs of the other mapper types.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
```diff
@@ -36,20 +45,14 @@ public IEnumerable<NamespaceMapper> GetNamespaces()
...
-                    AddOrCreateMappers(Right[0], ElementSide.Right);
-                }
-                else
+                for (int i = 0; i < Right.Length; i++)
                 {
-                    for (int i = 0; i < Right.Length; i++)
-                    {
-                        AddOrCreateMappers(Right[i], ElementSide.Right, i);
-                    }
+                    AddOrCreateMappers(Right[i], ElementSide.Right, i);
```
ViktorHofer, line 50: Same clean-up as in https://github.com/dotnet/sdk/commit/73045163de3dab3eb1affb7e18f738d45ddd4e04. The if/else isn't required here.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/MemberMapper.cs
```diff
@@ -12,6 +10,11 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// </summary>
     public class MemberMapper : ElementMapper<ISymbol>
     {
+        /// <summary>
+        /// The containg type of this member.
+        /// </summary>
+        internal TypeMapper ContainingType { get; }
```
ViktorHofer, line 16: style nit: placing the property before the ctor.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -47,53 +46,42 @@ public class ApiComparer : IApiComparer
...
-            mapper.AddElement(left.Element, ElementSide.Left);
 
+            // Retrieve the right names
             string[] rightNames = new string[rightCount];
             for (int i = 0; i < rightCount; i++)
             {
-                ElementContainer<IAssemblySymbol> element = right[i];
-                rightNames[i] = element.MetadataInformation.DisplayString;
-                mapper.AddElement(element.Element, ElementSide.Right, i);
+                rightNames[i] = right[i].MetadataInformation.DisplayString;
```
ViktorHofer, line 68: Constructing the rightNames before initializing the AssemblyMapper to avoid instantiating a throw-away ComparingSettings object.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -47,53 +46,42 @@ public class ApiComparer : IApiComparer
...
-            }
-
             AssemblyMapper mapper = new(GetComparingSettingsCore(leftName, rightName != null ? new[] { rightName } : null));
             mapper.AddElement(left, ElementSide.Left);
             mapper.AddElement(right, ElementSide.Right);
 
             DifferenceVisitor visitor = new();
             visitor.Visit(mapper);
-            return visitor.DiagnosticCollections.First();
+            return visitor.DiagnosticCollections[0];
```
ViktorHofer, line 55: Avoiding Linq as the DiagnosticCollections are now exposed as a IReadOnlyList<T> which allows to index on it.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/TypeMapper.cs
```diff
@@ -18,31 +15,35 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
...
     public class TypeMapper : ElementMapper<ITypeSymbol>
     {
-        private readonly TypeMapper _containingType;
-        private Dictionary<ITypeSymbol, TypeMapper> _nestedTypes;
-        private Dictionary<ISymbol, MemberMapper> _members;
+        private Dictionary<ITypeSymbol, TypeMapper>? _nestedTypes;
+        private Dictionary<ISymbol, MemberMapper>? _members;
+
+        /// <summary>
+        /// The containg type of this type. Null if the type isn't nested.
```
smasher164, line 22: s/containg/containing/

joperezr:
After this, is there anything left to annotate or is this it?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -38,7 +37,7 @@ public class ApiComparer : IApiComparer
...
```
joperezr, line 40: I know this is not new from your work, but do we have to check if this is null? or if the size of the DiagnosticCollection has at least 1 elemnt?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
joperezr: That seems fine by me. One small comment is that I'm not sure I understand why is that relevant in the nullability changes.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/RuleRunner.cs
```diff
@@ -50,14 +51,20 @@ public IReadOnlyList<IEnumerable<CompatDifference>> Run<T>(ElementMapper<T> mapp
...
+                        // ContainingType Left and Right cannot be null, as otherwise, the above condition would be false.
+                        Debug.Assert(mm.ContainingType.Left != null);
+                        Debug.Assert(mm.ContainingType.Right[rightIndex] != null);
+
                         _context.RunOnMemberSymbolActions(
                             mm.Left,
                             mm.Right[rightIndex],
-                            mm.ContainingType.Left,
-                            mm.ContainingType.Right[rightIndex],
+                            mm.ContainingType.Left!,
```
joperezr, line 62: The reason why you still have to add a `!` here is because even though you have a `Debug.Assert` above, you are not saving that into a local field, so Roslyn can't guarantee that from the time you run `Debug.Assert` to when you use it the value has not changed. If you want to get rid of this (and also do a minor perf optimization) you can save the left and right above into a local, Debug.Assert on the locals, and then pass that to `RunOnMemberSymbolActions`

ViktorHofer:
> After this, is there anything left to annotate or is this it?

The underlying issue will get closed when this gets merged. All the shipping code under src/ApiCompat now use the nullable reference type compiler feature (except for the underlying msbuild task infrastructure that comes via the sdk). Also, the feature switch is already enabled for all product code libraries:
- https://github.com/dotnet/sdk/blob/1c6bef9ef27528bb2dbf48dd890c8668641182a1/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj#L6
- https://github.com/dotnet/sdk/blob/1c6bef9ef27528bb2dbf48dd890c8668641182a1/src/ApiCompat/Microsoft.DotNet.PackageValidation/Microsoft.DotNet.PackageValidation.csproj#L6
- https://github.com/dotnet/sdk/blob/1c6bef9ef27528bb2dbf48dd890c8668641182a1/src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/Microsoft.DotNet.ApiCompat.Task.csproj#L11
- https://github.com/dotnet/sdk/blob/1c6bef9ef27528bb2dbf48dd890c8668641182a1/src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Microsoft.DotNet.ApiCompat.Tool.csproj#L6

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
ViktorHofer: The size of the `DiagnosticCollection` is governed by the `rightCount` ctor parameterof the `DifferenceVisitor` type and that value defaults to 1: https://github.com/dotnet/sdk/blob/1c6bef9ef27528bb2dbf48dd890c8668641182a1/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/DifferenceVisitor.cs#L28.

Also, the individual values in the collection are already initialized to an empty `HashSet<CompatDifference>`, so this could never be null.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
ViktorHofer: Yes, this isn't related to the nullable reference type feature. This is general code clean-up (same as for the other comments that I left on this PR) that I bundled into this PR as I already touched the files.

smasher164 (approved):

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/RuleRunner.cs
ViktorHofer: Apparently this only works for debug builds as on release, the call sites are removed.

PR #27011, opened 2022-08-05
Fix print download link bug

gkulin:
addresses bug from #26369 where print download link would check for updates on same feature band twice, rather than falling back


marcpopMSFT:
I ported the change to 6.0.4xx with some minor tweaks and it flowed back with this PR so closing as this PR is no longer needed: https://github.com/dotnet/sdk/pull/27164

PR #27020, opened 2022-08-08
CFS tool configuration for templating testfile

JanKrivanek:
## Problem:
* CFS Scan tool issues due to templating configuration file

## Solution:
* Fix the invalid xml commenting style and missing `config` node in test nuget.config
* Add cfs configuration file


vlada-shubina (approved):
It's very unfortunate that we need to configure the exception repo wide.
@dotnet/dotnet-cli fyi.

JanKrivanek:
> It's very unfortunate that we need to configure the exception repo wide. @dotnet/dotnet-cli fyi.

100% agreed.
I'll keep an eye on progress of CFS Scan tool introducing a local exceptions support

PR #27024, opened 2022-08-08
Support left and right names for assembly set

ViktorHofer:
The existing logic allows to specify a left's name and multiple rights'
names. When comparing an assembly set against one or multiple other
assembly sets, the left and right names are different (for each
assembly).

To support multi assembly mode (assembly set comparison), remove the
hardcoded left and right names values in the runner and instead infer
the name from the mapped (containing) assembly. To faciltate that,
create a hierarchy in the element mapper types.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/ElementContainer.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/ElementContainer.cs
@@ -9,17 +9,17 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// Class to wrap an Element of T with it's <see cref="MetadataInformation"/>.
     /// </summary>
     /// <typeparam name="T">The type of the Element that is holded</typeparam>
-    public readonly struct ElementContainer<T> where T : ISymbol
+    public class ElementContainer<T> where T : ISymbol
     {
         /// <summary>
         /// The element that the container is holding.
         /// </summary>
-        public T Element { get; }
+        public readonly T Element;
 
         /// <summary>
         /// The metadata associated to the element.
         /// </summary>
-        public MetadataInformation MetadataInformation { get; }
+        public readonly MetadataInformation MetadataInformation;
 
         /// <summary>
         /// Instantiates a new object with the <paramref name="element"/> and <paramref name="metadataInformation"/> used.
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
@@ -7,14 +7,19 @@ using Microsoft.CodeAnalysis;
 namespace Microsoft.DotNet.ApiCompatibility.Abstractions
 {
     /// <summary>
-    /// Object that represents a mapping between two <see cref="IAssemblySymbol"/> objects.
+    /// Object that represents a mapping between multiple <see cref="IAssemblySymbol"/> objects.
     /// This also holds a list of <see cref="NamespaceMapper"/> to represent the mapping of namespaces in between
     /// <see cref="ElementMapper{T}.Left"/> and <see cref="ElementMapper{T}.Right"/>.
     /// </summary>
-    public class AssemblyMapper : ElementMapper<IAssemblySymbol>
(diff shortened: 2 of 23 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/ElementContainer.cs
```diff
@@ -9,17 +9,17 @@ namespace Microsoft.DotNet.ApiCompatibility.Abstractions
     /// Class to wrap an Element of T with it's <see cref="MetadataInformation"/>.
     /// </summary>
     /// <typeparam name="T">The type of the Element that is holded</typeparam>
-    public readonly struct ElementContainer<T> where T : ISymbol
+    public class ElementContainer<T> where T : ISymbol
```
ViktorHofer, line 12: Moving to a struct as ElementContainer needs to be nullable. In the case of a struct, it would be wrapped into a Nullable<T> which isn't what we generally want. Also this shouldn't matter in any way as the numer of ElementContainers is limited (its equal to the number of assemblies passed in).

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/AssemblyMapper.cs
```diff
@@ -7,15 +7,20 @@
 namespace Microsoft.DotNet.ApiCompatibility.Abstractions
 {
     /// <summary>
-    /// Object that represents a mapping between two <see cref="IAssemblySymbol"/> objects.
+    /// Object that represents a mapping between multiple <see cref="IAssemblySymbol"/> objects.
     /// This also holds a list of <see cref="NamespaceMapper"/> to represent the mapping of namespaces in between
     /// <see cref="ElementMapper{T}.Left"/> and <see cref="ElementMapper{T}.Right"/>.
     /// </summary>
-    public class AssemblyMapper : ElementMapper<IAssemblySymbol>
+    public class AssemblyMapper : ElementMapper<ElementContainer<IAssemblySymbol>>
```
ViktorHofer, line 14: This means that the AssemblyMapper now isn't just a collection of `IAssemblySymbol` but a tuple of `IAssemblySymbol` and `MetadataInformation`. Ideally this would be flattened to just be `ElementMapper<IAssemblySymbol, MetadataInformation` but that's not easily possible with the current inheritance structure of the other mappers. So nesting is the current best trade-off.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/MemberMapper.cs
```diff
@@ -13,7 +13,7 @@ public class MemberMapper : ElementMapper<ISymbol>
         /// <summary>
         /// The containg type of this member.
         /// </summary>
-        internal TypeMapper ContainingType { get; }
+        public TypeMapper ContainingType { get; }
```
ViktorHofer, line 16: Making this public so that external consumers can also reason about the mapping inheritance structure. No need to only expose this internally given that everything else is public.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -23,15 +24,38 @@ public class ApiComparer : IApiComparer
...
+            get => _comparingSettings ?? new ComparingSettings(includeInternalSymbols: IncludeInternalSymbols,
+                strictMode: StrictMode,
+                warnOnMissingReferences: WarnOnMissingReferences);
+            set => _comparingSettings = value;
+        }
+
+        public ApiComparer(bool includeInternalSymbols = false,
+            bool strictMode = false,
+            bool warnOnMissingReferences = false,
+            ComparingSettings? comparingSettings = null)
```
ViktorHofer, line 38: Adding this constructor as it wasn't yet there and makes it less code to instantiate and configure an `ApiComparer`.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/IApiComparer.cs
```diff
@@ -29,35 +26,50 @@ public interface IApiComparer
         bool StrictMode { get; set; }
 
         /// <summary>
-        /// Flag indicating whether the ApiComparison will warn if there are missing references
+        /// Flag indicating whether the api comparison should warn when references are missing.
```
smasher164, line 29: Capitalize Api

smasher164 (approved):

PR #27030, opened 2022-08-08
Fix the loc build and the CI build

marcpopMSFT:
CI build is currently broken so I'll merge this on red (tested with an internal build).


marcpopMSFT:
@cristianosuzuki77 I think I figured out why the PRs were going to 6.0.4xx. I fixed when the loc build ran but not which branch it listed as the MirrorBranch. I'll let you know if the next PR we get doesn't target main.

PR #27087, opened 2022-08-10
Adding support for VS Designtime targets for worker projects

vijayrkn:
- Adding support for VS Design time targets for worker SDK
- Adding DynamicFileNestingEnabled capability for Worker projects

### DIFF
--- a/src/WebSdk/Worker/Targets/Microsoft.NET.Sdk.Worker.targets
+++ b/src/WebSdk/Worker/Targets/Microsoft.NET.Sdk.Worker.targets
@@ -14,15 +14,24 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup>
     <EnableDefaultContentItems Condition=" '$(EnableDefaultContentItems)' == '' ">true</EnableDefaultContentItems>
     <RunWorkingDirectory Condition=" '$(RunWorkingDirectory)' == '' and '$(EnableDefaultRunWorkingDirectory)' != 'false' ">$(MSBuildProjectDirectory)</RunWorkingDirectory>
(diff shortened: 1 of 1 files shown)


src/WebSdk/Worker/Targets/Microsoft.NET.Sdk.Worker.targets
```diff
@@ -14,15 +14,24 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-    <ProjectCapability Include="DynamicDependentFile" />
-    <ProjectCapability Include="DynamicFileNesting" />
-    <ProjectCapability Include="LocalUserSecrets" />
-    <ProjectCapability Include="WebNestingDefaults" />
-  </ItemGroup>
+  <Import Project="$(MSBuildWorkerTargetsPath)Microsoft.Worker.Designtime.targets" 
+         Condition="Exists('$(MSBuildWorkerTargetsPath)Microsoft.Worker.Designtime.targets')" />
+
+  <Choose>
+    <When Condition="!Exists('$(MSBuildWorkerTargetsPath)Microsoft.Worker.Designtime.targets')" >
```
BillHiebert, line 24: I don't think we should condition these on the targets file. Other IDE's could rely on these as well

BillHiebert (changes requested):

src/WebSdk/Worker/Targets/Microsoft.NET.Sdk.Worker.targets
vijayrkn: I was thinking this might give us a way to remove capabilities from VS if we need but your point is valid. I will remove the condition.

BillHiebert (approved):

PR #27093, opened 2022-08-11
Move dev templates to artifacts temp folder

GangWang01:
Move dev templates to artifacts temp folder so that not affect shipping official packages.


JanKrivanek (approved):

GangWang01:
Drop it since #27091 has similar fix.

Issue #27032, opened 2022-08-09
'dotnet new list' crashes for x64 SDK installed on Windows ARM64


AntonLapounov:
Install the x64 SDK (tested on 7.0.100-rc.1.22408.3) on a Windows 11 ARM64 machine and run `dotnet new list`:
```
C:\>"C:\Program Files\dotnet\x64\dotnet.exe" new list
Unhandled exception: System.BadImageFormatException: Retrieving the COM class factory for component with CLSID {C01B9BA0-BEA7-41BA-B604-D0A36F469133} failed due to the following error: 800700c1  is not a valid Win32 application. (0x800700C1).
   at Microsoft.DotNet.Installer.Windows.InstallerBase..ctor(InstallElevationContextBase elevationContext, ISetupLogger logger, Boolean verifySignatures)
   at Microsoft.DotNet.Installer.Windows.MsiInstallerBase..ctor(InstallElevationContextBase elevationContext, ISetupLogger logger, Boolean verifySignatures, IReporter reporter)
   at Microsoft.DotNet.Workloads.Workload.Install.NetSdkMsiInstallerClient..ctor(InstallElevationContextBase elevationContext, ISetupLogger logger, Boolean verifySignatures, IWorkloadResolver workloadResolver, SdkFeatureBand sdkFeatureBand, INuGetPackageDownloader nugetPackageDownloader, VerbosityOptions verbosity, PackageSourceLocation packageSourceLocation, IReporter reporter)
   at Microsoft.DotNet.Workloads.Workload.Install.NetSdkMsiInstallerClient.Create(Boolean verifySignatures, SdkFeatureBand sdkFeatureBand, IWorkloadResolver workloadResolver, INuGetPackageDownloader nugetPackageDownloader, VerbosityOptions verbosity, PackageSourceLocation packageSourceLocation, IReporter reporter, String tempDirPath, RestoreActionConfig restoreActionConfig)
   at Microsoft.DotNet.Workloads.Workload.Install.WorkloadInstallerFactory.GetWorkloadInstaller(IReporter reporter, SdkFeatureBand sdkFeatureBand, IWorkloadResolver workloadResolver, VerbosityOptions verbosity, String userProfileDir, Boolean verifySignatures, INuGetPackageDownloader nugetPackageDownloader, String dotnetDir, String tempDirPath, PackageSourceLocation packageSourceLocation, RestoreActionConfig restoreActionConfig, Boolean elevationRequired)
[...]

PR #27095, opened 2022-08-11
Fix registry redirection

joeloff:
Previous fix for #27032 did not account for redirection between the 32/64 bit hives.


AntonLapounov (approved):
Thanks!

PR #27159, opened 2022-08-14
Corresponding PublishAOT changes to match the runtime cleanup changes

LakshanF:
Supporting `PublishAOT` scenarios as described in https://github.com/dotnet/runtime/issues/72415 specifically, in the comments [here](https://github.com/dotnet/runtime/issues/72415#issuecomment-1205540035).

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -99,10 +99,10 @@ namespace Microsoft.NET.Build.Tasks
         public ITaskItem[] Crossgen2Packs { get; set; }
 
         [Output]
-        public ITaskItem[] ILCompilerPacks { get; set; }
+        public ITaskItem[] HostILCompilerPacks { get; set; }
 
         [Output]
-        public ITaskItem[] ILCompilerPacks2 { get; set; }
+        public ITaskItem[] TargetILCompilerPacks { get; set; }
 
         //  Runtime packs which aren't available for the specified RuntimeIdentifier
         [Output]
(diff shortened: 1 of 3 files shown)


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -274,6 +309,103 @@ public void NativeAot_hw_runs_with_PackageReference_PublishAot_is_enabled(string
...
+
+                // This will add a reference to a package that will also be automatically imported by the SDK
+                testProject.PackageReferences.Add(new TestPackageReference("Microsoft.DotNet.ILCompiler", "7.0.0-*"));
+
+                // Linux symbol files are embedded and require additional steps to be stripped to a separate file
+                // assumes /bin (or /usr/bin) are in the PATH
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    testProject.AdditionalProperties["StripSymbols"] = "true";
+                    testProject.AdditionalProperties["ObjCopyName"] = "objcopy";
```
MichalStrehovsky, line 331: I got rid of this workaround in #26687. Why do we need it?

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -274,6 +309,103 @@ public void NativeAot_hw_runs_with_PackageReference_PublishAot_is_enabled(string
...
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                var projectName = "HellowWorldNativeAotApp";
+                var rid = EnvironmentInfo.GetCompatibleRid(targetFramework);
+
+                var testProject = CreateHelloWorldTestProject(targetFramework, projectName, true);
+
+                // This will add a reference to a package that will also be automatically imported by the SDK
+                testProject.PackageReferences.Add(new TestPackageReference("Microsoft.DotNet.ILCompiler", "7.0.0-*"));
```
MichalStrehovsky, line 324: Did you have a chance to think about how we could remove this source of non-determinism (https://github.com/dotnet/sdk/pull/27098#issuecomment-1212598845)?

I see two options:
* Make it so that we know the version here. Pass it from the build somehow.
* Make it so that we know the version at test build time. AFAIK this gets spilled into a csproj, so the version string could also be `$(TheILCompilerPackageVersion)` and it will expand then at the time the test project is built. We just need something in the SDK to tell us what ILCompiler package it's including.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -274,6 +309,103 @@ public void NativeAot_hw_runs_with_PackageReference_PublishAot_is_enabled(string
...
+            {
+                var projectName = "HellowWorldNativeAotApp";
+                var rid = "win-arm64";
+
+                var testProject = CreateHelloWorldTestProject(targetFramework, projectName, true);
+                testProject.AdditionalProperties["PublishAot"] = "true";
+
+                // This will add a reference to a package that will also be automatically imported by the SDK
+                testProject.PackageReferences.Add(new TestPackageReference("Microsoft.DotNet.ILCompiler", "7.0.0-*"));
+                testProject.PackageReferences.Add(new TestPackageReference("runtime.win-x64.Microsoft.DotNet.ILCompiler", "7.0.0-*"));
```
MichalStrehovsky, line 374: I would expect crossbuild with PublishAot not to require an ILCompiler PackageReference.

People won't know what package version to include. The SDK should include all the necessary packages for them. The experience of managing ILCompiler packages version is subpar - we only want it for hotfixes, not as a way for people to consume PublishAot.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
MichalStrehovsky: (Maybe this doesn't have to be addressed here, just asking to make sure you saw it)

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -99,7 +99,10 @@ public class ProcessFrameworkReferences : TaskBase
         public ITaskItem[] Crossgen2Packs { get; set; }
 
         [Output]
-        public ITaskItem[] ILCompilerPacks { get; set; }        
+        public ITaskItem[] ILCompilerPacks { get; set; }
+
+        [Output]
+        public ITaskItem[] ILCompilerPacks2 { get; set; }
```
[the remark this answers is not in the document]
LakshanF: I wanted to avoid the compat issues with the runtime dependency (`ResolvedILCompilerPack`) but changing these make sense

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
LakshanF: Not sure why I had a stale copy on this

sbomer (approved):

Issue #22538, opened 2021-11-10
Race condition in ResolvePackageAssets leading to InvalidDataException


dfederm:
### Describe the bug
`ResolvePackageAssets` throws a `InvalidDataException` in certain race conditions.

### To Reproduce
Assume there are 3 projects: `src\A\A.csproj` (A), `src\B\B.csproj` (B), and `src\C\C.csproj` (C).

A and B both depend on C, but via different directory symlinks. A references C via some `ASandbox\C\C.csproj` (`ASandbox\` is a directory symlink to `src\`) and similarly B references C via `BSandbox\C\C.csproj` (`BSandbox\` is a directory symlink to `src\`)

Intermittently we'll see the following when building both A and B at the same time:
```
The "ResolvePackageAssets" task failed unexpectedly.
System.IO.InvalidDataException: Found invalid data while decoding.
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheReader.ValidateHeader(BinaryReader reader, Byte[] settingsHash)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheReader.OpenCacheStream(Stream stream, Byte[] settingsHash)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheReader.CreateReaderFromDisk(ResolvePackageAssets task, Byte[] settingsHash)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.CacheReader..ctor(ResolvePackageAssets task)
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.ReadItemGroups()
   at Microsoft.NET.Build.Tasks.ResolvePackageAssets.ExecuteCore()
   at Microsoft.NET.Build.Tasks.TaskBase.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
```

It seems like the race specifically is on the second read of the cache file. On first read the cache file, the exception is caught. Then the cache file is written to disk and the re-read from disk. If that 2nd read fails, the exception is not caught and fails the build. The race condition seems to be if there are two of these "write+2nd reads" at the same time. I believe the symlinks are important here since the project's path is part of the "settings hash" used in the cache file's header.

### Further technical details
This is a CloudBuild scenario

[...]

PR #27171, opened 2022-08-15
Handle InvalidDataException when trying to write to cache and then read

ocallesp:
Based on the description provided in [#22538](https://github.com/dotnet/sdk/issues/22538)
It seems like : 
1 ) race condition specifically is on the second read of the cache file. 
2) On first read the cache file, the exception is caught.
3) Then the cache file is written to disk and the re-read from disk. 
4) If that 2nd read fails, the exception is not caught and fails the build.

After a code review we can see that `CreateReaderFromMemory()` can throw the exception `InvalidDataException`, which is not handled, when reader == null and `CanWriteToCacheFile` is true.

The change:
Catch the exception `InvalidDataException` in `CacheReader.ctor` so that we can try to execute again `CreateReaderFromMemory()` when `_reader == null`

What this change is saying: "If we fail to read from cache file in the second try due to Invalid data, then continue and use the memory cache".
Probably this is hiding an issue, and it would be good log it somewhere.

---------------------------
Note:
If handling the exception is not enough, then we can implement to retry N number of times before throwing an exception.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
@@ -469,6 +469,10 @@ namespace Microsoft.NET.Build.Tasks
                     // for the same project configured with the same intermediate directory. This can
                     // (for example) happen when design-time builds and real builds overlap.
                     //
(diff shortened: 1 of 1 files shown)


ocallesp:
A workaround was proposed for this

PR #27178, opened 2022-08-15
Port pre-download changes into the 6.0.4xx branch

marcpopMSFT:
We're seeing time-outs in our internal codeflow PRs to the 6.0 branches. Our timeouts mostly went away in main after this set of changes so let's try porting these downlevel to see if that helps as well.


nagilson (approved):
LGTM with ephemeral added

PR #27179, opened 2022-08-15
Lock net5 sdk packages

j0shuams:

PR #27199, opened 2022-08-16
Support different Windows SDK package versions for different target frameworks

dsplaisted:
If there are multiple matching WindowsSdkSupportedTargetPlatformVersion items, choose the one with the highest minimum version.  That way it is possible to use older packages when targeting older versions of .NET, and newer packages for newer versions of .NET.

Alternate implementation for #27179, and targeting release/6.0.4xx branch instead of main.

## Description

The Windows SDK projection package included by the .NET SDK has been shipped as .NET 5 binaries up until now.  The Windows SDK projection package is going to start shipping .NET 6 binaries, but right now the .NET SDK uses the same version of the package for all .NET TFMs (.NET 5, 6, 7, etc.).  Given someone can use the .NET 6 SDK to still target .NET 5, this can cause issues.  Due to this, this change allows to lock the version of the Windows SDK projection package used for .NET 5 while allowing for the version used for later TFMs to still be updated.

## Customer Impact
Enlighten the .NET SDK to the different Windows SDK packages so customers do not experience build failures when targeting net5 with .NET 6 or later SDK.

## Regression

No

## Risk 
Low. We added some locked (constant) versions for the .NET 5 packages, and added a few lines of code to the SDK to pick the package version where constraints on the target framework version are met.

## Testing 
The test for this (included in change) is a test app that targets .NET 6 and .NET 5. we see that in the case of .NET 6 we get the latest Windows SDK package version, and in the case of .NET 5 the last known good version of the Windows SDK package is used.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/CreateWindowsSdkKnownFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/CreateWindowsSdkKnownFrameworkReferences.cs
@@ -55,6 +55,8 @@ namespace Microsoft.NET.Build.Tasks
             else
             {
                 var normalizedTargetFrameworkVersion = ProcessFrameworkReferences.NormalizeVersion(new Version(TargetFrameworkVersion));
+
+                var matchingKnownFrameworkReferences = new List<(Version minimumNetVersion, TaskItem knownFrameworkReference)>();
                 foreach (var supportedWindowsVersion in WindowsSdkSupportedTargetPlatformVersions)
                 {
                     var windowsSdkPackageVersion = supportedWindowsVersion.GetMetadata("WindowsSdkPackageVersion");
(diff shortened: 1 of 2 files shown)


j0shuams (approved):
LGTM

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
```diff
@@ -403,6 +403,50 @@ public void ItUsesCorrectWindowsSdkPackVersion(string targetFramework, bool? use
...
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProject)
+                .WithProjectChanges(project =>
+                {
+                    //  Add items for available SDK versions for test
+                    var testItems = XElement.Parse(@"
+  <ItemGroup>
+    <WindowsSdkSupportedTargetPlatformVersion Remove=""@(WindowsSdkSupportedTargetPlatformVersion)"" />
+
+    <WindowsSdkSupportedTargetPlatformVersion Include=""10.0.22000.0"" WindowsSdkPackageVersion=""10.0.22000.26"" MinimumNETVersion=""6.0"" />
```
j0shuams, line 436: If we had multiple net5 versions here, and the TFM is net6-*, would we still get the range of net5 options? Or just the net6?

src/Tasks/Microsoft.NET.Build.Tasks/CreateWindowsSdkKnownFrameworkReferences.cs
```diff
@@ -55,25 +55,34 @@ protected override void ExecuteCore()
...
 
-                        knownFrameworkReferences.Add(CreateKnownFrameworkReference(windowsSdkPackageVersion, TargetFrameworkVersion, supportedWindowsVersion.ItemSpec));
+                        matchingKnownFrameworkReferences.Add((normalizedMinimumVersion, CreateKnownFrameworkReference(windowsSdkPackageVersion, TargetFrameworkVersion, supportedWindowsVersion.ItemSpec)));
+                        
                     }
                 }
+
+                //  If there are multiple matching WindowsSdkSupportedTargetPlatformVersion items, choose the one with the highest minimum version.  That way it is possible to
+                //  use older packages when targeting older versions of .NET, and newer packages for newer versions of .NET
+                var highestMinimumVersion = matchingKnownFrameworkReferences.Max(t => t.minimumNetVersion);
```
manodasanW, line 84: I think you might need to also take into account the OS version too.  For instance, lets say we added a new OS TFM that is only supported in like .NET 7 and we added an minimumNetVersion entry for it. I think that would mean that would be the only one in the list when targeting .NET 7.

As an example, if you take a look at what we have right now, 10.0.22621.0 is defined for .NET 6 or higher.  I feel like with this change, only that would get selected and not the others when it should be the highest for the .NET version taking into account OS version.

https://github.com/dotnet/installer/blob/b3e1a28fb3579e3636c8918842cc5d6031ac3629/src/redist/targets/GenerateBundledVersions.targets#L874

manodasanW (approved):

PR #27201, opened 2022-08-16
Update Blazor WASM Baselines for new files

TanayParikh:
Updated baselines for https://github.com/dotnet/sdk/pull/27149.


TanayParikh:
@javiercn your diffset for https://github.com/dotnet/sdk/pull/26966 was a lot more concise. Did you do additional post-processing after rerunning the tests with `GENERATE_SWA_BASELINES` defined?

javiercn:
@TanayParikh no, I did not.

TanayParikh:
> @TanayParikh no, I did not.

So is this fine to merge in that case? Surprised the diff is so large.

lewing:
The identity changes look a little odd, do we know what is causing them?

TanayParikh:
> do we know what is causing them?

cc @pavelsavara @radical @maraf

javiercn (approved):
~~Changes look good, provided we want the files there~~

Not sure why the diffs are changing so much. @TanayParikh we need to investigate, probably debugging through a test is the easiest way

lewing:
The new assemblies are showing up for a known reason, we're working on a solution to that but it shouldn't block rc1

MackinnonBuck:
@lewing Are we clear to merge this, then?

lewing:
@MackinnonBuck the count differences are fine but I'm not sure why the paths are different in this commit I'd like someone on the aspnetcore side to take a look asap.

@kg can you take a look from the runtime side and double check that we're publishing what we think we are.

kg:
System.Transactions.Local makes no sense to me and I can't tell where it's coming from, but the threading ones make sense.

lewing (approved):

lewing:
this should probably be merged into all the flow prs

PR #27206, opened 2022-08-16
[release/6.0.3xx] Support different Windows SDK package versions for different target frameworks

j0shuams:
If there are multiple matching WindowsSdkSupportedTargetPlatformVersion items, choose the one with the highest minimum version. That way it is possible to use older packages when targeting older versions of .NET, and newer packages for newer versions of .NET.

Same as #27199 but targeting 6.0.3xx

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
@@ -426,7 +426,7 @@ namespace Microsoft.NET.Build.Tests
                 TargetFrameworks = targetFramework
             };
 
-            var testAsset = _testAssetsManager.CreateTestProject(testProject)
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: targetFramework)
                 .WithProjectChanges(project =>
                 {
                     //  Add items for available SDK versions for test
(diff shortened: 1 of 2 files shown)


dsplaisted (approved):

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAWindowsDesktopProject.cs
```diff
@@ -403,6 +403,53 @@ public void ItUsesCorrectWindowsSdkPackVersion(string targetFramework, bool? use
...
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: targetFramework)
+                .WithProjectChanges(project =>
+                {
+                    //  Add items for available SDK versions for test
+                    var testItems = XElement.Parse(@"
+  <ItemGroup>
+    <WindowsSdkSupportedTargetPlatformVersion Remove=""@(WindowsSdkSupportedTargetPlatformVersion)"" />
+    <WindowsSdkSupportedTargetPlatformVersion Include=""10.0.22621.0"" WindowsSdkPackageVersion=""10.0.22621.26"" MinimumNETVersion=""6.0"" />
+    <WindowsSdkSupportedTargetPlatformVersion Include=""10.0.22000.0"" WindowsSdkPackageVersion=""10.0.22000.26"" MinimumNETVersion=""6.0"" />
+j0shuams marked this conversation as resolved.
```
dsplaisted, line 438: This line needs to be removed

dsplaisted (changes requested):

dsplaisted (approved):

PR #27209, opened 2022-08-16
Port predownload changes into .3xx

marcpopMSFT:


nagilson (approved):
LGTM with ephemeral added

PR #27213, opened 2022-08-16
Quote file paths for the BrotliCompress tool

baronfel:
**Description**

Quotes file paths passed to the BrotliCompress tool as command line arguments so that the generated response file has the expected format for paths that include spaces.

Closes https://github.com/dotnet/sdk/issues/26026

**Customer Impact** 

Customers with Blazor WebAssembly projects that have a space in the path to the project could not publish their projects at all. They had to change the folder structure to no longer include a space to make publishing work again.

**Regression**
**Yes**, this regressed in 6.0.301 due to a change in System.CommandLine response file parsing. We fixed this for the CLI overall in an earlier fix, but missed this tool due to lack of test coverage for this particular scenario, as well as these standalone tools not using the same CLI processing logic as the 'core' `dotnet` tool (for good reason, since tools like the brotli compress tool are internal-only and should be free to change). This is present in 6.0.400 and onwards as well.

**Risk** 

**Low** - the change only impacts file paths that are not already quoted and uses an existing mechanism for quoting. Automated tests have also been added to cover building/publishing in paths that contain a space.

### DIFF
--- a/src/BlazorWasmSdk/Tasks/BrotliCompress.cs
+++ b/src/BlazorWasmSdk/Tasks/BrotliCompress.cs
@@ -111,10 +111,10 @@ namespace Microsoft.NET.Sdk.BlazorWebAssembly
                 }
 
                 builder.AppendLine("-s");
-                builder.AppendLine(inputFullPath);
+                builder.AppendLine(Quote(inputFullPath));
 
                 builder.AppendLine("-o");
-                builder.AppendLine(outputFullPath);
+                builder.AppendLine(Quote(outputFullPath));
             }
 
             return builder.ToString();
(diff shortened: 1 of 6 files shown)


baronfel:
I should use [this test](https://github.com/dotnet/sdk/blob/aacf586d77bfde8226e0a8ab08a045eb46d045b7/src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/WasmBuildIntegrationTest.cs#L47-L72) as a base to make sure that that we can exercise these code paths. We should duplicate this with a space-ified project path to ensure coverage.

baronfel:
Having a hard time verifying that the test I added _actually_ results in a test asset with a space in it. VS doesn't like me at the moment to debug.

javiercn (approved):
Looks good.

rbhanda:
Approved for 6.0.4xx as well

baronfel:
Branding PR has merged bumping to 6.0.305, so merging this.

PR #27260, opened 2022-08-18
Rev HostModel dependencies' versions

am11:
This is to unblock #27149.

cc @rainersigwald


am11:
@jkoritzinsky has a better fix (removing the pinned version). Closing.

PR #27262, opened 2022-08-18
[23250] - dotnet test boolean flags options fixed

### LINKED ISSUE #23250: `dotnet test` string-backed, flag-style parameters require arguments incorrectly
### Describe the bug

There are several forwarded parameters in the `dotnet test` command parser that are of type `Option<string>` but act like boolean flags. When set these forward along multiple string arguments to a child process.  An example of this would be the `--blame-crash-collect-always` flag. This flag, and others like it, error in SDK `6.0.100` when called with no argument parameter, which isn't in line with their intended use.  These flags should be changed to not require an argument in order to work.

### To Reproduce

In a project on the 6.0.100 SDK, run `dotnet test --blame-crash-collect-always`

### Exceptions (if any)

> Required argument missing for option: --blame-crash-collect-always

### Further technical details
.NET SDK 6.0.100 release

There are two separate problems that I think we'll run into:
* The current implementation doesn't specify arity for these string options, and so System.CommandLine is defaulting to an arity of ExactlyOne for these string-typed parameters. This is what's at the core of requiring the argument. This can be fixed in a few different ways:
  * Explicitly set arity in the constructors of these specific options to ZeroOrOne
  * Implicitly set arity by changing the type of the options to `bool` or `bool?` (because bools and any nullable value type are set to ZeroOrOne arity by the library)
* However, If you change the arity to ZeroOrOne, the interaction of the `ForwardAsMany` lambdas changes so that those values are _always_ passed along. To change this the lambda need to be changed to check the value of the parsed argument being passed in, and if the value is `false` (for `bool` options), or `null` (for nullable options) then return `Enumerable.Empty<string>()` instead of the filled list.

Tunduk:
Fixes #23250 
@baronfel I saw your comment with the custom parser function but I think that custom parser functions are little bit complicated. When we use this functions we should care about error handling and localization. So, I resolved the issue in another way. 

Also, I just fixed issue for the `dotnet test` but there are another commands that should be fixed too

### DIFF
--- a/src/Cli/dotnet/OptionForwardingExtensions.cs
+++ b/src/Cli/dotnet/OptionForwardingExtensions.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.CommandLine;
-using System.CommandLine.Help;
 using System.CommandLine.Parsing;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


Tunduk:
@baronfel Hi!
Sorry for bothering you again, but could you please review this PR?

Tunduk:
If I understand correctly `dotnet test` and `dotnet test --blame-crash-collect-always false` should be equals but it's not true. 
`ForwardAsFlag` also checks a value of the option, so the option doesn't pass at all for 'false' value

Tunduk:
@baronfel Hi! I'm sorry for tagging you again, but I would like to discuss this issue again. As I checked there is still the problem with passing false to the command. So, the main problem with the scripting use case is not solved.

marcpopMSFT:

PR #27269, opened 2022-08-18
Don't run CG in public builds

marcpopMSFT:
This is so that CG stops tracking our public branches as we already track in our internal branches.


AraHaan:
for a second I thought CG meant crossgen2 and was scared 😄.

danmoseley:
Component governance ... Microsoft runs tools that check for vulnerable dependencies and other bad things.

nagilson (approved):

PR #27296, opened 2022-08-19
Switch localization to 7.0.1xx release branch

marcpopMSFT:
Since we've forked the  7.0.1xx branch, we want to move our localization there as well as presumably, this is where most new strings will come in. We'll stay on 7.0.Nxx branches until after 7.0.4xx ships.

Internal loc change request here: https://ceapex.visualstudio.com/CEINTL/_workitems/edit/694147

### DIFF
--- a/eng/build.yml
+++ b/eng/build.yml
@@ -15,13 +15,13 @@ parameters:
   timeoutInMinutes: 180
 
 jobs:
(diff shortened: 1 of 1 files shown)


nagilson:
LGTM, thanks for providing some context here and the doc change

nagilson (approved):

PR #27306, opened 2022-08-19
Update our helix pre-download code to try to pre-download for non-windows legs as well.

marcpopMSFT:


build/RunTestsOnHelix.sh
```diff
@@ -18,3 +18,8 @@ cp -a $HELIX_CORRELATION_PAYLOAD/t/TestExecutionDirectoryFiles/. $TestExecutionD
 dotnet new --debug:ephemeral-hive
 # avoid potetial concurrency issues when nuget is creating nuget.config
 dotnet nuget list source
+# We downloaded a special zip of files to the .nuget folder so add that as a source
+dotnet new nugetconfig
+dotnet nuget add source $DOTNET_ROOT/.nuget --configfile nuget.config
+ls $DOTNET_ROOT/.nuget
```
nagilson, line 24: Do we want to keep the ls here for debugging purposes or should that be removed?

nagilson (approved):

build/RunTestsOnHelix.sh
marcpopMSFT: I'll go ahead and remove it so it doesn't pollute the logs. I'll keep the source list in there as it's fairly small and can be very useful for debugging whereas the ls is only really for debugging this change.

PR #27322, opened 2022-08-21
Only reference the needed packages

ViktorHofer:
The Microsoft.CodeAnalysis package contains much that isn't needed by APICompat. Instead of referencing the "all-in-one package", use Microsoft.CodeAnalysis.Common and Microsoft.CodeAnalysis.CSharp.

This reduces the payload size of the tool package (which contains the CodeAnalysis assemblies) and minimizes dependencies.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -20,7 +20,7 @@
   </PropertyGroup>
   <!-- Production Dependencies -->
   <PropertyGroup>
-    <JabVersion>0.8.0</JabVersion>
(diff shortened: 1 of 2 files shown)


ViktorHofer:
This exposed a bug in Jab which I submitted a fix for: https://github.com/pakrym/jab/pull/118. We will need to wait until that one is in and then update the JabVersion that we use.

smasher164 (approved):

PR #27347, opened 2022-08-22
Add wasm-tools-net6 to the list of workloads that will pull in netweb in VS.

marcpopMSFT:


marcpopMSFT:
@baronfel , this will have to be updated once we settle on what we want the 7.0.100 workloads to look like in VS.  I'm having a hunch we'll want wasm-tools to install both net6 and net7 and we'll want a net6 and net7 version of wasm.

baronfel (approved):

Issue #16919, opened 2021-04-14
ApiCompatibility.AssemblySymbolLoader design should be revisited


safern:
The current model for AssemblySymbolLoader is that you create an instance of the object and then that object keeps track of the loaded assemblies and search paths. Also, the `IEnumerable<IAssemblySymbol> LoadAssemblies` APIs run synchronously and don't do this load lazily. The reason to do that synchronously is because it has a `HasLoadWarnings(out IEnumerable<string> warnings)` api, and when loading assemblies it populates that array. However that array is stored for the life time of the object and if you call multiple times `LoadAssemblies` it would return the aggregate of warnings found.

Also, we might be able to do a lazy loading of the assemblies using the enumerator model if that meets the customer expectations.

Some options for the context is having the user pass down a context as a parameter, that way we don't have customers accidentally cross the context boundaries and get the assembly that was previously loaded when they wanted a new assembly (assemblies are cached by name). 

Also the way we are loading references might need to be improved: https://github.com/dotnet/sdk/pull/16817#discussion_r612588244

cc: @ericstj @Anipik

PR #27362, opened 2022-08-23
Fix resolving assembly reference paths from dirs

ViktorHofer:
This fixes the previous behavior that only probed for assemblies in the passed in assemblies' directories. This regressed with https://github.com/dotnet/sdk/commit/c4bf096954a96e0a2be7cfcbfe0c4aacca35c038.

Making sure that directories from the passed in assemblies and their references are used to probe for new reference assemblies.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
```diff
@@ -253,30 +254,31 @@ private IReadOnlyList<MetadataReference> LoadFromPaths(IEnumerable<string> paths
...
                         result.Add(CreateOrGetMetadataReferenceFromPath(assembly, referenceAssemblyNamesToIgnore));
                     }
                 }
                 else if (File.Exists(path))
                 {
-                    directory = Path.GetDirectoryName(path);
+                    string? directory = Path.GetDirectoryName(path);
+                    // If a directory is passed in as a path, add that to the reference paths.
+                    // Otherwise, if a file is passed in, add its parent directory to the reference paths.
+                    if (!string.IsNullOrEmpty(directory))
```
akoeplinger, line 273: is there actually any scenario where a file exists but we can't determine its directory or is this more a defense-in-depth check?

akoeplinger (approved):

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs
ViktorHofer: Path.GetDirectoryName returns a nullable string which made me check for it for two reasons:
1. Make the compiler happy (lol)
2. Even though the code doesn't look like it would ever return null with these given inputs, this could theoretically change as the method already returns a nullable string.

So yes, a defense-in-depth check.

joperezr:
I know this wasn't introduced by your change, but I wonder if instead of us having our own resolvers, if we should just leverage something from the framework that already exists, like MetadataLoadContext, or ReferenceAssemblyPathResolver.

ViktorHofer:
Absolutely. That is tracked in https://github.com/dotnet/sdk/issues/16919 (see the comment that it links to, in which @ericstj talks about MetadataLoadContext.

PR #27367, opened 2022-08-23
Fix AttributesMustMatch InvalidOperationException

ViktorHofer:
Two fixes:
1. Don't run the rule if no exclusion files are supplied. Currently as msbuild provides an empty item group, the IEnumerable<T> would never be null.
2. The AttributesMustMatch.ConstantComparer type's Equals method must check the TypedConstant's kind as the Value property throws an InvalidOperationException if the kind is array and vice versa the Values property throws if the kind is not an array.

@smasher164 I noticed this when I ran your changes in dotnet/runtime. It might make sense to add an additional test with an attribute for which the TypedConstant is an array and a type.


akoeplinger (approved):

ViktorHofer:
@smasher164 just making sure that you saw this PR and my comment above:

> @smasher164 I noticed this when I ran your changes in dotnet/runtime. It might make sense to add an additional test with an attribute for which the TypedConstant is an array and a type.

PR #27380, opened 2022-08-23
add rule to detect change in method parameter names

### LINKED ISSUE #23908: [API Compat] add rule to make sure parameter names don't change
This rule should make sure that parameter names don't change.

I think this rule should probably be optional and be disabled by default.

smasher164:
The rule is disabled by default, and enabled via the option `enableRuleCannotChangeParameterName`.

Fixes #23908


src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
```diff
@@ -41,6 +41,7 @@ internal static class ValidateAssemblies
             string? suppressionFile,
             string? noWarn,
             string[]? excludeAttributesFiles,
+            bool enableRuleCannotChangeParameterName,
```
carlossanlop, line 44: Do we expect to keep adding more arguments to this Run method as we add more rules?

I wonder if instead of adding more arguments to enable/disable rules, and since this code is internal, we instead create a class with boolean properties, each defining a rule to enable/disable. That way you only pass one object to this Run method and any new rule gets added to the class. Similar suggestion of passing such object would apply for the RuleFactory constructor below.

carlossanlop: Example: We stopped adding new arguments to the FileStream class constructors. Instead, we created the FileStreamOptions class, and any new configuration options get added there.
https://docs.microsoft.com/en-us/dotnet/api/system.io.filestreamoptions.-ctor?view=net-6.0

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/ParameterNamesCannotChangeTests.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        public static TheoryData<string, string, CompatDifference[]> TestCases => new()
+        {
+            // Method
+            {
+                 @"
+namespace CompatTests
+{
+  public class First {
+    public void F(int a, string s) {}
```
carlossanlop, line 22: Would it be possible to add tests now that diff parameters that became nullable, or that are decorated with nullability attributes? Or are those not yet supported?

smasher164: I think nullability attributes would be covered by another rule that needs to be added. This rule only cares about parameter names.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
smasher164: I agree with your suggestion to add a class with all of the `enableRuleXX` options. However, I'd like to make sure I encounter at least one other instance where I need to add an `enableRule` flag before refactoring with that class.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/ParameterNamesCannotChangeTests.cs
ViktorHofer: Nullable operators are filtered out in all existing rules: https://github.com/dotnet/sdk/blob/1b6da85abc592947ca0bb4c7de01d106466258da/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs#L26.

That's because we will add a specific rule that checks for nullability.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
ViktorHofer: IMO the current system is OK for now (just adding new parameters when they are needed) and for .NET 8 we would ideally spend some time on a configuration system that allows to specify configurations more dynamically (i.e. via a json file).

ViktorHofer:
@smasher164 why would this rule be disabled by default? Is it because developers tend to change parameter names more often than other public API changes? Also when this rule is disabled by default, who would go the extra mile and explicitly enable it?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/RuleFactory.cs
```diff
@@ -34,7 +37,8 @@ public IRule[] CreateRules(RuleSettings settings, IRuleRegistrationContext conte
                 new CannotSealType(settings, context),
                 new EnumsMustMatch(settings, context),
                 new MembersMustExist(settings, context),
-                new AttributesMustMatch(settings, context, _excludeAttributesFiles)
+                new AttributesMustMatch(settings, context, _excludeAttributesFiles),
+                new CannotChangeParameterName(settings, context, _enableRuleCannotChangeParameterName),
```
ViktorHofer, line 41: Wouldn't it make more sense to not even instantiate the rule if `_enableRuleCannotChangeParameterName` is false here?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/RuleFactory.cs
```diff
@@ -15,10 +15,13 @@ public class RuleFactory : IRuleFactory
 
         private readonly IReadOnlyCollection<string>? _excludeAttributesFiles;
 
-        public RuleFactory(ICompatibilityLogger log, IReadOnlyCollection<string>? excludeAttributesFiles = null)
+        private readonly bool _enableRuleCannotChangeParameterName;
```
ViktorHofer, line 18: nit: we usually don't put an empty line between private members.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/ParameterNamesCannotChangeTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        [Theory]
+        [MemberData(nameof(TestCases))]
+        public void EnsureDiagnosticIsReported(string leftSyntax, string rightSyntax, CompatDifference[] expected)
+        {
+            TestRuleFactory s_ruleFactory = new((settings, context) => new CannotChangeParameterName(settings, context, true));
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory);
+            IEnumerable<CompatDifference> actual = differ.GetDifferences(left, right);
+            Assert.Equal(expected, actual);
```
ViktorHofer, line 93: For readability please follow the "Array Act Assert" best practice and separate the logical from each other:

```suggestion
            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
            TestRuleFactory s_ruleFactory = new((settings, context) => new CannotChangeParameterName(settings, context, true));
            ApiComparer differ = new(s_ruleFactory);

            IEnumerable<CompatDifference> actual = differ.GetDifferences(left, right);

            Assert.Equal(expected, actual);
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/ParameterNamesCannotChangeTests.cs
```diff
@@ -0,0 +1,96 @@
...
+",
+                 new CompatDifference[] {}
+            }
+        };
+
+        [Theory]
+        [MemberData(nameof(TestCases))]
+        public void EnsureDiagnosticIsReported(string leftSyntax, string rightSyntax, CompatDifference[] expected)
+        {
+            TestRuleFactory s_ruleFactory = new((settings, context) => new CannotChangeParameterName(settings, context, true));
```
ViktorHofer, line 88: I think this member should be a static readonly field declared in the scope of the class so that when more test methods get added, they can reuse the rule factory.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeParameterName.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            if (left is not IMethodSymbol leftMethod || right is not IMethodSymbol rightMethod)
+            {
+                return;
+            }
+
+            for (int i = 0; i < leftMethod.Parameters.Length; i++)
+            {
+                IParameterSymbol leftParam = leftMethod.Parameters[i];
+                IParameterSymbol rightParam = rightMethod.Parameters[i];
```
ViktorHofer, line 43: Don't you need to check that the rightMethod.Parameters.Length is equal to leftMethod.Parameters.Length?

smasher164:
> why would this rule be disabled by default? 

In the previous APICompat, this rule was optional. It is also described as such in the issue.

> Is it because developers tend to change parameter names more often than other public API changes?

As far as why it should be optional, yeah I believe changing parameter names is too common to flag.

>  Also if this rule is disabled by default, do you think that customers would go the extra mile and explicitly enable it?

If we aggregate all of these optional checks under one flag `--enable-optional-checks`, they might? Also, this rule is currently enabled in the runtime, correct?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeParameterName.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.DotNet.ApiCompatibility.Abstractions;
+
+namespace Microsoft.DotNet.ApiCompatibility.Rules
+{
+    /// <summary>
+    /// This class implements a rule to check that the parameter names between public methods do not change.
+    /// </summary>
+    public class CannotChangeParameterName : IRule
+    {
+        private readonly RuleSettings _settings;
```
ViktorHofer, line 15: It looks like you don't use settings anywhere and then you also don't need the instance field.

ViktorHofer:
> As far as why it should be optional, yeah I believe changing parameter names is too common to flag.

OK, yeah then it makes sense to make it optional. What I worry about is that this will become a rule that only we in dotnet/runtime know about and customers will never turn it on. Maybe we just need good documentation for all the available msbuild / CLI switches.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeParameterName.cs
smasher164: I thought so too, but when testing, I couldn't get the RuleRunner to pass in methods with a different number of parameters, even when the methods were otherwise identical.

However, I can add a check defensively. It just seems that the RuleRunner is taking care of that issue for us by not invoking the RunOnMemberSymbol for methods with a different number of parameters.

ViktorHofer: Yes, that's handled by the mapping system (more precisely by the passed in equality comparer): https://github.com/dotnet/sdk/blob/1b6da85abc592947ca0bb4c7de01d106466258da/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Abstractions/Mappers/TypeMapper.cs#L138.

That said, it's an "outside" component (in the meaning of outside the rule), therefor I would add a Debug.Assert check to formalize that requirement in debug code.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/RuleFactory.cs
smasher164: Depending on the way you approach this, that may require changing the `CreateRules` method's return type to a `List<IRule>`, since we won't know its size ahead of time. Depending on our decision there, I think we should save that for another change.

ViktorHofer (approved):

PR #27421, opened 2022-08-24
emit diagnostic for adding sealed to interface member

### LINKED ISSUE #26169: clarify API Compatibility rules for virtual keyword
The policy around adding a `virtual` keyword was decided back in https://github.com/dotnet/runtime/issues/21750. While this may technically be a breaking change, this method of making a method virtual has been used in the past in a backwards-compatible way. C# users in particular have treated adding a virtual keyword as a legal API change.

---

Interface members are dynamically dispatched, regardless of whether the `virtual` keyword is present. However, the dispatch for a `sealed` interface method can be `call`, although the C# compiler generates `callvirt`.

Even if the dispatch is always the same, this could *potentially* be a breaking change if an interface member is changed from `sealed` to `virtual` or vice-versa. For example:
```C#
using System;

B b = new C();
b.F(); // prints "B" if sealed, "C" otherwise

interface B {
    public sealed void F() {
        Console.WriteLine("B");
    }
}

public class C: B {
    public void F() {
        Console.WriteLine("C");
    }
}
```

---

Perhaps API compatibility checkers should introduce the above checks when an additional flag is passed? `Strict` mode is already taken to mean bidirectional checking, so maybe a `Full` flag to indicate that all of the rules are verified?

/cc @ericstj @MichalStrehovsky @joperezr @ViktorHofer

smasher164:
Adding the `sealed` keyword to a default interface method no longer makes it virtual.
This change adds a new diagnostic to flag such cases, although it is implemented as
part of the Virtual keyword rule.

The compatibility guidelines will have to be updated to cover this diagnostic.

Fixes #26169


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotAddOrRemoveVirtualKeyword.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
     public class CannotAddOrRemoveVirtualKeyword : IRule
     {
         private readonly RuleSettings _settings;
+        private static bool isSealed(ISymbol sym) => sym.IsSealed || (!sym.IsVirtual && !sym.IsAbstract);
```
ViktorHofer, line 17: nit: Private methods (and properties, which are also methods) are pascal cased in C#: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions.

ViktorHofer: And another nit, we usually separate methods from fields via a new line.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualKeywordTests.cs
```diff
@@ -203,5 +205,41 @@ public class First {
...
             Assert.Equal(expected, differences);
         }
+
+        [Fact]
+        public static void EnsureDiagnosticWhenAddingSealedToInterfaceMember()
+        {
+            if (IsNetFramework)
+            {
+                return;
+            }
```
ViktorHofer, line 215: The test project already compiles separately for .NET Framework: https://github.com/dotnet/sdk/blob/cae591210531dc36615f9cd78fc79af9834e1084/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Microsoft.DotNet.ApiCompatibility.Tests.csproj#L4

So you can just add a preprocessor symbol and exclude the whole method from the .NETFramework build:

```C#
// .NET Framework doesn't support default interface methods.
#if !NETFRAMEWORK
[Fact]
public static void EnsureDiagnosticWhenAddingSealedToInterfaceMember()
{
...
}
#endif
```

MichalStrehovsky:
As I wrote in the issue, if you want to consider this breaking:

```diff
using System;

B b = new C();
b.F(); // prints "B" if sealed, "C" otherwise

interface B {
-    public void F() {
+    public sealed void F() {
        Console.WriteLine("B");
    }
}

public class C: B {
    public void F() {
        Console.WriteLine("C");
    }
}
```

But this non-breaking:

```diff
using System;

B b = new C();
b.F(); // prints "B"

B d = new D();
d.F(); // prints "D", which is very likely not what the author intended because the author never knew of B.F.

interface B {
+    public void F() {
+        Console.WriteLine("B");
+    }
}

public class C: B {
}

public class D: B {
    public void F() {
        Console.WriteLine("D");
    }
}
```

It feels very inconsistent.

The problem is the introduction of a new virtual method (which is what dropping sealed keyword also does).

MichalStrehovsky:
Oh, wait, is this warning if `sealed` is added instead of removed?

Yes, that is absolutely breaking and not a gray area as was discussed in the issue:

```csharp
interface IFoo
{
    sealed void M() { }
}

interface IBar : IFoo
{
    // This will not compile once you add sealed to IFoo.M
    void IFoo.M() { }
}
```

smasher164:
> The problem is the introduction of a new virtual method (which is what dropping sealed keyword also does).

In that case, I can update this change to flag the removal of `sealed` on interface methods as well. Basically, if someone introduces an interface with a default method, they should keep it that way. Changing it to `sealed` or removing `sealed` is a breaking change.

smasher164:
> that is absolutely breaking and not a gray area

Right. My goal with this PR is to warn on a case that we decided was absolutely not a grey area.

MichalStrehovsky:
> In that case, I can update this change to flag the removal of `sealed` on interface methods as well.

If we want to flag that, we would want to flag introducing a new non-sealed default interface method as well - it's the consistency issue I called out in https://github.com/dotnet/sdk/pull/27421#issuecomment-1226604982.

FWIW, I don't think we consider that one a braking change in the runtime repo, even though it could introduce breaks in theory if the signature matches an unrelated existing method on an implementing class. This problem doesn't exist when introducing new virtual methods in classes because that one is a very clear source build break (the C# compiler will complain that you need a `new` keyword to hide the base implementation). The problem is specific to default interface methods. Cc @stephentoub on whether we would want to update breaking change rules. That one feels like gray area.

smasher164:
@MichalStrehovsky Regardless of whether we decide to flag the removal of `sealed` or introducing a non-`sealed` default interface method, do you think the diagnostic in this PR can be added independently (since we both agree that this is absolutely breaking)?

MichalStrehovsky:
> @MichalStrehovsky Regardless of whether we decide to flag the removal of `sealed` or introducing a non-`sealed` default interface method, do you think the diagnostic in this PR can be added independently (since we both agree that this is absolutely breaking)?

Yup.

ViktorHofer (approved):

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualKeywordTests.cs
```diff
@@ -15,6 +15,8 @@ public class CannotAddOrRemoveVirtualKeywordTests
     {
         private static readonly TestRuleFactory s_ruleFactory = new((settings, context) => new CannotAddOrRemoveVirtualKeyword(settings, context));
 
+        private static bool IsNetFramework => RuntimeInformation.FrameworkDescription.StartsWith(".NET Framework", StringComparison.OrdinalIgnoreCase);
+
```
ViktorHofer, line 19:
```suggestion
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/CannotAddOrRemoveVirtualKeywordTests.cs
```diff
@@ -203,5 +203,40 @@ public class First {
...
+";
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory, new ApiComparerSettings());
+            CompatDifference[] expected = new[]
+            {
+                CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.CannotAddSealedToInterfaceMember, string.Empty, DifferenceType.Added, "M:CompatTests.First.F")
+            };
+
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(new[] { left }, new[] { right });
```
ViktorHofer, line 236:
```suggestion
            IEnumerable<CompatDifference> differences = differ.GetDifferences(left, right);
```

ViktorHofer (approved):
I added minor upates to the test code (remove dead code and update styling).

PR #27432, opened 2022-08-24
Add error when failing to resolve SDK with workload sdk resolver

Forgind:
If one SDK resolver fails to load an SDK, it's no big deal, but if they all fail, it typically indicates the user did something wrong. We just resolve a generic "something went wrong" kind of error. We should instead log real errors for each SDK resolver indicating what went wrong. This does so for the workload SDK resolver.


nagilson (commented):
May you add a test to show this behavior logs a specific error now with the path (localization agnostic)? If we could include the version in the error, that would be optimal

marcpopMSFT:
Actually, we removed all loc testing we had as it wasn't finding issues. That being said, it's still good to target the error number as then you can edit the error message at any time without having to update a test.

nagilson:
Indeed -- I was vaguely implying to add a test that doesn't check for any specific content besides a determined path and version number, so it would be localization agnostic. I should've been more specific.

I like the idea of adding a specific error code. But awkwardly, while we have one mega file with all of the NETSDK Errors [here](https://github.com/dotnet/sdk/blob/main/src/Tasks/Common/Resources/Strings.resx), there are a ton of smaller CLI errors without error codes distributed in different places, doing that here may side-step the pattern.

marcpopMSFT:
@Forgind can you take a look at whether this is in a state to get merged?

marcpopMSFT:
He replied offline and see we could close this for now.

PR #27454, opened 2022-08-25
Add EnableRuleAttributesMustMatch switch

### LINKED ISSUE #27453: [ApiCompat] Add a feature flag to enable the AttributeMustMatch rule
Today we conditionally run the `AttributeMustMatch` rule if a attribute exclusion files are passed in. Instead, it makes more sense to add another feature flag in the frontends and only instantiate the rule if that feature flag is set. The attribute exclusion files input is just another input to the rule but shouldn't impact the lifetime of the rule.

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/27453


src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/build/Microsoft.DotNet.ApiCompat.ValidateAssemblies.Common.targets
```diff
@@ -19,8 +19,9 @@
       GenerateCompatibilitySuppressionFile="$(GenerateCompatibilitySuppressionFile)"
       CompatibilitySuppressionFilePath="$(CompatibilitySuppressionFilePath)"
       NoWarn="$(NoWarn)"
+      EnableRuleAttributesMustMatch="$(ApiCompatEnableRuleAttributesMustMatch)"
       ExcludeAttributesFiles="@(ApiCompatExcludeAttributesFile)"
-      EnableRuleCannotChangeParameterName="@(ApiCompatEnableRuleCannotChangeParameterName)"
+      EnableRuleCannotChangeParameterName="$(ApiCompatEnableRuleCannotChangeParameterName)"
```
ViktorHofer, line 24: @smasher164 oversight from your PR, in the msbuild you use `$` for referencing properties and `@` for referencing items.

smasher164 (approved):

PR #27500, opened 2022-08-26
apicompat: sort suppressions

smasher164:
Sort suppressions before writing them to disk.
1. DiagnosticId by id number (PKV < CP)
2. Left (null < non-null)
3. Right (null < non-null)
4. Target (null < non-null)

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


ViktorHofer (changes requested):
Let's not use Regex for sorting elements. That's fragile and could result in denial of service attacks if a timeout isn't provided. Imagine someone would provide a default suppression list with diagnostic ids that would result in the regex to never evaluate and compute indefinitely at the time they are being serialized again to disk.

I think the usage of Linq's OrderBy and ThenBy is a valid use case here (especially as this isn't a hot code path).

This should suffice:
```c#
            Suppression[] orderedSuppressions = _validationSuppressions
                .OrderBy(sup => sup.DiagnosticId)
                .ThenBy(sup => sup.Left)
                .ThenBy(sup => sup.Right)
                .ThenBy(sup => sup.Target)
                .ToArray();
```

ViktorHofer (approved):
Thanks 👌

PR #27537, opened 2022-08-29
Hot reload script injection improvements

MackinnonBuck:
### Summary

Improved hot reload script injection to be more reliable and provide better log messages when failure occurs.

### Description

Implemented the following improvements to hot reload script injection:
* The closing body tag will now be identified...
  * if some characters are uppercase or extra whitespace is inserted.
  * if the tag spans across multiple writes to the response stream (this is achieved by buffering the entire response before performing scirpt injection).
* New log messages will...
  * suggest disabling response compression if script injection fails and the `Content-Encoding` header exists.
  * suggest manually adding the script tag if the cause for failure is unknown.

Fixes https://github.com/dotnet/aspnetcore/issues/41772


javiercn (approved):
Looks great!

PR #27555, opened 2022-08-30
Add Dependancy on Microsoft.TemplateEngine.Mocks

dkurepa:
Microsoft.TemplateEngine.Mocks isn't a shipping package, so it will cause build errors once we turn on stable branding, TemplateEngine.Mocks won't be getting that.
The change was tested in the .NET 7 Test Build, on the 7.0.1xx-test branch


marek-safar:
Is it still needed?

dkurepa:
It is, we'll be getting build errors when we stabilize the branding if we don't fix it. @mmitche, could you take a look please

mmitche (approved):

PR #19860, opened 2021-08-18
Check that SDK and app archs match before setting `DOTNET_ROOT` when using `dotnet run`

mateoatr:
Fixes #19743. I tested this locally. The steps for reproing the fix are already in the above issue. For checking the app's architecture, this first checks for the `RuntimeIdentifier` property of the project to run, if not set, it'll grab the architecture from the `DefaultAppHostRuntimeIdentifier` prop.

Issue #19743, opened 2021-08-15
Preview 7 UX for Arm64/x64 co-existence on macOS


richlander:
Let's take a look at the experience that is offered for Arm64/x64 coexistence on macOS.

I will use this app for the test. https://github.com/dotnet/core/tree/main/samples/dotnet-runtimeinfo

## Installing .NET (correctly)

We haven't updated the installers yet, so I had to take some special steps to setup my machine as if they did. Here's what I did.

- Delete .NET from my machine
- Install various x64 .NET versions (whichever ones I want).
- Move all the files in `dotnet` to `dotnet\x64`
- Install .NET 6 Arm64.

That gets me a working install.

Here are the actual commands I used (skipping over the `.pkg` installs).

```bash
rich@MacBook-Air-M1-2020 share % pwd
/usr/local/share
rich@MacBook-Air-M1-2020 share % sudo -r dotnet
# install x64 versions
rich@MacBook-Air-M1-2020 share % sudo mv dotnet x64
rich@MacBook-Air-M1-2020 share % sudo mkdir dotnet
rich@MacBook-Air-M1-2020 share % sudo mv x64 dotnet
# install .NET 6 Arm64
```

## Running Arm64 apps

```bash
rich@MacBook-Air-M1-2020 dotnet-runtimeinfo % pwd 
/Users/rich/git/core/samples/dotnet-runtimeinfo
rich@MacBook-Air-M1-2020 dotnet-runtimeinfo % cat dotnet-runtimeinfo.csproj | grep Target
    <TargetFramework>net6.0</TargetFramework>
rich@MacBook-Air-M1-2020 dotnet-runtimeinfo %
rich@MacBook-Air-M1-2020 dotnet-runtimeinfo % dotnet run
**.NET information
Version: 6.0.0
FrameworkDescription: .NET 6.0.0-preview.7.21377.19
Libraries version: 6.0.0-preview.7.21377.19
Libraries hash: 91ba01788d4d83475fec3aea7c830376e08585da

**Environment information
OSDescription: Darwin 21.0.0 Darwin Kernel Version 21.0.0: Thu Jul 22 19:52:32 PDT 2021; root:xnu-8019.0.72.141.3~2/RELEASE_ARM64_T8101
OSVersion: Unix 12.0.0
OSArchitecture: Arm64
ProcessorCount: 8
```

That's good. That's expected.

## Running x64 apps -- take one

```bash
[...]

PR #27566, opened 2022-08-30
Use architecture specific DOTNET_ROOT variable name

tmat:
Corrects parsing of runtime identifier added by https://github.com/dotnet/sdk/pull/19860 and reuses the code in dotnet watch to do the same.

Other than parsing the rid the behavior of https://github.com/dotnet/sdk/pull/19860 is preserved. However, it doesn't seem  consistent with https://github.com/dotnet/sdk/issues/19743#issuecomment-901404323 in the case the architecture specified in rid is unknown. The implementation in this case sets DOTNET_ROOT variable, while it seems it shouldn't since an unknown architecture doesn't match the current architecture.

Implements https://github.com/dotnet/aspnetcore/issues/35793

### DIFF
--- a/src/Common/EnvironmentVariableNames.cs
+++ b/src/Common/EnvironmentVariableNames.cs
@@ -1,6 +1,12 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#pragma warning disable IDE0240 // Nullable directive is redundant (when file is included to a project that already enables nullable
+#nullable enable
+
+using System.Runtime.InteropServices;
+using System;
+
 namespace Microsoft.DotNet.Cli
 {
     static class EnvironmentVariableNames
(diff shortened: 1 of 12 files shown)


src/Common/EnvironmentVariableNames.cs
```diff
@@ -12,5 +17,50 @@ static class EnvironmentVariableNames
...
+            }
+
+            return runtimeIdentifier.Substring(separator + 1).ToLowerInvariant() switch
+            {
+                "arm" => Architecture.Arm,
+                "arm64" => Architecture.Arm64,
+                "x64" => Architecture.X64,
+                "x86" => Architecture.X86,
+                _ => null
+            };
```
kasperk81, line 57: avoid maintaining list of supported architectures supported by using the enum provided by runtime:
```suggestion
            string architecture = runtimeIdentifier.Substring(separator + 1).ToLowerInvariant();
            return Enum.TryParse<Architecture>(architecture, true, out _) ? architecture : null;
```

tmat: AFAIK, there is no guarantee that the names of the enum fields match the runtime identifier.

kasperk81: dotnet host also parses the string using strongly typed enum: https://github.com/dotnet/runtime/blob/d5e3a5c2ca46691d65c81d520cb95f13f7a94652/src/installer/managed/Microsoft.NET.HostModel/Bundle/TargetInfo.cs#L26 so it is not possible to do anything useful with unknown architecture by the time control reaches this method.

tmat: Thanks for the pointer. The code is parsing the runtime identifier like so: https://github.com/dotnet/runtime/blob/9d6396deb02161f5ee47af72ccac52c2e1bae458/src/installer/tests/Microsoft.NET.HostModel.Tests/Helpers/BundleHelper.cs#L136

kasperk81: i doubt bundle tests run anywhere other than those few ci machines, but the source of truth is this public enum: https://github.com/dotnet/runtime/blob/7a45201181d37329b7ab0bee541ed7c42b5d94b6/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Architecture.cs#L9 and it is only growing

tmat:
@mateoatr PTAL

src/Tests/dotnet.Tests/EnvironmentVariableNamesTests.cs
```diff
@@ -0,0 +1,78 @@
...
+    [InlineData("os-unknown", "v6.0", Architecture.Wasm, true, "DOTNET_ROOT_WASM")]
+    [InlineData("os-unknown", "v6.0", Architecture.Wasm, false, "DOTNET_ROOT_WASM")]
+    [InlineData("os-x86", null, Architecture.X86, true, "DOTNET_ROOT")]
+    [InlineData("os-x86", null, Architecture.X86, false, "DOTNET_ROOT(x86)")]
+    [InlineData("os-x86", "v5.0", Architecture.X86, true, "DOTNET_ROOT")]
+    [InlineData("os-x86", "v5.0", Architecture.X86, false, "DOTNET_ROOT(x86)")]
+    [InlineData("os-x86", "v6.0", Architecture.X86, true, "DOTNET_ROOT_X86")]
+    [InlineData("os-x86", "v6.0", Architecture.X86, false, "DOTNET_ROOT_X86")]
+    [InlineData("os-x64", "v6.0", Architecture.X86, false, null)]
+    public static void TryGetDotNetRootVariableName(string rid, string frameworkVersion, Architecture currentArchitecture, bool is64bit, string expected)
```
vitek-karas, line 72: Could you please add a test for the most common cases:
os-x64, (5.0 and 6.0), arch.x64, true - is by far the most common case
os-arm64, (6.0), arch.arm64, true - important due to osx-arm64 and win-arm64

src/Tests/dotnet.Tests/EnvironmentVariableNamesTests.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        var parsedVersion = EnvironmentVariableNames.TryParseTargetFrameworkVersion(frameworkVersion);
+        Assert.Equal(expected, EnvironmentVariableNames.TryGetDotNetRootVariableNameImpl(rid, "", parsedVersion, currentArchitecture, is64bit));
+        Assert.Equal(expected, EnvironmentVariableNames.TryGetDotNetRootVariableNameImpl("", rid, parsedVersion, currentArchitecture, is64bit));
+    }
```
vitek-karas, line 77: Add a case where target rid is passed in, and default rid is something different like `os-armv6`.

vitek-karas (approved):

vitek-karas:
> However, it doesn't seem consistent with https://github.com/dotnet/sdk/issues/19743#issuecomment-901404323 in the case the architecture specified in rid is unknown. The implementation in this case sets DOTNET_ROOT variable, while it seems it shouldn't since an unknown architecture doesn't match the current architecture.

That comment was mainly about the case where both architectures are known and the SDK architecture is different from the target architecture - in that case SDK should not set anything, since it doesn't know anything about the target's architecture situation on the machine - thus it should fallback to existing state on the machine to figure it out.
The case where the target architecture is unknown typically means in the "dotnet run" case that it was not specified (I think that can happen), in which case the desire for "dotnet run" is to run the app with the SDK environment - the app should default to SDK's RID in such case basically (which should be the behavior of "dotnet build" as well). I don't know if this is something which actually happens, since I think the msbuild targets will eventually default the target RID to the SDK RID, and thus the command should see the defaulted RID, but maybe it doesn't happen always.

I don't know what is the behavior for "dotnet watch" in this case. The logic should be:
* If the target is to run on the same architecture as the SDK -> modify the environment to use the SDK's runtime location
* If the target is to run on a different architecture -> don't modify anything, as the SDK has no clue what that architecture situation is.

I don't feel strongly about the behavior when the target architecture is unknown (I think and hope that it's rare anyway), but since "dotnet run" already defaults to "assume SDK's architecture", then I guess "dotnet watch" should be consistent.

vitek-karas:
@MarcoRossignoli as an FYI, since he implemented (and maintains) a similar logic in "dotnet test".

tmat:
@dotnet/dotnet-cli PTAL

marcpopMSFT:
@tmat the linked issue is from a year ago and listed for .net 8. Is this not too risky to take in rc2 at this time?

tmat:
@marcpopMSFT I'm fine moving to .NET 8. Note though that this also corrects the parsing of RID in `dotnet run` originally implemented in https://github.com/dotnet/sdk/pull/19860.

MarcoRossignoli:
cc: @Evangelink

tmat:
Moving to main.

dsplaisted:
@tmat Looks like some of the new EnvironmentVariableNamesTests are failing.

PR #27568, opened 2022-08-30
Merge 7.0.1xx rc1 to rc2

dsplaisted:


dsplaisted:
@dotnet/templating-engine-maintainers Could someone look at the failures here?  On Mac OS only, it looks like there's one failing tab completion test, and another unrelated test is failing with permission denied:

> System.UnauthorizedAccessException: Access to the path '/Users/helix-runner/.templateengine/dotnetcli/v7.0.100-ci' is denied.

marcpopMSFT:
@dsplaisted any test running in helix that does a dotnet new needs to include "--debug:ephemeral-hive". that'll fix the permissions issue.  Prague folks should still look at the tab completion failure.

dsplaisted:
> any test running in helix that does a dotnet new needs to include "--debug:ephemeral-hive". that'll fix the permissions issue. Prague folks should still look at the tab completion failure.

@vlada-shubina @dotnet/templating-engine-maintainers Should we make it so that when running tests by default we always use `--debug:ephemeral-hive`, so that we don't have to remember to do this for each test that uses templates?

donJoseLuis:
@JanKrivanek please take a look

JanKrivanek:
> > any test running in helix that does a dotnet new needs to include "--debug:ephemeral-hive". that'll fix the permissions issue. Prague folks should still look at the tab completion failure.
> 
> @vlada-shubina Vlada Shubina FTE @dotnet/templating-engine-maintainers Should we make it so that when running tests by default we always use `--debug:ephemeral-hive`, so that we don't have to remember to do this for each test that uses templates?

`DotnetNewCommand` throws if custom or virtualized location is not set explicitly - to prevent this exact situation (unintentional omission of the setting). Some tests uses virtualized location, some custom location - so I wouldn't force either as default.

dsplaisted:
This is green now, would someone like to review it for me?  @dotnet/dotnet-cli

nagilson:
@joeloff and I reviewed this together this morning and it looks good, assuming the only thing we care to backport are the MSBuild changes. @lewing's manifest changes may or may not need to go in, but I think Runtime is already doing that.

nagilson (approved):

PR #27580, opened 2022-08-31
Add output DebugSymbolsFiles to ResolvePackageAssets

### LINKED ISSUE #1458: New project system doesn't copy PDBs from packages
_From @bording on July 27, 2017 3:12_

When a NuGet package includes PDBs in its lib folder, the legacy project system copies the PDB as well as the assembly into the build output folder.

The new project system only copies the assembly.

We've recently added [source link support ](https://github.com/Particular/NServiceBus/pull/4862), and when using the new project system with a .NET Core target, everything works fine since the PDB is referenced correctly from the global NuGet package folder.

However, because the PDB isn't copied for a .NET Framework target, we don't get any of the benefit of our sourcelinked PDB.

It would be great if the new project system would behave like the legacy project system in this case, and also copy the PDB into the project's build output folder.

_Copied from original issue: dotnet/project-system#2667_

EDIT: We are also using this issue to track copying XML files

ocallesp:
fixes #1458

`DebugSymbolsFiles` outputs the list of absolute path to symbols files (.pdb) from runtime assemblies.
`ReferenceDocumentationFiles` outputs the list  of absolute path to xml files that enables debugging features.

**The change:**
Added `DebugSymbolsFiles` and `ReferenceDocumentationFiles` and `ResolvePackagesAssets` was modified to read/write `DebugSymbolsFiles` and `ReferenceDocumentationFiles` from/to the cache file. 

The debug symbols  files will be copied to the output directory when `CopyDebugSymbolFilesFromPackages` quals to true.

The reference documentation files will be copied to the output directory when `CopyDocumentationFilesFromPackages` quals to true.

------

For manual testing, I was using this sample project https://github.com/dotnet/sdk/issues/1383 and checking that the pdb are copied to the output directory of the consumer.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
@@ -12,7 +12,6 @@ using Microsoft.Build.Evaluation;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using NuGet.Common;
-using NuGet.Frameworks;
 using NuGet.ProjectModel;
 using NuGet.Versioning;
 
@@ -156,6 +155,9 @@ namespace Microsoft.NET.Build.Tasks
         [Required]
         public string DotNetAppHostExecutableNameWithoutExtension { get; set; }
 
+        /// <summary>
+        /// True indicates we are doing a design-time build. Otherwise we are in a build.
+        /// </summary>
         public bool DesignTimeBuild { get; set; }
 
         /// <summary>
@@ -230,6 +232,24 @@ namespace Microsoft.NET.Build.Tasks
 
         [Output]
         public ITaskItem[] PackageDependencies { get; private set; }
+
+        /// <summary>
+        /// List of symbol files (.pdb) related to NuGet packages.
+        /// </summary>
+        /// <remarks>
+        /// Pdb files to be copied to the output directory
+        /// </remarks>
+        [Output]
+        public ITaskItem[] DebugSymbolsFiles { get; private set;}
+
+        /// <summary>
+        /// List of xml files related to NuGet packages.
+        /// </summary>
+        /// <remarks>
+        ///  The XML files should only be included in the publish output if PublishReferencesDocumentationFiles is true
+        /// </remarks>
+        [Output]
+        public ITaskItem[] ReferenceDocumentationFiles { get; private set; }
 
         /// <summary>
         /// Messages from the assets file.
@@ -311,11 +331,13 @@ namespace Microsoft.NET.Build.Tasks
                 ApphostsForShimRuntimeIdentifiers = reader.ReadItemGroup();
                 CompileTimeAssemblies = reader.ReadItemGroup();
                 ContentFilesToPreprocess = reader.ReadItemGroup();
+                DebugSymbolsFiles = reader.ReadItemGroup();
                 FrameworkAssemblies = reader.ReadItemGroup();
                 FrameworkReferences = reader.ReadItemGroup();
                 NativeLibraries = reader.ReadItemGroup();
                 PackageDependencies = reader.ReadItemGroup();
                 PackageFolders = reader.ReadItemGroup();
+                ReferenceDocumentationFiles = reader.ReadItemGroup();
                 ResourceAssemblies = reader.ReadItemGroup();
                 RuntimeAssemblies = reader.ReadItemGroup();
                 RuntimeTargets = reader.ReadItemGroup();
@@ -510,7 +532,7 @@ namespace Microsoft.NET.Build.Tasks
                 BinaryReader reader = null;
                 try
                 {
-                    if (File.GetLastWriteTimeUtc(task.ProjectAssetsCacheFile) > File.GetLastWriteTimeUtc(task.ProjectAssetsFile))
+                    if (IsCacheFileUpToDate())
                     {
                         reader = OpenCacheFile(task.ProjectAssetsCacheFile, settingsHash);
                     }
@@ -537,6 +559,8 @@ namespace Microsoft.NET.Build.Tasks
                 }
 
                 return reader;
+
+                bool IsCacheFileUpToDate() => File.GetLastWriteTimeUtc(task.ProjectAssetsCacheFile) > File.GetLastWriteTimeUtc(task.ProjectAssetsFile);
             }
 
             private static BinaryReader OpenCacheStream(Stream stream, byte[] settingsHash)
@@ -650,6 +674,8 @@ namespace Microsoft.NET.Build.Tasks
             private bool MismatchedAssetsFile => !CanWriteToCacheFile;
 
             private const string NetCorePlatformLibrary = "Microsoft.NETCore.App";
+
+            private const char RelatedPropertySeparator = ';';
 
             public CacheWriter(ResolvePackageAssets task)
             {
@@ -776,11 +802,13 @@ namespace Microsoft.NET.Build.Tasks
                 WriteItemGroup(WriteApphostsForShimRuntimeIdentifiers);
                 WriteItemGroup(WriteCompileTimeAssemblies);
                 WriteItemGroup(WriteContentFilesToPreprocess);
+                WriteItemGroup(WriteDebugSymbolsFiles);
                 WriteItemGroup(WriteFrameworkAssemblies);
                 WriteItemGroup(WriteFrameworkReferences);
                 WriteItemGroup(WriteNativeLibraries);
                 WriteItemGroup(WritePackageDependencies);
-                WriteItemGroup(WritePackageFolders);                
+                WriteItemGroup(WritePackageFolders);
+                WriteItemGroup(WriteReferenceDocumentationFiles);
                 WriteItemGroup(WriteResourceAssemblies);
                 WriteItemGroup(WriteRuntimeAssemblies);
                 WriteItemGroup(WriteRuntimeTargets);
@@ -1089,6 +1117,61 @@ namespace Microsoft.NET.Build.Tasks
                         WriteMetadata(MetadataKeys.OutputPath, asset.OutputPath);
                         WriteMetadata(MetadataKeys.CodeLanguage, asset.CodeLanguage);
                     });
+            }
+
+            private void WriteDebugSymbolsFiles()
+            {
+                WriteDebugItems(
+                    p => p.RuntimeAssemblies,
+                    MetadataKeys.PdbExtension);
+            }
+
+            private void WriteReferenceDocumentationFiles()
+            {
+                WriteDebugItems(
+                    p => p.CompileTimeAssemblies,
+                    MetadataKeys.XmlExtension);
+            }
+
+            private void WriteDebugItems(
+                Func<LockFileTargetLibrary, IEnumerable<LockFileItem>> getAssets,
+                string extension)
+            {
+                foreach (var library in _runtimeTarget.Libraries)
+                {
+                    if (!library.IsPackage())
+                    {
+                        continue;
+                    }
+
+                    foreach (LockFileItem asset in getAssets(library))
+                    {
+                        if (asset.IsPlaceholderFile() || !asset.Properties.ContainsKey(MetadataKeys.RelatedProperty))
+                        {
+                            continue;
+                        }
+
+                        string itemSpec = _packageResolver.ResolvePackageAssetPath(library, asset.Path);
+
+                        string relatedExtensions = asset.Properties[MetadataKeys.RelatedProperty];
+
+                        foreach (string fileExtension in relatedExtensions.Split(RelatedPropertySeparator))
+                        {
+                            if (fileExtension.ToLower() == extension)
+                            {
+                                string xmlFilePath = Path.ChangeExtension(itemSpec, fileExtension);
+                                if (File.Exists(xmlFilePath))
+                                {
+                                    WriteItem(xmlFilePath, library);
+                                }
+                                else
+                                {
+                                    _task.Log.LogWarning(Strings.AssetsFileNotFound, xmlFilePath);
+                                }
+                            }
+                        }
+                    }
+                }
             }
 
             private void WriteFrameworkAssemblies()
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
@@ -297,6 +297,8 @@ Copyright (c) .NET Foundation. All rights reserved.
       <Output TaskParameter="Analyzers" ItemName="ResolvedAnalyzers" />
       <Output TaskParameter="ApphostsForShimRuntimeIdentifiers" ItemName="_ApphostsForShimRuntimeIdentifiersResolvePackageAssets" />
       <Output TaskParameter="ContentFilesToPreprocess" ItemName="_ContentFilesToPreprocess" />
+      <Output TaskParameter="DebugSymbolsFiles" ItemName="_DebugSymbolsFiles" />
+      <Output TaskParameter="ReferenceDocumentationFiles" ItemName="_ReferenceDocumentationFiles" />
       <Output TaskParameter="FrameworkAssemblies" ItemName="ResolvedFrameworkAssemblies" />
       <Output TaskParameter="FrameworkReferences" ItemName="TransitiveFrameworkReference" />
       <Output TaskParameter="NativeLibraries" ItemName="NativeCopyLocalItems" />
@@ -308,6 +310,14 @@ Copyright (c) .NET Foundation. All rights reserved.
       <Output TaskParameter="PackageFolders" ItemName="AssetsFilePackageFolder" />
       <Output TaskParameter="PackageDependencies" ItemName="PackageDependencies" />
     </ResolvePackageAssets>
+
+    <ItemGroup Condition="'$(CopyDebugSymbolFilesFromPackages)' == 'true'">
+      <ReferenceCopyLocalPaths Include="@(_DebugSymbolsFiles)"></ReferenceCopyLocalPaths>
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(CopyDocumentationFilesFromPackages)' == 'true'">
+      <ReferenceCopyLocalPaths Include="@(_ReferenceDocumentationFiles)"></ReferenceCopyLocalPaths>
+    </ItemGroup>
 
     <ItemGroup Condition="'$(UseAppHostFromAssetsFile)' == 'true'">
       <_NativeRestoredAppHostNETCore Include="@(NativeCopyLocalItems)"
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
@@ -19,6 +19,8 @@ using Xunit;
 using Xunit.Abstractions;
 using System.Text.RegularExpressions;
 using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Reflection.Metadata;
 
 namespace Microsoft.NET.Build.Tests
 {
@@ -530,7 +532,7 @@ namespace DefaultReferences
 
             testProject.PackageReferences.Add(new TestPackageReference("System.Net.Http", httpPackageVersion));
 
-            testProject.SourceFiles["Program.cs"] = 
+            testProject.SourceFiles["Program.cs"] =
                 (useAlias ? "extern alias snh;" + Environment.NewLine : "") +
 
                 @"using System;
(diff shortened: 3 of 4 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1091,6 +1104,27 @@ private void WriteContentFilesToPreprocess()
...
+            private void WriteDebugSymbolsFiles()
+            {
+                WriteItems(
+                    _runtimeTarget,
+                    package => package.RuntimeAssemblies,
+                    filter: asset => asset.Properties.ContainsKey(MetadataKeys.Related),
+                    writeMetadata: (package, asset) =>
+                    {
+                        asset.Properties.TryGetValue(MetadataKeys.Related, out string propertyValue);
+                        if (string.Compare(propertyValue, MetadataKeys.Pdb, true) == 0)
```
ocallesp, line 1116: Change this code to accept related = "pdb;xml"

ocallesp: https://github.com/NuGet/Home/blob/dev/proposed/2021/NuGetConsumePDBInPackageReference.md

ocallesp: @heng-liu clarified this off line. The correct is ".pdb;.xml"

dsplaisted:
Thanks @ocallesp, this is a good start.  Can you rebase and retarget this PR to the `release/7.0.1xx` branch?

Other feedback:

- I think there should be separate output items on the task for the PDB and XML files.  The XML files should only be included in the publish output if [`PublishReferencesDocumentationFiles`](https://github.com/dotnet/sdk/pull/1062) is true (I think that for the build output they should always be included).
- The XML files should come from the `compile` assets in the project.assets.json file, not the `runtime` assets.

Here's a sample assets file:

```json
  "targets": {
    "net7.0": {
      "LibraryTest/1.0.0": {
        "type": "package",
        "compile": {
          "ref/net7.0/LibraryTest.dll": {
            "related": ".xml"
          }
        },
        "runtime": {
          "lib/net7.0/LibraryTest.dll": {
            "related": ".pdb"
          }
        }
      },
      "Newtonsoft.Json/4.5.4": {
        "type": "package",
        "compile": {
          "lib/net40/Newtonsoft.Json.dll": {
            "related": ".pdb;.xml"
          }
        },
        "runtime": {
          "lib/net40/Newtonsoft.Json.dll": {
            "related": ".pdb;.xml"
          }
        }
      }
    }
  },
```

dsplaisted:
> If we apply this change to release/7.0.1xx branch, does the pipeline/insertions also apply this change to main ?

Yes, the code will eventually flow forward to later branches.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
```diff
@@ -830,6 +830,27 @@ public void It_places_package_satellites_correctly(bool crossTarget)
             outputDirectory.Should().HaveFile(@"fr\FluentValidation.resources.dll");
         }
 
+        [WindowsOnlyFact]
+        public void It_places_package_pdb_file_correctly()
+        {
+            var testAsset = _testAssetsManager
```
ocallesp, line 836: I will add two more unit-tests with 

PublishReferencesDocumentationFiles=true  Xml should exist in the output directory
PublishReferencesDocumentationFiles=false Xml should NOT exist in the output directory

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1165,11 +1166,12 @@ private void WriteDebugXmlFiles()
 
                         asset.Properties.TryGetValue(MetadataKeys.RelatedProperty, out string relatedExtensions);
 
-                        foreach (string fileExtension in relatedExtensions.Split(RelatedPropertySeparator))
+                        foreach (string relatedExtension in relatedExtensions.Split(RelatedPropertySeparator))
                         {
-                            if (fileExtension.ToLower() == extension)
+                            string ext = Path.GetExtension(relatedExtension);
```
ocallesp, line 1171: Handle "related": ".Http.xml;.HttpListener.xml;.Mail.xml;.NameResolution.xml;.NetworkInformation.xml;.Ping.xml;.Primitives.xml;.Requests.xml;.Security.xml;.ServicePoint.xml;.Sockets.xml;.WebClient.xml;.WebHeaderCollection.xml;.WebProxy.xml;.WebSockets.Client.xml;.WebSockets.xml"

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1091,6 +1125,57 @@ private void WriteContentFilesToPreprocess()
...
+                            continue;
+                        }
+
+                        string itemSpec = _packageResolver.ResolvePackageAssetPath(library, asset.Path);
+
+                        asset.Properties.TryGetValue(MetadataKeys.RelatedProperty, out string relatedExtensions);
+
+                        foreach (string fileExtension in relatedExtensions.Split(RelatedPropertySeparator))
+                        {
+                            if (fileExtension.ToLower() == extension)
```
ocallesp, line 1170: @dsplaisted 
My understanding is that only <assemblyName>.pdb, <assemblyName>.xml

I am not sure if we should handle xml extension like the in the example here:
"related": ".Http.xml;.HttpListener.xml;.Mail.xml;.NameResolution.xml;.NetworkInformation.xml;.Ping.xml;.Primitives.xml;.Requests.xml;.Security.xml;.ServicePoint.xml;.Sockets.xml;.WebClient.xml;.WebHeaderCollection.xml;.WebProxy.xml;.WebSockets.Client.xml;.WebSockets.xml"

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
```diff
@@ -830,6 +830,50 @@ public void It_places_package_satellites_correctly(bool crossTarget)
...
```
dsplaisted, line 836: Rather than checking in test projects into the `Assets` folder, if they are no super complicated I prefer creating them with the `TestProject` APIs.  See for example https://github.com/dotnet/sdk/blob/868f9e2b7494f1e1e44cd15da3d5741e1e27e4a4/src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs#L307-L318

The test projects should probably target `ToolsetInfo.CurrentTargetFramework`.

We should also be testing behavior for both build and publish.  I believe that the XML files should always be copied to the output directory for build, but only be copied to the publish directory if `PublishReferencesDocumentationFiles` is true.  So we should verify whether that's true (by checking the project to project reference behavior) and test the different cases.

dsplaisted (commented):
Looking good so far.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
dsplaisted: This may be a bug in NuGet (or a bug in how we specc'ed the behaivor).  Where did you get this example from?

src/Assets/TestProjects/DesktopUsingPackageWithPdbWithoutXml/DesktopUsingPackageWithPdbWithoutXml.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net46</TargetFramework>
+    <PublishReferencesDocumentationFiles>false</PublishReferencesDocumentationFiles>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NServiceBus" Version="6.4.0" />
```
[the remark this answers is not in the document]
ocallesp: Got it. I'll change the project name to "DesktopUsingPackageWithPdbAndXml.csproj"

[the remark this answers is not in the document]
ocallesp: Fixed.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -156,8 +155,16 @@ public sealed class ResolvePackageAssets : TaskBase
...
 
+        /// <summary>
+        /// True indicates we are doing a design-time build. Otherwise we are in a build.
+        /// </summary>
         public bool DesignTimeBuild { get; set; }
 
+        /// <summary>
+        /// Enable publishing XML documentation
+        /// </summary>
+        public bool PublishReferencesDocumentationFiles { get; set; }
```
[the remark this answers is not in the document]
ocallesp: Removed this input property.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1091,6 +1125,57 @@ private void WriteContentFilesToPreprocess()
...
+                {
+                    WriteDebugItems(
+                        p => p.CompileTimeAssemblies,
+                        MetadataKeys.XmlExtension);
+                }
+            }
+
+            private void WriteDebugItems<T>(
+                Func<LockFileTargetLibrary, IEnumerable<T>> getAssets,
+                string extension)
```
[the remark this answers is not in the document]
ocallesp: remove generic

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1091,6 +1125,57 @@ private void WriteContentFilesToPreprocess()
...
+                    foreach (T asset in getAssets(library))
+                    {
+                        if (asset.IsPlaceholderFile() || !asset.Properties.ContainsKey(MetadataKeys.RelatedProperty))
+                        {
+                            continue;
+                        }
+
+                        string itemSpec = _packageResolver.ResolvePackageAssetPath(library, asset.Path);
+
+                        asset.Properties.TryGetValue(MetadataKeys.RelatedProperty, out string relatedExtensions);
```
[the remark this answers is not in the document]
ocallesp: changed

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
ocallesp: The last commit is using `TestProject ` apis in unit-tests

ocallesp:
"only be copied to the publish directory if PublishReferencesDocumentationFiles is true. So we should verify whether that's true (by checking the project to project reference behavior) and test the different cases."

I was thinking about checking the publish directory to have the xml files when PublishReferencesDocumentationFiles  is true. So now I got confused with "checking the project to project reference behavior". Can you elaborate more on this ?

dsplaisted:
> @dsplaisted "_only be copied to the publish directory if PublishReferencesDocumentationFiles is true. So we should verify whether that's true (by checking the project to project reference behavior) and test the different cases._"
> 
> I was thinking about checking the publish directory to have the xml files when PublishReferencesDocumentationFiles is true. So now I got confused with "_checking the project to project reference behavior_". Can you elaborate more on this ?

Say you have an Exe project with a `ProjectReference` to a library project.  I believe that when you build the Exe project, that you will get the XML documentation file from the referenced project in the Exe project's output directory, but that XML file won't be in the publish directory unless you have `PublishReferencesDocumentationFiles` set to true.

We should verify whether that is indeed the case, and then make sure that with your fixes, a package reference behaves the same way in this respect as a project reference.

dsplaisted:
@ocalles Great work so far.

All those test changes made me reconsider whether we should copy PDBs from NuGet packages by default.  Any change we make in behavior has the potential to be breaking.  Since we are fairly late in the release cycle for .NET 7, I think that for .NET 7, we should not copy the PDBs from packages by default, but only do it if a project opts in via a property such as `CopyDebugSymbolsFromPackages`.  Early in .NET previews we can then enable that property by default, either for all projects or based on the TargetFramework depending on how we assess the risk and what feedback we can gather.  FYI @baronfel @marcpopMSFT @dotnet/dotnet-cli 

That would mean also mean that you wouldn't have to update all those tests.

Also it looks like the current version of your code doesn't respect the `PublishReferencesDocumentationFiles` property.

marcpopMSFT:
Having this off by default in 7.0.1xx would be unfortunate but understandable. We could enable it in 7.0.2xx or tie it to the 8.0 TFM and enable it early in 8.0. I'll defer to @baronfel somewhat on that decision as it's a tradeoff of risk versus customer value.

ocallesp:
This change broke many unit-tests due to the new xml documentation files added to the output directory.

ocallesp:
There are the list of broken tests:

------

Tests using Should().BeEquivalentTo() also fail
    It_provides_runtime_configuration_and_shadow_copy_files_via_outputgroup

-------
There is the list of unit-tests failing using OnlyHaveFiles()
1   It_copies_local_package_dependencies_on_build
1   It_does_not_copy_local_package_dependencies_when_requested_not_to
1   It_copies_local_specific_runtime_package_dependencies_on_build
1   It_does_not_copy_local_package_dependencies_for_lib_projects
1   It_copies_local_package_dependencies_for_lib_projects_when_requested_to
1   It_does_not_copy_local_package_dependencies_for_netstandard_projects
1   It_copies_local_package_dependencies_for_netstandard_projects_when_requested_to
1   It_does_not_copy_local_runtime_dependencies_for_netframework_projects
1   It_does_not_include_facades_from_nuget_packages
1   It_includes_shims_when_net471_app_references_netstandard16
1   It_stores_when_targeting_netcoreapp3
1   It_publishes_all_satellites_when_not_filtered
1   It_copies_nuget_package_dependencies
1   It_publishes_projects_with_simple_dependencies

31  compose_dependencies
31  compose_dependencies_noopt

36  compose_multifile

38  It_builds_the_project_successfully
38  It_builds_the_project_successfully_twice
38  The_clean_target_removes_all_files_from_the_output_folder
38  The_clean_target_removes_all_files_from_the_output_folder
38  It_builds_the_project_successfully

40  It_creates_a_documentation_file
40  It_allows_us_to_override_the_documentation_file_name
40  It_does_not_create_a_documentation_file_if_GenerateDocumentationFile_property_is_false
40  It_builds_the_library_successfully
40  Default_items_have_the_correct_relative_paths

43  The_clean_target_removes_all_files_from_the_output_folder
43  It_publishes_a_framework_dependent_RID_specific_runnable_output
43  It_builds_a_framework_dependent_RID_specific_runnable_output

47  It_builds_desktop_library_successfully_on_windows

50  It_builds_nondesktop_library_successfully_on_all_platforms
50  It_packs_nondesktop_library_successfully_on_all_platforms

68  Properly_changes_implicit_defines

72  It_publishes_the_project_correctly
72  It_does_not_build_when_nobuild_is_set

104  It_builds_the_project_successfully_when_RAR_does_not_find_all_references
104 Properly_changes_implicit_defines
104 It_builds_a_simple_vb_project
104 It_publishes_with_or_without_apphost
104 It_builds_the_library_successfully 
104 It_does_not_build_with_an_apphost_by_default_before_netcoreapp_3
104 It_builds_the_project_successfully_when_RAR_finds_all_references
104 It_builds_the_project_successfully_with_static_graph_and_isolation
104 It_cleans_the_project_successfully_with_static_graph_and_isolation

119 It_does_not_treat_nuget_refs_as_copylocal_false
119 It_builds_a_runnable_output

120 It_only_publish_selected_ResourceLanguages
120 It_copies_all_satellites_when_not_filtered
120 It_publishes_projects_with_simple_dependencies_with_filter_profile
120 It_does_not_publish_a_PackageReference_with_PrivateAssets_All
120 It_does_not_publish_a_PackageReference_with_Publish_false
120 It_publishes_a_PackageReference_with_PrivateAssets_All_and_Publish_true
120 It_only_publishes_selected_ResourceLanguages

324 It_should_publish_framework_dependent_for_2x

ocallesp:
@dsplaisted 
Copying pdb and xml to output and publish directories are now disable by default.

To enable it set `EnableCopySymbolsAndXmlDocsToOutputDir` to true.

ocallesp:
@dsplaisted changes done as suggested.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
```diff
@@ -850,5 +852,154 @@ public void It_allows_TargetFrameworkVersion_to_be_capitalized()
                 .Should()
                 .Pass();
         }
+
+        [WindowsOnlyTheory]
+        [InlineData("true", "true")]
+        [InlineData("true", "false")]
+        [InlineData("false", "true")]
+        [InlineData("false", "false")]
+        public void It_places_package_pdb_and_xml_files_in_output_directory(string enableCopyDebugSymbolFilesFromPackages, string enableDocumentationFilesFromPackages)
```
dsplaisted, line 861: I think it would be better to change these parameters to `bool` values.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
```diff
@@ -850,5 +852,154 @@ public void It_allows_TargetFrameworkVersion_to_be_capitalized()
...
+                TargetFrameworks = ToolsetInfo.CurrentTargetFramework,
+                IsExe = true,
+            };
+
+            testProject.PackageReferences.Add(new TestPackageReference("Microsoft.Build", "17.3.1"));
+            
+            testProject.AdditionalProperties.Add("CopyDebugSymbolFilesFromPackages", enableCopyDebugSymbolFilesFromPackages);
+            testProject.AdditionalProperties.Add("CopyDocumentationFilesFromPackages", enableDocumentationFilesFromPackages);
+
+            TestAsset testAsset = _testAssetsManager.CreateTestProject(testProject, testProject.Name);
```
dsplaisted, line 875: You'll need to pass an identifier to make sure the different instances of the theory use different test paths:

```suggestion
            TestAsset testAsset = _testAssetsManager.CreateTestProject(testProject, testProject.Name, enableCopyDebugSymbolFilesFromPackages.ToString() + enableDocumentationFilesFromPackages.ToString());
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
```diff
@@ -850,5 +852,154 @@ public void It_allows_TargetFrameworkVersion_to_be_capitalized()
...
+        public void It_places_package_pdb_and_xml_files_in_output_directory(string enableCopyDebugSymbolFilesFromPackages, string enableDocumentationFilesFromPackages)
+        {
+            var testProject = new TestProject()
+            {
+                Name = "DesktopUsingPackageWithPdbAndXml",
+                TargetFrameworks = ToolsetInfo.CurrentTargetFramework,
+                IsExe = true,
+            };
+
+            testProject.PackageReferences.Add(new TestPackageReference("Microsoft.Build", "17.3.1"));
```
dsplaisted, line 870: Is there a package that has an XML file in the ref folder but not in the lib folder?  That would also be a good test case.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
```diff
@@ -850,5 +852,154 @@ public void It_allows_TargetFrameworkVersion_to_be_capitalized()
...
+
+            HelperCheckPdbAndDocumentation(outputDirectory, "Microsoft.Build", enableCopyDebugSymbolFilesFromPackages, enableDocumentationFilesFromPackages);
+        }
+
+        [WindowsOnlyTheory]
+        [InlineData("true", "true")]
+        [InlineData("true", "false")]
+        [InlineData("false", "true")]
+        [InlineData("false", "false")]
+        public void It_places_package_pdb_and_xml_files_from_project_references_in_output_directory(string enableCopyDebugSymbolFilesFromPackages, string enableDocumentationFilesFromPackages)
```
dsplaisted, line 894: Given how NuGet works, I think we can probably delete this test.  There's not really a difference between a direct and indirect package reference from the SDK's point of view, they both get written to the assets file the same way.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -309,6 +311,14 @@ Copyright (c) .NET Foundation. All rights reserved.
       <Output TaskParameter="PackageDependencies" ItemName="PackageDependencies" />
     </ResolvePackageAssets>
 
+    <ItemGroup Condition="'$(CopyDebugSymbolFilesFromPackages)' == 'true'">
+      <ReferenceCopyLocalPaths Include="@(_DebugSymbolsFiles)"></ReferenceCopyLocalPaths>
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(CopyDocumentationFilesFromPackages)' == 'true'">
```
dsplaisted, line 318: If `PublishReferencesDocumentationFiles` is true, then we should also set `CopyDocumentationFilesFromPackages` to true, if it isn't already set.

That will also add a few more test cases.

dsplaisted (commented):
Looking very good, and pretty close to done I think.  Thanks!

dsplaisted:
@baronfel, What do you think about the property names `CopyDebugSymbolFilesFromPackages` and `CopyDocumentationFilesFromPackages`?  Should we consider something like `CopyPdbFilesFromPackages` and `CopyXmlFilesFromPackages`?  The current names try to match the existing `PublishReferencesDocumentationFiles` property.

Also pinging @dotnet/dotnet-cli in case anyone else has feedback.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
ocallesp: This will break many unit-tests https://github.com/dotnet/sdk/pull/27580#issuecomment-1240884830 because
`PublishReferencesDocumentationFiles` is set to true by default.

dsplaisted: In that case you're right, we shouldn't make the change I suggested.

Probably we should test the publish behavior when `PublishReferencesDocumentationFiles` is false but `CopyDocumentationFilesFromPackages` is true.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
ocallesp: I think we can leave it as strings since they represent a value assigned to a property and not a value to be used in a logic condition.

I do see that most of the unit-tests are using true instead of "true", they use this values in a logic condition.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
ocallesp: Not sure about this.

The original bug was reported using this scenario 
https://github.com/dotnet/sdk/issues/1383

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
ocallesp: done in last commit

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildADesktopExe.cs
ocallesp: I found one "system.diagnostics.debug" Version="4.3.0"

ocallesp: done in last commit

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
ocallesp: done in last commit

baronfel:
I'm ok with the current names - I'd rather align with the existing terminology to stay consistent. The `PublishDocumentationFile(s)` and `PublishReferencesDocumentationFiles` properties don't seem to be documented yet, so when this changes I'll bundle documenting them and these new properties in a single PR to dotnet/docs so that we're up-to-date and maybe get some SEO on the change.

I also agree with the plan to not enable the flag by default in this release, due to timing. We should explore turning it on for 7.0.200 and beyond, though - I think the existence of the `PublishReferencesDocumentationFiles` property and it being set to `true` by default is a strong enough signal to change this mid-release-cycle rather than waiting for the 8.0.x SDK series.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1091,6 +1119,61 @@ private void WriteContentFilesToPreprocess()
...
+                            continue;
+                        }
+
+                        string itemSpec = _packageResolver.ResolvePackageAssetPath(library, asset.Path);
+
+                        string relatedExtensions = asset.Properties[MetadataKeys.RelatedProperty];
+
+                        foreach (string fileExtension in relatedExtensions.Split(RelatedPropertySeparator))
+                        {
+                            if (fileExtension.ToLower() == extension)
```
dsplaisted, line 1160: This should probably use invariant comparison

```suggestion
                            if (fileExtension.ToLowerInvariant() == extension)
```

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1091,6 +1119,61 @@ private void WriteContentFilesToPreprocess()
...
+                            {
+                                string xmlFilePath = Path.ChangeExtension(itemSpec, fileExtension);
+                                if (File.Exists(xmlFilePath))
+                                {
+                                    WriteItem(xmlFilePath, library);
+                                }
+                                else
+                                {
+                                    _task.Log.LogWarning(Strings.AssetsFileNotFound, xmlFilePath);
```
dsplaisted, line 1169: This isn't really the right error message here.  Do we need to log a warning at all here?

dsplaisted (approved):
2 minor comments, looks good otherwise.

PR #27601, opened 2022-08-31
apicompat: treat IntPtr as nint and UIntPtr as nuint

### LINKED ISSUE #25566: Microsoft.Dotnet.Compatibility analyzer doesn't work with the NumericIntPtr feature
https://github.com/dotnet/runtime/pull/69627 updated the dotnet/runtime repo to utilize the `NumericIntPtr` feature from C# 11: https://github.com/dotnet/csharplang/issues/6065

This resulted in the compatibility analyzer misfiring with many errors such as:
```
/__w/1/s/.packages/microsoft.dotnet.compatibility/2.0.0-alpha.1.21525.11/build/Microsoft.NET.Compatibility.Common.targets(33,5): error CP0002: Member 'Microsoft.Extensions.Caching.Memory.PostEvictionDelegate.PostEvictionDelegate(object, System.IntPtr)' exists on lib/net6.0/Microsoft.Extensions.Caching.Abstractions.dll but not on lib/net7.0/Microsoft.Extensions.Caching.Abstractions.dll [/__w/1/s/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/Microsoft.Extensions.Caching.Abstractions.csproj]
##[error].packages/microsoft.dotnet.compatibility/2.0.0-alpha.1.21525.11/build/Microsoft.NET.Compatibility.Common.targets(33,5): error CP0002: (NETCORE_ENGINEERING_TELEMETRY=Build) Member 'Microsoft.Extensions.Caching.Memory.PostEvictionDelegate.PostEvictionDelegate(object, System.IntPtr)' exists on lib/net6.0/Microsoft.Extensions.Caching.Abstractions.dll but not on lib/net7.0/Microsoft.Extensions.Caching.Abstractions.dll
```

This is despite the underlying assemblies still containing the correct signatures. I noticed the version was out of date and so grabbed the latest package available, 2.0.0-preview.4.22252.4, to see the same error.

smasher164:
Previously, ApiCompat would consider `IntPtr` and `nint` (as well as `UIntPtr` and `nuint`) as different types, emitting diagnostics where there shouldn't have been. This change updates the display string used by the SymbolComparer to replace all instances of `IntPtr` with `nint` and `UIntPtr` with `nuint`.

Fixes #25566

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
@@ -29,7 +29,10 @@ namespace Microsoft.DotNet.ApiCompatibility.Extensions
             return format.WithParameterOptions(format.ParameterOptions & ~SymbolDisplayParameterOptions.IncludeParamsRefOut);
         }
 
-        internal static string ToComparisonDisplayString(this ISymbol symbol) => symbol.ToDisplayString(Format);//.Replace("System.IntPtr", "nint");
+        internal static string ToComparisonDisplayString(this ISymbol symbol) =>
+            symbol.ToDisplayString(Format)
+                  .Replace("System.IntPtr", "nint") // Treat IntPtr and nint as the same
+                  .Replace("System.UIntPtr", "nuint"); // Treat UIntPtr and nuint as the same
 
         internal static IEnumerable<ITypeSymbol> GetAllBaseTypes(this ITypeSymbol type)
         {
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Extensions
             return format.WithParameterOptions(format.ParameterOptions & ~SymbolDisplayParameterOptions.IncludeParamsRefOut);
         }
 
-        internal static string ToComparisonDisplayString(this ISymbol symbol) => symbol.ToDisplayString(Format);
+        internal static string ToComparisonDisplayString(this ISymbol symbol) => symbol.ToDisplayString(Format);//.Replace("System.IntPtr", "nint");
 
         internal static IEnumerable<ITypeSymbol> GetAllBaseTypes(this ITypeSymbol type)
         {
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
@@ -466,5 +466,41 @@ namespace CompatTests
             };
             Assert.Equal(expected, differences);
         }
+
+        [Fact]
+        public void NumericPtrFlaggedOnlyBeforeNet7()
+        {
+            string leftSyntax = @"
+namespace CompatTests
+{
+  using System;
+
+  public class First
+  {
+    public void F(IntPtr p) {}
+  }
+}
+";
+            string rightSyntax = @"
+namespace CompatTests
+{
+  public class First
+  {
+    public void F(nint p) {}
+  }
+}
+";
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory);
+
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(left, right);
+
+            CompatDifference[] expected =
+            {
+            };
+            // ReferenceID on .NET 7 for left member is "M:CompatTests.First.F(System.IntPtr)"
+            Assert.Equal(expected, differences);
+        }
     }
 }
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
@@ -468,7 +468,7 @@ namespace CompatTests
         }
 
         [Fact]
-        public void NumericPtrFlaggedOnlyBeforeNet7()
+        public void NumericPtrNotFlagged()
         {
             string leftSyntax = @"
 namespace CompatTests
@@ -478,6 +478,7 @@ namespace CompatTests
   public class First
   {
     public void F(IntPtr p) {}
+    public void G(UIntPtr p) {}
   }
 }
 ";
@@ -487,6 +488,7 @@ namespace CompatTests
   public class First
   {
     public void F(nint p) {}
+    public void G(nuint p) {}
   }
 }
 ";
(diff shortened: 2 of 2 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
```diff
@@ -29,7 +29,10 @@ private static SymbolDisplayFormat GetSymbolDisplayFormat()
             return format.WithParameterOptions(format.ParameterOptions & ~SymbolDisplayParameterOptions.IncludeParamsRefOut);
         }
 
-        internal static string ToComparisonDisplayString(this ISymbol symbol) => symbol.ToDisplayString(Format);
+        internal static string ToComparisonDisplayString(this ISymbol symbol) =>
+            symbol.ToDisplayString(Format)
+                  .Replace("System.IntPtr", "nint") // Treat IntPtr and nint as the same
```
ericstj, line 34: Will this create confusing diagnostics when someone is actually using `IntPtr`?  Say I had an API that used `IntPtr` and I removed it on the right, what diagnostic do we emit?

ViktorHofer: This code path doesn't impact the `Target` that is being emitted by rules and APICompat. The `ToComparisonDisplayString` helper function is only used by the mapping infrastructure to map members together.

I.e. https://github.com/dotnet/sdk/blob/0975222d0666c8a5a0240e07830a71a900e33b96/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs#L73 doesn't use that function to for encoding the target.

smasher164: This string is only used for comparing symbols, so the diagnostic that is emitted in that case is still `"CP0002 : Member 'CompatTests.First.H(System.IntPtr)' exists on left but not on right"`.

ViktorHofer (approved):
Looks good. Thanks for adding a test for it.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
ericstj: If we do end up having to normalize this in comparisons, it might be better to do here:
https://github.com/dotnet/sdk/blob/0975222d0666c8a5a0240e07830a71a900e33b96/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/DefaultSymbolsEqualityComparer.cs#L14

And only after the equality check fails, and only if references are not provided.  I'd like us to first see if we can get roslyn to behave as if the feature is supported.

ViktorHofer: IMO the `ToComparisonDisplayString` method is the right place as the name already suggests that it calculates a display string **for comparison**.

ericstj: I see.  That makes sense.  I was mistaken about where this was used.  Is there any concern about applying this transformation even when we have references?

smasher164:
@ericstj
>What if you did the replacement the other way?

This still works. Is there a reason we might prefer one direction over the other?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
ViktorHofer: This is interesting. I also considered doing the replacement the other way but imagined that `Replace("nint", ...)` would have more false positives than `Replace("System.IntPtr", ...)` because of length of the string and the "System classification".

Are dots `.` allowed in many members types in C#/VB? I.e. a method comparison string `Companint` would be replaced to `CompaSystem.IntPtr`.

ericstj: > This still works. Is there a reason we might prefer one direction over the other?

`IntPtr` has been around forever so there will be a lot more uses of it that actually mean `IntPtr` and not `nint`.  `nint` is new and while it is important, I don't expect it would be used as much as `IntPtr`.  From the same perspective, if we can make this an optional normalization that's only applied when references are absent it could further reduce the amount of unnecessary work that is done.

smasher164: Testing this on .NET 472 and .NET 7,
```C#
if (RuntimeFeature.IsSupported("NumericIntPtr"))
{
    return symbol.ToDisplayString(Format);
}
return symbol.ToDisplayString(Format)
    .Replace("nint", "System.IntPtr") // Treat nint and IntPtr as the same
    .Replace("nuint", "System.UIntPtr"); // Treat nuint and UIntPtr as the same
```
tests fail on .NET 7. It seems like even when NumericIntPtr is defined, Roslyn returns different display strings.

bartonjs: Off the top of my head the two concerning places are nested types (which probably here use a `+` instead of `.`) and explicit interface implementations (e.g. the full method name of `IEnumerable.GetEnumerator() => GetEnumerator();` I believe is `System.IEnumerable.GetEnumerator`.  So while "nint" => "System.IntPtr" probably wouldn't create any false positives, '.' probably isn't a guaranteed "eh, it's fine".

ViktorHofer: Agreed that nint is fairly new but it's not unlikely that `nint` and `nuint` word parts are used in a member's name. I would expect that replacing "nint" with "System.IntPtr" would result be more false positives than the other way around. That said, these false positives shouldn't have any negative impact as far as I can tell.

ericstj: >if (RuntimeFeature.IsSupported("NumericIntPtr"))

This is mixing *tooling runtime* with target runtime.  The check would need to be against whatever the target references that were used when compiling the assembly.  Calling the feature `RuntimeFeature.IsSupported` in the tool runtime has no relevance since we might be using an older runtime to test assemblies built for a newer one. 

The way I was imagining this to work was we would add some parameter to `DefaultSymbolsEqualityComparer` like `bool honorRuntimeFeatures` and set that as `withReferences == true`.  Then we could say that when it is false we'd do the hack-up in `DefaultSymbolsEqualityComparer.GetKey` to normalize those symbols.

ericstj: Hmm, the premise of the `withReferences` suggestion was based on the fact that we only face this problem when we don't have the matching references.  However, we probably are in this state even when references are provided, since we don't have references for the left-hand-side and that might be the runtime that doesn't support the feature :(

I was really trying to get this to a place where we didn't need this hack indefinitely but I guess that might not be possible at the moment.

ericstj (approved):

ericstj:
Approved this as-is.  I think given our current architecture (references present only for right hand side) and the only source for this feature information in the compiler coming from references - we'll always be in a state where we need to normalize these types of metadata differences ourselves.  That's unfortunate but I can't think of a better solution.

PR #27646, opened 2022-09-02
Simplify error message

jkotas:
The "When targeting .NET 6/7 or higher" part of the error message is superfluous


jkotas:
Do I need to do something special for localized versions of the string?

LakshanF (approved):
Thanks!

PR #27684, opened 2022-09-05
Bugfix: Report error on conditional params authoring errors

JanKrivanek:
**Problem:** Template authoring issues with conditional parameters were not clearly surfaced to user (generic internal error [return code 70](https://github.com/dotnet/templating/blob/main/docs/Exit-Codes.md#70) was returned instead)

**Solution:** Properly handle error codes returned from `TemplateCreator`


vlada-shubina (approved):
Having a test on this would be good.

PR #27699, opened 2022-09-06
Remove various timeouts

### LINKED ISSUE #40587: CLI Hot reload Blazor wasm just doesn't work.
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

There were many issues about dotnet watch not working with blazor wasm. #38561, #36723, #40091 etc. It was mentioned there that the issue is fixed in .net 6.0.2. But it's not. The issue is there still. Furthermore, there are no responses now on those issues when it was informed that in .net6.0.2 it's not working. Very disappointing behavior.

dotnet watch --verbose logs
```
watch : Project supports hot reload and was configured to run with the default run-command. Watching with hot-reload
watch : Hot reload enabled. For a list of supported edits, see https://aka.ms/dotnet/hot-reload. Press "Ctrl + R" to restart.
watch : Building...
watch : Started 'C:\Program Files\dotnet\dotnet.exe' '' with process id 9496
  Restored C:\Users\ziaul\Sources\BlazorWasmApp\BlazorWasmApp.csproj (in 1.39 sec).
  BlazorWasmApp -> C:\Users\ziaul\Sources\BlazorWasmApp\bin\Debug\net6.0\BlazorWasmApp.dll
  BlazorWasmApp (Blazor output) -> C:\Users\ziaul\Sources\BlazorWasmApp\bin\Debug\net6.0\wwwroot
watch : Process id 9496 ran for 27116ms
watch : Running MSBuild target 'GenerateWatchList' on 'C:\Users\ziaul\Sources\BlazorWasmApp\BlazorWasmApp.csproj'
watch : Started 'C:\Program Files\dotnet\dotnet.exe' '' with process id 9176
watch : Process id 9176 ran for 3657ms
watch : Watching 28 file(s) for changes
watch : dotnet-watch is configured to launch a browser on ASP.NET Core application startup.
watch : Configuring the app to use browser-refresh middleware.
watch : Refresh server running at wss://localhost:58192,ws://localhost:58193.
watch : HotReloadProfile: BlazorWebAssembly.
watch : Waiting for a browser to connect
watch : Started 'dotnet' '"C:\Users\ziaul\.nuget\packages\microsoft.aspnetcore.components.webassembly.devserver\6.0.2\build\../tools/blazor-devserver.dll" --applicationpath "C:\Users\ziaul\Sources\BlazorWasmApp\bin\Debug\net6.0\BlazorWasmApp.dll"' with process id 8976
watch : Running dotnet with the following arguments: run
watch : Started
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: https://localhost:7020
watch : Launching browser.
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5152
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: C:\Users\ziaul\Sources\BlazorWasmApp
watch : Hot reload capabilities: .
watch : File changed: C:\Users\ziaul\Sources\BlazorWasmApp\Pages\Index.razor.
watch : No deltas modified. Applying changes to clear diagnostics.
watch : Received 1 from browser in [Count: 1, MessageType: Binary, EndOfMessage: True].
watch : No hot reload changes to apply.
watch : Hot reload change handled in 6599.4355ms.
```

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version

6.0.200

### Anything else?

_No response_

tmat:
From the discussion on https://github.com/dotnet/aspnetcore/issues/40587 it appears that these timeouts are causing failures in real world usage of dotnet watch. 

I don't think we should use arbitrary timeouts on these code paths as long as they are cancellable. If some communication gets stuck during Hot Reload for some reason the user can always trigger cancelation of the change application.

Fixes https://github.com/dotnet/aspnetcore/issues/40587

### DIFF
--- a/src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
@@ -55,8 +55,7 @@ namespace Microsoft.DotNet.Watcher.Tools
                 var buffer = ArrayPool<byte>.Shared.Rent(32 * 1024);
                 try
                 {
-                    // We'll query the browser and ask it send capabilities. If the browser does not respond in a short duration, we'll assume something is amiss and return
-                    // baseline capabilities.
+                    // We'll query the browser and ask it send capabilities.
                     var response = await context.BrowserRefreshServer.ReceiveAsync(buffer, cancellationToken);
                     if (!response.HasValue || !response.Value.EndOfMessage || response.Value.MessageType != WebSocketMessageType.Text)
                     {
--- a/src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
@@ -24,8 +24,6 @@ namespace Microsoft.DotNet.Watcher.Tools
         private readonly IReporter _reporter;
         private int _sequenceId;
 
-        private static readonly TimeSpan VerifyDeltaTimeout = TimeSpan.FromSeconds(5);
-
         public BlazorWebAssemblyDeltaApplier(IReporter reporter)
         {
             _reporter = reporter;
@@ -59,9 +57,7 @@ namespace Microsoft.DotNet.Watcher.Tools
                 {
                     // We'll query the browser and ask it send capabilities. If the browser does not respond in a short duration, we'll assume something is amiss and return
                     // baseline capabilities.
-                    var response = await context.BrowserRefreshServer.ReceiveAsync(buffer, cancellationToken)
-                        .AsTask()
-                        .WaitAsync(TimeSpan.FromSeconds(15), cancellationToken);
+                    var response = await context.BrowserRefreshServer.ReceiveAsync(buffer, cancellationToken);
                     if (!response.HasValue || !response.Value.EndOfMessage || response.Value.MessageType != WebSocketMessageType.Text)
                     {
                         return _baselineCapabilities;
@@ -74,15 +70,10 @@ namespace Microsoft.DotNet.Watcher.Tools
                     var result = values.Split(' ').ToImmutableArray();
                     return result;
                 }
-                catch (TimeoutException)
-                {
-                }
                 finally
                 {
                     ArrayPool<byte>.Shared.Return(buffer);
                 }
-
-                return _baselineCapabilities;
             }
         }
 
@@ -104,7 +95,7 @@ namespace Microsoft.DotNet.Watcher.Tools
             });
 
             await context.BrowserRefreshServer.SendJsonWithSecret(sharedSecret => new UpdatePayload { SharedSecret = sharedSecret, Deltas = deltas }, cancellationToken);
-            return await VerifyDeltaApplied(context, cancellationToken).WaitAsync(VerifyDeltaTimeout, cancellationToken);
+            return await VerifyDeltaApplied(context, cancellationToken);
         }
 
         public async ValueTask ReportDiagnosticsAsync(DotNetWatchContext context, IEnumerable<string> diagnostics, CancellationToken cancellationToken)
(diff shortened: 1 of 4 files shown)


tmat:
@javiercn @SteveSandersonMS

arunchndr (approved):

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
```diff
@@ -59,9 +57,7 @@ async Task<ImmutableArray<string>> GetApplyUpdateCapabilitiesCoreAsync()
                 {
                     // We'll query the browser and ask it send capabilities. If the browser does not respond in a short duration, we'll assume something is amiss and return
                     // baseline capabilities.
-                    var response = await context.BrowserRefreshServer.ReceiveAsync(buffer, cancellationToken)
-                        .AsTask()
-                        .WaitAsync(TimeSpan.FromSeconds(15), cancellationToken);
+                    var response = await context.BrowserRefreshServer.ReceiveAsync(buffer, cancellationToken);
```
mkArtakMSFT, line 60: How safe is it to remove this timeout? It was there for a reason?
Looking at the above comment as well as the code, it looks like there is a possibility for this call to hang for a long period of time. Feels like this can stall the execution.

tmat: I don't think there is a reason for having the timeout at all. The user can cancel any of these calls.

tmat: If this call times out the rest of dotnet watch functionality is broken.

tmat: If I run `dotnet watch` on a Bazor app and it is not able to connect to the browser in a given time then is it preferable to terminate `dotnet watch` with a message that it's not able to connect, or wait for the browser indefinitely with "Waiting to connect ..." message printed out to the console?

tmat: The user can hit Cltr+C any time btw to cancel.

mkArtakMSFT: I think it is still beneficial to have some timeout and print "unable to connect to browser after waiting for X minutes". But it's your call, @tmat.

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
```diff
@@ -123,33 +113,18 @@ public async ValueTask ReportDiagnosticsAsync(DotNetWatchContext context, IEnume
...
-                    }
-
-                    if (IsDeltaReceivedMessage(result.Value))
-                    {
-                        // 1 indicates success.
-                        return _receiveBuffer[0] == 1;
-                    }
-                }
+                // A null result indicates no clients are connected. No deltas could have been applied in this state.
+                _reporter.Verbose("No client is connected to ack deltas");
```
javiercn, line 120: Should this be reported as verbose? Wouldn't it be better as a warning/error. If I understand this correctly, at this point things are in a broken state, aren't they?

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
SteveSandersonMS: As long as it's clear to the user that the problem is "we haven't yet managed to connect" then TBH I don't think it matters whether we eventually give up or wait for a ctrl+c.

In both cases it's not working and the user knows what the problem is, so it's more or less equivalent.

SteveSandersonMS (approved):
While I don't have loads of context on this, I understand you're looking for someone on the Blazor team to approve, and it looks reasonable to me so approving.

It would be great to address [@javiercn's comment](https://github.com/dotnet/sdk/pull/27699#pullrequestreview-1100441708) if possible though.

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
tmat: I have added code that prints "Waiting for application connection ..." warning every 5s until the application connection is established (prints nothing if it takes less than 5s). I also added regular output "Application ready for Hot Reload" once initialization is complete. Although changes can be made to source files earlier they can't be applied until we finish initialization, so having output indicate that seems useful.

image

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
tmat: ReceiveAsync already reports any errors.

I'll clean this up a bit later on. It's not clear to me why this [catch](https://github.com/dotnet/sdk/blob/main/src/BuiltInTools/dotnet-watch/Filters/BrowserRefreshServer.cs#L225) is in the loop over `_clientSockets`. I would expect it to be outside of the loop. I'd expect the semantics of the method be: find the first socket that's open and use it to receive data. If it fails then don't try another one and just fail the whole thing.

PR #27700, opened 2022-09-06
Remove timeouts from tests

tmat:
The timeouts are arbitrary and make tests flaky.


arkalyanms:
Consider: Test timeouts have been useful in capturing product code delays. That said, I am not familiar with the sensitivity of the scenario to recommend a replacement time that makes sense for these scenarios. 5 minutes for browser launch seems like a good ceiling value.

arunchndr (approved):

tmat:
I'd prefer we have dedicated performance tests for doing that instead of timeouts in tests that run on random VMs that do not use dedicated resources and thus their execution time varies a lot.

PR #27742, opened 2022-09-07
Remove Jab dependency

eerhardt:
The DependencyInjection usage isn't necessary in ApiCompat. It can simplified to using Lazy<T> for singleton services, and just creating objects directly when necessary.

This removes the dependency on an external library, and simplifies source-build, since we no longer need to build another repo.


ViktorHofer (approved):
As discussed offline, this makes sense and looks good! Thanks a lot

eerhardt:
Test failures are #27767. Merging.

joperezr:
> As discussed offline, this makes sense and looks good! 

I'd be interested in knowing the conclusions of the chat, since as you may recall I gave this feedback when the dependency was originally added.

joperezr (approved):
Changes look good, thanks @eerhardt!

ViktorHofer:
> I'd be interested in knowing the conclusions of the chat, since as you may recall I gave similar feedback when the dependency was originally added.

Shared the offline Teams conversation around that with you.

PR #27757, opened 2022-09-08
Don't read from suppression file it doesn't exist.

ViktorHofer:
Don't read from the passed in suppression file if errors will be baselined and the file doesn't exist. I.e. when invoking the CLI and passing in `--generate-suppression-file` and `suppression-file ...` path, if the file doesn't exist this would result in a FileNotFoundException instead of not reading from it.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
@@ -163,7 +163,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
 
         private HashSet<Suppression> ParseSuppressionFile(string? suppressionFile)
         {
-            if (string.IsNullOrEmpty(suppressionFile?.Trim()))
+            if (string.IsNullOrWhiteSpace(suppressionFile))
             {
                 return new HashSet<Suppression>();
             }
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
@@ -170,7 +170,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Logging
 
             try
             {
-                using Stream reader = GetReadableStream(suppressionFile);
+                using Stream reader = GetReadableStream(suppressionFile!);
                 if (_serializer.Deserialize(reader) is Suppression[] deserializedSuppressions)
                 {
                     return new HashSet<Suppression>(deserializedSuppressions);
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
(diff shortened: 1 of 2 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
```
smasher164, line 1: Is this a whitespace change? What caused this diff?

smasher164 (approved):

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -30,7 +30,10 @@ public class SuppressionEngine : ISuppressionEngine
             string? noWarn = null,
             bool baselineAllErrors = false)
         {
-            _validationSuppressions = ParseSuppressionFile(suppressionFile);
+            // Don't read from the passed in suppression file if errors will be baselined and the file doesn't exist.
+            _validationSuppressions = baselineAllErrors && !File.Exists(suppressionFile) ?
+                new HashSet<Suppression>() :
+                ParseSuppressionFile(suppressionFile);
```
carlossanlop, line 36: Maybe the `File.Exists` check could be added to the `ParseSuppressionFile` method at the beginning, and if there is no file, let it return an empty `HashSet`.

I suggest this because if there are other locations where the `ParseSuppressionFile` method needs to be used in the future, you would have to add similar `File.Exists` checks there.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/SuppressionEngine.cs
```diff
@@ -30,7 +30,10 @@ public class SuppressionEngine : ISuppressionEngine
             string? noWarn = null,
             bool baselineAllErrors = false)
         {
-            _validationSuppressions = ParseSuppressionFile(suppressionFile);
+            // Don't read from the passed in suppression file if errors will be baselined and the file doesn't exist.
+            _validationSuppressions = baselineAllErrors && !File.Exists(suppressionFile) ?
```
joperezr, line 34: >_validationSuppressions [](http://example.com/codeflow?start=12&length=23)

Even if the file does exist, if baselineAllErrors is true technically we could also skip the parsing of the file right?
Or is the problem there that if we do that we would expand global suppressions in that case?

bartonjs: To me, it seems like

_validationSuppressions = new HashSet<Suppression>();

try
{
    ParseSuppressionFile(suppressionFile, _validationSuppressions);
}
catch (FileNotFoundException) when (baselineAllErrors)
{
}
```

is a better fix, since it avoids the trap of using File.Exists.

ViktorHofer: > Even if the file does exist, if baselineAllErrors is true technically we could also skip the parsing of the file right?
Or is the problem there that if we do that we would expand global suppressions in that case?

If the suppression file exists but isn't read, global suppressions in it would be lost. [This logic was already part of APICompat v1](https://github.com/dotnet/sdk/blob/79a92b4eb087b07494034d0a09168dcf323f1efe/src/Compatibility/Microsoft.DotNet.Compatibility/CompatibilityLogger.cs#L24) but was lost during the addition of the CLI frontend.

I noticed this just last week when I passed in a suppression-file option to the CLI frontend which resulted in an uncaught FileNotFoundException as the file didn't exist.

bartonjs (approved):

ViktorHofer:
Added one more nit change: Instead of `string.IsNullOrEmpty("<string>"?.Trim())`, use `string.IsNullOrWhitespace("<string>")`

PR #27759, opened 2022-09-08
Don't use Linq in TypeMustExistTest

ViktorHofer:
The use of the ElementAt Linq extension method isn't necessary anymore as right already is a IReadOnlyList that can be indexed.

### DIFF
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/TypeMustExistTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/TypeMustExistTests.cs
@@ -1,9 +1,7 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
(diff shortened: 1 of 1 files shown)


smasher164 (approved):

carlossanlop (approved):
For my own education: I assume this would help improve perf, and removes one more dependency on Linq? Or is there another reason for this change that I'm missing?

joperezr (approved):

joperezr:
> I assume this would help improve perf, and removes one more dependency on Linq? Or is there another reason for this change that I'm missing?

That is pretty much it 😄. Linq is handy as it is readable and very user-friendly when working with collections and particularly when perf isn't super important. We try not to use it much in the BCL or internal tooling since for most of our cases, perf is very important.

ViktorHofer:
> For my own education: I assume this would help improve perf, and removes one more dependency on Linq? Or is there another reason for this change that I'm missing?

In this case perf isn't really important as its just test code but a general rule that I usually follow is to avoid Linq when possible. Someone could debate if it's even worth opening a PR for this change ;)

PR #27760, opened 2022-09-08
Fix single assembly mode when invoking ApiComparer

ViktorHofer:
Use the correct overload when invoking ApiComparer to compare left and right assemblies, and assembly sets with each other. 

Also fix the underlying ApiCompatRunnerWorkItem type to allow multiple right assembly sets for a given left assembly set and rename the left and right property to follow the naming convention in ApiComparer.
Update a few tests because of the data structure change in the ApiCompatRunnerWorkItem type.

In addition to that, fixing ApiComparer not returning compat differences for assemblies that aren't passed in (i.e. when a type forward for a left reference can't be resolved). Previously the result was joined with the passed in left/right which would exclude anything that isn't part of left/right. I'm now instead doing a manual left outer join.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
@@ -109,7 +109,25 @@ namespace Microsoft.DotNet.ApiCompatibility
             visitor.Visit(mapper);
 
             // Sort the compat differences by the order of the passed in rights.
-            return right.Join(visitor.CompatDifferences, i => i.MetadataInformation, d => d.Right, (i, d) => d).ToArray();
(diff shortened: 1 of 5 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparer.cs
```diff
@@ -126,8 +144,26 @@ public class ApiComparer : IApiComparer
...
+
+            foreach (IGrouping<MetadataInformation, CompatDifference> compatDifferenceGroup in compatDifferenceLookup)
+            {
+                if (processedMetadata.Contains(compatDifferenceGroup.Key))
+                    continue;
+
+                compatDifferences.AddRange(compatDifferenceGroup);
+            }
+
+            return compatDifferences;
```
smasher164, line 166: It seems you are doing the same logic for `right` and `left` to add them to `compatDifferences` and `processedMetadata`. Perhaps extract it into a function?

smasher164 (approved):

PR #27648, opened 2022-09-02
do not emit diagnostic when attribute argument is internal type

smasher164:

PR #27774, opened 2022-09-08
Revert "do not emit diagnostic when attribute argument is internal type"

smasher164:
The diagnostic emitted that this PR suppressed accurately identified a mismatch in the type arguments to TypeConverterAttribute. PR https://github.com/dotnet/runtime/pull/75235 addresses that mismatch, and it's unclear if internal type arguments actually need to be excluded from the attribute rule.

Reverts dotnet/sdk#27648

### DIFF
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
@@ -219,42 +219,6 @@ namespace CompatTests
 new CompatDifference[] {
     CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.CannotChangeAttribute, "", DifferenceType.Changed, "T:CompatTests.First:[T:CompatTests.FooAttribute]")
 }
-            },
-            // Attributes with internal type arguments
-            {
-                @"
-namespace CompatTests
-{
-  using System;
-
-  [AttributeUsage(AttributeTargets.All, AllowMultiple = true)]
-  public class FooAttribute : Attribute {
-    public FooAttribute(Type type) {}
-    public Type A;
-  }
-
-  internal class Bar {}
-
-  [Foo(typeof(Bar), A = typeof(int))]
-  public class First {}
-}
-",
-                @"
-namespace CompatTests
-{
-  using System;
-
-  [AttributeUsage(AttributeTargets.All, AllowMultiple = true)]
-  public class FooAttribute : Attribute {
-    public FooAttribute(Type type) {}
-    public Type A;
-  }
-
-  [Foo(typeof(int), A = typeof(int))]
-  public class First {}
-}
-",
-new CompatDifference[] {}
             }
         };
 
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
@@ -219,6 +219,47 @@ namespace CompatTests
 new CompatDifference[] {
     CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.CannotChangeAttribute, "", DifferenceType.Changed, "T:CompatTests.First:[T:CompatTests.FooAttribute]")
 }
+            },
+            // Attributes on internal type arguments
+            {
+                @"
+namespace CompatTests
+{
+  using System;
+  using CompatTestsSecondNamespace;
+
+  [AttributeUsage(AttributeTargets.All, AllowMultiple = true)]
+  public class FooAttribute : Attribute {
+    public FooAttribute(Type type) {}
+  }
+
+  [Foo(typeof(Bar))]
+  public class First {}
+}
+
+namespace CompatTestsSecondNamespace {
+  internal class Bar {}
+}
+",
+                @"
+namespace CompatTests
+{
+  using System;
+  [AttributeUsage(AttributeTargets.All, AllowMultiple = true)]
+  public class FooAttribute : Attribute {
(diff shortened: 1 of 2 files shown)


ViktorHofer:
> PR https://github.com/dotnet/runtime/pull/75235 addresses that mismatch, and it's unclear if internal type arguments actually need to be excluded from the attribute rule.

Do you know why this was the case in the old ApiCompat?

smasher164:
> > PR [dotnet/runtime#75235](https://github.com/dotnet/runtime/pull/75235) addresses that mismatch, and it's unclear if internal type arguments actually need to be excluded from the attribute rule.
> 
> Do you know why this was the case in the old ApiCompat?

After speaking with @ericstj, it was apparently due to the fact that the `PublicOnlyCciFilter` didn't even visit attribute arguments invisible outside the assembly. See https://github.com/dotnet/arcade/blob/e7ede87875f41a9b3df898ae08da5ebc96e24f56/src/Microsoft.Cci.Extensions/Filters/PublicOnlyCciFilter.cs#L72. I'm not sure why that was, but it has something to do with the fact that the Cci filters were used in code generation?

ericstj:
> I'm not sure why that was, but it has something to do with the fact that the Cci filters were used in code generation?

That’s my best guess.  Otherwise GenAPI would emit code that didn’t compile.  There wouldn’t be any override to tell GenAPI to exclude that one applied attribute - or include the internal type. 

For API Compat I think it makes more sense to include here if the user is telling us that they think the Attribute should be tested.  They have opted into treating uses of that attribute on public API as public.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
```diff
@@ -219,42 +219,6 @@ public class First {}
 new CompatDifference[] {
     CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.CannotChangeAttribute, "", DifferenceType.Changed, "T:CompatTests.First:[T:CompatTests.FooAttribute]")
 }
```
ericstj, line 221: Since we desire this behavior can we add an explicit test for it?  That should avoid having this discussion again in the future :)

smasher164: To clarify, do you mean adding a test that asserts that changes to internal type arguments are caught?

ericstj: Yeah, that might catch if we had an inconsistent namespace between the internal type argument on left and right.

ViktorHofer: Shouldn't we only validate internal members when this flag is set to true? https://github.com/dotnet/sdk/blob/412df6f0ccee8f0f309d88b674b9c8d6d88c58ea/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/ApiComparerSettings.cs#L23-L26

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
```diff
@@ -220,41 +220,46 @@ public class First {}
...
-  [Foo(typeof(int), A = typeof(int))]
+  internal class Bar {}
+
+  [Foo(typeof(Bar))]
   public class First {}
 }
 ",
-new CompatDifference[] {}
+new CompatDifference[] {
+    CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.CannotChangeAttribute, "", DifferenceType.Changed, "T:CompatTests.First:[T:CompatTests.FooAttribute]"),
```
ericstj, line 261: Thank you for adding the test - you may want to also include some comments that explain why it's important to examine internal types.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
smasher164: > The concept of internal/public on attributes is not well defined

Based on the previous behavior, right now, `IncludeInternalSymbols` applies to the attribute class as well. If we didn't want to do that, then I can remove that behavior as well.

ViktorHofer:
@smasher164 do you want to get this into release/7.0?

smasher164:
> @smasher164 do you want to get this into release/7.0?

Yes. I believe we decided that internal type arguments to a public attribute shouldn't be filtered out by `typ.IsVisibleOutsideOfAssembly`.

ViktorHofer:
Cool. I was asking because we are past RC2 now and still getting changes into .NET 7 becomes harder with every day. Do you know what is missing to get this one in? Do we still need to reach consensus?

ViktorHofer:
@smasher164 gentle ping regarding status. I just yesterday backported an APICompat change into release/7.0.1xx and that required Tactics approval (even for a non binary impacting change).

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
ViktorHofer: [We still filter out attributes which aren't visible outside of the assembly](https://github.com/dotnet/sdk/blob/65c6944d81aff5bdb7c9cf14637b8657a0c0a44d/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs#L60-L63), even with this PR; but we don't filter out visible attributes which contain internal members anymore. That's the desired behavior, correct?

ViktorHofer (approved):

PR #27776, opened 2022-09-08
Produce a warning and ignore Prefer32Bit for netcoreapp 

VSadov:
Fixes:https://github.com/dotnet/runtime/issues/70776

`Prefer32Bit` is a way to control how OS loader loads a managed `.exe`. It does not make sense for a `.dll` and the loader would refuse to load such `.dll`. 
For netcoreapp target `Prefer32Bit` results in an app that cannot run. 

With this change we produce a warning when building for netcoreapp >= 7.0 target and ignore the setting.


vitek-karas (approved):

VSadov:
Thanks!!

PR #27798, opened 2022-09-09
Mysterious PR you should ignore for now

nagilson:
Add publish-self-contained back


nagilson:
Ok cool, so for the sake of history this run shows https://github.com/dotnet/sdk/runs/8381377847 that no tests are broken by making `SelfContained` `treataslocalproperty`. This build is adjacent to everything I'm doing here but runs without a conditional `PublishedSelfContained` check, to confirm it didn't break any test.

PR #27822, opened 2022-09-12
[APICompat] Don't filter package assets and handle placeholder files

### LINKED ISSUE #18165: PackageValidation task should not filter package assets.
cc: @Anipik @safern @ericstj 

While testing package validation I noticed that apparently we are filtering the assets that we consider when parsing a package here:

https://github.com/dotnet/sdk/blob/2347c7209d6974b0622b52bd6968bb6aae00fc07/src/Compatibility/Microsoft.DotNet.PackageValidation/NupkgParser.cs#L45

The problem with this, is that we will only consider those filtered files when calculating the best compile and runtime asset to verify APICompat for, but that is not always accurate, as a package could have an asset like `lib/net461/_._` which should win as a runtime asset over a `lib/netstandard2.0/Foo.dll` when evaluating for net461, and that is not happening today. This is important as the previously mentioned example, means that effectively I would be intentionally dropping support for net461, but package validation won't catch it as it would evaluate APICompat against the netstandard2.0 asset.

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/18165

Instead of manually filtering package assets, let NuGet calculate the runtime and compile time assets and handle placeholder files which are returned by NuGet.

TODO:
- [ ] Add tests for packages with placeholder files.

### DIFF
--- a/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
+++ b/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using Microsoft.DotNet.ApiCompatibility;
 using Microsoft.DotNet.ApiCompatibility.Logging;
 using Microsoft.DotNet.ApiCompatibility.Runner;
+using Microsoft.DotNet.ApiSymbolExtensions.Logging;
 using NuGet.ContentModel;
 using NuGet.Frameworks;
 
@@ -30,6 +31,19 @@ namespace Microsoft.DotNet.PackageValidation
             {
                 return;
             }
+
+            // Don't perform api compatibility checks on placeholder files.
+            if (leftContentItems.IsPlaceholderFile() && rightContentItems.IsPlaceholderFile())
+            {
+                leftContentItems[0].Properties.TryGetValue("tfm", out object? tfmRaw);
+                string tfm = tfmRaw?.ToString() ?? string.Empty;
+
+                log.LogMessage(MessageImportance.Normal, string.Format(Resources.SkipApiCompatForPlaceholderFiles, tfm));
+            }
+
+            // Make sure placeholder package files aren't enqueued as api comparison check work items.
+            Debug.Assert(!leftContentItems.IsPlaceholderFile() && !rightContentItems.IsPlaceholderFile(),
+                "Placeholder files must not be enqueued for api comparison checks.");
 
             // If a right hand side package is provided in addition to the left side, package validation runs in baseline comparison mode.
             // The assumption stands that the right hand side is "current" and has more information than the left, i.e. assembly references.
--- a/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Package.cs
+++ b/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Package.cs
@@ -6,7 +6,6 @@ using NuGet.Client;
 using NuGet.ContentModel;
 using NuGet.Frameworks;
 using NuGet.Packaging;
-using NuGet.Packaging.Core;
 using NuGet.RuntimeModel;
 
 namespace Microsoft.DotNet.PackageValidation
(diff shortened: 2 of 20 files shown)


marcpopMSFT:

ViktorHofer:
Yes, let me close this an get back to it when I have more time for APICompat.

src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
```diff
@@ -31,6 +32,19 @@ internal static class ApiCompatRunnerExtensions
...
+            if (leftContentItems.IsPlaceholderFile() && rightContentItems.IsPlaceholderFile())
+            {
+                leftContentItems[0].Properties.TryGetValue("tfm", out object? tfmRaw);
+                string tfm = tfmRaw?.ToString() ?? string.Empty;
+
+                log.LogMessage(MessageImportance.Normal, string.Format(Resources.SkipApiCompatForPlaceholderFiles, tfm));
+            }
+
+            // Make sure placeholder package files aren't enqueued as api comparison check work items.
+            Debug.Assert(!leftContentItems.IsPlaceholderFile() && !rightContentItems.IsPlaceholderFile(),
```
ericstj, line 45: I don't think this is an accurate assert.  What if a package explicitly adds a placeholder in a compatible framework?

It might do this if a) the library became part of that framework and we could compare the package library to the framework library (in references) or b) as an intentional breaking change in which case we should still compare and raise the missing file.

src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
```diff
@@ -31,6 +32,19 @@ internal static class ApiCompatRunnerExtensions
                 return;
             }
 
+            // Don't perform api compatibility checks on placeholder files.
+            if (leftContentItems.IsPlaceholderFile() && rightContentItems.IsPlaceholderFile())
```
ericstj, line 36: Maybe the condition here should be if left is a placeholder and we're not in strict mode?

When in strict mode, then if left is placeholder right must also be placeholder.

Perhaps we cover these cases by just fleshing out the comparisons with tests involving placeholders.

src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Validators/BaselinePackageValidator.cs
```diff
@@ -43,13 +43,21 @@ public void Validate(PackageValidatorOption options)
...
+                        (latestCompileAssets.IsPlaceholderFile() && !baselineCompileAssets.IsPlaceholderFile()))
                     {
                         log.LogError(new Suppression(DiagnosticIds.TargetFrameworkDropped) { Target = baselineTargetFramework.ToString() },
                             DiagnosticIds.TargetFrameworkDropped,
                             string.Format(Resources.MissingTargetFramework,
                                 baselineTargetFramework));
                     }
+                    else if (baselineCompileAssets.IsPlaceholderFile() && !latestCompileAssets.IsPlaceholderFile())
+                    {
+                        // Ignore the newly added compile time asset in the latest package.
```
ericstj, line 59: In strict mode I think we would want to flag this.

src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Package.cs
```diff
@@ -136,7 +135,7 @@ public static Package Create(string? packagePath, IReadOnlyDictionary<NuGetFrame
             NuspecReader nuspecReader = packageReader.NuspecReader;
             string packageId = nuspecReader.GetId();
             string version = nuspecReader.GetVersion().ToString();
-            IEnumerable<string> packageAssets = packageReader.GetFiles().Where(t => t.EndsWith(".dll")).ToArray();
+            IEnumerable<string> packageAssets = packageReader.GetFiles().ToArray();
```
ericstj, line 138: Make sure you add a couple tests to ensure we can handle the other file types that NuGet will give us.
https://github.com/NuGet/NuGet.Client/blob/46f80a3f1a83f3c5f33a20fa9fb72871e1e2d4eb/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs#L28

I wonder if we should ignore exe, or allow folks to specify something to ignore exe's?  I guess they could add a suppression to ignore all things from the comparison of an exe.

PR #27824, opened 2022-09-12
Trusted roots:  August 2022 CTL

dtivel:
This is the August 2022 CTL update from the Microsoft Trusted Root Program team.

Previously, codesignctl.pem contained roots which were valid for both code signing (CS) and timestamping (TS), which means that roots valid for one but not the other were not included.

With this PR, there are separate CS and TS certificate bundles, which contain all roots valid for that purpose.

CC @richlander, @aortiz-msft


aortiz-msft (approved):
LGTM but we may need eyes from @marcpopMSFT or someone in his team.

marcpopMSFT:
@aortiz-msft I mean, I was hoping your team would own these PRs end to end so my team didn't have to review/merge them. Other teams that contribute to the sdk repo tend to manage their own code as it's gotten too big for my team to track it all (there are 10+ teams making changes). 

The PR looks fine to me other than what branch it should go into as this is going into 8.0.1xx.

PR #27866, opened 2022-09-13
Try to make all the the trimmer tests happy

lewing:
Untested but I think the warning checks ignore missing ones so this might be enough to make the test happy against runtime main and runtime release/7.0


sbomer (approved):
Thank you, LGTM!

PR #27942, opened 2022-09-15
Expand existing warning to target 5.0

marcpopMSFT:
## Description

The 5.0 runtime went out of support in May. At the time, we discussed when to add the warning message to customer builds when targeting 5.0.  We decided it made sense to align that with the next major release so we're planning on doing this for the November releases: https://github.com/dotnet/sdk/pull/27943

## Customer Impact
Customer will get a build warning for any project that targets 5.0.

## Regression

No

## Risk 
Low. Customers can add CheckEolTargetFramework=false to avoid this warning.

## Testing 
Automated

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.EolTargetFrameworks.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.EolTargetFrameworks.targets
@@ -20,7 +20,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     receive servicing updates and security fixes.
   -->
   <ItemGroup>
-    <_EolNetCoreTargetFrameworkVersions Include="1.0;1.1;2.0;2.1;2.2;3.0" />
+    <_EolNetCoreTargetFrameworkVersions Include="1.0;1.1;2.0;2.1;2.2;3.0;5.0" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.EolTargetFrameworks.targets
```diff
@@ -20,7 +20,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     receive servicing updates and security fixes.
   -->
   <ItemGroup>
-    <_EolNetCoreTargetFrameworkVersions Include="1.0;1.1;2.0;2.1;2.2;3.0" />
+    <_EolNetCoreTargetFrameworkVersions Include="1.0;1.1;2.0;2.1;2.2;3.0;5.0" />
```
kasperk81, line 23: 3.1's life will end in [less than 90 days](https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core). should we add it here?

marcpopMSFT: Customers consider new warnings in existing products to be highly impactful.  We included a 2.1 warning the month after it went out of support and got a lot of feedback. After discussion internally when 5.0 went out of support, we decided to delay turning on this warning for 6 months after 5.0 went out of support to align with our 7.0 release to give customers more time before we rolled out the warning. I expect 3.1 will also be delayed by a few months before we add that warning.

baronfel (approved):

PR #27961, opened 2022-09-16
Handle test parser forwarded args

### LINKED ISSUE #4014: dotnet test does not forward MSBuild properties to msbuild in .NET 7 RC1
## Description

Starting with .NET SDK `7.0.100-rc.1.22431.12` MSBuild parameters specified on the command line to `dotnet test` are no longer forwarded to MSBuild.

The issue was not present in .NET 7 preview 7.

## Steps to reproduce

Clone the [martincostello/api](https://github.com/martincostello/api) repository at commit [`61ec79c1c1e773b721f6ebdf28ed189d78d63ae7`](https://github.com/martincostello/api/commit/61ec79c1c1e773b721f6ebdf28ed189d78d63ae7) and then run `dotnet test` with the arguments specified below.

```sh
git clone https://github.com/martincostello/api
cd api
git checkout 61ec79c1c1e773b721f6ebdf28ed189d78d63ae7
dotnet test ./tests/API.Tests --configuration Release --filter Category=EndToEnd /p:CollectCoverage=false
```

## Expected behavior

`dotnet test` exits with an exit code of `0`.

## Actual behavior

`dotnet test` returns a non-zero exit code due to the `CollectCoverage` property not being set to `false`, causing the minimum coverage requirement of coverlet to not be satisfied.

```sh
+--------+------+--------+--------+
| Module | Line | Branch | Method |
+--------+------+--------+--------+
| API    | 0%   | 0%     | 0%     |
+--------+------+--------+--------+

+---------+------+--------+--------+
|         | Line | Branch | Method |
+---------+------+--------+--------+
| Total   | 0%   | 0%     | 0%     |
+---------+------+--------+--------+
| Average | 0%   | 0%     | 0%     |
+---------+------+--------+--------+

C:\Users\martin.costello\.nuget\packages\coverlet.msbuild\3.1.2\build\coverlet.msbuild.targets(71,5): error : The minimum line coverage is below the specified 55 [C:\Coding\martincostello\api\tests\API.Tests\API.Tests.csproj] C:\Users\martin.costello\.nuget\packages\coverlet.msbuild\3.1.2\build\coverlet.msbuild.targets(71,5): error : The minimum branch coverage is below the specified 55 [C:\Coding\martincostello\api\tests\API.Tests\API.Tests.csproj]
C:\Users\martin.costello\.nuget\packages\coverlet.msbuild\3.1.2\build\coverlet.msbuild.targets(71,5): error : The minimum method coverage is below the specified 55 [C:\Coding\martincostello\api\tests\API.Tests\API.Tests.csproj]
C:\Users\martin.costello\.nuget\packages\coverlet.msbuild\3.1.2\build\coverlet.msbuild.targets(71,5): error :  [C:\Coding\martincostello\api\tests\API.Tests\API.Tests.csproj]
C:\Users\martin.costello\.nuget\packages\coverlet.msbuild\3.1.2\build\coverlet.msbuild.targets(71,5): error :    at Coverlet.MSbuild.Tasks.CoverageResultTask.Execute() in /_/src/coverlet.msbuild.tasks/CoverageResultTask.cs:line 247 [C:\Coding\martincostello\api\tests\API.Tests\API.Tests.csproj]
```

## Environment

```sh
❯ dotnet --info
.NET SDK:
 Version:   7.0.100-rc.1.22431.12
 Commit:    f1cf61e1c0

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.100-rc.1.22431.12\

Host:
  Version:      7.0.0-rc.1.22426.10
  Architecture: x64
  Commit:       06aceb7015

.NET SDKs installed:
  2.1.818 [C:\Program Files\dotnet\sdk]
  3.1.120 [C:\Program Files\dotnet\sdk]
  3.1.201 [C:\Program Files\dotnet\sdk]
  3.1.302 [C:\Program Files\dotnet\sdk]
  3.1.303 [C:\Program Files\dotnet\sdk]
  3.1.403 [C:\Program Files\dotnet\sdk]
  3.1.423 [C:\Program Files\dotnet\sdk]
  5.0.103 [C:\Program Files\dotnet\sdk]
  5.0.104 [C:\Program Files\dotnet\sdk]
  5.0.203 [C:\Program Files\dotnet\sdk]
  5.0.214 [C:\Program Files\dotnet\sdk]
  5.0.303 [C:\Program Files\dotnet\sdk]
  5.0.408 [C:\Program Files\dotnet\sdk]
  5.0.411 [C:\Program Files\dotnet\sdk]
  6.0.100-rc.1.21458.32 [C:\Program Files\dotnet\sdk]
  6.0.102 [C:\Program Files\dotnet\sdk]
  6.0.109 [C:\Program Files\dotnet\sdk]
  6.0.202 [C:\Program Files\dotnet\sdk]
  6.0.203 [C:\Program Files\dotnet\sdk]
  6.0.304 [C:\Program Files\dotnet\sdk]
  6.0.400 [C:\Program Files\dotnet\sdk]
  6.0.401 [C:\Program Files\dotnet\sdk]
  7.0.100-rc.1.22431.12 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.21 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.25 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.21 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.25 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.11 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.20 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.28 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.0-preview.8.20414.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.0-preview.7.22376.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.0-rc.1.22427.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.21 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.25 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.11 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.20 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.28 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.0-preview.8.20407.11 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.0-preview.5.21301.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.0-rc.1.21451.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.0-rc.2.21480.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.8 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0-preview.7.22375.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0-rc.1.22426.10 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.11 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.20 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.28 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.0-preview.8.20411.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.3 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.0-preview.5.21301.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.0-rc.1.21451.3 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.0-rc.2.21501.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.8 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.0-preview.7.22377.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.0-rc.1.22427.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  arm64 [C:\Program Files\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\arm64\InstallLocation]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  Not set

global.json file:
  C:\Coding\martincostello\api\global.json

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
```

[AB#1663078](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1663078)

baronfel:
## Description

A tech debt PR for the CLI's command line parsing resulted in a regression where unknown tokens were no longer forwarded along in the `dotnet test` command. This is behavior we want, since commands that forward to MSBuild shouldn't have to embed the entire MSBuild CLI syntax in their own parsers. As a result of this regression, users were unable to pass MSBuild arguments directly, instead having to use workarounds like setting environment variables instead.  This was reported in https://github.com/microsoft/vstest/issues/4014.

## Regression

**Yes**, this worked in preview 7.

## Risk

**Low**, this uses the same mechanism we have in other locations in the codebase to forward the tokens along

## Testing

- [x] manual testing
- [x] automated tests to cover msbuild argument forwarding for this command

## Details

Closes https://github.com/microsoft/vstest/issues/4014

With the switch away from the legacy `--` handling, users of System.CommandLine have two options for handling unrecognized arguments:
* adding an explicit argument to capture them and doing something explicit with them, or
* using the `ParseResult.UnmatchedTokens` property to get unmatched tokens (tokens that the parser doesn't explicitly have handling for) and do something explicit with _them_

In this PR I've taken the ~former~ latter approach because the former approach results in a help output that is too noisy.

### DIFF
--- a/src/Assets/TestProjects/VSTestMSBuildParameters/Tests.cs
+++ b/src/Assets/TestProjects/VSTestMSBuildParameters/Tests.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections;
+using System.Reflection;
+
+namespace TestNamespace
+{
+    [TestClass]
+    public class Tests
+    {
+        [TestMethod]
+        public void TestMSBuildParameters()
+        {
+            var assemblyInfoVersion = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+            Assert.AreEqual(assemblyInfoVersion, "1.2.3");
+        }
+    }
+}
--- a/src/Assets/TestProjects/VSTestMSBuildParameters/Tests.cs
+++ b/src/Assets/TestProjects/VSTestMSBuildParameters/Tests.cs
@@ -15,7 +15,7 @@ namespace TestNamespace
         public void TestMSBuildParameters()
         {
             var assemblyInfoVersion = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
-            Assert.AreEqual(assemblyInfoVersion, "1.2.3");
+            Assert.AreEqual("1.2.3", assemblyInfoVersion);
         }
     }
 }
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -33,15 +33,6 @@ namespace Microsoft.DotNet.Tools.Test
             // from the VSTest side.
             string testSessionCorrelationId = $"{Environment.ProcessId}_{Guid.NewGuid()}";
 
-            if (!Debugger.IsAttached)
-            {
-                Debugger.Launch();
-            }
-            else
-            {
-                Debugger.Break();
-            }
-
             string[] args = parseResult.GetArguments();
 
             if (VSTestTrace.TraceEnabled)
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -121,7 +121,8 @@ namespace Microsoft.DotNet.Tools.Test
 
             // Extra msbuild properties won't be parsed and so end up in the UnmatchedTokens list. In addition to those
             // properties, all the test settings properties are also considered as unmatched but we don't want to forward
-            // these as-is to msbuild. So we filter out the test settings properties from the unmatched tokens.
+            // these as-is to msbuild. So we filter out the test settings properties from the unmatched tokens,
+            // by only taking values until the first item after `--`. (`--` is not present in the UnmatchedTokens ).
             var unMatchedNonSettingsArgs = settings.Length > 1
                 ? result.UnmatchedTokens.TakeWhile(x => x != settings[1])
                 : result.UnmatchedTokens;
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -120,7 +120,7 @@ namespace Microsoft.DotNet.Tools.Test
 
 
             var parsedArgs =
-                    result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
+                result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
                     .Concat(result.UnmatchedTokens); // all tokens that the test-parser doesn't explicitly track
 
             VSTestTrace.SafeWriteTrace(() => $"MSBuild args from forwarded options: {String.Join(", ", parsedArgs)}" );
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -119,9 +119,13 @@ namespace Microsoft.DotNet.Tools.Test
                 "-nologo"
             };
 
+            var unMatchedNonSettingsArgs = settings.Length > 1
+                ? result.UnmatchedTokens.TakeWhile(x => x != settings[1])
+                : result.UnmatchedTokens;
+
             var parsedArgs =
                 result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
-                    .Concat(args); // all tokens that the test-parser doesn't explicitly track (minus the settings tokens)
+                    .Concat(unMatchedNonSettingsArgs); // all tokens that the test-parser doesn't explicitly track (minus the settings tokens)
 
             VSTestTrace.SafeWriteTrace(() => $"MSBuild args from forwarded options: {String.Join(", ", parsedArgs)}" );
             msbuildArgs.AddRange(parsedArgs);
@@ -129,14 +133,14 @@ namespace Microsoft.DotNet.Tools.Test
             if (settings.Any())
             {
                 //workaround for correct -- logic
-                var commandArgument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
-                // TODO: @baronfel, @Evangelink
-                // If the project or solution is not specified we would expect null here but we actually
-                // get some of the parameter (last one before --) instead. 
-                if (!string.IsNullOrWhiteSpace(commandArgument) && !settings.Contains(commandArgument))
-                {
-                    msbuildArgs.Add(commandArgument);
-                }
+                //var commandArgument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
+                //// TODO: @baronfel, @Evangelink
+                //// If the project or solution is not specified we would expect null here but we actually
+                //// get some of the parameter (last one before --) instead. 
+                //if (!string.IsNullOrWhiteSpace(commandArgument) && !settings.Contains(commandArgument))
+                //{
+                //    msbuildArgs.Add(commandArgument);
+                //}
 
                 // skip '--' and escape every \ to be \\ and every " to be \" to survive the next hop
                 string[] escaped = settings.Skip(1).Select(s => s.Replace("\\", "\\\\").Replace("\"", "\\\"")).ToArray();
@@ -146,11 +150,11 @@ namespace Microsoft.DotNet.Tools.Test
             }
             else
             {
-                var argument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
-                if (!string.IsNullOrWhiteSpace(argument))
-                {
-                    msbuildArgs.Add(argument);
-                }
+                //var argument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
+                //if (!string.IsNullOrWhiteSpace(argument))
+                //{
+                //    msbuildArgs.Add(argument);
+                //}
             }
 
             string verbosityArg = result.ForwardedOptionValues<IReadOnlyCollection<string>>(TestCommandParser.GetCommand(), "verbosity")?.SingleOrDefault() ?? null;
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -32,7 +32,7 @@ namespace Microsoft.DotNet.Tools.Test
             // from the VSTest side.
             string testSessionCorrelationId = $"{Environment.ProcessId}_{Guid.NewGuid()}";
 
-            string[] args = parseResult.GetArguments();
+            string[] args = parseResult.GetArguments().Concat(parseResult.UnmatchedTokens).ToArray();
 
             if (VSTestTrace.TraceEnabled)
             {
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -32,7 +32,7 @@ namespace Microsoft.DotNet.Tools.Test
             // from the VSTest side.
             string testSessionCorrelationId = $"{Environment.ProcessId}_{Guid.NewGuid()}";
 
-            string[] args = parseResult.GetArguments().Concat(parseResult.UnmatchedTokens).ToArray();
+            string[] args = parseResult.GetArguments();
 
             if (VSTestTrace.TraceEnabled)
             {
@@ -118,7 +118,13 @@ namespace Microsoft.DotNet.Tools.Test
                 "-nologo"
             };
 
-            msbuildArgs.AddRange(result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()));
+
+            var parsedArgs =
+                    result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
+                    .Concat(result.UnmatchedTokens); // all tokens that the test-parser doesn't explicitly track
+
+            VSTestTrace.SafeWriteTrace(() => $"MSBuild args from forwarded options: {String.Join(", ", parsedArgs)}" );
+            msbuildArgs.AddRange(parsedArgs);
 
             if (settings.Any())
             {
@@ -126,7 +132,7 @@ namespace Microsoft.DotNet.Tools.Test
                 var commandArgument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
                 if(!string.IsNullOrWhiteSpace(commandArgument) && !settings.Contains(commandArgument))
                 {
-                    msbuildArgs.Add(result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument));
+                    msbuildArgs.Add(commandArgument);
                 }
 
                 // skip '--' and escape every \ to be \\ and every " to be \" to survive the next hop
@@ -138,8 +144,10 @@ namespace Microsoft.DotNet.Tools.Test
             else
             {
                 var argument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
-                if(!string.IsNullOrWhiteSpace(argument))
+                if (!string.IsNullOrWhiteSpace(argument))
+                {
                     msbuildArgs.Add(argument);
+                }
             }
 
             string verbosityArg = result.ForwardedOptionValues<IReadOnlyCollection<string>>(TestCommandParser.GetCommand(), "verbosity")?.SingleOrDefault() ?? null;
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -68,7 +68,7 @@ namespace Microsoft.DotNet.Tools.Test
             try
             {
                 Environment.SetEnvironmentVariable(NodeWindowEnvironmentName, "1");
-                int exitCode = FromParseResult(parseResult, args, settings, testSessionCorrelationId).Execute();
+                int exitCode = FromParseResult(parseResult, settings, testSessionCorrelationId).Execute();
 
                 // We run post processing also if execution is failed for possible partial successful result to post process.
                 exitCode |= RunArtifactPostProcessingIfNeeded(testSessionCorrelationId, parseResult, FeatureFlag.Instance);
@@ -108,7 +108,7 @@ namespace Microsoft.DotNet.Tools.Test
             return exitCode;
         }
 
-        private static TestCommand FromParseResult(ParseResult result, string[] args, string[] settings, string testSessionCorrelationId, string msbuildPath = null)
+        private static TestCommand FromParseResult(ParseResult result, string[] settings, string testSessionCorrelationId, string msbuildPath = null)
         {
             result.ShowHelpOrErrorIfAppropriate();
 
@@ -119,6 +119,9 @@ namespace Microsoft.DotNet.Tools.Test
                 "-nologo"
             };
 
+            // Extra msbuild properties won't be parsed and so end up in the UnmatchedTokens list. In addition to those
+            // properties, all the test settings properties are also considered as unmatched but we don't want to forward
+            // these as-is to msbuild. So we filter out the test settings properties from the unmatched tokens.
             var unMatchedNonSettingsArgs = settings.Length > 1
                 ? result.UnmatchedTokens.TakeWhile(x => x != settings[1])
                 : result.UnmatchedTokens;
@@ -132,29 +135,11 @@ namespace Microsoft.DotNet.Tools.Test
 
             if (settings.Any())
             {
-                //workaround for correct -- logic
-                //var commandArgument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
-                //// TODO: @baronfel, @Evangelink
-                //// If the project or solution is not specified we would expect null here but we actually
-                //// get some of the parameter (last one before --) instead. 
-                //if (!string.IsNullOrWhiteSpace(commandArgument) && !settings.Contains(commandArgument))
-                //{
-                //    msbuildArgs.Add(commandArgument);
-                //}
-
                 // skip '--' and escape every \ to be \\ and every " to be \" to survive the next hop
                 string[] escaped = settings.Skip(1).Select(s => s.Replace("\\", "\\\\").Replace("\"", "\\\"")).ToArray();
 
                 string runSettingsArg = string.Join(";", escaped);
                 msbuildArgs.Add($"-property:VSTestCLIRunSettings=\"{runSettingsArg}\"");
-            }
-            else
-            {
-                //var argument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
-                //if (!string.IsNullOrWhiteSpace(argument))
-                //{
-                //    msbuildArgs.Add(argument);
-                //}
             }
 
             string verbosityArg = result.ForwardedOptionValues<IReadOnlyCollection<string>>(TestCommandParser.GetCommand(), "verbosity")?.SingleOrDefault() ?? null;
--- a/src/Cli/dotnet/commands/dotnet-test/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-test/Program.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Parsing;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using Microsoft.DotNet.Cli;
@@ -32,6 +33,15 @@ namespace Microsoft.DotNet.Tools.Test
             // from the VSTest side.
             string testSessionCorrelationId = $"{Environment.ProcessId}_{Guid.NewGuid()}";
 
+            if (!Debugger.IsAttached)
+            {
+                Debugger.Launch();
+            }
+            else
+            {
+                Debugger.Break();
+            }
+
             string[] args = parseResult.GetArguments();
 
             if (VSTestTrace.TraceEnabled)
@@ -56,10 +66,10 @@ namespace Microsoft.DotNet.Tools.Test
                 return ForwardToVSTestConsole(parseResult, args, settings, testSessionCorrelationId);
             }
 
-            return ForwardToMsbuild(parseResult, settings, testSessionCorrelationId);
+            return ForwardToMsbuild(parseResult, args, settings, testSessionCorrelationId);
         }
 
-        private static int ForwardToMsbuild(ParseResult parseResult, string[] settings, string testSessionCorrelationId)
+        private static int ForwardToMsbuild(ParseResult parseResult, string[] args, string[] settings, string testSessionCorrelationId)
         {
             // Workaround for https://github.com/Microsoft/vstest/issues/1503
             const string NodeWindowEnvironmentName = "MSBUILDENSURESTDOUTFORTASKPROCESSES";
@@ -67,7 +77,7 @@ namespace Microsoft.DotNet.Tools.Test
             try
             {
                 Environment.SetEnvironmentVariable(NodeWindowEnvironmentName, "1");
-                int exitCode = FromParseResult(parseResult, settings, testSessionCorrelationId).Execute();
+                int exitCode = FromParseResult(parseResult, args, settings, testSessionCorrelationId).Execute();
 
                 // We run post processing also if execution is failed for possible partial successful result to post process.
                 exitCode |= RunArtifactPostProcessingIfNeeded(testSessionCorrelationId, parseResult, FeatureFlag.Instance);
@@ -107,7 +117,7 @@ namespace Microsoft.DotNet.Tools.Test
             return exitCode;
         }
 
-        private static TestCommand FromParseResult(ParseResult result, string[] settings, string testSessionCorrelationId, string msbuildPath = null)
+        private static TestCommand FromParseResult(ParseResult result, string[] args, string[] settings, string testSessionCorrelationId, string msbuildPath = null)
         {
             result.ShowHelpOrErrorIfAppropriate();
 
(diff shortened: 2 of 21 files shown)


baronfel:
The new test output has some duplication that I can't track down the source of:


```
Description:
  .NET Test Driver                                                                                                                                                                                                                                                                                                                                                                                                                Usage:
  dotnet test [<PROJECT | SOLUTION> [<FORWARDED_ARGS>...]] [options] [[--] <additional arguments>...]]

Arguments:
  <PROJECT | SOLUTION>  The project or solution file to operate on. If a file is not specified, the command will search the current directory for one.
  <FORWARDED_ARGS>      Arguments to forward to MSBuild.

Options:
  -s, --settings <SETTINGS_FILE>                             The settings file to use when running tests.
  -t, --list-tests                                           List the discovered tests instead of running the tests.
  -e, --environment <NAME="VALUE">                           Sets the value of an environment variable.
                                                             Creates the variable if it does not exist, overrides if it does.
                                                             This will force the tests to be run in an isolated process.
                                                             This argument can be specified multiple times to provide multiple variables.

                                                             Examples:
                                                             -e VARIABLE=abc
                                                             -e VARIABLE="value with spaces"
                                                             -e VARIABLE="value;seperated with;semicolons"
                                                             -e VAR1=abc -e VAR2=def -e VAR3=ghi
  --filter <EXPRESSION>                                      Run tests that match the given expression.
                                                                                                     Examples:
                                                                                                     Run tests with priority set to 1: --filter "Priority = 1"
                                                                                                     Run a test with the specified full name: --filter "FullyQualifiedName=Namespace.ClassName.MethodName"
                                                                                                     Run tests that contain the specified name: --filter "FullyQualifiedName~Namespace.Class"
                                                                                                     See https://aka.ms/vstest-filtering for more information on filtering support.
  --test-adapter-path <ADAPTER_PATH>                         The path to the custom adapters to use for the test run.
  -l, --logger <LOGGER>                                      The logger to use for test results.
                                                                                                     Examples:
                                                                                                     Log in trx format using a unique file name: --logger trx
                                                                                                     Log in trx format using the specified file name: --logger "trx;LogFileName=<TestResults.trx>"
                                                                                                     See https://aka.ms/vstest-report for more information on logger arguments.
  -o, --output <OUTPUT_DIR>                                  The output directory to place built artifacts in.
  -d, --diag <LOG_FILE>                                      Enable verbose logging to the specified file.
  --no-build                                                 Do not build the project before testing. Implies --no-restore.
  --results-directory <RESULTS_DIR>                          The directory where the test results will be placed.
                                                             The specified directory will be created if it does not exist.
  --collect <DATA_COLLECTOR_NAME>                            The friendly name of the data collector to use for the test run.
                                                                                                     More info here: https://aka.ms/vstest-collect
  --blame                                                    Runs the tests in blame mode. This option is helpful in isolating problematic tests that cause the test host to crash or hang.
                                                             When a crash is detected, it creates an sequence file in TestResults/guid/guid_Sequence.xml that captures the order of tests that were run before
                                                             the crash.
                                                             Based on the additional settings, hang dump or crash dump can also be collected.
                                                             Example:
                                                               Timeout the test run when test takes more than the default timeout of 1 hour, and collect crash dump when the test host exits unexpectedly.
                                                               (Crash dumps require additional setup, see below.)
                                                               dotnet test --blame-hang --blame-crash
                                                             Example:
                                                               Timeout the test run when a test takes more than 20 minutes and collect hang dump.
                                                               dotnet test --blame-hang-timeout 20min
  --blame-crash                                              Runs the tests in blame mode and enables collecting crash dump when testhost exits unexpectedly.
                                                             This option is currently only supported on Windows, and requires procdump.exe and procdump64.exe to be available in PATH.
                                                             Or PROCDUMP_PATH environment variable to be set, and point to a directory that contains procdump.exe and procdump64.exe.
                                                             The tools can be downloaded here: https://docs.microsoft.com/sysinternals/downloads/procdump
                                                             Implies --blame.
  --blame-crash-dump-type <blame-crash-dump-type>            The type of crash dump to be collected. Implies --blame-crash.
  --blame-crash-collect-always <blame-crash-collect-always>  Enables collecting crash dump on expected as well as unexpected testhost exit.
  --blame-hang                                               Run the tests in blame mode and enables collecting hang dump when test exceeds the given timeout. Implies --blame-hang.
  --blame-hang-dump-type <blame-hang-dump-type>              The type of crash dump to be collected. When None, is used then test host is terminated on timeout, but no dump is collected. Implies --blame-hang.
  --blame-hang-timeout <TIMESPAN>                            Per-test timeout, after which hang dump is triggered and the testhost process is terminated.
                                                             The timeout value is specified in the following format: 1.5h / 90m / 5400s / 5400000ms. When no unit is used (e.g. 5400000), the value is assumed
                                                             to be in milliseconds.
                                                             When used together with data driven tests, the timeout behavior depends on the test adapter used. For xUnit and NUnit the timeout is renewed after
                                                             every test case,
                                                             For MSTest, the timeout is used for all testcases.
                                                             This option is currently supported only on Windows together with netcoreapp2.1 and newer. And on Linux with netcoreapp3.1 and newer. OSX and UWP
                                                             are not supported.
  --nologo                                                   Run test(s), without displaying Microsoft Testplatform banner
  -c, --configuration <CONFIGURATION>                        The configuration to use for running tests. The default for most projects is 'Debug'.
  -f, --framework <FRAMEWORK>                                The target framework to run tests for. The target framework must also be specified in the project file.
  -r, --runtime <RUNTIME_IDENTIFIER>                         The target runtime to test for.
  --no-restore                                               Do not restore the project before building.
  --interactive                                              Allows the command to stop and wait for user input or action (for example to complete authentication).
  -v, --verbosity <LEVEL>                                    Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
  -a, --arch <arch>                                          The target architecture.
  --os <os>                                                  The target operating system.
  -?, -h, --help                                             Show command line help.


Additional Arguments:
  Arguments passed to the application that is being run.
```

The 'additional arguments' portion seems to have resources defined, but I don't see where those resources are being accessed.

baronfel:
Those are defined in the core Localized resources of System.CommandLine, so unless we remove them for all commands there's no getting around it.

baronfel:
Using the unmatched tokens approach is perhaps cleaner - it doesn't pollute the help output, and it is easy to union the unmatched tokens on the total list of known arguments. The final result seems to work well:

```
2:39:05 ❯ dotnet test .\demo.csproj /p:Configuration=Release -bl
D:\Code\dotnet-sdk\.dotnet\sdk\7.0.100-preview.7.22377.5\MSBuild.dll -nologo -bl -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,D:\Code\dotnet-sdk\.dotnet\sdk\7.0.100-preview.7.22377.5\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,D:\Code\dotnet-sdk\.dotnet\sdk\7.0.100-preview.7.22377.5\dotnet.dll -maxcpucount -nodereuse:false -restore -target:VSTest -verbosity:m /p:Configuration=Release -property:VSTestArtifactsProcessingMode=collect -property:VSTestSessionCorrelationId=307080_ee2e0d4f-c859-4f43-860e-f21092c97c4b .\demo.csproj
```

In this example you can see both the `/p` property and the `-bl` flag forwarded to the test target.

baronfel:
@nohwnd I pinged you for review because you're familiar with everything going on here.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -32,7 +32,7 @@ public static int Run(ParseResult parseResult)
             // from the VSTest side.
             string testSessionCorrelationId = $"{Environment.ProcessId}_{Guid.NewGuid()}";
 
-            string[] args = parseResult.GetArguments();
+            string[] args = parseResult.GetArguments().Concat(parseResult.UnmatchedTokens).ToArray();
```
nohwnd, line 35: As we discussed in chat this adds duplicate parameters to the args that are passed directly to VSTest. 

Instead we need fixing in the path that is a bit lower that sends the parameters to msbuild. 

  .WithEnvironmentVariable("DOTNET_CLI_VSTEST_TRACE", "1") writes how we call msbuild to screen, and can aid in debugging this. 
  
There seems to be special handling for CollectCoverage, and it also differs when a test project is provided or not. 

Will have another look tomorow.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -118,15 +118,21 @@ private static TestCommand FromParseResult(ParseResult result, string[] settings
                 "-nologo"
             };
 
-            msbuildArgs.AddRange(result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()));
+
+            var parsedArgs =
+                    result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
+                    .Concat(result.UnmatchedTokens); // all tokens that the test-parser doesn't explicitly track
```
baronfel, line 124: this is the real fix - this method doesn't call `ParseResult.GetArguments()` and so it only receives the 'understood' options/arguments on the command line. Since properties aren't added to the grammar for `dotnet test` (yet), the unknown tokens that they end up in should be forwarded along to MSBuild unchanged.

src/Tests/dotnet-test.Tests/GivenDotnetTestContainsMSBuildParameters.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        private const string TestAppName = "VSTestMSBuildParameters";
+        private const string MSBuildParameter = "/p:Version=1.2.3";
+
+        public GivenDotnetTestContainsMSBuildParameters(ITestOutputHelper log) : base(log)
+        {
+        }
+
+        [InlineData($"{TestAppName}.csproj")]
+        [InlineData(null)]
```
baronfel, line 27: We hit different branches in the `dotnet test` command logic based on if there is a project/sln/dll specified or not, so we need to have test cases for both.

src/Cli/dotnet/commands/dotnet-test/Program.cs
Evangelink, line 124: Nit: Indent is too deep

```suggestion
            var parsedArgs =
                result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
                    .Concat(result.UnmatchedTokens); // all tokens that the test-parser doesn't explicitly track
```

Evangelink: Or if you prefer to keep `.Concat` on the same indent.

```suggestion
            var parsedArgs =
                result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
                .Concat(result.UnmatchedTokens); // all tokens that the test-parser doesn't explicitly track
```

src/Tests/dotnet-test.Tests/GivenDotnetTestContainsMSBuildParameters.cs
Evangelink: Shall we have a 3rd entry for the `.dll` case?

src/Assets/TestProjects/VSTestMSBuildParameters/Tests.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace TestNamespace
+{
+    [TestClass]
+    public class Tests
+    {
+        [TestMethod]
+        public void TestMSBuildParameters()
+        {
+            var assemblyInfoVersion = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+            Assert.AreEqual(assemblyInfoVersion, "1.2.3");
```
nohwnd, line 18: Expected is the first parameter.

```suggestion
            Assert.AreEqual("1.2.3", assemblyInfoVersion);
```

src/Tests/dotnet-test.Tests/GivenDotnetTestContainsMSBuildParameters.cs
baronfel: The .dll case doesn't hit the same code paths at all, and doesn't forward MSBuild parameters (because there's no build happening) so at best it would be a negative test. Is that valuable?

nohwnd:
@baronfel I will be leaving for paternity leave some time soon, @Evangelink will take over this for me. We would like to get it into 7.0.1xx servicing.

baronfel:
@nohwnd congratulations and best of luck! I'll work with @Evangelink to get this in. A little slammed ATM with .NET Conf prep, but we'll make it happen.

src/Tests/dotnet-test.Tests/GivenDotnetTestContainsMSBuildParameters.cs
Evangelink: Doesn't feel valuable.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -118,15 +119,23 @@ private static TestCommand FromParseResult(ParseResult result, string[] settings
                 "-nologo"
             };
 
-            msbuildArgs.AddRange(result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()));
+            var parsedArgs =
+                result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()) // all msbuild-recognized tokens
+                    .Concat(args); // all tokens that the test-parser doesn't explicitly track (minus the settings tokens)
```
Evangelink, line 124: @baronfel I have updated this part because we should not pass the "settings" arguments to msbuild (at least not directly).

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -118,15 +119,23 @@ private static TestCommand FromParseResult(ParseResult result, string[] settings
...
+
+            VSTestTrace.SafeWriteTrace(() => $"MSBuild args from forwarded options: {String.Join(", ", parsedArgs)}" );
+            msbuildArgs.AddRange(parsedArgs);
 
             if (settings.Any())
             {
                 //workaround for correct -- logic
                 var commandArgument = result.GetValueForArgument(TestCommandParser.SlnOrProjectArgument);
-                if(!string.IsNullOrWhiteSpace(commandArgument) && !settings.Contains(commandArgument))
+                // TODO: @baronfel, @Evangelink
```
Evangelink, line 133: @baronfel There is something still off with this (see comment below).

Evangelink (commented):
I will check with team tomorrow for help but I cannot run/debug tests from VS nor from command line using `dotnet` although I am following the steps described in https://github.com/dotnet/sdk/blob/main/documentation/project-docs/developer-guide.md.

So I pushed only a partial change in case you have some time to push forward. Otherwise, I will resume my work tomorrow.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -118,29 +119,28 @@ private static TestCommand FromParseResult(ParseResult result, string[] settings
                 "-nologo"
             };
 
-            msbuildArgs.AddRange(result.OptionValuesToBeForwarded(TestCommandParser.GetCommand()));
+            // Extra msbuild properties won't be parsed and so end up in the UnmatchedTokens list. In addition to those
+            // properties, all the test settings properties are also considered as unmatched but we don't want to forward
+            // these as-is to msbuild. So we filter out the test settings properties from the unmatched tokens.
```
nohwnd, line 124:
```suggestion
            // these as-is to msbuild. So we filter out the test settings properties from the unmatched tokens,
            // by only taking values until the first item after `--`. (`--` is not present in the UnmatchedTokens ).
```

nohwnd (approved):

eng/Version.Details.xml
```diff
@@ -42,56 +42,56 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>c623d96fe33cc360bd9981c9e6f6a646d7881d9e</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.Build" Version="17.4.0-preview-22464-02">
```
baronfel, line 45: I'm a little worried about these version changes - does the PR need to be rebased or something to prevent incorporating bad versions?

Evangelink: @baronfel Changes to `Version.Details.xml` and `Versions.props` were done by you (i.e. were there in the PR when I took it over) so I assumed they were needed. It might be worth rebasing from the branch as you suggested.

baronfel: ok cool - glad to hear it. I'll fix that up now then if that's ok?

Evangelink: Sure go for it! Thank you

baronfel: @Evangelink done with the rebase. Can you take a quick look and make sure that nothing changed with your fixes?  The rebase applied cleanly, so I don't think there were any issues, but additional verification would be great.

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -56,10 +57,10 @@ public static int Run(ParseResult parseResult)
                 return ForwardToVSTestConsole(parseResult, args, settings, testSessionCorrelationId);
             }
 
-            return ForwardToMsbuild(parseResult, settings, testSessionCorrelationId);
+            return ForwardToMsbuild(parseResult, args, settings, testSessionCorrelationId);
         }
 
-        private static int ForwardToMsbuild(ParseResult parseResult, string[] settings, string testSessionCorrelationId)
+        private static int ForwardToMsbuild(ParseResult parseResult, string[] args, string[] settings, string testSessionCorrelationId)
```
Evangelink, line 63: Forgot to cleanup this part:

```suggestion
        private static int ForwardToMsbuild(ParseResult parseResult, string[] settings, string testSessionCorrelationId)
```

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -56,10 +57,10 @@ public static int Run(ParseResult parseResult)
                 return ForwardToVSTestConsole(parseResult, args, settings, testSessionCorrelationId);
             }
 
-            return ForwardToMsbuild(parseResult, settings, testSessionCorrelationId);
+            return ForwardToMsbuild(parseResult, args, settings, testSessionCorrelationId);
```
Evangelink, line 60: Forgot to cleanup this part:

```suggestion
            return ForwardToMsbuild(parseResult, settings, testSessionCorrelationId);
```

src/Cli/dotnet/commands/dotnet-test/Program.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Parsing;
+using System.Diagnostics;
```
Evangelink, line 8: Forgot to cleanup this part:

```suggestion
```

src/Cli/dotnet/commands/dotnet-test/TestCommandParser.cs
```diff
@@ -15,13 +15,7 @@ namespace Microsoft.DotNet.Cli
...
     internal static class TestCommandParser
     {
-        public static readonly string DocsLink = "https://aka.ms/dotnet-test";
-
-        public static readonly Argument<string> SlnOrProjectArgument = new Argument<string>(CommonLocalizableStrings.SolutionOrProjectArgumentName)
-        {
-            Description = CommonLocalizableStrings.SolutionOrProjectArgumentDescription,
-            Arity = ArgumentArity.ZeroOrOne
-        };
+        public static readonly string DocsLink = "https://aka.ms/dotnet-test";
```
Evangelink, line 18: Nit: Looks like I somehow added extra trailing spaces.

```suggestion
        public static readonly string DocsLink = "https://aka.ms/dotnet-test";
```

Evangelink (approved):
@baronfel I added some nitpicks, could you please apply them?

Evangelink (approved):

PR #27962, opened 2022-09-16
Make mapping infrastructure unit testable, Ignore effectively sealed members and types and add tests

ViktorHofer:
[Make mapping infrastructure unit testable](https://github.com/dotnet/sdk/commit/f504751f76b858de00287d120f47047bc3ec6631)

Define interfaces for the mapping types in order for them to not depend
on concrete implementations in order for them to be unit test-able.

[Ignore effectively sealed members and types](https://github.com/dotnet/sdk/commit/6871f5ac87585c0b5b814bbb7798d31b65f43d89) 

Don't map effectively sealed members and types so that individual rules
don't have to special case them.

[Add unit tests for mapper types and eff sealed case](https://github.com/dotnet/sdk/commit/8870ebff977ee12bd8f503b7282b5b41a2a88432)

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
@@ -62,7 +62,7 @@ namespace Microsoft.DotNet.ApiCompatibility.Extensions
             {
                 foreach (IMethodSymbol constructor in namedType.Constructors)
                 {
-                    if (!constructor.IsStatic && constructor.IsVisibleOutsideOfAssembly(includeInternals))
+                    if (!constructor.IsStatic && constructor.IsVisibleOutsideOfAssembly(includeInternals, includeEffectivelySealedSymbols: true))
                         return true;
                 }
             }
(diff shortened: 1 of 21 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs
```diff
@@ -62,7 +62,7 @@ private static bool HasVisibleConstructor(ITypeSymbol type, bool includeInternal
             {
                 foreach (IMethodSymbol constructor in namedType.Constructors)
                 {
-                    if (!constructor.IsStatic && constructor.IsVisibleOutsideOfAssembly(includeInternals))
+                    if (!constructor.IsStatic && constructor.IsVisibleOutsideOfAssembly(includeInternals, includeEffectivelySealedSymbols: true))
```
smasher164, line 65: Although the containing type is effectively sealed, would you describe the members that way too? I feel like `includeEffectivelyPrivateSymbols` is more accurate, however a user won't really know what that means unless we explain it. So I guess this is fine.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Mappers/AssemblyMapperTests.cs
```diff
@@ -0,0 +1,82 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiCompatibility.Abstractions;
+using Microsoft.DotNet.ApiCompatibility.Rules;
+using Moq;
```
smasher164, line 11: Is using `Moq` preferable over implementing our own `IAssemblyMapper` that keeps track of the settings, size, etc.?
Likewise for the other interfaces.

ViktorHofer: Using [Moq](https://github.com/moq/moq4#moq) avoids implementing interfaces yourself when you just need a simple (or even a complex) implementation. Without Moq, dependencies would need to be manually implemented.

In code above, you just want to mock the dependency without specifying any explicit behavior, because the dependency isn't used as part of the AssemblyMapper. You just want to verify that what is passed into the ctor is correctly set as a property.

For other units (not part of this PR), the dependencies are actually used, i.e. to retrieve members from a type and such mocks need to have some behavior attached which is also straightforward with Moq. I.e. https://github.com/dotnet/sdk/blob/fb2651af5fec4a910ddcd9b8007000debceaab18/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Runner/ApiCompatRunnerTests.cs#L20-L26

smasher164 (approved):

PR #27964, opened 2022-09-16
Avoid extra copies of dotnet-watch build outputs in the SDK directory

tmat:
`DotNetWatch.targets` was copied to the output dir of `redist.csproj` and from there to multiple other directories where they shouldn't be.


tmat:
@javiercn

dsplaisted (commented):
Does this change what we ship in the SDK, or just what happens for local builds?

Is this needed for RC2?  If it makes an actual change to the product I think we are at the point where we would avoid doing it for RC2 unless we really need it, due to risk.

Overall this sounds like a good change, I'm just not sure what branch it should go into.

marcpopMSFT:
I think this is probably too risky at this point. All RC2 changes now require approval and the RTM branch won't have any more previews. We'll have a 7.0.2xx branch open in the coming weeks or main should be fine.

tmat:
Moving to main.

dsplaisted (approved):
I believe we will be creating a 7.0.2xx branch next week, you may want to wait to merge it there (otherwise it will only go in 8.0 unless we also backport it).

tmat:
@dsplaisted Rebased to 7.0.2xx

PR #27969, opened 2022-09-16
Implement publish-self-contained

nagilson:
Adds PublishSelfContained (PSC):

1. You can add 
```xml
<PublishSelfContained>true</PublishSelfContained>
```
to publish self-contained by default but not build self-contained by default. You can of course also forward the property.

2. If you do `dotnet publish --no-sc -p:PublishSelfContained=true`, `--no-sc` will win. Or if `PublishSelfContained` is in the properties.

4. Had a large group discussion about getting these things to work in MSBuild with `t:/Publish` -- it's not possible. But we are going to do work to get it to align with VS.

5. `PublishSelfContained` will also imply a RID.


nagilson:
GH isn't updating the PR correctly, just going to make a new one

Issue #28022, opened 2022-09-20
[NetSDKE2E]Running "dotnet publish -f net7.0" with .net7 rc2 SDK,the output shown contains an invalid character "-".


v-ainigao:
Environment:  7.0.100-rc.2.22466.28(runtime-7.0.0-rc.2.22464.1)
Repro steps:
1.Install 7.0.100-rc.2.22466.28(runtime-7.0.0-rc.2.22464.1)
2.create a console in cli
3.run"dotnet publish -f net7.0"
Expected:
The output shown doesn't contains an invalid character "-".
Actual:
The output shown contains an invalid character "-".
![MicrosoftTeams-image (6)](https://user-images.githubusercontent.com/106054229/191155042-fd840952-4356-4d70-8b36-0fa2352d1d9f.png)

dotnet --info
C:\Users\v-jiexu1>dotnet --info
.NET SDK:
Version:   7.0.100-rc.2.22466.28
Commit:    60fdbb471d

Runtime Environment:
OS Name:     Windows
OS Version:  10.0.20348
OS Platform: Windows
RID:         win10-x64
Base Path:   C:\Program Files\dotnet\sdk\7.0.100-rc.2.22466.28\

Host:
  Version:      7.0.0-rc.2.22464.1
  Architecture: x64
  Commit:       c623d96fe3

.NET SDKs installed:
  7.0.100-rc.2.22466.28 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.0-rc.2.22464.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
[...]

PR #28059, opened 2022-09-20
Fix Property Parsing in Publish

nagilson:
Resolves https://github.com/dotnet/sdk/issues/28022
Backport: 
https://github.com/dotnet/sdk/pull/28061 **(please approve this too!)**

Post-Moderm:

I intercepted the `MSBuildPropertyParser` code stack too many interface layers in, so for some edge cases, properties passed via `/p` and `-p` were not fully parsed as it wasn't running through the `Options` Forwarding code. The ones I had tested did get parsed but for example in the above case they weren't fully parsed. Also removes an erroneous comment from the code that was old and slipped in. 

Risk:
Medium

Reason To Accept: 
Fixes a bug which will cause erroneous errors inside of `dotnet publish`. The PublishRelease code will do an MSBuild evaluation with faulty properties which could have any number of side effects, but in general it would be unrelated to what actually gets published. Still, this is high impact and should be merged in. 

Testing: [OK]
`dotnet publish -p:PublishRelease=true -property Configuration=Debug`
`dotnet publish -p:PublishRelease=true -p    Configuration=Debug`
Also 
`dotnet publish -f net7.0`
Also
`dotnet publish` with `PublishRelease` inside of a `.csproj`

### DIFF
--- a/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
+++ b/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
@@ -39,7 +39,7 @@ namespace Microsoft.DotNet.Cli
             )
         {
             ProjectInstance project = null;
-            var globalProperties = GetGlobalPropertiesFromUserArgs(userPropertyArgs);
+            var globalProperties = GetGlobalPropertiesFromUserArgs(parseResult);
 
             if (parseResult.HasOption(configOption) || globalProperties.ContainsKey(MSBuildPropertyNames.CONFIGURATION))
                 yield break;
@@ -186,17 +186,14 @@ namespace Microsoft.DotNet.Cli
         }
 
         /// <returns>A case-insensitive dictionary of any properties passed from the user and their values.</returns>
-        private Dictionary<string, string> GetGlobalPropertiesFromUserArgs(IEnumerable<string> userPropertyArgs)
+        private Dictionary<string, string> GetGlobalPropertiesFromUserArgs(ParseResult parseResult)
         {
             Dictionary<string, string> globalProperties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
-            var forwardingFunction = (CommonOptions.PropertiesOption as ForwardedOption<string[]>).GetForwardingFunction();
-            IEnumerable<string> globalPropEnumerable = forwardingFunction(CommonOptions.PropertiesOption.Parse(String.Join(";", userPropertyArgs)));
+            string[] globalPropEnumerable = parseResult.GetValueForOption(CommonOptions.PropertiesOption);
 
-            foreach (var propertyString in globalPropEnumerable)
+            foreach (var keyEqVal in globalPropEnumerable)
             {
-                // The parser puts keys into the format --property:Key=Value no matter the user input, so we can expect that pattern and extract the key.
-                string keyEqVal = propertyString.Split("--property:").Last();
                 string[] keyValuePair = keyEqVal.Split("=");
                 globalProperties[keyValuePair.First()] = keyValuePair.Last();
             }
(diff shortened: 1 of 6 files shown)


nagilson:
Darwin pipeline failing due to infrastructure issues. Opening PR for review

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -563,6 +563,29 @@ public void It_warns_if_PublishRelease_set_on_sln_but_env_var_not_used()
                 .HaveStdOutContaining("NETSDK1190");
         }
 
+        [Theory]
+        [InlineData("-f net7.0")]
+        [InlineData("/p:TargetFramework=net7.0")]
```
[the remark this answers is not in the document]
nagilson: Good point, thanks for generating these. I'll incorporate those tests.

nagilson:
Approved via email

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -186,15 +185,16 @@ private bool IsValidProjectFilePath(string path)
...
 
-            // The parser puts keys into the format --property:Key no matter the user input, so we can expect that pattern and extract the key.
-            foreach (var kvPair in globalPropEnumerable)
+            string[] globalPropEnumerable = parseResult.GetValueForOption(CommonOptions.PropertiesOption);
+
+            foreach (var keyEqVal in globalPropEnumerable)
             {
-                globalProperties[kvPair.key.Split(new string[] { "--property:" }, StringSplitOptions.None).Last()] = kvPair.value;
+                string[] keyValuePair = keyEqVal.Split("=");
+                globalProperties[keyValuePair.First()] = keyValuePair.Last();
```
dsplaisted, line 197: I'd suggest indexing directly into the array here:

```suggestion
                globalProperties[keyValuePair[0]] = keyValuePair[1];
```

dsplaisted (approved):

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -186,15 +185,16 @@ private bool IsValidProjectFilePath(string path)
...
-            IEnumerable<(string key, string value)> globalPropEnumerable = MSBuildPropertyParser.ParseProperties(String.Join(";", userPropertyArgs));
 
-            // The parser puts keys into the format --property:Key no matter the user input, so we can expect that pattern and extract the key.
-            foreach (var kvPair in globalPropEnumerable)
+            string[] globalPropEnumerable = parseResult.GetValueForOption(CommonOptions.PropertiesOption);
+
+            foreach (var keyEqVal in globalPropEnumerable)
             {
-                globalProperties[kvPair.key.Split(new string[] { "--property:" }, StringSplitOptions.None).Last()] = kvPair.value;
+                string[] keyValuePair = keyEqVal.Split("=");
```
baronfel, line 196: This split concerns me. a property value like `RepositoryUrl="https://repo.mycorp.com/repos?id=321432asdfs=="` would split incorrectly and we'd mangle the value.  I think you should use `keyEqVal.Split("=", 2)` to ensure that only two substrings are returns (i.e. that only the first `=` is split on).

baronfel (changes requested):
One note, but the rest of it looks completely reasonable to me.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Done!

baronfel (approved):
Thanks for following up on that, let's merge this!

PR #28060, opened 2022-09-20
Gate warnings for localization asset locales on TFM >= 7

### LINKED ISSUE #28058: Invalid and unknown locale warnings for package assets should be gated on TFM
### Describe the bug
In https://github.com/dotnet/sdk/pull/26729 we introduced a new warning to .NET 7 SDKs, but we did not gate that warning on a TFM change. This means that users that work on case-insensitive file systems are very likely to start hitting the new warning on an SDK update, instead of a TFM update. We want to not do that so that we encourage folks to update SDKs as soon as they come out.

baronfel:
## Description
In https://github.com/dotnet/sdk/pull/26729 we introduced a new warning to .NET 7 SDKs around inconsistent locale casing for resources found in NuGet packages, but we did not gate that warning on a TFM change. This means that users that work on case-insensitive file systems are very likely to start hitting the new warning on an SDK update, instead of a TFM update. We want to not do that so that we encourage folks to update SDKs as soon as they come out. We have had reports of a couple of internal customers 

## Regression

**No**, instead we want to change the default. I suppose you could call it a user-experience regression?

## Risk

**Low**, this uses existing patterns to check an already-available piece of data.

## Testing

- [x] manual testing
- [x] automated tests to cover the matrix of TFM versions and error codes we added

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
@@ -4,7 +4,6 @@
 using FluentAssertions;
 using Microsoft.Build.Framework;
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -159,7 +158,7 @@ namespace Microsoft.NET.Build.Tasks.UnitTests
             var task = InitializeTask(out _);
             task.ProjectAssetsFile = projectAssetsJsonPath;
             task.TargetFramework = tfm;
-            var writer = new CacheWriter(task);
+            var writer = new CacheWriter(task, new MockPackageResolver());
             writer.WriteToMemoryStream();
             var engine = task.BuildEngine as MockBuildEngine;
 
@@ -167,7 +166,7 @@ namespace Microsoft.NET.Build.Tasks.UnitTests
             invalidContextWarnings.Should().HaveCount(shouldHaveWarnings ? 1 : 0);
 
             var invalidContextMessages = engine.Messages.Where(msg => msg.Code == "NETSDK1188");
-            invalidContextWarnings.Should().HaveCount(!shouldHaveWarnings ? 1 : 0);
+            invalidContextMessages.Should().HaveCount(!shouldHaveWarnings ? 1 : 0);
 
         }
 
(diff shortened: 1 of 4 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/IPackageResolver.cs
```diff
@@ -9,5 +10,6 @@ internal interface IPackageResolver
     {
         string GetPackageDirectory(string packageId, NuGetVersion version);
         string GetPackageDirectory(string packageId, NuGetVersion version, out string packageRoot);
+        string ResolvePackageAssetPath(LockFileTargetLibrary package, string relativePath);
```
dsplaisted, line 13: Is this new method actually used?

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -677,6 +677,11 @@ internal sealed class CacheWriter : IDisposable
 
             private const char RelatedPropertySeparator = ';';
 
+            public CacheWriter(ResolvePackageAssets task, IPackageResolver resolver) : this(task)
```
dsplaisted, line 680: Suggestion: add a comment here that this constructor is for tests.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
```diff
@@ -136,59 +131,65 @@ public void It_warns_on_invalid_culture_codes_of_resources()
...
+            task.TargetFramework = tfm;
+            var writer = new CacheWriter(task, new MockPackageResolver());
+            writer.WriteToMemoryStream();
+            var engine = task.BuildEngine as MockBuildEngine;
+
+            var invalidContextWarnings = engine.Warnings.Where(msg => msg.Code == "NETSDK1187");
+            invalidContextWarnings.Should().HaveCount(shouldHaveWarnings ? 1 : 0);
+
+            var invalidContextMessages = engine.Messages.Where(msg => msg.Code == "NETSDK1187");
+            invalidContextMessages.Should().HaveCount(!shouldHaveWarnings ? 1 : 0);
```
dsplaisted, line 192: Nit: I think this is a bit clearer

```suggestion
            invalidContextMessages.Should().HaveCount(shouldHaveWarnings ? 0 : 1);
```

dsplaisted (approved):

src/Tasks/Microsoft.NET.Build.Tasks/IPackageResolver.cs
baronfel: It's used on line 1036 (not part of these changes) but since the type of `packageResolver` changed from `NuGetPackageResolver` to `IPackageResolver` to allow for mocks, the interface needed to learn the method.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
baronfel: good call - pushed just now.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
baronfel: Good call - did this in both places with the inverted condition

baronfel:
Approved via email

PR #28167, opened 2022-09-26
[release/7.0.1xx] Provide package description for APICompat packages

ViktorHofer:
Backport of https://github.com/dotnet/sdk/pull/28102 to release/7.0.1xx

## Customer Impact
Customers who search for the APICompat packages on frontends (website, VS) will notice that these packages are missing a description, and in the worst-case wonder if these are legit packages. By providing a description, the intent of the package is made clear.

## Testing
I verified that the produced package's metadata now contains a package description.

## Risk
Low. The compile assemblies aren't impacted by this change. Only the metadata in the NuGet nuspec is impacted.


ViktorHofer:
@ericstj This is adding a package description to the two shipping APICopmat packages. Can you help me understand what the current bar for such changes is?

ericstj:
This seems like a reasonable change to make.  Fit and finish, low risk.  I don't think this impacts localization.  It does impact the product bits so I wouldn't classify as "infrastructure".

WRT process @marcpopMSFT can we do this as tell-mode or is SDK in ask mode for product changes right now?

carlossanlop (approved):
LGTM.

Pedantic nit: I would've used "API" instead of "api". But the main PR has already been merged, so it's ok.

marcpopMSFT:
> Unsure what "QB mode" stands for but I guess it means "Tactics approval"?

Correct, you'll need tactics approval (fill in the template, mark servicing consider, send mail/show up to tactics).

PR #28224, opened 2022-09-28
Add warning when targeting 7.0 in 17.3 and remove 7.0 from the list of supported targets

marcpopMSFT:
## Description

Targeting 7.0 will not work in 17.3 because of source generators that will light up when targeting net7.0. We're not sure the exact error experience. This PR will provide a warning message to customers and make that option not show up in the property pages either.

## Customer Impact
New warning when targeting 7.0 and using 17.3 MSBuild.  Customers won't be able to swtich to use net7.0 from the VS properties.

## Regression

No

## Risk 
Low. This is already not supported

## Testing 
manual/automated

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -871,4 +871,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1191: A runtime identifier for the property '{0}' couldn't be inferred. Specify a rid explicitly.</value>
       <comment>{StrBegin="NETSDK1191: "}</comment>
   </data>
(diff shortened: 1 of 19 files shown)


dsplaisted (approved):

marcpopMSFT:
@dsplaisted it turns out VisualStudioVersion is set to 17.0 even in 17.4. We can't rely on that property. I switched to using MSBuild instead as that seems to work but pinging you to raise visibility here for this.

dsplaisted:
> @dsplaisted it turns out VisualStudioVersion is set to 17.0 even in 17.4. We can't rely on that property. I switched to using MSBuild instead as that seems to work but pinging you to raise visibility here for this.

Makes sense.

PR #28233, opened 2022-09-28
[release/7.0.1xx] Revert "do not emit diagnostic when attribute argument is internal type (#27648)"

smasher164:
Backport of https://github.com/dotnet/sdk/pull/27774 to release/7.0.1xx

## Customer Impact
TL;DR: Missing functionality in a new opt-in 7.0 feature.

Customers who turn on the `AttributesMustMatch` rule expect that internal type arguments are compared between two attributes. However, without this change, those arguments are ignored when the attributes themselves are public. This would cause breaking changes for certain attribute arguments to not get flagged.

## Testing
I verified that internal type arguments are no longer ignored.

## Risk
Low. This would only affect users who turn on the `AttribtuesMustMatch` rule and have not set the `IncludeInternalSymbols` flag to `true`.


ViktorHofer (approved):

ericstj:
Approved over email.

PR #28243, opened 2022-09-29
don't emit "attribute added" diagnostic when not in strict mode

smasher164:


ViktorHofer:
The PR title is misleading. Shouldn't it say "... in non strict mode"?

ViktorHofer (approved):

PR #28290, opened 2022-10-01
apicompat: attribute changed on right must only be emitted in strict mode

smasher164:
When an attribute is repeated on the right hand side, with more arguments, it shouldn't be considered a change outside of strict mode. This is because the attributes on the left are a subset of the ones on the right. For example,
```C#
// Left
[Foo("a")]
void F() {}

// Right
[Foo("a")]
[Foo("b")]
void F() {}
```


ViktorHofer:
Would it make sense to add an additional test that has a single attribute only on the left i.e. `[Serializable]` but two identical `[Serializable]` attributes on the right?

smasher164:
@ViktorHofer 

> Would it make sense to add an additional test that has a single attribute only on the left i.e. `[Serializable]` but two identical `[Serializable]` attributes on the right?

That test case would fail to compile, since the compiler will report that `[Serializable]` is duplicated. I *could* add a test case that duplicates `[Foo("S")]`. Based on this rule, that would be permissible in non-strict mode, and produce a warning in strict mode.

ViktorHofer (approved):

Issue #28300, opened 2022-10-03
APICompat.Task targets file should be validated in CI


ViktorHofer:
Apparently, the APICompat.Task.targets file in the `build` folder isn't validated on CI. That was noticed in https://github.com/dotnet/sdk/commit/0a1f5cdffe8f9e4badc6a4c9da22644b74bc9b18.

ericstj:
Is it straight forward to produce a test assembly which references this (either through package or direct targets import) in the SDK's test infrastructure?

ViktorHofer:
I believe existing tests already do validate the msbuild file but are disabled because of https://github.com/dotnet/sdk/issues/23533.

smasher164:
Closing as duplicate of #23533.

PR #28301, opened 2022-10-03
[release/7.0.1xx] Fix msbuild evaluation of ApiCompat task pkg

ViktorHofer:
Backport of https://github.com/dotnet/sdk/commit/0a1f5cdffe8f9e4badc6a4c9da22644b74bc9b18.

## Customer Impact
Customers who reference the `Microsoft.DotNet.ApiCopmat.Task` package to make use of the very latest APICompat features (which aren't yet available in the SDK) will notice an msbuild evaluation error that will effectively break their build. Fixing the broken msbuild condition which recently regressed as part of https://github.com/dotnet/sdk/commit/e99f2b75d33bb0cc1bb9f019f447896e2077d58c. 

## Testing
I verified that adding a reference to the `Microsoft.DotNet.ApiCompat.Task` nuget package doesn't result in an msbuild evaluation error anymore and that the component works as intended. Filed https://github.com/dotnet/sdk/issues/28300 for adding CI protection.

## Risk
Very low. The msbuild evaluation is unblocked and the condition still works as expected.


ericstj (approved):
LGTM, thank you for thinking about adding CI protection.

ViktorHofer:
Approved offline by Tactics via mail.

PR #28338, opened 2022-10-05
[release/7.0.1xx] Add missing EnableRuleAttributesMustMatch task property

ViktorHofer:
Backport of https://github.com/dotnet/sdk/pull/28336 & https://github.com/dotnet/sdk/pull/28344

## Customer Impact
Customers who use [APICompat's PackageValidation feature](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview) can't enable the optional `AttributesMustMatch` rule (in the msbuild frontend) that validates that attributes are compatible between the contract and the implementation. The property/parameter is already exposed in the task but the targets file never passed the msbuild property `ApiCompatEnableRuleAttributesMustMatch` into the task.

## Testing
I verified that the property is now passed into the `ValidatePackage` task correctly and filed https://github.com/dotnet/sdk/issues/28337 for adding CI protection.

## Risk
Low. This is an opt-in rule and customers need to explicitly set the msbuild property to true.


smasher164 (approved):

ViktorHofer:
Approved offline by Tactics

PR #28348, opened 2022-10-05
normalize accessibility when not including internal symbols

smasher164:
Previously, the comparison between `Accessibility` values returned by Roslyn relied on comparing their numerical enum values.
However, that ordering neither considered `protected` of higher accessibility than `internal`, nor did it treat `internal` members as `private`, when internal symbols are excluded.

This meant that when internal symbols were excluded, the accessibility of internal symbols would still affect the compatibility checker, and it would emit unwanted diagnostics in those circumstances.

This change defines a custom comparator for `Accessibility` values that takes these factors into account.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Runtime;
 using Microsoft.CodeAnalysis;
@@ -24,6 +25,15 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
             context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
         }
 
+        private Accessibility normalizeAccessibility(Accessibility a) =>
+            (_settings.IncludeInternalSymbols, a) switch
+            {
+                (false, Accessibility.ProtectedAndInternal
+                        or Accessibility.Internal
+                        or Accessibility.ProtectedOrInternal) => Accessibility.Protected,
+                _ => a,
+            };
+
         private void RunOnSymbol(
             ISymbol? left,
             ISymbol? right,
@@ -37,8 +47,8 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
                 return;
             }
 
-            Accessibility leftAccess = left.DeclaredAccessibility;
-            Accessibility rightAccess = right.DeclaredAccessibility;
+            Accessibility leftAccess = normalizeAccessibility(left.DeclaredAccessibility);
+            Accessibility rightAccess = normalizeAccessibility(right.DeclaredAccessibility);
 
             if (leftAccess > rightAccess)
             {
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
@@ -25,14 +25,39 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
             context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
         }
 
-        private Accessibility normalizeAccessibility(Accessibility a) =>
-            (_settings.IncludeInternalSymbols, a) switch
+        private static Accessibility normalizeInternals(Accessibility a) => a switch
+        {
+            Accessibility.ProtectedOrInternal => Accessibility.Protected,
+            Accessibility.ProtectedAndInternal or Accessibility.Internal => Accessibility.Private,
+            _ => a,
+        };
+
+        private int compareAccessibility(Accessibility a, Accessibility b)
+        {
+            if (!_settings.IncludeInternalSymbols)
             {
-                (false, Accessibility.ProtectedAndInternal
-                        or Accessibility.Internal
-                        or Accessibility.ProtectedOrInternal) => Accessibility.Protected,
-                _ => a,
+                a = CannotChangeVisibility.normalizeInternals(a);
+                b = CannotChangeVisibility.normalizeInternals(b);
+            }
+
+            if (a == b)
+            {
+                return 0;
+            }
+
+            return (a, b) switch
+            {
+                (Accessibility.Public, _) => 1,
+                (_, Accessibility.Public) => -1,
+                (Accessibility.ProtectedOrInternal, _) => 1,
+                (_, Accessibility.ProtectedOrInternal) => -1,
+                (Accessibility.Protected or Accessibility.Internal, _) => 1,
+                (_, Accessibility.Protected or Accessibility.Internal) => -1,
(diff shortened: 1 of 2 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
```diff
@@ -24,6 +25,15 @@ public CannotChangeVisibility(RuleSettings settings, IRuleRegistrationContext co
             context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
         }
 
+        private Accessibility normalizeAccessibility(Accessibility a) =>
+            (_settings.IncludeInternalSymbols, a) switch
+            {
+                (false, Accessibility.ProtectedAndInternal
```
ericstj, line 31: Shouldn't `ProtectedAndInternal` and `Internal` be *excluded* when IncludeInternalSymbols is false?  IOW these should be mapped to Private.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
```diff
@@ -37,8 +47,8 @@ public CannotChangeVisibility(RuleSettings settings, IRuleRegistrationContext co
                 return;
             }
 
-            Accessibility leftAccess = left.DeclaredAccessibility;
-            Accessibility rightAccess = right.DeclaredAccessibility;
+            Accessibility leftAccess = normalizeAccessibility(left.DeclaredAccessibility);
+            Accessibility rightAccess = normalizeAccessibility(right.DeclaredAccessibility);
 
             if (leftAccess > rightAccess)
```
ericstj, line 53: When someone is looking at internal members, this would not flag an API breaking change from `protected` to `internal`

ericstj (approved):
LGTM. Thank you for addressing my feedback

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
```diff
@@ -24,6 +25,40 @@ public CannotChangeVisibility(RuleSettings settings, IRuleRegistrationContext co
             context.RegisterOnTypeSymbolAction(RunOnTypeSymbol);
         }
 
+        private static Accessibility normalizeInternals(Accessibility a) => a switch
```
ViktorHofer, line 28: Please start following the [C# Coding Coventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions) sooner than later. Even though the method is private, we use `CamelCase` in C#.

ViktorHofer (changes requested):
Left a comment regarding Coding Conventions.

ViktorHofer:
@smasher164 please either add a detailed commit description (and make sure that it shows up in the PR description) or make sure that you have a tracking issue with more details that is then being fixed as part of your changes. 

Changes like these aren't self-explanatory and I myself fail to understand their **exact** intent without more information. Try to author your PRs to that non domain experts would understand their purpose. I assume that you had a side conversation with @ericstj about an incorrect behavior, but unfortunately neither I nor others who want to provide feedback have that information available.

This comment's intent is to make our collaboration more efficient and your changes more inclusive to the open-source community. Please don't get me wrong as I enjoy your work and I think you are doing a great job and we just want to fine-tune the messaging of your changes.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
ericstj: If you change this, it might also be good to comment these around why this method exists.
Something like: When not comparing internals remove internal from the Accesssibility value.  You might also consider calling it "RemoveInternalAccessibility".

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
```diff
@@ -24,6 +25,40 @@ public CannotChangeVisibility(RuleSettings settings, IRuleRegistrationContext co
...
+            Accessibility.ProtectedAndInternal or Accessibility.Internal => Accessibility.Private,
+            _ => a,
+        };
+
+        private int CompareAccessibility(Accessibility a, Accessibility b)
+        {
+            if (!_settings.IncludeInternalSymbols)
+            {
+                a = CannotChangeVisibility.NormalizeInternals(a);
+                b = CannotChangeVisibility.NormalizeInternals(b);
```
ViktorHofer, line 40: nit: we usually don't specify the class's name from within that class.

```suggestion
                a = NormalizeInternals(a);
                b = NormalizeInternals(b);
```

smasher164: oh that's weird. i think the automated refactoring just did that. my bad

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/CannotChangeVisibility.cs
```diff
@@ -24,6 +25,40 @@ public CannotChangeVisibility(RuleSettings settings, IRuleRegistrationContext co
...
+            {
+                (Accessibility.Public, _) => 1,
+                (_, Accessibility.Public) => -1,
+                (Accessibility.ProtectedOrInternal, _) => 1,
+                (_, Accessibility.ProtectedOrInternal) => -1,
+                (Accessibility.Protected or Accessibility.Internal, _) => 1,
+                (_, Accessibility.Protected or Accessibility.Internal) => -1,
+                (Accessibility.ProtectedAndInternal, _) => 1,
+                (_, Accessibility.ProtectedAndInternal) => -1,
+                _ => throw new NotImplementedException(),
```
ViktorHofer, line 58: I'm a little bit worried about this line. This would cause a NotImplementedException being thrown when a new visibility is introduced by Roslyn. Is there a good default that we can provide for unknown accessibilities?

Also, if we would keep it, we would probably throw `ArgumentOutOfRangeException` instead.

smasher164: I'm not sure there is a default that applies to an `Accessibility` no matter where it is placed in the hierarchy. In particular, if it's an `Accessibility` like `Protected` or `Internal`, then it might special-casing, since all of this is a partial order.

`ArgumentOutOfRangeException` seems better suited for this scenario.

ViktorHofer: I would follow the approach that we take in SymbolExtensions to not diverge in behavior: https://github.com/dotnet/sdk/blob/30dfe0a0726d0f9fd3ffb7c528fd1f0cc8ec6ec8/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Extensions/SymbolExtensions.cs#L94

ViktorHofer: Hmm, actually would it make sense to just do a `_ => a > b ? 1 : b > a ? -1 : 0` as the default for new unknown visibility members? Then you could also remove L43-45 as it would be handled by that default switch statement.

ViktorHofer (approved):

PR #28400, opened 2022-10-07
Add an official builder assembly attribute

marcpopMSFT:
This setting was inverted in RC1 and we still haven't figured out why. Adding an attribute we can check for in the installer repo codeflow. If somehow this ends up different than the actual telemetry setting, it'll be a sign of a massive compiler failure so assume this is enough.

### DIFF
--- a/src/Common/CompileOptions.cs
+++ b/src/Common/CompileOptions.cs
@@ -3,7 +3,7 @@
 using System.Reflection;
 
 #if MICROSOFT_ENABLE_TELEMETRY
-      [assembly: AssemblyMetadata("OfficialBuilder", "Microsoft")]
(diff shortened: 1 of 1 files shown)


src/Common/CompileOptions.cs
```diff
@@ -1,5 +1,10 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Reflection;
+
+#if MICROSOFT_ENABLE_TELEMETRY
+      [assembly: AssemblyMetadata("OfficialBuilder", "Microsoft")]
```
dsplaisted, line 6: I was thinking of something like this, but no need to change it if you're happy with the OfficialBuilder metadata.

```c#
      [assembly: AssemblyMetadata("TelemetryOptOutDefault", CompileOptions.TelemetryOptOutDefault)]
```

dsplaisted (approved):

PR #28425, opened 2022-10-10
[release/7.0.1xx] Reference net7.0 version of msbuild for source-build

lbussell:
Fixes https://github.com/dotnet/source-build/issues/3048

Backport of this source-build patch: https://github.com/dotnet/installer/blob/860067bade6be3fbda19a3162abde99286c82df1/src/SourceBuild/tarball/patches/sdk/0002-Look-for-msbuild-net7.0-artifacts.patch

We changed msbuild to [target net7.0](https://github.com/dotnet/msbuild/pull/7790) but we still need to use an older version ([17.3.0 in this case](https://github.com/dotnet/sdk/blob/39b57e7c15ec09a0c230dade5f014d17be960b42/src/Layout/redist/minimumMSBuildVersion#L1)) in the Microsoft-built SDK for VS and VS for Mac compatibility. This means we need a couple of source-build conditions for this change. The change to `MicrosoftBuildPackageVersion` only affects the repo's source-build CI leg and not the source-build tarball build.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -112,7 +112,7 @@
           so target one that matches the version in minimumMSBuildVersion.
 
           This avoids the need to juggle references to packages that have been updated in newer MSBuild. -->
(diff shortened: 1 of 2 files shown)


MichaelSimons (approved):

crummel (approved):

lbussell:
@rainersigwald or @dsplaisted, can someone from the SDK or MSBuild teams review this so that this can make it into 7.0 GA? Thanks in advance.

dsplaisted (approved):

rainersigwald (approved):

PR #28430, opened 2022-10-10
6.0.1xx branding and test disablement

marcpopMSFT:


src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/BlazorWasmStaticWebAssetsIntegrationTest.cs
```diff
@@ -90,7 +90,7 @@ public void StaticWebAssets_PublishMinimal_Works()
                 intermediateOutputPath);
         }
 
-        [Fact]
+        [Fact(Skip="https://github.com/dotnet/sdk/issues/28429")]
```
dsplaisted, line 93: You could use a `RequiredMSBuildVersionFact` instead so that the test will automatically start running again.

dsplaisted (approved):

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/BlazorWasmStaticWebAssetsIntegrationTest.cs
marcpopMSFT: do you know which MSBuild version the fix is in and can we do preview versions like that?

TanayParikh (approved):

TanayParikh:
It's green!

PR #28476, opened 2022-10-11
[Internal -> External Port (7.0.1xx-RC2)] Fix SDK Temp Paths

nagilson:
Confirming file permissions are using 700:
The first image shows the temp path created by file utilities in the biggest exploit (Nuget Packages under sudo).
The second image shows the permissions on the file to be drwx-----

[Images redacted for internal only]


nagilson:
@dsplaisted We're not sure if we need this for RC2 yet.

nagilson:
We don't need it for rc2

Issue #28048, opened 2022-09-20
Code Coverage for RuntimeIdentifiers


nagilson:
Based on https://github.com/dotnet/sdk/issues/27985 which was resolved by https://github.com/dotnet/sdk/pull/28006, it would be helpful if we had a regression test with projects that leverage RuntimeIdentifiers but not a RuntimeIdentifier. In addition, we may be able to provide a better patch in the long-term now that the other builds are presumably unblocked. In discussion with @dsplaisted, we should likely move it to another task/target.

nagilson:
Huzzah, this can be closed

Issue #27985, opened 2022-09-17
NETSDK1191 error when building crossgen2.csproj with RC2 SDK


MichaelSimons:
This issue was encountered with source-build.  In source-build, you must be able to take the SDK from a source-build and use it to rebuild the source.  It was during this process the following build error was surfaced while prepping for RC2.

```
  /repos/rc2-artifacts/dotnet/sdk/7.0.100-rc.2.22466.1/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.RuntimeIdentifierInference.targets(169,5): error NETSDK1191: A runtime identifier for the property 'PublishReadyToRun' couldn't be inferred. Specify a rid explicitly. [/repos/tarball-boot/src/runtime/artifacts/source-build/self/src/src/coreclr/tools/aot/crossgen2/crossgen2.csproj]
```

The error check appears to be new to the RC2 sdk as I don't find it in the RC1 sdk - https://github.com/dotnet/sdk/commit/1fe65d2e6d2a539f357f016af4202218c6922255#diff-a34c0c3c174905dc5ce1e883006893f45b46b08fcac6ba51535c464cb952c6a2

From /repos/runtime/.dotnet/sdk/7.0.100-rc.2.22466.18/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.RuntimeIdentifierInference.targets

```
    <NETSdkError Condition="'$(PublishReadyToRun)' == 'true' and '$(RuntimeIdentifier)' == ''"
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishReadyToRun"/>
```

Here is a binlog from a source-build.
[runtime.boot.sourcebuild.zip](https://github.com/dotnet/runtime/files/9590742/runtime.boot.sourcebuild.zip)

Failing source-build CI - Microsoft internal link - https://dev.azure.com/dnceng/internal/_build/results?buildId=1996153&view=logs&j=05a957c8-6537-5d6d-69e3-d72786d35c31&t=8f2f8c4f-4a49-5a74-0521-8049d6bde132

This is a blocker for rc2.

MichaelSimons:
cc @marcpopMSFT

nagilson:
Thanks @am11, that sounds right to me. I'll take a look and fix it. Sorry to block it.

ViktorHofer:
[...]

PR #28484, opened 2022-10-11
RuntimeIdentifiers Code Coverage + Implicit RuntimeIdentifiers (not just runtimeIdentifier)

nagilson:
# Description

Resolves https://github.com/dotnet/sdk/issues/28048, follow up to https://github.com/dotnet/sdk/issues/27985 
We can imply a runtime identifier even if `RuntimeIdentifiers` is added and add a test to show that this feature works with not just one RID but `RuntimeIdentifiers`. The RID here is used to publish the project which will then use runtime identifiers. 

# Customer Impact

Low, this makes it so projects using RuntimeIdentifiers don't need to specify a RuntimeIdentifier, but probably they did anyways. It is a cleaner solution to the one we put out for source build a while ago, but both solutions fix the issue. We are raising this because we're making changes to this file anyways for a much more severe break, which is https://github.com/dotnet/sdk/pull/28628.

# Regression 

No.

# Risk

Somewhere between low and medium, closer to low than medium.

# Testing

We added a test to test the scenario that the previous fix... well, fixed, but we haven't run it through source-build who were the initial consumers that discovered the problem with how we handled RuntimeIdentifiers.

### DIFF
--- a/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Xml.Linq;
 using FluentAssertions;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.NET.TestFramework;
@@ -217,6 +218,37 @@ namespace Microsoft.NET.Publish.Tests
                 .Should()
                 .Pass();
 
+        }
+
+        [Fact]
+        public void It_Publishes_Successfully_With_RID_Requiring_Properties_And_RuntimeIdentifierS_but_no_RuntimeIdentifier()
+        {
+            var targetFramework = ToolsetInfo.CurrentTargetFramework;
+            var runtimeIdentifier = EnvironmentInfo.GetCompatibleRid(targetFramework);
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("HelloWorld")
+                .WithSource()
+                .WithProjectChanges(project =>
+                {
+                    var ns = project.Root.Name.Namespace;
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifiers", runtimeIdentifier));
+                    propertyGroup.Add(new XElement(ns + "PublishReadyToRun", "true"));
+                });
+
+            var buildCommand = new BuildCommand(testAsset);
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
+
+            var outputDirectory = buildCommand.GetOutputDirectory(targetFramework, runtimeIdentifier: runtimeIdentifier);
+
+            var publishCommand = new PublishCommand(testAsset);
+            publishCommand
+                .Execute()
+                .Should()
+                .Pass();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -218,5 +219,36 @@ public void DuplicateRuntimeIdentifiers()
...
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("HelloWorld")
+                .WithSource()
+                .WithProjectChanges(project =>
+                {
+                    var ns = project.Root.Name.Namespace;
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifiers", runtimeIdentifier));
+                    propertyGroup.Add(new XElement(ns + "PublishReadyToRun", "true"));
+                });
```
dsplaisted, line 237: Any reason to start with "HelloWorld" instead of something like this?

```C#
{
    OutputType = "Exe",
    TargetFrameworks = ToolsetInfo.CurrentTargetFramework
};

testProject.AdditionalProperties["RuntimeIdentifiers"] = runtimeIdentifiers;
testProject.AdditionalProperties["PublishReadyToRun"] = "true";
var testAsset = _testAssetsManager.CreateTestProject(testProject);
```

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -218,5 +219,36 @@ public void DuplicateRuntimeIdentifiers()
...
+
+            var buildCommand = new BuildCommand(testAsset);
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
+
+            var outputDirectory = buildCommand.GetOutputDirectory(targetFramework, runtimeIdentifier: runtimeIdentifier);
```
dsplaisted, line 245: If you're not using this directory for an assertion then we don't need to calculate it at all.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -218,5 +219,36 @@ public void DuplicateRuntimeIdentifiers()
...
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifiers", runtimeIdentifier));
+                    propertyGroup.Add(new XElement(ns + "PublishReadyToRun", "true"));
+                });
+
+            var buildCommand = new BuildCommand(testAsset);
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
```
dsplaisted, line 243: It's not required to run build before running publish, so for this scenario I would probably just delete this and only do a publish.

dsplaisted (approved):

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
nagilson: Nah, I thought it was the simplest way to make a test project but that's not true.

nagilson:
@dsplaisted Right now this is targeting 7.0.1xx which will go into GA IIRC; I know we weren't sure if we wanted to merge this into GA or not. Seems lower priority than the workloads issue you have atm. I'll wait to merge this until we get confirmation from someone.

Issue #28475, opened 2022-10-11
Trying to build with PublishAot=true on Mac ARM results in restore failure


agocke:
This occurs even if you explicitly specify `-r osx-x64`.

```
dotnet restore -r osx-x64
  Determining projects to restore...
/Users/angocke/tmp/aot/aot.csproj : error NU1101: Unable to find package runtime.osx-arm64.Microsoft.DotNet.ILCompiler. No packages exist with this id in source(s): nuget.org
  Failed to restore /Users/angocke/tmp/aot/aot.csproj (in 132 ms).
```

agocke:
fyi @LakshanF 

I'll take a look at this -- because it's blocking `dotnet build`, I'd consider this servicing-worthy.

agocke:
This looks like it's fixed.

PR #28495, opened 2022-10-12
Allow building with PublishAot set

agocke:
Fixes https://github.com/dotnet/sdk/issues/28475

The current PublishAot behavior is to attempt to restore the runtime pack for NativeAOT immediately. This doesn't work for osx-arm64, since there's no runtime pack for that RID. This causes even restore to fail, when it should succeed (as should build).

The new behavior is to only attempt to restore the runtime pack when publish is actually being run. An error message is also printed if an unsupported RID is used, although an "escape hatch" property is also provided to skip that error.


LakshanF (approved):
LGTM! ITs good that users can bypass the rid check in the SDK if needed when ILC packages with new rids get added

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
```diff
@@ -81,6 +81,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <PropertyGroup>
       <_NuGetRestoreSupported Condition="('$(Language)' == 'C++' and '$(_EnablePackageReferencesInVCProjects)' != 'true')">false</_NuGetRestoreSupported>
+      <_AotEnabled Condition="('$(PublishAot)' == 'true') and ('$(_IsPublishing)' == 'true')">true</_AotEnabled>
```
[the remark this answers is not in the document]
nagilson: FYI `_IsPublishing` is only for CLI scenarios ATM, we are discussing if we can get it to work in VS Publishing today. We don't think we can get it to work for `t:/Publish` and plan on evaluating it in a Monday partner sync or in a separate call, I was planning on adding you @agocke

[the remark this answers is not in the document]
agocke: Recommendation on what I should use instead?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -82,6 +82,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     </PropertyGroup>
   </Target>
 
+  <PropertyGroup>
+    <_AotSupportedRids>linux-x64;linux-arm64;linux-musl-x64;linux-musl-arm64;windows-x64;windows-arm64</_AotSupportedRids>
```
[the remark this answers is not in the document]
agocke: That list isn't correct -- it includes the macos RIDs. I need the real list here, as I want to provide an error for mac, not just let it go.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -111,10 +115,15 @@ Copyright (c) .NET Foundation. All rights reserved.
                             '$(EnableCompressionInSingleFile)' == 'true' And
                             '$(SelfContained)' != 'true'"
                  ResourceName="CompressionInSingleFileRequiresSelfContained" />
+    <NETSdkError Condition="'$(PublishAot)' == 'true' And
+                            '$(_SkipAotSupportedRidCheck)' != 'true' And
+                            $(_AotSupportedRids.Contains('$(RuntimeIdentifier)')) != 'true'"
```
AntonLapounov, line 120: This will allow identifiers like `linux-arm` and arbitrary substrings. You might want to do `Contains(';$(RuntimeIdentifier);')` instead.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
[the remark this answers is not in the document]
nagilson: I'm sorry to say I don't have any immediate recommendations, I am curious as to how `PublishTrimmed`, single file, aot, etc currently work. The dilemma we're facing right now is detecting if we are publishing early in evaluation seems to be incompatible with MSBuild as it currently exists, because of custom Publish targets that call publish later on, --no-build publishes, etc. It seems like it should be so simple and something we really want; we have been going back and forth on this for a few months now!

agocke:
Talking with @sbomer offline, maybe the right answer is just to not fail restore at all. We can give a better error message later.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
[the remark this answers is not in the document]
agocke: OK, this logic is getting convoluted, but I think this split is what we want now.

We want the full list in restore, so that restore will try to grab it.

But we want the restricted list here, so that you get a clear error when you're doing something unsupported.

I might be missing something though.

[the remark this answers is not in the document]
agocke: So, the experience I think we want (based on previous changes in this code) is:

1. PublishAot on Mac = error unsupported.
2. PublishAot on Mac w/ explicit bypass = works.

If that's true, the list of acceptable RIDs in (1) does not include osx, but it does in (2). I don't see how those could be the same list.

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -362,11 +362,9 @@ var runtimeRequiredByDeployment
             
             if (AotEnabled)
             {
-                if (!AddAotOrR2RRuntimePackage(AotPackageType.ILCompiler, _normalizedTargetFrameworkVersion, packagesToDownload))
-                {
-                    Log.LogError(Strings.AotNoValidRuntimePackageError);
-                    return;
-                }
+                // Try to download the package. If it fails, we don't error out, as failing a restore
```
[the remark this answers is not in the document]
agocke: Yup, I'll adjust the comment.

agocke:
Yup, the decision in this PR is that restore will never fail, even if the eventual publish certainly will.

I think this PR now represents a workable solution. Publish will, by default, check that the RID is supported and error if it is not.

If the check is bypassed, publish will attempt to use a downloaded package (which will be downloaded during restore if we've added it to the list of available packages, or if the user adds a new `KnownILCompilerPack`).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -111,10 +115,15 @@ Copyright (c) .NET Foundation. All rights reserved.
                             '$(EnableCompressionInSingleFile)' == 'true' And
                             '$(SelfContained)' != 'true'"
                  ResourceName="CompressionInSingleFileRequiresSelfContained" />
+    <NETSdkError Condition="'$(PublishAot)' == 'true' And
+                            '$(_SkipAotSupportedRidCheck)' != 'true' And
```
jkotas, line 119: We have similar property in 
https://github.com/dotnet/runtime/blob/cb6fb60dfc253bc13655fafa15686cb864fd0ea2/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets#L58 . 

Would it make sense to use one property to suppress all these supported checks?

- Rename `_SkipAotSupportedRidCheck` -> `SkipAotSupportedCheck` here
- Rename `DisableUnsupportedError` to `SkipAotSupportedCheck` in dotnet/runtime.

agocke: That makes sense, but I'd like to do that as a separate change, as I intend for this to go into .NET 7 as a servicing fix.

agocke:
@sbomer Anything else here?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
jkotas: If you agree that we should use the same property name in all places, the property should be renamed in this servicing fix.

src/Layout/toolset-tasks/OverrideAndCreateBundledNETCoreAppPackageVersion.cs
```diff
@@ -123,6 +123,12 @@ void CheckAndReplaceAttribute(XAttribute attribute)
                 crossgen2Rids.Value += ";osx-x64";
             }
 
+            // Currently has too many RIDs. We only want to include the RIDs of actual packages we
+            // produce, as if don't try to download a missing package then restore will succeed and
+            // we can give a better error during publish that the given RID is not supported.
+            var ilcompilerRids = itemGroup.Elements(ns + "KnownILCompilerPack").First().Attribute("ILCompilerRuntimeIdentifiers");
+            ilcompilerRids.Value = "linux-x64;linux-musl-x64;linux-arm64;linux-musl-arm64;win-x64;win-arm64;osx-x64";
```
dsplaisted, line 130: Ideally we shouldn't hard-code the supported RuntimeIdentifiers in a C# task.  What's the reason for the mismatch between the values coming from the item and what's actually supported?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -82,6 +82,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     </PropertyGroup>
   </Target>
 
+  <PropertyGroup>
+    <_AotSupportedRids>;linux-x64;linux-arm64;linux-musl-x64;linux-musl-arm64;win-x64;win-arm64;</_AotSupportedRids>
```
dsplaisted, line 86: Usually when we add support for new RuntimeIdentifiers that ends up getting authored in the installer repo and flowing into items such as the KnownFrameworkReference or other items in Microsoft.NETCoreSdk.BundledVersions.props.  Should we do that for this?  How would we support it if we add support for new RuntimeIdentifiers for AoT, and the list of supported RIDs depends on the target framework?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -111,10 +115,15 @@ Copyright (c) .NET Foundation. All rights reserved.
                             '$(EnableCompressionInSingleFile)' == 'true' And
                             '$(SelfContained)' != 'true'"
                  ResourceName="CompressionInSingleFileRequiresSelfContained" />
+    <NETSdkError Condition="'$(PublishAot)' == 'true' And
+                            '$(_SkipAotSupportedRidCheck)' != 'true' And
+                            $(_AotSupportedRids.Contains(';$(NETCoreSdkRuntimeIdentifier);')) != 'true'"
```
dsplaisted, line 120: Instead of checking whether a string contains the RuntimeIdentifier, shouldn't we be checking to see whether there is a valid `HostILCompilerPack` or `TargetILCompilerPack` instead for this error?

src/Layout/toolset-tasks/OverrideAndCreateBundledNETCoreAppPackageVersion.cs
agocke: The mismatch is that we were trying to avoid blocking people who wanted to use unsupported, but available packages.

But being too broad here actually causes problems because attempting to download a missing package blocks restore/build.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
agocke: Note that the list above (which is the supported list that flows from /installer), is not *quite* the same as the supported rids here, namely that osx-x64 is in the "to download" list but not in the "supported" list.

jkotas: The only thing that matters for unsupported architectures is that there is some way for people who want to give it try. It does not need to be pretty. It is ok if it requires extra boilerplate like explicit reference to the osx package.

Does the new SkipAotSupportedCheck property enable that? If yes, I think it would be fine to get rid of this special-casing for available, but officially unsupported architectures.

src/Layout/toolset-tasks/OverrideAndCreateBundledNETCoreAppPackageVersion.cs
kant2002: We may want to have `browser-wasm` here for NativeAOT-LLVM support.

PR #28515, opened 2022-10-12
Add the Microsoft* package and see if it's small enough to be downloaded.

marcpopMSFT:
We hit a size limit last time but this is slightly smaller so maybe it'll work. we're still seeing a lot of timeouts so it's worth a shot.


marcpopMSFT:
Sigh, still too large: {"Id":"UnzipError","Message":"Archive File too large. Unable to unpack C:\\h\\w\\A1D1094B\\w\\SDKTestPackages.Microsoft.zip.partial in C:\\h\\w\\A1D1094B\\p\\d/.nuget. Execution may be compromised."},

PR #28541, opened 2022-10-13
Trusted roots:  2022-09 CTL

dtivel:
Cherry-picking https://github.com/dotnet/sdk/pull/28512 into release/7.0.1xx.

 ## Servicing Template. 
 
 ## Description
 This PR in combination with https://github.com/NuGet/NuGet.Client/pull/4850 updates CTLs for NuGet signed package verification on Linux and macOS to use 2 separate fallback certificate bundles instead of 1.

Prior to this change, NuGet would use a single fallback certificate bundle which contained root certificates valid for both code signing and timestamping. Roots valid for only code signing or only timestamping were not in the certificate bundle because a consumer had no way of knowing which certificates were valid for which purpose(s).

Whatever trust anchors we use must be valid for CS and TS. Actually, it's a bit more nuanced in that whatever set of trust anchors we use for purpose X must be valid for purpose X. The set of trust anchors valid for CS and the set of trust anchors valid for TS intersect but do not necessarily overlap. Using a trust anchor that is only valid for TS (or TLS, for that matter) to verify a CS certificate would be incorrect.

## Customer Impact 

Not having separate CTLs fails the verification of a set of packages on nuget.org. The expectation is that *all* packages form NuGet.org pass validation. This change allows verification to enabled as opt-in and have confidence that it'll work.

## Regression

No

## Risk

Low - Feature not enabled by default.

## Link the PR to the original issue and to the PR to main

NuGet side: https://github.com/NuGet/Home/issues/12033 and https://github.com/NuGet/Home/issues/12027

Main PR: https://github.com/dotnet/sdk/pull/28512

 ## Packaging impact

None.

# Ref pack impact

None.


nkolev92:
I have updated this PR to be used for servicing approval. 

This PR has a companion PR NuGet side, which needs to flow together.

nkolev92 (approved):

PR #28552, opened 2022-10-13
The loc build changes didn't flow to main so going directly.

marcpopMSFT:


nagilson (approved):
Marc, may you please confirm the target branch for this is supposed to be `main` and not `7.0.2`? I assume you are trying to get localization changes from internal 7.0.2 to go into main but wanted to be sure this wasn't meant to trigger for 7.0.2xx.

PR #28559, opened 2022-10-13
Change the build machine to have a larger hard drive

marcpopMSFT:


ghost (approved):
I think this looks right. I think we'll have some confirmation on the public side during the PR

PR #28596, opened 2022-10-14
Disable the windows SDK automated references for c++

### LINKED ISSUE #26203: Don't reference Microsoft.Windows.SDK.NET.Ref for C++/CLI projects
C++/CLI projects always have a target platform of some version of Windows 10 or higher.  This is causing the SDK to add framework references for Microsoft.Windows.SDK.NET.Ref, which provides access to Windows WinRT APIs via CSWinRT-generated shims.  However, this wasn't intended for, and may not work with C++/CLI.  So we should probably disable this implicit reference for C++/CLI.

Probably this means removing the second condition from this line: https://github.com/dotnet/sdk/blob/b7e2f112f78ccaa1d779fb3769aa1b9a2eaccc93/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Windows.targets#L22

marcpopMSFT:
Even with package references enabled in c++ projects, the windows references should not be included by default.

Fixes #26203


dsplaisted (approved):
Do we have. any tests that cover C++ PackageReference projects?  Did anyone manually test this at all?

marcpopMSFT:
I don't think our c++/cli tests are currently even enabled. I'm having a difficult time getting them to run or to manually test this. Any suggestions?

dsplaisted:
They've been disabled for a long time (see https://github.com/dotnet/sdk/issues/11008).  I think you should be able to enable them locally and try running them in VS to get some projects that work.

marcpopMSFT:
@dsplaisted I turned on the existing cppcl itests and added a new one. Want to review the new test?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
```diff
@@ -40,7 +45,7 @@ public void It_builds_and_runs()
             var exe = Path.Combine( //find the platform directory
                 new DirectoryInfo(Path.Combine(testAsset.TestRoot, "CSConsoleApp", "bin")).GetDirectories().Single().FullName,
                 "Debug",
-                "net5.0",
+                ToolsetInfo.CurrentTargetFramework,
```
dsplaisted, line 48: This should either be hardcoded at `net7.0` to match the test asset, or the test asset should be updated to use CurrentTargetFramework.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
```diff
@@ -166,5 +191,79 @@ private void ChangeTargetFramework(string projectPath, XDocument project, string
...
+//    <ItemGroup>
+//{propertiesElements}
+//    </ItemGroup>
+//    <WriteLinesToFile
+//      File='$(BaseIntermediateOutputPath)\$(Configuration)\$(TargetFramework)\PropertyValues.txt'
+//      Lines='@(LinesToWrite)'
+//      Overwrite='true'
+//      Encoding='Unicode'
+//      />
+//  </Target>";
```
dsplaisted, line 234: It might be easier to read and less redundant if you did an `XElement.Parse` on a string literal with the whole target instead of building it up in code.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
```diff
@@ -52,7 +57,27 @@ public void It_builds_and_runs()
...
+        [FullMSBuildOnlyFact]
+        public void It_builds_and_runs_with_package_reference()
+        {
+            var targetFramework = ToolsetInfo.CurrentTargetFramework + "-windows";
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("NetCoreCsharpAppReferenceCppCliLib")
+                .WithSource()
+                .WithProjectChanges((projectPath, project) => ConfigureProject(projectPath, project, "NewtonSoft.Json","13.0.1", targetFramework, new string[] { "_EnablePackageReferencesInVCProjects" , "IncludeWindowsSDKRefFrameworkReferences" }));
+
+            // build projects separately with BuildProjectReferences=false to simulate VS build behavior
```
dsplaisted, line 69: Is this comment supposed to be here?  I don't see anything setting BuildProjectReferences to false.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
```diff
@@ -52,7 +57,27 @@ public void It_builds_and_runs()
...
+            new BuildCommand(testAsset, "NETCoreCppCliTest")
+                .Execute("-p:Platform=x64", "-p:EnableManagedpackageReferenceSupport=true", "/bl")
```
dsplaisted, line 71: Nit: Probably don't want to record a binlog except when we're running / debugging a test locally.

```suggestion
                .Execute("-p:Platform=x64", "-p:EnableManagedpackageReferenceSupport=true")
```

dsplaisted (approved):

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
marcpopMSFT: The problem I found is that XElement.parse ends up creating an empty namespace (xmlns="") which then breaks the build. I've tried a few different options but I end up setting the namespace on the top level target but still have the empty namespace on the sub nodes.  I found some examples for removing the namespace from every entry but not one for setting it.  Any ideas?

marcpopMSFT: ok, found a way to do clean up the namespaces and cleaned up the code to do it in the more readable way.

PR #28608, opened 2022-10-15
Ensure workload installers can be read by other users

### LINKED ISSUE #28450: [NETSDKE2E]Error happens "Workload installation failed: Access to the path 'C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.MonoAOTCompiler.Task.Msi.x64\7.0.0-rtm.22507.1\msi.json' is denied." when installing workload using another account.
1.Install RTM SDK(7.0.100-rtm.22510.12) and then install workload with below command in CLI
2.dotnet workload install wasm-tools --skip-manifest-update
3. **add another domain account as a admin into current machine**
image

4. Login with anther domain account and then install workload with below command in CLI
dotnet workload install maccatalyst --skip-manifest-update
**Below error happens**
Workload installation failed: Access to the path 'C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.MonoAOTCompiler.Task.Msi.x64\7.0.0-rtm.22507.1\msi.json' is denied.
image

Note:
1. Folder 7.0.0-rtm.22507.1 permission as below.
image

image

2. Run cmd as an administrator and then install workload successfully without any issue.

joeloff:
For MSI based installs of workloads, the MSI and its manifest are extracted to a random temporary folder and then moved into the package cache under Program Data. Checking the cache for the installer happens on the client (non-elevated) side. When one user installed a workload and another user logs on and tries to update installed workloads, the check fails because the second user has no read permissions against the file in the cache.

Because the file is extracted by the client and then copied by the server, the first user permissions are attached to the file.

For example, when the file is moved, we end up with descriptor that includes the original user's SID: `O:S-1-5-21-1004336348-1177238915-682003330-512G:DUD:P(A;;FA;;;SY)(A;;FA;;;BA)(A;;FA;;;S-1-5-21-1004336348-1177238915-682003330-512)`

We want the owner to be the built-in administrators and restrict built-in users and everyone to only have read and execute permissions: `O:BAG:DUD:AI(A;;0x1200a9;;;WD)(A;ID;FA;;;SY)(A;ID;FA;;;BA)(A;ID;0x1200a9;;;BU)`

Fixes #28450

### DIFF
--- a/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
+++ b/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
@@ -154,8 +154,8 @@ namespace Microsoft.DotNet.Installer.Windows
                 string cachedMsiPath = Path.Combine(packageDirectory, Path.GetFileName(msiPath));
                 string cachedManifestPath = Path.Combine(packageDirectory, Path.GetFileName(manifestPath));
 
-                MoveFile(manifestPath, cachedManifestPath);
-                MoveFile(msiPath, cachedMsiPath);
+                MoveAndSecureFile(manifestPath, cachedManifestPath);
+                MoveAndSecureFile(msiPath, cachedMsiPath);
             }
             else if (IsClient)
             {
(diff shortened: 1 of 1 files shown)


joeloff:
@eerhardt would you mind taking a peek at this change?

nagilson (commented):
I don't know enough to give this a proper review, but a few comments:

1) The file is moved before the permissions are set. This should be fine if it has lower or the same permissions via move, but is there slight potential for a Race condition where the permissions were not set but the file had been moved? If its the later maybe its so late we dont care. 

2) `O:BAG:DUD:AI(A;ID;FA;;;SY)(A;ID;FA;;;BA)(A;ID;0x1200a9;;;BU)(A;ID;0x1200a9;;;WD)` is really difficult for me to digest, could there instead be something like 
const string ADMIN_READ_PERMISSION = 'O:BAG:DUD:AI
const string ADMIN_WRITE_PERMISSION = '(A;ID;FA;;;SY)`
...
const string CORRECT_PERMISSIONS = ADMIN_READ_PERMISSION + ADMIN_WRITE_PERMISSION + ...; 
SetSecurityDescriptorSddlForm(CORRECT_PERMISSIONS)?

Probably a pain but it would be super helpful imo.

joeloff:
> I don't know enough to give this a proper review, but a few comments:

1. The file has to be moved before the permissions are set because it is moved from the client (non-elevated) to a privileged location and will carry the user's SID when moved which we don't want because the cache is machine wide and should not be associated with a specific user. I suppose there is potential for race conditions, e.g. if two users on the same machine tries to install the same workload and we try to cache it at the same time. That's a very unlikely scenario. I don't think it will break anything and users will encounter other issues like running concurrent installs which will fail since MSIs cannot execute in parallel.
2. I moved everything to a const with doc comments that explain everything in more detail. Naming the constants for the ACE items would become extremely verbose I think to spell out the individual fields we're setting.

nagilson:
Thanks @joeloff for following up, that resolves my concerns and it's much easier to understand now. Would appreciate if someone could give a final approval.

joeloff:
@eerhardt Any other concerns here?

eerhardt (approved):
Looks good to me.

PR #28627, opened 2022-10-17
Return workload manifests in defined, controllable order

### LINKED ISSUE #22358: Workloads should be able to control the relative import order of .props and .targets
Right now the order that the WorkloadManifest.targets and AutoImport.props files are imported is not defined.  However, workloads may need to have dependencies on what the evaluation import ordering of these files are.

An example of this is that the Maui workloads will probably want to remove implicit usings that are added by the iOS and Android workloads (since they all define `Button`, for example).

So we should probably define what the import order is by default (ie alphabetical), and add a way to allow the workload manifests to specify that something should be evaluated after something else.

### LINKED ISSUE #28607: WorkloadManifest.targets should be imported in a pre-determined order for `7.0`
With the latest `7.0.0` sdk, the `WorkloadManifest.targets` for the `wasm-tools*` packs are imported in a different order from earlier which breaks the wasm workloads. We started hitting this with the [stable branding change](https://github.com/dotnet/runtime/pull/76849) in `dotnet/runtime`.

In the working case:
- `microsoft.net.workload.mono.toolchain.net6`, `microsoft.net.workload.mono.toolchain.net7`, `microsoft.net.workload.emscripten.net6`, `microsoft.net.workload.emscripten.net7`

In the broken case:
- `microsoft.net.workload.emscripten.net7`,`microsoft.net.workload.emscripten.net6`, `microsoft.net.workload.mono.toolchain.net7`, `microsoft.net.workload.mono.toolchain.net6`

Mono.Toolchain.net*, and Emscripten*'s  WorkloadManifest.targets are sensitive to the order, so we have a correct order specified in `IncludedWorkloadManifests.txt`.

We have been trying to fix this since the issue was discovered, but the solutions are non-trivial, and require a lot of non-trivial testing. It would be preferable to enforce the order in `IncludedWorkloadManifests.txt`. This would mean that all the testing that has been done, till we switched to stable branding, would still be relevant.
This change can be restricted to `7.0` .

cc @lewing @steveisok @marcpopMSFT @jonathanpeppers

dsplaisted:
# Description

The WASM workload stopped working when we switched to stable branding.  This is because the order in which we imported workload manifest targets was not guaranteed and ended up changing when the branding changed, but the WASM workload manifest targets did depend on the order they were imported.

To fix this, we will return workload manifests in a stable order.  Manifests in the IncludedWorkloadManifests.txt file will be first, and in the same order they appear in that file.  Then the rest of the manifests (if any) will be returned in (ordinal case-insensitive) alphabetical order.

Fixes #22358
Fixes #28607

Alternatively, we could have changed the WASM workload manifest targets to be resilient to being imported in an arbitrary order.  That would have been a lot more risky: The change itself would be more complex, it would be harder to test, and the time to flow the change through the build graph would be longer.

# Customer Impact

Without this change (or another more risky change to the WASM workload manifest targets), the WASM workload won't work in stable versioned .NET SDKs.

# Regression

Yes, this worked in earlier prereleases, but dropping the prerelease specifier from the version triggered this regression.

# Risk

Low

# Testing

Added unit test.  Planning to validate E2E scenario with WASM manually.

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.IO;
 using System.Linq;
 using Microsoft.DotNet.Cli;
@@ -17,7 +18,7 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
         private readonly string [] _manifestDirectories;
         private static HashSet<string> _outdatedManifestIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "microsoft.net.workload.android", "microsoft.net.workload.blazorwebassembly", "microsoft.net.workload.ios",
             "microsoft.net.workload.maccatalyst", "microsoft.net.workload.macos", "microsoft.net.workload.tvos" };
-        private readonly HashSet<string>? _knownManifestIds;
+        private readonly Dictionary<string, int>? _knownManifestIdsAndOrder;
 
         public SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVersion, string? userProfileDir)
             : this(sdkRootPath, sdkVersion, Environment.GetEnvironmentVariable, userProfileDir)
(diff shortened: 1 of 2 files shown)


marcpopMSFT (approved):

nagilson (approved):

radical:
@dsplaisted how do I test this? Our wasm tests use the installer via `dotnet-install.sh`, and then install `wasm-tools` workload on that.

lewing (approved):

radical:
I tested this locally by copying over `lib/net6.0/Microsoft.NET.Sdk.WorkloadManifestReader.dll`, and `lib/net6.0/Microsoft.NET.Sdk.WorkloadMSBuildSdkResolver.dll` from a locally build of this branch, to `<dotnet>/sdk/7.0.100`. And the logs show the expected import order.

PR #28674, opened 2022-10-20
Make PublishAot an optional feature because Mono does not support it.

### LINKED ISSUE #3076: Missing Microsoft.DotNet.ILCompiler package on s390x
When running an iterated source-build on s390x (i.e. using the output of the previous source-build as input set of package instead of the tarball), building the SDK fails due to a missing package `Microsoft.DotNet.ILCompiler`.

On Intel, this package is generated by the `runtime` repository, but on s390x this does not happen due to a check in `eng/Subsets.props`:
```
        <ProjectToBuild Condition="'$(NativeAotSupported)' == 'true' and '$(RuntimeFlavor)' != 'Mono'" Include="$(InstallerProjectRoot)\pkg\projects\nativeaot-packages.proj" Category="packs" />
```
Both the `NativeAotSupported` check and the `RuntimeFlavor` check fail on s390x.

Fixing this in the runtime build seems non-trivial since the actual artifacts to be packaged in that package are only created as part of the CoreCLR build, which doesn't happen on s390x.

Also, given that this package - while it itself seems to be platform-agnostic - simply adds support to the SDK to pull in various other packages needed for the NativeAOT build, and those are not available on s390x either, it seems this package shouldn't be really needed on s390x.   We accept that the NativeAOT feature will not be available on the platform.

I thought this would be a simple matter of making the `Microsoft.DotNet.ILCompiler` line in `src/Layout/redist/targets/BundledSdks.targets` in the SDK repo conditional on the platform.  But it turns out this isn't easily possible as the SDK repo build as part of the source-build process doesn't appear to even be aware which platform is being targeted.  None of the `Architecture`, `TargetRid` or `Platform` properties (or any other property I could identify) are set to indicate a s390x target architecture.

We could disable the line for building with source-build generally, but I understand this is undesirable, as there is supposed to be NativeAOT support on Intel (even though the compiler packages aren't build with source-build, they would be pulled in from nuget at runtime if the SDK support is available).

@crummel any suggestions on how to best address this problem?

FYI - @tmds @omajid @janani66

tmds:
# Customer impact

The sdk assumes native aot is always available. This is not the case when .NET 7 is source-built with mono runtime for architectures like s390x and ppc64le.

# Testing

The builds for RHEL and Fedora already include this patch.

# Risk

Low, as it is already used.

### DIFF
--- a/eng/SourceBuild.props
+++ b/eng/SourceBuild.props
@@ -8,6 +8,7 @@
     <PropertyGroup>
       <InnerBuildArgs>$(InnerBuildArgs) /p:Projects="$(InnerSourceBuildRepoRoot)\source-build.slnf"</InnerBuildArgs>
       <InnerBuildArgs>$(InnerBuildArgs) /p:UseSharedCompilation=false</InnerBuildArgs>
+      <InnerBuildArgs Condition="'$(SourceBuildUseMonoRuntime)' == 'true'">$(InnerBuildArgs) /p:NativeAotSupported=false</InnerBuildArgs>
     </PropertyGroup>
   </Target>
 
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -836,7 +836,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1182: "}</comment>
   </data>
   <data name="AotNoValidRuntimePackageError" xml:space="preserve">
-    <value>NETSDK1183: Unable to optimize assemblies for Ahead of time compilation: a valid runtime package was not found. Either set the PublishAot property to false, or use a supported runtime identifier when publishing. When targeting .NET 7 or higher, make sure to restore packages with the PublishAot property set to true.</value>
+    <value>NETSDK1183: Unable to optimize assemblies for ahead-of-time compilation: a valid runtime package was not found. Either set the PublishAot property to false, or use a supported runtime identifier when publishing. When targeting .NET 7 or higher, make sure to restore packages with the PublishAot property set to true.</value>
     <comment>{StrBegin="NETSDK1183: "}</comment>
   </data>
   <data name="TargetingPackNotRestored_TransitiveDisabled" xml:space="preserve">
@@ -876,7 +876,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1192: "}</comment>
   </data>
   <data name="AotNotSupported" xml:space="preserve">
-    <value>NETSDK1193: The SDK does not support Ahead of time compilation. Set the PublishAot property to false.</value>
+    <value>NETSDK1193: The SDK does not support ahead-of-time compilation. Set the PublishAot property to false.</value>
       <comment>{StrBegin="NETSDK1193: "}</comment>
   </data>
 </root>
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -875,4 +875,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1192: Targeting .NET 7.0 or higher in Visual Studio 2022 17.3 is not supported.</value>
     <comment>{StrBegin="NETSDK1192: "}</comment>
   </data>
+  <data name="AotNotSupported" xml:space="preserve">
+    <value>NETSDK1193: The SDK does not support Ahead of time compilation. Set the PublishAot property to false.</value>
+      <comment>{StrBegin="NETSDK1193: "}</comment>
+  </data>
 </root>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -182,9 +182,6 @@ Copyright (c) .NET Foundation. All rights reserved.
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishAot"/>
 
-    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(PublishAotSupported)' != 'true'"
-                ResourceName="AotNotSupported" />
-
     <!-- End of implicit RID resolver checks.-->
     <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(UseAppHost)' != 'true' and '$(_RuntimeIdentifierUsesAppHost)' == 'true'"
                  ResourceName="CannotUseSelfContainedWithoutAppHost" />
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -182,6 +182,9 @@ Copyright (c) .NET Foundation. All rights reserved.
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishAot"/>
 
+    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(PublishAotSupported)' != 'true'"
+                ResourceName="AotNotSupported" />
+
     <!-- End of implicit RID resolver checks.-->
     <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(UseAppHost)' != 'true' and '$(_RuntimeIdentifierUsesAppHost)' == 'true'"
                  ResourceName="CannotUseSelfContainedWithoutAppHost" />
--- a/src/Layout/redist/targets/BundledSdks.targets
+++ b/src/Layout/redist/targets/BundledSdks.targets
@@ -5,6 +5,6 @@
     <BundledSdk Include="FSharp.NET.Sdk" Version="1.0.4-bundled-0100" />
     <BundledSdk Include="Microsoft.Docker.Sdk" Version="1.1.0" />
     <BundledSdk Include="Microsoft.NET.ILLink.Tasks" Version="$(MicrosoftNETILLinkTasksPackageVersion)" />
-    <BundledSdk Include="Microsoft.DotNet.ILCompiler" Version="$(MicrosoftDotNetILCompilerPackageVersion)" />
+    <BundledSdk Include="Microsoft.DotNet.ILCompiler" Version="$(MicrosoftDotNetILCompilerPackageVersion)" Condition="'$(NativeAotSupported)' != 'false'" />
   </ItemGroup>
 </Project>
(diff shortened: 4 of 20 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -178,6 +178,9 @@ Copyright (c) .NET Foundation. All rights reserved.
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishAot"/>
 
+    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(PublishAotSupported)' != 'true'"
```
tmds, line 181: I've put this with the rid checks, but probably there is a better place for it.

dsplaisted (approved):

agocke (approved):
LGTM

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -875,4 +875,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1192: Targeting .NET 7.0 or higher in Visual Studio 2022 17.3 is not supported.</value>
     <comment>{StrBegin="NETSDK1192: "}</comment>
   </data>
+  <data name="AotNotSupported" xml:space="preserve">
+    <value>NETSDK1193: The SDK does not support Ahead of time compilation. Set the PublishAot property to false.</value>
```
AntonLapounov, line 879: Nit: Why is the initial `A` in `Ahead` capitalized?  It is usually spelled "ahead-of-time compilation", (e.g., in [Wikipedia](https://en.wikipedia.org/wiki/Ahead-of-time_compilation)).

tmds: I took the spelling from

https://github.com/dotnet/sdk/blob/1624c6a446d875622c3e45b46d374db7ff027b6c/src/Tasks/Common/Resources/Strings.resx#L839

tmds: @AntonLapounov should I update both values?

crummel (approved):
LGTM from the source-build perspective.

tmds:
@agocke @crummel https://github.com/dotnet/sdk/pull/28690 has invalidated my PR.
Should I target the 7.0 branch instead?

tmds:
> Should I target the 7.0 branch instead?

What branch should I target? 7.0.1xx? Will that get merged to 7.0.2xx?

For the `main` branch, because https://github.com/dotnet/sdk/pull/28690 removes the `BundledSdk`, we should be able to compile for mono runtime with the branch as-is.
There won't be a nice error message when a user tries to publish aot, but we can look into that later.
Should I make a PR against main that adds `NETSDK1193` so the identifier doesn't get used for anything else?

MichaelSimons:
> What branch should I target? 7.0.1xx? Will that get merged to 7.0.2xx?

I would say 7.0.1xx.  Changes to 7.0.1xx will get auto merged (via PR) to the higher feature bands (e.g. 7.0.2xx)

tmds:
I've changed this to target 7.0.1xx.

@agocke when CI is green, can you merge this?

For `main` we don't need anything atm, except perhaps 'reserving' the `NETSDK1193` error. Should I make another PR for that?

agocke:
@tmds This will need to go through servicing approval for 7.0. Have you already done that?

tmds:
> @tmds This will need to go through servicing approval for 7.0. Have you already done that?

@agocke usually maintainers take care of this. I've updated the top-level comment with some additional information.

tmds:
Is this good to merge?

agocke:
Unfortunately couldn't get to it in Tactics today. I'll try to do this over email.

src/Tasks/Common/Resources/Strings.resx
AntonLapounov: Sorry, missed the question. I would update both values.

tmds: I've replaced it everywhere where the language is English. I left the other ones untouched.
```
Strings.es.xlf:        <target state="translated">NETSDK1183: No se pueden optimizar los ensamblados para la compilación Ahead of time: no se ha encontrado un paquete en tiempo de ejecución válido. Establezca la propiedad PublishAot en false o use un identificador de tiempo de ejecución compatible al publicar. Cuando el destino sea .NET 7 o una versión posterior, asegúrese de restaurar los paquetes con la propiedad PublishAot establecida en true.</target>
Strings.fr.xlf:        <target state="translated">NETSDK1183: Impossible d'optimiser les assemblys pour la compilation Ahead of time : un package d'exécution valide n'a pas été trouvé. Définissez la propriété PublishAot sur false ou utilisez un identificateur d'exécution pris en charge lors de la publication. Lorsque vous ciblez .NET 7 ou supérieur, assurez-vous de restaurer les packages avec la propriété PublishAot définie sur true.</target>
Strings.it.xlf:        <target state="translated">NETSDK1183: non è possibile ottimizzare gli assembly per la compilazione Ahead Of Time perché non è stato trovato alcun pacchetto di runtime valido. Impostare la proprietà PublishAot su false oppure usare un identificatore di runtime supportato durante la pubblicazione. Quando si usa .NET 7 o versioni successive, assicurarsi di ripristinare i pacchetti con la proprietà PublishAot impostata su true.</target>
Strings.ja.xlf:        <target state="translated">NETSDK1183: Ahead Of Time コンパイル用にアセンブリを最適化できません: 有効なランタイム パッケージが見つかりませんでした。PublishAot プロパティを false に設定するか、公開時に、サポートされているランタイム識別子を使用してください。.NET 7 以降を対象とする場合は、必ず PublishAot プロパティを true に設定してパッケージを復元してください。</target>
Strings.ko.xlf:        <target state="translated">NETSDK1183: Ahead of Time 컴파일을 위해 어셈블리를 최적화할 수 없습니다. 유효한 런타임 패키지를 찾을 수 없습니다. PublishAot 속성을 false로 설정하거나 게시할 때 지원되는 런타임 식별자를 사용하세요. .NET 7 이상을 대상으로 하는 경우 PublishAot 속성이 true로 설정된 패키지를 복원해야 합니다.</target>
Strings.ru.xlf:        <target state="translated">NETSDK1183: не удалось оптимизировать сборки для компиляции Ahead Of Time: не найден допустимый пакет среды выполнения. Задайте для свойства PublishAot значение false либо используйте поддерживаемый идентификатор среды выполнения при публикации. При выборе .NET 7 или более поздней версии в качестве цели восстановите пакеты со свойством PublishAot со значением true.</target>
```

ayakael:
@tmds Have you encountered any errors building SDK with this patch. Our Alpine build on `armv7` fails when building on itself:
```
    /home/buildozer/aports/community/dotnet7-build/src/dotnet-v7.0.100-rtm.22521.12/src/sdk/artifacts/source-build/self/src/src/Layout/redist/targets/sdks/sdks.csproj : error NU1101: Unable to find package Microsoft.DotNet.ILCompiler. No packages exist with this id in source(s): prebuilt, previously-source-built, reference-packages, source-built
    ##vso[task.logissue type=error;sourcepath=/home/buildozer/aports/community/dotnet7-build/src/dotnet-v7.0.100-rtm.22521.12/src/sdk/artifacts/source-build/self/src/src/Layout/redist/targets/sdks/sdks.csproj;linenumber=0;columnnumber=0;code=NU1101;](NETCORE_ENGINEERING_TELEMETRY=Build) Unable to find package Microsoft.DotNet.ILCompiler. No packages exist with this id in source(s): prebuilt, previously-source-built, reference-packages, source-built
      Failed to restore /home/buildozer/aports/community/dotnet7-build/src/dotnet-v7.0.100-rtm.22521.12/src/sdk/artifacts/source-build/self/src/src/Layout/redist/targets/sdks/sdks.csproj (in 38 ms)
```
ILCompiler nupkg doesn't get added in the artifacts tarball for some reason.

tmds:
> @tmds Have you encountered any errors building SDK with this patch. Our Alpine build on armv7 fails when building on itself:

Is this using mono runtime?

I verified this patch works in builds with coreclr runtime and with mono runtime

With mono runtime, the `ILCompiler` package doesn't get built, and this patch makes the necessary changes for that.

ayakael:
> > @tmds Have you encountered any errors building SDK with this patch. Our Alpine build on armv7 fails when building on itself:
> 
> Is this using mono runtime?
> 
> I verified this patch works in builds with coreclr runtime and with mono runtime
> 
> With mono runtime, the `ILCompiler` package doesn't get built, and this patch makes the necessary changes for that.

Debugging further has confirmed that this patch isn't the cause. Coreclr-flavored runtime build on `armv7` never builds `runtime.alpine.3.17-arm.Microsoft.DotNet.ILCompiler.7.0.0.nupkg` thus the SDK never builds `Microsoft.DotNet.ILCompiler` on the first bootstrap build.

eng/SourceBuild.props
```diff
@@ -8,6 +8,7 @@
     <PropertyGroup>
       <InnerBuildArgs>$(InnerBuildArgs) /p:Projects="$(InnerSourceBuildRepoRoot)\source-build.slnf"</InnerBuildArgs>
       <InnerBuildArgs>$(InnerBuildArgs) /p:UseSharedCompilation=false</InnerBuildArgs>
+      <InnerBuildArgs Condition="'$(SourceBuildUseMonoRuntime)' == 'true'">$(InnerBuildArgs) /p:NativeAotSupported=false</InnerBuildArgs>
```
ayakael, line 11: Some coreclr-supported platforms do not support NativeAot as well (`arm`). Does SDK have a platform variable? I couldn't find any thus I had to find another way to set `NativeAotSupported` when arch is `arm`.  On Alpine, I removed this part of the modification, and moved the check to `repos/sdk.proj` as `$(BuildArchitecture)` does not reach `eng/SourceBuild.props`.

Of course, you can keep this check and source-build can set `NativeAotSupported=false` when `$(BuildArchitecture) == arm`.

Thus, in https://github.com/dotnet/installer/pull/14792:
```diff
diff --git a/src/SourceBuild/tarball/content/repos/sdk.proj b/src/SourceBuild/tarball/content/repos/sdk.proj
index c18e00dd61..6567b41983 100644
--- a/src/SourceBuild/tarball/content/repos/sdk.proj
+++ b/src/SourceBuild/tarball/content/repos/sdk.proj
@@ -6,6 +6,7 @@
     <BuildCommandArgs>$(BuildCommandArgs) $(FlagParameterPrefix)nodereuse $(ArcadeFalseBoolBuildArg)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:PackageProjectUrl=https://github.com/dotnet/sdk</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:PublishCompressedFilesPathPrefix=$(SourceBuiltToolsetDir)</BuildCommandArgs>
+    <BuildCommandArgs Condition="'$(SourceBuildUseMonoRuntime)' == 'true' OR '$(BuildArchitecture)' == 'arm'">$(BuildCommandArgs) /p:NativeAotSupported=false</BuildCommandArgs>
 
     <LogVerbosityOptOut>true</LogVerbosityOptOut>
     <BuildCommandArgs>$(BuildCommandArgs) -v $(LogVerbosity)</BuildCommandArgs>
```

marcpopMSFT:
@agocke @crummel this was approved by tactics for 7.0.3 which is open now. Can this be merged and if so, it needs to be this week?

eng/SourceBuild.props
crummel: > Does SDK have a platform variable?

I don't think so.  This approach looks fine to me.

crummel (approved):

crummel:
> @agocke @crummel this was approved by tactics for 7.0.3 which is open now. Can this be merged and if so, it needs to be this week?

This LGTM to merge.

PR #28696, opened 2022-10-20
Improve Tests for Implicit RIDs and Libraries

nagilson:
As a follow-up to https://github.com/dotnet/sdk/pull/28628, this PR responds to PR feedback for tests that we didn't manage to get ready in time, as it would have delayed builds too long and we needed to get the fix in ASAP. Confirmed that the new tests  fail in both regards without the change and pass in both regards (library paths and RID properties are set correctly) in the new scenario. We could target it to 7.0.1xx but I think that's locked down atm. 

Drafting until 7.0.1xx changes have merged into 7.0.2xx. 
cc @dsplaisted


dsplaisted:
I think we can probably target this to 7.0.1xx.  Test changes don't have to be approved.  We would probably just have to wait until the branch is open in to merge the PR, to avoid triggering unwanted builds.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var targetFramework = ToolsetInfo.CurrentTargetFramework;
+            string rid = "win-x86";
+
+            List<string> args = new List<string>{"/p:PublishSingleFile=true", $"/p:RuntimeIdenitifer={rid}"};
+            var testAsset = Publish(args, testIdentifierName: "PublishSingleFileLibrary");
+
+            var assetFolder = testAsset.TestRoot;
+            var ridlessLibraryDllPath = Path.Combine(assetFolder, "Library", "bin", "Debug", targetFramework, "Library.dll");
+            Assert.True(File.Exists(ridlessLibraryDllPath));
```
dsplaisted, line 65: I'm not sure you need both this file existence check and the `RuntimteIdentifier` property value check.  I think that's mostly redundant.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            var properties = _referencedProject.GetPropertyValues(testAsset.TestRoot, targetFramework: targetFramework);
```
dsplaisted, line 67: I'd suggest setting `AppendRuntimeIdentifierToOutputPath` to false in the referenced project.  Otherwise you have to pass the RuntimeIdentifier to the `GetPropertyValues` method so that it will look in the right output path for the file that's written with the values.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -193,6 +193,30 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
             }
         }
 
+        [Fact]
+        public void ImplicitRuntimeIdentifierOptOutCorrecltyOptsOut()
```
dsplaisted, line 197: Typo:

```suggestion
        public void ImplicitRuntimeIdentifierOptOutCorrectlyOptsOut()
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.NET.TestFramework.ProjectConstruction;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.NET.Publish.Tests
+{
+    public class GivenThatWeWantToPublishASingleFileLibrary : SdkTest
+    {
+        TestProject _testProject;
+        TestProject _referencedProject;
```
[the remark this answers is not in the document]
nagilson: That's correct!

But I was considering adding the publish and build command to the TestProject class:
Something like `public CommandResult TestProject::Build(List<string> args,  bool shouldPass=true, (args would be optional through an overload without args))` and  `public CommandResult TestProject::Publish(List<string> args, bool shouldPass=true)`, and probably they would call a more generic function like `internal CommandResult TestProject::DotnetCommand(command, bool shouldPass=true)`? 

What do you think about doing that? I can see myself writing the same code a lot (and a lot of test code currently does this).

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
nagilson: Yeah, might as well get rid of it (ha, funny pun) to make the path changes easier

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
nagilson: I don't think we should do this, for `_testProject` it would need the RID but for `_referenceProject` the folder without the RID contains `PropertyValuesFile.txt` (or w/e it was called) even without `AppendruntimeIdentifierToOutputPath`. And if `AppendRuntimeIdentifierToOutputPath` was false then it would be a less comprehensive test, as that was one part of the problem that caused the failure in VS

nagilson: We discussed this and the reason is that if the test fails it will be harder to tell why it failed with `AppendRuntimeIdentifierToOutputPath` so I should set it to false.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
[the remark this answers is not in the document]
nagilson: We discussed this and I'm going to try to do it but it might be a problem because of the `_testAssetsManager`.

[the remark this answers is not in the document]
nagilson: it's not worth doing the above proposal, just going to merge into one function

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileLibrary.cs
nagilson: We dont need to worry about this when we move rids out of propertyvalues.txt path.

nagilson:
Will pass once https://github.com/dotnet/sdk/pull/28717 is merged. cc @dsplaisted  to approve. Confirmed the test failed before and passes with the newer rid inference code.

dsplaisted (approved):

PR #28714, opened 2022-10-21
Introduce PublishSelfContained

nagilson:
# Summary 
Adds PublishSelfContained (PSC):

1. You can add 
```xml
<PublishSelfContained>true</PublishSelfContained>
```
to publish self-contained by default but not build self-contained by default. You can of course also forward the property.

# Examples

`dotnet publish --sc=false -p:PublishSelfContained=true` -> Not Self Contained
`dotnet publish /p:SelfContained=false -p:PublishSelfContained=true` -> Not Self Contained
`dotnet publish -p:PublishSelfContained=true` -> Self Contained
`dotnet publish -p:PublishSelfContained=false` -> Not Self Contained

In a project file: 

```xml
<PublishSelfContained>true</PublishSelfContained> 
```
`dotnet publish` -> SelfContained 

```xml
<PublishSelfContained>false</PublishSelfContained> 
```
`dotnet publish` -> Not SelfContained 

`dotnet build -p:PublishSelfContained=true` -> Not Self Contained

# Long Form Explanation of changes

2. If you do `dotnet publish --no-sc -p:PublishSelfContained=true`, `--no-sc` will win. Also true if `PublishSelfContained` is in the properties. Also true for other variants of `--no-sc` like `/p:SelfContained=false`. I got it to work the other way where `PSC` would win but we'd really, really prefer not to do it like that because it could break a lot of stuff. 

3. If you do `dotnet publish` and `<SelfContained>false</SC>` in the project file but `PublishSelfContained` is `true` in the project file or in the CLI, then `PSC` will win. Likewise, If `PSC` is `false` but `SC` is `true` in the project file, `PSC` will win. Like stated above, if `SC` is stated in the command invocation, `SC` will win. 

# Blurbs about the logistical things 

4. Had a large group discussion about getting `_IsPublishing` to work in MSBuild with `t:/Publish` -- it's not possible. (We may not actually need `_IsPublishing` to do this one, but we do for PublishRID and PublishRelease, so it's probably better to keep them aligned.) But we are going to do work to get it to align with VS. We'll need to make changes in their repo(s), but it shouldn't affect anything here, I'd hope. Not sure if we want to wait to merge until we can tell folks they can use this in VS.

5. `PublishSelfContained` will also imply a RID but only if we're publishing, since this property should only do anything if you're publishing. IIRC other publish properties already had a history of failing build without a RID.

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -876,7 +876,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1192: "}</comment>
   </data>
   <data name="PublishSelfContainedMustBeBool" xml:space="preserve">
-    <value>NETSDK1193: PublishSelfContained must be either true or false. The value given was '{0}'.</value>
+    <value>NETSDK1193: If PublishSelfContained is set, it must be either true or false. The value given was '{0}'.</value>
     <comment>{StrBegin="NETSDK1193: "}</comment>
   </data>
 </root>
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -875,4 +875,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1192: Targeting .NET 7.0 or higher in Visual Studio 2022 17.3 is not supported.</value>
     <comment>{StrBegin="NETSDK1192: "}</comment>
   </data>
+  <data name="PublishSelfContainedMustBeBool" xml:space="preserve">
+    <value>NETSDK1193: PublishSelfContained must be either true or false. The value given was '{0}'.</value>
+    <comment>{StrBegin="NETSDK1193: "}</comment>
+  </data>
 </root>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -71,7 +71,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                     '$(PublishReadyToRun)' == 'true' or
                     '$(PublishSingleFile)' == 'true' or
                     '$(PublishAot)' == 'true' or
-                    ('$(_IsPublishing)' == 'true' and '$(PublishSelfContained)' == 'true')
+                    ('$(_CommandLineDefinedSelfContained)' != 'true' and '$(_IsPublishing)' == 'true' and '$(PublishSelfContained)' == 'true')
                   )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -61,14 +61,14 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RuntimeIdentifier Condition="'$(PlatformTarget)' == 'x86' or '$(PlatformTarget)' == ''">win7-x86</RuntimeIdentifier>
   </PropertyGroup>
 
-  <!-- Check if variable is global, resets the value to it's original state if it was local.-->
+  <!-- Check if SelfContained is global, resets the value to it's original state if it was local.-->
   <PropertyGroup>
     <_OriginalSelfContainedValue>$(SelfContained)</_OriginalSelfContainedValue>
     <SelfContained></SelfContained>
   </PropertyGroup>
 
   <Choose>
-    <!-- Assumes a global variable cannot be an undefined variable; not setting it to 'temp' as that may cause type resolution failures.-->
+    <!-- Assumes a global variable cannot be an undefined variable.-->
     <When Condition="'$(SelfContained)' != '' and $(SelfContained) == $(_OriginalSelfContainedValue)">
       <PropertyGroup>
         <_SelfContainedIsGlobal>true</_SelfContainedIsGlobal>
@@ -82,13 +82,19 @@ Copyright (c) .NET Foundation. All rights reserved.
   </Choose>
 
   <PropertyGroup>
-    <SelfContained>$(OriginalSelfContainedValue)</SelfContained>
+    <SelfContained>$(_OriginalSelfContainedValue)</SelfContained>
     <_OriginalSelfContainedValue></_OriginalSelfContainedValue>
   </PropertyGroup>
 
   <!-- PubishSelfContained may be set but not change self contained if we are not publishing, or self contained was already set globally. This condition tracks whether it will take effect.-->
-  <PropertyGroup Condition=" '$(_CommandLineDefinedSelfContained)' != 'true' and '$(_SelfContainedIsGlobal)' != 'true' and '$(_IsPublishing)' == 'true' and '$(PublishSelfContained)' == 'true'">
+  <PropertyGroup Condition=" '$(_CommandLineDefinedSelfContained)' != 'true' and '$(_SelfContainedIsGlobal)' != 'true' and '$(_IsPublishing)' == 'true' and ('$(PublishSelfContained)' == 'true' or '$(PublishSelfContained)' == 'false')">
     <_PublishSelfContainedIsActive>true</_PublishSelfContainedIsActive>
+  </PropertyGroup>
+
+  <!-- Edit SelfContained to match the value of PublishSelfContained if we are publishing.
+       This Won't affect t:/Publish (because of _IsPublishing), and also won't override a global SelfContained property.-->
+  <PropertyGroup Condition="'$(_PublishSelfContainedIsActive)' == 'true'">
+    <SelfContained>$(PublishSelfContained)</SelfContained>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == ''">
@@ -101,7 +107,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                     '$(PublishReadyToRun)' == 'true' or
                     '$(PublishSingleFile)' == 'true' or
                     '$(PublishAot)' == 'true' or
-                    '$(_PublishSelfContainedIsActive)' == 'true'
+                    ('$(_PublishSelfContainedIsActive)' == 'true' and '$(PublishSelfContained)' == 'true')
                   )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
@@ -200,7 +206,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="SelfContained"/>
 
-    <NETSdkError Condition="'$(_PublishSelfContainedIsActive)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == ''"
+    <NETSdkError Condition="'$(_PublishSelfContainedIsActive)' == 'true' and '$(PublishSelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == ''"
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="PublishSelfContained"/>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -71,7 +71,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                     '$(PublishReadyToRun)' == 'true' or
                     '$(PublishSingleFile)' == 'true' or
                     '$(PublishAot)' == 'true' or
-                    '$(PublishSelfContained)' == 'true'
+                    ('$(_IsPublishing)' == 'true' and '$(PublishSelfContained)' == 'true')
                   )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
@@ -170,7 +170,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="SelfContained"/>
 
-    <NETSdkError Condition="'$(PublishSelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == ''"
+    <NETSdkError Condition="'$(_IsPublishing)' == 'true' and '$(PublishSelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == ''"
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="PublishSelfContained"/>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -192,6 +192,11 @@ Copyright (c) .NET Foundation. All rights reserved.
                 FormatArguments="PublishAot"/>
 
     <!-- End of implicit RID resolver checks.-->
+
+    <NETSdkError Condition="'$(PublishSelfContained)' != 'true' and '$(PublishSelfContained)' != 'false' and '$(PublishSelfContained)' != ''"
+                ResourceName="PublishSelfContainedMustBeBool"
+                FormatArguments="$(PublishSelfContained)"/>
+
     <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(UseAppHost)' != 'true' and '$(_RuntimeIdentifierUsesAppHost)' == 'true'"
                  ResourceName="CannotUseSelfContainedWithoutAppHost" />
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -175,10 +175,6 @@ Copyright (c) .NET Foundation. All rights reserved.
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="SelfContained"/>
 
-    <NETSdkError Condition="'$(_PublishSelfContainedIsActive)' == 'true' and '$(PublishSelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == ''"
-                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
-                 FormatArguments="PublishSelfContained"/>
-
     <NETSdkError Condition="'$(PublishReadyToRun)' == 'true' and '$(RuntimeIdentifier)' == ''"
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="PublishReadyToRun"/>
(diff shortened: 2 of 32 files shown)


nagilson:
Pipeline failure is due to infrastructure.
```

##[error].packages/microsoft.dotnet.arcade.sdk/7.0.0-beta.22464.4/tools/Tools.proj(0,0): error NU1301: (NETCORE_ENGINEERING_TELEMETRY=Restore) Failed to retrieve information about 'Microsoft.SymbolUploader.Build.Task' from remote source 'https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/3e04f704-20f0-4deb-929d-a189c3a4dae4/nuget/v3/flat2/microsoft.symboluploader.build.task/index.json'.
  Retrying 'FindPackagesByIdAsync' for source 'https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/49e5305d-d845-4a14-9d69-6f5dbfb9570c/nuget/v3/flat2/sn/index.json'.
  Response status code does not indicate success: 429 ().
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -61,6 +61,42 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <PropertyGroup>
+        <_SelfContainedIsGlobal>false</_SelfContainedIsGlobal>
+      </PropertyGroup>
+    </Otherwise>
+  </Choose>
+
+  <PropertyGroup>
+    <SelfContained>$(_OriginalSelfContainedValue)</SelfContained>
+    <_OriginalSelfContainedValue></_OriginalSelfContainedValue>
+  </PropertyGroup>
```
dsplaisted, line 87: Any reason to use a `Choose` block instead of something like this?

```suggestion
  <PropertyGroup>
    <_OriginalSelfContainedValue>$(SelfContained)</_OriginalSelfContainedValue>
    <SelfContained></SelfContained>

    <_SelfContainedIsGlobal>false</_SelfContainedIsGlobal>
    <!-- Assumes a global variable cannot be an undefined variable.-->
    <_SelfContainedIsGlobal Condition="'$(SelfContained)' != '' and $(SelfContained) == $(_OriginalSelfContainedValue)">true</_SelfContainedIsGlobal>

    <SelfContained>$(_OriginalSelfContainedValue)</SelfContained>
    <_OriginalSelfContainedValue></_OriginalSelfContainedValue>
  </PropertyGroup>
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -61,6 +61,42 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+
+  <!-- PubishSelfContained may be set but not change self contained if we are not publishing, or self contained was already set globally. This condition tracks whether it will take effect.-->
+  <PropertyGroup Condition=" '$(_CommandLineDefinedSelfContained)' != 'true' and '$(_SelfContainedIsGlobal)' != 'true' and '$(_IsPublishing)' == 'true' and ('$(PublishSelfContained)' == 'true' or '$(PublishSelfContained)' == 'false')">
```
dsplaisted, line 90: Is the `_CommandLineDefinedSelfContained` redundant here?  I think if `_CommandLineDefinedSelfContained` is true then that would imply `_SelfContainedIsGlobal` would also be true.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -71,7 +107,8 @@ Copyright (c) .NET Foundation. All rights reserved.
                     '$(SelfContained)' == 'true' or
                     '$(PublishReadyToRun)' == 'true' or
                     '$(PublishSingleFile)' == 'true' or
-                    '$(PublishAot)' == 'true'
+                    '$(PublishAot)' == 'true' or
+                    ('$(_PublishSelfContainedIsActive)' == 'true' and '$(PublishSelfContained)' == 'true')
```
dsplaisted, line 111: Is this extra clause necessary?  It seems if these two properties are true then `SelfContained` would also be true, which is already covered earlier on.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -61,6 +61,42 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <!-- PubishSelfContained may be set but not change self contained if we are not publishing, or self contained was already set globally. This condition tracks whether it will take effect.-->
+  <PropertyGroup Condition=" '$(_CommandLineDefinedSelfContained)' != 'true' and '$(_SelfContainedIsGlobal)' != 'true' and '$(_IsPublishing)' == 'true' and ('$(PublishSelfContained)' == 'true' or '$(PublishSelfContained)' == 'false')">
+    <_PublishSelfContainedIsActive>true</_PublishSelfContainedIsActive>
+  </PropertyGroup>
+
+  <!-- Edit SelfContained to match the value of PublishSelfContained if we are publishing.
+       This Won't affect t:/Publish (because of _IsPublishing), and also won't override a global SelfContained property.-->
+  <PropertyGroup Condition="'$(_PublishSelfContainedIsActive)' == 'true'">
+    <SelfContained>$(PublishSelfContained)</SelfContained>
+  </PropertyGroup>
```
dsplaisted, line 98: Do we need `_PublishSelfContainedIsActive` and whether `SelfContained` was specified in a global property at all?  Could we do the following instead?

```suggestion
  <PropertyGroup Condition="'$(PublishSelfContained)' != ''">
    <SelfContained>$(PublishSelfContained)</SelfContained>
  </PropertyGroup>
```

This would have no effect if `SelfContained` was a global property, but we wouldn't need all that extra logic to check.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -170,6 +207,10 @@ Copyright (c) .NET Foundation. All rights reserved.
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="SelfContained"/>
 
+    <NETSdkError Condition="'$(_PublishSelfContainedIsActive)' == 'true' and '$(PublishSelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' == ''"
```
dsplaisted, line 210: Can we merge the changes from https://github.com/dotnet/sdk/pull/28484 first, and follow that pattern in the new condition here (ie no need to check `RuntimeIdentifiers`?

dsplaisted (commented):
I might be missing something, but I think the logic can be simplified a lot to not need to even check if `SelfContained` is a global property.

I didn't look at the tests much yet.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: I don't think we need this error anymore and can just rely on the `SelfContained` one. Let me know if you still wanna merge that one first.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: I don't think we can do that exact proposal, we still need to check `_IsPublishing`. And I don't think we can just check `!= ''` because it would fail if someone passed a value besides `true` or `false`. 

But overall, I agree we can slim this down a lot and achieve the same result. I tried to do this somewhere else to decouple SC from RidInference but ended up needing to move it here anyways and some stuff got left.

dsplaisted: I think the valid values for `PublishSelfContained` are the same as the valid values for `SelfContained`: basically True, False, and empty.  So I'm not sure if we need to worry about whether or not we copy an invalid value.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: Thanks, this is helpful for my learning... Trying to think of MSBuild in if/elses is not the best idea it seems

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: Ya, thanks

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: If we do the check as true or false, it will ignore PublishSelfContained silently. If we do the check with `!= ''` it produces the following:
```go
MSBuild version 17.4.0+18d5aef85 for .NET
C:\Users\\sdk\artifacts\bin\redist\Debug\dotnet\sdk\7.0.100-dev\Sdks\Microsoft.
NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(99,33): error MSB4
030: "Blah" is an invalid value for the "SelfContained" parameter of the "ProcessFrameworkR
eferences" task. The "SelfContained" parameter is of type "System.Boolean". [C:\Users\noa
hgilson\sdk\newc\newc.csproj]****
```

So I'll add a new error for `PublishSelfContained`. Thank you for your help, I made this way too complicated

nagilson:
@dsplaisted Responded to the feedback, please take a look again when you have availability. Thanks!
(btw the test code is old so it uses HelloWorld, in the future I plan to use the testproject more)

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -875,4 +875,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1192: Targeting .NET 7.0 or higher in Visual Studio 2022 17.3 is not supported.</value>
     <comment>{StrBegin="NETSDK1192: "}</comment>
   </data>
+  <data name="PublishSelfContainedMustBeBool" xml:space="preserve">
+    <value>NETSDK1193: PublishSelfContained must be either true or false. The value given was '{0}'.</value>
```
dsplaisted, line 879:
```suggestion
    <value>NETSDK1193: If PublishSelfContained is set, it must be either true or false. The value given was '{0}'.</value>
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -374,6 +374,35 @@ static int Last2DigitsTo0(int versionBuild)
...
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("HelloWorld", identifier: "ItDoesNotBuildSCDueToPSC")
+                .WithSource()
+                .WithTargetFramework(targetFramework)
+                .WithProjectChanges(project =>
+                {
+                    var ns = project.Root.Name.Namespace;
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifier", runtimeIdentifier));
+                    propertyGroup.Add(new XElement(ns + "SelfContained", "false"));
```
dsplaisted, line 391: Any reason to set the `RuntimeIdentifier` and `SelfContained` properties here?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -374,6 +374,35 @@ static int Last2DigitsTo0(int versionBuild)
             publishCommand.Execute(new [] {"-property:SelfContained=true", "-property:_CommandLineDefinedSelfContained=true", $"-property:RuntimeIdentifier={rid}", "-property:_CommandLineDefinedRuntimeIdentifier=true" }).Should().Pass().And.NotHaveStdOutContaining("warning");
         }
 
+        [Theory]
+        [InlineData("net7.0")]
```
dsplaisted, line 378: Do we want a Theory with only one instance?  Should we have a fact that just uses `ToolsetInfo.CurrentTargetFramework` instead?

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -137,6 +140,83 @@ public void ItPublishesSelfContainedWithRid(string args)
...
+
+            var testAsset = _testAssetsManager
+            .CopyTestAsset("HelloWorld", identifier: $"PSC-OVERRIDES-{publishSelfContained}-{selfContainedIsGlobal}-{publishSelfContainedIsGlobal}")
+            .WithSource()
+            .WithProjectChanges(project =>
+            {
+                var ns = project.Root.Name.Namespace;
+                var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                propertyGroup.Add(new XElement(ns + "SelfContained", (!publishSelfContained).ToString()));
+                propertyGroup.Add(new XElement(ns + "PublishSelfContained", publishSelfContained.ToString()));
```
dsplaisted, line 179: If these properties are global you probably don't want to set them in the project here.

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -137,6 +140,83 @@ public void ItPublishesSelfContainedWithRid(string args)
...
+            });
+
+            var publishCommand = new PublishCommand(testAsset);
+            List<string> args = new List<string>
+            {
+                $"/p:RuntimeIdentifier={rid}",
+                $"/p:_IsPublishing=true", // Normally this would be set by the CLI (OR VS Soon TM), but this calls directly into t:/Publish.
```
dsplaisted, line 186: Instead of using PublishCommand, I would probably use `dotnet publish` here so that you don't have to explicitly add this property.  You can do something like

```c#
new DotnetCommand(Log, "publish")
  .WithWorkingDirectory(testAsset.Path)
  .Execute(args)
  .Should()
  .Pass();
```

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -137,6 +140,83 @@ public void ItPublishesSelfContainedWithRid(string args)
...
+                var ns = project.Root.Name.Namespace;
+                var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                propertyGroup.Add(new XElement(ns + "SelfContained", (!publishSelfContained).ToString()));
+                propertyGroup.Add(new XElement(ns + "PublishSelfContained", publishSelfContained.ToString()));
+            });
+
+            var publishCommand = new PublishCommand(testAsset);
+            List<string> args = new List<string>
+            {
+                $"/p:RuntimeIdentifier={rid}",
```
dsplaisted, line 185: Can we do the test without passing in the RuntimeIdentifier?  Setting the RuntimeIdentifier will default SelfContained to true (though this is probably changing for .NET 8), which kind of interferes with what we're trying to test here.

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -137,6 +140,83 @@ public void ItPublishesSelfContainedWithRid(string args)
...
+            publishCommand
+                .Execute(args.ToArray())
+                .Should()
+                .Pass();
+
+            var publishedDirectory = publishCommand.GetOutputDirectory(targetFramework: targetFramework, runtimeIdentifier: rid);
+            var expectedFiles = new[] { "HelloWorld.dll" };
+
+            if (publishSelfContained && !selfContainedIsGlobal)
+                expectedFiles.Append("System.dll"); // System.dll should only exist if self contained
```
dsplaisted, line 200: Is there a better way to test for self-contained than this?

dsplaisted (approved):

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
nagilson: Removed them.
(It was so the output path could be given a RID but SC wouldn't be inferred from the RID.)

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
nagilson: Good point

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
nagilson: Yeah, the only way I could think of was to use the other test framework so I migrated it to that

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
nagilson: Made this change, thanks for suggesting it! It's a bit messy because the RID wouldn't be known in the path so it needs to be deduced, but it works out OK.

nagilson:
@dsplaisted I implemented all of the requested test changes. (Thanks, I should be less sloppy with my tests). I think the one thing you may object to / the only other thing worth looking at is that I added https://github.com/dotnet/sdk/pull/28714/files#diff-95fe3aebd6a9400190035038d2c2976fae64579c63dc1c5d4d912f7bc53790bbR478 this change to `TestProject` where you can provide the `GetPropertyValues` function a hard-path instead of a test directory + properties. The alternative I saw was to support automatic RID folders in that change, or to add back in the RID into the test, but I think this would be useful in a few of the other tests I wrote. 

Obvious problems with this solution:
It encourages hardcoding the path which will make the test migration harder if this is used a lot more
it may be the first overload people see instead of the testRoot overload which is what they generally should use, almost always.

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -137,6 +141,89 @@ public void ItPublishesSelfContainedWithRid(string args)
...
+            publishCommand
+                .WithWorkingDirectory(Path.Combine(testAsset.Path, "MainProject"))
+                .Execute(args.ToArray())
+                .Should()
+                .Pass();
+
+            var publishedDirectory = testAsset.TestRoot;
+            publishedDirectory = Path.Combine(publishedDirectory, "MainProject", "obj", "Debug", targetFramework);
+            if (resultShouldBeSelfContained)
+                publishedDirectory = Directory.GetDirectories(publishedDirectory).FirstOrDefault(); // get the rid in the directory
```
dsplaisted, line 203: If there is more than one RID directory, then it would be better to throw an exception here rather than to return null and fail later.

```suggestion
                publishedDirectory = Directory.GetDirectories(publishedDirectory).First(); // get the rid in the directory
```

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestProject.cs
```diff
@@ -458,15 +458,28 @@ public void RecordProperties(params string[] propertyNames)
...
+
+        /// <summary>
+        /// Returns a dictionary of property name to value mappings.
+        /// </summary>
+        /// <param name="finalOutputPath">
+        /// The final output path after executing a dotnet command where propertyvalues.txt will be generated.
+        /// If the output path does not need to be hardcoded, use the overload requesting testRoot instead.
+        /// </param>
+        /// <returns></returns>
+        public Dictionary<string, string> GetPropertyValues(string finalOutputPath)
```
dsplaisted, line 478: The alternative I was thinking of was to not write the file in a RID-specific path.  IE update the injected target in `TestProject.Create` to something like this:

```c#
                string injectTargetContents =
    $@"<Project>
  <Target Name=`WritePropertyValues` BeforeTargets=`AfterBuild`>
    <ItemGroup>
{propertiesElements}
    </ItemGroup>
    <WriteLinesToFile
      File=`$(BaseIntermediateOutputPath)\$(Configuration)\$(TargetFramework)\PropertyValues.txt`
      Lines=`@(LinesToWrite)`
      Overwrite=`true`
      Encoding=`Unicode`
      />
  </Target>
</Project>";
```

src/Tests/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
nagilson: Good point, thanks. With the below change this won't be needed.

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestProject.cs
nagilson: That makes sense. The `PublishRuntimeIdentifier` tests need this too and I added it there. I think I should retarget both of these to 7.0.2xx. And then merge the publish RID PR. Then merge this with 7.0.2xx and change the test.

PR #28717, opened 2022-10-21
Introduce PublishRuntimeIdentifier

nagilson:
# Changes Examples

`dotnet publish  --use-current-runtime -p:PublishRuntimeIdentifier=tizen` will publish with tizen, not the ucr.
`dotnet publish --self-contained -p:PublishRuntimeIdentifier=win-x86` will publish with rid: win-x86
`dotnet publish -p:PublishRuntimeIdentifier=win-x86` will publish with win-x86 rid.

In a project file:
```xml
<RuntimeIdentifier>win-x86</RuntimeIdentifier>
<PublishRuntimeIdentifier>win-x64</PublishRuntimeIdentifier>
```
`dotnet build` -> win-x86 rid.
`dotnet publish` -> win-x64 rid. 

`dotnet publish -r win-x86 /p:PublishRuntimeIdentifier=tizen` will get rid: win-x86. Not ideal, we can do it the other way but it will have potentially far-reaching consequences. 
`dotnet publish --property:RuntimeIdentifier=win-x86 /p:PublishRuntimeIdentifier=tizen`: gives win-x86 rid.

# Summary 
- `PublishRuntimeIdentifier` will override `RuntimeIdentifier` when publishing unless it is global, like passed in with `dotnet publish --r RID /p:PulishRuntimeIdentifier=RID_2`
- `PublishRuntimeIdentifier` will override the RID if UCR is used.
- Same thing where it doesn't work with t:/Publish but we plan to make it work in VS Publish outside of this PR. 

The rest is pretty self explanatory...

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1086,7 +1086,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <AdditionalTargetFrameworkInfoProperty Include="ShouldBeValidatedAsExecutableReference"/>
     <AdditionalTargetFrameworkInfoProperty Include="_SelfContainedWasSpecified"/>
   </ItemGroup>
-  
+
   <PropertyGroup Condition="'$(RuntimeIdentifier)' == 'browser-wasm'">
     <!-- Don't generate a NETSDK1151 error if a non self-contained Exe references a Blazor wasm Exe -->
     <ShouldBeValidatedAsExecutableReference>false</ShouldBeValidatedAsExecutableReference>
@@ -1132,15 +1132,6 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Target Name="ValidateCommandLineProperties"
           AfterTargets="_CheckForInvalidConfigurationAndPlatform" >
-
-    <NETSdkWarning Condition="'$(_SelfContainedWasSpecified)' != 'true' and
-                   '$(_CommandLineDefinedRuntimeIdentifier)' == 'true' and
-                   '$(RuntimeIdentifier)' != '' and
-                   '$(PublishAot)' != 'true' and
-                   '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and
-                   $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0'))"
-               ResourceName="SelfContainedOptionShouldBeUsedWithRuntime" />
-
   </Target>
 
   <!--
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1086,7 +1086,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <AdditionalTargetFrameworkInfoProperty Include="ShouldBeValidatedAsExecutableReference"/>
     <AdditionalTargetFrameworkInfoProperty Include="_SelfContainedWasSpecified"/>
   </ItemGroup>
-
+  
   <PropertyGroup Condition="'$(RuntimeIdentifier)' == 'browser-wasm'">
     <!-- Don't generate a NETSDK1151 error if a non self-contained Exe references a Blazor wasm Exe -->
     <ShouldBeValidatedAsExecutableReference>false</ShouldBeValidatedAsExecutableReference>
@@ -1132,6 +1132,15 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Target Name="ValidateCommandLineProperties"
           AfterTargets="_CheckForInvalidConfigurationAndPlatform" >
+
+    <NETSdkWarning Condition="'$(_SelfContainedWasSpecified)' != 'true' and
+                   '$(_CommandLineDefinedRuntimeIdentifier)' == 'true' and
+                   '$(RuntimeIdentifier)' != '' and
+                   '$(PublishAot)' != 'true' and
+                   '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and
+                   $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0'))"
+               ResourceName="SelfContainedOptionShouldBeUsedWithRuntime" />
+
   </Target>
 
   <!--
--- a/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
@@ -230,7 +230,7 @@ namespace Microsoft.NET.Publish.Tests
             Assert.True(finalRid == expectedRid); // This assert is theoretically worthless as the above code will fail if the RID path is wrong.
         }
 
-        [WindowsOnlyFact]
+        [WindowsOnlyTheory]
         [InlineData("net7.0", "tizen")] // tizen is an arbitrary nonwindows rid, picked because it will be different from a windows rid.
         public void PublishRuntimeIdentifierDoesNotOverrideUseCurrentRuntime(string tfm, string publishRid)
         {
--- a/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
@@ -243,6 +243,7 @@ namespace Microsoft.NET.Publish.Tests
             testProject.AdditionalProperties["UseCurrentRuntimeIdentifier"] = "true";
             testProject.AdditionalProperties["PublishRuntimeIdentifier"] = publishRid;
             testProject.RecordProperties("RuntimeIdentifier");
+            testProject.RecordProperties("NETCoreSdkPortableRuntimeIdentifier");
 
             var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: "UCR_PUBLISH_RID_OVERRIDES");
             var publishCommand = new PublishCommand(testAsset);
@@ -251,11 +252,16 @@ namespace Microsoft.NET.Publish.Tests
                 .Should()
                 .Pass();
 
-            var currentRid = EnvironmentInfo.GetCompatibleRid(testProject.TargetFrameworks);
-            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: tfm, runtimeIdentifier: currentRid);
-            var finalRid = properties["RuntimeIdentifier"];
+            var projectTfmPath = Path.Combine(testAsset.Path, System.Reflection.MethodBase.GetCurrentMethod().Name, "bin", "Debug", tfm);
+            var projectPath = Directory.GetDirectories(projectTfmPath).FirstOrDefault();
+            var testResolvedRid = Path.GetFileName(projectPath);
 
-            Assert.True(finalRid == currentRid); // This assert is theoretically worthless as the above code will fail if the RID path is wrong.
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: tfm, runtimeIdentifier: testResolvedRid);
+            var finalRid = properties["RuntimeIdentifier"];
+            var expectedRid = properties["NETCoreSdkPortableRuntimeIdentifier"];
+
+            Assert.True(expectedRid == testResolvedRid);
+            Assert.True(finalRid == expectedRid); // This assert is theoretically worthless as the above code will fail if the RID path is wrong.
         }
 
         [Fact]
--- a/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
@@ -195,30 +195,37 @@ namespace Microsoft.NET.Publish.Tests
         }
 
         [WindowsOnlyTheory]
-        [InlineData("net7.0", "win-x64", "win-x86", false, false)]
-        [InlineData("net7.0", "win-x64", "win-x86", true, false)]
-        [InlineData("net7.0", "win-x64", "win-x86", true, true)]
-        public void PublishRuntimeIdentifierSetsRuntimeIdentifierAndDoesOrDoesntOverrideRID(string tfm, string publishRuntimeIdentifier, string runtimeIdentifier, bool runtimeIdentifierIsGlobal, bool publishRuntimeIdentifierIsGlobal)
+        [InlineData(false, false)]
+        [InlineData(true, false)]
+        [InlineData(true, true)]
+        public void PublishRuntimeIdentifierSetsRuntimeIdentifierAndDoesOrDoesntOverrideRID(bool runtimeIdentifierIsGlobal, bool publishRuntimeIdentifierIsGlobal)
         {
+            string tfm = ToolsetInfo.CurrentTargetFramework;
+            string publishRuntimeIdentifier = "win-x64";
+            string runtimeIdentifier = "win-x86";
+
             var testProject = new TestProject()
             {
                 IsExe = true,
                 TargetFrameworks = tfm
             };
-            testProject.AdditionalProperties["RuntimeIdentifier"] = runtimeIdentifier;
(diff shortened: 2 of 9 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1132,15 +1132,6 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   <Target Name="ValidateCommandLineProperties"
           AfterTargets="_CheckForInvalidConfigurationAndPlatform" >
-
-    <NETSdkWarning Condition="'$(_SelfContainedWasSpecified)' != 'true' and
-                   '$(_CommandLineDefinedRuntimeIdentifier)' == 'true' and
-                   '$(RuntimeIdentifier)' != '' and
-                   '$(PublishAot)' != 'true' and
-                   '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and
-                   $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '6.0'))"
-               ResourceName="SelfContainedOptionShouldBeUsedWithRuntime" />
```
dsplaisted, line 1142: I think we probably shouldn't be deleting this yet.  I think we planned to change the behavior so that specifying the RuntimeIdentifier didn't automatically set SelfContained to true.  I think we_ would change this warning when we changed that, but we would probably only make the new logic apply to .NET 8 and higher, so this warning should probably be left in place for .NET 6 and 7.

nagilson: Ah I see, thanks for double checking it. Let me revert that change

nagilson:
@dsplaisted PTAL when you're free

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -193,6 +193,76 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+            testProject.RecordProperties("RuntimeIdentifier");
+
+            List<string> args = new List<string>
+            {
+                $"/p:_IsPublishing=true", // Normally this would be set by the CLI (OR VS Soon TM), but this calls directly into t:/Publish.
+                runtimeIdentifierIsGlobal ? $"/p:RuntimeIdentifier={runtimeIdentifier}" : "",
+                publishRuntimeIdentifierIsGlobal ? $"/p:PublishRuntimeIdentifier={publishRuntimeIdentifier}" : ""
+            };
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: $"{publishRuntimeIdentifierIsGlobal}-{runtimeIdentifierIsGlobal}");
```
dsplaisted, line 218: We should include all of the theory parameters in the identifier here.  If some values don't change (like the RuntimeIdentifiers), then we may not need to include them as test theory parameters at all, we can just hard-code them in the test.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -193,6 +193,76 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+            var publishCommand = new PublishCommand(testAsset);
+            publishCommand
+                .Execute(args.ToArray())
+                .Should()
+                .Pass();
+
+            string expectedRid = runtimeIdentifierIsGlobal ? runtimeIdentifier : publishRuntimeIdentifier;
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: tfm, runtimeIdentifier: expectedRid);
```
dsplaisted, line 226: Similar to what we talked about on another test, if the RuntimeIdentifier that the project is built with is wrong, this won't be able to read the property value.  The test should still fail but it won't be as clear why it's failing.  It's OK to leave like this, but maybe we should change the property recording logic so that it doesn't put the property values in a RuntimeIdentifier-specific folder.

EDIT: I see you noted some of this in a comment below.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -193,6 +193,76 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+            var finalRid = properties["RuntimeIdentifier"];
+
+            Assert.True(finalRid == expectedRid); // This assert is theoretically worthless as the above code will fail if the RID path is wrong.
+        }
+
+        [WindowsOnlyTheory]
+        [InlineData("net7.0", "tizen")] // tizen is an arbitrary nonwindows rid, picked because it will be different from a windows rid.
```
dsplaisted, line 233: I would probably use `linux-x64` as the RID here.

Also if there's only one instance it should probably be a `Fact` instead of a `Theory`.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -193,6 +193,76 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+                IsExe = true,
+                TargetFrameworks = tfm
+            };
+
+            testProject.AdditionalProperties["UseCurrentRuntimeIdentifier"] = "true";
+            testProject.AdditionalProperties["PublishRuntimeIdentifier"] = publishRid;
+            testProject.RecordProperties("RuntimeIdentifier");
+            testProject.RecordProperties("NETCoreSdkPortableRuntimeIdentifier");
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: "UCR_PUBLISH_RID_OVERRIDES");
```
dsplaisted, line 247: We don't need to hard-code the identifier.  By default the folder created will depend on the test method name, and the identifier is used to avoid collisions for different instances of a theory.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -193,6 +193,76 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+            string expectedRid = runtimeIdentifierIsGlobal ? runtimeIdentifier : publishRuntimeIdentifier;
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: tfm, runtimeIdentifier: expectedRid);
+            var finalRid = properties["RuntimeIdentifier"];
+
+            Assert.True(finalRid == expectedRid); // This assert is theoretically worthless as the above code will fail if the RID path is wrong.
+        }
+
+        [WindowsOnlyTheory]
+        [InlineData("net7.0", "tizen")] // tizen is an arbitrary nonwindows rid, picked because it will be different from a windows rid.
+        public void PublishRuntimeIdentifierDoesNotOverrideUseCurrentRuntime(string tfm, string publishRid)
```
dsplaisted, line 234: It feels like this behavior is backwards from what it should be.  If you specify both `UseCurrentRuntimeIdentifier` and `PublishRuntimeIdentifier` in the project file, it is probably because you want to build for the development machine, but publish for a different target.

@richlander @baronfel Thoughts?

nagilson: I agree, not sure what I was thinking before. I'll change it for now but leave this thread open.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
nagilson: Made that change. It's going to have to go in another PR because PublishSelfContained also relies on it.

nagilson:
@dsplaisted PTAL. I've made all of the requested changes and a few others where the same comment applied to other parts. Let's merge this and then use this testProject change in the PublishSelfContained PR.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -194,6 +195,81 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+            {
+                runtimeIdentifierIsGlobal ? $"/p:RuntimeIdentifier={runtimeIdentifier}" : "",
+                publishRuntimeIdentifierIsGlobal ? $"/p:PublishRuntimeIdentifier={publishRuntimeIdentifier}" : ""
+            };
+
+            string identifier = $"PublishRuntimeIdentifierOverrides-{publishRuntimeIdentifierIsGlobal}-{runtimeIdentifierIsGlobal}";
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: identifier);
+            var publishCommand = new DotnetPublishCommand(Log);
+            publishCommand
+                .WithWorkingDirectory(Path.Combine(testAsset.TestRoot, MethodBase.GetCurrentMethod().Name))
```
dsplaisted, line 229: You should be able to do this instead of getting the method name via reflection:

```suggestion
                .WithWorkingDirectory(Path.Combine(testAsset.TestRoot, testProject.Name))
```

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -194,6 +195,81 @@ public void PublishWithRuntimeIdentifier(bool publishNoBuild)
...
+                .Pass();
+
+            string expectedRid = runtimeIdentifierIsGlobal ? runtimeIdentifier : publishRuntimeIdentifier;
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: tfm);
+            var finalRid = properties["RuntimeIdentifier"];
+
+            Assert.True(finalRid == expectedRid);
+        }
+
+        [WindowsOnlyFact]
```
dsplaisted, line 241: This test can probably also run on all operating systems.  You might need to choose a `publishRid` that isn't going to be the default RID anywhere.

dsplaisted (approved):

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
nagilson: Hm, while we could use something obscure and specific, like `rocky-arm64`, I don't think there's much added value in running this test on more than one platform for the potential ways this can go wrong.

nagilson:
@dsplaisted Thanks for taking a look, I appreciate your time. I added some comments in the test code. As incorrect comments can be quite insidious, lmk if you see anything wrong.

src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
```diff
@@ -21,6 +21,9 @@ public class TestAsset : TestDirectory
 
         public string TestRoot => Path;
 
+        /// <summary>
+        /// The hashed test name (so file paths do not become too long) of a particular test. This will be the root of the test folder.
```
dsplaisted: The hashing happens in `TestAssetsManager`, so I would probably note that here.  Otherwise it might be confusing to see this comment but not the code that does the hashing in `TestAsset`.

src/Tests/Microsoft.NET.TestFramework/TestAssetsManager.cs
```diff
@@ -52,6 +52,15 @@ public TestAssetsManager(ITestOutputHelper log)
             return testAsset;
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="testProject">The testProject used to create a teatAsset with.</param>
+        /// <param name="callingMethod">Defaults to the name of the caller function (presumably the test). Used to prevent file collisions on tests which share the same test project.</param>
+        /// <param name="identifier">An ID for the project that should be unique for every test, including each theory child test.
```
dsplaisted: The ID doesn't need to be different for tests where the `callingMethod` is different.

```suggestion
        /// <param name="identifier">An ID that should be different for every child test of a theory, to ensure that each test uses a different test folder.
```

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestProject.cs
```diff
@@ -21,14 +21,18 @@ public TestProject([CallerMemberName] string name = null)
...
+        /// By default, it is the unhashed name of the function that instantiated the TestProject object.
         /// </summary>
         public string Name { get; set; }
 
         public bool IsSdkProject { get; set; } = true;
 
         public bool IsExe { get; set; }
 
+        /// <summary>
+        /// The project is a WPF or Windows Form App Executable.
```
dsplaisted, line 34: This actually just controls whether the `OutputType` in the generated project is set to `WinExe`.  It's not directly tied to the UI framework (WPF, Windows Forms, WinUI, or something else).

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestProject.cs
```diff
@@ -61,6 +65,9 @@ public TestProject([CallerMemberName] string name = null)
 
         public Dictionary<string, string> EmbeddedResources { get; } = new Dictionary<string, string>();
 
+        /// <summary>
+        /// Use this dictionary to set a property key to a desired value before evaluation.
```
dsplaisted, line 69:
```suggestion
        /// Use this dictionary to set add properties and values to the created project.
```

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestProject.cs
```diff
@@ -459,6 +466,10 @@ public void RecordProperties(params string[] propertyNames)
             PropertiesToRecord.AddRange(propertyNames);
         }
 
+        /// <returns>
+        /// A dictionary of property keys to property value strings, case sensitive.
+        /// Only properties added to the PropertiesToRecord member will be observed.<see cref="PropertiesToRecord"/>
```
dsplaisted, line 471:
```suggestion
        /// Only properties added to the <see cref="PropertiesToRecord"/> member will be observed.
```

src/Tests/Microsoft.NET.TestFramework/TestAssetsManager.cs
```diff
@@ -53,14 +53,16 @@ public TestAssetsManager(ITestOutputHelper log)
...
+        /// Writes an in-memory test project onto the disk.
         /// </summary>
-        /// <param name="testProject">The testProject used to create a teatAsset with.</param>
-        /// <param name="callingMethod">Defaults to the name of the caller function (presumably the test). Used to prevent file collisions on tests which share the same test project.</param>
-        /// <param name="identifier">An ID for the project that should be unique for every test, including each theory child test.
+        /// <param name="testProject">The testProject used to create a testAsset with.</param>
+        /// <param name="callingMethod">Defaults to the name of the caller function (presumably the test).
+        /// Used to prevent file collisions on tests which share the same test project.</param>
+        /// <param name="identifier">Use this for theories.
+        /// The Identifier is an ID for the project that should be unique for every test, including each theory child test.
```
dsplaisted, line 62:
```suggestion
        /// The Identifier is used to distinguish between theory child tests.  Generally it should be created using a combination of all of the theory parameter values.
```

src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
```diff
@@ -13,6 +13,10 @@
 
 namespace Microsoft.NET.TestFramework
 {
+    /// <summary>
+    /// A directory wrapper around the <see cref="TestProject"/> class.
+    /// It manages the on-disk files of the TestProject and provides additional functionality to edit projects.
+    /// </summary>
```
dsplaisted, line 19: This isn't only a wrapper over a `TestProject`.  It may also be a wrapper over files that are copied from the `src\Assets` folder.  It is created via `TestAssetsManager.CopyTestAsset` or `TestAssetsManager.CreateTestProject`, depending on which is the case.

dsplaisted (approved):

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestProject.cs
nagilson: Suggestion: ` /// Use this dictionary to set add properties and values to the created project.` -->
Accepted Revision: ` /// Use this dictionary to set a property (the key) to a value for the created project.`

src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
nagilson: Gotcha!

dsplaisted (approved):

PR #28769, opened 2022-10-25
add console message for package validation success or failure

### LINKED ISSUE #19214: [Package-Validation] Provide confirmation  that the package validation has run successfully
A couple of folks have asked if there is a way to check if the package validation has run successfully. 
 Currently we dont do anything if the package validation run successfully.

cc @safern @ericstj @ViktorHofer

### LINKED ISSUE #28083: apicompat/tool: log message to explain how to generate suppressions
When the APICompat CLI tool produces a list of diagnostics, log a message explaining how to re-run it with the \p:GenerateCompatibilitySuppressions flag.

/cc @ViktorHofer

smasher164:
This change prints a console message for each execution of `ValidatePackage` or `ValidateAssembly`. If there were no suppressions logged, then it prints "success". Otherwise, it explains how to rebuild to update the suppression files.

Fixes #28083.
Fixes #19214.


src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/MSBuildCompatibilityLogger.cs
```diff
@@ -38,9 +40,13 @@ private bool LogSuppressableMessage(MessageLevel messageLevel, Suppression suppr
...
                 return false;
 
+            _suppressionWasLogged = true;
+
             _log.Log(new Message(messageLevel, string.Format(format, args), code));
 
             return true;
         }
+
+        public bool SuppressionWasLogged => _suppressionWasLogged;
```
ViktorHofer, line 50:
```suggestion
        public bool SuppressionWasLogged { get; private set; }
```

src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/MSBuildCompatibilityLogger.cs
```diff
@@ -13,12 +13,14 @@ internal sealed class MSBuildCompatibilityLogger : ICompatibilityLogger
...
         private readonly Logger _log;
         private readonly ISuppressionEngine _suppressionEngine;
+        private bool _suppressionWasLogged;
 
         public MSBuildCompatibilityLogger(Logger log,
             ISuppressionEngine suppressionEngine)
         {
             _log = log;
             _suppressionEngine = suppressionEngine;
+            _suppressionWasLogged = false;
```
ViktorHofer, line 23:
```suggestion
```

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/ICompatibilityLogger.cs
```diff
@@ -35,5 +35,7 @@ public interface ICompatibilityLogger
         /// <param name="format">The message format/param>
         /// <param name="args">The message format arguments</param>
         void LogMessage(MessageImportance importance, string format, params string[] args);
+
+        bool SuppressionWasLogged { get; }
```
ViktorHofer, line 39: Please add tripple slash public api comments

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
```diff
@@ -87,6 +88,23 @@ internal static class ValidateAssemblies
...
+                {
+                    serviceProvider.CompatibilityLogger.LogMessage(
+                        MessageImportance.High,
+                        "Either fix breaking change, or update compatibility suppression files, rebuild with \\p:GenerateCompatibilitySuppressions or --generate-suppression-file.");
+                }
+            }
+            else
+            {
+                serviceProvider.CompatibilityLogger.LogMessage(
+                    MessageImportance.High,
```
ViktorHofer, line 103: Lower the verbosity in order to not print the message in the Task frontend.

```suggestion
                    MessageImportance.Normal,
```

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
```diff
@@ -77,6 +78,23 @@ internal static class ValidatePackage
...
+                    {
+                        serviceProvider.CompatibilityLogger.LogMessage(
+                        MessageImportance.High,
+                        "Either fix breaking change, or update compatibility suppression files, rebuild with \\p:GenerateCompatibilitySuppressions or --generate-suppression-file.");
+                    }
+                }
+                else
+                {
+                    serviceProvider.CompatibilityLogger.LogMessage(
+                        MessageImportance.High,
```
ViktorHofer, line 94:
```suggestion
                        MessageImportance.Normal,
```

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
```diff
@@ -87,6 +88,23 @@ internal static class ValidateAssemblies
             // Execute the enqueued work item(s).
             apiCompatRunner.ExecuteWorkItems();
 
+            if (serviceProvider.CompatibilityLogger.SuppressionWasLogged)
+            {
+                if (!generateSuppressionFile)
+                {
+                    serviceProvider.CompatibilityLogger.LogMessage(
+                        MessageImportance.High,
+                        "Either fix breaking change, or update compatibility suppression files, rebuild with \\p:GenerateCompatibilitySuppressions or --generate-suppression-file.");
```
ViktorHofer, line 97: Please localize the message (there's already a Resources.resx in the Shared project)

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
```diff
@@ -77,6 +78,23 @@ internal static class ValidatePackage
...
                 // Execute the work items that were enqueued.
                 serviceProvider.ApiCompatRunner.ExecuteWorkItems();
+
+                if (serviceProvider.CompatibilityLogger.SuppressionWasLogged)
+                {
+                    if (!generateSuppressionFile)
+                    {
+                        serviceProvider.CompatibilityLogger.LogMessage(
+                        MessageImportance.High,
+                        "Either fix breaking change, or update compatibility suppression files, rebuild with \\p:GenerateCompatibilitySuppressions or --generate-suppression-file.");
```
ViktorHofer, line 88: Please localize the message.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/ConsoleCompatibilityLogger.cs
```diff
@@ -44,10 +46,14 @@ private bool LogSuppressableMessage(bool errorOutput, Suppression suppression, s
...
 
+            _suppressionWasLogged = true;
+
             TextWriter textWriter = errorOutput ? Console.Error : Console.Out;
             textWriter.WriteLine(code + ": " + string.Format(format, args));
 
             return true;
         }
+
+        public bool SuppressionWasLogged => _suppressionWasLogged;
```
ViktorHofer, line 57:
```suggestion
        /// <inheritdoc />
        public bool SuppressionWasLogged { get; private set; }
```

src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/ConsoleCompatibilityLogger.cs
```diff
@@ -14,12 +14,14 @@ internal sealed class ConsoleCompatibilityLogger : ICompatibilityLogger
...
         private readonly ISuppressionEngine _suppressionEngine;
         private readonly MessageImportance _messageImportance;
+        private bool _suppressionWasLogged;
 
         public ConsoleCompatibilityLogger(ISuppressionEngine suppressionEngine,
             MessageImportance messageImportance)
         {
             _suppressionEngine = suppressionEngine;
             _messageImportance = messageImportance;
+            _suppressionWasLogged = false;
```
ViktorHofer, line 24:
```suggestion
```

src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/ConsoleCompatibilityLogger.cs
```diff
@@ -14,12 +14,14 @@ internal sealed class ConsoleCompatibilityLogger : ICompatibilityLogger
     {
         private readonly ISuppressionEngine _suppressionEngine;
         private readonly MessageImportance _messageImportance;
+        private bool _suppressionWasLogged;
```
ViktorHofer, line 17:
```suggestion
```

src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/MSBuildCompatibilityLogger.cs
ViktorHofer: You don't need to save a state for the msbuild logger as it already has a property on the `_log` field, called `HasLoggedAnyErrors` (or similar).

ViktorHofer (commented):
A few high-level observations:
- We want to log a message in high importance only if we have unhandled compatibility differences. Otherwise, we can log a message but that shouldn't be done with high importance to avoid noise. We should also investigate if we want to downgrade the default console verbosity to include "normal" importance messages by default.
- I wonder if the suppression engine should actually be the place to know if an unhandled compatibility difference exists. That would likely also require changing how we emit suppressions today to the frontend.
- Please use RESX localization infrastructure.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/MSBuildCompatibilityLogger.cs
```diff
@@ -40,13 +38,11 @@ private bool LogSuppressableMessage(MessageLevel messageLevel, Suppression suppr
...
 
-            _suppressionWasLogged = true;
-
             _log.Log(new Message(messageLevel, string.Format(format, args), code));
 
             return true;
         }
 
-        public bool SuppressionWasLogged => _suppressionWasLogged;
+        public bool SuppressionWasLogged => _log.HasLoggedErrors;
```
ViktorHofer, line 46: Sorry for the confusion. Can you please revert this change to what you had before? Asking as with the proposed logging consolidation with GenAPI we will be able to log errors that aren't related to suppressions.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidateAssemblies.cs
```diff
@@ -87,6 +88,22 @@ internal static class ValidateAssemblies
...
+                        MessageImportance.High,
+                        CommonResources.BreakingChangesFound);
+                }
+            }
+            else
+            {
+                serviceProvider.CompatibilityLogger.LogMessage(
+                    MessageImportance.Normal,
+                    CommonResources.NoBreakingChangesFound);
+            }
```
ViktorHofer, line 105: Should we move this into a helper function similar to the below `GenerateSuppressionFile` helper function to avoid the repetition between the two tasks?

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/CommonResources.resx
```diff
@@ -117,12 +117,18 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="BreakingChangesFound" xml:space="preserve">
+    <value>Either fix breaking change, or update compatibility suppression files, rebuild with \p:GenerateCompatibilitySuppressions or --generate-suppression-file.</value>
```
ViktorHofer, line 121: I would rephrase this message and start with "API compatibility differences found. If those are intentional, the APICompat baseline file can be updated by running `dotnet build /p:ApiCompatGenerateSuppressionFile=true`.

ViktorHofer: Of course, as I'm not a native speaker, feel free to refine my suggestion.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/CommonResources.resx
```diff
@@ -117,12 +117,18 @@
...
+  </data>
   <data name="CreateWorkItemPerAssemblyAssembliesNotEqual" xml:space="preserve">
     <value>When enqueuing work items per assembly, the number of passed in left and right assemblies must be equal.</value>
   </data>
   <data name="InvalidRexegStringTransformationPattern" xml:space="preserve">
     <value>Invalid transformation pattern provided: '{0} - {1}'.</value>
   </data>
+  <data name="NoBreakingChangesFound" xml:space="preserve">
+    <value>Package validation was successful. No breaking changes.</value>
```
ViktorHofer, line 130: This isn't just about Package Validation. The message will also be printed when assemblies are being compared. I would rephrase to `APICompat ran successfully without finding any breaking changes.`.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/CommonResources.resx
```diff
@@ -117,12 +117,18 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="BreakingChangesFound" xml:space="preserve">
+    <value>API breaking changes found. If those are intentional, the APICompat suppression file can be updated by running dotnet build /p:GenerateCompatibilitySuppressions=true.</value>
```
ViktorHofer, line 121: The MSBuild property was just recently renamed to `ApiCompatGenerateSuppressionFile`: https://github.com/dotnet/sdk/blob/39fcce04a117339a8087060201b1f7797b6167e8/src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/build/Microsoft.DotNet.ApiCompat.ValidateAssemblies.Common.targets#L20

The old property still works for compat reasons but let's use the new one going forward.

ViktorHofer:
```suggestion
    <value>API breaking changes found. If those are intentional, the APICompat suppression file can be updated by running "dotnet build /p:ApiCompatGenerateSuppressionFile=true".</value>
```

ViktorHofer: This command only applies to the MSBuild task frontend. The CLI requires a different command. Can you please add something like "or --generate-suppression-file"? I think you already had that before.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/SuppressionFileHelper.cs
```diff
@@ -32,5 +32,25 @@ internal static class SuppressionFileHelper
                 log.LogMessage(MessageImportance.High, CommonResources.WroteSuppressions, suppressionOutputFile);
             }
         }
+
+        // Log whether or not we found breaking changes. If we are writing to a suppression file, no need to log anything.
+        public static void LogAPICompatSuccessOrFailure(bool generateSuppressionFile, ApiCompatServiceProvider serviceProvider)
```
ViktorHofer, line 37: nit: Please don't flow the entire ApiCompatServiceProvider into this method. It has access to much more than what we need here. Instead consider just flowing the compatibility logger interface in.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/SuppressionFileHelper.cs
```diff
@@ -32,5 +32,25 @@ internal static class SuppressionFileHelper
                 log.LogMessage(MessageImportance.High, CommonResources.WroteSuppressions, suppressionOutputFile);
             }
         }
+
+        // Log whether or not we found breaking changes. If we are writing to a suppression file, no need to log anything.
```
ViktorHofer, line 36: Please make this a tripple slash xml doc instead as it's a public method.

ViktorHofer (approved):
Left some more feedback. Aside from that, looks good.

Issue #25868, opened 2022-06-06
The --launch-profile option of `dotnet run` and `dotnet watch run` should fallback to a case-insensitive match if a case-sensitive match isn't found


DamianEdwards:
Today, `dotnet run --launch-profile profilename` and `dotnet watch run --launch-profile profilename` are case-sensitive and if a matching profile is not found in `launchSettings.json` an error message is displayed but the project is run anyway, e.g.:

**launchSettings.json**

```json
{
  "profiles": {
    "http": {
      "commandName": "Project",
      "dotnetRunMessages": true,
      "launchBrowser": true,
      "applicationUrl": "http://localhost:5130",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      }
    },
    "HTTPS": {
      "commandName": "Project",
      "dotnetRunMessages": true,
      "launchBrowser": true,
      "applicationUrl": "https://localhost:7175;http://localhost:5130",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      }
    }
  }
}
```

```console
$ dotnet run --launch-profile https
The launch profile "https" could not be applied.
A profile with the specified name could not be found or is not a valid JSON object.
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5000
```

It would be more user-friendly in cases like this to simply fallback to a case-insensitive match if a case-sensitive match is not found first.

DamianEdwards:
VS shows them as distinct profiles 
<img width="337" alt="image" src="https://user-images.githubusercontent.com/249088/173685152-fe7c0a4d-de71-415e-bbe7-32361e2ab0a2.png">

baronfel:
What about something like this:
* if the name is case-insensitively unique (i.e. `http` in your image above), then accept any case
* if the name has case-sensitive variants, require case-sensitive matches or else error
* completions always suggest the user-entered spellings exactly

DamianEdwards:
Yep that's inline with what I was thinking (fallback to case-insensitive if case-sensitive is not found).

nagilson:
>  I do think this would be a good area for @nagilson after he tackles the various Release configuration/RID inference changes for 7 if nothing else crazy pops up.

Ah yes, something crazy  _definitely_ did NOT pop up. This was a good suggestion and while I'm sorry for the delay, glad I finally had time to work on this.

nagilson:
[...]

PR #28908, opened 2022-11-05
Support calling unique launch profiles with a case-insensitive nature & Improve errors around launch profiles

nagilson:
# Changes:
- Separates the error for a corrupt json file and a non-existent launch profile
- Adds a more specific error for if the launch settings WAS found but the Profile was not found 
image

image
(The last sentence is no longer there ^) 

- The new no launch profile error specifies where it tried to look for the launch profile
- If you have one launch profile like https and try to call it with HTTPS it will be OK
image

- If you have multiple launch profiles like https and HTTPS and try to use the profile, you will get an error

image

Resolves: https://github.com/dotnet/sdk/issues/25868

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
@@ -60,8 +60,8 @@ namespace Microsoft.DotNet.Tools.Run.LaunchSettings
 
                             if (caseInsensitiveProfileMatches.Count() > 1)
                             {
-                                return new LaunchSettingsApplyResult(false, string.Format(LocalizableStrings.DuplicateCaseInsensitiveLaunchProfileNames,
-                                    String.Join(",", caseInsensitiveProfileMatches.ToArray())));
+                                throw new GracefulException(LocalizableStrings.DuplicateCaseInsensitiveLaunchProfileNames,
+                                    String.Join(",", caseInsensitiveProfileMatches.ToArray()));
                             }
                             else if (!caseInsensitiveProfileMatches.Any())
                             {
--- a/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
@@ -52,9 +52,30 @@ namespace Microsoft.DotNet.Tools.Run.LaunchSettings
                     }
                     else
                     {
-                        if (!profilesObject.TryGetProperty(profileName, out profileObject) || profileObject.ValueKind != JsonValueKind.Object)
+                        if (!profilesObject.TryGetProperty(profileName, out profileObject))
                         {
-                            return new LaunchSettingsApplyResult(false, LocalizableStrings.LaunchProfileIsNotAJsonObject);
+                            IEnumerable<JsonProperty> caseInsensitiveProfileMatches = profilesObject
+                                .EnumerateObject()
+                                .Where(p => string.Compare(p.Name, profileName, StringComparison.OrdinalIgnoreCase) == 0); // p.Name should never fail, as profileObject enumerables are only created from a JsonObject.
+
+                            if (caseInsensitiveProfileMatches.Count() > 1)
+                            {
+                                return new LaunchSettingsApplyResult(false, string.Format(LocalizableStrings.DuplicateCaseInsensitiveLaunchProfileNames,
+                                    String.Join(",", caseInsensitiveProfileMatches.ToArray())));
+                            }
+                            else if (!caseInsensitiveProfileMatches.Any())
+                            {
+                                return new LaunchSettingsApplyResult(false, string.Format(LocalizableStrings.LaunchProfileDoesNotExist, profileName));
+                            }
+                            else
+                            {
+                                profileObject = profilesObject.GetProperty(caseInsensitiveProfileMatches.First().Name);
+                            }
+
+                            if (profileObject.ValueKind != JsonValueKind.Object)
+                            {
+                                return new LaunchSettingsApplyResult(false, LocalizableStrings.LaunchProfileIsNotAJsonObject);
+                            }
                         }
                     }
 
--- a/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
@@ -54,7 +54,8 @@ namespace Microsoft.DotNet.Tools.Run.LaunchSettings
                     {
                         IEnumerable<JsonProperty> caseInsensitiveProfileMatches = profilesObject
                             .EnumerateObject() // p.Name shouldn't fail, as profileObject enumerables here are only created from an existing JsonObject
-                            .Where(p => string.Compare(p.Name, profileName, StringComparison.OrdinalIgnoreCase) == 0);
+                            .Where(p => string.Equals(p.Name, profileName, StringComparison.OrdinalIgnoreCase))
+                            .ToList();
 
                         if (caseInsensitiveProfileMatches.Count() > 1)
                         {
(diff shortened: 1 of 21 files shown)


baronfel:
> completions always suggest the user-entered spellings exactly

I should have phrased this differently. It should be more like

> completions for the 'launch profile' option should be the exact profile names json keys from the launch settings file

The intent here is that the CLI wouldn't change the user-entered names at all.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
```diff
@@ -50,9 +50,27 @@ public static LaunchSettingsApplyResult TryApplyLaunchSettings(string launchSett
                             .EnumerateObject()
                             .FirstOrDefault(IsDefaultProfileType).Value;
                     }
-                    else
+                    else // Find a profile match for the given profileName
                     {
-                        if (!profilesObject.TryGetProperty(profileName, out profileObject) || profileObject.ValueKind != JsonValueKind.Object)
+                        IEnumerable<JsonProperty> caseInsensitiveProfileMatches = profilesObject
```
nagilson, line 55: It would be nice to refactor this into its own function, but that is a bit troublesome due to the required `out` parameter `profileObject`, potentially returning `null` or not for the launch settings application failure, and the fact that this interface is `static`. It turned out to be messier than not to have this code all-inlined.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
```diff
@@ -50,9 +50,27 @@ public static LaunchSettingsApplyResult TryApplyLaunchSettings(string launchSett
...
+                            .EnumerateObject() // p.Name shouldn't fail, as profileObject enumerables here are only created from an existing JsonObject
+                            .Where(p => string.Compare(p.Name, profileName, StringComparison.OrdinalIgnoreCase) == 0);
+
+                        if (caseInsensitiveProfileMatches.Count() > 1)
```
nagilson, line 59: This could be converted into a switch statement, but that doesn't seem too valuable.

src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
```diff
@@ -173,7 +173,7 @@ The current {1} is '{2}'.</value>
     <value>Arguments passed to the application that is being run.</value>
   </data>
   <data name="RunCommandExceptionCouldNotLocateALaunchSettingsFile" xml:space="preserve">
-    <value>The specified launch profile could not be located.</value>
+    <value>The specified launch profile '{0}' could not be located. Is it of the form ./Properties/launchSettings.json?</value>
```
nagilson, line 176: `Is it of the form ./Properties/launchSettings.json?` This should be removed as you can have a custom output path and I assume we don't expect people to use file-based launch profiles that often (VS is the interface between the two)

src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
```diff
@@ -186,7 +186,7 @@ The current {1} is '{2}'.</value>
     <value>Using launch settings from {0}...</value>
   </data>
   <data name="LaunchProfileIsNotAJsonObject" xml:space="preserve">
-    <value>A profile with the specified name could not be found or is not a valid JSON object.</value>
+    <value>A profile with the specified name is not a valid JSON object.</value>
```
nagilson, line 189: Nit: contraction, is not -> isn't

src/Cli/dotnet/commands/dotnet-run/LocalizableStrings.resx
```diff
@@ -226,4 +226,12 @@ The current {1} is '{2}'.</value>
   <data name="OnlyOneProjectAllowed" xml:space="preserve">
     <value>Only one project can be specified at a time using the `-p` option.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="DuplicateCaseInsensitiveLaunchProfileNames" xml:space="preserve">
+    <value>There are several launch profiles with case-sensitive names, which is not permitted:
```
nagilson, line 230: Nit: Contraction is not -> isn't

src/Tests/dotnet-run.Tests/GivenDotnetRunRunsVbProj.cs
```diff
@@ -27,31 +29,75 @@ public void ItGivesAnErrorWhenAttemptingToUseALaunchProfileThatDoesNotExistWhenT
...
-            new DotnetCommand(Log, "run")
+            var runResult = new DotnetCommand(Log, "run")
                 .WithWorkingDirectory(testProjectDirectory)
-                .Execute("--launch-profile", "test")
-                .Should().Pass()
-                         .And.HaveStdOutContaining("Hello World!")
-                         .And.HaveStdErrContaining(LocalizableStrings.RunCommandExceptionCouldNotLocateALaunchSettingsFile);
+                .Execute("--launch-profile", "test");
+
+            string[] expectedErrorWords = LocalizableStrings.RunCommandExceptionCouldNotLocateALaunchSettingsFile.Replace("\'{0}\'", "").Split(" ");
```
nagilson, line 36: For this test in particular, you can probably just format the profile name in instead of removing the {0}

nagilson: Ah, jk this is a path test that's duplicated for VB and CS projs.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
```diff
@@ -50,9 +50,27 @@ public static LaunchSettingsApplyResult TryApplyLaunchSettings(string launchSett
                             .EnumerateObject()
                             .FirstOrDefault(IsDefaultProfileType).Value;
                     }
-                    else
+                    else // Find a profile match for the given profileName
                     {
-                        if (!profilesObject.TryGetProperty(profileName, out profileObject) || profileObject.ValueKind != JsonValueKind.Object)
+                        IEnumerable<JsonProperty> caseInsensitiveProfileMatches = profilesObject
+                            .EnumerateObject() // p.Name shouldn't fail, as profileObject enumerables here are only created from an existing JsonObject
+                            .Where(p => string.Compare(p.Name, profileName, StringComparison.OrdinalIgnoreCase) == 0);
```
dsplaisted, line 57: Nit: I think `Equals` is easier to read than `Compare`.

Also, it's probably a good idea to add a `ToList()` to the end of this so that the enumeration / comparison doesn't have to run multiple times (for the `Count()` call, then the `Any()` call, then the `First()` call).

```suggestion
                            .Where(p => string.Equals(p.Name, profileName, StringComparison.OrdinalIgnoreCase))
                            .ToList();
```

src/Tests/dotnet-run.Tests/GivenDotnetRunRunsCsProj.cs
```diff
@@ -379,12 +380,18 @@ public void ItGivesAnErrorWhenAttemptingToUseALaunchProfileThatDoesNotExistWhenT
...
-            new DotnetCommand(Log, "run")
+            var runResult = new DotnetCommand(Log, "run")
                 .WithWorkingDirectory(testProjectDirectory)
-                .Execute("--launch-profile", "test")
-                .Should().Pass()
-                         .And.HaveStdOutContaining("Hello World!")
-                         .And.HaveStdErrContaining(LocalizableStrings.RunCommandExceptionCouldNotLocateALaunchSettingsFile);
+                .Execute("--launch-profile", "test");
+
+            string[] expectedErrorWords = LocalizableStrings.RunCommandExceptionCouldNotLocateALaunchSettingsFile.Replace("\'{0}\'", "").Split(" ");
```
dsplaisted, line 387: Would it be simpler just to format the message with the expected parameters and verify that the output contains that, instead of checking that it has all the words (but not verifying order or anything)?

dsplaisted (approved):

src/Tests/dotnet-run.Tests/GivenDotnetRunRunsCsProj.cs
nagilson: This is a good option, though it would require extracting `testRoot`, specifying the TFM, and hardcoding parts of the launchSettings path into the test. I prefer this option for this test since that information isn't relevant to the test, makes the test more flexible to changes in the asset, and can be done in less lines of code

PR #28934, opened 2022-11-08
Support multiple suppression files

### LINKED ISSUE #26746: [ApiCompat] Support passing in multiple suppression files
see title. See https://github.com/dotnet/runtime/issues/72595 in which CoreLib would want to provide multiple suppression files.

We probably also want to be able to write to a different suppression file than to the ones that we pass in to APICompat.

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/26746

Changes the frontends to allow multiple suppression files and writing to an output suppression path while retaining the existing msbuild properties and items to not break existing customers.


smasher164 (approved):
LGTM

Issue #28891, opened 2022-11-03
Project References that are not listed in assets file aren't included in deps.json


wrall:
### Describe the bug
When there is a reference to a project that is not included in the project.assets.json file for some reason (e.g. it has opted out of nuget restore), it does not get included into deps.json.

### To Reproduce
<ProjectReference Include="ProjectA.csproj" />, and opt ProjectA out of using NuGet (e.g. force _IsProjectRestoreSupported target to return nothing).

### Exceptions (if any)
None.

### Further technical details
Happens on all versions.

PR #28963, opened 2022-11-09
Deps.json should include project references that aren't present in project.assets.json

wrall:
Addresses #28891.

Although most projects are included in project.assets.json after a restore takes place, there are some (rare) scenarios when this is not the case. When that happens, project references end up being skipped and excluded from deps.json.

One can work around this by adding a binary Reference, that is inconvenient and can lead to issues when one forgets. Instead, it would be good to check whether the project reference is included in project.assets.json before skipping it when generating the list of dependencies.

Also added a functional test (.NET 6.0 project referencing a .NET Framework project) that failed previously and succeeds now.

(Continues #28892, retargeting release/7.0.2xx)


wrall:
(@marcpopMSFT @dsplaisted FYI.)

src/Tasks/Microsoft.NET.Build.Tasks/ReferenceInfo.cs
```diff
@@ -54,17 +54,36 @@ public static IEnumerable<ReferenceInfo> CreateReferenceInfos(IEnumerable<ITaskI
...
+                }
+
+                string projectName = referencePath.GetMetadata(MetadataKeys.MSBuildSourceProjectFile);
+                if (string.IsNullOrEmpty(projectName))
+                {
+                    projectName = Path.GetFileNameWithoutExtension(referencePath.ItemSpec);
+                    if (string.IsNullOrEmpty(projectName))
+                    {
+                        return true;
+                    }
```
dsplaisted, line 79: Under what conditions would this be null or empty, and should we be returning true here?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAnAppWithTransitiveNonSdkProjectRefs.cs
```diff
@@ -0,0 +1,192 @@
...
+                        string projectFileName = Path.GetFileNameWithoutExtension(projectName);
+                        if (StringComparer.OrdinalIgnoreCase.Equals(projectFileName, "AuxLibrary") ||
+                            StringComparer.OrdinalIgnoreCase.Equals(projectFileName, "MainLibrary"))
+                        {
+                            var ns = project.Root.Name.Namespace;
+
+                            if (!string.IsNullOrEmpty(prefix))
+                            {
+                                XElement propertyGroup = project.Root.Element(XName.Get("PropertyGroup", ns.NamespaceName));
+                                XElement assemblyName = propertyGroup.Element(XName.Get("AssemblyName", ns.NamespaceName));
```
dsplaisted, line 61: I believe using the namespace can be simplified here:

```suggestion
                                XElement propertyGroup = project.Root.Element(ns + "PropertyGroup");
                                XElement assemblyName = propertyGroup.Element(ns + "AssemblyName");
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAnAppWithTransitiveNonSdkProjectRefs.cs
```diff
@@ -0,0 +1,192 @@
...
+                                assemblyName.RemoveAll();
+                                assemblyName.Add("TestApp." + projectFileName);
```
dsplaisted, line 63: Nit: Looks like it's not actually using the value of the parameter here.  Probably this should be `prefix + projectFileName`, or the parameter should be a `bool` specifying whether assembly name should be different from the project file.

dsplaisted (approved):
This is looking great, thanks!

dsplaisted:
Looks like there are some test failures in CI: `GivenThatWeWantToPublishAProjectWithAllFeatures.It_fails_when_nobuild_is_set_and_build_was_not_performed_previously` and `GivenThatWeWantToPreserveCompilationContextForBuild.It_supports_copylocal_false_references`.  Let us know if you need help fixing those.

Thanks!

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAnAppWithTransitiveNonSdkProjectRefs.cs
wrall: Yep, you're right.  Fixed

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildAnAppWithTransitiveNonSdkProjectRefs.cs
wrall: Yep!

wrall:
@dsplaisted said:
> Looks like there are some test failures in CI: `GivenThatWeWantToPublishAProjectWithAllFeatures.It_fails_when_nobuild_is_set_and_build_was_not_performed_previously` and `GivenThatWeWantToPreserveCompilationContextForBuild.It_supports_copylocal_false_references`.

These should be fixed now - there was an issue with how I resolved some of the comments from the last PR.

wrall:
FYI, latest update was to hide the new behavior behind a new MSBuild property.  (Note that the test case I added fails without the new switch being provided, but I did not add a negative test).

StephenBonikowsky:
@dsplaisted and @marcpopMSFT is this ready to be merged?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -125,6 +125,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <MetadataUpdaterSupport Condition="'$(MetadataUpdaterSupport)' == '' and '$(Configuration)' != 'Debug'">false</MetadataUpdaterSupport>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Turn off IncludeProjectsNotInAssetsFileInDepsFile by default. -->
+    <IncludeProjectsNotInAssetsFileInDepsFile Condition="'$(IncludeProjectsNotInAssetsFileInDepsFile)' == ''">false</IncludeProjectsNotInAssetsFileInDepsFile>
```
dsplaisted, line 130: I'd suggest changing the default value here to true for the change that goes in 7.0.2xx (the next feature release).

Then we can backport the change to servicing branches, and in those we should set the default to false.

Otherwise, I think this looks all ready to merge.  Thanks!

dsplaisted (approved):

marcpopMSFT (approved):

PR #28967, opened 2022-11-09
Introduce Roslyn-based GenAPI

andriipatsula:
The GenAPI is a set of tools (a MSBuild task, a CLI tool) to synthesize C# source tree out of a DLL. Originally the GenAPI tool was based on the CCi and live in dotnet/arcade repository. The CCI is deprecated, and the main idea is to utilize Roslyn API to write new tool from scratch and revise the list of options required for Source Build Reference Package generator and for the Runtime team.


andriipatsula:
@ViktorHofer , @smasher164 here is a PR for Roslyn-based GenAPI. I am still working on UTs and continue working on functionality. Please take a look.

andriipatsula:
@ViktorHofer, as we agreed, the first commit is the version of GenAPI from dotnet/arcade repository (the one I was working on)

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.CommandLine.Parsing;
+using Microsoft.DotNet.GenAPI.Shared;
+
```
ViktorHofer, line 8:
```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Microsoft.DotNet.GenAPI.Tool.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(DotNetBuildFromSource)' == 'true'">net7.0</TargetFrameworks>
```
ViktorHofer, line 4:
```suggestion
    <TargetFramework>net6.0</TargetFramework>
    <TargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net7.0</TargetFramework>
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Microsoft.DotNet.GenAPI.Shared.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
ViktorHofer, line 1: This project isn't really a Visual Studio Shared project. Instead, it's just a class library. Only use the "Shared" name suffix if it's a VS Shared project. I think it makes sense to keep this a class library and rename the project and folder from `Microsoft.DotNet.GenAPI.Shared` to `Microsoft.DotNet.GenAPI.Core`.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Microsoft.DotNet.GenAPI.Shared.csproj
```diff
@@ -0,0 +1,23 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(DotNetBuildFromSource)' == 'true'">net7.0</TargetFrameworks>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
ViktorHofer, line 5: We want to consume the library in a source build environment.

```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/SyntaxGeneratorExtensions.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
```
ViktorHofer, line 4:
```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/SyntaxGeneratorExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+                                type.Name,
+                                accessibility: type.DeclaredAccessibility,
+                                interfaceTypes: type.Interfaces.Select(i => syntaxGenerator.TypeExpression(i)));
+                        case TypeKind.Enum:
+                            return syntaxGenerator.EnumDeclaration(
+                                type.Name,
+                                accessibility: type.DeclaredAccessibility);
+                    }
+                    break;
+            }
```
ViktorHofer, line 50: The outer switch/case is weird, as it only does logic if `symbol.Kind == SymbolKind.NamedType`. I would just use an `if` for that instead.

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        static int Main(string[] args)
+        {
+            // Global options
+            Option<string> assemblyOption = new("--assembly",
+                "Delimited (',' or ';') set of paths for assemblies or directories to get all assemblies.")
+            {
+                ArgumentHelpName = "file,dir,...",
+                IsRequired = true
+            };
```
ViktorHofer, line 25: System.CommandLine should already take care of parsing the input (which includes splitting multiple values that are separated by a delimiter). Can you please follow what we do in APICompat: https://github.com/dotnet/sdk/blob/4bd9b67d81fbf6d65012ef08c9d8cf2a27953057/src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs#L47-L49

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    /// <summary>
+    /// CLI frontend for the Roslyn-based GenAPI.
+    /// </summary>
+    class Program
+    {
+        static int Main(string[] args)
+        {
+            // Global options
+            Option<string> assemblyOption = new("--assembly",
```
ViktorHofer, line 20: Assembly is a required argument today in the existing GenAPI: https://github.com/dotnet/arcade/blob/7ab9e65d08b9a31183684781e21c3195f91ea1a8/src/Microsoft.DotNet.GenAPI/GenAPITask.cs#L42

Wouldn't you use `Argument<>` here instead of `Option<>`?

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            return pathSet.Split(new char[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries);
+        }
+
+        /// <summary>
+        /// Read the header file if specified, or use default one.
+        /// </summary>
+        /// <param name="headerFile">File with an alternate header content to prepend to output</param>
+        /// <returns></returns>
+        public static string ReadHeaderFile(string? headerFile)
```
ViktorHofer, line 147: Although the method isn't used today, I would assume that it should be private.

```suggestion
        private static string ReadHeaderFile(string? headerFile)
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// Splits delimiter separated list of pathes represented as a string to a List of paths.
+        /// </summary>
+        /// <param name="pathSet">Delimiter separated list of paths.</param>
+        /// <returns></returns>
+        private static string[] SplitPaths(string? pathSet)
+        {
+            if (pathSet == null) return Array.Empty<string>();
+
+            return pathSet.Split(new char[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries);
+        }
```
ViktorHofer, line 140: As mentioned in the frontend code, we shouldn't do this here and instead let the frontend do the parsing and accept a string array instead for all parameters that potentially contain multiple values.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            /// <summary>
+            /// Delimited (',' or ';') set of paths for assemblies or directories to get all assemblies.
+            /// </summary>
+            public string? Assembly { get; set; }
+
+            /// <summary>
+            /// If true, tries to resolve assembly reference.
+            /// </summary>
+            public bool ResolveAssemblyReferences { get; set; } = false;
```
ViktorHofer, line 29: False is already the default for a boolean. No need to explicitly set it.

```suggestion
            public bool ResolveAssemblyReferences { get; set; }
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
```diff
@@ -0,0 +1,178 @@
...
+    ///     Shared between CLI and MSBuild tasks frontends.
+    /// </summary>
+    public static class GenAPIApp
+    {
+        public class Context
+        {
+            /// <summary>
+            /// Delimited (',' or ';') set of paths for assemblies or directories to get all assemblies.
+            /// </summary>
+            public string? Assembly { get; set; }
```
ViktorHofer, line 24: Shouldn't be nullable but an array instead.

```suggestion
            public string[] Assembly { get; set; }
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
```diff
@@ -0,0 +1,178 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiSymbolExtensions;
+
+namespace Microsoft.DotNet.GenAPI.Shared
+{
+
```
ViktorHofer, line 12:
```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Filters/FilterOutAttributes.cs
```diff
@@ -0,0 +1,53 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.IO;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiSymbolExtensions;
+
+namespace Microsoft.DotNet.GenAPI.Shared
+{
+    internal class FilterOutAttributes : ISymbolFilter
```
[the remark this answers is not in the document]
smasher164: Yes, we should just move that code out of the AttributesMustMatch rule into this shared class.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
andriipatsula: I wanted to have as much as possible common code in one place. Now, I'd need to have SplitPaths in both frontends or introduce some utility class for that. Updated

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Microsoft.DotNet.GenAPI.Tool.csproj
ViktorHofer: As GenAPI is not part of the SDK, yes. But ApiSymbolExtensions and anything ApiCompat related needs to stay on net6.0 and net472.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Microsoft.DotNet.GenAPI.Shared.csproj
ViktorHofer: Yes as we want to be able to use it during source build in dotnet/runtime.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/GenAPIApp.cs
ViktorHofer: System.CommandLine takes care of that for you. You only need to do this in the msbuild frontend. Take a look at how we solved this in the APICompat frontends.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Filters/FilterOutAttributes.cs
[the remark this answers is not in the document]
ViktorHofer: > Updated, but modified to !id.StartsWith("#") && !id.StartsWith("//") -> id.StartsWith("T:"). And moved to Microsoft.DotNet.ApiSymbolExtensions

We need the existing behavior in ApiCompat as attributes not just exist on types but also on members and parameters.

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -0,0 +1,69 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Build.Framework;
+using Microsoft.DotNet.Build.Tasks;
+using Microsoft.DotNet.GenAPI.Shared;
+
+namespace Microsoft.DotNet.GenAPI.Tasks;
+
+#nullable enable
```
ViktorHofer, line 10: Turn on `<Nullable>enable</Nullable>` for the entire project and remove this line.

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -0,0 +1,69 @@
...
+/// <summary>
+/// MSBuild task frontend for the Roslyn-based GenAPI.
+/// </summary>
+public class GenAPITask : BuildTask
+{
+    /// <summary>
+    /// Delimited (',' or ';') set of paths for assemblies or directories to get all assemblies.
+    /// </summary>
+    [Required]
+    public string? Assembly { get; set; }
```
ViktorHofer, line 21: This should be a string array and named plural to indicate that multiple assemblies can be passed in:

```suggestion
    public string[]? Assemblies { get; set; }
```

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// <summary>
+    /// If true, tries to resolve assembly reference.
+    /// </summary>
+    public bool? ResolveAssemblyReferences { get; set; }
```
ViktorHofer, line 26: Booleans very rarely need to be nullable. You don't need a third state here, false and true should be sufficient:

```suggestion
    public bool ResolveAssemblyReferences { get; set; }
```

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -0,0 +1,69 @@
...
+    
+    /// <summary>
+    /// Delimited (',' or ';') set of paths to use for resolving assembly references.
+    /// </summary>
+    public string? LibPath { get; set; }
```
ViktorHofer, line 31: Same as assemblies, should be a string array and named plural. Also, would it make sense to follow what we have in ApiCompat by naming it `AssemblyReferences`?

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// <summary>
+    /// Method bodies should throw PlatformNotSupportedException.
+    /// </summary>
+    public string? ExceptionMessage { get; set; }
+
+    /// <summary>
+    /// Specify a path to a file with the list in the DocId format of which attributes should be excluded from being applied on apis.
+    /// </summary>
+    public string? ExcludeAttributesList { get; set; }
```
ViktorHofer, line 52: Allow passing in multiple attributes exclusion files. Please follow what we already have in ApiCompat: https://github.com/dotnet/sdk/blob/f14f84010ae85a54d557fe735bb4bcc14bcd5382/src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/ValidateAssembliesTask.cs#L57-L60

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(TargetFrameworkForNETSDK)</TargetFrameworks>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
ViktorHofer, line 5: The task should be source buildable

```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PackageType>MSBuildSdk</PackageType>
```
ViktorHofer, line 6: We don't want this package to be an msbuild sdk package. `Microsoft.NET.Sdk` or `Microsoft.Build.NoTargets` are examples of msbuild sdks. This package is just a tool that will be consumed  via msbuild props and targets files but not as an sdk.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,185 @@
...
+    ///     Shared between CLI and MSBuild tasks frontends.
+    /// </summary>
+    public static class GenAPIApp
+    {
+        public sealed class Context
+        {
+            /// <summary>
+            /// The path to one or more assemblies or directories with assemblies.
+            /// </summary>
+            public string[] Assembly { get; set; } = new string[0];
```
ViktorHofer, line 23:
```suggestion
            public required string[] Assemblies { get; set; }
```

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,185 @@
...
+
+namespace Microsoft.DotNet.GenAPI
+{
+    /// <summary>
+    /// Class to standertize initilization and running of GenAPI tool.
+    ///     Shared between CLI and MSBuild tasks frontends.
+    /// </summary>
+    public static class GenAPIApp
+    {
+        public sealed class Context
```
ViktorHofer, line 18: I don't think that this extra class is necessary, I would just pass all these in as parameters to GenAPIApp but if you want to keep it, make it a readonly struct instead (for better perf).

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            /// <summary>
+            /// The path to one or more attribute exclusion files with types in DocId format.
+            /// </summary>
+            public string[]? ExcludeAttributesFiles { get; set; }
+
+            /// <summary>
+            /// Include all API's not just public APIs. Default is public only.
+            /// </summary>
+            public bool IncludeVisibleOutsideOfAssembly { get; set; } = false;
```
ViktorHofer, line 69: nit: boolean already defaults to false

```suggestion
            public bool IncludeVisibleOutsideOfAssembly { get; set; }
```

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// <summary>
+        /// Initialize and run Roslyn-based GenAPI tool.
+        /// </summary>
+        public static void Run(Context context)
+        {
+            IAssemblySymbolLoader loader = new AssemblySymbolLoader(context.ResolveAssemblyReferences);
+
+            if (context.LibPath is not null)
+            {
+                loader.AddReferenceSearchPaths(context.LibPath);
```
ViktorHofer, line 81: Assembly references can be different per assembly. Do you expect that the passed in references are a super-set or should GenAPI be able to handle different references per assembly? I.e. consider when both a net6.0 and net7.0 compiled assembly is passed in. Those would have entirely different assembly references.

src/GenAPI/Microsoft.DotNet.GenAPI/Filters/IntersectionFilter.cs
```diff
@@ -0,0 +1,30 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiSymbolExtensions;
+
+namespace Microsoft.DotNet.GenAPI
+{
+    internal class IntersectionFilter : ISymbolFilter
```
ViktorHofer, line 11: This essentially is the composite pattern: https://en.wikipedia.org/wiki/Composite_pattern.

The English word "intersection" means something different than a composite. Hence, I would call it `CompositeFilter`.

ViktorHofer: Can you please also share this filter with ApiCopmat via ApiSymbolExtensions? As part of that, when you make the type public, please also add triple slash xml documentation to the type itself.

src/GenAPI/Microsoft.DotNet.GenAPI/Filters/ImplicitSymbolsFilter.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiSymbolExtensions;
+
+namespace Microsoft.DotNet.GenAPI
+{
+    internal class ImplicitSymbolsFilter : ISymbolFilter
+    {
+        /// <inheritdoc />
+        public bool Include(ISymbol member)
```
ViktorHofer, line 12: Instead of inheritdoc, can you please add a triple slash xml comment to explain what this filter does? Just from looking at the code, I have no idea why this filter is required for GenAPI.

src/GenAPI/Microsoft.DotNet.GenAPI/Filters/DelegateMembersFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.ApiSymbolExtensions;
+
+namespace Microsoft.DotNet.GenAPI
+{
+    internal class DelegateMembersFilter : ISymbolFilter
+    {
+        /// <inheritdoc />
+        public bool Include(ISymbol member)
```
ViktorHofer, line 12: Same as in the ImplicitSymbolsFilter, please add a tiple slash comment here instead of inheritdoc to explain what this filter does.

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Filters/FilterOutAttributes.cs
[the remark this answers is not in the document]
ViktorHofer: @smasher164 @ericstj do you know why we didn't just filter on `StartsWith("T:")` in ApiCompat's attribute exclusion parsing logic?

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
andriipatsula: Getting `Compiler Error CS0656`. Will need some time to investigate the problem

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Microsoft.DotNet.GenAPI.Tool.csproj
andriipatsula: Just in case: I noticed, that ApiContat.Tool targets net6.0 only

src/GenAPI/Microsoft.DotNet.GenAPI.Shared/Filters/FilterOutAttributes.cs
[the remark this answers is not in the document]
ericstj: Probably it didn't want to just silently ignore bad syntax.  It'd ignore comment lines, but everything else it wanted to treat as legitimate input.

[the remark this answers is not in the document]
ViktorHofer: That would make perfect sense if the input would be parsed in any way. Lines are just string compared against a type's DocId. Maybe that was different at some point and the lines were actually parsed and resolved.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpSyntaxWriterTests.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+using Microsoft.DotNet.GenAPI;
+
+namespace Microsoft.DotNet.GenAPI.Tests
+{
+    public class CSharpSyntaxWriterTests
```
smasher164, line 9: Missing tests?

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                description: "The path to one or more assemblies or directories with assemblies.",
+                parseArgument: ParseAssemblyArgument)
+            {
+                AllowMultipleArgumentsPerToken = true,
+                Arity = ArgumentArity.OneOrMore,
+                IsRequired = true
+            };
+
+            Option<bool> resolveAssemblyReferencesOption = new("--resolve-assembly-references",
+                "Default is `false`. If true, tries to resolve assembly reference.");
```
ViktorHofer, line 29: Remove as we don't use the approximation anymore:

```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    /// CLI frontend for the Roslyn-based GenAPI.
+    /// </summary>
+    class Program
+    {
+        static int Main(string[] args)
+        {
+            // Global options
+            Option<string[]> assembliesOption = new("--assemblies",
```
ViktorHofer, line 19: Singular option name:

```suggestion
            Option<string[]> assembliesOption = new("--assembly",
```

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                AllowMultipleArgumentsPerToken = true,
+                Arity = ArgumentArity.OneOrMore,
+                IsRequired = true
+            };
+
+            Option<bool> resolveAssemblyReferencesOption = new("--resolve-assembly-references",
+                "Default is `false`. If true, tries to resolve assembly reference.");
+
+            Option<string[]?> assemblyReferencesOption = new("--assembly-references",
```
ViktorHofer, line 31: Singular option name:

```suggestion
            Option<string[]?> assemblyReferencesOption = new("--assembly-reference",
```

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                description: "Paths to assembly references or their underlying directories for a specific target framework in the package.",
+                parseArgument: ParseAssemblyArgument)
+            {
+                AllowMultipleArgumentsPerToken = true,
+                Arity = ArgumentArity.ZeroOrMore
+            };
+
+            Option<string[]?> excludeAttributesFilesOption = new("--exclude-attributes-files",
```
ViktorHofer, line 39: Singular option name:

```suggestion
            Option<string[]?> excludeAttributesFilesOption = new("--exclude-attributes-file",
```

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            /// <summary>
+            /// The path to one or more assemblies or directories with assemblies.
+            /// </summary>
+            public string[] Assemblies { get; }
+
+            /// <summary>
+            /// Paths to assembly references or their underlying directories for a specific target framework in the package.
+            /// </summary>
+            public string[]? AssemblyReference { get; }
```
ViktorHofer, line 46:
```suggestion
            public string[]? AssemblyReferences { get; }
```

andriipatsula: Updated.

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,98 @@
...
+            Option<string[]?> assemblyReferencesOption = new("--assembly-reference",
+                description: "Paths to assembly references or their underlying directories for a specific target framework in the package.",
+                parseArgument: ParseAssemblyArgument)
+            {
+                Arity = ArgumentArity.ZeroOrMore
+            };
+
+            Option<string[]?> excludeAttributesFilesOption = new("--exclude-attributes-file",
+                description: "The path to one or more attribute exclusion files with types in DocId format.",
+                parseArgument: ParseAssemblyArgument)
```
ViktorHofer, line 36: nit: I don' think that we want the assembly parsing logic for attributes files. It's very rare that more than one or two exclusion files are passed in. In such cases, the consumer would just pass them by specifying the option multiple times: `--exclude-attributes-file a.txt --exclude-attributes-file b.txt`.

```suggestion
            Option<string[]?> excludeAttributesFilesOption = new("--exclude-attributes-file",
                "The path to one or more attribute exclusion files with types in DocId format.");
```

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
```diff
@@ -0,0 +1,98 @@
...
+            and then a file will be created for each assembly with the matching name of the assembly.");
+
+            Option<string?> headerFileOption = new("--header-file",
+                "Specify a file with an alternate header content to prepend to output.");
+
+            Option<string?> exceptionMessageOption = new("--exception-message",
+                "If specified - method bodies should throw PlatformNotSupportedException, else `throw null`.");
+
+            Option<bool> includeVisibleOutsideOfAssembly = new("--include-visible-outside",
+                "Include all API's not just public APIs. Default is public only. Default is `false`.");
```
ViktorHofer, line 52: nit: description

```suggestion
            Option<bool> includeVisibleOutsideOfAssembly = new("--include-visible-outside",
                "Include all API's not just public APIs. The default is public only.");
```

src/Microsoft.DotNet.ApiSymbolExtensions/Filtering/AttributeSymbolFilter.cs
```diff
@@ -0,0 +1,51 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.IO;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.DotNet.ApiSymbolExtensions
+{
+    public class AttributeSymbolFilter : ISymbolFilter
```
ViktorHofer, line 10: As this class is public, please add an xml comment similar to the CompositeFilter.

src/GenAPI/Microsoft.DotNet.GenAPI/IAssemblySymbolWriter.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.DotNet.GenAPI
+{
+    /// <summary>
+    /// Interface incapsulates logic for processing symbol assemblies.
```
ViktorHofer, line 9:
```suggestion
    /// Interface encapsulates logic for processing symbol assemblies.
```

src/GenAPI/Microsoft.DotNet.GenAPI/IAssemblySymbolWriter.cs
```diff
@@ -0,0 +1,15 @@
...
+    /// </summary>
+    internal interface IAssemblySymbolWriter
+    {
+        void WriteAssembly(IAssemblySymbol assemblySymbol);
```
ViktorHofer, line 13: Please add triple slash xml comments for this method. (as it's public)

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,158 @@
...
+        public static void Run(Context context)
+        {
+            IAssemblySymbolLoader loader = new AssemblySymbolLoader(true);
+
+            if (context.AssemblyReferences is not null)
+            {
+                loader.AddReferenceSearchPaths(context.AssemblyReferences);
+            }
+
+            var intersectionFilter = new CompositeFilter()
```
ViktorHofer, line 87: nit: Please rename this variable as we don't use "intersection" as the name anymore.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,158 @@
...
+            /// </summary>
+            public bool IncludeVisibleOutsideOfAssembly { get; }
+        }
+
+        /// <summary>
+        /// Initialize and run Roslyn-based GenAPI tool.
+        /// </summary>
+        public static void Run(Context context)
+        {
+            IAssemblySymbolLoader loader = new AssemblySymbolLoader(true);
```
ViktorHofer, line 80: I think the ASsemblySymbolLoader parameter should be dependent on if assembly references are passed in.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,158 @@
...
+            var intersectionFilter = new CompositeFilter()
+                .Add<ImplicitSymbolsFilter>()
+                .Add(new SymbolAccessibilityBasedFilter(context.IncludeVisibleOutsideOfAssembly));
+
+            if (context.ExcludeAttributesFiles != null)
+            {
+                intersectionFilter.Add(new AttributeSymbolFilter(context.ExcludeAttributesFiles));
+            }
+
+            var assemblySymbols = loader.LoadAssemblies(context.Assemblies);
```
ViktorHofer, line 96: We don't like to use `var` as it makes code reviews harder as it's not obvious what the actual type is.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,158 @@
...
+            foreach (var assemblySymbol in assemblySymbols)
+            {
+                if (assemblySymbol == null) continue;
```
ViktorHofer, line 99: I would add a "// TODO: Add logging for the assembly symbol loading failure".

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,158 @@
...
+
+                using var fileBuilder = new CSharpFileBuilder(
```
ViktorHofer, line 101:
```suggestion
                using CSharpFileBuilder fileBuilder = new(
```

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+            BlockSyntax node = GetMethodBodyFromText(SyntaxFactory.Space.ToString(), newLine);
+            return node.WithOpenBraceToken(node.OpenBraceToken.WithTrailingTrivia(SyntaxFactory.Space));
+        }
+
+        private BlockSyntax GetThrowNullBody(bool newLine = false)
+        {
+            if (_exceptionMessage is string exceptionMessage)
+            {
+                return GetMethodBodyFromText(string.Format(" throw new PlatformNotSupportedException(\"{0}\"); ", exceptionMessage), newLine);
+            }
```
ViktorHofer, line 103: No need to assign _exceptionMessage to a new variable. And use $"" which in this case is more readable.

```suggestion
            if (_exceptionMessage is not null)
            {
                return GetMethodBodyFromText($" throw new PlatformNotSupportedException(\"{_exceptionMessage}\"); "), newLine);
            }
```

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+            return GetMethodBodyFromText(" throw null; ", newLine);
+        }
+
+        private BlockSyntax GetMethodBodyFromText(string text, bool newLine = false) =>
+            SyntaxFactory.Block(SyntaxFactory.ParseStatement(text))
+                         .WithTrailingTrivia(newLine ? SyntaxFactory.CarriageReturnLineFeed : SyntaxFactory.Space);
+    }
+
+
```
ViktorHofer, line 112: nit: remove empty lines

```suggestion
```

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
```diff
@@ -0,0 +1,113 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.DotNet.GenAPI
+{
+    public class CSharpSyntaxWriter : CSharpSyntaxRewriter
```
ViktorHofer, line 11: Please add triple slash comments to the class and add `<inheritdoc />` statements to the public override methods below.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            _textWriter = textWriter;
+            _symbolFilter = symbolFilter;
+            _syntaxRewriter = syntaxRewriter;
+
+            _adhocWorkspace = new AdhocWorkspace();
+            _syntaxGenerator = SyntaxGenerator.GetGenerator(_adhocWorkspace, LanguageNames.CSharp);
+        }
+
+        public void WriteAssembly(IAssemblySymbol assembly) => Visit(assembly);
```
ViktorHofer, line 41: Add `<inheritdoc />`

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -0,0 +1,142 @@
...
+                yield return current;
+
+                foreach (var subNamespace in current.GetNamespaceMembers())
+                {
+                    stack.Push(subNamespace);
+                }
+            }
+        }
+
+        public void Dispose() => _textWriter.Dispose();
```
ViktorHofer, line 140: Add `<inheritdoc />`

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -0,0 +1,142 @@
...
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.Simplification;
+using Microsoft.DotNet.ApiSymbolExtensions;
+
+namespace Microsoft.DotNet.GenAPI
+{
+    /// <summary>
+    /// Processes assemly symbols to build correspoding structures in C# language.
+    /// Depend on ISyntaxWriter impelemtenation, the result could be C# source file, xml etc.
```
ViktorHofer, line 19: Is this line still valid? What ISyntaxWriter is meant here? Also, this class is specific to only emit C#, so xml or other languages don't need to be named here.

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
andriipatsula: Let's have the same behavior for all `plural` options

src/GenAPI/Microsoft.DotNet.GenAPI.Tool/Program.cs
andriipatsula: Thanks!

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
andriipatsula: No, the comment is no longer valid. Updated

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
andriipatsula: Added. Plus, added comment to base class.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
andriipatsula: added

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
andriipatsula: Updated:
```
    /// <summary>
    /// Represents a <see cref="CSharpSyntaxVisitor{TResult}"/> which descends an entire <see cref="CSharpSyntaxNode"/> graph and
    /// modify visited constructor, class, interface, struct, method declarations SyntaxNodes in depth-first order.
    /// </summary>
```

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
andriipatsula: Thanks, nice formatting feature `$"{variable}"`

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
andriipatsula: removed

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
andriipatsula:
```
bool resolveAssemblyReferences = context.AssemblyReferences is not null &&
    context.AssemblyReferences.Any();
```

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
andriipatsula: updated

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
andriipatsula: nice idea

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
andriipatsula: Ok. agree, sometimes it is good to have the full signature. In places like `foreach (var ..)` may be sutable

src/GenAPI/Microsoft.DotNet.GenAPI/IAssemblySymbolWriter.cs
andriipatsula: updated

src/GenAPI/Microsoft.DotNet.GenAPI/IAssemblySymbolWriter.cs
andriipatsula: this is done

src/Microsoft.DotNet.ApiSymbolExtensions/Filtering/AttributeSymbolFilter.cs
andriipatsula: updated

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
andriipatsula: added

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -0,0 +1,164 @@
...
+            public bool IncludeVisibleOutsideOfAssembly { get; }
+        }
+
+        /// <summary>
+        /// Initialize and run Roslyn-based GenAPI tool.
+        /// </summary>
+        public static void Run(Context context)
+        {
+            bool resolveAssemblyReferences = context.AssemblyReferences is not null &&
+                context.AssemblyReferences.Any();
```
ViktorHofer, line 82: Try to avoid Linq (in this case the Any() function) when possible as it's noticeable slower than simple array operations. Also the below suggestion is shorter :)

```suggestion
            bool resolveAssemblyReferences = context.AssemblyReferences?.Length > 0;
```

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -0,0 +1,144 @@
...
+    internal class CSharpFileBuilder : IAssemblySymbolWriter, IDisposable
+    {
+        private readonly TextWriter _textWriter;
+        private readonly ISymbolFilter _symbolFilter;
+        private readonly CSharpSyntaxRewriter _syntaxRewriter;
+
+        private readonly AdhocWorkspace _adhocWorkspace;
+        private readonly SyntaxGenerator _syntaxGenerator;
+
+
```
ViktorHofer, line 30: nit: remove empty line

```suggestion
```

ViktorHofer (approved):

andriipatsula:
@ViktorHofer I noticed, you just merged this PR. previously we agreed to save the history and have 2 commits (not a squash commit) - now we have lost this history (first commit). The merge commit is without description that was specified in this PR, is it expected? The commit description looks like:
>Introduce Roslyn-based GenAPI (https://github.com/dotnet/sdk/pull/28967)
> * Migrate GenAPI from dotnet/arcade repository.
> * Introduce the Roslyn-based GenAPI

ViktorHofer:
OMG facepalm, I'm an idiot ... Sorry for that. We can't really undo that. I think it's fine though. It would have been nice to preserve the first commit separately but it's OK.

PR #27012, opened 2022-08-05
Target net7.0 in additional projects

lbussell:
These net6.0 target frameworks are still bringing in prebuilts in source-build. They need to be changed before we can ship net7.0 without prebuilts.

PR #28984, opened 2022-11-09
Retarget Microsoft.Extensions.DotNetDeltaApplier to netstandard2.1

tmat:
## Description

`Microsoft.Extensions.DotNetDeltaApplier.dll` from the current .NET SDK version can be loaded to an app that targets any version of .NET >= 6.0 and equal or lower than the current version. We could target the dll to `net6.0` to satisfy these requirements, but that would add more prebuilds to source build. Instead, we target `netstandard2.1`.

Fixes https://github.com/dotnet/aspnetcore/issues/45129, https://github.com/aspnet/AspNetCore-ManualTests/issues/1635

## Customer Impact

Customers can't use `dotnet-watch` from .NET 7.0 when .NET 6.0 is also installed.

## Regression?

- [x] Yes
- [ ] No

Regressed by https://github.com/dotnet/sdk/pull/27012

## Risk

- [ ] High
- [ ] Medium
- [x] Low

The change is small and the affected code paths are well covered by tests.

## Verification

- [x] Manual (required)
- [x] Automated

- [x] Yes
- [ ] No
- [ ] N/A

### DIFF
--- a/src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
+++ b/src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
@@ -4,26 +4,52 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
-using System.Reflection.Metadata;
 
 namespace Microsoft.Extensions.HotReload
 {
     internal sealed class HotReloadAgent : IDisposable
     {
+        private delegate void ApplyUpdateDelegate(Assembly assembly, ReadOnlySpan<byte> metadataDelta, ReadOnlySpan<byte> ilDelta, ReadOnlySpan<byte> pdbDelta);
+
         private readonly Action<string> _log;
         private readonly AssemblyLoadEventHandler _assemblyLoad;
         private readonly ConcurrentDictionary<Guid, List<UpdateDelta>> _deltas = new();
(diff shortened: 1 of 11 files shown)


src/BuiltInTools/DotNetDeltaApplier/HotReloadAgent.cs
```diff
@@ -4,27 +4,53 @@
...
+        private readonly string? _capabilities;
         private volatile UpdateHandlerActions? _handlerActions;
 
         public HotReloadAgent(Action<string> log)
         {
+            var metadataUpdater = Type.GetType("System.Reflection.Metadata.MetadataUpdater, System.Runtime.Loader", throwOnError: false);
+
+            if (metadataUpdater != null)
+            {
+                _applyUpdate = (ApplyUpdateDelegate?)metadataUpdater.GetMethod("ApplyUpdate", BindingFlags.Public | BindingFlags.Static, binder: null,
```
mkArtakMSFT, line 31: So, this is really the main fix. Because this is happening at runtime now - it is picking up the version of the method from within the same version as the running code?

tmat: Not quite. The fix is to remove all metadata dependencies on 7.0.0 specific assemblies by targeting netstandard2.1.
Switching to reflection is only necessary because netstandard2.1 does not have the API.

mkArtakMSFT (approved):
Thanks @tmat.
I don't have any particular feedback but added @javiercn in case he will have any. Don't block on him, though.

tmat:
@lbussell PTAL

PR #29009, opened 2022-11-10
Fix TargetPlatformVersion normalization for Windows

dsplaisted:
This broke with https://github.com/dotnet/sdk/pull/27199


eng/common/sdk-task.ps1
```diff
@@ -64,7 +64,7 @@ try {
       $GlobalJson.tools | Add-Member -Name "vs" -Value (ConvertFrom-Json "{ `"version`": `"16.5`" }") -MemberType NoteProperty
     }
     if( -not ($GlobalJson.tools.PSObject.Properties.Name -match "xcopy-msbuild" )) {
-      $GlobalJson.tools | Add-Member -Name "xcopy-msbuild" -Value "17.2.1" -MemberType NoteProperty
```
nagilson, line 67: Is it possible ( and a good idea? ) to automate this version bumping?

dsplaisted: Ugh, looks like this was a bad merge or something, let me rebase it out.

dsplaisted: Sorry for the confusion, what's actually happening is that #28591 was an automated update with some test failures.  It turned out there were some breaking changes which the tests were catching, so I created a separate PR starting with the changes from the automated one to fix the regressions.

eng/common/tools.ps1
```diff
@@ -365,8 +365,8 @@ function InitializeVisualStudioMSBuild([bool]$install, [object]$vsRequirements =
 
   # If the version of msbuild is going to be xcopied,
   # use this version. Version matches a package here:
-  # https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-eng&package=RoslynTools.MSBuild&protocolType=NuGet&version=17.2.1&view=overview
-  $defaultXCopyMSBuildVersion = '17.2.1'
+  # https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-eng&package=RoslynTools.MSBuild&protocolType=NuGet&version=17.3.1view=overview
```
joeloff, line 368: Are we missing an `&` between `17.3` and `view=`?

dsplaisted: Yes, though this is in comment in a file that is synced from the arcade repo, so it would need to be fixed there: https://github.com/dotnet/arcade/blob/release/7.0/eng/common/tools.ps1#L368  @mmitche

dsplaisted:
FYI @j0shuams

nagilson (approved):
@dsplaisted I looked at the code and it seemed good, but I didn't see any test that tested this specific regression:
`A TargetPlatformVersion property for Windows without a trailing “.0” won’t be normalized to 4 version sections anymore (for example from 10.0.18362 to 10.0.18362.0).` Am I wrong that this is related to that?

The test changes I saw were just handling case-sensitivity, was a regression test added to check for the .0 behavior? I know some other tests (might) fail if it's wrong already.

dsplaisted:
@nagilson The test is `It_succeeds_if_windows_target_platform_version_does_not_have_trailing_zeros`.  There's no update to it in this PR, because this PR fixes the regression that caused the test to fail.  You can see it is one of the tests disabled in the 7.0.1xx PR: https://github.com/dotnet/sdk/pull/28591

nagilson:
Gotcha, thanks!

PR #29031, opened 2022-11-11
Don't Infer RID outside of dotnet publish for PublishSingleFile/PublishAot

nagilson:
# Summary

With the new implicit runtime identifier changes, if you do something like `PublishSingleFile=true` and build (on windows) it could work locally by inferring your local RID, but then you push to Azure (linux) and it wouldn't be for the correct runtime. Previously it just wouldn't work if you didn't set the RID, but this is a confusing behavior that emerged since these properties shouldn't affect operations outside of publishing. Thank you @DamianEdwards for pointing this behavior out.

## Concerns 

This should be highly considered for December servicing @marcpopMSFT.
But the risk that it breaks something is very high until we lab this out more. Like, does this break the restore feature with publish single file in that it will require another restore as the first would be RID Agnostic, and when publishing it wouldn't be? If so we need to add `_IsRestoring` and if we do that then the RID inference for these properties would be locked to CLI only and not work in MSBuild / VS. 

Another consideration is: are there other times when `PublishAot`, `PublishSingleFile`, or `PublishReadyToRun` need to be active outside of publish? I have pinged @ CLR AppModel Team to comment on this after talking with @LakshanF. 

EDIT: This will break publish with `-no-restore` if someone migrated to using a RIDless restore (or started using no-restore) unless we add `_IsRestoring`: 
image

Concern: How will `SelfContained` and friends work in VS if they no longer error but don't get a RID? 

cc @richlander

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -80,12 +80,14 @@ Copyright (c) .NET Foundation. All rights reserved.
                  '$(RuntimeIdentifier)' == '' and
                  '$(_IsExecutable)' == 'true' and '$(IsTestProject)' != 'true' and
                  '$(IsRidAgnostic)' != 'true' and
-                 ('$(_IsPublishing)' == 'true' or '$(SelfContained)' == 'true') and
-                  (
-                    '$(SelfContained)' == 'true' or
-                    '$(PublishReadyToRun)' == 'true' or
-                    '$(PublishSingleFile)' == 'true' or
-                    '$(PublishAot)' == 'true'
+                 (
+                   '$(SelfContained)' == 'true' or
+                   ('$(_IsPublishing)' == 'true' and
+                    (
+                      '$(PublishReadyToRun)' == 'true' or
+                      '$(PublishSingleFile)' == 'true' or
+                      '$(PublishAot)' == 'true'
+                    )
                   )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
@@ -184,7 +186,7 @@ Copyright (c) .NET Foundation. All rights reserved.
           Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
 
     <!-- The following RID errors are asserts, and we don't expect them to ever occur. The error message is added as a safeguard.-->
-    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == '''"
+    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == ''"
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="SelfContained"/>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -83,12 +83,13 @@ Copyright (c) .NET Foundation. All rights reserved.
                  (
                    '$(SelfContained)' == 'true' or
                    ('$(_IsPublishing)' == 'true' and
-                    (
-                      '$(PublishReadyToRun)' == 'true' or
-                      '$(PublishSingleFile)' == 'true' or
-                      '$(PublishAot)' == 'true'
-                    )
-                  )">true</UseCurrentRuntimeIdentifier>
+                      (
+                        '$(PublishReadyToRun)' == 'true' or
+                        '$(PublishSingleFile)' == 'true' or
+                        '$(PublishAot)' == 'true'
+                      )
+                   )
+                 )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == 'true'">
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -176,19 +176,19 @@ Copyright (c) .NET Foundation. All rights reserved.
           Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
 
     <!-- The following RID errors are asserts, and we don't expect them to ever occur. The error message is added as a safeguard.-->
-    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == ''"
+    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true'"
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="SelfContained"/>
 
-    <NETSdkError Condition="'$(PublishReadyToRun)' == 'true' and '$(RuntimeIdentifier)' == ''"
+    <NETSdkError Condition="'$(PublishReadyToRun)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true'"
                  ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                  FormatArguments="PublishReadyToRun"/>
 
-    <NETSdkError Condition="'$(PublishSingleFile)' == 'true' and '$(RuntimeIdentifier)' == ''"
+    <NETSdkError Condition="'$(PublishSingleFile)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true'"
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishSingleFile"/>
 
-    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(RuntimeIdentifier)' == ''"
+    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true'"
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishAot"/>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -72,11 +72,12 @@ Copyright (c) .NET Foundation. All rights reserved.
                  '$(RuntimeIdentifier)' == '' and
                  '$(_IsExecutable)' == 'true' and '$(IsTestProject)' != 'true' and
                  '$(IsRidAgnostic)' != 'true' and
+                 '$(_IsPublishing)' == 'true' and
                   (
                     '$(SelfContained)' == 'true' or
-                    ('$(_IsPublishing)' == 'true' and '$(PublishReadyToRun)' == 'true') or
-                    ('$(_IsPublishing)' == 'true' and '$(PublishSingleFile)' == 'true') or
-                    ('$(_IsPublishing)' == 'true' and '$(PublishAot)' == 'true')
+                    '$(PublishReadyToRun)' == 'true' or
+                    '$(PublishSingleFile)' == 'true' or
+                    '$(PublishAot)' == 'true'
                   )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -74,9 +74,9 @@ Copyright (c) .NET Foundation. All rights reserved.
                  '$(IsRidAgnostic)' != 'true' and
                   (
                     '$(SelfContained)' == 'true' or
-                    '$(PublishReadyToRun)' == 'true' or
-                    '$(PublishSingleFile)' == 'true' or
-                    '$(PublishAot)' == 'true'
+                    ('$(_IsPublishing)' == 'true' and '$(PublishReadyToRun)' == 'true') or
+                    ('$(_IsPublishing)' == 'true' and '$(PublishSingleFile)' == 'true') or
+                    ('$(_IsPublishing)' == 'true' and '$(PublishAot)' == 'true')
                   )">true</UseCurrentRuntimeIdentifier>
   </PropertyGroup>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -67,6 +67,14 @@ Copyright (c) .NET Foundation. All rights reserved.
     <SelfContained>$(PublishSelfContained)</SelfContained>
   </PropertyGroup>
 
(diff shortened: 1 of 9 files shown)


DamianEdwards:
The place this seemed to impact me the most was when running in Visual Studio with a project that has set `PublishSingleFIle="true"` in the project file, and selecting to launch with WSL. This appears to result in the build being RID-specific for win10-x64 and then the output copied to the WSL image where it (obviously) fails to launch.

I haven't tried the VS publish scenario to try publishing to Azure App Service for Linux but I think it knows the target RID in that case.

nagilson:
Thank you for clarifying, that is a pretty confusing behavior.. 

Some correspondence in the email thread: the RID shouldn't be necessary for doing a build with these properties, but their reach extends far outside of just dotnet publish. We're thinking more about this one. 

Based on the attached issue, we may want to change self-contained to not interact outside of publish, and a lot of these properties rely on self-contained.

Also, since there were discussions of making apps RID Specific by default in .NET 8, we should get together to discuss that at some point.

nagilson:
An enumeration of the current problem:

1. Before the change, VS would fail to build because PublishSingleFile requires you to specify a RID. Now it doesn't, but sometimes you want it to know of another RID. 

2. There are 2 components to the issue, the first is what Damian mentioned above for build and friends, the other is for publishing / restoring. 

3. For build and friends, we can limit the inference to only occur when the properties should be active (publishing and restoring). SelfContained and other publish* properties were meant for publish, and we would have to make a breaking change to make them no longer work (become a NO-OP) outside of dotnet publish which is what Andy raised to us above. If we do this change, we will also need work from Visual Studio to align the information from SDK Invokers conveying we are doing a restore/publish. If we don't, publish will suddenly take 2x more time because it will need to do 2 restores, or we would sacrifice the inference all together for VS, which would go to the previous behavior of failure. 

4.  For publish and restore, there is no real way to solve this, but at least in VS publishing wizard the RID is a more explicit item. 

5. Another problem is we are planning to make apps RID specific anyways as mentioned above, # 3 will become irrelevant for post net 7 projects if we do this. Do the pros outweigh the cons? Again, sounds like something we need to discuss before taking further action.

nagilson:
I take back some of that because it seems VS can do a separate restore anyways. I need to edit some errors to also only happen if `_IsPublishing` and perhaps we will go ahead with this change.

nagilson:
@dsplaisted This could be merged after review, though I am still a bit weary of the potential to break things, per our conversation it should be OK since the implicit RID is a very new thing. It also is a breaking change for dotnet restore as commented on above, but the impact is likely low.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -72,6 +72,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                  '$(RuntimeIdentifier)' == '' and
                  '$(_IsExecutable)' == 'true' and '$(IsTestProject)' != 'true' and
                  '$(IsRidAgnostic)' != 'true' and
+                 '$(_IsPublishing)' == 'true' and
                   (
                     '$(SelfContained)' == 'true' or
```
dsplaisted, line 77: If `SelfContained` is set to true, we should automatically set the RuntimeIdentifier even if not publishing.  If there's a `PublishSelfContained` property (maybe it's in 8.0 or just planned), then in that case the RuntimeIdentifier should only be set when publishing.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -175,19 +176,19 @@ Copyright (c) .NET Foundation. All rights reserved.
           Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
 
     <!-- The following RID errors are asserts, and we don't expect them to ever occur. The error message is added as a safeguard.-->
-    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == ''"
+    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true'"
```
dsplaisted, line 179: If we're handling `SelfContained` properly, I'm not sure this message makes sense here.

dsplaisted (commented):
I didn't look at the tests much, they may need to change once you change the `SelfContained` logic.

Also I'm kind of surprised it looks like we don't have a test that validates that if you build with `SelfContained` set to true that you get an implicit RuntimeIdentifier.  We should probably add one.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: May you elaborate a bit on this please? The message can occur if you turn off the RID inference by setting `RuntimeSpecific=false`. In addition, this error is acting as an assert like described in the comment above to make sure the inference is working correctly.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: Gotcha, implemented that behavior. `PublishSelfContained` has been around for a bit now in 7.0.200, you reviewed it! 😉

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -67,13 +67,22 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+
+  Finally, library projects and non-executable projects have awkward interactions here so they are excluded.-->
   <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == ''">
     <UseCurrentRuntimeIdentifier Condition="
                  '$(RuntimeIdentifier)' == '' and
                  '$(_IsExecutable)' == 'true' and '$(IsTestProject)' != 'true' and
                  '$(IsRidAgnostic)' != 'true' and
                  ('$(_IsPublishing)' == 'true' or '$(SelfContained)' == 'true') and
                   (
+                    '$(SelfContained)' == 'true' or
```
dsplaisted, line 85: The `SelfContained` check is partially duplicated, maybe this is a bit simpler:

```xml
  <PropertyGroup Condition="'$(UseCurrentRuntimeIdentifier)' == ''">
    <UseCurrentRuntimeIdentifier Condition="
                 '$(RuntimeIdentifier)' == '' and
                 '$(_IsExecutable)' == 'true' and '$(IsTestProject)' != 'true' and
                 '$(IsRidAgnostic)' != 'true' and
                 ('$(SelfContained)' == 'true' or
                   ('$(_IsPublishing)' == 'true') and
                     (
                        '$(PublishReadyToRun)' == 'true' or
                        '$(PublishSingleFile)' == 'true' or
                        '$(PublishAot)' == 'true'
                      )
                  )">true</UseCurrentRuntimeIdentifier>
  </PropertyGroup>
```

It's still complicated with a bunch of nested clauses though.  Could we set `SelfContained` to true if `_IsPublishing` is true and any of `PublishReadyToRun`, `PublishSingleFile`, or `PublishAot` are true?  If we did that beforehand, then we could remove the whole `_IsPublishing` clause from this condition and it would be a lot simpler and easier to understand.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -175,19 +184,19 @@ Copyright (c) .NET Foundation. All rights reserved.
           Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
 
     <!-- The following RID errors are asserts, and we don't expect them to ever occur. The error message is added as a safeguard.-->
-    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == ''"
+    <NETSdkError Condition="'$(SelfContained)' == 'true' and '$(RuntimeIdentifier)' == '''"
```
dsplaisted, line 187: Is the single quote supposed to be added here?  It looks like it's incorrect as far as I can see from trying to match them visually.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
dsplaisted: OK, sounds good.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -404,6 +404,32 @@ public void It_does_not_build_SelfContained_due_to_PublishSelfContained_being_tr
             outputDirectory.Should().NotHaveFile("hostfxr.dll"); // This file will only appear if SelfContained. 
         }
 
+        [Theory]
+        [InlineData("--p:PublishReadyToRun=true")]
+        [InlineData("-p:PublishSingleFile=true")]
+        [InlineData("-p:PublishSelfContained=true")]
+        [InlineData("-p:PublishAot=true")]
```
dsplaisted, line 411: Is there a reason the test is passing these as command line arguments instead of specifying them in the project?  It doesn't really make sense to pass these publish properties on the command line if you're just doing a `dotnet build`.  The expected scenario would be that they are in the project so when you do `dotnet publish` they take effect, but don't affect `dotnet build`.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -404,6 +404,32 @@ public void It_does_not_build_SelfContained_due_to_PublishSelfContained_being_tr
...
+        public void It_builds_without_implicit_rid_with_RuntimeIdentifier_specific_during_publish_only_properties(string property)
+        {
+            var tfm = ToolsetInfo.CurrentTargetFramework;
+            var testProject = new TestProject()
+            {
+                IsExe = true,
+                TargetFrameworks = tfm,
+            };
+            testProject.RecordProperties("RuntimeIdentifier");
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: $"ItBuildsWithoutImplicitRIDOnRIDRequiringPropertiesDuringPublish_{property}");
```
dsplaisted, line 421: You shouldn't need to add a bunch of stuff to the identifier, it just needs to be unique within the test name.

```suggestion
            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: property);
```

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
```diff
@@ -86,7 +86,7 @@ public void TestGlobalPropertyFlowToLibrary(bool passSelfContained, bool passRun
 
             bool buildingSelfContained = passSelfContained || passRuntimeIdentifier;
 
-            ValidateProperties(testAsset, _testProject, expectSelfContained: buildingSelfContained, expectRuntimeIdentifier: buildingSelfContained);
+            ValidateProperties(testAsset, _testProject, expectSelfContained: buildingSelfContained, expectRuntimeIdentifier: passRuntimeIdentifier);
```
dsplaisted, line 89: I'm not sure this change is right.  If we pass `--self-contained` as a command-line option, then the RuntimeIdentifier should be inferred.

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
```diff
@@ -103,8 +103,8 @@ public void TestGlobalPropertyFlowToExe(bool passSelfContained, bool passRuntime
 
             bool buildingSelfContained = passSelfContained || passRuntimeIdentifier;
 
-            ValidateProperties(testAsset, _testProject, expectSelfContained: buildingSelfContained, expectRuntimeIdentifier: buildingSelfContained);
-            ValidateProperties(testAsset, _referencedProject, expectSelfContained: buildingSelfContained, expectRuntimeIdentifier: buildingSelfContained);
+            ValidateProperties(testAsset, _testProject, expectSelfContained: buildingSelfContained, expectRuntimeIdentifier: passRuntimeIdentifier);
+            ValidateProperties(testAsset, _referencedProject, expectSelfContained: buildingSelfContained, expectRuntimeIdentifier: passRuntimeIdentifier);
```
dsplaisted, line 107: Same thing here, I'm not sure these lines should be changing.

dsplaisted (commented):
I'm not sure about a lot of the global property flow test changes.  We probably need to look at those closely and see if we need to document any breaking changes.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: I like that simplification, thanks for suggesting it.

In response to your question where the 3 `Publish*` properties imply `SelfContained`, I think not. based on this quote from @vitek-karas. 

```
PublishAot actually implies self contained, PublishTrimmed also currently implies self contained (but may not in the future),
PublishSingle file doesn’t imply self-contained but it behaves similarly (needs to download RID specific assets and can only publish for a specific RID),
PublishReadToRun is similar to PublishSingleFile in this regard (needs to download RID specific tooling and can only publish for a specific RID)
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: Yeah, that is definitely not supposed to be there. Removed

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
nagilson: Oh!

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
nagilson: You are right; It was very much not right, reverted this.

nagilson:
@dsplaisted Feedback has all been responded to.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetOsArchOptions.cs
```diff
@@ -143,36 +143,5 @@ public void CommandsRunWithArchOption(string command)
                 .Should()
                 .Pass();
         }
-
-        [Fact]
-        public void ItUsesImplicitRidWhenNoneIsSpecifiedForSelfContained()
```
dsplaisted, line 148: Why are these tests getting deleted?

Maybe we're moving some of the RID inference from the CLI to the SDK, but in that case I would expect to also see some code in the CLI command parsing deleted together with these tests.

dsplaisted (approved):
Looks good.

From a breaking changes perspective, my understanding is that for some property combinations:

- In 6.x, the project would fail to build
- In 7.0.1xx, the project would build with an implicit RuntimeIdentifier
- In 7.0.2xx (with these changes), the project will build without a RuntimeIdentifier

Is that correct?  I don't think the change from 6 to 7 needs to be documented, but we should document the breaking change from 7.0.1xx to 7.0.2xx.

src/Tests/dotnet.Tests/dotnet-msbuild/GivenDotnetOsArchOptions.cs
nagilson: I deleted the CLI command parsing code when I first added the RID inference but didn't see these tests until now

nagilson:
> Is that correct? I don't think the change from 6 to 7 needs to be documented, but we should document the breaking change from 7.0.1xx to 7.0.2xx.

It looks good to me.

I had already documented the `6 --> 7` change because there were people who somehow did not specify a `RuntimeIdentifier` but got past the error, but when we added the RID it started changing their behavior. That's what we were talking about on Teams and I asked them for more binlogs to investigate.

> we should document the breaking change from 7.0.1xx to 7.0.2xx. 

Agreed. But there is a SECOND breaking change in addition to the one you mentioned:
This will break publish with `-no-restore` if someone migrated to using a RIDless restore (or started using no-restore) unless we add `_IsRestoring`.

image

Please give your thoughts on this break @dsplaisted, we can avoid it but it will be not pretty.

nagilson:
Going to merge this as I think the benefit of this PR outweighs the cost of the breaking change in 7.0.200.

PR #29041, opened 2022-11-12
Migrate the nuget state from earlier SDKs

marcpopMSFT:


src/Cli/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
```diff
@@ -58,6 +58,9 @@ public void Configure()
             if (ShouldPrintFirstTimeUseNotice())
             {
                 Stopwatch beforeFirstTimeUseNotice = Stopwatch.StartNew();
+                // Migrate the nuget state from earlier SDKs
+                NuGet.Common.Migrations.MigrationRunner.Run();
```
kartheekp-ms, line 62: Please consider adding an interface and injecting it as a dependency into `DotnetFirstTimeUseConfigurer` class so that we can add a test in https://github.com/dotnet/sdk/blob/main/src/Tests/Microsoft.DotNet.Configurer.UnitTests/GivenADotnetFirstTimeUseConfigurer.cs to ensure that NuGet's migration logic is invoked when a dotnet sdk command is invoked for the first time.

```cs
interface INuGetMigrationRunner
{
 void Run();
}

class NuGetMigrationRunner : INuGetMigrationRunner
{
 public void Run() 
  {
     NuGet.Common.Migrations.MigrationRunner.Run();
  }
}
```

kartheekp-ms (approved):

src/Cli/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
marcpopMSFT: Please file a separate issue for that as we're not adding that interface and test during servicing.

PR #29065, opened 2022-11-14
Add error when specifying --output option for a solution

### LINKED ISSUE #15607: Error, warn, or fix behavior when specifying output path when building a solution
Specifying the output path (commonly by using the `-o` command line option to `dotnet build` or `dotnet publish`) when building or publishing a solution can lead to issues, as multiple projects building to the same path may override each other's files.

We should either block specifying the output path for a solution, or do something to fix the behavior (for example by appending the project name to the specified output path for each project).

Related:

- #3650
- #11953
- #13369
- #12126

dsplaisted:
Fixes #15607


src/Cli/dotnet/OptionForwardingExtensions.cs
```diff
@@ -18,6 +18,25 @@ public static class OptionForwardingExtensions
 
         public static ForwardedOption<T> ForwardAsSingle<T>(this ForwardedOption<T> option, Func<T, string> format) => option.SetForwardingFunction(format);
 
+        public static ForwardedOption<string> ForwardAsOutputPath(this ForwardedOption<string> option, string outputPropertyName, bool surroundWithDoubleQuotes = false)
```
nagilson, line 21: I think we should add a docstring that explains the `surroundWithDoubleQuotes` parameter, maybe just noting it's intended for `dotnet test` to preserve old behavior and remove the comment.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -879,4 +879,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1193: If PublishSelfContained is set, it must be either true or false. The value given was '{0}'.</value>
     <comment>{StrBegin="NETSDK1193: "}</comment>
   </data>
+    <data name="CannotHaveSolutionLevelOutputPath" xml:space="preserve">
+    <value>NETSDK1194: Building a solution with a specific OutputPath is not supported. Specifiy the output path at the individual project level instead.</value>
```
nagilson, line 883: nit: `is not` --> `isn't` would create a shorter error message

nagilson, line 883: Typo: `specifiy` -> `Specify`

nagilson, line 883: It would be valuable to describe how to specify the output path at the project level. Would I use the `OutputPath` property in the project file for that?

nagilson, line 883: nit: `at the individual` could be cleaner with `at an individual`.

src/Tests/dotnet.Tests/OutputPathOptionTests.cs
```diff
@@ -0,0 +1,98 @@
...
+namespace dotnet.Tests
+{
+    public class OutputPathOptionTests : SdkTest
+    {
+        public OutputPathOptionTests(ITestOutputHelper log) : base(log)
+        {
+        }
+
+        [Theory]
+        [InlineData("build")]
```
nagilson, line 29: This seems to have changed dotnet store too, but it isn't reflected in test here.

src/Tests/dotnet.Tests/OutputPathOptionTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+            else
+            {
+                new DotnetCommand(Log)
+                    .WithWorkingDirectory(slnDirectory)
+                    .Execute(command, $"--property:OutputPath={outputDirectory}")
+                    .Should()
+                    .Pass();
+            }
+
```
nagilson, line 94: nit: remove the empty line

src/Cli/dotnet/OptionForwardingExtensions.cs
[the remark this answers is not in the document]
dsplaisted: I'm not sure what you mean.  Visual Studio doesn't pass the `-o` option.  This PR is about what happens if you build a solution from the command line and specify that option.

src/Tasks/Common/Resources/Strings.resx
dsplaisted: It felt too complicated to explain in the error message so I just simplified the text to:

> The "--output" option isn't supported when building a solution.

src/Tests/dotnet.Tests/OutputPathOptionTests.cs
dsplaisted: Yes, `dotnet store` is complicated to test and sort of deprecated, so I just skipped testing it here.

src/Cli/dotnet/OptionForwardingExtensions.cs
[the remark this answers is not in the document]
nagilson: Sorry, I was thinking of some hypothetical implementation of VS that invoked an SDK CLI process with `-o ./project-name/`, in which case this would break that. But that doesn't happen. 

If that would be all it takes to implement the feature from the SDK side though, that would be pretty awesome

nagilson (approved):

aaronasmith:
This seems to have broken `dotnet pack` for solution files when specifying an output.  Even when called with nobuild.

ctyar:
Yeah, you have to specify a project:
dotnet pack **path\to\project.csproj** -o artifacts

kekyo:
Too many my projects are broken when building nuget packaging with `dotnet nuget pack -o <path>` usage...

dave-yotta:
This has broken our build, thanks.

GregTheMad:
Can confirmed --output works with 102 (pulled from repos locally), and is now broken with 200 (pulled by our Azure pipeline). (eventual consistency, ho!)

dave-yotta:
That didn't even work, specifying the folder(s) I just get #20 3.922 /usr/share/dotnet/sdk/7.0.200/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(221,5): error NU5026: The file '...' to be packed was not found on disk.

baronfel:
> This has broken our build, thanks. @ctyar we're packing the whole solution not a single nupkg.  Not sure why this has happened overnight but probabbly `dotnet-install.sh` has just started pointing to it or the dotnet buster images have just updated to .200?
> 
> This is a breaking change it's not ok to release these on a patch version.

@dave-yotta for the install-scriptsnspecifically you can limit them to the 100 series if that's a requirement for you.

For clarity, the 200 series is not a patch release from the 100-series of SDKs, it is a 'feature band' release where we do often introduce new features. The .NET SDK, and .Net overall, do not use SemVer for versions. While we do strive to make feature band' updates as painless as possible, we decided this work was worth doing, as we had seen many users end up with inconsistent builds due to the unclear semantics of publish/pack at the solution level.

Now having said that, we should have had a breaking change notice up on the docs and missed that - that's 100% our fault and we'll get that updated today.

dave-yotta:
I'm using net 7.0, I need the sdk to match that, without pulling any breaking changes.

dmelinosky:
Can confirm is is breaking one of our builds as well.

bbrandt-plx:
This is also breaking our build all of the sudden within Github Actions. We had a successful build a few days ago under MSBuild 17.4.1+9a89d02ff but now fails under MSBuild 17.5.0-preview-23061-01+040e2a90e

We have a solution with multiple projects that all produce nuget packages.

Here's the basic format of the dotnet Pack command we're running:

`dotnet pack /pathTo/ourSolution.sln --configuration Release --no-build --no-restore --output /output/artifacts/nuget --no-dependencies`

The build fails with: "error NETSDK1194: The "--output" option isn't supported when building a solution."

Any recommendations on how we can change our build commands to build a solution and publish all the nuget packages to one place?

ascott18:
To keep old behavior if you actually need all your packages to be outputted into one place (e.g. to publish as a build artifact rather than feeding straight to nuget push:

`-o out/packages` becomes `-p:PackageOutputPath="$(pwd)/out/packages"`

xt0rted:
This is a pretty disappointing change. We've been publishing all of our projects in one go using `dotnet publish ./project.sln --no-build --output ./artifacts` and that would publish each website and console app into its own artifacts folder which was passed to DevOps for deploying. This would take 2 seconds to run and required no extra effort when adding new projects that needed publishing. Now we have to publish each project individually which bumps the time to 20+ seconds. If we take `--no-build` off it takes minutes to build and publish artifacts.

jozefizso:
This change randomly broke our build system. Apparently the default SDK for GHA was .NET 6 where this option was working correctly and it is broken in .NET 7 which is default now.

Why we cannot use the `--output` option when there is directly replacement with `/p:PublishDir`?

jozefizso:
> For clarity, the 200 series is not a patch release from the 100-series of SDKs, it is a 'feature band' release where we do often introduce new features.

This is a breaking change and not a feature.

jozefizso:
> For pack specifically you should be able to do a combination of `dotnet pack <sln> <other options except --output>` followed by `dotnet nuget push **/bin/**/*.nupkg` - the push command handles globbing natively. That said, this does feel like a change that had too much unintended impact and the team is discussing mitigations at the moment.

This is really bad workaround. Pushing random nupkgs from all over the place to nuget?

baronfel:
Hi @jozefizso - it should be possible to update to 7.0.201, where we've pushed out a hotfix that addresses this regression. Especially removing it from `pack`, where it never should have been applied. There's a more in-depth example of the reasoning for the new diagnostic [on the change notification](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid#reason-for-change) if you'd like some more background on why setting properties like `PublishDir` at the solution level isn't something we recommend.

PR #29083, opened 2022-11-16
Fix unhandled exception during test discovery on macOS

akoeplinger:
While debugging https://github.com/dotnet/sdk/pull/29029 I noticed that we hit an unhandled exception because later macOS versions don't have an extra dot in the RID version number so the string split failed.

Switched to using float parsing for the version like we do for Ubuntu.

### DIFF
--- a/src/Tests/Microsoft.NET.TestFramework/EnvironmentInfo.cs
+++ b/src/Tests/Microsoft.NET.TestFramework/EnvironmentInfo.cs
@@ -125,9 +125,9 @@ namespace Microsoft.NET.TestFramework
             {
                 string restOfRid = currentRid.Substring(ridOS.Length + 1);
                 string ubuntuVersionString = restOfRid.Split('-')[0];
-                if (float.TryParse(ubuntuVersionString, out float ubuntuVersion))
+                if (float.TryParse(ubuntuVersionString, System.Globalization.CultureInfo.InvariantCulture, out float ubuntuVersion))
                 {
-                    if (ubuntuVersion > 16.04)
+                    if (ubuntuVersion > 16.04f)
                     {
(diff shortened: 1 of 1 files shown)


src/Tests/Microsoft.NET.TestFramework/EnvironmentInfo.cs
```diff
@@ -144,6 +144,18 @@ public static bool SupportsTargetFramework(string targetFramework)
             {
                 string restOfRid = currentRid.Substring(ridOS.Length + 1);
                 string osxVersionString = restOfRid.Split('-')[0];
+
+                //  Starting with macOS 12.0 (Monterey), the RID is "osx.12-x64" instead of the "osx.10.14-x64" format used before, so the version won't have a dot in it
+                if (!osxVersionString.Contains('.'))
```
dsplaisted, line 149: Instead of relying on the fact that the RID doesn't contain a period in it, could we make the OS parsing resilient to not having a dot?  Then this condition would be something like `macOSMajorVersion >= 12`.

akoeplinger: @dsplaisted good point, I switched to using float parsing for the version like we do for Ubuntu.

dsplaisted:
@akoeplinger I'm not sure float parsing is a good idea here either, since floats can't represent decimals exactly and you might get a result something like 16.0400000000001.  I was going to suggest something like `Version.Parse`, but that wouldn't be able to tell the difference between 16.04 and 16.4.  So probably `Decimal.Parse` would be better.

akoeplinger:
besides that, 10.12 is _ancient_ so we could probably get rid of that specific version check :)

dsplaisted (approved):

PR #29093, opened 2022-11-16
[ASP.NET] Remove versions from baselines

javiercn:
This PR changes how we perform baseline comparisons by removing the artifacts from the generated baselines and comparing the templates against each other instead of applying certain values to the templates


src/Tests/Microsoft.NET.Sdk.Razor.Tests/StaticWebAssetsBaselineComparer.cs
```diff
@@ -0,0 +1,251 @@
...
+        // Otherwise, do a property level comparison of all assets
+        switch (comparisonMode)
+        {
+            case GroupComparisonMode.Exact:
+                break;
+            case GroupComparisonMode.AllowAdditionalAssets:
+                break;
+            default:
+                break;
+        }
```
javiercn, line 84: TODO: we plan to allow tests to pass even if the baselines do not exactly match in some circumstances (like if there are additional dlls and so on). 

For that, we will compare the list of assets by groups of assets instead of as a flat list and apply different validation rules to each group.

TanayParikh (approved):
🏆

javiercn:
There seem to be a set of unrelated tests that are consistently failing.

image

Going to merge as these changes should not affect those tests.

nagilson:
Thanks for this PR, the watch tests are unrelated and being tracked here: https://github.com/dotnet/sdk/pull/29099

javiercn:
@nagilson unfortunately this is blocking me from merging.

mkArtakMSFT:
@marcpopMSFT can you force-merge this given that the cause for failures is known and is being tracked separately?

marcpopMSFT:
I have used the force

PR #29120, opened 2022-11-18
[ASP.NET] Remove versions from baselines (#29093)

javiercn:
* Updates the way the baselines are compared to generate a template of the baselines from the current manifest and compare it against the existing template instead of applying the current versions to the baseline and comparing it against the generated manifest.
* Removes all the versions and hashes from all the files Wed generate.
* Splits the code for generating and comparing the baselines into their own classes for better maintenance.
* Updates the baselines to reflect the new format.


MackinnonBuck (approved):

javiercn:
We have to wait a bit to merge on this branch

PR #29137, opened 2022-11-20
Add switch to disable apphost pack restore

ViktorHofer:
Contributes to https://github.com/dotnet/runtime/issues/58109
Mimics https://github.com/dotnet/sdk/pull/15944

In dotnet/runtime we live build targeting packs, runtime packs and app host packs. For the former two, switches already exist to disable nuget restore:
- EnableTargetingPackDownload
- EnableRuntimePackDownload

The latter one doesn't yet have such a switch and therefore is always restored by NuGet. In the cases where want to use the live built apphost pack, restore fails, i.e. from a runtime build:

`artifacts/bin/trimmingTests/projects/Microsoft.Extensions.DependencyInjection.TrimmingTests/ActivatorUtilitiesTests/osx-x64/project.csproj(0,0): error NU1102: (NETCORE_ENGINEERING_TELEMETRY=Build) Unable to find package Microsoft.NETCore.App.Host.osx-x64 with version (= 8.0.0)`

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolveAppHosts.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolveAppHosts.cs
@@ -54,6 +54,8 @@ namespace Microsoft.NET.Build.Tasks
         public bool NuGetRestoreSupported { get; set; } = true;
 
         public string NetCoreTargetingPackRoot { get; set; }
+        
+        public bool EnableAppHostPackDownload { get; set; }
 
         [Output]
         public ITaskItem[] PackagesToDownload { get; set; }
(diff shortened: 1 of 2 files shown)


akoeplinger (approved):

src/Tasks/Microsoft.NET.Build.Tasks/ResolveAppHosts.cs
```diff
@@ -54,6 +54,8 @@ public class ResolveAppHosts : TaskBase
         public bool NuGetRestoreSupported { get; set; } = true;
 
         public string NetCoreTargetingPackRoot { get; set; }
+        
+        public bool EnableAppHostPackDownload { get; set; } = true;
```
dsplaisted, line 58: Normally we set the default value for something like this in an MSBuild property.  I'm not entirely sure that this works correctly if the MSBuild property is empty, though if it didn't work there would probably some test failures.

PR #29155, opened 2022-11-22
Make `dotnet publish` use `Release` by default in 8.0+ Projects

nagilson:
# Changes:
- dotnet publish *(AND `dotnet pack`)* use a configuration of `Release` by default now if the TFM is above 8. `dotnet pack` will use `Release` REGARDLESS*** of the TFM.

- The environment variables to opt in or out for solutions have been removed. The error before you'd get using them was tighter than the new error, and while you won't be warned you need to turn on the ENV var anymore, you just don't need to.
- There's a new Environment variable that disables the features all-together called `DOTNET_CLI_DISABLE_PUBLISH_AND_PACK_RELEASE`
- `<Configuration>` in a project file never worked to begin with. This no longer considers if you set `<Configuration>` in your project file. Frankly, if that was supported, it would have dramatic consequences. 
-  If you multi-target with 8.0, I initially wrote it so you're going to get the `Release` behavior for each TFM in `dotnet pack`. No way to go around that besides making this an 8.0 SDK thing and not a TFM thing. FWIW `dotnet publish` with `TargetFrameworks` won't work so it picks whatever you used with `-f`.


src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -179,17 +164,49 @@ private ProjectInstance TryGetProjectInstance(string projectPath, Dictionary<str
...
+        /// Returns the target framework of the project.
+        /// This will return 0 if there are no TFM of the form netX.0.
+        /// If the project is multi-targeted, it will return the maximum target framework number (e.g. 8 for net8.0) of all modern TFMS in the project.
+        /// </returns>
+        /// <remarks>If the naming schema of TFM changes, this will not work. Yeah, it's not great.</remarks>
+        private float GetProjectMaxModernTargetFramework(ProjectInstance targetProject)
+        {
+            string multiTargetingFrameworks = targetProject.GetPropertyValue(MSBuildPropertyNames.TARGET_FRAMEWORKS);
+            string targetFramework = targetProject.GetPropertyValue(MSBuildPropertyNames.TARGET_FRAMEWORK);
+            if (!string.IsNullOrEmpty(multiTargetingFrameworks))
```
nagilson, line 184: One tested concern: 
If you have 8.0 as your publish TFM but it's not in your `TargetFrameworks` then it will fail anyways, so this behaves as expected (will use Release for 8.0 if you specify both `TargetFrameworks` and `TargetFramework`)

nagilson:
@dotnet/dotnet-cli Gentle ping reminder that this is available for review. Merge conflicts are just NETSDK Errors being added to the .resx in the time I was gone.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -112,11 +112,10 @@ Copyright (c) .NET Foundation. All rights reserved.
                             '$(SelfContained)' != 'true'"
                  ResourceName="CompressionInSingleFileRequiresSelfContained" />
 
-    <!-- Enable warning for trying to use PublishRelease or PackRelease with a solution if env-var is not set.-->
-    <NETSdkWarning Condition="'$(PublishRelease)' != '' and '$(SolutionExt)' == '.sln' and '$(DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS)' == ''"
-                   ResourceName="PReleaseRequiresEnvVarOnSln" 
-                   FormatArguments="PublishRelease;DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS"
-    />
+    <!-- Projects in a solution cannot have conflicting configurations. This checks if PublishRelease conflicted at runtime to avoid extra project evaluations. -->
+    <NETSdkError  Condition="'$(_SolutionLevelPublishRelease)' != '' and '$(_SolutionLevelPublishRelease)' != '$(PublishRelease)'"
```
dsplaisted, line 116: Do we ever set `PublishRelease` (or `PackRelease`) in targets?  It doesn't look like it, so I think we will always generate these errors when publishing or packing a solution unless `PublishRelease` is specified, even if the solution only has a single project.

I think we may need to go back to setting the default values to these properties in the .targets files, so that we can generate the errors when there is a conflict but otherwise not generate them.

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,19 +88,21 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
-        [Fact]
-        public void It_packs_with_release_if_PackRelease_property_set()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void It_packs_with_release_if_PackRelease_property_set(bool optedOut)
         {
+            if (optedOut)
+            {
+                Environment.SetEnvironmentVariable(EnvironmentVariableNames.DISABLE_PUBLISH_AND_PACK_RELEASE, "true");
```
dsplaisted, line 98: We shouldn't be setting environment variables in the test process.  You can do the following instead:

```c#
            packCommand
                .WithEnvironmentVariable(EnvironmentVariableNames.DISABLE_PUBLISH_AND_PACK_RELEASE, "true")
                .Execute()
                .Should()
                .Pass();
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,19 +88,21 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
+            }
+
             var helloWorldAsset = _testAssetsManager
-               .CopyTestAsset("HelloWorld", "PackReleaseHelloWorld")
+               .CopyTestAsset("HelloWorld", $"PackReleaseBasicTest-Opted-Out-{optedOut}")
```
dsplaisted, line 102:
```suggestion
               .CopyTestAsset("HelloWorld", identifier: optedOut)
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,19 +88,21 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
                .WithSource();
 
-            System.IO.File.WriteAllText(helloWorldAsset.Path + "/Directory.Build.props", "<Project><PropertyGroup><PackRelease>true</PackRelease></PropertyGroup></Project>");
-
-            new BuildCommand(helloWorldAsset)
-               .Execute()
-               .Should()
-               .Pass();
+            File.WriteAllText(helloWorldAsset.Path + "/Directory.Build.props", "<Project><PropertyGroup><PackRelease>true</PackRelease></PropertyGroup></Project>");
```
dsplaisted, line 105: Nit: Use `Path.Combine`

```suggestion
            File.WriteAllText(Path.Combine(helloWorldAsset.Path, "Directory.Build.props"), "<Project><PropertyGroup><PackRelease>true</PackRelease></PropertyGroup></Project>");
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -123,70 +130,98 @@ public void It_packs_with_release_if_PackRelease_property_set_in_csproj()
...
 
-        [Fact(Skip = "https://github.com/dotnet/sdk/issues/27066")]
-        public void It_warns_if_PackRelease_set_on_sln_but_env_var_not_used()
+        [Theory]
+        [InlineData("net7.0")]
+        [InlineData("net8.0")]
+        public void It_packs_with_release_by_default_in_net_8_but_not_in_net_7(string tfm)
+        {
+            var helloWorldAsset = _testAssetsManager
+               .CopyTestAsset("HelloWorld", $"PackReleaseDefault-{tfm}")
```
dsplaisted, line 150:
```suggestion
               .CopyTestAsset("HelloWorld", identifier: tfm)
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -123,70 +130,98 @@ public void It_packs_with_release_if_PackRelease_property_set_in_csproj()
...
+
+            new DotnetPackCommand(Log)
+                .WithWorkingDirectory(Path.Combine(testAsset.TestRoot, testProject.Name))
+                .Execute()
+                .Should()
+                .Pass();
+
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: "net7.0", configuration: "Release"); // this will fail if configuration is debug and TFM code didn't work.
+            var finalConfiguration = properties["Configuration"];
+            Assert.True(finalConfiguration == "Release");
```
dsplaisted, line 183:
```suggestion
            finalConfiguration.Should().Equal("Release");
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -426,141 +426,130 @@ public void It_fails_for_unsupported_rid()
...
         {
-            var helloWorldAsset = _testAssetsManager
-               .CopyTestAsset("HelloWorld", "PublishReleaseHelloWorld")
-               .WithSource()
-               .WithTargetFramework(ToolsetInfo.CurrentTargetFramework);
-
-            System.IO.File.WriteAllText(helloWorldAsset.Path + "/Directory.Build.props", "<Project><PropertyGroup><PublishRelease>true</PublishRelease></PropertyGroup></Project>");
+            if (optedOut)
+            {
+                Environment.SetEnvironmentVariable(EnvironmentVariableNames.DISABLE_PUBLISH_AND_PACK_RELEASE, "true");
```
dsplaisted, line 436: As before, use `WithEnvironmentVariable` instead.

dsplaisted (commented):
Good progress.

I think we are missing tests for publishing and packing solutions where the `PublishRelease` or `PackRelease` properties aren't specified but there's no conflict between projects.  I think currently that will fail.

To get it to work I think we may need to move the default values for the properties back into the .targets files.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -879,4 +875,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1193: If PublishSelfContained is set, it must be either true or false. The value given was '{0}'.</value>
     <comment>{StrBegin="NETSDK1193: "}</comment>
   </data>
+  <data name="SolutionProjectConfigurationsConflict" xml:space="preserve">
+    <value>NETSDK1194: Multiple solution project(s) contain conflicting '{0}' values; ensure the values match. Consider using a Directory.build.props file to set all project configurations. Conflicting project:</value>
```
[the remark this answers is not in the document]
nagilson: The project appears to be put at the end of the output by default, so I think this still makes sense. 
image

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -18,60 +17,85 @@
 
 namespace Microsoft.DotNet.Cli
 {
+    /// <summary>
+    /// This class is used to enable configuration changes at the project level.
+    /// Configuration evaluation occurs before a project file is evaluated, and the project file may have dependencies on the configuration.
+    /// Because of this, it is 'impossible' for the project file to correctly influence the value of Configuration.
+    /// This class allows evaluation of Configuration properties set in the project file before build time by giving back a global Configuration property to inject while building.
```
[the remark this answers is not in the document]
nagilson: Changed:
```
 /// This class is used to enable properties that edit the Configuration property inside of a .*proj file.
    /// Properties such as DebugSymbols are evaluated based on the Configuration set before a project file is evaluated, and the project file may have dependencies on the configuration.
    ```

File: src/Cli/dotnet/ReleasePropertyProjectLocator.cs [hunk 11]
```diff
@@ -83,31 +107,23 @@ public ProjectInstance GetTargetedProject(IEnumerable<string> slnOrProjectArgs,
...
-
-            return string.IsNullOrEmpty(potentialProject) ? null : TryGetProjectInstance(potentialProject, globalProps);
+            return null;  // If nothing can be found: that's caught by MSBuild XMake::ProcessProjectSwitch -- don't change the behavior by failing here. 
         }
 
-        /// <returns>The executable project (first if multiple exist) in a SLN. Returns null if no executable project. Throws exception if two executable projects disagree
-        /// in the configuration property to check.</returns>
-        public ProjectInstance GetSlnProject(string slnPath, Dictionary<string, string> globalProps, string slnProjectConfigPropertytoCheck = "")
+        /// <returns>A random existant project in a solution file. Returns null if no projects exist.</returns>
+        public ProjectInstance GetRandomSlnProject(string slnPath, Dictionary<string, string> globalProps)
```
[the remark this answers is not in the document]
nagilson: The joys of random vs arbitrary!

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -18,60 +17,85 @@
...
-
-                if (!string.IsNullOrEmpty(configurationToUse) && !ProjectHasUserCustomizedConfiguration(project, defaultedConfigurationProperty))
-                    yield return $"-property:{MSBuildPropertyNames.CONFIGURATION}={configurationToUse}";
+                string releasePropertyFlag = project.GetPropertyValue(_defaultedConfigurationProperty);
+                if (!string.IsNullOrEmpty(releasePropertyFlag)) // The project set PublishRelease or PackRelease itself.
+                {
+                    string configurationToUse = releasePropertyFlag.Equals("true", StringComparison.OrdinalIgnoreCase) ? MSBuildPropertyNames.CONFIGURATION_RELEASE_VALUE : MSBuildPropertyNames.CONFIGURATION_DEBUG_VALUE;
+                    return new List<string> {
+                        $"-property:{MSBuildPropertyNames.CONFIGURATION}={configurationToUse}",
+                        _isPublishingSolution ? $"-property:_SolutionLevel{_defaultedConfigurationProperty}=true" : "" // Allows us to spot conflicting configuration values during evaluation.
```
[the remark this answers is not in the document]
nagilson: I rewrote it to not have the entry, though the code is less readable, I tried to add some comments.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -18,60 +17,85 @@
...
-            ParseResult parseResult,
-            string defaultedConfigurationProperty,
-            IEnumerable<string> slnOrProjectArgs,
-            Option<string> configOption
-            )
+        public IEnumerable<string> GetCustomDefaultConfigurationValueIfSpecified()
         {
-            ProjectInstance project = null;
-            var globalProperties = GetGlobalPropertiesFromUserArgs(parseResult);
+            Debug.Assert(_defaultedConfigurationProperty == MSBuildPropertyNames.PUBLISH_RELEASE || _defaultedConfigurationProperty == MSBuildPropertyNames.PACK_RELEASE, "Only PackRelease or PublishRelease are currently expected.");
```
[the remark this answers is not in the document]
nagilson: I would prefer that the check is compiled out of production so I will leave this as is for now. Please let me know if you disagree and explain a bit as to why, I am guessing it's for clarity in the argumentexception, but I think this is clear enough.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -18,60 +17,85 @@
...
+                string releasePropertyFlag = project.GetPropertyValue(_defaultedConfigurationProperty);
+                if (!string.IsNullOrEmpty(releasePropertyFlag)) // The project set PublishRelease or PackRelease itself.
+                {
+                    string configurationToUse = releasePropertyFlag.Equals("true", StringComparison.OrdinalIgnoreCase) ? MSBuildPropertyNames.CONFIGURATION_RELEASE_VALUE : MSBuildPropertyNames.CONFIGURATION_DEBUG_VALUE;
+                    return new List<string> {
+                        $"-property:{MSBuildPropertyNames.CONFIGURATION}={configurationToUse}",
+                        _isPublishingSolution ? $"-property:_SolutionLevel{_defaultedConfigurationProperty}=true" : "" // Allows us to spot conflicting configuration values during evaluation.
+                    };
+                }
+                else if (GetProjectMaxModernTargetFramework(project) >= 8) // For 8.0, these properties are enabled by default.
```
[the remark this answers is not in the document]
nagilson: That's a good point, sadly we couldn't think of a better alternative. @/baronfel and @/richlander and I discussed these 3 evils:

For PublishRelease and .NET 8, which of the following is a better choice?

1) When multitargeting, if net8.0 is in your set of tfms, the default for publish will be Release no matter the TFM
2) When multitargeting, Publish will always default to Debug
3) When using the 8.0 SDK, Release will be the default of Publish regardless of the TFM

We Chet said we didn't want 2. And given only about 5% of customers multi-target, I think 1) is better. 

Even if we handle the default value in targets I think it will be too late to pivot. We can add the default value but I don't think we can gate the behavior on TFM.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -179,17 +164,49 @@ private ProjectInstance TryGetProjectInstance(string projectPath, Dictionary<str
...
         }
 
+        /// <param name="targetProject">The project which we want to get the potential modern (NET 5.0/Core+) TFM of.</param>
+        /// <returns>
+        /// Returns the target framework of the project.
+        /// This will return 0 if there are no TFM of the form netX.0.
+        /// If the project is multi-targeted, it will return the maximum target framework number (e.g. 8 for net8.0) of all modern TFMS in the project.
+        /// </returns>
+        /// <remarks>If the naming schema of TFM changes, this will not work. Yeah, it's not great.</remarks>
+        private float GetProjectMaxModernTargetFramework(ProjectInstance targetProject)
```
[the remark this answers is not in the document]
nagilson: Luckily we don't need to parse it at all anymore... this code has been removed.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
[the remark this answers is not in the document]
nagilson: Just for record keeping, now this has been changed and it should respect the `-f` option above all else. Multitargeting isn't even an issue with publish and pack will have it own by default.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
nagilson: They've been moved to targets, good idea. Not sure what I was thinking when I did it like the old implementation. This is no longer an issue, have tested it.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -18,60 +17,85 @@
...
+            string defaultedConfigurationProperty,
+            IEnumerable<string> slnOrProjectArgs,
+            Option<string> configOption
+         )
+         => (_parseResult, _defaultedConfigurationProperty, _slnOrProjectArgs, _configOption) = (parseResult, defaultedConfigurationProperty, slnOrProjectArgs, configOption);
 
         /// <param name="slnProjectPropertytoCheck">A property to enforce if we are looking into SLN files. If projects disagree on the property, throws exception.</param>
         /// <returns>A project instance that will be targeted to publish/pack, etc. null if one does not exist.</returns>
-        public ProjectInstance GetTargetedProject(IEnumerable<string> slnOrProjectArgs, Dictionary<string, string> globalProps, string slnProjectPropertytoCheck = "")
+        public ProjectInstance GetTargetedProject(Dictionary<string, string> globalProps)
```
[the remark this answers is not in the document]
nagilson: Nope, that doesn't work. Thanks, made a change and adding a test for this.

src/Assets/TestProjects/TestAppWithSlnUsingPublishReleaseConflictingValues/App/App.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <PublishRelease>false</PublishRelease>
+    <PackRelease>true</PackRelease>
```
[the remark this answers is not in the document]
nagilson: Good point, we didn't. I'm working on a LOT of tests for solutions... see `ItFailsWithNet8ProjectAndNet7ProjectSolutionWithPublishReleaseUndefined` once I push it up

richlander:
> <Configuration> in a project file never worked to begin with.

Does that mean that this pattern is broken/bad? @tmat gave it to me.

https://github.com/dotnet/core/blob/2ad3ba7e14816c3a8cb8459259e9aa88895c9052/samples/Directory.Build.props#L30-L34

> dotnet publish *(AND `dotnet pack`)* use a configuration of `Release` by default now if the TFM is above 8. `dotnet pack` will use `Release` REGARDLESS*** of the TFM.

Why the difference in behavior?

nagilson:
> Does that mean that this pattern is broken/bad? @tmat gave it to me.

The pattern you linked is in `Directory.build.props` so it works fine as that's the first thing imported. If it's in the project file itself like `foo.csproj` then it is broken/bad. Note that it actually changes the path even if it doesn't change the underlying configuration of the assets which is pretty dubious! 

> Why the difference in behavior?

This I will explain in a set of breaking change documents I need to write. It is .. ugh, complicated. But the short answer is pack allows multitarget without a specific target framework while publish doesn't, and there are complications around solutions with different projects with different TFMS + multitargeting combined together and figuring out which `Configuration` to pick.

nagilson:
@dsplaisted This is ready for review again. Let me know of any concerns. I've implemented the other design where the publish got slower but less people were broken (adding an 8.0 project to a solution with existing projects of older TFMs where `PRelease` (pack or publish) is undefined is now valid)

src/Assets/TestProjects/TestAppWithSlnUsingPublishReleaseConflictingValues/App.sln
```diff
@@ -0,0 +1,48 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
dsplaisted: Should this file be deleted now that you have the `TestSolution` class?

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -18,96 +18,132 @@
...
-                string configurationToUse = "";
-                string releasePropertyFlag = project.GetPropertyValue(defaultedConfigurationProperty);
-                if (!string.IsNullOrEmpty(releasePropertyFlag))
-                    configurationToUse = releasePropertyFlag.Equals("true", StringComparison.OrdinalIgnoreCase) ? MSBuildPropertyNames.CONFIGURATION_RELEASE_VALUE : "";
+                string propertyToCheckValue = project.GetPropertyValue(_propertyToCheck);
+                if (!string.IsNullOrEmpty(propertyToCheckValue))
+                {
+                    var newConfigurationArgs = new List<string> {
+                        $"-property:{MSBuildPropertyNames.CONFIGURATION}={
+                            (propertyToCheckValue.Equals("true", StringComparison.OrdinalIgnoreCase) ? MSBuildPropertyNames.CONFIGURATION_RELEASE_VALUE : MSBuildPropertyNames.CONFIGURATION_DEBUG_VALUE)
```
dsplaisted, line 95: Looking at this, I think that we only want to explicitly set the Configuration to Release if the propertyToCheck is true.  If propertyToCheck is set to false, then we don't need to explicitly set the Configuration to Debug, we can just let the default value take effect, which is normally going to be Debug.

Does this make sense?  I'm not sure if I've considered this before.

dsplaisted (commented):
Per offline discussion, adding comments I have so far.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: This is definitely something we could do. With that being said, I think this is preferable. If someone adds `Configuration` to the project file, it will impact the output path if this code for `Debug` is removed. Obviously, this is a bad operation in the first place, but it will not impact the output path with this change added.

src/Assets/TestProjects/TestAppWithSlnUsingPublishReleaseConflictingValues/App.sln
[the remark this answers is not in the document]
nagilson: I would prefer not to, as it helps to increase code coverage -- there is the situation with locating a solution where you provide a hard-coded solution path, but the solution path is not in the same working directory as the test/project, which having the test asset allows us to cover better, as it will be in a different folder, unless I am wrong. Admittedly, we could test this in different ways, but combined with this and the number of tests required to rewrite without this file, I would prefer to keep it. `TestSolution` is still saving us from 17 other versions of the same solution. You may push back on this and we can remove it if need be though.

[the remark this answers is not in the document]
nagilson: Ugh, maybe the copy of it makes it so it doesn't. Hold up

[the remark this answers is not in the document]
nagilson: Ya it doesn't. Removing this

nagilson:
@dsplaisted PTAL. I renamed the environment variable, FYI. I thought the old name didn't communicate the purpose of it too well. Now `DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS`

AntonLapounov:
> Now `DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS`

What does 'lazy' mean here?

nagilson:
Waiting for Daniel to approve of the name but @AntonLapounov in this instance it means it will only evaluate one project in the solution for `Configuration` correctness instead of all of them... saving efficiency/speed, but it cannot as precisely distinguish what is an error and what is not this way. By that, I mean if you add any `8.0` project to an existing solution containing not `8.0` projects, that does not define `PublishRelease`, give `-c`, or disable the new feature, it will fail. This will be off by default but can be turned on if the performance hit of evaluating every project is not desired.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -118,55 +161,93 @@ public ProjectInstance GetSlnProject(string slnPath, Dictionary<string, string>
...
-            string executableProjectOutputType = MSBuildPropertyNames.OUTPUT_TYPE_EXECUTABLE; // Note that even on Unix when we don't produce exe this is still an exe, same for ASP
-            const string solutionFolderGuid = "{2150E333-8FDC-42A3-9474-1A3956D46DE8}";
-            const string sharedProjectGuid = "{D954291E-2A0B-460D-934E-DC6B0785DB48}";
+            if (String.Equals(Environment.GetEnvironmentVariable(EnvironmentVariableNames.DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS), "true", StringComparison.OrdinalIgnoreCase))
+            {
+                // Evaluate only one project for speed if this environment variable is used. Will break more customers if enabled (adding 8.0 project to SLN with other project TFMs with no Publish or PackRelease.)
+                return GetSingleProjectFromSolution(sln, globalProps);
+            }
 
             Parallel.ForEach(sln.Projects.AsEnumerable(), (project, state) =>
```
dsplaisted, line 175: I think we want to check the TargetFramework(s) of the projects in this method too.  If none of the projects explicitly define PublishRelease, then we should check to see if any projects target .NET 8 or greater, and if so treat it as if PublishRelease is true.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -117,12 +117,16 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-                   ResourceName="PReleaseRequiresEnvVarOnSln" 
-                   FormatArguments="PublishRelease;DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS"
-    />
-
+    <!-- Projects in a solution cannot have conflicting configurations. This checks if PublishRelease conflicted at runtime to avoid extra project evaluations.
+    SolutionExt is used to check if we are publishing a solution. Will be undefined if not.-->
+    <NETSdkError Condition="'$(_IsPublishing)' == 'true' and
+                            '$(DOTNET_CLI_DISABLE_PUBLISH_AND_PACK_RELEASE)' != 'true' and
+                            '$(DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS)' == 'true' and
+                            '$(SolutionExt)' == '.sln' and
```
dsplaisted, line 125: Probably rather than checking `SolutionExt`, you can check whether `_SolutionLevelPublishRelease` is set.

```suggestion
                            '$(_SolutionLevelPublishRelease)' != '' and
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
```diff
@@ -153,4 +153,21 @@ Copyright (c) .NET Foundation. All rights reserved.
     <NETSdkWarning ResourceName="UnnecessaryWindowsDesktopSDK" />
   </Target>
 
+  <!-- Default the value of PublishRelease for 8.0+ TFMs. Note that this is pre-evaluated by the CLI in an evaluation before evaluation (see ReleasePropertyProjectLocator.cs).-->
+  <PropertyGroup Condition="'$(PublishRelease)' == '' and '$(_TargetFrameworkVersionWithoutV)' >= '8.0'">
+    <PublishRelease>true</PublishRelease>
```
dsplaisted, line 158: If we want the CLI to handle the case where you have a solution with a .NET 7 and .NET 8 project and neither explicitly sets `PublishRelease`, we need a way for the CLI to detect whether `PublishRelease` was explicitly set.  You could either move the defaulting logic to the CLI, or you could set a property such as `_UsingDefaultPublishRelease` to true here.  I'd probably go with the second option.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props
```diff
@@ -24,6 +24,15 @@ Copyright (c) .NET Foundation. All rights reserved.
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
   </PropertyGroup>
 
+  <!-- Default the value of PackRelease for every project. Note that this is pre-evaluated by the CLI in an evaluation before evaluation (see ReleasePropertyProjectLocator.cs).
+  Multitargeting pack splits a single 2 TFM project into 9 separate evaluations, with different target imports. Microsoft.NET.SDK.DefaultItems.Targets ...
+  is where PublishRelease is defaulted because it depends on _TargetFrameworkVersionWithoutV. Some targets do not run in some instances of dotnet pack.
+  So, we must set PackRelease here. Even though this occurs before project import, the project import can override PackRelease later, before the pre-evaluation, so this is fine.-->
+  <PropertyGroup Condition="'$(PackRelease)' == ''">
+    <!-- v Do not depend on this value being correct until after project evaluation.-->
+    <PackRelease>true</PackRelease>
```
dsplaisted, line 33: Same here, you probably need to set something like `_UsingDefaultPackRelease` to let the CLI know whether this was explicitly specified.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
nagilson: asking rainer on this one

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: We don't need to do this, as discussed offline, the pre-evaluation evaluates the defaulting code on a `TargetFramework` basis. And the `-f` required on publish multi-target will flow into this pre-evaluation. In addition, `dotnet pack` will default for all TFM, so multi-targeting is not an issue. And while we may be able to evaluate the TFMS, the `nuget` code does not seem to have the needed functionality, and in addition, it would require us to have different default TFM behaviors depending on the environment variable.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
nagilson: As discussed offline, this occurs in the pre-evaluation, so it is not necessary.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props
nagilson: As discussed offline, this occurs in the pre-evaluation, so it is not necessary.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
nagilson: Per Rainier, regarding `SolutionExt` being correct: 
>  it's basically a guarantee right now

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -199,19 +288,22 @@ private bool IsValidProjectFilePath(string path)
...
         }
 
-        /// <param name="propertyToDisableIfTrue">A property that will be disabled (NOT BY THIS FUCTION) based on the condition.</param>
-        /// <returns>Returns true if Configuration on a project is not Debug or Release. Note if someone explicitly set Debug, we don't detect that here.</returns>
-        /// <remarks>Will log that the property will not work if this is true.</remarks>
-        private bool ProjectHasUserCustomizedConfiguration(ProjectInstance project, string propertyToDisableIfTrue)
+        /// <summary>
+        /// Because command-line-forwarded properties aren't in the global arguments from Properties, we need to add the TFM to the collection.
+        /// The TFM is the only property besides Configuration that isn't an MSBuild property that could affect the pre-evaluation.
+        /// This allows the pre-evaluation to correctly deduce its Publish or PackRelease value because it will know the actual TFM being used.
```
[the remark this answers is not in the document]
nagilson: The second line does not mean the only property per se.. it means that it is the only CommandLine Option (in the CLI) that you can give that doesn't appear in the msbuild property dictionary, that we would expect to affect the evaluation. I can change this

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -118,55 +161,93 @@ public ProjectInstance GetSlnProject(string slnPath, Dictionary<string, string>
...
+                // 3) This code only gets hit if we are in a solution publish setting, so we don't need to worry about it failing other publish scenarios
+                throw new GracefulException(Strings.SolutionProjectConfigurationsConflict, _propertyToCheck, String.Join("\n", (configuredProjects).Select(x => x.FullPath)));
             }
 
-            return shouldReturnNull ? null : configuredProjects.FirstOrDefault();
+            return configuredProjects.Any() ? configuredProjects.First() : null;
+        }
+
+        /// <summary>
+        /// Lazily gets an arbitrary project for the solution. This will not catch errors unlike where we evaluate all projects in parallel.
```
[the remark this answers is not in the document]
nagilson: Improving this comment, it was rather lazily written.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -199,19 +288,22 @@ private bool IsValidProjectFilePath(string path)
...
-        /// <remarks>Will log that the property will not work if this is true.</remarks>
-        private bool ProjectHasUserCustomizedConfiguration(ProjectInstance project, string propertyToDisableIfTrue)
+        /// <summary>
+        /// Because command-line-forwarded properties aren't in the global arguments from Properties, we need to add the TFM to the collection.
+        /// The TFM is the only property besides Configuration that isn't an MSBuild property that could affect the pre-evaluation.
+        /// This allows the pre-evaluation to correctly deduce its Publish or PackRelease value because it will know the actual TFM being used.
+        /// </summary>
+        /// <param name="oldGlobalProperties">The set of MSBuild properties that were specified explicitly like -p:Property=Foo or in other syntax sugars.</param>
+        /// <returns>The same set of global properties for the project, but with the new potential TFM based on -f or --framework.</returns>
+        private Dictionary<string, string> InjectTargetFrameworkIntoGlobalProperties(Dictionary<string, string> oldGlobalProperties)
```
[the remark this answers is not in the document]
nagilson: Done. My mind apparently still defaults to C++ land ;)

nagilson:
@dsplaisted  I have responded to the feedback here. Thanks!

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Ah... that's pretty bad. Changed, thanks for catching this

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -199,19 +289,21 @@ private bool IsValidProjectFilePath(string path)
...
-        /// <remarks>Will log that the property will not work if this is true.</remarks>
-        private bool ProjectHasUserCustomizedConfiguration(ProjectInstance project, string propertyToDisableIfTrue)
+        /// <summary>
+        /// Because command-line options that translate to MSBuild properties aren't in the global arguments from Properties, we need to add the TargetFramework to the collection.
+        /// The TargetFramework is the only command-line option besides Configuration that could affect the pre-evaluation.
+        /// This allows the pre-evaluation to correctly deduce its Publish or PackRelease value because it will know the actual TargetFramework being used.
+        /// </summary>
+        /// <param name="oldGlobalProperties">The set of MSBuild properties that were specified explicitly like -p:Property=Foo or in other syntax sugars.</param>
+        /// <returns>The same set of global properties for the project, but with the new potential TFM based on -f or --framework.</returns>
+        void InjectTargetFrameworkIntoGlobalProperties(Dictionary<string, string> oldGlobalProperties)
```
dsplaisted, line 299: Name change suggestion:

```suggestion
        void InjectTargetFrameworkIntoGlobalProperties(Dictionary<string, string> globalProperties)
```

src/Tests/Microsoft.DotNet.PackageInstall.Tests/Microsoft.DotNet.PackageInstall.Tests.csproj
```diff
@@ -64,14 +64,11 @@
       <testAssetSourceRoot>$(BaseOutputPath)/TestAsset/SampleGlobalTool</testAssetSourceRoot>
     </PropertyGroup>
     <Copy SourceFiles="SampleGlobalTool/DotnetToolSettings.xml" DestinationFolder="$(testAssetSourceRoot)" />
-    <MSBuild BuildInParallel="False" Projects="SampleGlobalTool/consoledemo.csproj" Targets="Restore" Properties="Configuration=Release;BaseOutputPath=$(testAssetSourceRoot)/bin/;BaseIntermediateOutputPath=$(testAssetSourceRoot)/obj/;ImportDirectoryBuildProps=false;ImportDirectoryBuildTargets=false;ForceRestoreToEvaluateSeparately=1"
```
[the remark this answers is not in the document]
dsplaisted: If all changes in this file are just whitespace changes, should they be reverted?

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,108 +88,109 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
             var packCommand = new DotnetPackCommand(Log, helloWorldAsset.TestRoot);
 
             packCommand
                 .Execute()
                 .Should()
                 .Pass();
 
-            var expectedAssetPath = System.IO.Path.Combine(helloWorldAsset.Path, "bin", "Release", "HelloWorld.1.0.0.nupkg");
+            var expectedAssetPath = Path.Combine(helloWorldAsset.Path, "bin", valueOfPackRelease == "true" ? "Release" : "Debug", "HelloWorld.1.0.0.nupkg");
             Assert.True(File.Exists(expectedAssetPath));
```
dsplaisted, line 137: Doing asserts like this will give better messages (including the path) if they fail:

```suggestion
            new FileInfo(expectedAssetPath).Should().Exist();
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,108 +88,109 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
         }
 
-        [Fact(Skip = "https://github.com/dotnet/sdk/issues/27066")]
-        public void It_warns_if_PackRelease_set_on_sln_but_env_var_not_used()
+        [InlineData("")]
+        [InlineData("false")]
+        [Theory]
+        public void It_packs_successfully_with_Mutlitargeting_where_net_8_and_net_7_project_defines_PackRelease_or_not(string packReleaseValue)
```
dsplaisted, line 143: Small typo

```suggestion
        public void It_packs_successfully_with_Multitargeting_where_net_8_and_net_7_project_defines_PackRelease_or_not(string packReleaseValue)
```

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,108 +88,109 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
                 .Should()
-                .Pass()
-                .And
-                .HaveStdOutContaining("NETSDK1190");
-        }
+                .Pass();
 
+            string expectedConfiguration = packReleaseValue == "false" ? "Debug" : "Release";
+            var expectedAssetPath = Path.Combine(helloWorldAsset.Path, "bin", expectedConfiguration, "HelloWorld.1.0.0.nupkg");
+            Assert.True(File.Exists(expectedAssetPath));
```
dsplaisted, line 169:
```suggestion
            new FileInfo(expectedAssetPath).Should().Exist();
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -427,141 +435,104 @@ public void It_fails_for_unsupported_rid()
...
         {
             var helloWorldAsset = _testAssetsManager
-               .CopyTestAsset("HelloWorld", "PublishReleaseHelloWorld")
-               .WithSource()
-               .WithTargetFramework(ToolsetInfo.CurrentTargetFramework);
+               .CopyTestAsset("HelloWorld", $"{optedOut}")
+               .WithSource();
 
-            System.IO.File.WriteAllText(helloWorldAsset.Path + "/Directory.Build.props", "<Project><PropertyGroup><PublishRelease>true</PublishRelease></PropertyGroup></Project>");
+            File.WriteAllText(helloWorldAsset.Path + "/Directory.Build.props", "<Project><PropertyGroup><PublishRelease>true</PublishRelease></PropertyGroup></Project>");
```
dsplaisted, line 447:
```suggestion
            File.WriteAllText(Path.Combine(helloWorldAsset.Path, "/Directory.Build.props"), "<Project><PropertyGroup><PublishRelease>true</PublishRelease></PropertyGroup></Project>");
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -427,141 +435,104 @@ public void It_fails_for_unsupported_rid()
...
 
-            new BuildCommand(Log, slnDir, "App.sln")
-               .Execute()
-               .Should()
-               .Pass();
-
-            var publishCommand = new DotnetCommand(Log)
+            new DotnetCommand(Log)
                 .WithWorkingDirectory(slnDir)
                 .Execute(@"dotnet", "publish")
```
dsplaisted, line 531: Won't this essentially run `dotnet dotnet publish`?  It seems like we wouldn't want to pass "dotnet" as the first argument to `Execute`

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -589,7 +560,7 @@ public void It_publishes_correctly_in_PublishRelease_evaluation_despite_option_f
         public void It_publishes_on_release_if_PublishRelease_property_set_in_csproj()
         {
             var helloWorldAsset = _testAssetsManager
-               .CopyTestAsset("HelloWorld", "PublishReleaseHelloWorldCsProj")
+               .CopyTestAsset("HelloWorld")
                .WithSource()
                .WithTargetFramework(ToolsetInfo.CurrentTargetFramework)
```
dsplaisted, line 565: You might want to target "net7.0" here, if you want to test that PublishRelease in the project file has an effect, since if you target .NET 8 then it will be set to true anyway and the value in the project doesn't really change anything.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -702,73 +598,61 @@ public void PublishRelease_does_not_override_Configuration_property_across_forma
...
 
-            var expectedAssetPath = System.IO.Path.Combine(helloWorldAsset.Path, "bin", "Debug", ToolsetInfo.CurrentTargetFramework, "HelloWorld.dll");
-            Assert.True(File.Exists(expectedAssetPath));
-            var releaseAssetPath = System.IO.Path.Combine(helloWorldAsset.Path, "bin", "Release", ToolsetInfo.CurrentTargetFramework, "HelloWorld.dll");
-            Assert.False(File.Exists(releaseAssetPath)); // build will produce a debug asset, need to make sure this doesn't exist either.
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: tfm, configuration: tfm == "net7.0" ? "Debug" : "Release");
+            var finalConfiguration = properties["Configuration"];
+            var finalDebugSymbols = properties["DebugSymbols"];
+            Assert.True(finalConfiguration == (tfm == "net7.0" ? "Debug" : "Release"));
+            Assert.True(finalDebugSymbols == (tfm == "net7.0" ? "true" : "false"));
```
dsplaisted, line 637: `Assert.Equals` is generally better in situations like this, because if it fails you get better information in the failure message.

```suggestion
            Assert.Equal((tfm == "net7.0" ? "Debug" : "Release"), finalConfiguration);
            Assert.Equal((tfm == "net7.0" ? "true" : "false"), finalDebugSymbols);
```

dsplaisted (commented):
I haven't made it through all the tests, but this is looking good so far.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props
```diff
@@ -24,6 +24,15 @@ Copyright (c) .NET Foundation. All rights reserved.
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
   </PropertyGroup>
 
+  <!-- Default the value of PackRelease for every project. Note that this is pre-evaluated by the CLI in an evaluation before evaluation (see ReleasePropertyProjectLocator.cs).
+  Multitargeting pack splits a single 2 TFM project into 9 separate evaluations, with different target imports. Microsoft.NET.SDK.DefaultItems.Targets ...
+  is where PublishRelease is defaulted because it depends on _TargetFrameworkVersionWithoutV. Some targets do not run in some instances of dotnet pack.
+  So, we must set PackRelease here. Even though this occurs before project import, the project import can override PackRelease later, before the pre-evaluation, so this is fine.-->
+  <PropertyGroup Condition="'$(PackRelease)' == ''">
+    <!-- v Do not depend on this value being correct until after project evaluation.-->
```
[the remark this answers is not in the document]
nagilson: It's supposed to be a down arrow.... can remove

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Update parameter name to reflect the pass by reference default of C#

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -891,4 +887,9 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1196: The SDK does not support ahead-of-time compilation. Set the PublishAot property to false.</value>
     <comment>{StrBegin="NETSDK1196: "}</comment>
   </data>
+  <data name="SolutionProjectConfigurationsConflict" xml:space="preserve">
+    <value>NETSDK1197: Multiple solution project(s) contain conflicting '{0}' values; ensure the values match. Consider using a Directory.build.props file to set all project configurations. Conflicting projects:
```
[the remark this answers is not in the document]
nagilson: Be clear that Configuration > PublishRelease or PackRelease in the error message.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+            {
+                expectedBooleanValue = "false";
+            }
+
+
+            Assert.Equal(expectedBooleanValue, finalProperties.First()[Optimize]);
+            Assert.Equal(expectedConfiguration, finalProperties.First()[Configuration]);
+
+            Assert.Equal(expectedBooleanValue, finalProperties.ElementAt(1)[Optimize]);
+            Assert.Equal(expectedConfiguration, finalProperties.ElementAt(1)[Configuration]);
```
dsplaisted, line 392: I'd use `finalProperties[0]` and `finalProperties[1]` instead of `.First()` and `ElementAt(1)`.  The methods work on enumerables, but for lists and arrays you can use direct indexing.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+                .Pass();
+
+            var properties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: "net7.0", configuration: "Release"); // this will fail if configuration is debug and TFM code didn't work.
+            string finalConfiguration = properties["Configuration"];
+            finalConfiguration.Should().BeEquivalentTo("Release");
+        }
+
+        private void VerifyCorrectConfiguration(List<Dictionary<string, string>> finalProperties, string expectedConfiguration)
+        {
+            string expectedBooleanValue = "true";
```
dsplaisted, line 381: I suggest renaming this to `expectedOptimizeValue` to make it more clear.

src/Tests/Microsoft.NET.TestFramework/ProjectConstruction/TestSolution.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        ///  Gives the property values for each project in the test solution.
+        ///  Can throw exceptions if the path to the project provided by tfm and configuration is not correct.
+        /// </summary>
+        /// <param name="targetFrameworksAndConfigurationsInOrderPerProject">A pair, first starting with the targetframework of the designated project.
+        /// The second item should be the Configuration expected of the second project.
+        /// This should match the order in which projects were added to the solution.
+        /// </param>
+        /// <returns>A dictionary of property -> value mappings for every subproject in the solution.</returns>
+        public List<Dictionary<string, string>> ProjectProperties(List<Tuple<string, string>> targetFrameworksAndConfigurationsInOrderPerProject = null)
```
dsplaisted, line 82: Suggest using tuples with named elements:

```suggestion
        public List<Dictionary<string, string>> ProjectProperties(List<(string targetFramework, string configuration)> targetFrameworksAndConfigurationsInOrderPerProject = null)
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+            testProjects.Add(referencedProject);
+            return new Tuple<TestSolution, List<TestProject>>(sln, testProjects);
+        }
+
+
+
+        [InlineData(PublishRelease, "net8.0", true)]
+        [InlineData(PublishRelease, "-p:TargetFramework=net8.0", false)]
+        [Theory]
+        public void ItUsesReleaseWithATargetFrameworkOptionNet8ForNet6AndNet7MultitargetingProjectWithPReleaseUndefined(string releaseProperty, string args, bool passDashF)
```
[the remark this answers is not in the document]
dsplaisted: Also perhaps the parameters should change to use `ToolsetInfo.CurrentTargetFramework`, so that when that revs you're not passing `-f net8.0` for a project that doesn't normally target `net8.0`.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            VerifyCorrectConfiguration(finalPropertyResults, expectedConfiguration);
+        }
+
+
+        [InlineData("true", PublishRelease)]
+        [InlineData("false", PublishRelease)]
+        [InlineData("", PublishRelease)]
+        [InlineData("true", PackRelease)]
+        // [InlineData("false", PackRelease)] This case we would expect to fail as PackRelease is enabled regardless of TFM.
```
dsplaisted, line 219: If we expect this to fail, do we have a test case covering that?  Should we?

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+            var solutionAndProjects = Setup(Log, new List<string> { firstProjectTfm }, new List<string> { secondProjectTfm }, PublishRelease, "false", "");
+            var sln = solutionAndProjects.Item1;
+
+            var dotnetCommand = new DotnetCommand(Log, publish);
+            dotnetCommand
+                .Execute(sln.SolutionPath)
+                .Should()
+                .Fail()
+                .And
+                .HaveStdErrContaining(String.Format(Strings.SolutionProjectConfigurationsConflict, PublishRelease, ""));;
```
dsplaisted, line 288: Any reason this isn't just checking for the error code like the previous test?

dsplaisted (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props
```diff
@@ -24,6 +24,15 @@ Copyright (c) .NET Foundation. All rights reserved.
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
   </PropertyGroup>
 
+  <!-- Default the value of PackRelease for every project. Note that this is pre-evaluated by the CLI in an evaluation before evaluation (see ReleasePropertyProjectLocator.cs).
+  Multitargeting pack splits a single 2 TFM project into 9 separate evaluations, with different target imports. Microsoft.NET.SDK.DefaultItems.Targets ...
```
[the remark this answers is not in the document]
nagilson: Not sure... but run dotnet pack on a new console with 2 TFMs and you get this:
image

src/Tests/Microsoft.NET.Pack.Tests/GivenThatWeWantToPackAHelloWorldProject.cs
```diff
@@ -85,108 +88,109 @@ public void It_fails_if_nobuild_was_requested_but_build_was_invoked()
...
+                .WithTargetFrameworks("net8.0;net7.0")
+                .WithProjectChanges(project =>
+                {
+                    var ns = project.Root.Name.Namespace;
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                     propertyGroup.Add(packReleaseValue != "" ? 
+                         new XElement(ns + "PackRelease", packReleaseValue)
+                         :
+                         new XElement(ns + "Foo", packReleaseValue)
+                    );
```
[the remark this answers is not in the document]
nagilson: Nope. Not picking this change because != null won't quite work the same.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+                IsExe = true
+            };
+            testProject.RecordProperties("Configuration", "Optimize", PReleaseProperty);
+            if (exePReleaseValue != "")
+            {
+                testProject.AdditionalProperties[PReleaseProperty] = exePReleaseValue;
+            }
+            var mainProject = _testAssetsManager.CreateTestProject(testProject, callingMethod: testPath, identifier: string.Join("", exeProjTfms) + PReleaseProperty);
+
+            var referencedProject = new TestProject("ReferencedProject")
```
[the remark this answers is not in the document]
nagilson: It doesn't matter one way or the other, so ill go with the simpler method of a library only.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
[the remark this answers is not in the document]
nagilson: Thanks, this used to be a pack test too but turned into just publish

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
```diff
@@ -0,0 +1,395 @@
...
+        public void ItUsesReleaseWithATargetFrameworkOptionNet8ForNet6AndNet7MultitargetingProjectWithPReleaseUndefined(string releaseProperty, string args, bool passDashF)
+        {
+            var secondProjectTfm = ToolsetInfo.CurrentTargetFramework; // Net8 here is a 'net 8+' project
+            var expectedConfiguration = Release;
+            var expectedTfm = "net8.0";
+
+            var solutionAndProjects = Setup(Log, new List<string> { "net6.0", "net7.0", "net8.0" }, new List<string> { secondProjectTfm }, releaseProperty, "", "");
+            var sln = solutionAndProjects.Item1;
+
+            var dotnetCommand = new DotnetCommand(Log, releaseProperty == PublishRelease ? publish : pack);
```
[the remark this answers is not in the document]
nagilson: Initially yes, then I realized it didn't make sense for pack to be included in this scenario because of `-f` differences

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToTestAMultitargetedSolutionWithPublishReleaseOrPackRelease.cs
nagilson: Yes, since `PackRelease` is enabled by default for all TFMs. I have changed the test to reflect the following.

dsplaisted:
🎉

Issue #27658, opened 2022-09-02
Ubuntu_22_04 Build_Release CI build fails to copy asset file intermittently


tmat:
### Describe the bug

Build:
https://dev.azure.com/dnceng-public/public/_build/results?buildId=3171&view=results

Output:
```
##[error].dotnet/sdk/7.0.100-preview.7.22377.5/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Publish.targets(248,5): 
error MSB3030: (NETCORE_ENGINEERING_TELEMETRY=Build) Could not copy the file 
"/home/vsts/work/1/s/src/Assets/TestPackages/dotnet-outputsframeworkversion/dotnet-outputsframeworkversion-netcoreapp1.0/obj/0f271755-2e7a-40ea-aea3-7b25896734fe.tmp" because it was not found.
```

### Further technical details
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and its version

nagilson:
Either the build got deleted by pipeline storage management or I cannot access it @tmat? But I can confirm this happens to me too from time to time and I just rerun the pipeline and that generally works.

marcpopMSFT:
Likely the pipeline was rerun or replaced. This is a timing issue in the build. If it happens again, we should have someone take a look without rerunning.

nagilson:
Oh hey! Looks like it already happened again https://dev.azure.com/dnceng-public/public/_build/results?buildId=18746&view=logs&j=92885c4a-db2e-5086-f9ba-51524576e2ac&t=21c5c3c8-000e-5461-4a70-e7939905edf6 but in Windows NT.
`D:\a\1\s\.dotnet\sdk\7.0.100-preview.7.22377.5\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(248,5): error MSB3030: Could not copy the file "D:\a\1\s\src\Assets\TestPackages\PackageWithFakeNativeDep\obj\6f72467f-20b3-44a8-bbf9-b2ad1476579f.tmp" because it was not found. [D:\a\1\s\src\Tests\Microsoft.NET.Sdk.BlazorWebAssembly.AoT.Tests\Microsoft.NET.Sdk.BlazorWebAssembly.AoT.Tests.csproj]`

@marcpopMSFT For my PRs it seems to preserve the pipeline runs for at least a little while, I don't why that pipeline from @tmat  would have been deleted.

[...]

PR #29180, opened 2022-11-22
Remove tmp from the default includes for tests

marcpopMSFT:
https://github.com/dotnet/sdk/issues/27658

Test builds have been failing occasionally trying to copy a .tmp file that vanishes during the build. Let's try excluding that from the test builds and see if that helps.


dsplaisted (approved):
This may help, but it would probably be better to exclude all of the `bin` and `obj` folders, or better yet figure out why they are getting created and used in the first place (instead of the repo `artifacts` folder).

PR #29258, opened 2022-11-28
Add base class for all test classes using WatchableApp.

tmat:


src/Tests/dotnet-watch.Tests/Watch/Utilities/DotNetWatchTestBase.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using Microsoft.NET.TestFramework;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Watcher.Tests;
+
+public abstract class DotNetWatchTestBase : IDisposable
```
arkalyanms, line 13: nit: documentation that can be inherited as well would be useful

arunchndr (approved):

PR #29264, opened 2022-11-29
Add tracking of the nuget cache and update to a flag zip list

marcpopMSFT:
The goal of this PR is to switch the .zip files that had the nuget chache format to a zip with a flat nupkg list of files. This includes more files than previously as we only had two zips before but were including the fully expanded nuget cache which doubled the space of the files.

I've run this before the new zip files and after.  The results are not consistent but they do show an improvement. This is comparing the before and after Count when running the dotent nuget locals command in this PR.

Before **10541** (from November) and **8449** from #31050. 
For after, I have **1080** and **1115** for the two runs.

That means the number of files getting downloaded is a few thousand less based on this data. I think this is ready to be reviewed and merged.


marcpopMSFT:
@mmitche this is the result of the conversation we had with Nikolche back in November. I'd previously removed extra feeds to reduce the pressure on nuget and this is to simplify the pre-downloaded files. Assuming I'm measuring the data right, it appears to be a significant improvement.

mmitche (approved):

PR #29310, opened 2022-11-30
Update .NET Framework reference assemblies version

ViktorHofer:
Targeting 7.0.2xx. Please let me know if I should target an even older branch instead.

The `Microsoft.NETFramework.ReferenceAssemblies` package version is hardcoded and should be updated to 1.0.3 since a new version was published over three months ago.


marcpopMSFT (approved):

marcpopMSFT:
@ViktorHofer do you know what's in the update and whether it's important for servicing or not?

ViktorHofer:
Unfortunately not. It could just be the addition of the net481 reference assemblies but am not sure.

PR #29316, opened 2022-11-30
Update target framework version in toolsetinfo for NET 8.

nagilson:
It would be ideal if this could be done automatically, though per Marc it could make the migration to 9 harder so probably not.


marcpopMSFT:
You may want nexttargetframework to stay as 8.0. Jason added that for the transition from 7 to 8.0 as we had some tests that required 7 still and some that required 8 already. I don't think those tests will work targeting 9. The alternative is to move all tests that use next target framework currently back to current with this PR. 

I like that alternative better.

nagilson:
@marcpopMSFT  Pretty sure that's already been done, there are no references to either of the nextTFM properties.
image

dsplaisted:
I don't think that "Find all references" search is working.  I see references still here:

https://github.com/dotnet/sdk/blob/5a5e3364d262dd3cd97ee1666e52cb61ee137ac6/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs#L373-L375
https://github.com/dotnet/sdk/blob/5a5e3364d262dd3cd97ee1666e52cb61ee137ac6/src/Tests/Microsoft.NET.Publish.Tests/PublishTestUtils.cs

v-wuzhai:
The NextTargetFramework constant is also used in #29158.

nagilson:
@dotnet/templating-engine-maintainers PTAL at the failure:
`Microsoft.DotNet.Cli.New.IntegrationTests.CommonTemplatesTests.FeaturesSupport(name: \"console\", buildPass: True, framework: \"net7.0\", langVersion: null, langVersionUnsupported: False, language: null, supportsNullable: True, supportsTopLevel: True, forceDisableTopLevel: False, supportsImplicitUsings: True, supportsFileScopedNs: True)` with the new `ToolsetInfo.CurrentTargetFramework`. 

https://dev.azure.com/dnceng-public/public/_build/results?buildId=110027&view=ms.vss-test-web.build-test-results-tab

nagilson:
@dotnet/dotnet-cli PTAL

baronfel:
I'm glad that we caught some build-logic updates (warning levels) with this bump :) How concerned should I be that it looks like we're losing a lot of coverage on `net7.0` TFMs on tests?

nagilson:
_I_ am too concerned since they are still running in the 7x branches and theoretically any gap in 8 and 7 should be specifically tested with 7 and 8. (I'm not saying they are, but stylistically it would be the correct thing to do and it's what I've been doing for 8.0 breaking changes.)

marcpopMSFT (approved):

baronfel:
@nagilson impeccable response, you are of course correct on all counts. Thank you :)

nagilson:
Lol! I have gotten better at covering my bases.

marcpopMSFT:
+1 to what Noah said. We should probably identify a specific set of tests that target all supported runtimes as a baseline but everything else we should roll forward each release.

PR #29350, opened 2022-12-02
extend typo correction functionality with StartsWith and Contains

YuliiaKovalova:
### Problem

Fixes [#5531](https://github.com/dotnet/templating/issues/5531)

### Solution
The list with suggestions will contain **max 10** elements that are collected based on priorities:
        1st: Starts with input token - ordered by distance = possible match length - token length
        2nd: Contains substring - the call is restricted with input token length check and ordered by distance = possible match length - token length
        3rd: Levenshtein algorithm.

### Checks
[*] Added unit tests
[ ] Added `#nullable enable` to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style) **N/A**


src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -12,35 +12,75 @@ namespace Microsoft.DotNet.Cli.Utils
 {
     public static class TypoCorrection
     {
+        private static readonly int s_minCurrentTokenLength = 3;
+        private static readonly int s_maxNumberOfSuggestions = 10;
```
vlada-shubina, line 16: I would make it method optional parameters instead.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -12,35 +12,75 @@ namespace Microsoft.DotNet.Cli.Utils
...
         /// <param name="possibleTokens">List of tokens to select from.</param>
         /// <param name="currentToken">The token that is being compared.</param>
         /// <param name="maxLevenshteinDistance">the difference between two strings, default: 3.</param>
         /// <returns>The enumerator to tokens similar to <paramref name="currentToken"/>.</returns>
         public static IEnumerable<string> GetSimilarTokens(IEnumerable<string> possibleTokens, string currentToken, int maxLevenshteinDistance = 3)
         {
-            int? bestDistance = null;
-            return possibleTokens
-                .Select(possibleMatch => (possibleMatch, distance: GetDistance(currentToken, possibleMatch)))
+            var possibleMatchesTuples = possibleTokens.Select(possibleMatch => (possibleMatch, distance: GetDistance(currentToken, possibleMatch)));
```
vlada-shubina, line 31: performance wise i suggest to do it in this order:
- `StartsWith` matches
- if > 10, return first 10 ordered by distance
- add `Contains` matches (if >= 3 chars)
- if > 10, return first 10 ordered by rules
- do Levenshtein algorithm

Distance for `StartsWith` and `Contains` is: `candidate.Length - criteria.Length`.

vlada-shubina (dismissed):
Please check the comments, there are couple of possible optimizations.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
YuliiaKovalova: done
please take a look

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -13,34 +13,97 @@ namespace Microsoft.DotNet.Cli.Utils
...
         /// <param name="currentToken">The token that is being compared.</param>
         /// <param name="maxLevenshteinDistance">the difference between two strings, default: 3.</param>
         /// <returns>The enumerator to tokens similar to <paramref name="currentToken"/>.</returns>
         public static IEnumerable<string> GetSimilarTokens(IEnumerable<string> possibleTokens, string currentToken, int maxLevenshteinDistance = 3)
         {
-            int? bestDistance = null;
-            return possibleTokens
-                .Select(possibleMatch => (possibleMatch, distance: GetDistance(currentToken, possibleMatch)))
+            var minCurrentTokenLength = 3;
+            var maxNumberOfSuggestions = 10;
```
vlada-shubina, line 30: optional parameters will be better

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -13,34 +13,97 @@ namespace Microsoft.DotNet.Cli.Utils
...
+            var currentTokenLength = currentToken.Length;
+            var startsWithAndContainsSpecificDistanceFunc = (string possibleMatch)
+                => (possibleMatch, distance: possibleMatch.Length - currentTokenLength);
+
+            var possibleMatchesTuples = possibleTokens
+                .Select(startsWithAndContainsSpecificDistanceFunc);
+
+            var matchByStartsWithTuples = possibleMatchesTuples
+               .Where(tuple => tuple.possibleMatch.StartsWith(currentToken))
+               .OrderBy(tuple => tuple.distance);
```
vlada-shubina, line 41: it would be good to do ordering of results only once where possible

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -13,34 +13,97 @@ namespace Microsoft.DotNet.Cli.Utils
...
+
+            var numberOfStartsWithSuggestions = matchByStartsWithTuples.Count();
+            if (numberOfStartsWithSuggestions >= maxNumberOfSuggestions)
+            {
+                return matchByStartsWithTuples
+                    .Take(maxNumberOfSuggestions)
+                    .Select(tuple => tuple.possibleMatch);
+            }
+
+            possibleMatchesTuples = possibleMatchesTuples.Except(matchByStartsWithTuples);
```
vlada-shubina, line 51: can you think of algorithm without using `Except`? `Except` is time-consuming.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
YuliiaKovalova: I have applied your code changes. Please take a look

vlada-shubina:
@JanKrivanek - could you please check this PR? we were working on the PR together and I would prefer an independent look. Thanks

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -2,45 +2,99 @@
...
-                .ThenByDescending(tuple => GetStartsWithDistance(currentToken, tuple.possibleMatch))
-                .TakeWhile(tuple =>
+
+            return possibleTokens.Select(token => GetTokenRank(token, currentToken, minCurrentTokenLength, maxLevenshteinDistance))
+                .Where(rankedToken => rankedToken.Rank != ComparisonRank.NoMatch)
+                .OrderBy(rankedToken => rankedToken.Rank)
+                .ThenBy(rankedToken => rankedToken.Distance)
+                .ThenBy(rankedToken => rankedToken.Rank == ComparisonRank.Levenshtein
+                    ? GetStartsWithDistance(currentToken, rankedToken.Token)
+                    : 0)
```
JanKrivanek, line 44: I like the readability and brevity of this code!

Though I must say that I'm unsure that it won't attempt to perform the unwanted sorts. I'd hope so, but quikly peeking into implementation of `ThenBy` and `OrderedEnumerable` it seems to me that it'll do the sort upfront: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq/src/System/Linq/OrderedEnumerable.cs#L21-L32

But I haven't tried - can you try to doublecheck that Levenshtein distance is not counted in case we have enough matches by previous 2 criteria? (just manually create a cese and set breakpoint into the `GetStartsWithDistance`, to verify it's not hit)

JanKrivanek (approved):
Looks good - but please verify and confirm that `GetDistance` is not being called if we have enough matches by simple string comparisons

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
YuliiaKovalova: Jan, I went through the steps and it seems to me that we call `GetDistance` in the `Where` statement when filtering out `rankedToken.Rank != ComparisonRank.NoMatch;`

YuliiaKovalova: For the sorting, we go down with the evaluated distance for all entries.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -7,40 +7,101 @@
...
+                if (numberOfSuggestions >= maxNumberOfSuggestions)
                 {
-                    (string _, int distance) = tuple;
-                    bestDistance ??= distance;
-                    return distance == bestDistance;
-                })
-                .Select(tuple => tuple.possibleMatch);
+                    return matchByStartsWith
+                        .Concat(matchByContains)
+                        .Take(maxNumberOfSuggestions)
```
ladipro, line 65: Perf suggestion: Similar to above, move the `Take` to the previous statement. `Take(maxNumberOfSuggestions - numberOfSuggestions)`

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -7,40 +7,101 @@
...
+                        .Select(s => s.PossibleMatch);
+                }
+            }
+
+            var matchByLevenshteinDistance = possibleSuggestions
+                .Select(s => new Suggestion(s.PossibleMatch, GetDistance(s.PossibleMatch, currentToken)))
+                .Where(s => !(matchByStartsWith.Contains(s) || matchByContains.Contains(s)) && s.Distance <= maxLevenshteinDistance)
```
ladipro, line 72: The first part of the condition `!(matchByStartsWith.Contains(s) || matchByContains.Contains(s))` should be moved before the `Select` to save us from calculating distance for strings that will not be used.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -7,40 +7,101 @@
...
+                .OrderBy(s => s.Distance)
+                .ThenByDescending(s => GetStartsWithDistance(currentToken, s.PossibleMatch))
```
ladipro, line 74: We are doing this even for matches that are filtered out on the following line. Suggestion: swap the two lines.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -110,7 +168,25 @@ private static int GetDistance(string first, string second)
...
+                Distance = distance;
+            }
+
+            public string PossibleMatch { get; }
+
+            public int Distance { get; }
+
+            public override bool Equals(object? obj) => obj is Suggestion suggestion && Equals(suggestion);
+
+            public bool Equals(Suggestion other) => string.Equals(PossibleMatch, other.PossibleMatch, StringComparison.OrdinalIgnoreCase);
```
ladipro, line 187: nit: This is the only case-insensitive code in this PR, all other comparisons are case-sensitive. Is it needed?

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -7,40 +7,101 @@
...
+
+        // The method takes the matches with the shortest distance
+        // e.g. (razor, 2), (pazor, 2), (pazors, 3) => (razor, 2), (pazor, 2)
+        private static IEnumerable<Suggestion> FilterByShortestDistance(this IEnumerable<Suggestion> possibleMatches)
+        {
+            int? bestDistance = null;
+
+            return possibleMatches.TakeWhile(s =>
+            {
+                (string _, int distance) = (s.PossibleMatch, s.Distance);
```
ladipro, line 92: nit: Why not simply:

```suggestion
                int distance = s.Distance;
```

ladipro (commented):
I have left perf suggestions inline.

ladipro:
Also, one minor comment on the `List.Contains` calls, which make it look like O(N^2). I understand that in practice the size of the results is very small (hard-coded to 10 currently) but still, there should be a more elegant and future-proof way of doing it. You could, for example, mark the "already consumed" suggestions with a flag, making it possible to skip them in subsequent queries in constant time. That would actually make a case for keeping `Suggestion` as a mutable class.

Completely optional though, if the max number of suggestions stays low, there is no problem.

src/Cli/Microsoft.DotNet.Cli.Utils/TypoCorrection.cs
```diff
@@ -110,7 +168,25 @@ private static int GetDistance(string first, string second)
 
             // Return the computed edit distance
             return rows[curRow][m];
+        }
+
+        internal class Suggestion
```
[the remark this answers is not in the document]
YuliiaKovalova: resolved with adding IsSelected property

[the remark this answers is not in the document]
ladipro: Thank you! In line with our aspiration to a higher code quality, please consider commenting the class. Also super-nit: the class can be `sealed`.

ladipro (approved):

PR #29356, opened 2022-12-02
Add support for default values of methods parameters in the GenAPI.

andriipatsula:
- https://github.com/dotnet/arcade/issues/11814
- https://github.com/dotnet/arcade/issues/11734
- Add and update Unit Tests
- Introduce new parameter `IncludeVisibleOutsideOfAssembly` to the CLI and MSBuild frontends.
- Upgrade the version of Microsoft.CodeAnalysis to `4.5.0-3.22601.3`


src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
```diff
@@ -69,18 +68,31 @@ public CSharpSyntaxWriter(string? exceptionMessage)
...
         }
 
+        /// <inheritdoc />
+        public override SyntaxNode? VisitOperatorDeclaration(OperatorDeclarationSyntax node)
+        {
+            // visit subtree first to normalize type names.
+            OperatorDeclarationSyntax? rs = (OperatorDeclarationSyntax?)base.VisitOperatorDeclaration(node);
+            if (rs is null) return rs;
+
+            return rs
```
premun, line 91:
```suggestion

            return rs?
```

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpSyntaxWriter.cs
```diff
@@ -69,18 +68,31 @@ public CSharpSyntaxWriter(string? exceptionMessage)
...
             }
 
             return rs.WithParameterList(rs.ParameterList.WithTrailingTrivia(SyntaxFactory.Space));
         }
 
+        /// <inheritdoc />
+        public override SyntaxNode? VisitOperatorDeclaration(OperatorDeclarationSyntax node)
+        {
+            // visit subtree first to normalize type names.
+            OperatorDeclarationSyntax? rs = (OperatorDeclarationSyntax?)base.VisitOperatorDeclaration(node);
```
premun, line 88: Should this be?
```suggestion
            OperatorDeclarationSyntax? rs = base.VisitOperatorDeclaration(node) as OperatorDeclarationSyntax;
```

If it returns `null`, I don't know if the cast would fail always? (I haven't tried casting into a nullable)

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -44,10 +44,16 @@ public class GenAPITask : TaskBase
...
-        /// Include all API's not just public APIs. The default is public only.
+        /// Include internal API's. Default is false
         /// </summary>
         public bool IncludeVisibleOutsideOfAssembly { get; set; }
 
+
+        /// <summary>
+        /// Include public, protected and private API's. Default is public only.
+        /// </summary>
+        public bool IncludeEffectivelyPrivateSymbols { get; set;  }
```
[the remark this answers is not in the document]
RikkiGibson: note that protected APIs are not effectively private if they are contained in a non-sealed class which is accessible from the other assembly.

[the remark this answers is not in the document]
ViktorHofer: Why don't you unconditionally pass in `includeEffectivelyPrivateSymbols: true` to the [`IsVisibleOutsideOfAssembly`](https://github.com/dotnet/sdk/blob/bf4da7e01882fe5b680eaa9d30d6f33fbdf1f23d/src/Microsoft.DotNet.ApiSymbolExtensions/SymbolExtensions.cs#L87-L95) method? Do you need that parameter in the frontend? Would that ever be set to false?

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
```diff
@@ -10,6 +10,8 @@
     <NoWarn>$(NoWarn);NU5128</NoWarn>
     <Nullable>enable</Nullable>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);_AddBuildOutputToPackageCore;_AddBuildOutputToPackageDesktop</TargetsForTfmSpecificContentInPackage>
+    <_MicrosoftCodeAnalysisVersion>4.5.0-3.22601.3</_MicrosoftCodeAnalysisVersion>
+    <_MicrosoftCodeAnalysisVersion Condition="'$(DotNetBuildFromSource)' == 'true'">$(MicrosoftCodeAnalysisPackageVersion)</_MicrosoftCodeAnalysisVersion>
```
ViktorHofer, line 14: These two properties are dead code in this project file as the PackageReferences to Microsoft.CodeAnalysis aren't defined here anymore.

This should instead now be changed here: https://github.com/dotnet/sdk/blob/bf4da7e01882fe5b680eaa9d30d6f33fbdf1f23d/src/GenAPI/Microsoft.DotNet.GenAPI/Microsoft.DotNet.GenAPI.csproj#L9-L10

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
```diff
@@ -214,8 +214,10 @@ namespace Foo
                 {
                     public class Car
                     {
-                        public int? Wheels { get; }
+                        public int? Drivers { get; }
+                        public int Wheels { get => 4; }
                         public bool IsRunning { get; set; }
+                        public bool Is4x4 { get => false; set { } }
```
ViktorHofer, line 220: Can you please describe the ordering algorithm? Is it the exact same as in the old GenAPI? Asking as we want to preserve the current ordering for the reference source files in dotnet/runtime, i.e. [System.Runtime.cs](https://raw.githubusercontent.com/dotnet/runtime/main/src/libraries/System.Runtime/ref/System.Runtime.cs).

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
andriipatsula: updated. thanks

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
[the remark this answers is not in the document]
andriipatsula: Nope, I don't need it. I thought you might need to have it set to false. If not - I'll set to true.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
ViktorHofer: I was confused that the tests (which weren't correctly sorted) didn't fail. I assume we don't yet have roundtripping tests?

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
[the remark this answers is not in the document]
ViktorHofer: Yes, we need `includeEffectivelyPrivateSymbols` be false for ApiCompat but that shouldn't require you to expose the parameter as a task / CLI option in GenAPI.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
andriipatsula: >I was confused that the tests (which weren't correctly sorted) didn't fail.
Nope, the tests for previous iteration did fail. That's why I made changes to expected code snippets in the last iteration.

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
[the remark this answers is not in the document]
andriipatsula: I meant, for use cases like `Generate reference source files from implementation assemblies` in the Runtime repository; do you expect to have parameter `includeEffectivelyPrivateSymbols` set to true? - If yes, I'll remove options from MSBuild task and CLI tool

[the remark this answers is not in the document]
ViktorHofer: Yes, that would always be true for our use-case in dotnet/runtime.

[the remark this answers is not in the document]
andriipatsula: Updated

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -44,16 +44,10 @@ public class GenAPITask : TaskBase
         public string[]? ExcludeAttributesFiles { get; set; }
 
         /// <summary>
-        /// Include internal API's. Default is false
+        /// Include private and internal API's. Default is false
```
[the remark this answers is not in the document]
andriipatsula: will revert to `Include internal API's. Default is false`

premun (approved):

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
```diff
@@ -44,7 +44,7 @@ public class GenAPITask : TaskBase
         public string[]? ExcludeAttributesFiles { get; set; }
 
         /// <summary>
-        /// Include all API's not just public APIs. The default is public only.
+        /// IncludeInclude internal API's. Default is false.
```
ViktorHofer, line 47: Typo: double "Include"

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
ViktorHofer, line 1: There weren't any changes in this file except that you changed the BOM. Let's try to avoid this in the future.

src/GenAPI/Microsoft.DotNet.GenAPI/AssemblySymbolOrderer.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// Provides ordering for namespaces, types and members.
+    /// </summary>
+    public static class AssemblySymbolOrderer
+    {
+        /// <summary>
+        /// Sorts the elements of a INamespaceSymbol.
+        /// </summary>
+        /// <param name="namespaces">List of namespaces to be sorted.</param>
+        /// <returns>Returns namespaces in sorted order.</returns>
+        public static IEnumerable<INamespaceSymbol> Order(IEnumerable<INamespaceSymbol> namespaces)
```
ViktorHofer, line 21: These methods would be perfect candidates for extension methods. Did you consider such instead of static access via the `AssemblySymbolOrderer` class?

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -74,8 +76,9 @@ private void Visit(IAssemblySymbol assembly)
 
             Document document = project.AddDocument(assembly.Name, compilationUnit);
             document = Simplifier.ReduceAsync(document).Result;
-            compilationUnit = document.GetSyntaxRootAsync().Result!;
+            document = Formatter.FormatAsync(document, DefineFormattingOptions()).Result;
```
ViktorHofer, line 79: Isn't there a synchronous API that you could call instead? Calling onto Result of the async call is an anti-pattern afaik.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -89,7 +89,9 @@ public static void Run(Context context)
 
             var compositeFilter = new CompositeFilter()
                 .Add<ImplicitSymbolsFilter>()
-                .Add(new SymbolAccessibilityBasedFilter(context.IncludeVisibleOutsideOfAssembly));
+                .Add(new SymbolAccessibilityBasedFilter(
+                    context.IncludeVisibleOutsideOfAssembly,
+                    /*IncludeEffectivelyPrivateSymbols*/true));
```
ViktorHofer, line 94: There's an in-built syntax for that in C# that avoids comment as they can get out-of-sync soon (paramName: value):

L94:
```
includeEffectivelyPrivateSymbols: true));
```

ViktorHofer (commented):
Left some more post-merge feedback.

src/GenAPI/Microsoft.DotNet.GenAPI.Task/GenAPITask.cs
andriipatsula: Updated. Thanks. will update in https://github.com/dotnet/sdk/pull/29471

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
andriipatsula: Ok.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
andriipatsula: There is a synchronous API for `Formatter` class, but there are not for `Simplifier` class. Plus, synchronous API works with `SyntaxNode` directly and asynchronous - with Document. In order to work with the `Formatter` and the `Simplifier` in the same way, I'd prefer to have:
```C#
Document document = project.AddDocument(assembly.Name, compilationUnit);
document = Simplifier.ReduceAsync(document).Result;
document = Formatter.FormatAsync(document, DefineFormattingOptions()).Result;
```
rather than
```C#
Document document = project.AddDocument(assembly.Name, compilationUnit);
document = Simplifier.ReduceAsync(document).Result;
compilationUnit = document.GetSyntaxRootAsync().Result!;
compilationUnit = Formatter.Format(compilationUnit, Formatter.Annotation, _adhocWorkspace, DefineFormattingOptions());
```

src/GenAPI/Microsoft.DotNet.GenAPI/AssemblySymbolOrderer.cs
andriipatsula: Thanks. Will update it https://github.com/dotnet/sdk/pull/29471

PR #29369, opened 2022-12-02
Change the cppcli tests to use the package reference support 

marcpopMSFT:
so they don't need the matching apphost


src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
```diff
@@ -24,21 +24,22 @@ public GivenThatWeWantToBuildACppCliProject(ITestOutputHelper log) : base(log)
...
             var testAsset = _testAssetsManager
                 .CopyTestAsset("NetCoreCsharpAppReferenceCppCliLib")
-                .WithSource();
+                .WithSource()
+                .WithProjectChanges((projectPath, project) => AddPackageReference(projectPath, project, "NewtonSoft.Json", "13.0.1"));
 
             // build projects separately with BuildProjectReferences=false to simulate VS build behavior
             new BuildCommand(testAsset, "NETCoreCppCliTest")
-                .Execute("-p:Platform=x64")
+                .Execute("-p:Platform=x64", "-p:EnableManagedpackageReferenceSupport=true")
```
dsplaisted, line 37: Instead of passing this on the command line, can we make it part of the project, either by including it directly in the test asset, or else in the `WithProjectChanges` call?

Same with other instances of this.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildACppCliProject.cs
```diff
@@ -76,28 +77,30 @@ public void It_builds_and_runs_with_package_reference()
...
         {
             var testAsset = _testAssetsManager
                 .CopyTestAsset("CppCliLibWithWpfFrameworkReference")
-                .WithSource();
+                .WithSource()
+                .WithProjectChanges((projectPath, project) => AddPackageReference(projectPath, project, "NewtonSoft.Json", "13.0.1"));
 
             new BuildCommand(testAsset)
-                .Execute("-p:Platform=x64")
+                .Execute("-p:Platform=x64", "-p:EnableManagedpackageReferenceSupport=true", @"/bl:c:\repos\sdk\my.binlog")
```
dsplaisted, line 103: Left over debugging argument

```suggestion
                .Execute("-p:Platform=x64", "-p:EnableManagedpackageReferenceSupport=true")
```

dsplaisted (approved):

PR #29371, opened 2022-12-02
Remove unnecessary sources from the test used nuget.config file to reduce 429s

marcpopMSFT:
Every package is simultaneously queried on every source. We believe this is what's causing us to get throttled and results in 429s and 503 errors in our tests when there are lots of PRs open at once.

### DIFF
--- a/build/RunTestsOnHelix.cmd
+++ b/build/RunTestsOnHelix.cmd
@@ -28,7 +28,18 @@ robocopy %HELIX_CORRELATION_PAYLOAD%\t\TestExecutionDirectoryFiles %TestExecutio
 REM call dotnet new so the first run message doesn't interfere with the first test
 dotnet new --debug:ephemeral-hive
 REM avoid potetial concurrency issues when nuget is creating nuget.config
-dotnet nuget list source
+dotnet nuget list source --configfile %TestExecutionDirectory%\nuget.config
 REM We downloaded a special zip of files to the .nuget folder so add that as a source
-dotnet new nugetconfig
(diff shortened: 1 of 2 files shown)


marcpopMSFT:
@dtivel The first iteration of this PR had an odd build failure caused by the trustedroots.Tests:
https://dev.azure.com/dnceng-public/public/_build/results?buildId=100567&view=logs&j=1a72c28d-7a68-5d41-adbe-a7fdc2aec7fb&t=14e40ebc-73fd-5974-cd7e-50927e145076&l=603
`D:\a\1\s\.dotnet\sdk\7.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(254,5): error MSB3026: Could not copy "D:\a\1\s\artifacts\bin\Tests\Microsoft.DotNet.Cli.Utils\Debug\net7.0\MSBuild.dll" to "D:\a\1\s\artifacts\bin\redist\Debug\publish\MSBuild.dll". Beginning retry 1 in 1000ms. The process cannot access the file 'D:\a\1\s\artifacts\bin\redist\Debug\publish\MSBuild.dll' because it is being used by another process.  [D:\a\1\s\src\Tests\trustedroots.Tests\trustedroots.Tests.csproj]`

From poking around the code, it's because both test projects have the same output path:
    `<OutputPath>$(ArtifactsBinDir)redist\$(Configuration)</OutputPath>`

Do the trusted root tests need to go to that location? The dotnet.Tests go there for in-process tests (so says the comment in the .csproj file though Daniel may know the history there)? It'd be better if the output paths were not overlapping like that as it can lead to copy errors.

mmitche (approved):

PR #29379, opened 2022-12-03
extract common logging functionality into ApiSymbolExtensions

smasher164:
This change refactors APICompat and GenAPI to use a common set of logging abstractions. These include `ILog`, `ISuppressableLog`, `MSBuildLog`, `ConsoleLog`, `SuppressableConsoleLog`, `SuppressableMSBuildLog`, and `SuppressableTestLog`.

```
ILog
├─ConsoleLog────────────┬─SuppressableConsoleLog
├─ISuppressableLog──────┤
│ └─SuppressableTestLog │
└─MSBuildLog────────────┴─SuppressableMSBuildLog
```

Because we don't want the console logger to depend on `Microsoft.Build.*` packages, we exclude files that depend on it from `ApiSymbolExtensions`'s build. The depending packages (in this case `GenAPI.Task` and `ApiCompat.Task`) are required to include those files explicitly.


ViktorHofer:
The console logger must function without a reference to the msbuild packages. We must not mix frontend stacks. That's why we had the copy of the MessageImportance class to work on both msbuild and the CLI.

That also means that the ApiSymbolExtensions must not have a PackageReference to msbuild packages. We would either need to introduce an additional project ie ApiSymbolExtensions.MSBuild.csproj or just have the msbuild loggers in ApiSymbolExtensions, not compile them and then manually include them in the ApiCompat and GenAPI frontends.

src/Microsoft.DotNet.ApiSymbolExtensions/Logging/ILog.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Build.Framework;
```
andriipatsula, line 4: I agree with Viktor, mixing frontends is not the best option. I believe, we must remove the usage of `Microsoft.Build.Framework` from here.

src/GenAPI/Microsoft.DotNet.GenAPI/Microsoft.DotNet.GenAPI.csproj
```diff
@@ -9,10 +9,10 @@
     <_MicrosoftCodeAnalysisVersion>4.0.1</_MicrosoftCodeAnalysisVersion>
     <_MicrosoftCodeAnalysisVersion Condition="'$(DotNetBuildFromSource)' == 'true'">$(MicrosoftCodeAnalysisPackageVersion)</_MicrosoftCodeAnalysisVersion>
   </PropertyGroup>
-
+
```
andriipatsula, line 12:
```suggestion

```

src/GenAPI/Microsoft.DotNet.GenAPI.Task/Microsoft.DotNet.GenAPI.Task.csproj
```diff
@@ -22,11 +22,17 @@
...
+  <ItemGroup>
+    <Compile Include="..\..\Microsoft.DotNet.ApiSymbolExtensions\Logging\MSBuildLog.cs" />
+  </ItemGroup>
 
   <ItemGroup>
     <!-- ExcludeAssets="Runtime" is being set to avoid adding package references and dependencies of package and project references into the package. -->
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildPackageVersion)" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" ExcludeAssets="Runtime" />
     <ProjectReference Include="..\Microsoft.DotNet.GenAPI\Microsoft.DotNet.GenAPI.csproj" ExcludeAssets="Runtime" />
+    <ProjectReference Include="..\..\Microsoft.DotNet.ApiSymbolExtensions\Microsoft.DotNet.ApiSymbolExtensions.csproj" ExcludeAssets="runtime" />
```
andriipatsula, line 35: We have a ProjectReference to `Microsoft.DotNet.ApiSymbolExtensions.csproj` in `Microsoft.DotNet.GenAPI`. I think, we can omit this reference here

src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/Microsoft.DotNet.ApiCompat.Task.csproj
```diff
@@ -23,10 +23,16 @@
...
+  <!-- Include MSBuild logger -->
+  <ItemGroup>
+    <Compile Include="..\..\Microsoft.DotNet.ApiSymbolExtensions\Logging\MSBuildLog.cs" />
+  </ItemGroup>
 
   <ItemGroup>
     <!-- ExcludeAssets="Runtime" is being set to avoid adding package references and dependencies of package and project references into the package. -->
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildPackageVersion)" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" ExcludeAssets="Runtime" />
+    <ProjectReference Include="..\..\Microsoft.DotNet.ApiSymbolExtensions\Microsoft.DotNet.ApiSymbolExtensions.csproj" ExcludeAssets="Runtime" />
```
andriipatsula, line 35: we have reference to `Microsoft.DotNet.ApiSymbolExtensions.csproj` in `Microsoft.DotNet.ApiCompatibility.csproj`. I think we can omit this line here

andriipatsula (approved):

smasher164:
A lot of the tests are failing with the error `Could not load file or assembly 'Microsoft.CodeAnalysis`. I need to dig in to see if there's a code change here that affected that dependency from being loaded.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/SuppressableConsoleLog.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.DotNet.ApiCompat.Tool
+{
+    /// <summary>
+    /// Class that can log Suppressions to the Console, by implementing ConsoleLog and ISuppressableLog.
+    /// </summary>
+    internal class SuppressableConsoleLog : ConsoleLog, ISuppressableLog
+    {
+        private readonly ISuppressionEngine _suppressionEngine;
+        public bool SuppressionWasLogged { get; private set; }
```
ViktorHofer, line 17: `/// <inheritdoc />` is missing here

Also please add an empty line to separate the private field from the property (code convention).

src/Microsoft.DotNet.ApiSymbolExtensions/Logging/MessageLevel.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.ApiSymbolExtensions.Logging
+{
+    public enum MessageLevel
```
ViktorHofer, line 6: Is that a new type? And if yes, what is it needed for?

ViktorHofer (commented):
Aside from a few questions, looks great. Thanks

PR #29393, opened 2022-12-05
Add net8.0 ILCompiler tests

sbomer:
Now that the SDK can target net8.0, we can test 8.0 scenarios (and scenarios where the 8.0 SDK targets 7.0).
I also added a check that that the built-in PublishAot scenario doesn't produce packagereference warnings. This depends on https://github.com/dotnet/runtime/pull/79240.


LakshanF (approved):
Thanks for doing this! Neat way of checking FX and AOT package versions! 

Not related to this PR but to this test is general: Its bad that we might have to revisit to change ExplicitPackageVersion but at least its only on one place. The test scenarios are complex and are going to be challenging to follow but can't see an obvious way to make these easier to understand. Perhaps when we revisit testing OSX support, I can create a high-level test spec for this.

agocke (approved):
LGTM

PR #29416, opened 2022-12-06
Update minimum msbuild to 17.4

marcpopMSFT:
Updating the minimum version per this document: https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules


baronfel (approved):
🚂

PR #29435, opened 2022-12-07
Fix bad merge that incorrectly didn't exclude some packages

### LINKED ISSUE #16447: Suspicious code found inside the ResolvePackageAssets task
While investigating what files are copied into the publish directory, I ended up down the rabbit hole reading the source code of the `ResolvePackageAssets` task. There, I stumbled on [three lines](https://github.com/dotnet/sdk/blob/14707360491225abea19a7c00f5b1ac17696659e/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs#L1426-L1429) which seem very suspicious:

```c#
// publishPackageExclusions is a HashSet<string>
if (publishPackageExclusions.Contains(library.Name))
{
    publishPackageExclusions.Add(library.Name);
}
```

Testing if a set contains an element to add it in that set doesn't make sense. Most probably the `Contains` test was meant to be performed on another collection. Maybe it is the result of a bad merge? 

/cc @nguerrera @dsplaisted

marcpopMSFT:
Fixes #16447

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
@@ -1747,7 +1747,7 @@ namespace Microsoft.NET.Build.Tasks
                         {
                             //  Libraries explicitly marked as exclude from publish should be excluded from
                             //  publish even if there are other transitive dependencies to them
-                            if (publishPackageExclusions.Contains(library.Name))
+                            if (excludeFromPublishPackageIds.Contains(library.Name))
                             {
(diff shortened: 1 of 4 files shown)


dsplaisted (approved):
Do we know if there's any observable behavior change as a result of this?

marcpopMSFT:
> Do we know if there's any observable behavior change as a result of this?
It does. I'll see about adding a test for this. If you have projects A->B and both have package references to the same package and A had publish=false, this change will make it so the package doesn't get published when publishing A.  Prior to this, we would always publish transitive references. The behavior worked correctly if the direct reference was the only one.

PR #29441, opened 2022-12-08
Use linker matching TFM

### LINKED ISSUE #3029: PublishTrimmed should use a linker that matches the TargetFramework
We need a long-term solution for back-compat issues when using a newer SDK to trim apps targeting older TFMs. Some issues encountered so far:
- https://github.com/dotnet/linker/issues/2998
- https://github.com/dotnet/runtime/pull/72143#issuecomment-1233715839

We will hit a similar situation with planned breaking changes in 8.0, where it would be difficult to maintain a back-compat path. Maybe there are more examples, but this one comes to mind:
- https://github.com/dotnet/runtime/issues/117157

The solution will most likely (pending the outcome of broader discussion) be to use the 6.0 linker when targeting .net6, etc. This issue tracks the technical side of that proposed solution.

sbomer:
Use a linker based on the current TargetFramework. This relies on KnownILLinkPack and the implicit packagereference mechanism used also for NativeAot to restore the right version of illink. See the KnownILLinkPack definitions here: https://github.com/dotnet/installer/pull/15106.

Fixes https://github.com/dotnet/linker/issues/3029

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -631,7 +631,8 @@ namespace Microsoft.NET.Build.Tasks
 
             
             // Crossgen and ILCompiler have RID-specific bits.
-            if (toolPackType is ToolPackType.Crossgen2 or ToolPackType.ILCompiler) {
+            if (toolPackType is ToolPackType.Crossgen2 or ToolPackType.ILCompiler)
+            {
                 var packNamePattern = knownPack.GetMetadata(packName + "PackNamePattern");
                 var packSupportedRuntimeIdentifiers = knownPack.GetMetadata(packName + "RuntimeIdentifiers").Split(';');
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -377,8 +377,8 @@ namespace Microsoft.NET.Build.Tasks
                     return;
                 }
             }
-            // Only add ILLink for non-AOT trimming.
-            else if (TrimmingEnabled)
+
+            if (TrimmingEnabled)
             {
                 if (!AddToolPack(ToolPackType.ILLink, _normalizedTargetFrameworkVersion, packagesToDownload, implicitPackageReferences))
                 {
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -44,6 +44,8 @@ namespace Microsoft.NET.Build.Tasks
 
         public bool ReadyToRunUseCrossgen2 { get; set; }
 
+        public bool TrimmingEnabled { get; set; }
+
         public bool AotEnabled { get; set; }
 
         public string RuntimeIdentifier { get; set; }
@@ -72,7 +74,9 @@ namespace Microsoft.NET.Build.Tasks
 
         public ITaskItem[] KnownCrossgen2Packs { get; set; } = Array.Empty<ITaskItem>();
 
-        public ITaskItem[] KnownILCompilerPacks { get; set; } = Array.Empty<ITaskItem>();        
+        public ITaskItem[] KnownILCompilerPacks { get; set; } = Array.Empty<ITaskItem>();
+
+        public ITaskItem[] KnownILLinkPacks { get; set; } = Array.Empty<ITaskItem>();
 
         [Required]
         public string NETCoreSdkRuntimeIdentifier { get; set; }
@@ -354,9 +358,11 @@ namespace Microsoft.NET.Build.Tasks
                 }
             }
 
+            List<ITaskItem> implicitPackageReferences = new List<ITaskItem>();
+
             if (ReadyToRunEnabled && ReadyToRunUseCrossgen2)
             {
-                if (!AddAotOrR2RRuntimePackage(AotPackageType.Crossgen2, _normalizedTargetFrameworkVersion, packagesToDownload))
+                if (!AddToolPack(ToolPackType.Crossgen2, _normalizedTargetFrameworkVersion, packagesToDownload, implicitPackageReferences))
                 {
                     Log.LogError(Strings.ReadyToRunNoValidRuntimePackageError);
                     return;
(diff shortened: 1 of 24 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -590,68 +622,79 @@ private bool AddAotOrR2RRuntimePackage(AotPackageType packageType, Version norma
...
-            if (EnableRuntimePackDownload)
-            {
-                // We need to download the runtime pack
-                TaskItem runtimePackToDownload = new TaskItem(runtimePackName);
-                runtimePackToDownload.SetMetadata(MetadataKeys.Version, packVersion);
-                packagesToDownload.Add(runtimePackToDownload);
-            }
+            
+            // Crossgen and ILCompiler have RID-specific bits.
+            if (toolPackType is ToolPackType.Crossgen2 or ToolPackType.ILCompiler) {
```
dsplaisted, line 634: Nit: Usually we put opening braces on separate lines.

```suggestion
            if (toolPackType is ToolPackType.Crossgen2 or ToolPackType.ILCompiler)
            {
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
```diff
@@ -111,6 +112,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                                 DisableTransitiveFrameworkReferenceDownloads="$(DisableTransitiveFrameworkReferenceDownloads)"
                                 KnownCrossgen2Packs="@(KnownCrossgen2Pack)"
                                 KnownILCompilerPacks="@(KnownILCompilerPack)"
+                                KnownILLinkPacks="@(KnownILLinkPack)"
```
dsplaisted, line 115: Where is `KnownILLinkPack` consumed?  Is it supposed to be converted to a `PackageReference`?  How are the .props and .targets inside it (which I assume used to be in the `Microsoft.NET.ILLink.Tasks` SDK) imported?

dsplaisted (commented):
Looks good, but I'm missing how the ILLink targets actually end up getting imported.

Also it looks like this means that ILLink will always need to be downloaded in a NuGet restore instead of being included in the SDK.  That means the scenario won't work "offline" anymore.  Probably this is OK, but I wanted to make sure that's expected.

dsplaisted (approved):

vitek-karas:
If we're always going to download ILLink - should we remove it from the SDK isntaller then? What's the point of having it there (or maybe I don't understand something).

dsplaisted:
> If we're always going to download ILLink - should we remove it from the SDK isntaller then? What's the point of having it there (or maybe I don't understand something).

I believe this PR does remove it from being bundled in the SDK installer, by removing this line:

```xml
 <BundledSdk Include="Microsoft.NET.ILLink.Tasks" Version="$(MicrosoftNETILLinkTasksPackageVersion)" />
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1182,7 +1182,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.Sdk.VisualBasic.targets" Condition="'$(Language)' == 'VB'" />
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.Sdk.FSharp.targets" Condition="'$(Language)' == 'F#'" />
   <Import Project="$(ILCompilerTargetsPath)" Condition="'$(PublishAot)' == 'true'"/>
-  <Import Project="$(ILLinkTargetsPath)" Condition="'$(Language)' != 'C++'" />
+  <Import Project="$(ILLinkTargetsPath)" Condition="'$(ILLinkTargetsPath)' != '' And '$(Language)' != 'C++'" />
```
ViktorHofer, line 1185: @sbomer given that the nuget package is now always used, why do we still need this and not just make sure that the targets file gets imported from the nuget package automatically via the file naming convention?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -15,6 +15,10 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DefaultCopyToPublishDirectoryMetadata Condition="'$(DefaultCopyToPublishDirectoryMetadata)' == ''">true</DefaultCopyToPublishDirectoryMetadata>
     <_GetChildProjectCopyToPublishDirectoryItems Condition="'$(_GetChildProjectCopyToPublishDirectoryItems)' == ''">true</_GetChildProjectCopyToPublishDirectoryItems>
     <IsPublishable Condition="'$(IsPublishable)' == ''">true</IsPublishable>
+    <!-- PublishAot depends on PublishTrimmed. This must be set early enough for the KnownILLinkPack to be restored. -->
+    <PublishTrimmed Condition="'$(PublishTrimmed)' == '' And '$(PublishAot)' == 'true'">true</PublishTrimmed>
+    <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</_IsTrimmingEnabled>
```
jonathanpeppers, line 20: @sbomer dotnet/maui has a `netstandard2.0` class library where `$(IsTrimmable)` is set to true, because it is safe to aggressively trim.

This causes the build error:

    Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(90,5): error NETSDK1195: Unable to optimize assemblies for size: a valid runtime package was not found. Either set the PublishTrimmed property to false, or use a supported target framework when publishing.

We had to turn off `$(IsTrimmable)` for the `netstandard2.0` build (this assembly is used from an MSBuild/VS context, but trimming doesn't matter there).

Is this an expected case?

richlander:
> I believe this PR does remove it from being bundled in the SDK installer,

Is that a breaking change? If so, has it been documented?

dsplaisted:
> > I believe this PR does remove it from being bundled in the SDK installer,
> 
> Is that a breaking change? If so, has it been documented?

I don't think we consider it a breaking change.  The change is that instead of being included directly in the .NET SDK, the linker will be downloaded during NuGet restore.  That could break offline scenarios, but I don't think that's something we track.

PR #29450, opened 2022-12-08
Update Microsoft.NET.ApiCompat.targets for net8.0

SteveSandersonMS:
It looks like there's now an incorrect path inside `Microsoft.NET.ApiCompat.targets`, and it's blocking the `aspnetcore` repo from ingesting SDK updates (https://github.com/dotnet/aspnetcore/pull/44925).


SteveSandersonMS:
You might want to use some MSBuild property to get the current target TFM instead of hardcoding `net8.0`, but I'll leave that up to the SDK people.

wtgodbe (approved):
@dsplaisted @marcpopMSFT any reason we shouldn't take this?

PR #29506, opened 2022-12-12
Set server GC for F# FSC/FSI

vzarytovskii:
As we figured out [here](https://github.com/dotnet/fsharp/pull/13740#issuecomment-1304870562), SDK is not getting `runtimeconfig.json` files from F# itself, but from the https://github.com/dotnet/sdk/blob/release/7.0.2xx/src/Layout/tool_fsharp/tool_fsc.csproj

https://github.com/dotnet/sdk/blob/a50339473c041a9099a27b808b09e1faf506920b/src/Layout/redist/targets/GenerateLayout.targets#L254-L262

Hopefully, it's the proper place to make the change, if not, please advise what would be thecorrect one.

**Update:** will also need to verify that these changes do affect runtimeconfig, when build is complete.

### DIFF
--- a/src/Layout/tool_fsharp/tool_fsc.csproj
+++ b/src/Layout/tool_fsharp/tool_fsc.csproj
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFramework>$(SdkTargetFramework)</TargetFramework>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
   </PropertyGroup>
 
   <ItemGroup>


baronfel:
I cloned this down, built it, and verified that the server GC was set in the runtimeconfig.json:
 

```
../FSharp on vzarytovskii-patch-1
11:34:56 ❯ cat *.runtime*
{
  "runtimeOptions": {
    "tfm": "net8.0",
    "framework": {
      "version": "8.0.0-alpha.1.22605.1"
    },
    "configProperties": {
      "System.GC.Server": true
    }
  }
}
{
  "runtimeOptions": {
    "tfm": "net8.0",
    "framework": {
      "version": "8.0.0-alpha.1.22605.1"
    },
    "configProperties": {
      "System.GC.Server": true
    }
  }
}
```

This was on the main-targeted form of this PR, but the result should be the same.

baronfel (approved):

PR #29508, opened 2022-12-12
Fix garbage collection for aliased workload packs

### LINKED ISSUE #29212: [NETE2ESDK]With .NET 7.0.101 SDK installed, Warning "Workload garbage collection failed with error: An item with the same key has already been added. Key: (Microsoft.NETCore.App.Runtime.Mono.android-arm, 6.0.12) displayed when installing andriod workload in CLI.
Repro steps:
1.installed .NET7.0.101 SDK(December update)
2.In CLI, execute: dotnet workload install android --skip-manifest-update

Expected Result:
workload should be installed successfully, and no warnings should be displayed in the output.

Actual Result:
Warning "Workload garbage collection failed with error: An item with the same key has already been added. Key: (Microsoft.NETCore.App.Runtime.Mono.android-arm, 6.0.12) displayed

image

dsplaisted:
Fixes #29212

### DIFF
--- a/src/Assets/TestReleases/TestRelease/releases-index.json
+++ b/src/Assets/TestReleases/TestRelease/releases-index.json
@@ -20,7 +20,7 @@
             "latest-sdk": "3.1.404",
             "product": ".NET Core",
             "support-phase": "lts",
-            "eol-date": "2022-12-03",
+            "eol-date": null,
             "releases.json": "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/3.1/releases.json"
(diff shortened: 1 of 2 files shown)


joeloff (approved):
I'm going to start looking at these this week so we can update all the branches with the new library and test data.

joeloff (approved):

dsplaisted:
We decided not to include this for servicing, as the workload manifests have been fixed not to hit this issue.  The fix is already checked in to 7.0.200.

PR #29540, opened 2022-12-14
Use File.OpenRead when loading tool configuration

### LINKED ISSUE #29535: "dotnet tool restore" failing when run concurrently in CI -- DotnetToolSettings.xml locked
### Describe the bug

Our client's CI process typically kicks off two builds in parallel - the branch without being merged in, and the branch after being merged in. (we're working to help modernize their process.)

But, this means that oftentimes in parallel, our build is running the following in a `bat` command in Jenkins:

`dotnet tool restore --tool-manifest code/src/REDACTED/.config/dotnet-tools.json`

Occasionally, the builds apparently run _too much in parallel_ 😄 and we see an error on both builds around a given tool (in this case, Swashbuckle, though we also use MinVer in our build and may see the error for that tool depending on the instance):

> Failed to retrieve tool configuration: The process cannot access the file 'C:\Users\REDACTED\.nuget\packages\swashbuckle.aspnetcore.cli\6.3.1\tools\net6.0\any\DotnetToolSettings.xml' because it is being used by another process

I've looked into the `--disable-parallel` flag but AFAICT that appears to be for parallel restores for a set of projects, not across invocations.

What is the recommendation for using dotnet tool restore in a CI setting, particularly where parallel builds are concerned?

(Side note: I'm happy to consider adding a PR to the code or the docs with a point in the right direction.)

### To Reproduce

* Create a tools file in a given project, e.g. `dotnet-tools.json`
* Create copies of the folder containing that tool setup
* In parallel, call `dotnet tool restore --tool-manifest path/folder1/to/dotnet-tools.json` and `dotnet tool restore --tool-manifest path/folder2/to/dotnet-tools.json`.

Expected behavior: the tools for both folders restore
Actual behavior: Error message about `DotnetToolSettings.xml` being locked.

## Backport PRs to create

* [x] 6.0.1xx #29579
* [x] 6.0.3xx #29583
* [x] 6.0.4xx: #29548
* [x] 7.0.1xx #29584
* [x] 7.0.2xx #29549

SeanKilleen:
Supports #29535. As noted, this explicitly opens the file with `Read`, rather than `ReadWrite`. This should prevent conflicts when the configuration files are accessed in parallel, such as when building multiple branches in a CI environment.

CC @baronfel in case you've any particular input in this case. (and thank you for the quick diagnosis in the original issue!)

### DIFF
--- a/src/Cli/dotnet/ToolPackage/ToolConfigurationDeserializer.cs
+++ b/src/Cli/dotnet/ToolPackage/ToolConfigurationDeserializer.cs
@@ -25,7 +25,7 @@ namespace Microsoft.DotNet.ToolPackage
 
             try
             {
-                using (var fs = new FileStream(pathToXml, FileMode.Open))
+                using (var fs = File.OpenRead(pathToXml))
                 {
                     var reader = XmlReader.Create(fs);
                     dotNetCliTool = (DotNetCliTool)serializer.Deserialize(reader);


baronfel:
This seems correct on its face, so I'm not sure what kind of testing we can do here. I think the other question would be if this is considered a bug fix. Currently this PR is targeting main, which is already building for .NET 8 previews. If we want this any sooner (and I suspect we do), then we'll need to re-open the PR against the appropriate `release/*` branches. @SeanKilleen - what version of the SDK was your client using? We are more likely to get fixes through the servicing process if we can point to specific customers/versions impacted by the issue.

SeanKilleen:
@baronfel thanks for the quick response again!

Client is on 6.x; big org and pinning to the LTS. Typically I'd say this could wait until 8 but given the implications for CI in larger environments it probably makes sense to backport to 7.x and 6.x.

Haven't done that in this context before but happy to see it through. Is it separate PRs against the release branches all linked to the same issue? It's a small enough change that it seems doable without risk.

baronfel:
Typically we start at the furthest back (so the 6.0.1xx branch I think?) and let the automatic codeflow we have set up flow the change to all supported 'downstream' branches. Let me check with the team for more details.

nagilson (approved):

baronfel:
Chatted with the team and we will take this to servicing. @SeanKilleen I see that you have PRs open for 6.0.4xx and 7.0.2xx, can you also open for 6.0.1xx, 6.0.3xx, and 7.0.1xx? These are all in support now.

Once the PRs are open, I'll update them all with what we call the 'servicing template' - a description format that includes a Description of the issue, estimated customer impact (what categories of customers hit the issue, if the issue was user reported, etc), if this is a Regression or not, risk, etc.

If you'd rather not cherry-pick and open the other PRs, I can of course do that for you - just wanted to give you the chance to get some sweet PR metrics :D

baronfel:
In the interim, I'll merge this for 8.0 as-is. Thank you for taking the initiative!

SeanKilleen:
@baronfel hah! I'm happy to do it; you've got enough on your plate. 

I'll open those PRs and then ping you on the issue so you can ship it.

PR #29559, opened 2022-12-14
fix locale-sensitive comparison of UnmatchedArgument error strings

### LINKED ISSUE #28975: .NET 7.0.100 `dotnet test` cannot parse MSBuild switches from .rsp or command line if .NET CLI language is not `en`
### Description
`dotnet test` command cannot parse parameters from provided .rsp file  if .NET CLI language is `de` or `ru` but works fine with `en` on .NET SDK 7.0.100. It leads to an error `Unrecognized command or argument ...`

This bug is not reproducible  on the .NET SDK 6.0.402.

### To Reproduce
1. Create any test project in my case it's xUnit project targeting `net7.0.` (please, check test project detail below)
2. Create any 1.rsp file with at least one switch, for example 
```
-p:abc=123
```
3. Set environment variable `DOTNET_CLI_UI_LANGUAGE` to value `de` (German). In PowerShell
``` PowerShell
PS > $env:DOTNET_CLI_UI_LANGUAGE = 'de'
```
4. Running the following command
``` PowerShell
PS > dotnet test .\xUnit.Tests.csproj `@1.rsp
```
gives 
```
Befehl oder Argument "-p:abc=123" nicht erkannt
Description:
  .NET-Testtreiber

Usage:
  dotnet test [<PROJECT | SOLUTION>] [options] [[--] <additional arguments>...]]

Arguments:
  <PROJECT | SOLUTION>  Das Projekt oder die Projektmappendatei, die verwendet werden soll. Wenn keine Datei angegeben ist,
                        durchsucht der Befehl das aktuelle Verzeichnis nach einer Datei.

Options:
...
```
`Befehl oder Argument "-p:abc=123" nicht erkannt` relates to `Unrecognized command or argument "-p:abc=123"` in English.

5. Set environment variable `DOTNET_CLI_UI_LANGUAGE` to value `en` (English):
``` PowerShell
PS > $env:DOTNET_CLI_UI_LANGUAGE = 'en'
```
6. Running the following command
``` PowerShell
PS > dotnet test .\xUnit.Tests.csproj `@1.rsp
```
gives
```
  Restored C:\Users\Maxim.Chechulnikov\sources\small-amount-of-dotnet-tests\mixed\xUnit.Tests\xUnit.Tests.csproj (in 274 ms).
  xUnit.Tests -> C:\Users\Maxim.Chechulnikov\sources\small-amount-of-dotnet-tests\mixed\xUnit.Tests\bin\Debug\net7.0\xUnit.Tests.dl
  l
Test run for C:\Users\Maxim.Chechulnikov\sources\small-amount-of-dotnet-tests\mixed\xUnit.Tests\bin\Debug\net7.0\xUnit.Tests.dll (.NETCoreApp,Version=v7.0)
Microsoft (R) Test Execution Command Line Tool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:   100, Skipped:     0, Total:   100, Duration: 13 ms - xUnit.Tests.dll (net7.0)
```

#### An example of simplest test project 
<details>
<summary>xUnit.Tests.csproj content</summary>
Content

``` xml

    <PropertyGroup>
        <Nullable>enable</Nullable>

        <IsPackable>false</IsPackable>
    </PropertyGroup>

    <ItemGroup>
        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
        <PackageReference Include="xunit" Version="2.4.1" />
        <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
            <PrivateAssets>all</PrivateAssets>
        </PackageReference>
        <PackageReference Include="coverlet.collector" Version="3.1.2">
            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
            <PrivateAssets>all</PrivateAssets>
        </PackageReference>
    </ItemGroup>

</Project>
```
</details>

<details>
<summary>UnitTests0.cs content</summary>
Content

``` C#
namespace xUnit.Tests;

public class UnitTest0
{
    [Fact]
    public void Test0()
    {
        Assert.True(true);
    }

    ...  // I have 100 same tests
}
```
</details>

<details>
<summary>Usings.cs content</summary>
Content

``` c#
global using Xunit;
```
</details>

<details>
<summary>1.rsp content</summary>
Content

```
-p:abc=123
```
</details>

### Further technical details
<details>
<summary>`dotnet --info` output:</summary>
```
.NET SDK:
 Version:   7.0.100
 Commit:    e12b7af219

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.100\

Host:
  Version:      7.0.0
  Architecture: x64
  Commit:       d099f075e4

.NET SDKs installed:
  3.1.424 [C:\Program Files\dotnet\sdk]
  6.0.401 [C:\Program Files\dotnet\sdk]
  6.0.402 [C:\Program Files\dotnet\sdk]
  7.0.100 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0-rc.2.22472.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.0 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:

Environment variables:
  Not set

global.json file:
  Not found

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
```
</details>

### LINKED ISSUE #29543: dotnet test does no longer accept a project path after updating to NET 7.0.101
### Describe the bug
After updating NET 7.0.100 to NET 7.0.101 `dotnet test` does no longer accept a project path as an argument.

### To Reproduce

Sample test project:
```xml

  <PropertyGroup>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.4.0" />
    <PackageReference Include="xunit" Version="2.4.2" />
    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>

</Project>
```

```c#
using Xunit;

namespace ClassLibrary1
{
    public class Class1
    {
        [Fact]
        public void Test()
        {

        }
    }
}
```

With NET 7.0.101 this does not work
```
C:\>dotnet --version
7.0.101
C:\>dotnet test C:\Users\Administrator\Desktop\ClassLibrary1.csproj
Befehl oder Argument "C:\Users\Administrator\Desktop\ClassLibrary1.csproj" nicht erkannt
```

With NET 7.0.100 this worked

```
C:\>dotnet --version
7.0.100
C:\>dotnet test C:\Users\Administrator\Desktop\ClassLibrary1.csproj
  All projects are up-to-date for restore.
  ClassLibrary1 -> C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll
Test run for C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll (.NETCoreApp,Version=v7.0)
Microsoft (R) Test Execution Command Line Tool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: < 1 ms - ClassLibrary1.dll (net7.0)
```

This also affects the built in dotnet command in Azure Dev Ops.
image

### Workarounds

- Set the environment variable `DOTNET_CLI_UI_LANGUAGE=en-US`.

- Running `donet test` directly in the project directory does work:

```
C:\Users\Administrator\Desktop>dotnet test
  Wiederherzustellende Projekte werden ermittelt...
  Alle Projekte sind für die Wiederherstellung auf dem neuesten Stand.
  ClassLibrary1 -> C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll
Testlauf für "C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll" (.NETCoreApp,Version=v7.0)
Microsoft (R) Testausführungs-Befehlszeilentool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation. Alle Rechte vorbehalten.

Die Testausführung wird gestartet, bitte warten...
Insgesamt 1 Testdateien stimmten mit dem angegebenen Muster überein.

Bestanden!   : Fehler:     0, erfolgreich:     1, übersprungen:     0, gesamt:     1, Dauer: < 1 ms - ClassLibrary1.dll (net7.0)
```

- Running `donet test` referencing a DLL or exe does work:

```
C:\Users\Administrator\Desktop>dotnet test C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll
  Wiederherzustellende Projekte werden ermittelt...
  Alle Projekte sind für die Wiederherstellung auf dem neuesten Stand.
  ClassLibrary1 -> C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll
Testlauf für "C:\Users\Administrator\Desktop\bin\Debug\net7.0\ClassLibrary1.dll" (.NETCoreApp,Version=v7.0)
Microsoft (R) Testausführungs-Befehlszeilentool Version 17.4.0 (x64)
Copyright (c) Microsoft Corporation. Alle Rechte vorbehalten.

Die Testausführung wird gestartet, bitte warten...
Insgesamt 1 Testdateien stimmten mit dem angegebenen Muster überein.

Bestanden!   : Fehler:     0, erfolgreich:     1, übersprungen:     0, gesamt:     1, Dauer: < 1 ms - ClassLibrary1.dll (net7.0)
```

### Further technical details
```
C:\>dotnet --info
.NET SDK:
 Version:   7.0.101
 Commit:    bb24aafa11

Laufzeitumgebung:
 OS Name:     Windows
 OS Version:  10.0.20348
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.101\

Host:
  Version:      7.0.1
  Architecture: x64
  Commit:       97203d38ba

.NET SDKs installed:
  7.0.101 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 7.0.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 7.0.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 7.0.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  None

Environment variables:
  Not set

global.json file:
  Not found
```

baronfel:
## Description

Closes #29543 
Closes #28975 

by fixing a helper method comparison that was intended to filter out unmatched token errors. Since errors in System.CommandLine parsing are strings, discovering the source of a particular string is hard. This method attempted to see if a 'masked' error string was entirely contained in the error string. This works for languages and grammars that put the argument name in single quotes at the end of the string, but for locales like `de-DE` both of these invariants were broken.

The new method breaks the format string at the  .NET format placeholders, then ensures that each of those literal portions exists in order in the final localized error message string.

This issue **was customer reported**.

## Customer Impact
This potentially impacts build, pack, test, restore, list references, and other CLI Commands that use this same call to check for unparsed arguments. Users that used these commands in locales with non-English arrangements of locale strings would not be able to execute those commands. One **workaround** would be to force the use of English (or similarly-structured) locales.

## Regression
**Yes**, but hidden under a couple other issues that we fixed in 7.0.101.

## Risk
**Low** - the new checking logic seems to be stable under different locales, and has tests covering the locales that users have reported failing.

## Testing
I locally tested this by using the DOTNET_CLI_UI_LANGUAGE environment variable to setup the triggering situation.
I added automated testing of different locales to validate the change.

## Follow up

This kind of string manipulation is very error-prone, so I've reached out to the System.CommandLine API design group to consider a more structured approach.


src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -34,18 +34,29 @@ public static void ShowHelp(this ParseResult parseResult)
             Parser.Instance.Parse(tokenList).Invoke();
         }
 
+        private static string StripPlaceholdersFromString(string input) {
```
Evangelink, line 37: Nitpick

```suggestion
        private static string StripPlaceholdersFromString(string input)
        {
```

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -34,18 +34,29 @@ public static void ShowHelp(this ParseResult parseResult)
...
+            return System.Text.RegularExpressions.Regex.Replace(input, @"['""](.*)['""]", string.Empty);
+        }
+
         public static void ShowHelpOrErrorIfAppropriate(this ParseResult parseResult)
         {
             if (parseResult.Errors.Any())
             {
-                var unrecognizedTokenErrors = parseResult.Errors.Where(error =>
-                    error.Message.Contains(Parser.Instance.Configuration.LocalizationResources.UnrecognizedCommandOrArgument(string.Empty).Replace("'", string.Empty)));
+                var unrecognizedTokenErrors = parseResult.Errors.Where(error => {
```
Evangelink, line 49: Nitpick:

```suggestion
                var unrecognizedTokenErrors = parseResult.Errors.Where(error => 
                {
```

Evangelink (commented):
Shall we add the test with `dotnet test` and non-US locale as part of this PR or a follow-up one?

baronfel:
Just added a test that
* sets a locale
* parses a `dotnet test` command with an unmatched argument
* asserts that parse errors exist, and
* asserts that our throw helper doesn't throw (because the errors are UnmatchedToken errors

This test works with en-US and de-DE cultures, which seem to cover a number of the differences in culture strings here.

Evangelink (approved):

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -34,18 +34,31 @@ public static void ShowHelp(this ParseResult parseResult)
             Parser.Instance.Parse(tokenList).Invoke();
         }
 
+        private static string StripPlaceholdersFromString(string input)
+        {
+            // Find and remove all format placeholder strings (and filled runtime strings!) of the form ['"].*['"]
+            // from an input string.
+            // This lets us do fair comparisons against strings that are translated, since the placeholders
+            // may be in various locations, and may have different quote characters in different languages.
+            return System.Text.RegularExpressions.Regex.Replace(input, @"['""](.*)['""]", string.Empty);
```
danmoseley, line 43: This is greedy .. it will change `Foo "bar" baz "biz" boo`
To `Foo boo` won't it?

You can change to `(.*?)` or `([^'""])*` if you don't want that.

baronfel: Thanks! I'll do so. I want to remove each instance of the pattern.

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -34,18 +34,31 @@ public static void ShowHelp(this ParseResult parseResult)
             Parser.Instance.Parse(tokenList).Invoke();
         }
 
+        private static string StripPlaceholdersFromString(string input)
+        {
+            // Find and remove all format placeholder strings (and filled runtime strings!) of the form ['"].*['"]
+            // from an input string.
+            // This lets us do fair comparisons against strings that are translated, since the placeholders
+            // may be in various locations, and may have different quote characters in different languages.
+            return System.Text.RegularExpressions.Regex.Replace(input, @"['""](.*?)['""]", string.Empty);
```
dsplaisted, line 43: I'm bad at regex, but doesn't this still assume that the quoting character is either a single or a double quote?

Would it be better to parse the format string into its components, and check that the error message contains the non-substituted components of the format string in the right order?

IE the string `Unrecognized command or argument '{0}'` has these components:

- `Unrecognized command or argument '` - literal
- '{0}' - substitution
- `'` - literal

So we would check that the error string started with the first component, and that the third component was found somewhere after the end of the first one.

That way I don't think we'd be dependent on localization or on the substituted arguments having quotes around them at all.

src/Tests/dotnet.Tests/ParserTests/TestParserTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            System.Globalization.CultureInfo.CurrentCulture = this.startingCulture;
+        }
+
+        [InlineData("en-US")]
+        [InlineData("de-DE")]
+        [Theory]
+        public void TestParserCanGetArgumentFromDoubleDash(string culture)
+        {
+            System.Globalization.CultureInfo.CurrentCulture = new System.Globalization.CultureInfo(culture);
```
dsplaisted, line 35: Tests are run in parallel, and this affects global state, so this may cause flakiness in our tests.

MarcoRossignoli:
@baronfel  any news on this one?

src/Cli/dotnet/ParseResultExtensions.cs
baronfel: Yeah, that's a solid plan. I'll work on that today.

src/Tests/dotnet.Tests/ParserTests/TestParserTests.cs
baronfel: Yeah, I know - I couldn't find a way around this aside from reworking a lot of the internals of the CLI parser to use some kind of 'locale provider' abstraction instead. Definitely open to thoughts here.

MarcoRossignoli: You can go out of proc like runtime tests https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.RemoteExecutor @ViktorHofer is there some guide on how to integrate?

ViktorHofer: Unfortunately we don't have a README but you should just need a PackageReference pointing to the package and then invoke the API.

baronfel: I suppose an alternative here would be to use the existing `DotnetTestCommand` wrappers that actually invoke the CLI on an actual project - but that is _a lot_ of additional work for what should just be a test of the parser.

baronfel: I ended up using out-of-proc test executions against a known, simple test project to validate this.

src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
```diff
@@ -4,11 +4,11 @@
 using System;
 using System.Globalization;
 
-namespace Microsoft.DotNet.Cli
+namespace Microsoft.DotNet.Cli.Utils
```
baronfel, line 7: I moved this file to Utils because a) much of the shared utility logic for the CLI was already in this project and b) this project already had internals visible to the test framework, where I wanted to add a `WithCulture` extension method.

baronfel:
I think we're in a good space now with the algorithm and the tests for this check.  I'll want to get another check from @dsplaisted to make sure he's comfortable with the implementation, but then we're ready to go.

A note about the timing of this fix: we are already beyond the merge window for the January servicing release of .NET 7 (which would be 7.0.102) due to the holiday season. As a result, this fix (when approved and merged) will only be available in the Februrary servicing release at earliest (so 7.0.103). This feels really bad, and isn't a great response time given the very fast time to reporting that users had, but it's the best I can do.

In the meantime, the best thing I can suggest for users that are hitting this error is to force the CLI to the English local for `dotnet test`. The easiest way to do that is to set the `DOTNET_CLI_UI_LANGUAGE` environment variable in your shell to `en-US` for the duration of the test command. e.g. in Bash you'd run `DOTNET_CLI_UI_LANGUAGE='en-us' dotnet test .....`. Full documentation for this environment variable can be found [here](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_cli_ui_language).

MarcoRossignoli:
@baronfel thanks for the fix!

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -34,21 +35,51 @@ public static void ShowHelp(this ParseResult parseResult)
...
-                                             parseResult.Errors.Select(e => e.Message)), 
+                                             parseResult.Errors.Select(e => e.Message)),
                         parseResult: parseResult);
                 }
             }
+
+            ///<summary>Splits a .NET format string by the format placeholders (the {N} parts) to get an array of the literal parts, to be used in message-checking</summary> 
+            static string[] DistinctFormatStringParts(string formatString)
+            {
+                return Regex.Split(formatString, @"{[0-9]+}"); // match the literal '{', followed by any of 0-9 one or more times, followed by the literal '}'
```
dsplaisted, line 62: Are braces not special characters in .NET regexes?  Do they need to be escaped?

src/Cli/dotnet/ParseResultExtensions.cs
```diff
@@ -34,21 +35,51 @@ public static void ShowHelp(this ParseResult parseResult)
...
+
+
+            /// <summary>given a string and a series of parts, ensures that all parts are present in the string in sequential order</summary>
+            static bool ErrorContainsAllParts(ReadOnlySpan<char> error, string[] parts)
+            {
+                foreach(var part in parts) {
+                    var foundIndex = error.IndexOf(part);
+                    if (foundIndex != -1)
+                    {
+                        error = error.Slice(foundIndex);
```
dsplaisted, line 73: Should this be the following?

```suggestion
                        error = error.Slice(foundIndex + part.Length);
```

dsplaisted (approved):

src/Cli/dotnet/ParseResultExtensions.cs
baronfel: Correct, they are not special-cased:

image

You can see here that the `Value` of the Capture is the literal brace-wrapped number.

src/Cli/dotnet/ParseResultExtensions.cs
baronfel: good eye!

marcpopMSFT:
@baronfel this was approved for February. Please ensure the checks get resolved and this gets merged by eow.

baronfel:
Both of these check failures are due to the macos helix image expiration, which I think has already been handled in an earlier PR. Trying to kick these off again to see.

PR #29566, opened 2022-12-15
Update the hardcoded to the current framework

v-wuzhai:


marcpopMSFT (approved):

marcpopMSFT:
I think now that I moved the cppcli tests to use the packagereference support, they should be ok when we move to 9 but it'll be something we watch out for in 9 months.

PR #29599, opened 2022-12-16
Add support for new output path format

dsplaisted:
Add support for simplified output paths and `RootOutputPath` property.

Based on this design spec: https://github.com/dotnet/designs/pull/281.  The design doc needs to be updated.  The main changes from the current version:

- Switching away from the "artifacts" name, and putting the new output paths under `bin` instead.  This is discussed in [this comment](https://github.com/dotnet/designs/pull/281)
- The new output path format is not enabled by default on .NET 8 and higher.  You have to opt in to it.  This is due to some of the issues in the previously referenced comment, as well as concerns about builds (especially Maui builds) running into Windows path length issues.

I needed a name for a property that would enable the new output paths.  I went with `UseStandardOutputPaths`, which was the best I could think of.  Better name suggestions are welcome.

The property that puts all output for a solution under a single folder is `RootOutputPath`, which I think is a reasonably good name.  Still open to suggestions though.


nagilson (commented):
Thank you for wrangling up all of those tests... *wow* that was a lot of hard-coded paths. 

It looks like there are still a lot of test failures, including a few of the newer tests, so I won't look at the MSBuild code and try it out until those are fixed. Also hoping to get a little more engagement on the design spec

`UseStandardOutputPaths` seems like it has a similar-ish problem in what is "standard" might change... (I am thinking of netstandard rids.)
Some brainstormed ideas:
`UseSimplifiedOutputPaths`
`UseExperimentalOutputPaths`

I like `RootOutputPath`. Maybe also: `FlatOutputPath`?

baronfel (approved):
Very excited to see this land for previews and user feedback!

Re: the name of the opt-in property, I agree with @nagilson that `Standard`'s meaning can change over time, but I don't have a compelling alternative. Maybe `UseUnifiedOutputPaths`? This is something we can potentially update across previews.

Nirmal4G:
Why not `BuildDir`?

My proposal: https://github.com/dotnet/designs/pull/281#issuecomment-1381205054

I personally think we should start from existing paths and modify it to cover all the goals presented in the new proposal. I'm not good at explaining things over words, so I could create a PR on both MSBuild repo and here to explain the changes better!

Nirmal4G (changes requested):
Why not `BuildDir`?

My proposals:

 1. [Introduce `BuildDir` as a top-level directory for all things build](https://github.com/dotnet/designs/pull/281#issuecomment-1381205054)
2. [Fix output path customization limitations](https://github.com/dotnet/designs/pull/281#issuecomment-1409717669)

I personally think we should start from existing paths and modify it to cover all the goals presented in the new proposal. I'm not good at explaining things over words, so I could create a PR on both MSBuild repo and here to explain the changes better!

joeloff:
For the `Standard` naming issue, I'll toss another suggestion out, `Legacy`

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
```diff
@@ -11,29 +11,60 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+
+    <BaseOutputPath>$(BaseStandardOutputPath)build\</BaseOutputPath>
+
+    <!-- If RootOutputPath is set, then projects share the BaseStandardOutputPath, and the project name should be appended to the output path -->
+    <BaseOutputPath Condition="'$(RootOutputPath)' != ''">$(BaseOutputPath)$(MSBuildProjectName)\</BaseOutputPath>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(UseStandardOutputPaths)' != 'true'">
     <BaseOutputPath Condition="'$(BaseOutputPath)' == ''">bin\</BaseOutputPath>
     <BaseOutputPath Condition="!HasTrailingSlash('$(BaseOutputPath)')">$(BaseOutputPath)\</BaseOutputPath>
```
joeloff, line 59: I know this isn't part of your change, but any reason we use the condition and manually add a slash when we can use `EnsureTrailingSlash`

richlander:
Can you add an example to your initial comment what default and opt-in will look like? That would help a lot.

Nirmal4G:
Is the comment directed to me?

dsplaisted:
@joeloff @nagilson @baronfel @rainersigwald I've updated the implementation here to (I believe) match the latest version of the [design proposal](https://github.com/dotnet/designs/pull/281).

I still need to update tests and probably add some more test coverage, but the implementation is ready to review, I believe.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
```diff
@@ -11,29 +11,109 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <!-- ArtifactsPivots instead? -->
+    <ArtifactsPivots>$(Configuration.ToLowerInvariant())</ArtifactsPivots>
+
+    <!-- Include the TargetFramework in the pivots if the project is multi-targeted (ie TargetFrameworks) is defined -->
+    <ArtifactsPivots Condition="'$(TargetFrameworks)' != ''"
+                     >$(ArtifactsPivots)_$(TargetFramework.ToLowerInvariant())</ArtifactsPivots>
+
+    <!-- This targets file is evaluated before RuntimeIdentifierInference.targets, so this will only include the
+         RuntimeIdentifier in the path if it was explicitly specified, not if it was inferred.  This is the
+         behavior we want.
```
nagilson, line 74: Thank you for this comment 🔥

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
```diff
@@ -11,29 +11,109 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+
+         The BlazorWebAssembly .props file sets the RuntimeIdentifier to blazor-wasm, so treat that as a special case.
```
nagilson, line 76:
```suggestion
         The BlazorWebAssembly .props file sets the RuntimeIdentifier to browser-wasm, so treat that as a special case.
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
```diff
@@ -11,29 +11,109 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    duplication and more automatic defaults, we always set a default OutputPath.
    -->
-  <PropertyGroup>
-    <Configuration Condition="'$(Configuration)'==''">Debug</Configuration>
-    <Platform Condition="'$(Platform)'==''">AnyCPU</Platform>
-    <PlatformName Condition="'$(PlatformName)' == ''">$(Platform)</PlatformName>
 
+  <!-- Projects which don't use Microsoft.NET.Sdk will typically define the OutputPath directly (usually in a
+       Configuration-specific PropertyGroup), so in that case we won't append to it by default. -->
+  <PropertyGroup Condition="'$(UsingNETSdkDefaults)' == 'true'">
```
nagilson, line 24: I believe this is a breaking change to existing paths? Do we want this for the existing behavior as well?

src/Tests/Microsoft.NET.Build.Tests/ArtifactsOutputPathTests.cs
```diff
@@ -0,0 +1,216 @@
...
+                    """);
+            }
+
+            return (testProjects, testAsset);
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ItUsesArtifactsOutputPathForBuild(bool useDirectoryBuildProps)
```
nagilson, line 82: Do we have a test for a custom name of  `<ArtifactsPath>foo</ArtifactsPath>`?

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -32,6 +32,59 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <!-- We need to put the UseArtifactsOutput logic after the import of Directory.Build.props, but before the MSBuild Project Extensions .props import.
+       However, both of these things happen in Microsoft.Common.props with no opportunity to insert logic in between them.
+
+       So what we do here is duplicate the Directory.Build.props import logic from Microsoft.Common.props, and then set ImportDirectoryBuildProps to
+       false so that it doesn't get imported twice.
+
+       Alternatively, we could add a hook in MSBuild to define a file to import at the right location, to avoid duplicating the logic here.
+       -->
+  <PropertyGroup>
+    <ImportDirectoryBuildProps Condition="'$(ImportDirectoryBuildProps)' == ''">true</ImportDirectoryBuildProps>
```
nagilson, line 46: When we communicate this feature to users we should be very clear they shouldn't add this to their project files.

src/Tests/Microsoft.NET.TestFramework/Commands/MSBuildCommand.cs
```diff
@@ -72,8 +75,13 @@ internal static string FindProjectFile(ref string projectRootPath, string relati
             return buildProjectFiles[0];
         }
 
-        public virtual DirectoryInfo GetOutputDirectory(string targetFramework, string configuration = "Debug", string runtimeIdentifier = "")
+        public virtual DirectoryInfo GetOutputDirectory(string targetFramework = null, string configuration = "Debug", string runtimeIdentifier = null)
```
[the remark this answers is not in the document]
dsplaisted: It makes for more concise syntax later on:

```c#
targetFramework = targetFramework ?? TargetFramework ?? string.Empty;
```

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -32,6 +32,59 @@ Copyright (c) .NET Foundation. All rights reserved.
     <UsingNETSdkDefaults>true</UsingNETSdkDefaults>
   </PropertyGroup>
```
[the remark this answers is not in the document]
dsplaisted: Thanks, yes, it looks like I missed adding that.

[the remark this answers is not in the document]
dsplaisted: It's not really possible to do this, because where would we store the cache of the old value?  It would normally go in the intermediate directory, but that may also have changed.

I don't think it's a common scenario to choose a different artifacts / output path within the project folder.  It's more likely that if you are overriding this folder, you will be putting it somewhere further up in the directory tree.  So I don't think it will be too common to hit this.

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -32,6 +32,59 @@ Copyright (c) .NET Foundation. All rights reserved.
...
 
+
+  
+  <!-- We need to put the UseArtifactsOutput logic after the import of Directory.Build.props, but before the MSBuild Project Extensions .props import.
+       However, both of these things happen in Microsoft.Common.props with no opportunity to insert logic in between them.
+
+       So what we do here is duplicate the Directory.Build.props import logic from Microsoft.Common.props, and then set ImportDirectoryBuildProps to
+       false so that it doesn't get imported twice.
+
+       Alternatively, we could add a hook in MSBuild to define a file to import at the right location, to avoid duplicating the logic here.
```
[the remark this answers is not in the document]
dsplaisted: I plan to add the hook to MSBuild.  The code change itself probably won't be hard, the more complicated part will be syncing the changes between the two different repos.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
dsplaisted: This is behavior which already existed, it's just been moved around to a different .targets file.  I believe there shouldn't be any breaking change with this.

src/Tests/Microsoft.NET.Build.Tests/ArtifactsOutputPathTests.cs
dsplaisted: Not yet.

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
dsplaisted: This is implementation that users generally won't see.  The ImportDirectoryBuildProps property is something that MSBuild has supported since (I believe) the support for Directory.Build.props was added.

src/Tests/Microsoft.NET.TestFramework/OutputPathCalculator.cs
```diff
@@ -0,0 +1,252 @@
...
+            //}
+            //else
+            //{
+            //    return false;
+            //}
+
+            return false;
+        }
+
+        public string GetOutputDirectory(string targetFramework = null, string configuration = "Debug", string runtimeIdentifier = "")
```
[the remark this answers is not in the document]
dsplaisted: When using artifacts output format, the logic is very similar and can be merged.  But otherwise the logic is different, so I thought it was simplest to leave them as separate functions.

src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/WasmBuildIntegrationTest.cs
```diff
@@ -93,7 +93,7 @@ public void Build_Works_WithLibraryUsingHintPath()
                     reference.Name = "Reference";
                     reference.Add(new XElement(
                         "HintPath",
-                        Path.Combine("..", "razorclasslibrary", "bin", "Debug", DefaultTfm, "RazorClassLibrary.dll")));
```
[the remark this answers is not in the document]
dsplaisted: Since it's a relative path, it seemed like that would be more trouble than it's worth.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
```diff
@@ -11,29 +11,60 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <!-- Set BaseOutputPath for standard output format if it wasn't set in Sdk.props -->
+  <PropertyGroup Condition="'$(UseStandardOutputPaths)' == 'true' And '$(BaseOutputPath)' == ''">
+    <BaseStandardOutputPath Condition="'$(BaseStandardOutputPath)' == ''">bin\</BaseStandardOutputPath>
+    <BaseStandardOutputPath>$([MSBuild]::EnsureTrailingSlash($(BaseStandardOutputPath)))</BaseStandardOutputPath>
+
+    <BaseOutputPath>$(BaseStandardOutputPath)build\</BaseOutputPath>
+
+    <!-- If RootOutputPath is set, then projects share the BaseStandardOutputPath, and the project name should be appended to the output path -->
+    <BaseOutputPath Condition="'$(RootOutputPath)' != ''">$(BaseOutputPath)$(MSBuildProjectName)\</BaseOutputPath>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
dsplaisted: No, this is for the case where you set properties (which have changed names since this) in your project file instead of in a `Directory.Build.props` file.  So when `Sdk.props` is evaluated, those properties aren't set yet.

src/Tests/dotnet-sln.Tests/GivenDotnetSlnRemove.cs
```diff
@@ -619,13 +619,13 @@ public void WhenReferenceIsRemovedSlnBuilds()
...
-            releaseDirectory.Count().Should().Be(1, $"App {reasonString}");
-            Directory.EnumerateFiles(releaseDirectory.Single(), "App.dll", SearchOption.AllDirectories)
-                .Count().Should().Be(1, $"App {reasonString}");
+            var outputCalculator = OutputPathCalculator.FromTestAsset(Path.Combine(projectDirectory, "App"));
+
+            new DirectoryInfo(outputCalculator.GetOutputDirectory(configuration: "Debug")).Should().NotExist(reasonString);
+
+            var outputDirectory = new DirectoryInfo(outputCalculator.GetOutputDirectory(configuration: "Release"));
+            outputDirectory.Should().Exist();
+            outputDirectory.Should().HaveFile("App.dll");
```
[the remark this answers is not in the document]
dsplaisted: It basically tests the same thing because it verifies that the Debug output folder doesn't exist.

All those output files are expected in the output path I believe, and would have been there before any of my changes too.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
dsplaisted: I don't know, my guess is that when this was originally implemented that either `EnsureTrailingSlash` didn't exist or whoever implemented this didn't know about it.

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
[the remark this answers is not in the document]
dsplaisted: There's nothing that definitively defines the root of a repository.  When we add new functionality we can use a heuristic (such as using the Directory.Build.props folder, looking for a `.git` folder, or for a `.sln` file), but we can't change existing functionality like the Directory.Build.props search without breaking people.

We could add a warning or something if lots of people were having problems with this, but I don't think we've heard anything to indicate that's happening.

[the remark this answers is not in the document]
dsplaisted: While testing this out for a demo, I did run into issues where after switching to artifacts output format, the intermediate folders from the old `obj` directory were picked up, causing errors.  I had thought they would have been excluded.  I've updated the logic so that when using artifacts output folder the `bin` and `obj` folders in the project are always ignored, and added a test case for switching back and forth between formats.

src/Tests/Microsoft.NET.Build.Tests/ArtifactsOutputPathTests.cs
dsplaisted: OK, it should be added now.

dsplaisted:
> Why not `BuildDir`?
> 
> My proposals:
> 
> 1. [Introduce `BuildDir` as a top-level directory for all things build](https://github.com/dotnet/designs/pull/281#issuecomment-1381205054)
> 2. [Fix output path customization limitations](https://github.com/dotnet/designs/pull/281#issuecomment-1409717669)
> 
> I personally think we should start from existing paths and modify it to cover all the goals presented in the new proposal. I'm not good at explaining things over words, so I could create a PR on both MSBuild repo and here to explain the changes better!

@Nirmal4G I just noticed this comment today.  It can be hard to clearly explain thoughts on technical things like this.  Unfortunately, sending a PR without the discussion often doesn't help much, as the code changes can be complicated and it isn't clear without explanation why the PR is doing what it's doing.

I've asked some questions over [on the designs proposal](https://github.com/dotnet/designs/pull/281#issuecomment-1428943439) to try to better understand what you are looking for and why.

Thanks!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
nagilson: You're right, there is no breaking change. The first concern would be rid inference but this is moving the change to earlier in evaluation (defaultoutpathtargets) instead of later (after ridinference.targets). Thanks!

src/Cli/dotnet/CommonOptions.cs
```diff
@@ -51,6 +51,15 @@ internal static class CommonOptions
             }.ForwardAsSingle(o => $"-property:TargetFramework={o}")
             .AddCompletions(Complete.TargetFrameworksFromProjectFile);
 
+        public static Option<string> ArtifactsPathOption =
+            new ForwardedOption<string>(
+                //  --artifacts-path is pretty verbose, should we use --artifacts instead (or possibly support both)?
```
nagilson, line 56: I like supporting both. We should ideally mention this in the doc's options for each command, such as in https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-build

nagilson (approved):
LGTM. Thanks for responding to my feedback and sorry it took a bit for me to get back to it. Excited for this to go in! 🎉

src/Cli/dotnet/CommonOptions.cs
joeloff: Wouldn't artifacts imply the produced output instead of their location?

joeloff:
Might be totally unrelated, but does this potentially impact things like the build configuration dialogs in Visual Studio? What happens when a user launches VS directly/from a console that don't have the properties set for using a single output path or is the assumption that they're either specified as environment variables or defined in a directory.build.props?

src/Cli/dotnet/CommonOptions.cs
nagilson: That's true. Nvm on the --artifacts path, still think we should include this in the docs tho!

marcpopMSFT:
@dsplaisted fyi, we had a bad branch creation for preview 3 so we deleted it and repushed with the right commit. I had to close and reopen your PR during that activity (and fixed a merge conflict)

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -32,6 +32,60 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <PropertyGroup Condition="'$(ArtifactsPath)' != ''">
+    <UseArtifactsOutput Condition="'$(UseArtifactsOutput)' == ''">true</UseArtifactsOutput>
+    <IncludeProjectNameInArtifactsPaths Condition="'$(IncludeProjectNameInArtifactsPaths)' == ''">true</IncludeProjectNameInArtifactsPaths>
+  </PropertyGroup>
+
+  <!-- Set up base output folders if UseArtifactsOutput is set -->
+  <PropertyGroup Condition="'$(UseArtifactsOutput)' == 'true' And '$(ArtifactsPath)' == '' And '$(_DirectoryBuildPropsBasePath)' != ''">
+    <!-- Default ArtifactsPath to be in the directory where Directory.Build.props is found
+         Note that we do not append a backslash to the ArtifactsPath as we do with most paths, because it may be a global property passed in on the command-line which we can't easily change -->
+    <ArtifactsPath>$(_DirectoryBuildPropsBasePath)\.artifacts</ArtifactsPath>
```
rainersigwald, line 70: If you stick with `.artifacts` I think you should add it to the `.gitignore` template

https://github.com/dotnet/sdk/blob/09d8fec0cd3cfd622fa8e53dba5509d6decd2deb/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Gitignore/.gitignore#L19

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
```diff
@@ -11,29 +11,109 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <UseArtifactsOutput>true</UseArtifactsOutput>
+  </PropertyGroup>
+  -->
+
+  <!-- Handle ArtifactsPath if it wasn't set in Sdk.props -->
+  <PropertyGroup Condition="'$(ArtifactsPath)' != ''">
+    <UseArtifactsOutput Condition="'$(UseArtifactsOutput)' == ''">true</UseArtifactsOutput>
+  </PropertyGroup>
+
+  <!-- Repeat ArtifactsPath logic from Sdk.props here, in case UseArtifactsOutput was set in the project file -->
```
rainersigwald, line 45: Could we error in that case instead? I don't like the idea of accidentally diverging the implementation over time.

rainersigwald (commented):
I didn't look at all the tests, but have a few small comments on the MSBuild logic.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -708,4 +708,10 @@ The default is 'true' if a runtime identifier is specified.</value>
   <data name="CorruptSolutionProjectFolderStructure" xml:space="preserve">
     <value>The solution file '{0}' is missing EndProject tags or has invalid child-parent project folder mappings around project GUID: '{1}'. Manually repair the solution or try to open and save it in Visual Studio."</value>
   </data>
+  <data name="ArtifactsPathOptionDescription" xml:space="preserve">
```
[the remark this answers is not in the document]
dsplaisted: It does include intermediate files.  I didn't include that in the description because I don't expect people to be looking for the intermediate path in order to do something with the files there.

Note that if you set the `ArtifactsPath` property in the project file instead of in Directory.Build.props or on the command line, that it doesn't affect the intermediate output path.

dsplaisted:
> Might be totally unrelated, but does this potentially impact things like the build configuration dialogs in Visual Studio? What happens when a user launches VS directly/from a console that don't have the properties set for using a single output path or is the assumption that they're either specified as environment variables or defined in a directory.build.props?

@joeloff  I think the main scenario is to put it in Directory.Build.props or to specify it on the command-line.  If `ArtifactsPath` is specified in Directory.Build.props then Visual Studio should respect it.  Do you mean the Configuration Manager dialog?  It doesn't look like that shows output paths so I don't think it would be affected at all.

If you only specify the artifacts path on the command-line, then Visual Studio would build using the default (old) paths.  The different output paths shouldn't interfere with each other though (the test that verifies that you can switch between output path formats helps verify that).  Also, I think specifying the artifacts path on the command line may be more of a CI scenario so it isn't as likely to happen together with Visual Studio.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets
dsplaisted: So `ArtifactsPath` just wouldn't be allowed in the project file at all?

We go through some hoops to make `BaseIntermediateOutputPath` work whether it's set in the project file or beforehand.  If it's set in the project file then it doesn't affect `MSBuildProjectExtensionsPath`.  This seemed analogous.

I just had a crazy idea: Maybe we could de-duplicate some of the logic by having a .props file that's conditionally imported in two different places depending on where `ArtifactsPath` is defined?

dsplaisted:
Followup items:

- [ ] Port changes to `main` branch
- [ ] Add `.artifacts` to Visual Studio default .gitignore file.
- [ ] De-duplicate default output path logic depending on whether `ArtifactsPath` is defined in project file or beforehand
- [ ] Add hook to MSBuild to inject targets after `Directory.Build.props` import, and use that in SDK instead of duplicate import logic.

PR #29636, opened 2022-12-21
[release/7.0.1xx] Inverted language version conditions in console / classlib templates to support new language features for newer language versions

YuliiaKovalova:
## Problem
fixes https://github.com/dotnet/templating/issues/5668

## Description
In the fix the condition on language version was reverted, so the default behavior is new syntax and that the C#11 and further language versions will use new syntax.

## Customer Impact
After the fix, the template that utilizes the latest language features (top-level statements, file-scoped namespaces etc) will be used.

## Testing
Automated

## Risk
Very low
Those conditions are already used in .NET 8 templates and passing the tests

## Checks:
 Added unit tests - existing tests cover the change
 Added #nullable enable to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style) - NA


JanKrivanek (approved):
Looks good!

PR #29644, opened 2022-12-22
Update the macOs version

v-wuzhai:


marek-safar (approved):

am11:
It seems like some OSX test failures are related to tests using `osx.{version}-{architecture}` RIDs, where major version is 10. We can update them to use version-less RID (where applicable) and update the rest to use 12.x.

(FWIW, tests in dontet/runtime mostly use version-less RIDs, except for RID compatibility tests)

v-wuzhai:
Superceded by #29660

PR #29716, opened 2023-01-03
cover template samples with tests + fix floating issue with test run

YuliiaKovalova:
cover template samples with tests + fix floating issue with test run
`The API 'Microsoft.DotNet.Cli.New.IntegrationTests.VerifySettingsFixture..ctor' must be called prior to any Verify has run. Usually this is done in a [ModuleInitializer].`


src/Tests/dotnet-new.Tests/TemplateEngineSamplesTest.cs
```diff
@@ -0,0 +1,98 @@
...
+        private readonly ILogger _log;
+
+        public TemplateEngineSamplesTest(SharedHomeDirectory sharedHome, ITestOutputHelper log)
+            : base(log)
+        {
+            _log = new XunitLoggerProvider(log).CreateLogger("TestRun");
+            _sharedHome = sharedHome;
+            _sharedHome.InstallPackage(
+               "Microsoft.TemplateEngine.Samples",
+               nugetSource: @"https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json");
```
JanKrivanek, line 26: Can we omit the explicit nuget source and let the `dotnet new install` use the sources from nuget.config in repo root?

src/Tests/dotnet-new.Tests/TemplateEngineSamplesTest.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                sb.Append('.');
+
+                for (int index = 0; index < args.Length; index += 2)
+                {
+                    sb.Append($"{args[index].Replace("--", "")}={args[index + 1]}");
+                    if (index < args.Length - 2)
+                    {
+                        sb.Append('.');
+                    }
```
JanKrivanek, line 91: This is completely fine as is. 

Just a super optional (and opinionated) suggestion on shortening the code - the string can be composed unconditionally with the separating dot (`sb.Append($"{args[index].Replace("--", "")}={args[index + 1]}.");`) and then just in the final `ToString` we can slash the last dot (`sb.ToString(0, Math.Max(0,sb.Length-2))`)

JanKrivanek (changes requested):
Looks good!

Please have a look if hardcoded package source can be removed - that would improve maintainability.

src/Tests/dotnet-new.Tests/TemplateEngineSamplesTest.cs
```diff
@@ -52,6 +53,7 @@ public TemplateEngineSamplesTest(SharedHomeDirectory sharedHome, ITestOutputHelp
             _log.LogInformation($"Template with {caseDescription}");
             Dictionary<string, string> environmentUnderTest = new() { ["DOTNET_NOLOGO"] = false.ToString() };
             TestContext.Current.AddTestEnvironmentVariables(environmentUnderTest);
+            FileExtensions.AddTextExtension(".cshtml");
```
JanKrivanek, line 54: I believe you need to specify the extensions without the dot in this API

JanKrivanek: Though the build succeeds now, which is quite strange

JanKrivanek: Oh - it actualy strips the dot (https://github.com/VerifyTests/EmptyFiles/blob/79678f2f459d7969425fbdbb0cad3cc88677d728/src/EmptyFiles/FileExtensions.cs#L17), unlinke some other Verify APIs.

So all is good

JanKrivanek (approved):

PR #29717, opened 2023-01-03
fix the floating issue with tests run in VerifySettingFixture..ctor

YuliiaKovalova:
fix the floating issue with tests run:
` "The API 'Microsoft.DotNet.Cli.New.IntegrationTests.VerifySettingsFixture..ctor' must be called prior to any Verify has run. Usually this is done in a [ModuleInitializer]."`


JanKrivanek (approved):
Thank you!

Issue #29151, opened 2022-11-21
`dotnet publish` shows garbled characters when running on PMC on OS with Traditional Chinese language


dominoFire:
Bug transferred from VS Feedback

### Describe the bug

_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/程序包管理器控制台主机版本-640111-乱码/10203703)._

---
程序包管理器控制台主机版本 6.4.0.111 乱码

dotnet publish 命令乱码

![2022-11-16 22 08 50.png](https://aka.ms/dc/image?name=Bd157fd5af3e24cdd90128823fb39370c638042046055504960_2022-11-16_22_08_50.png&tid=d157fd5af3e24cdd90128823fb39370c638042046055504960)

---
#### Feedback Bot on 11/16/2022, 06:37 PM: 

(private comment, text removed)
#### Feedback Bot on 11/17/2022, 00:07 PM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

### To Reproduce

1. Set Windows language to Traiditional Chinese
1. Set Visual Studio Language to same as OS system.
1. Create a New Console project
1. Open Package Manager Console in the recently created Console App
1. Run "dotnet publish" in Package Manager Console.

### Exceptions (if any)

![image](https://user-images.githubusercontent.com/1192347/203138288-92d9699f-232a-41ec-a834-185f182fda6a.png)

### Further technical details
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and its version

luantian-terark:
I got same problem.
it happends in dotnet tools version 7 and my projects sdk versions are dotnet 6（run **dotnet publish**）.  My codepage in powershell is 936(GBK), found garbled characters and got correct characters in code page 65001(UTF8).
note: i got correct characters in `dotnet --info` but garbage in `dotnet build` with code page 936

```
PS D:\> dotnet --info
.NET SDK:
 Version:   7.0.100
 Commit:    e12b7af219

运行时环境:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.100\

Host:
  Version:      7.0.0
  Architecture: x64
  Commit:       d099f075e4

.NET SDKs installed:
  7.0.100 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
[...]

Issue #24757, opened 2022-04-08
[Loc]With .NET 7 Preview 3 SDK zip installed on local OS, some garbled characters appear when installing wasm-tools for the first time


WenJunJi01:
**Environment:** .NET SDK 7.0.100-preview.3.22179.4(runtime-7.0.0-preview.3.22175.4)

**Repro steps:**

1. Set up a clean local OS machine with .NET SDK 7.0.100-preview.3.22179.4 zip installation
2. Run dotnet workload install wasm-tools --source https://pkgs.dev.azure.com/dnceng/public/_packaging/7.0.100-preview.3.22179.4-shipping-1/nuget/v3/index.json

**Expected Result:**
wasm-tools workload install successfully in CLI and strings are localized well.

**Actual Result:**
Some garbled characters appear when installing wasm-tools for the first time.
SC:
![image](https://user-images.githubusercontent.com/87680983/162380014-b7ecc5a4-3fd1-4343-8dc0-abcd6594bfbb.png)
DE:
![image](https://user-images.githubusercontent.com/87680983/162379849-0cd19c60-59e8-4b05-ae1e-e19cfb8e09e9.png)
TK:
![image](https://user-images.githubusercontent.com/87680983/162380242-094f5028-f4f4-47e8-be11-746e98bc2d70.png)

**Note:**
After uninstalling and reinstalling it, the localized characters can be displayed normally.
![image](https://user-images.githubusercontent.com/87680983/162379281-91d1c16d-d6e3-4f63-bb2f-cd642d6435d1.png)

**Dotnet info:**
.NET SDK (gemäß "global.json"):
 Version:   7.0.100-preview.3.22179.4
 Commit:    c48f2c30ee

Laufzeitumgebung:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\tt\sdk\7.0.100-preview.3.22179.4\

Host (useful for support):
  Version: 7.0.0-preview.3.22175.4
  Commit:  162f83657c

.NET SDKs installed:
  7.0.100-preview.3.22179.4 [C:\tt\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 7.0.0-preview.3.22178.4 [C:\tt\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 7.0.0-preview.3.22175.4 [C:\tt\shared\Microsoft.NETCore.App]
[...]

PR #29755, opened 2023-01-05
Use UTF-8 Encoding by Default On Windows On Non English Languages

nagilson:
Fix for: https://github.com/dotnet/sdk/issues/8833, https://github.com/dotnet/sdk/issues/24757, https://github.com/dotnet/sdk/issues/29151, and @rawieo's remark: https://github.com/microsoft/vstest/issues/821

Before: 
image

After: 
image

The main fix is setting the CHCP and OutputEncoding to UTF8 by default on Windows, if a custom language is set.

Things to know:

- [x] Make sure this works in CMD and Powershell (Works in Powershell, no easy way to test it in CMD. Hasn't been tested, but Powershell seems to the trickier one because it has its own custom preference variable)
- [x] Make this take effect if the system language is not english as well. Note that this could be weird if VS has different behavior, but VSLANG is respected above OS lang.
- [x] Add tests if possible ( I did not add tests, they would be a massive pain. There is a fairly close attempt in PR history But we would have to change the SDK to make it testable.)
- [x] `dotnet new` duplicated a lot of the old code... but it's also fixed because this runs through `Main` in the CLI. We could consider removing that code, but I didn't to lower risk of breaks. 
- [x] MSBuild does not respect the environment variables so that's still english ATM. We should do the work to align those once this is approved, I can migrate/copy this code to MSBuild.  https://github.com/dotnet/msbuild/issues/1596
- [x] `dotnet --debug` will have broken encoding, because it runs before the encoding is set. Otherwise the debugger would not be able to attach to the encoding changing code. We don't expect many people to be impacted by this, probably not worth fixing.
image
- [x] UTF8 was not super-officially supported until Win10. We should only do this if it's supported to not break Win7 below, especially.


nagilson:
> Does this affect the emoji that `dotnet watch` outputs? The OEM codepage of the culture might not include those characters.
> 
> https://github.com/dotnet/sdk/blob/882a576873b701f79d6584e4a0cf84036fdcd83f/src/BuiltInTools/dotnet-watch/Internal/ConsoleReporter.cs#L61

@KalleOlaviNiemitalo That's a really good point. I was considering setting the page to UTF-8. It does indeed break the emoji in certain languages. UTF-8, meanwhile, is fine, though it's somewhat concerning in legacy scenarios.

nagilson:
Not sure, but you're right in that InputEncoding doesn't work for certain `System.Runtime.Versioning.UnsupportedOSPlatform`, like `tvos` or `android`, so I was planning to see what happens when you try to use it on an unsupported OS.

nagilson:
Yes, this does not appear to repro on Linux. I'll check other platforms I can access in a bit, but limiting it to Windows for now 👍

dsplaisted (commented):
Looks good to me, but I don't know glob/loc details such as the different between CurrentCulture and CurrentUICulture, so we should get someone who does to review this.

dsplaisted:
@tarekgh can probably review this or point us to someone who can.

src/Cli/dotnet/UILanguageOverride.cs
```diff
@@ -24,7 +26,20 @@ public static void Setup()
 
         private static void ApplyOverrideToCurrentProcess(CultureInfo language)
         {
+
+            CultureInfo.DefaultThreadCurrentCulture = language;
```
tarekgh, line 30: >CultureInfo.DefaultThreadCurrentCulture = language; [](http://example.com/codeflow?start=12&length=51)

Why you are changing this?

src/Cli/dotnet/UILanguageOverride.cs
```diff
@@ -61,6 +82,11 @@ private static CultureInfo GetOverriddenUILanguage()
                 catch (CultureNotFoundException) { }
             }
 
+            if (!CultureInfo.InstalledUICulture.TwoLetterISOLanguageName.Equals("en", StringComparison.InvariantCultureIgnoreCase))
```
tarekgh, line 85: what you are trying to do here? CultureInfo.InstalledUICulture not necessary will tell the user UI language. I don't think this logic correct and I am not sure what you are trying to achieve here.


[https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs,422](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs,422)

src/Cli/dotnet/UILanguageOverride.cs
```diff
@@ -24,7 +26,20 @@ public static void Setup()
...
             CultureInfo.DefaultThreadCurrentUICulture = language;
+
+            // The CurrentUICulture can be supposedly be left incorrectly unchanged even when DefaultThreadCurrentUICulture is changed:
+            // https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.defaultthreadcurrentculture?redirectedfrom=MSDN&view=net-7.0#remarks
+            CultureInfo.CurrentUICulture = language;
+
+            if (OperatingSystem.IsWindows() && Environment.OSVersion.Version.Major >= 10) // Encoding is only an issue on Windows, UTF-8 is only officially supported on 10+.
+            {
+                Console.OutputEncoding = DefaultMultilingualEncoding;
+                Console.InputEncoding = DefaultMultilingualEncoding; // Setting both encodings causes a change in the CHCP, making it so we dont need to P-Invoke ourselves.
```
tarekgh, line 40: @adamsitnik would setting UTF-8 encoding to the console be ok?

src/Cli/dotnet/UILanguageOverride.cs
nagilson: We are getting the OS System Language here to check if it is english

tarekgh: What you mean by the system language? User locale? UI locale? system locale? and if it is English, why you force the UI langue to English at that time? 

InstalledUILanguage returns the user locale which is not necessarily the default UI language on the system. I am wondering why you don't just run with whatever UI language set by the user. Why do you want to change this setting even when not using any config switch to override the user settings?

src/Cli/dotnet/UILanguageOverride.cs
nagilson: This is so the date formatting and such are aligned to the same culture as the UI Culture, https://stackoverflow.com/questions/329033/what-is-the-difference-between-currentculture-and-currentuiculture-properties-of

src/Cli/dotnet/UILanguageOverride.cs
nagilson: Thanks for taking the time to review!

dsplaisted: I think the intent here is to set the output encoding to UTF8 if we're using a non-English language, either because it's the default or because it's overridden using `DOTNET_CLI_UI_LANGUAGE` or `VSLANG` environment variables.

nagilson: By system language I mean this:
image

We are setting the encoding if it is not english to support non english chars. If the system language is JP then dotnet should also be in JP and shouldn't show ??? because it would be in US english chcp page without this change. If my system is set to JP and I don't have english installed on my system dotnet shouldn't be in english by default.

src/Cli/dotnet/UILanguageOverride.cs
tarekgh: It is not right to force the User and UI languages to be same. Imagine I am a user like to have English UI and use Italian language for the date/time formatting. This change is going to break that.

src/Cli/dotnet/UILanguageOverride.cs
tarekgh: I hope we focus more here on the user scenarios to know what exactly we need to achieve. If it is around the console encoding, I am not seeing any reason we need to change anything handling User/UI language.

nagilson: > I am not seeing any reason we need to change anything handling User/UI language

Maybe this is referring to the code above... but the InstalledUICulture is referring to the OS Display Language. https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.installeduiculture?view=netframework-4.8 This is focused on a specific user scenario, which is what's stated above: currently if my OS is in a not english language `dotnet` will still be in english, this makes it so it uses the language of the OS.

src/Cli/dotnet/UILanguageOverride.cs
nagilson: I see, I'll remove this part. I didn't realize this was controlled by a flag/option in Windows

src/Cli/dotnet/UILanguageOverride.cs
tarekgh: `InstalledUICulture` is not doing what you are looking for. Imagine you install Windows machine which has UI language like Japanese and setting the user locale to English. `InstalledUICulture` will have English at that time and not Japanese. If you want get the UI language, just use `CultureInfo.CurrentUICulture`.

nagilson: Thanks again for taking the time to review and explain 😄

src/Cli/dotnet/UILanguageOverride.cs
nagilson: According to this, https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page, after the May 2019 Windows 10 update UTF-8 is officially supported. Since this only happens in Windows 10+, and when using languages besides English, which already have a bad encoding without this change, I think we can take this change for 8. 

cc @dsplaisted  for potential approval.

dsplaisted (approved):
👍

PR #29785, opened 2023-01-06
Add DynamicCodeSupport Feature Switch

eerhardt:
Adds support in the SDK for the new RuntimeFeature.IsDynamicCodeSupported feature switch added in  https://github.com/dotnet/runtime/issues/39806. When PublishAot is set in a project, DynamicCodeSupport gets defaulted to false.

### DIFF
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -25,7 +25,7 @@
     <NullabilityInfoContextSupport>false</NullabilityInfoContextSupport>
     <CustomResourceTypesSupport>false</CustomResourceTypesSupport>
     <UseSystemResourceKeys>true</UseSystemResourceKeys>
-    <DynamicCodeSupport>false</DynamicCodeSupport>
+    <DynamicCodeSupport>true</DynamicCodeSupport>
     <BuiltInComInteropSupport>false</BuiltInComInteropSupport>
     <_EnableConsumingManagedCodeFromNativeHosting>false</_EnableConsumingManagedCodeFromNativeHosting>
     <EnableCppCLIHostActivation>false</EnableCppCLIHostActivation>
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -25,6 +25,7 @@
     <NullabilityInfoContextSupport>false</NullabilityInfoContextSupport>
     <CustomResourceTypesSupport>false</CustomResourceTypesSupport>
     <UseSystemResourceKeys>true</UseSystemResourceKeys>
+    <DynamicCodeSupport>false</DynamicCodeSupport>
     <BuiltInComInteropSupport>false</BuiltInComInteropSupport>
     <_EnableConsumingManagedCodeFromNativeHosting>false</_EnableConsumingManagedCodeFromNativeHosting>
     <EnableCppCLIHostActivation>false</EnableCppCLIHostActivation>
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
@@ -705,6 +705,34 @@ namespace Microsoft.NET.Publish.Tests
             }
         }
 
+        [RequiresMSBuildVersionTheory("17.0.0.32901")]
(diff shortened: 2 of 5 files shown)


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -705,6 +705,34 @@ public void It_publishes_with_implicit_rid_with_NativeAotApp(string targetFramew
...
 
+        [RequiresMSBuildVersionTheory("17.0.0.32901")]
+        [InlineData(ToolsetInfo.CurrentTargetFramework)]
+        public void It_builds_with_dynamiccodesupport_false(string targetFramework)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                var projectName = "DynamicCodeSupportFalseApp";
+                var testProject = CreateHelloWorldTestProject(targetFramework, projectName, true);
+                testProject.AdditionalProperties["PublishAot"] = "true";
```
MichalStrehovsky, line 716: Did you mean to set DynamicCodeSupport to false instead?

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -705,6 +705,34 @@ public void It_publishes_with_implicit_rid_with_NativeAotApp(string targetFramew
             }
         }
 
+        [RequiresMSBuildVersionTheory("17.0.0.32901")]
+        [InlineData(ToolsetInfo.CurrentTargetFramework)]
+        public void It_builds_with_dynamiccodesupport_false(string targetFramework)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
MichalStrehovsky, line 712: Is it a pattern to restrict to these platforms?

src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
```diff
@@ -25,6 +25,7 @@
     <NullabilityInfoContextSupport>false</NullabilityInfoContextSupport>
     <CustomResourceTypesSupport>false</CustomResourceTypesSupport>
     <UseSystemResourceKeys>true</UseSystemResourceKeys>
+    <DynamicCodeSupport>false</DynamicCodeSupport>
```
dsplaisted, line 28: This is a test project which is used in a lot of different tests, does it make sense to turn of dynamic code support for all the tests that use the project?

dsplaisted (approved):

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
eerhardt: No. This test is testing that when `PublishAot=true`, the `System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported` feature switch is defaulted to `false`.

vitek-karas: This way it tests that `PublishAot` implies turning it off, so I think we need this test. We could add another one which sets the property directly, but that one is less important in my mind.

vitek-karas (approved):

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
eerhardt: The rest of the tests in this file do, presumably because NativeAot isn't supported on macOS.

I can remove this check since we are just building the project here.

src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
eerhardt: I am just testing that the value flows through to the runtimeconfig.json. I can flip it to `true` to not affect other tests.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
eerhardt: > We could add another one which sets the property directly

That's the "KitchenSink" test changes above.

sbomer (approved):

AntonLapounov (approved):

MichalStrehovsky (approved):

PR #29787, opened 2023-01-06
Add PackageReference to new Roslyn package

Forgind:
## Description

Users can set the `BuildWithNetFrameworkHostedCompiler` property to `true` to automatically inherit the framework equivalent of the .NET compiler included in this SDK with a version matching the core compiler.

## Customer Impact

If the version of the SDK is different from what is used in VS, it can cause numerous problems as different components expect different versions hence APIs and functionality. This resolves the problem by providing a way for the SDK to acquire a .NET Framework version of Roslyn with the same version as that in the SDK. (The version is specified in the dotnet/installer repo.)

That functionality is turned off for non-Windows machines, as they do not support .NET Framework. It also provides a warning if you try to reference the package directly, as that is not intended behavior.

## Regression?
* [ ] Yes
* [x] No

## Risk
* [ ] High
* [ ] Medium
* [x] Low

The change itself is well-understood and available already in main. It's opt-in and only adds a warning if the user directly references a package jaredpar does not want users to reference. New automated tests were added to verify that the feature works as expected.

## Verification

* [x] manual
* [x] automated tests

(Automated tests are currently disabled because they rely on the version provided by the installer. The installer has not yet flowed to the SDK, so they do not work, but they will ship together, which means customers should get the correct version.)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -29,6 +29,10 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="GetAssemblyVersion" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="GenerateSatelliteAssemblies" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
+  <ItemGroup Condition="'$(BuildWithFrameworkRoslyn)' == 'true' and '$(MicrosoftNetCompilersToolsetPackageVersion)' != ''">
+    <PackageReference Include="Microsoft.Net.Compilers.Framework.Package" PrivateAssets="all" Version="$(MicrosoftNetCompilersToolsetPackageVersion)"/>
```
Forgind, line 33:
```suggestion
    <PackageReference Include="Microsoft.Net.Compilers.Toolset.Framework" PrivateAssets="all" Version="$(MicrosoftNetCompilersToolsetPackageVersion)"/>
```

joeloff (approved):

joeloff:
Sorry, forgot to ask - is there a chance the package reference might conflict, or is the intention that they can override it by setting the related package version property?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -29,6 +29,10 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="GetAssemblyVersion" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="GenerateSatelliteAssemblies" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
+  <ItemGroup Condition="'$(BuildWithFrameworkRoslyn)' == 'true' and '$(MicrosoftNetCompilersToolsetPackageVersion)' != ''">
```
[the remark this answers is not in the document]
mmitche: We also don't appear to mention the name "Roslyn" much as part of official property names. It's there in some APICompat code, but minimally otherwise. Should we be using something other than the original codename?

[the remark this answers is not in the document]
Forgind: Sorry I was late getting back to this. Do you have a suggestion for a better property name?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -29,6 +29,10 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="GetAssemblyVersion" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="GenerateSatelliteAssemblies" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
+  <ItemGroup Condition="'$(BuildWithFrameworkRoslyn)' == 'true' and '$(MicrosoftNetFrameworkCompilersToolsetPackageVersion)' != '' and @(PackageReference->AnyHaveMetadataValue('Identity', 'Microsoft.Net.Compilers.Toolset.Framework')) != 'true'">
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset.Framework" PrivateAssets="all" Version="$(MicrosoftNetFrameworkCompilersToolsetPackageVersion)"/>
```
mmitche, line 33: Throwing some out there:

`BuildWithFrameworkRoslyn` -> `BuildWithNetFrameworkHostedCompiler`
`MicrosoftNetFrameworkCompilersToolsetPackageVersion` -> `NetFrameworkHostedCompilersVersion`

jaredpar: > and '$(MicrosoftNetFrameworkCompilersToolsetPackageVersion)' != '' and @(PackageReference->AnyHaveMetadataValue('Identity', 'Microsoft.Net.Compilers.Toolset.Framework')) != 'true'"

I feel like this conditions end up masking problems and making it harder for customers to detect issues. Basically it creates two cases where a customer says `<BuildWithFrameworkRoslyn>true</BuildWithFrameworkRoslyn>` but we **silently** do not actually do it. 

Instead can we consider putting a diagnostic here when either of these conditions are met? Essentially very visibly alert the customer that a problem exists.

jaredpar: > MicrosoftNetFrameworkCompilersToolsetPackageVersion

How does this property get set on customer installations? Basically when a customer installs the .NET SDK what props file sets this value? Expecting to see a change here that pushes it but don't.

Forgind: > How does this property get set on customer installations? Basically when a customer installs the .NET SDK what props file sets this value? Expecting to see a change here that pushes it but don't.

That's [this](https://github.com/dotnet/installer/blob/4f3e7dba2ff24dff107813d0f85dc38ddb2b3cc9/src/redist/targets/GenerateBundledVersions.targets#L355).

jaredpar: > That's [this](https://github.com/dotnet/installer/blob/4f3e7dba2ff24dff107813d0f85dc38ddb2b3cc9/src/redist/targets/GenerateBundledVersions.targets#L355).

Ah okay. So what version of the .NET SDK can I see this value being present? Tried searching some older 8.0 previews and didn't see it. Wanted to just kinda get a feel for where this is. 

> That just means that someone wanted to pin to an older version of the framework compiler, so they added their own PackageReference before this was imported.

Correct but that is essentially the build file in conflict with itself. There are essentially two declarations saying "add this package" and we're silently picking one.

Forgind: 8.0.100-preview 3, I think. Does that sound right, @joeloff?

This feels like the most natural way to add an override to me. People are already familiar with adding PackageReferences to various packages and including versions as desired; this just ensures that if they do that, we don't download a wrong version. We could instead recommend that people overwrite the version provided by default in the SDK, but then they'd have to know about our particular version property's name. Do you have a different way you think an override should be introducible?

jaredpar: > Do you have a different way you think an override should be introducible?

If people want to manually override they should add the package reference directly and unset the property. The issue with the current design is when we see `<BuildWithFrameworkRoslyn>true</BuildWithFrameworkRoslyn>` we don't actually know what is happening. Customers could be using the framework compiler we introduced or they could be overriding it. Both are valid outcomes. 

That seems incorrect to me. They specified they wanted the framework roslyn, we should give that to them or fail because we can't give it to them. 

If the customer is both setting the property and adding a package reference then they've effectively set conflicting build properties. We should error and tell them to decide which they want.

Forgind: I talked with marcpopMSFT, and we're not convinced that's a problem we have to solve. I agree they could be using the framework compiler _version_ we introduced or overriding the _version_, but they're using the framework compiler either way—only the version shifted—so I still feel the property is valid. It isn't currently being used in this way, but I could easily imagine later using that property to affect other parts of the build.

The big question is how often customers will break themselves by doing something. In this case, it doesn't seem terribly harmful to have overridden the version in addition to asking us to use the default version if they don't override it. That said, not having a version is certainly a problem. I suggested that I'd just remove the check, in which case we try to resolve a PackageReference with no version, which leads to a NuGet warning.

jaredpar: > The big question is how often customers will break themselves by doing something

I strongly disagree. I think the question is how much effort does it take from us to stop the customer from hurting themselves in the first place. The effort here is _minimal_. The change to stop a bug from ever happening in the first place is orders of magnitude less than the text we've already written here. 

The change as written is inviting customers to write MSBuild code that does not do as it's stated. The documentation for `$(BuildWithFrameworkRoslyn)` will need to be 

> This might use the Roslyn compiler from the SDK or it might not. 

Why are we fighting making the customer story here significantly simpler?

joeloff: Is the `$(BuildWithFrameworkRoslyn)` used for anything else? If the `<PackageReference/>` exists, it's like an implicit version the user configured and that implies `BuildWithFrameworkRoslyn=true`. However, if the package doesn't exist, but the property is `true`, that indicates the user wants to use the default SDK values.

What if we named the property `BuildWithDefaultFrameworkRoslyn`. If `true` and the user defines a `PackageReference` we error out because there can only be one version, and this happens later in some target, post-evaluation.

I don't know enough to say what the default should be. Assuming it's the SDK copy, then setting the property to true would require an explicit gesture from a user to set it to false and provide a package reference that would also have a specific/latest version attached to it.

The version property might still be an issue. Can we rename it and prefix it with `_` - convention says that's an internal property that users should not override and if they do, they're on their own.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -29,6 +29,10 @@ Copyright (c) .NET Foundation. All rights reserved.
   <UsingTask TaskName="GetAssemblyVersion" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <UsingTask TaskName="GenerateSatelliteAssemblies" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
+  <ItemGroup Condition="'$(BuildWithFrameworkRoslyn)' == 'true' and '$(MicrosoftNetFrameworkCompilersToolsetPackageVersion)' != '' and @(PackageReference->AnyHaveMetadataValue('Identity', 'Microsoft.Net.Compilers.Toolset.Framework')) != 'true'">
```
ViktorHofer, line 32: Check if PackageReference is not empty as otherwise the `AnyHaveMetadataValue` msbuild function would throw an exception. Also no need to do the `!= true`.

```suggestion
  <ItemGroup Condition="'$(BuildWithFrameworkRoslyn)' == 'true' and '$(MicrosoftNetFrameworkCompilersToolsetPackageVersion)' != '' and ('@(PackageReference)' == '' or  !@(PackageReference->AnyHaveMetadataValue('Identity', 'Microsoft.Net.Compilers.Toolset.Framework')))">
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
ViktorHofer: FWIW the .NET Framework targeting pack auto package reference does the following: https://github.com/dotnet/sdk/blob/31d20a223b9d3442d0641bb6d54a1d0f56b682a2/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets#L462-L465

Would we good to align with that code to be consistent.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
jaredpar: This is the case where there are no package references?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
jaredpar: > Would we good to align with that code to be consistent. 

Where does the user set a property that explicitly asks for this package to be included? That is my key problem here. We're defining a property whose entire intent is "reference this package" then introducing a case where we silently fail to do that. This is going to make build more confusing, not less.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
ViktorHofer: Yes. If no package references exist, the `AnyHaveMetadataValue` msbuild function would fail. That's an unfortunate behavior with msbuild item functions that return a boolean.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
ViktorHofer: Right, .NET Framework targeting pack auto-references are opt-out instead of opt-in.

> We're defining a property whose entire intent is "reference this package" then introducing a case where we silently fail to do that. This is going to make build more confusing, not less.

Agreed. I suggest to verify if NuGet already emits an error for duplicate package references with `IsImplicitlyDefined` metadata and if not, emit an error ourselves.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
Forgind: I put in a fix specifically for AnyHaveMetadataValue to make it return the proper value even if there are no PackageReferences. Prior to that, it would just notice an empty list and return an empty list. != true handles both cases.

jaredpar (approved):

jaredpar:
> Question, will this also be downloaded when targeting NETCoreApp or only for NETFramework? 

That's a good point. As written it's going to download the package for `msbuild` or `dotnet build`. I think that is unavoidable given how you can interleave restore and build across the two tools. Perhaps there is a nuance that makes it safe but I'm not familiar enough to say. 

It does seem unfortunate we're downloading it on platforms where it can never take effect like Linux. 
> Ie when I set the property to true, do I then use the NETFramework Roslyn everywhere?

This will only use .NET Framework Roslyn when building on .NET Framework MSBuild. The [package itself](https://github.com/dotnet/roslyn/blob/main/src/NuGet/Microsoft.Net.Compilers.Toolset/Framework/build/Microsoft.Net.Compilers.Toolset.Framework.props#L10) has the logic for that.

ViktorHofer (commented):
Overall looks good (aside from the use of an item for the condition). I would put the logic inside a target so that PackageReference items that are added after evaluation are considered as well. Having a target also cleans up the logic:

```
<Target Name="AddMicrosoftNetCompilerToolsetFrameworkPackage"
        Condition="'$(BuildWithNetFrameworkHostedCompiler)' == 'true'"
        BeforeTargets="CollectPackageReferences">

  <PropertyGroup Condition="'@(PackageReference->AnyHaveMetadataValue('Identity', 'Microsoft.Net.Compilers.Toolset.Framework'))' == 'true'">
    <_UserDirectlyReferencedMicrosoftNetCompilersToolsetFrameworkPackage>true</_UserDirectlyReferencedMicrosoftNetCompilersToolsetFrameworkPackage>
  </PropertyGroup>

  <!-- Users should not be setting Microsoft.Net.Compilers.Toolset.Framework directly. If they do, and they also set BuildWithNetFrameworkHostedCompiler,
       we will have a duplicate PackageReference. This makes it more explicit that that is not supported. -->
  <NETSdkWarning ResourceName="CannotDirectlyReferenceMicrosoftNetCompilersToolsetFramework"
                 Condition="'$(_UserDirectlyReferencedMicrosoftNetCompilersToolsetFrameworkPackage)' == 'true'" />

  <ItemGroup Condition="'$(BuildWithNetFrameworkHostedCompiler)' == 'true' and '$(OS)' == 'Windows_NT'">
    <PackageReference Include="Microsoft.Net.Compilers.Toolset.Framework" Version="$(_NetFrameworkHostedCompilersVersion)" PrivateAssets="all" IsImplicitlyDefined="true" />
  </ItemGroup>
  
</Target>
```

src/Tests/Microsoft.NET.Restore.Tests/GivenThatWeWantToUseFrameworkRoslyn.cs
```diff
@@ -27,6 +28,8 @@ public void It_restores_Microsoft_Net_Compilers_Toolset_Framework_when_requested
                 TargetFrameworks = "net6.0",
             };
 
+            // Add an explicit version for the test. This will normally come from the installer.
+            project.AdditionalProperties.Add("_NetFrameworkHostedCompilersVersion", "4.7.0-2.23260.7");
```
ViktorHofer, line 32: nit: From which feed is that package restored? If it's a dotnetX feed, then we will need to rev the version when we drop that feed.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -27,7 +27,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+         If they do, and they also set BuildWithNetFrameworkHostedCompiler, we will have
+         a duplicate PackageReference. This makes it more explicit that that is not supported. -->
+    <NETSdkWarning ResourceName="CannotDirectlyReferenceMicrosoftNetCompilersToolsetFramework"
+        Condition="'@(PackageReference->AnyHaveMetadataValue('Identity', 'Microsoft.Net.Compilers.Toolset.Framework'))' == 'true'" /> 
+
+    <ItemGroup Condition="'$(BuildWithNetFrameworkHostedCompiler)' == 'true' and '$(OS)' == 'Windows_NT'">
+      <PackageReference Include="Microsoft.Net.Compilers.Toolset.Framework" PrivateAssets="all" Version="$(_NetFrameworkHostedCompilersVersion)" IsImplicitlyDefined="true" />
+    </ItemGroup>
+
+  </Target>
```
ViktorHofer, line 44: nit: Please move the target down to the other targets in this file, i.e. next to `_CheckForBuildWithNoBuild`.

src/Tests/Microsoft.NET.Restore.Tests/GivenThatWeWantToUseFrameworkRoslyn.cs
Forgind: dotnet-tools

Why are we dropping feeds? I would've thought the feed would be permanent even if the particular version may disappear some day.

joeloff (approved):
We could make the localization safer for the vendors by adding `{Locked="Microsoft.Net.Compilers.Toolset.Framework"}{Locked="BuildWithNetFrameworkHostedCompiler"}` to the comment field in the resx file (keeping the `StrBegin` as well), but I'm not going to block on that.

src/Tests/Microsoft.NET.Restore.Tests/GivenThatWeWantToUseFrameworkRoslyn.cs
ViktorHofer: Dotnet-tools is a permanent feed so that's good then. We are dropping dotnetX feeds in subsequent releases, ie dotnet6* feeds shouldn't be used when the stack uses dotnet8* feeds.

PR #29817, opened 2023-01-07
Do a two-phase publish to mimic VS behavior

baronfel:
I wanted to take a run at this because it just makes sense to me.

Publish profiles work in VS from a technical level because they operate in two phases - one to locate and parse build properties from a specified publish profile pubxml, and the second to run the build's Publish target with those properties applied. This allows users to put properties that are specific to that publish configuration in a single file, not cluttering their project file.

This PR emulates this behavior. It 
* does an evaluation, adhering to any user-specified CLI options that might be relevant to MSBuild evaluation, 
  * LastUsedBuildConfiguration
  * Configuration
  * Platform
  * TargetFramework
  * TargetFrameworks
  * RuntimeIdentifier
* checks the results of that evaluation to determine a) if a profile was imported at all and b) if so, the path to the profile that was chosen
* loads the properties from the profile that was chosen
* appends those properties to the end of the MSBuild publish invocation, so that those properties take precedence (since MSBuild CLI properties are a latest-takes-precedence ordering)

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-publish/Program.cs
+++ b/src/Cli/dotnet/commands/dotnet-publish/Program.cs
@@ -43,30 +43,58 @@ namespace Microsoft.DotNet.Tools.Publish
             parseResult.HandleDebugSwitch();
             parseResult.ShowHelpOrErrorIfAppropriate();
 
-            var msbuildArgs = new List<string>()
-            {
-                "-target:Publish",
-                "--property:_IsPublishing=true" // This property will not hold true for MSBuild /t:Publish. VS should also inject this property when publishing in the future.
-            };
-
-            IEnumerable<string> slnOrProjectArgs = parseResult.GetValueForArgument(PublishCommandParser.SlnOrProjectArgument);
-
-            CommonOptions.ValidateSelfContainedOptions(parseResult.HasOption(PublishCommandParser.SelfContainedOption),
-                parseResult.HasOption(PublishCommandParser.NoSelfContainedOption));
-
-            msbuildArgs.AddRange(parseResult.OptionValuesToBeForwarded(PublishCommandParser.GetCommand()));
-            ReleasePropertyProjectLocator projectLocator = new ReleasePropertyProjectLocator(Environment.GetEnvironmentVariable(EnvironmentVariableNames.ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS) != null);
-            msbuildArgs.AddRange(projectLocator.GetCustomDefaultConfigurationValueIfSpecified(parseResult, MSBuildPropertyNames.PUBLISH_RELEASE,
-                slnOrProjectArgs, PublishCommandParser.ConfigurationOption) ?? Array.Empty<string>());
-            msbuildArgs.AddRange(slnOrProjectArgs ?? Array.Empty<string>());
-
             bool noRestore = parseResult.HasOption(PublishCommandParser.NoRestoreOption)
                           || parseResult.HasOption(PublishCommandParser.NoBuildOption);
 
+
+            var publishProfileProperties = DiscoverPropertiesFromPublishProfile(parseResult);
+            Console.WriteLine($"Read publish properties: {string.Join(",", publishProfileProperties)}");
+            var standardMSbuildProperties = CreatePropertyListForPublishInvocation(parseResult);
             return new PublishCommand(
-                msbuildArgs,
+                // properties defined by the selected publish profile should override any other properties,
+                // so they should be added after the other properties.
+                standardMSbuildProperties.Concat(publishProfileProperties),
                 noRestore,
-                msbuildPath);
+                msbuildPath
+            );
+
+            List<string> CreatePropertyListForPublishInvocation(ParseResult parseResult) {
+                var msbuildArgs = new List<string>()
+                    {
+                        "-target:Publish",
+                        "--property:_IsPublishing=true" // This property will not hold true for MSBuild /t:Publish. VS should also inject this property when publishing in the future.
+                    };
+
+                IEnumerable<string> slnOrProjectArgs = parseResult.GetValueForArgument(PublishCommandParser.SlnOrProjectArgument);
+
+                CommonOptions.ValidateSelfContainedOptions(parseResult.HasOption(PublishCommandParser.SelfContainedOption),
+                    parseResult.HasOption(PublishCommandParser.NoSelfContainedOption));
+
+                msbuildArgs.AddRange(parseResult.OptionValuesToBeForwarded(PublishCommandParser.GetCommand()));
+                ReleasePropertyProjectLocator projectLocator = new ReleasePropertyProjectLocator(Environment.GetEnvironmentVariable(EnvironmentVariableNames.ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS) != null);
+                msbuildArgs.AddRange(projectLocator.GetCustomDefaultConfigurationValueIfSpecified(parseResult, MSBuildPropertyNames.PUBLISH_RELEASE, slnOrProjectArgs, PublishCommandParser.ConfigurationOption) ?? Array.Empty<string>());
+                msbuildArgs.AddRange(slnOrProjectArgs ?? Array.Empty<string>());
+
+                return msbuildArgs;
+            }
+
+            List<string> DiscoverPropertiesFromPublishProfile(ParseResult parseResult)
+            {
+                ReleasePropertyProjectLocator projectLocator = new ReleasePropertyProjectLocator(Environment.GetEnvironmentVariable(EnvironmentVariableNames.ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS) != null);
+                var cliProps = projectLocator.GetGlobalPropertiesFromUserArgs(parseResult);
+                var projectInstance = projectLocator.GetTargetedProject(parseResult.GetValueForArgument(PublishCommandParser.SlnOrProjectArgument), cliProps);
+                var importedPropValue = projectInstance.GetPropertyValue(MSBuildPropertyNames.PUBLISH_PROFILE_IMPORTED);
+                if (!String.IsNullOrEmpty(importedPropValue) && bool.TryParse(importedPropValue, out var wasImported) && wasImported) {
+                    try {
+                        if (projectInstance.GetPropertyValue(MSBuildPropertyNames.PUBLISH_PROFILE_FULL_PATH) is {} fullPathPropValue) {
+                            return new ProjectInstance(fullPathPropValue).ToProjectRootElement().PropertyGroups.First().Properties.Select(p => $"--property:{p.Name}=\"{p.Value}\"").ToList();
+                        }
+                    } catch (IOException) {
+                        return new List<string>();
+                    }
+                }
+                return new List<string>();
+            };
         }
 
         public static int Run(ParseResult parseResult)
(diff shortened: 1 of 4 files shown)


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -794,16 +794,8 @@ public void PublishRelease_interacts_similarly_with_PublishProfile_Configuration
...
-            {
-                Assert.True(File.Exists(releaseAssetPath)); // We ignore Debug configuration and override it, IF its custom though, we dont use publishrelease.
-            }
-            else
-            {
-                Assert.False(File.Exists(releaseAssetPath)); // build will produce a debug asset, need to make sure this doesn't exist either.       
-                publishOutput.Should().HaveStdOutContaining("PublishRelease");
-            }
+            var configAssetPath = System.IO.Path.Combine(helloWorldAsset.Path, "bin", config, ToolsetInfo.CurrentTargetFramework, rid, "HelloWorld.dll");
+            Assert.True(File.Exists(configAssetPath)); // We cannot ignore Debug configuration and override it, because it's set as a globalproperty during publish.
```
baronfel, line 798: This change is the most controversial part of the entire PR - due to the way that publish profile-driven properties are passed into the 'main' build (they are passed via the `-p` option, so are treated as global properties), PublishRelease can no longer have any effect on the build. Is this ok?  I contend yes - the point of publish release was to make publish use Release mode by default, and the defaults for VS-authored publish profiles is to set configuration to `Release`.

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -43,30 +64,64 @@ public static PublishCommand FromParseResult(ParseResult parseResult, string msb
...
+                // this can happen if the project wasn't loadable, or if we're at a solution context
+                if (projectInstance == null)
+                {
+                    return new List<string>();
+                }
+                var importedPropValue = projectInstance.GetPropertyValue(MSBuildPropertyNames.PUBLISH_PROFILE_IMPORTED);
+                if (!String.IsNullOrEmpty(importedPropValue) && bool.TryParse(importedPropValue, out var wasImported) && wasImported) {
+                    try {
+                        if (projectInstance.GetPropertyValue(MSBuildPropertyNames.PUBLISH_PROFILE_FULL_PATH) is {} fullPathPropValue) {
+                            var properties = new ProjectInstance(fullPathPropValue).ToProjectRootElement().PropertyGroups.First().Properties;
```
nagilson, line 115: While I haven't done the side-by-side comparison, I imagine that it would be faster to loop through the properties in the propertiesToForwardFromProfile, instead of loading in every property on the project into memory

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -43,30 +64,64 @@ public static PublishCommand FromParseResult(ParseResult parseResult, string msb
...
+                msbuildArgs.AddRange(slnOrProjectArgs ?? Array.Empty<string>());
+
+                return msbuildArgs;
+            }
+
+            List<string> DiscoverPropertiesFromPublishProfile(ParseResult parseResult)
+            {
+                ReleasePropertyProjectLocator projectLocator = new ReleasePropertyProjectLocator(Environment.GetEnvironmentVariable(EnvironmentVariableNames.ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS) != null);
+                var cliProps = projectLocator.GetGlobalPropertiesFromUserArgs(parseResult);
+                var projectInstance = projectLocator.GetTargetedProject(parseResult.GetValueForArgument(PublishCommandParser.SlnOrProjectArgument), cliProps, MSBuildPropertyNames.PUBLISH_RELEASE);
```
nagilson, line 105: Note that this will be another +~18-25% performace regression in time to publish which I don't think we want. If we were to do this, I think it would be better to create something like a singleton wrapper which only ever instantiates or pre-evaluates the project once.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
nagilson: This seems fine to me. But I also think you could write it such that this isn't the case by removing the `Configuration` from the Publish Profile Evaluated Properties and Injecting the one discovered by the ReleasePropertyProjectLocator

src/Cli/dotnet/commands/dotnet-publish/Program.cs
nagilson, line 105: My main concern here is the `GetTargetedProject` call. In NET 8 we choose an arbitrary project in the solution as opposed to the executable project. This means the publish profile import could be non-deterministic unless further changes are made, and those changes, while not an extreme amount of work, also will need to be very well thought-through to prevent bugs

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -23,6 +23,27 @@ namespace Microsoft.DotNet.Tools.Publish
...
+        /// The list of properties that should be forwarded from the publish profile to the publish invocation.
+        /// </summary>
+        /// <remarks>
+        /// While we could forward along every property, the intent of this array is to mimic the behavior of VS,
+        /// whose Publish operation only forwards along a few properties. Of particular interest are properties that
+        /// are set very early on in the build (RID, Configuration, etc).  The remainder will be imported during the
+        /// build via the Microsoft.Net.Sdk.Publish props and targets.
+        /// </remarks>
+        private static string[] PropertiesToForwardFromProfile = new [] {
+            "Configuration",
```
nagilson, line 37: You may want to use `MSBuildPropertyNames` here, but it doesn't really matter too much.

src/Cli/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -43,30 +64,64 @@ public static PublishCommand FromParseResult(ParseResult parseResult, string msb
...
+
+                msbuildArgs.AddRange(parseResult.OptionValuesToBeForwarded(PublishCommandParser.GetCommand()));
+                ReleasePropertyProjectLocator projectLocator = new ReleasePropertyProjectLocator(Environment.GetEnvironmentVariable(EnvironmentVariableNames.ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS) != null);
+                msbuildArgs.AddRange(projectLocator.GetCustomDefaultConfigurationValueIfSpecified(parseResult, MSBuildPropertyNames.PUBLISH_RELEASE, slnOrProjectArgs, PublishCommandParser.ConfigurationOption) ?? Array.Empty<string>());
+                msbuildArgs.AddRange(slnOrProjectArgs ?? Array.Empty<string>());
+
+                return msbuildArgs;
+            }
+
+            List<string> DiscoverPropertiesFromPublishProfile(ParseResult parseResult)
```
nagilson, line 101: Please add a docstring for this function.

src/Cli/dotnet/commands/dotnet-publish/Program.cs
nagilson, line 115: I would prefer to use TryGetProjectInstance as race conditions could occur and cause failure if the project file is changed on disk in between this and the above lines of code

nagilson (commented):
It would be a big win to support publish profiles in the SDK. Thanks for this investigation. I think more work would be required to insert this into production, but great work so far.

src/Cli/dotnet/commands/dotnet-publish/Program.cs
baronfel: interesting - so the API changed from "if I'm called with a solution I return null" to "if I'm called with a solution I pick a project" in 8? That's good to know - for publish specifically we'd want to not run at all if the argument is a solution, so I can change this call to explicitly check that condition before calling `GetTargetedProject`.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
baronfel: IMO the point of this is to get alignment with VS explicitly by using the subset of properties tracked in the array at the top of the file, so I think we should _want_ the breaking change.

baronfel:
After some consideration I'm going to retarget this to 8.0 (but it will still need to be documented as a breaking change!).

src/Cli/dotnet/commands/dotnet-publish/Program.cs
nagilson: Yeah. Docstring is updated here: https://github.com/dotnet/sdk/blob/0e9ba20184689a18972aa755edf522ee5c7fa9c8/src/Cli/dotnet/ReleasePropertyProjectLocator.cs#L120

Balkoth:
Is anywhere concrete and complete documentation available on what properties are working from the pubxml file? I only found [this](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-publish#pubxml-files) which only gives away that there might be other properties which are not supported.

baronfel:
@Balkoth those docs were added in part because of the investigation required to get this PR going - VS uses PubXml properties for both _building_ the assets to publish, as well as potentially orchestrating pushing those assets to another location. The .NET CLI will only do the first of these (building the assets to publish). Because of this, the properties listed are the only ones we care about at the current time.

In general, all properties from the pubxml will be included in the build - but due to timing/ordering of the `Import` statement that includes the pubxml file into the build those properties may not be available when you expect. That's why I made this PR in the first place.  If you're not seeing a property 'take' in the way you expect from a pubxml, I'd create a [binlog](aka.ms/binlog) and inspect it with the binlog viewer to verify the ordering/usage.

Balkoth:
> Because of this, the properties listed are the only ones we care about at the current time.

In the linked document i can't see any properties which you care about, only ones that "aren't supported"

> In general, all properties from the pubxml will be included in the build - but due to timing/ordering of the `Import` statement that includes the pubxml file into the build those properties may not be available when you expect. That's why I made this PR in the first place. If you're not seeing a property 'take' in the way you expect from a pubxml, I'd create a [binlog](aka.ms/binlog) and inspect it with the binlog viewer to verify the ordering/usage.

I am "simply" trying to bring publishing of some projects off from client machines to our build server and build them from the command line with dotnet build.

Sample FolderProfile.pubxml of one of our projects:
```xml
<Project>
  <PropertyGroup>
    <Configuration>Release</Configuration>
    <Platform>Any CPU</Platform>
    <PublishDir>bin\publish\</PublishDir>
    <PublishProtocol>FileSystem</PublishProtocol>
    <TargetFramework>net7.0-windows</TargetFramework>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
    <PublishSingleFile>true</PublishSingleFile>
    <PublishReadyToRun>false</PublishReadyToRun>
    <IncludeNativeLibrariesForSelfExtract>true</IncludeNativeLibrariesForSelfExtract>
  </PropertyGroup>
</Project>
```

Naive me was thinking i could just do it like this in a PowerShell script, but the document says otherwise.
```ps
$targetProjectFile = "C:\Source\Project\Project.csproj"
$targetOutputDir = "C:\Binaries\Project"
dotnet publish $targetProjectFile --configuration Release --output $targetOutputDir /p:PublishProfile=FolderProfile
```

Is it even possible to pass all the arguments to dotnet publish manually so that it does exactly what Visual Studio does when publishing through the UI?

baronfel:
There are minor differences but they should be functionally equivalent. I used the 'generic' form of passing properties like Configuration and Runtime because I was assuming you were going to do a mechanical translation of "property from pubxml file" to "property to pass on the command line". You can of course use the more specific flags as well.

We're getting a bit off of the topic of this PR, which should be focused on code review/etc - if you're having specific issues with a publish command, can you open a new issue for triage and further discussion?

marcpopMSFT:
Per Chet, this needs to be redone so closing for now.

PR #29818, opened 2023-01-08
update eol list

kasperk81:
3.1 to eol list


marcpopMSFT:
fwiw, our plan was to update 6 and 7 SDKs for this in ~6 months. We got feedback from when 2.1 went out of support that this warning was the most disruptive change made during eol (as we made it right away for that release) and so for 5, we delayed until 6 months after the EOL to add the warning to give customers some time to migrate.

That being said, I don't see a reason to delay this for 8.0 since it'll be that way when it ships and it's a preview.

marcpopMSFT (approved):

PR #29860, opened 2023-01-10
enforce codestyle on build by default

jmarolf:
For .NET 8 we want this option to finally be enabled.

This will have the benefit of style warnings that appear in the IDE will also show up on commandline builds.


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
```diff
@@ -84,7 +84,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <EnableAotAnalyzer Condition="'$(EnableAotAnalyzer)' == '' And '$(PublishAot)' == 'true'">true</EnableAotAnalyzer>
 
     <!-- EnforceCodeStyleInBuild Allows code style analyzers to be disabled in bulk via msbuild if the user wants to -->
-    <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == ''">false</EnforceCodeStyleInBuild>
+    <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == ''">true</EnforceCodeStyleInBuild>
```
baronfel, line 87: This could be pretty disruptive - could it instead be triggered off of a check for EffectiveAnalysisLevel > 8? Everything in the current PropertyGroup describes the defaults for 5.0, so perhaps another group specifically for the 8.0 changes so it's all clear where the opt-ins happened?

PR #29880, opened 2023-01-11
Add a PostAction to support modifying Json files in CLI templates

fgheysels:
The DotnetModifyJsonPostActionProcessor PostAction allows dotnet template authors to modify JSON (existing) JSON files when creating a new project.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=29880) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=29880) before we can accept your contribution.<br/><br/>:x: fgheysels [sign now](https://cla.dotnetfoundation.org/dotnet/sdk?pullRequest=29880)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/sdk?pullRequest=29880) it.</sub>

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetModifyJsonPostActionProcessor.cs
```diff
@@ -0,0 +1,237 @@
...
+                    return null;
+                }
+
+                node = node[property];
+            }
+
+            return node;
+        }
+
+        private static IReadOnlyList<string> FindFilesAtOrAbovePath(
```
fgheysels, line 175: I could get rid of this code if we could have an overload of the `FindFilesAtOrAbovePath` method that exists in the  `FileFindHelpers` class

maridematte: Feel free to open a PR on the Template Engine to add an over load for the `FindFilesAtAbovePath`. We can keep this function here until the Template Engine PR is complete for the overloading function.

maridematte:
I changed to the `main` branch, as 7.0.2 release's date is very soon and this will probably not make it in. In the future, depending on how soon we can complete this PR we will add to the next iteration (probably 7.0.3).

src/Tests/dotnet.Tests/dotnet-new/DotnetModifyJsonPostActionTests.cs
```diff
@@ -0,0 +1,306 @@
...
+                    Assert.NotNull(modifiedJsonContent["person"]["address"]);
+                    Assert.Equal("street name", modifiedJsonContent["person"]["address"]["street"].ToString());
+                }),
+
+            new("Can add property to document root",
+                @"{""firstProperty"": ""foo""}",
+                new Dictionary<string, string>
+                {
+                    ["jsonFileName"] = "jsonfile.json",
+                    ["parentPropertyPath"] = null,
```
maridematte, line 221: If this value is assumed null when not specified, I would not include this line on the test, so we can check if the minimum requirements are being met / not met. This relates to my previous comment on the `ProcessInternal` function.

src/Tests/dotnet.Tests/dotnet-new/DotnetModifyJsonPostActionTests.cs
```diff
@@ -0,0 +1,306 @@
...
+using System.Text.Json.Nodes;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.New.PostActionProcessors;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Mocks;
+using Microsoft.TemplateEngine.TestHelper;
+using Moq;
+using Xunit;
+
+namespace Microsoft.DotNet.Cli.New.Tests
```
maridematte, line 17: A test with multiple changes to the json file would be very helpful!

maridematte (commented):
I think this PR is going the right way, if there is a better solution for the naming problem that would be great, but otherwise I think this workaround is ok!

maridematte:
Also, I have a small nit on the name of the post action. I feel like `ModifyJson` name in this case is a bit confusing, as this post action is just an addition to the file. Maybe something like `AddJsonProperty` would be more fitting for this post action.

fgheysels:
> Also, I have a small nit on the name of the post action. I feel like `ModifyJson` name in this case is a bit confusing, as this post action is just an addition to the file. Maybe something like `AddJsonProperty` would be more fitting for this post action.

Naming, one of the most difficult aspects in programming.

src/Tests/dotnet.Tests/dotnet-new/DotnetModifyJsonPostActionTests.cs
fgheysels: Hmm, in hindsight, I don't think running the PostAction twice in the same template will deliver the desired results.  Good idea to create a test for this; I'll see to that.

fgheysels:
> Also, I have a small nit on the name of the post action. I feel like `ModifyJson` name in this case is a bit confusing, as this post action is just an addition to the file. Maybe something like `AddJsonProperty` would be more fitting for this post action.

I have changed this and renamed the type to `DotnetAddJsonPropertyPostActionProcessor`

src/Tests/dotnet.Tests/dotnet-new/DotnetModifyJsonPostActionTests.cs
fgheysels: Ah, I can do that.  I just thought to make it explicit that this property is not specified.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetModifyJsonPostActionProcessor.cs
```diff
@@ -0,0 +1,237 @@
...
+            }
+
+            if (jsonFiles.Count > 1)
+            {
+                Reporter.Error.WriteLine(string.Format(LocalizableStrings.PostAction_ModifyJson_Error_MultipleJsonFiles, jsonFileName));
+                return false;
+            }
+
+            // args required:
+            // parent property name in json file (if not specified, assume null and add the json section to the root)
```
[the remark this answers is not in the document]
fgheysels: It is not required, I'll remove these comments

src/Tests/dotnet.Tests/dotnet-new/DotnetModifyJsonPostActionTests.cs
fgheysels: I have added an additional end-to-end test which tests this.

src/Assets/TestPackages/dotnet-new/Microsoft.TemplateEngine.TestTemplates.csproj
```diff
@@ -21,4 +21,9 @@
     <Content Include="nupkg_templates\**" />
     <None Include="nupkg_templates\**" CopyToOutputDirectory="Always" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Remove="test_templates\PostActions\AddJsonProperty\WithAddMultipleProperties\.template.config\template.json" />
+    <Content Remove="test_templates\PostActions\AddJsonProperty\WithAddMultipleProperties\MyTestProject.csproj" />
+    <Content Remove="test_templates\PostActions\AddJsonProperty\WithAddMultipleProperties\testfile.json" />
+  </ItemGroup>
```
vlada-shubina, line 28: please remove this change, it is not needed.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Abstractions.PhysicalFileSystem;
+using Microsoft.TemplateEngine.Cli.PostActionProcessors;
+
+namespace Microsoft.DotNet.Tools.New.PostActionProcessors
+{
+    internal class DotnetAddJsonPropertyPostActionProcessor : PostActionProcessorBase
```
vlada-shubina, line 21: imo, this post action will be better in https://github.com/dotnet/sdk/tree/main/src/Cli/Microsoft.TemplateEngine.Cli/PostActionProcessors as it has no dependency to `dotnet`.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,235 @@
...
+            ICreationResult templateCreationResult,
+            string outputBasePath)
+        {
+            if (!action.Args.TryGetValue(JsonFileNameArgument, out string? jsonFileName))
+            {
+                Reporter.Error.WriteLine(string.Format(LocalizableStrings.PostAction_ModifyJson_Error_ArgumentNotConfigured, JsonFileNameArgument));
+                return false;
+            }
+
+            IReadOnlyList<string> jsonFiles = FindFilesAtOrAbovePath(environment.Host.FileSystem, outputBasePath, matchPattern: jsonFileName, maxAllowedAboveDirectories: 1);
```
vlada-shubina, line 57: imo, this is a misuse of this method as json is expected exactly on path.
the refactoring to two different methods with extraction of common functionality may be of benefit.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,235 @@
...
+
+                directory = Path.GetPathRoot(directory) != directory ? Directory.GetParent(directory)?.FullName : null;
+                numberOfUpLevels++;
+            }
+            while (directory != null && numberOfUpLevels <= maxAllowedAboveDirectories);
+
+            return new List<string>();
+        }
+
+        private static (string sourceName, string replaceValue)? DetermineSourceNameReplaceValue(ICreationEffects creationEffects)
```
vlada-shubina, line 209: `sourceName` is not limited to project file only.
I would recommend to consider all file renames when replacing args.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,235 @@
...
+                return false;
+            }
+
+            if (newJsonPropertyValue == null)
+            {
+                Reporter.Error.WriteLine(string.Format(LocalizableStrings.PostAction_ModifyJson_Error_ArgumentNotConfigured, NewJsonPropertyValueArgument));
+                return false;
+            }
+
+            var sourceNameReplaceValue = DetermineSourceNameReplaceValue(creationEffects);
```
vlada-shubina, line 89: The logic of determining the `sourceName` is not fully correct.
I would recommend to consider replacing all file renames for this properties, or better create an additional configuration which file renames to consider.
This configuration should contain the list of source filenames, check the final names for those and apply renames to `newJsonPropertyName` and `newJsonPropertyValue`

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,235 @@
...
+                    return null;
+                }
+
+                node = node[property];
+            }
+
+            return node;
+        }
+
+        private static IReadOnlyList<string> FindFilesAtOrAbovePath(
```
vlada-shubina, line 173: since this method is anyway local, is there any use of it when `maxAllowedAboveDirectories` is not 1?

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
```diff
@@ -308,4 +308,20 @@
...
-</root>
\ No newline at end of file
+  <data name="PostAction_ModifyJson_Error_ArgumentNotConfigured" xml:space="preserve">
+    <value>{0} not configured.</value>
+  </data>
+  <data name="PostAction_ModifyJson_Error_MultipleJsonFiles" xml:space="preserve">
+    <value>More than one file found that matches {0}.</value>
+  </data>
+  <data name="PostAction_ModifyJson_Error_NoJsonFile" xml:space="preserve">
+    <value>Unable to find the json file in the solution</value>
```
vlada-shubina, line 318:
```suggestion
    <value>Unable to find the json file '{0}' in the solution.</value>
```

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
```diff
@@ -308,4 +308,20 @@
     <value>Run 'dotnet workload search' to search workloads available to be installed on your SDK.</value>
     <comment>{Locked="dotnet workload search"}</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="PostAction_ModifyJson_Error_ArgumentNotConfigured" xml:space="preserve">
+    <value>{0} not configured.</value>
+  </data>
+  <data name="PostAction_ModifyJson_Error_MultipleJsonFiles" xml:space="preserve">
+    <value>More than one file found that matches {0}.</value>
```
vlada-shubina, line 315:
```suggestion
    <value>More than one file found that matches '{0}'.</value>
```

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
```diff
@@ -308,4 +308,20 @@
     <value>Run 'dotnet workload search' to search workloads available to be installed on your SDK.</value>
     <comment>{Locked="dotnet workload search"}</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="PostAction_ModifyJson_Error_ArgumentNotConfigured" xml:space="preserve">
+    <value>{0} not configured.</value>
```
vlada-shubina, line 312:
```suggestion
    <value>Post action argument '{0}' is mandatory, but not configured.</value>
```

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
```diff
@@ -308,4 +308,20 @@
...
+  </data>
+  <data name="PostAction_ModifyJson_Error_MultipleJsonFiles" xml:space="preserve">
+    <value>More than one file found that matches {0}.</value>
+  </data>
+  <data name="PostAction_ModifyJson_Error_NoJsonFile" xml:space="preserve">
+    <value>Unable to find the json file in the solution</value>
+  </data>
+  <data name="PostAction_ModifyJson_Error_ParentPropertyPathInvalid" xml:space="preserve">
+    <value>Parent property path could not be traversed.</value>
```
vlada-shubina, line 321:
```suggestion
    <value>Parent property path could not be traversed.</value>
```
```suggestion
    <value>Parent property path '{0}' could not be traversed.</value>
```

{0} - configured property path

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
```diff
@@ -308,4 +308,20 @@
...
+  </data>
+  <data name="PostAction_ModifyJson_Succeeded" xml:space="preserve">
+    <value>Successfully modified {0}.</value>
```
vlada-shubina, line 324:
```suggestion
    <value>Successfully modified '{0}'.</value>
```

src/Tests/dotnet-new.Tests/PostActionTests.cs
```diff
@@ -872,5 +872,198 @@ public void ItCanCreateTemplate_WithRestore()
...
+        public void AddJsonProperty_Basic()
+        {
+            const string templateLocation = "PostActions/AddJsonProperty/Basic";
+            const string templateName = "TestAssets.PostActions.AddJsonProperty.Basic";
+
+            string home = CreateTemporaryFolder(folderName: "Home");
+            string workingDirectory = CreateTemporaryFolder();
+            InstallTestTemplate(templateLocation, _log, home, workingDirectory);
+
+            new DotnetNewCommand(_log, templateName)
```
vlada-shubina, line 886: Applies to other tests as well.

I would recommend to use snapshot test to assert the whole output as here: https://github.com/dotnet/sdk/blob/ec94dfe5a77140541bb1b87165de4372e64be223/src/Tests/dotnet-new.Tests/PostActionTests.Approval.cs#L32-L42

It is possible to "scrub" the output that may change.
If you'd like to assert both stdout and strerr you can use `FormatOutputStreams` extension: https://github.com/dotnet/sdk/blob/ec94dfe5a77140541bb1b87165de4372e64be223/src/Tests/dotnet-new.Tests/Extensions.cs#L11

vlada-shubina (commented):
Thank you for the contribution! This is indeed very valuable post action, and I can see further uses of it.

Added some comments inline.
For "sourceName" issue discussed in https://github.com/dotnet/templating/issues/5831 suggesting to configure the source filenames to be replaced, to cover more possible scenarios.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: I can do that.  Do I need to move the tests to another location as well ?

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: I'm afraid I don't follow here.

What I was actually trying to do here, was to make sure that, if the `newJsonPropertyName` or `newJsonPropertyValue` is set to the value of the `sourceName` property in the template.json, the `newJsonPropertyValue` or `newJsonPropertyName` is set to the value that is overriden by the -n or --name parameter that was used in the dotnet new command.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: Is this related to your other [comment](https://github.com/dotnet/sdk/pull/29880/files/227889fb859648b37f026cfd31c83e5eb8bcacaa#r1072102296) ?  So what you're actually saying is that I should look at every file to determine this source name & replacement value ?

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
vlada-shubina: yes, please

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
vlada-shubina: Yes, it is related. 

There is no proven way to determine what is `sourceName` having the information available for the post action. 
Relying that project file is changed by `sourceName` is not fully correct, though it might be so for many cases. But since this is a generic action that modifies arbitrary json file, this assumption cannot be taken as granted.

Therefore I would recommend to consider either replacing all file renames for this properties, or better create an additional configuration which file renames to consider. This configuration should contain the list of source filenames, check the final names for those and apply renames to `newJsonPropertyName` and `newJsonPropertyValue`. 

We can improve API, so source name is known for post action context, but that a separate problem to solve.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: Any suggestion on where I could put those unit-tests ?
I've moved the `DotnetAddJsonPropertyPostActionProcessor` from `dotnet/commands/dotnet-new`  to `Microsoft.TemplateEngine.Cli/PostActionProcessors` .  There are already other post-action processors there, so I figured I'll put the unittests in the same location as the tests of the other postaction processors, but it looks like the other postactions (like ChmodPostActionProcessor) does not have any tests ? :?

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
fgheysels: Removed this localizable string and moved the (suggested) value to the Localizablestrings.resx file in `Microsoft.TemplateEngine.Cli`

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
fgheysels: Removed this localizable string and moved the (suggested) value to the Localizablestrings.resx file in `Microsoft.TemplateEngine.Cli`

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
fgheysels: Removed this localizable string and moved the (suggested) value to the Localizablestrings.resx file in `Microsoft.TemplateEngine.Cli`

src/Cli/dotnet/commands/dotnet-new/LocalizableStrings.resx
fgheysels: Removed this localizable string and moved the (suggested) value to the Localizablestrings.resx file in `Microsoft.TemplateEngine.Cli`

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: > We can improve API, so source name is known for post action context, but that a separate problem to solve.

I think this would be the best and most clear solution where everyone can benefit from.  However, I do realise this might not be something that can be fixed on short term.

fgheysels: > Therefore I would recommend to consider either replacing all file renames for this properties, or better create an additional configuration which file renames to consider. This configuration should contain the list of source filenames, check the final names for those and apply renames to `newJsonPropertyName` and `newJsonPropertyValue`.
> 

So, If I understand correctly, you would go over all the FileChanges that are listed in the `CreationEffects` and take all those changes into account for applying renames to `newJsonProperty` and `newJsonPropertyValue`?
I would assume though that most files will not be renamed.  Only files that contain the `sourceName` 'magic value' will be renamed.
Or you mean that I should not just assume that the project file would contain the `sourceName` value ?  It looks to me that this file has the most change of being renamed, but indeed, I can take into consideration all files.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: (There are some build errors right now because I need to move the unit-tests :) )

maridematte: It seems like the other post-process do not have tests. I have opened a issue about it so we can add them later (https://github.com/dotnet/templating/issues/6004). For now, you can add them to `sdk\src\Tests\Microsoft.TemplateEngine.Cli.UnitTests` .

maridematte: I have been corrected, the tests for those post actions are in `src/Tests/dotnet-new.Tests/PostActionTests.cs`, so you can add tests there.

fgheysels: These are integration-tests, but not unit-tests.
I do have integration-tests and unit-tests for this post-action, so for the moment, I'll put the unit-tests in `Microsoft.TemplateEngine.Cli.UnitTests`.

src/Tests/dotnet-new.Tests/PostActionTests.cs
fgheysels: TBH, I don't really see an advantage of this approach.  It's somewhat bloated and it is not immediately visible what you is being tested imho.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
vlada-shubina: > So, If I understand correctly, you would go over all the FileChanges that are listed in the CreationEffects and take all those changes into account for applying renames to newJsonProperty and newJsonPropertyValue?
I would assume though that most files will not be renamed. Only files that contain the sourceName 'magic value' will be renamed.

The files can be quite frequently renamed:
- `sourceName` as one of the case, likely most frequent
- `fileRename` configuration
- `source` - `rename` configuration

Assuming that project name is `sourceName` is not correct: every occurrence of `sourceName` is replaced, and project name may be different from it.

> But additionally, what is a file has the name sourceNameSomethingElse. This will be replaced to MyReplaceValueSomethingElse. We also need to be able to cope with that.

That's a valid concern, but mostly it's up to template author to ensure that's not happening. The replacement value for `sourceName` might be something very unique: as `MyValueForSourceNameThatCannotOccurSomewhereElse.1`, then accidental replacements are pretty safe.

src/Tests/dotnet-new.Tests/PostActionTests.cs
vlada-shubina: it's okay not to use it - we found it useful as it blocks unexpected output if the feature is broken in a  way.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
vlada-shubina: That will be correct place for unit tests

src/Tests/dotnet-new.Tests/PostActionTests.cs
fgheysels: Then I would opt to not use it.  It is then also consistent with the other integration-tests for other Post Actions that are in the same file.

src/Cli/dotnet/commands/dotnet-new/PostActions/DotnetAddJsonPropertyPostActionProcessor.cs
fgheysels: I've been thinkering about this for a while, and I'm actually convinced that the only proper way to correctly handle this, is to make sure that the `sourceName` and the value that should eventually be used to replace  `sourceName` should actually be available in the `PostActionProcessor`  instance.

I think this information can be part of the `ICreationEffects2` interface.

maridematte: Hey @fgheysels, just checking if you are continue working on this issue! We are very interested in this solution and are willing to help once we have more time. Let us know the situation so we can proceed in the best way :)

fgheysels: Hi @maridematte, thanks for reaching out.

I'm actually struggling a bit with how we could make sure that we take into account the `sourceName` replacement in the JSON property that we could add.

Let's say for instance that we want to add a JSON property to an existing JSON file, but the name of that property or the value of that property should be 'dynamic'. 
The templates support the `sourceName` property and when instantiating the template, the template-files that contain the value of `sourceName` are being replaced by the string that is specified by the `--name`  argument.
I want to have this behavior when specifying the new JSON property or JSON value as well.  Suppose:

```json
"sourceName":  "TheMagicWord"

"postActions": [
    {
        "args": {
               "newJsonPropertyName":  "TheMagicWord"
              ...
        }
    }
]
```
When instantiating the template and specifying `--name MyProjectName`, I would like to have that the new property that must be added to the json file is not `TheMagicWord` , but is `MyProjectName`.

For now, I've implemented this by checking the `creationEffects`  on the project-file as this is afaik the only way to get to know the original `sourceName` value and the value of the `--name` argument (see [here](https://github.com/dotnet/sdk/blob/497187db31346639dfc1e390256e64dbd4b3a7e7/src/Cli/Microsoft.TemplateEngine.Cli/PostActionProcessors/AddJsonPropertyPostActionProcessor.cs#L204).   However, @vlada-shubina  has a relevant remark on this [here](https://github.com/dotnet/sdk/pull/29880/#discussion_r1072100716), as there is indeed an assumption there that the name of the project file in the template is actually the same as the `sourceName` value.  It is true that we should not assume that, and that I need to look at other files as well.
Next to that, I also assume that the project file will not have a suffix or prefix to the `sourceName`.  (It can actually be `TheMagicWordSomeProject.csproj` for instance.
So actually, it would be better if I could somehow calculate the difference between the `SourceRelativePath` and `TargetRelativePath` of the creation-effects, but I don't see how this can be done.

As said in my comment just above yours, I think the best solution would be to extend the `ICreationEffects2` interface with the value of `sourceName` and the value of the `--name` argument that might have been given.

vlada-shubina: It would be good to think about scenarios other then `sourceName` here, as this is not only source of replacement.
I think https://github.com/dotnet/templating/issues/1600 would be beneficial in general if extended to all post action arguments, however unfortunately it's not on the radar yet. 

If this is implemented the example below:
```
  "sourceName": "MyProject",
  "postActions": [
    {
      "description": "Modify JSON file",
      "manualInstructions": [ { "text": "Modify the JSON file manually." } ],
      "args": {
        "jsonFileName": "deployment.template.json",
        "parentPropertyPath": "modulesContent:$edgeAgent:properties.desired:modules",
        "newJsonPropertyName": "MyProject",
        "newJsonPropertyValue": "{\"version\":\"1.0.0\", \"type\":\"docker\", \"settings\":{\"image\":\"${MODULEDIR<../MyProject>}\"}}"
      },
      "actionId": "695A3659-EB40-4FF5-A6A6-C9C4E629FCB0",
      "continueOnError": true
    }

```
will result in having `MyProject` replaced with whatever `sourceName` is in args collection and post action won't need to care about this.

Another suggestion might be to extend interface of post actions with actual replacements or operations doing replacements but it might be too heavy.

Having just `sourceName` in interface or replacing just `sourceName` might be an intermediate solution but we need to aim for full support further on.

fgheysels: I'm just wondering now how to proceed with this.
It seems like we're stuck with this for now, as we're waiting on some other functionality as mentioned by @vlada-shubina to be able to fully complete the implementation of this.

vlada-shubina: We are currently discussing how to proceed with it. 
One option on the table is to implement https://github.com/dotnet/templating/issues/1600 in near future (1-2 months). 
This PR may not wait for it, it will start to work for replacements once https://github.com/dotnet/templating/issues/1600 is merged or changed if needed then.

We are definitely interested in merging this implementation even without possibility to replace source name or other replacements. I know unfortunately it doesn't work for your case now, but we see value even without replacement for other uses.

Please allow us more time to discuss it.

YuliiaKovalova: Hi @fgheysels,

We have prioritized the work that has to be done in order to unblock you.
It is expected to implement within the next 1-2 months. You can track the progress in this ticket https://github.com/dotnet/templating/issues/1600

vlada-shubina:
@fgheysels could you please check https://github.com/dotnet/templating/pull/6312 - the e2e test and if the result is expected for your case? I based it on example from this PR.

If so, please consider to adapt this PR for the updated version of args in post action with file renames already applied. 

We should merge that PR within a week, and once done you can rebase to latest main to get those changes in for test in your branch.

fgheysels:
> @fgheysels could you please check [dotnet/templating#6312](https://github.com/dotnet/templating/pull/6312) - the e2e test and if the result is expected for your case? I based it on example from this PR.
> 
> If so, please consider to adapt this PR for the updated version of args in post action with file renames already applied.
> 
> We should merge that PR within a week, and once done you can rebase to latest main to get those changes in for test in your branch.

Hi @vlada-shubina , If I understand it correctly, 2 new arguments have been added to the configuration.  With one of these arguments you can specify that the 'renames' should be applied on other arguments in that configuration.
I think that will cover my scenario.

I have added some remarks to the PR though :)

YuliiaKovalova:
Could you add post-action info to templating repo?
https://github.com/dotnet/templating/wiki/Post-Action-Registry

src/Cli/Microsoft.TemplateEngine.Cli/PostActionProcessors/AddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,191 @@
...
+            action.Args.TryGetValue(NewJsonPropertyNameArgument, out string? newJsonPropertyName);
+            action.Args.TryGetValue(NewJsonPropertyValueArgument, out string? newJsonPropertyValue);
+
+            if (newJsonPropertyName == null)
+            {
+                Reporter.Error.WriteLine(string.Format(LocalizableStrings.PostAction_ModifyJson_Error_ArgumentNotConfigured, NewJsonPropertyNameArgument));
+                return false;
+            }
+
+            if (newJsonPropertyValue == null)
```
YuliiaKovalova, line 76: can we move these validation checks to a separate method in order to make this method easier to read?

fgheysels: In most cases, I'm in favor of doing this, however, in this case, I 'm not sure if it will improve readability.
I've made an attempt though in https://github.com/dotnet/sdk/pull/29880/commits/cb9979f497ba683d25307fbcdd3c717ba94dafe3

src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
```diff
@@ -15,9 +15,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="FakeItEasy" />
-    <PackageReference Include="Verify.XUnit" />
-    <PackageReference Include="Verify.DiffPlex" />
+    <PackageReference Include="FakeItEasy" Version="7.3.1" />
+    <PackageReference Include="Moq" Version="4.8.2" />
```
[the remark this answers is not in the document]
YuliiaKovalova: @vlada-shubina , I suspect it can be the undesired thing to add a new testing package, can't it?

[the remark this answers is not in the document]
vlada-shubina: Both are already used. Please use `$(MoqPackageVersion)` instead of version (it is defined here https://github.com/dotnet/sdk/blob/8546a2ea939238c0dbf7285f0a863794af73534f/eng/Versions.props#L191) and keep `FakeItEasy` without version (it is managed by dependabot). 
```suggestion
    <PackageReference Include="Moq" Version="$(MoqPackageVersion)" />
    <PackageReference Include="FakeItEasy" />
```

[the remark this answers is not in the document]
fgheysels: Done.

YuliiaKovalova:
looks good to me!

fgheysels:
> Could you add post-action info to templating repo? https://github.com/dotnet/templating/wiki/Post-Action-Registry

So, that's a PR In another repository ?
Wil do!

vlada-shubina (approved):
Thank you for the contribution!

src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
```diff
@@ -16,8 +16,9 @@
 
   <ItemGroup>
     <PackageReference Include="FakeItEasy" />
-    <PackageReference Include="Verify.XUnit" />
-    <PackageReference Include="Verify.DiffPlex" />
+    <PackageReference Include="Moq" Version="$(MoqPackageVersion)" />
+    <PackageReference Include="Verify.XUnit" Version="19.8.0" />
+    <PackageReference Include="Verify.DiffPlex" Version="2.1.0" />
```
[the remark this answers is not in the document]
fgheysels: Done!

src/Cli/Microsoft.TemplateEngine.Cli/PostActionProcessors/AddJsonPropertyPostActionProcessor.cs
```diff
@@ -0,0 +1,224 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
```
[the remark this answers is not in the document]
YuliiaKovalova: It's an unlucky coincidence - the header has been updated recently https://github.com/dotnet/sdk/pull/32141/files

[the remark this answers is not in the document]
fgheysels: Ah, thanks.  I would have seen it if I had merged with upstream again :)

fgheysels:
> > Could you add post-action info to templating repo? https://github.com/dotnet/templating/wiki/Post-Action-Registry
> 
> So, that's a PR In another repository ? Wil do!

I've cloned the wiki locally and added some documentation for the new postaction.
Afterwards, I wanted to push my branch, but got the message that I'm not allowed to push my branch to the remote.  So, I created a fork and pushed the changes to my fork.
Now, I want to create a PR for the wiki changes, but GitHub is a bit confused as it says that there is nothing to compare (commit histories are entirely different) although it also shows the changes I've made in a diff window.  Unfortunately, it doesn't allow me to create a PR:

image

So, any help here is appreciated :)

YuliiaKovalova:
> > > Could you add post-action info to templating repo? https://github.com/dotnet/templating/wiki/Post-Action-Registry
> > 
> > 
> > So, that's a PR In another repository ? Wil do!
> 
> I've cloned the wiki locally and added some documentation for the new postaction. Afterwards, I wanted to push my branch, but got the message that I'm not allowed to push my branch to the remote. So, I created a fork and pushed the changes to my fork. Now, I want to create a PR for the wiki changes, but GitHub is a bit confused as it says that there is nothing to compare (commit histories are entirely different) although it also shows the changes I've made in a diff window. Unfortunately, it doesn't allow me to create a PR:
> 
> image
> 
> So, any help here is appreciated :)

You need to fork the repo , commit changes and create PR from the fork to dotnet/templating repo.

fgheysels:
> > > > Could you add post-action info to templating repo? https://github.com/dotnet/templating/wiki/Post-Action-Registry
> > > 
> > > 
> > > So, that's a PR In another repository ? Wil do!
> > 
> > 
> > I've cloned the wiki locally and added some documentation for the new postaction. Afterwards, I wanted to push my branch, but got the message that I'm not allowed to push my branch to the remote. So, I created a fork and pushed the changes to my fork. Now, I want to create a PR for the wiki changes, but GitHub is a bit confused as it says that there is nothing to compare (commit histories are entirely different) although it also shows the changes I've made in a diff window. Unfortunately, it doesn't allow me to create a PR:
> > image
> > So, any help here is appreciated :)
> 
> You need to fork the repo , commit changes and create PR from the fork to dotnet/templating repo.

I've done that.  However, when you fork a repo, the wiki is not forked with it.
So, I've followed [these steps](https://stackoverflow.com/a/56480628/55774).  I've now the original wiki in my fork, and updated the docs in my fork.

However, I cannot create a PR as the changes to the wiki are not taken into account when creating a PR.  The wiki is a separate repository, so that makes sense.  There doesn't seem to be an option to create a PR for the wiki repo.

fgheysels:
> 

@dotnet-policy-service agree

src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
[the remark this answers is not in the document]
fgheysels: Yes, I had to verify something with my employer.  Done now.

YuliiaKovalova:
> > > > > Could you add post-action info to templating repo? https://github.com/dotnet/templating/wiki/Post-Action-Registry
> > > > 
> > > > 
> > > > So, that's a PR In another repository ? Wil do!
> > > 
> > > 
> > > I've cloned the wiki locally and added some documentation for the new postaction. Afterwards, I wanted to push my branch, but got the message that I'm not allowed to push my branch to the remote. So, I created a fork and pushed the changes to my fork. Now, I want to create a PR for the wiki changes, but GitHub is a bit confused as it says that there is nothing to compare (commit histories are entirely different) although it also shows the changes I've made in a diff window. Unfortunately, it doesn't allow me to create a PR:
> > > image
> > > So, any help here is appreciated :)
> > 
> > 
> > You need to fork the repo , commit changes and create PR from the fork to dotnet/templating repo.
> 
> I've done that. However, when you fork a repo, the wiki is not forked with it. So, I've followed [these steps](https://stackoverflow.com/a/56480628/55774). I've now the original wiki in my fork, and updated the docs in my fork.
> 
> However, I cannot create a PR as the changes to the wiki are not taken into account when creating a PR. The wiki is a separate repository, so that makes sense. There doesn't seem to be an option to create a PR for the wiki repo.

Please point me to the changes in your fork and I will transfer them to our repo.
Thank you for the contribution!

PR #29899, opened 2023-01-12
Ensure Blazor startup succeeds

TanayParikh:
Check for 200-OK for `/`


mkArtakMSFT:
@tmat I think this is what you were looking for, weren't you?

tmat:
As long as it exercises enough of the server code to be meaningful.

mkArtakMSFT:
> As long as it exercises enough of the server code to be meaningful.

It just makes sure that the server has started serving content and that a Get request to the root will return 200 OK code.

mkArtakMSFT (approved):

PR #29984, opened 2023-01-18
Use better AOT defaults in Web SDK

### LINKED ISSUE #29896: Disable inclusion of IIS hosting assets (ANCM & web.config) in publish output when PublishNativeAot=true
When publishing for Windows today, the web SDK includes assets required to run the project on IIS, namely the ANCM module (*aspnetcorev2_inprocess.dll*) and a *web.config* file. As part of the [work to support native AOT publish for ASP.NET Core](https://github.com/dotnet/aspnetcore/issues/45910), we should ensure that these assets are not included in the application output when the `<PublishAot>` property is set to true in the project file, as IIS hosting does not support native AOT published ASP.NET Core apps, i.e. set the default value of `<PublishIISAssets>` to false when `<PublishAot>` is true.

### LINKED ISSUE #30059: Microsoft.NET.Sdk.Web should default to TrimMode=full
Today, we are defaulting `TrimMode=partial` in `Microsoft.NET.Sdk.Web`:

https://github.com/dotnet/sdk/blob/67b59150e3407428c4114c8c39ade5170b98c3ad/src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets#L89-L90

```xml
    <!-- If TrimMode has not already been set, default to partial trimming, as AspNet is not currently fully trim-compatible -->
    <TrimMode Condition="'$(PublishTrimmed)' == 'true' and '$(TrimMode)' == ''">partial</TrimMode>
```

In the "api" template being added in https://github.com/dotnet/aspnetcore/pull/46064, we are explicitly setting TrimMode=full.

Instead, we should do one of the following:

1. Remove the above code, and just let the `TrimMode=full` default from the runtime be applied to Microsoft.NET.Sdk.Web projects.
2. Change the default to `partial` only if we aren't also publishing for NativeAOT. Basically adding `and '$(PublishAot)' != 'true'` above.

My goal would be we do (1). For places in ASP.NET that aren't fully trim-compatible, the app developer should get a warning (even from within ASP.NET code) to tell them what are things that aren't going to work.

We may consider just doing (2) in .NET 8, until things like MVC and Razor pages work correctly in fully trimmed apps.

cc @JamesNK @DamianEdwards @davidfowl

halter73:
This PR provides better defaults for the TrimMode and PublishIISAssets properties when the PublishAot property is set to true.

Once this is merged, we should be able to follow up by removing [these properties](https://github.com/dotnet/aspnetcore/blob/e5238763bddd7100823751c4a4ae0220d78160aa/src/ProjectTemplates/Web.ProjectTemplates/Api-CSharp.csproj.in#L12-L13) from the AOT version of the new "api" project template.

Fixes dotnet/sdk#30059
Fixes dotnet/sdk#29896

### DIFF
--- a/src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
+++ b/src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
@@ -20,7 +20,7 @@ namespace Microsoft.NET.Sdk.Web.Tests
         {
         }
 
-        [Theory()]
+        [Theory]
         [MemberData(nameof(SupportedTfms))]
         public void TrimmingOptions_Are_Defaulted_Correctly_On_Trimmed_Apps(string targetFramework)
         {
@@ -29,11 +29,16 @@ namespace Microsoft.NET.Sdk.Web.Tests
 
             var testProject = CreateTestProjectForILLinkTesting(targetFramework, projectName);
             testProject.AdditionalProperties["PublishTrimmed"] = "true";
+            testProject.AdditionalProperties["RuntimeIdentifier"] = rid;
+            testProject.PropertiesToRecord.Add("TrimMode");
 
             var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: projectName + targetFramework);
 
             var publishCommand = new PublishCommand(Log, Path.Combine(testAsset.TestRoot, testProject.Name));
-            publishCommand.Execute($"/p:RuntimeIdentifier={rid}").Should().Pass();
+            publishCommand.Execute().Should().Pass();
+
+            var buildProperties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework);
+            buildProperties["TrimMode"].Should().Be("partial");
 
             string outputDirectory = publishCommand.GetOutputDirectory(targetFramework: targetFramework, runtimeIdentifier: rid).FullName;
             string runtimeConfigFile = Path.Combine(outputDirectory, $"{projectName}.runtimeconfig.json");
@@ -46,7 +51,7 @@ namespace Microsoft.NET.Sdk.Web.Tests
                     .Should().BeTrue();
         }
 
-        [Theory()]
+        [Theory]
         [MemberData(nameof(SupportedTfms))]
         public void TrimmingOptions_Are_Defaulted_Correctly_On_Aot_Apps(string targetFramework)
         {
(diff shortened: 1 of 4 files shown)


JamesNK (commented):
LGTM

However, I'm not an msbuild/SDK person. I recommend a second opinion.

JamesNK (approved):

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
```diff
@@ -36,6 +36,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
              And '$(RuntimeIdentifier)' != ''
              And !$(RuntimeIdentifier.StartsWith('win'))
              And '$(PublishIISAssets)' == ''">false</PublishIISAssets>
+    <PublishIISAssets Condition="'$(PublishAot)' == 'true' And '$(PublishIISAssets)' == ''">false</PublishIISAssets>
```
mitchdenny, line 39: nit: ```and```

mitchdenny (approved):

vijayrkn (approved):

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
eerhardt, line 39:
```suggestion
    <PublishIISAssets Condition="'$(PublishIISAssets)' == '' and '$(PublishAot)' == 'true'">false</PublishIISAssets>
```

It reads better to put the "has this already been set?" check first.

eerhardt (commented):
Can we write a test or two that ensures these are set correctly?

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
```diff
@@ -87,7 +88,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <EFSQLScriptsFolderName Condition="$(EFSQLScriptsFolderName) == ''">EFSQLScripts</EFSQLScriptsFolderName>
 
     <!-- If TrimMode has not already been set, default to partial trimming, as AspNet is not currently fully trim-compatible -->
-    <TrimMode Condition="'$(PublishTrimmed)' == 'true' and '$(TrimMode)' == ''">partial</TrimMode>
+    <!-- We allow the full trimming default with PublishAot enabled, because the parts that are not trim-compatible are not aot-compatible either -->
+    <TrimMode Condition="'$(PublishTrimmed)' == 'true' and '$(PubishAot)' != 'true' and '$(TrimMode)' == ''">partial</TrimMode>
```
halter73, line 92:
```suggestion
    <TrimMode Condition="'$(TrimMode)' == '' and '$(PublishTrimmed)' == 'true' and '$(PubishAot)' != 'true'">partial</TrimMode>
```

src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
```diff
@@ -55,16 +60,26 @@ public void TrimmingOptions_Are_Defaulted_Correctly_On_Aot_Apps(string targetFra
...
+            testProject.PropertiesToRecord.Add("PublishIISAssets");
 
             var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: projectName + targetFramework);
             var publishCommand = new PublishCommand(Log, Path.Combine(testAsset.TestRoot, testProject.Name));
-            publishCommand.Execute($"/p:RuntimeIdentifier={rid}").Should().Pass();
+            publishCommand.Execute().Should().Pass();
+
+            var buildProperties = testProject.GetPropertyValues(testAsset.TestRoot, targetFramework);
+            buildProperties["PublishTrimmed"].Should().Be("true");
+            buildProperties["TrimMode"].Should().Be("");
```
halter73, line 74: This verifies we don't set TrimMode to "partial" like we do in the PublishTrimmed-only case. Does anyone know of a better way to verify we get "full" trimming?

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
```diff
@@ -87,7 +88,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <EFSQLScriptsFolderName Condition="$(EFSQLScriptsFolderName) == ''">EFSQLScripts</EFSQLScriptsFolderName>
 
     <!-- If TrimMode has not already been set, default to partial trimming, as AspNet is not currently fully trim-compatible -->
-    <TrimMode Condition="'$(PublishTrimmed)' == 'true' and '$(TrimMode)' == ''">partial</TrimMode>
+    <!-- We allow the full trimming default with PublishAot enabled, because the parts that are not trim-compatible are not aot-compatible either -->
+    <TrimMode Condition="'$(TrimMode)' == '' and '$(PublishTrimmed)' == 'true' and '$(PublishAot)' != 'true'">partial</TrimMode>
```
davidfowl, line 92: We should think about defaulting to full with `PublishTrimmed` true as well 

cc @DamianEdwards @eerhardt

DamianEdwards: yeah this came up in the original issue and discussion but it seems a bit premature right now given the state of trimmability across all of ASP.NET Core. Not ruling it out for .NET 8 though if we think we get far enough to make it the better default.

davidfowl: Can we track it in an issue and punt it to the next release if we decide?

halter73: I reopened https://github.com/dotnet/sdk/issues/30059 and put it in the Goldilocks project for tracking.

PR #30038, opened 2023-01-20
[Breaking] `RuntimeIdentifier` No Longer Implies `SelfContained` Apps by Default

nagilson:
# `RuntimeIdentifier` No Longer Implies `SelfContained` Apps by Default
- RIDS no longer imply Self-Contained for TFM >= 8.0
- Warns for apps that didn't set SC but set RIDs in TFM < 8 that they need to set SC explicitly to get the same behavior with TFM 8.0. Warning will go away after SC is set, either to true or false.
- PublishTrimmed, PublishSingleFile, and PublishAot will keep the old behavior of implying SC with 8+ TFMs, though they will only do so for publish which is a breaking change. PublishReadyToRun will not maintain this behavior and you will need to specify SelfContained explicitly to use it with PublishReadyToRun in TFMs above 7.

- Remove the old error that occurred when `-r` given without `-self-contained` as it did not catch property in project scenarios, and it's redundant with the breaking change warning. We also have a more understandable default now, (setting one property does not add another that is not necessary.)
- Update IL Link Tests and Publish* tests that expected the old behavior. Talked with that team, we don't want to add an implicit `SelfContained` there even if it will break customers as it limits the design potential in the future for these properties.
- Talked to Blazor WASM who would be broken by this, they have created work items to add `SelfContained` so customers aren't broken. I modified their tests as well to pass by adding `SelfContained`. 

This is the current VS behavior and we designed this in a NET 8 breaking changes list. It will help us align with VS behavior.

# Tested Behavior:
* RidNoLongerImpliesSelfContainedForTFM8+
* SelfContainedPlusRuntimeIdentifierDoesNotTriggerEightTFMFDDBreakingChangeWarning

# Questions:
- Why does `TargetFrameworkInference` sometimes not run? I accounted for that, but don't understand the contexts.


sbomer (approved):
ILLink test changes LGTM. We could probably clean up the tests by adding SelfContained to the project file, but no need to do so in this change.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -229,7 +230,7 @@ public void ILLink_respects_global_TrimMode(string targetFramework, string trimM
                 .WithProjectChanges(project => AddRootDescriptor(project, $"{referenceProjectName}.xml"));
 
             var publishCommand = new PublishCommand(testAsset);
-            publishCommand.Execute($"/p:RuntimeIdentifier={rid}", "/p:PublishTrimmed=true").Should().Pass();
+            publishCommand.Execute($"/p:RuntimeIdentifier={rid}", "/p:SelfContained=true", "/p:SelfContained=true", "/p:PublishTrimmed=true").Should().Pass();
```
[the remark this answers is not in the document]
nagilson: Whoops, sorry. I missed that mistake. Thanks for catching it.

richlander:
Is this still blocked?

nagilson:
Nope, our meeting from last week cleared up any blockers on this one. It's in the pipeline to take care of soon :)

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToRunILLink.cs
```diff
@@ -1555,15 +1558,17 @@ public void Build_respects_IsTrimmable_property(string targetFramework, bool isE
...
             var testProject = CreateTestProjectWithAnalysisWarnings(targetFramework, projectName, isExe);
             testProject.AdditionalProperties["IsTrimmable"] = "true";
+            var rid = EnvironmentInfo.GetCompatibleRid(targetFramework);
+            testProject.AdditionalProperties["RuntimeIdentifier"] = rid;
             var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: targetFramework + isExe);
 
             var buildCommand = new BuildCommand(testAsset);
             // IsTrimmable enables analysis warnings during build
-            buildCommand.Execute()
+            buildCommand.Execute("/bl")
```
nagilson, line 1567:
```suggestion
            buildCommand.Execute()
```

src/Assets/TestProjects/TrimmedAppWithReferences/App/App.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>$(CurrentTargetFramework)</TargetFramework>
     <PublishTrimmed>true</PublishTrimmed>
     <EnableTrimAnalyzer>false</EnableTrimAnalyzer>
+    <SelfContained>True</SelfContained>
```
nagilson, line 8:
```suggestion

```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -149,7 +163,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <_SelfContainedWasSpecified Condition="'$(SelfContained)' != ''">true</_SelfContainedWasSpecified>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and '$(HasRuntimeOutput)' == 'true'">
-    <SelfContained Condition="'$(SelfContained)' == '' and '$(RuntimeIdentifier)' != ''">true</SelfContained>
+    <!-- Breaking change in .NET 8: Projects with 8.0+ TFMS will no longer have RuntimeIdentifier imply SelfContained. Note that PublishReadyToRun will imply SelfContained in these versions. -->
+    <SelfContained Condition="'$(SelfContained)' == '' and '$(RuntimeIdentifier)' != '' and
+                              '$(_TargetFrameworkVersionWithoutV)' &lt; '8.0'">true</SelfContained>
```
nagilson, line 168: I dont think we need to check if _TargetFrameworkVersionWithoutV is empty by this point

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
```diff
@@ -248,11 +235,13 @@ public void TestGlobalPropertyFlowInSolution(bool passSelfContained, bool passRu
             }
         }
 
-        private static void ValidateProperties(TestAsset testAsset, TestProject testProject, bool expectSelfContained, bool expectRuntimeIdentifier, string targetFramework = null, string expectedRuntimeIdentifier = "")
+        private static void ValidateProperties(TestAsset testAsset, TestProject testProject, bool expectSelfContained, bool expectRuntimeIdentifier, string thisTargetFramework = null, string expectedRuntimeIdentifier = "")
```
nagilson, line 238: The parameter rename here is because the IDE was getting confused when doing type checking on which targetFramework property was which.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: Eh better be safe than sorry

dsplaisted: I'd suggest switching to `$([MSBuild]::VersionLessThan($(_TargetFrameworkVersionWithoutV), '8.0'))`.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -383,6 +385,75 @@ public void It_can_publish_runtime_specific_apps_with_library_dependencies_self_
...
+                    var ns = project.Root.Name.Namespace;
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifier", runtimeIdentifier));
+                    propertyGroup.Add(new XElement(ns + "SelfContained", defineSelfContained ? "true" : ""));
+                });
+
+            var buildCommand = new DotnetBuildCommand(Log);
+            var commandResult = buildCommand
+                .WithWorkingDirectory(testAsset.Path)
+                .Execute("/bl");
```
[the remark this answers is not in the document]
dsplaisted: I think we should avoid producing binlogs during CI test runs, they just take up extra space (and it would be a lot of extra space if all the tests were doing it).  When I'm debugging locally I'll add the `/bl` parameter if needed, and sometimes I'll accidentally check it in.  But I think we should avoid doing it intentionally and remove it when we notice it.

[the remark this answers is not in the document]
nagilson: Thanks for explaining. I agree.

baronfel (approved):
Approved the changes to the containers files on behalf of @dotnet/sdk-container-builds-maintainers - this change is actually something we're really excited about, because this old default meant that most folks were making large images by default!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
nagilson: Good point, I didnt know about that. Though we need to keep the empty string checks because that function will fail with an empty string

nagilson:
@dsplaisted @joeloff  @richlander  PTAL at reviewing when possible. Id like to try to push this into preview 4 if it looks good and I think theres really only one or two days left for that. Thank you!

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
```diff
@@ -86,7 +86,7 @@ public void TestGlobalPropertyFlowToLibrary(bool passSelfContained, bool passRun
 
             bool buildingSelfContained = passSelfContained || passRuntimeIdentifier;
```
dsplaisted, line 87: This should probably renamed or the logic reworked, since passing a RuntimeIdentifier doesn't imply that the project will be self-contained.  Same with other tests here.

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
```diff
@@ -182,18 +165,22 @@ public void TestGlobalPropertyFlowToMultitargetedProject(bool passSelfContained,
                 project.Root.Element("PropertyGroup").Add(XElement.Parse(@"<OutputType Condition=""'$(TargetFramework)' == 'net6.0'"">Library</OutputType>"));
             });
 
-            var testAsset = Build(passSelfContained, passRuntimeIdentifier, identifier: passSelfContained.ToString() + "_" + passRuntimeIdentifier);
+            string identifier = passSelfContained.ToString() + "_" + passRuntimeIdentifier;
 
-            bool buildingSelfContained = passSelfContained || passRuntimeIdentifier;
+                var testAsset = Build(passSelfContained, passRuntimeIdentifier, identifier: identifier);
```
dsplaisted, line 170: Looks like the indentation might be off here.

src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
```diff
@@ -271,6 +271,58 @@ public void PublishRuntimeIdentifierOverridesUseCurrentRuntime()
...
+        [InlineData("PublishReadyToRun", true)]
+        [InlineData("PublishSingleFile", true)]
+        [InlineData("PublishTrimmed", true)]
+        [InlineData("PublishAot", true)]
+        [InlineData("PublishReadyToRun", false)]
+        [InlineData("PublishSingleFile", false)]
+        [InlineData("PublishTrimmed", false)]
+        public void SomePublishPropertiesInferSelfContained(string property, bool useFrameworkDependentDefaultTargetFramework)
+        {
+            // Note: there is a bug with PublishAot I think where this test will fail for Aot if the testname is too long. Do not make it longer.
```
dsplaisted, line 284: In that case you can set the TestProject's Name to something shorter.

dsplaisted (approved):
Generally looks good.  I didn't look at all the test changes for things like global property flow and referencing Exes in as much detail as would be nice.

It sounds like we want to fix part of the breaking change for blazor-wasm, I think we should probably do that in this PR rather than working around it in the tests.

All the test changes highlight that this is a pretty high-impact change.  It helps that it's restricted to projects targeting .NET 8 or higher, but we probably need to make sure to get the word out about this and expect some impact from the change. @richlander @baronfel

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/CreateNewImageTests.cs
```diff
@@ -138,13 +138,13 @@ public void Tasks_EndToEnd_With_EnvironmentVariable_Validation()
 
         File.WriteAllText(Path.Combine(newProjectDir.FullName, "Program.cs"), $"Console.Write(Environment.GetEnvironmentVariable(\"GoodEnvVar\"));");
 
-        new DotnetCommand(_testOutput, "build", "--configuration", "release", "/p:runtimeidentifier=linux-x64")
+        new DotnetCommand(_testOutput, "build", "--configuration", "release", "/p:runtimeidentifier=linux-x64", $"/p:RuntimeFrameworkVersion=8.0.0-preview.3.23174.8")
```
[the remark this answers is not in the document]
nagilson: I agree, I believe @baronfel did this as a temporary patch to get the container tests to pass as there is no structure in place to make those tests pass while the preview 4 and preview 5 container bits havent been built.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/DockerRegistryManager.cs
```diff
@@ -10,12 +10,15 @@ namespace Microsoft.NET.Build.Containers.IntegrationTests;
...
+    public const string RuntimeBaseImage = "dotnet/runtime";
+    public const string AspNetBaseImage = "dotnet/aspnet";
     public const string BaseImageSource = "mcr.microsoft.com/";
     public const string Net6ImageTag = "6.0";
     public const string Net7ImageTag = "7.0";
+    public const string Net8PreviewImageTag = "8.0-preview";
     public const string LocalRegistry = "localhost:5010";
-    public const string FullyQualifiedBaseImageDefault = $"{BaseImageSource}{BaseImage}:{Net6ImageTag}";
+    public const string FullyQualifiedBaseImageDefault = $"{BaseImageSource}{RuntimeBaseImage}:{Net8PreviewImageTag}";
+    public const string FullyQualifiedBaseImageAspNet = $"{BaseImageSource}{AspNetBaseImage}:{Net8PreviewImageTag}";
```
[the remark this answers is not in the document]
nagilson: These are from https://github.com/dotnet/sdk/pull/30038/commits/72b31a43e8d1426380cac002fbff1178ff95d00a which was needed to get the tests to pass irrespective of this change I _think_, should we just put that in a separate PR @baronfel ?

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
```diff
@@ -383,6 +385,75 @@ public void It_can_publish_runtime_specific_apps_with_library_dependencies_self_
...
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("HelloWorld", identifier: targetFramework + defineSelfContained.ToString())
+                .WithSource()
+                .WithTargetFramework(targetFramework)
+                .WithProjectChanges(project =>
+                {
+                    var ns = project.Root.Name.Namespace;
+                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
+                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifier", runtimeIdentifier));
+                    propertyGroup.Add(new XElement(ns + "SelfContained", defineSelfContained ? "true" : ""));
```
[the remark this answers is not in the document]
nagilson: How would we best do that? The project change needs to be within a lambda expression. I dont think there is a way to wrap an if statement inside of a lambda expression in particular.. and I am not aware of any C# language feature where we could pass a statement into the lambda thats a no-op lest we create a function for it such as `defineSelfContained ? propertyGroup.Add(...) : doNothing()`?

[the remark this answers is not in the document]
dsplaisted: Unless I'm very confused you can have an if statement inside here no problem:

```c#
                    var ns = project.Root.Name.Namespace;
                    var propertyGroup = project.Root.Elements(ns + "PropertyGroup").First();
                    propertyGroup.Add(new XElement(ns + "RuntimeIdentifier", runtimeIdentifier));
                    if (defineSelfContained)
                    {
                        propertyGroup.Add(new XElement(ns + "SelfContained", "true"));
                    }
```

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/CreateNewImageTests.cs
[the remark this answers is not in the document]
baronfel: Echoing this - the container tests are in kind of an odd space (at least for the preview timeframe). The SDK version being used to _build_ containers carries along a certain RuntimeFrameworkVersion implicitly, but the containers made available on `mcr.microsoft.com` lag behind in terms of the runtimes installed on them (by design - they are only made in the run-ups to each preview release) and so in the SDK we tend to make framework-dependent apps that we actually _can't_ run in containers.  We (containers team) need a better strategy for this and so I added these properties to 'pin' the runtimes for projects targeting net8.0 as a temporary measure.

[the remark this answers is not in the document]
baronfel: I went ahead and extracted this comment^ to https://github.com/dotnet/sdk/issues/32099 for the containers team to follow-up on.

src/Tests/Microsoft.NET.Build.Tests/GlobalPropertyFlowTests.cs
nagilson: I changed it to `appIsSelfContainedOrRuntimeSpecific` which is longer and could be better, but I think it does make it clearer

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildASelfContainedApp.cs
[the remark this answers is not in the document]
nagilson: Nope, I was the very confused one. Lol

src/Tasks/Microsoft.NET.Build.Tasks/ValidateExecutableReferences.cs
```diff
@@ -75,14 +75,6 @@ protected override void ExecuteCore()
                     {
                         referencedProjectIsSelfContained = true;
                     }
-
-                    //  If the project is NOT RID agnostic, then a global RuntimeIdentifier will flow to it.
-                    //  If the project didn't explicitly specify a value for SelfContained, then this will
-                    //  set SelfContained to true
```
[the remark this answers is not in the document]
nagilson: That is a really good point. I think I've fixed this and added tests for it.

src/Assets/TestProjects/BlazorHosted/blazorhosted/blazorhosted.csproj
```diff
@@ -8,6 +8,11 @@
     <ProjectReference Include="..\blazorwasm\blazorwasm.csproj" />
   </ItemGroup>
 
+  <!-- Based on .NET 8 breaking change: Add SelfContained if RuntimeIdentifier is defined to match old behavior.. -->
+  <PropertyGroup Condition="'$(RuntimeIdentifier)' != '' and '$(SelfContained)' == ''">
+    <SelfContained>true</SelfContained>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
nagilson: Unfortunately, it looks like there are still edge cases where Blazor would be broken so I am coordinating with them. I don't think we want to release this and break Blazor in the process so it will probably get delayed to preview 5 unless we can do it quickly.

cc @javiercn @danroth27

src/Tests/Microsoft.NET.Build.Tests/ReferenceExeTests.cs
```diff
@@ -240,8 +265,12 @@ public void ReferencedExeCanRunWhenReferencesExeWithSelfContainedMismatchForDiff
 
                 project.Root.Element(ns + "PropertyGroup")
                     .Add(XElement.Parse($@"<RuntimeIdentifier Condition=""'$(TargetFramework)' == '{ToolsetInfo.CurrentTargetFramework}'"">" + EnvironmentInfo.GetCompatibleRid() + "</RuntimeIdentifier>"));
+
+                project.Root.Element(ns + "PropertyGroup")
+                    .Add(new XElement(ns + "SelfContained", "true"));
```
dsplaisted, line 270: To test the intended scenario, `SelfContained` should only be set to true for the non `windows` TargetFramework.  Here I've moved both properties into a new PropertyGroup with the appropriate condition.

```suggestion
                var propertyGroup = new XElement(ns + "PropertyGroup",
                    new XAttribute("Condition", $"'$(TargetFramework)' == '{ToolsetInfo.CurrentTargetFramework}'"));

                propertyGroup.Add(new XElement(ns + "RuntimeIdentifier", EnvironmentInfo.GetCompatibleRid()));
                propertyGroup.Add(new XElement(ns + "SelfContained", "true"));

                project.Root.Add(propertyGroup);
```

src/Tasks/Microsoft.NET.Build.Tasks/ValidateExecutableReferences.cs
```diff
@@ -54,6 +57,11 @@ protected override void ExecuteCore()
                 bool referencedProjectIsSelfContained = MSBuildUtilities.ConvertStringToBool(projectAdditionalProperties["SelfContained"]);
                 bool referencedProjectHadSelfContainedSpecified = MSBuildUtilities.ConvertStringToBool(projectAdditionalProperties["_SelfContainedWasSpecified"]);
 
+                // We can only access TargetFrameworks and NearestTargetFramework to find the referenced project "TargetFramework".
+                // We rely on the nearest one because it will pick the lowest 'most-compatible' tfm for the two projects.
+                // Since 'younger' TFMs are the ones that would error and are generally also what gets picked as 'most-copmaptible' we can use it.
+                var referencedProjectTargetFramework = NuGetFramework.Parse(nearestTargetFramework);
```
dsplaisted, line 63: We should not be parsing the TargetFramework here, that will not work well with aliased TargetFrameworks (see [here](https://github.com/NuGet/Home/issues/5154) and [here](https://github.com/NuGet/Home/pull/12124).

I think what we should instead be doing is getting the index of the NearestTargetFramework from the TargetFrameworks metadata, and then retrieving the value from the same index in the TargetFrameworkMonikers metadata.  You can then parse that moniker with the [FrameworkName](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.frameworkname.-ctor?view=net-8.0#system-runtime-versioning-frameworkname-ctor(system-string)) class, and check that the `Identifier` is `.NETCoreApp` and the `Version` is less than 8.

Here's a sample of what the data for a multi-targeted referenced project looks like from a binlog:

```
../ReferencedProject/ReferencedProject.csproj
    ReferenceSourceTarget = ProjectReference
    MSBuildSourceProjectFile = c:\git\dotnet-sdk-main\artifacts\tmp\Debug\ReferencedExe---C4AC8564\ReferencedProject\ReferencedProject.csproj
    HasSingleTargetFramework = false
    SetTargetFramework = TargetFramework=net8.0
    ReferenceOutputAssembly = true
    IsVcxOrNativeProj = false
    OutputItemType = 
    TargetPlatformMonikers = None;Windows,Version=7.0
    BuildReference = true
    Platform = AnyCPU
    AdditionalPropertiesFromProject = <AdditionalProjectProperties>
  <TargetFramework Name="net8.0">
    <_IsExecutable>true</_IsExecutable>
    <IsRidAgnostic>false</IsRidAgnostic>
    <ShouldBeValidatedAsExecutableReference></ShouldBeValidatedAsExecutableReference>
    <_SelfContainedWasSpecified>true</_SelfContainedWasSpecified>
  </TargetFramework>
  <TargetFramework Name="net8.0-windows">
    <_IsExecutable>true</_IsExecutable>
    <IsRidAgnostic>false</IsRidAgnostic>
    <ShouldBeValidatedAsExecutableReference></ShouldBeValidatedAsExecutableReference>
    <_SelfContainedWasSpecified></_SelfContainedWasSpecified>
  </TargetFramework>
</AdditionalProjectProperties>
    TargetFrameworkMonikers = .NETCoreApp,Version=v8.0;.NETCoreApp,Version=v8.0
    TargetFrameworks = net8.0;net8.0-windows
    IsRidAgnostic = false
    MSBuildSourceTargetName = GetTargetFrameworks
    NearestTargetFramework = net8.0
    Platforms = AnyCPU
    Targets = 
    OriginalItemSpec = ../ReferencedProject/ReferencedProject.csproj
```

nagilson: Thanks, I didn't consider this and pushed the change. I must admit I don't really get the difference between this and the TargetFrameworkElement retrived using `UseAttributeForTargetFrameworkInfoPropertyNames` and I didn't leverage that property. Please let me know if we need to use that as I don't really see what it is being used for

src/Tasks/Microsoft.NET.Build.Tasks/ValidateExecutableReferences.cs
```diff
@@ -76,10 +94,14 @@ protected override void ExecuteCore()
...
 
+                    //  We need to check if referenced project will become SelfContained because of its RuntimeIdentifier. This only happens on TargetFrameworks less than net8.0.
+                    var sdkVersionWhereRuntimeIdentifierNoLongerInfersSelfContained = new FrameworkName(".NETCoreApp,Version=v8.0");
+                    bool runtimeIdentifierInfersSelfContained = referencedProjectTargetFramework != null && referencedProjectTargetFramework.Identifier == ".NETCoreApp" && referencedProjectTargetFramework.Version < sdkVersionWhereRuntimeIdentifierNoLongerInfersSelfContained.Version;
+
                     //  If the project is NOT RID agnostic, then a global RuntimeIdentifier will flow to it.
                     //  If the project didn't explicitly specify a value for SelfContained, then this will
                     //  set SelfContained to true
-                    if (runtimeIdentifierIsGlobalProperty && !referencedProjectHadSelfContainedSpecified)
+                    if (RuntimeIdentifierInfersSelfContained && runtimeIdentifierIsGlobalProperty && !referencedProjectHadSelfContainedSpecified)
```
dsplaisted, line 104: I think this needs to refer to the local variable, rather than the property which doesn't ever seem to be set.

```suggestion
                    if (runtimeIdentifierInfersSelfContained && runtimeIdentifierIsGlobalProperty && !referencedProjectHadSelfContainedSpecified)
```

src/Tasks/Microsoft.NET.Build.Tasks/ValidateExecutableReferences.cs
```diff
@@ -76,10 +94,14 @@ protected override void ExecuteCore()
                         referencedProjectIsSelfContained = true;
                     }
 
+                    //  We need to check if referenced project will become SelfContained because of its RuntimeIdentifier. This only happens on TargetFrameworks less than net8.0.
+                    var sdkVersionWhereRuntimeIdentifierNoLongerInfersSelfContained = new FrameworkName(".NETCoreApp,Version=v8.0");
+                    bool runtimeIdentifierInfersSelfContained = referencedProjectTargetFramework != null && referencedProjectTargetFramework.Identifier == ".NETCoreApp" && referencedProjectTargetFramework.Version < sdkVersionWhereRuntimeIdentifierNoLongerInfersSelfContained.Version;
```
dsplaisted, line 99: I think this can be simplified:

```suggestion
                    bool runtimeIdentifierInfersSelfContained = referencedProjectTargetFramework != null && referencedProjectTargetFramework.Identifier == ".NETCoreApp" && referencedProjectTargetFramework.Version.Major < 8;
```

dsplaisted (approved):

nagilson:
Here's the breaking change doc: https://github.com/dotnet/docs/issues/33726

PR #30044, opened 2023-01-20
Prevent editors from auto-trimming trailing whitespaces in test approval files

GangWang01:
### Problem
https://github.com/dotnet/templating/issues/5805

### Solution
Disable trimming trailing whitespaces in .editorconfig for test approval or snapshot files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -501,3 +501,6 @@ dotnet_diagnostic.IDE0040.severity = warning
 
 [*.txt]
 insert_final_newline = false
+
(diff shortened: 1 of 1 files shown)


vlada-shubina (approved):

GangWang01:
> How about also adding a [`whitespace`](https://git-scm.com/docs/gitattributes#_checking_whitespace_errors) setting to `.gitattributes`?

@KalleOlaviNiemitalo Thanks for your suggestion! It's not relevant to the intention here. We need to avoid white spaces in files for the verification in tests being removed while editing in VS.

PR #30057, opened 2023-01-20
Port two better fixes for installer crashes from main

marcpopMSFT:
We are still seeing internal instability in the installer tests. My prior fix was not sufficient as it doesn't work on all legs. I believe the fixes from Alex and Kasper are better approaches long term so I'm cherry-picking them into 7.0.1xx to help improve stability in installer.


joeloff:
I'm seeing similar failures in 7.0.2xx and 7.0.3xx only on the Darwin legs in installer

dsplaisted (approved):
Can you link to the PRs you're cherry-picking from?

marcpopMSFT:
Merged into https://github.com/dotnet/sdk/pull/30081 so I could make the code match main more closely to avoid later merge conflicts.

PR #30077, opened 2023-01-23
BlazorWasmSdk: Serve .webcil as application/octet-stream

lambdageek:
WebCIL is a new container format for .NET assemblies when publishing to webassembly.  (See https://github.com/dotnet/runtime/pull/79416)

Related to https://github.com/dotnet/runtime/issues/80807


javiercn (approved):
We also need to update this within UseBlazorFrameworkFiles in the aspnetcore repo

PR #30091, opened 2023-01-24
Guard ILCompilerTargetsPath import

ViktorHofer:
Apparently, in some cases, msbuild fails when the Project attribute in an Import statement is empty.

from a binlog created by Jiri Cincura (during publish):
`D:\repos\performance\tools\dotnet\x64\sdk\8.0.100-alpha.1.23073.15\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(1189,3): error MSB4020: The value "" of the "Project" attribute in element <Import> is invalid. [D:\repos\performance\src\scenarios\emptyconsolenativeaot\app\emptyconsolenativeaot.csproj]`

vs. when I tried something similar on my machine (during restore):
`NoImport: $(ILCompilerTargetsPath) at (1184;3) empty expression`

I think it's best to guard the Import via a Condition as the `ILCompilerTargetsPath` property isn't availble during Restore (which doesn't import nuget props and targets) and possibly during design time builds. Still, we should follow-up and understand why MSBuild sometimes accepts an empty expression vs when it errors out.

cc @rainersigwald

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1186,8 +1186,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.Sdk.CSharp.targets" Condition="'$(Language)' == 'C#'" />
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.Sdk.VisualBasic.targets" Condition="'$(Language)' == 'VB'" />
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.Sdk.FSharp.targets" Condition="'$(Language)' == 'F#'" />
-  <Import Project="$(ILCompilerTargetsPath)" Condition="'$(PublishAot)' == 'true'"/>
-  <Import Project="$(ILLinkTargetsPath)" Condition="'$(ILLinkTargetsPath)' != '' And '$(Language)' != 'C++'" />
+  <Import Project="$(ILCompilerTargetsPath)" Condition="'$(ILCompilerTargetsPath)' != '' and '$(PublishAot)' == 'true'"/>
+  <Import Project="$(ILLinkTargetsPath)" Condition="'$(ILLinkTargetsPath)' != '' and '$(Language)' != 'C++'" />
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.NET.Sdk.Analyzers.targets" Condition="'$(Language)' == 'C#' or '$(Language)' == 'VB'" />
 
   <!-- Import WindowsDesktop targets if necessary -->


rainersigwald:
The MSBuild evaluator has a permissive mode, used in restore--the idea is that maybe we can't import the thing because it hasn't been restored yet, but if we just ignore it hopefully we'll collect enough information about the packages to restore it and then import it correctly on the "real" evaluation for build. That can sometimes cause surprising behavior around incorrectly-specified imports, as here.

ViktorHofer:
Based on Rainer's above response, a repro for this symptom would be a build that either doesn't invoke restore first or that explicitly sets NuGet's property to disable props and targets imports: `/p:ExcludeRestorePackageImports=true`.

LakshanF:
> Based on Rainer's above response, a repro for this symptom would be a build that either doesn't invoke restore first or that explicitly sets NuGet's property to disable props and targets imports: `/p:ExcludeRestorePackageImports=true`.

The guard seems reasonable and adding some context.

NativeAot assumes that the property, `ILCompilerTargetsPath`, is going to be always [set ](https://github.com/dotnet/runtime/blob/1c6752813014463c93ae753ee45994284878f007/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.props#L16) either directly by explicit NativeAOT package reference or via the [SDK](https://github.com/dotnet/runtime/blob/86a77421378927a8efaf3d41a22ed77d9066628b/src/installer/pkg/projects/Microsoft.DotNet.ILCompiler/sdk/Sdk.props#L20). This is to ensure that the 2 (build and runtime) NativeAOT packages are version compatible. We recently started using implicit package reference to download the NativeAOT built package and I will open a tracking issue to double check all scenarios are covered in setting the `ILCompilerTargetsPath` property.

LakshanF (approved):
Opened issue #30094 to track this on the NativeAot side.

ViktorHofer:
> NativeAot assumes that the property, ILCompilerTargetsPath, is going to be always [set ](https://github.com/dotnet/runtime/blob/1c6752813014463c93ae753ee45994284878f007/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.props#L16)either directly by explicit NativeAOT package reference or via the [SDK](https://github.com/dotnet/runtime/blob/86a77421378927a8efaf3d41a22ed77d9066628b/src/installer/pkg/projects/Microsoft.DotNet.ILCompiler/sdk/Sdk.props#L20).

AFAIK with the recent changes that switched the consumption to  an implicit package reference, the SDK itself doesn't set the `ILCompilerTargetsPath` anymore. The "SDK" link above points to the package's Sdk.props file which isn't part of the SDK but of the package.

Because of that, the Sdk imports fails whenever restore is broken or package imports aren't happening (because of `ExcludeRestorePackageImports`). A side question, probably unrelated to this discussion: Why does the Sdk import the targets file instead of just the nuget package via the usual naming convention given that we use an implicit package reference now?

LakshanF:
> AFAIK with the recent changes that switched the consumption to an implicit package reference, the SDK itself doesn't set the `ILCompilerTargetsPath` anymore. The "SDK" link above points to the package's Sdk.props file which isn't part of the SDK but of the package.
> 
> Because of that, the import in the Microsoft.Net.Sdk.targets fails whenever restore is broken or package imports aren't happening (because of `ExcludeRestorePackageImports`). A side question, probably unrelated to this discussion: Why does the Sdk import the targets file instead of just the nuget package via the usual naming convention given that we use an implicit package reference now?

Yes, we should do some cleanup after moving to implicit package reference including perhaps not needing 2 packages for NativeAOT now. I'll add the need to review post implicit work to the above issue.

PR #28405, opened 2022-10-07
Add PackageDependenciesDesignTime to cache assets file

ocallesp:
Fixes https://github.com/dotnet/sdk/issues/27738

Implements reading the PackageDefinitions and `PackageDependencies` items in `ResolvePackageAssets`, so that `ResolvePackageDependencies` doesn't need to be called at all unless `EmitLegacyAssetsFileItems` is true. This improve incremental build perf and also significantly improve the design time build perf (23%) for this large solution scenario by saving to the cache file the same items generated from `PreprocessPackageDependenciesDesignTime`.

----------------
Note:
This implementation was taken from `PreprocessPackageDependenciesDesignTime.ExecuteCore`

PR #30105, opened 2023-01-24
Skip package when resolving package dependencies if it appears multiple times

### LINKED ISSUE #86: The "ResolvePackageDependencies" task failed unexpectedly - An item with the same key has already been added
@jogibear9988 commented on [Sun, 19 Jul 2020 08:00:16 GMT](https://github.com/NuGet/Home/issues/9818)
Error	MSB4018	The "ResolvePackageDependencies" task failed unexpectedly.
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.GetPackageAndFileDependencies(LockFileTarget target)
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.RaiseLockFileTargets()
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.ExecuteCore()
   at Microsoft.NET.Build.Tasks.TaskBase.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()	
## Details about Problem

VS UI

## Detailed repro steps so we can see the same problem

1. I have a complex Solution. This error I get after some Nuget Updates during build.

2. How can I find out wich Nuget is causing the issue and why?

smitpatel:
Resolves https://github.com/dotnet/NuGet.BuildTasks/issues/86

NuGet shows error about conflict already so we just skip it

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
@@ -279,8 +279,26 @@ namespace Microsoft.NET.Build.Tasks
 
         private void GetPackageAndFileDependencies(LockFileTarget target)
         {
-            var resolvedPackageVersions = target.Libraries
-                .ToDictionary(pkg => pkg.Name, pkg => pkg.Version.ToNormalizedString(), StringComparer.OrdinalIgnoreCase);
+            // In case of there are duplicated packages in libraries with differing versions,
+            // We skip package altogether
+            var resolvedPackageVersions = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            var duplicatedPackages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (LockFileTargetLibrary library in target.Libraries)
+            {
+                string libraryName = library.Name;
+                if (duplicatedPackages.Contains(libraryName))
+                {
+                    continue;
+                }
+
+                if (resolvedPackageVersions.ContainsKey(libraryName))
+                {
+                    duplicatedPackages.Add(libraryName);
+                    resolvedPackageVersions.Remove(libraryName);
+                }
+
+                resolvedPackageVersions[libraryName] = library.Version.ToNormalizedString();
+            }
 
             string frameworkAlias = _targetNameToAliasMap[target.Name];
 


src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
```diff
@@ -279,8 +279,26 @@ private void RaiseLockFileTargets()
...
+                    continue;
+                }
+
+                if (resolvedPackageVersions.ContainsKey(libraryName))
+                {
+                    duplicatedPackages.Add(libraryName);
+                    resolvedPackageVersions.Remove(libraryName);
+                }
+
+                resolvedPackageVersions[libraryName] = library.Version.ToNormalizedString();
```
smitpatel, line 300: If we let first one pass-through then we can avoid hashset lookup & dict.Remove calls.
If we want to minimize the impact then perhaps try/catch could be useful so we only run additional code only in error case.

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
```diff
@@ -279,8 +279,26 @@ private void RaiseLockFileTargets()
...
+            // In case of there are duplicated packages in libraries with differing versions,
+            // We skip package altogether
+            var resolvedPackageVersions = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            var duplicatedPackages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (LockFileTargetLibrary library in target.Libraries)
+            {
+                string libraryName = library.Name;
+                if (duplicatedPackages.Contains(libraryName))
+                {
+                    continue;
```
ViktorHofer, line 291: Add some logging so that i.e. a binlog shows that duplicate packages are consumed?

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
```diff
@@ -279,8 +279,26 @@ private void RaiseLockFileTargets()
...
+                }
+
+                if (resolvedPackageVersions.ContainsKey(libraryName))
+                {
+                    duplicatedPackages.Add(libraryName);
+                    resolvedPackageVersions.Remove(libraryName);
+                }
```
marcpopMSFT, line 298: might be better with an else here as you can then avoid having to remove every time

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
```diff
@@ -279,8 +279,26 @@ private void RaiseLockFileTargets()
...
-            var resolvedPackageVersions = target.Libraries
-                .ToDictionary(pkg => pkg.Name, pkg => pkg.Version.ToNormalizedString(), StringComparer.OrdinalIgnoreCase);
+            // In case of there are duplicated packages in libraries with differing versions,
+            // We skip package altogether
+            var resolvedPackageVersions = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            var duplicatedPackages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (LockFileTargetLibrary library in target.Libraries)
+            {
+                string libraryName = library.Name;
+                if (duplicatedPackages.Contains(libraryName))
```
marcpopMSFT, line 289: If we only use the duplicatedPackages HashSet here, is it worth the perf hit of even having it.  You could just have the below as it simplifies the code (unless the hashset check is so much faster than the dictionary containsKey check that it's worth it)
```
  if (!resolvedPackageVersions.ContainsKey(libraryName))
                {
resolvedPackageVersions[libraryName] = library.Version.ToNormalizedString();
                }
```

src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs
smitpatel: Can you point me to some example of logging, that would be helpful.

smitpatel:
@dsplaisted - Can you also take a look at this?

dsplaisted:
While looking at this, I remembered that as of #28405, we don't run the `ResolvePackageDependencies` task at all by default anymore.  That logic has moved to `ResolvePackageAssets`.  So this is probably not necessary anymore.  It would be good to try to repro the issue and see if it still repros on 7.0.200.

PR #30115, opened 2023-01-25
Add tests to verify analyzer configuration with analysis properties 

mavasani:
Tests added:

1. Add test to verify mapping from analysis properties to analyzer config file
2. Add test to verify bulk escalation of CA rules with different analysis modes

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
@@ -260,19 +260,31 @@ namespace Microsoft.NET.Build.Tests
             computedEffectiveAnalysisLevel.Should().NotBe(analysisLevel);
         }
 
-        [InlineData("latest", "all", "false")]
-        [InlineData("latest", "", "true")]
-        [InlineData("5", "", "true")]
-        [InlineData("5.0", "minimum", "false")]
-        [InlineData("6", "recommended", "false")]
-        [InlineData("6.0", "", "true")]
-        [InlineData("7", "none", "true")]
-        [InlineData("7.0", "", "false")]
-        [InlineData("8", "default", "false")]
-        [InlineData("8.0", "", "true")]
+        [InlineData("latest", "all", "false", "")]
+        [InlineData("latest", "", "true", "")]
+        [InlineData("latest", "all", "false", "Design")]
+        [InlineData("latest", "", "true", "Documentation")]
+        [InlineData("5", "", "true", "")]
+        [InlineData("5.0", "minimum", "false", "")]
+        [InlineData("5", "", "true", "Globalization")]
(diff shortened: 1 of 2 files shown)


mavasani:
Tagging @jmarolf @Youssef1313 for review

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
```diff
@@ -259,5 +260,95 @@ static void Main()
...
+        [InlineData("5", "", "true")]
+        [InlineData("5.0", "minimum", "false")]
+        [InlineData("6", "recommended", "false")]
+        [InlineData("6.0", "", "true")]
+        [InlineData("7", "none", "true")]
+        [InlineData("7.0", "", "false")]
+        [InlineData("8", "default", "false")]
+        [InlineData("8.0", "", "true")]
+        [RequiresMSBuildVersionTheory("16.8")]
+        public void It_maps_analysis_properties_to_globalconfig(string analysisLevel, string analysisMode, string codeAnalysisTreatWarningsAsErrors)
```
baronfel, line 274: To make sure I understand - there is an implicit dependency between roslyn and SDK that we didn't quite get to when we bumped the TFM to net8.0, and this test is making the dependency explicit so we don't later this year when we start the road to 9.0?

mavasani: Yes, that is correct.

The NetAnalyzers NuGet package drops props and targets file and some analyzerconfig files along with the analyzer assemblies. These props and targets file have the functionality to map MSBuild properties being tested here to a specific analyzerConfig file that is then passed to the compiler command line for appropriate analyzer configuration. We need to ensure that whenever LatestAnalysisLevel is bumped to 9.0 in future, the tooling that generates and packs these analyzer config files into the repo should be aware of the version bump to produce 9.0 specific analyzer config files.

baronfel: Sounds perfect - love that we're canonicalizing all of these touchpoints with tests!

baronfel (approved):

baronfel:
Happy to take this whenever, figured that @jmarolf might want to see it too.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToFloatWarningLevels.cs
mavasani: Yes, I am planning to add a few more of these over the coming days.

mavasani:
@baronfel I added another test to the PR, can you please take a look? Please feel free to merge the PR once you are ready to approve. @jmarolf and me have discussed a bunch in the past about the need for these tests, so we should be good to add these.

baronfel:
Great stuff @mavasani - thanks for taking the time.

PR #30139, opened 2023-01-26
Add  Visual Basic module template

YuliiaKovalova:
Add  Visual Basic module template and cover it with tests


vlada-shubina:
@KlausLoeffelmann could you please review the VB module template? thanks

src/Tests/dotnet-new.Tests/DotnetClassTemplateTests.cs
```diff
@@ -121,6 +121,10 @@ await engine.Execute(options)
         [InlineData("interface", "11.0", "net7.0")]
         [InlineData("interface", "10.0", "net6.0")]
         [InlineData("interface", "9", "netstandard2.0")]
+        [InlineData("module")]
+        [InlineData("module", "11.0", "net7.0")]
+        [InlineData("module", "10.0", "net6.0")]
+        [InlineData("module", "9", "netstandard2.0")]
         public async void DotnetVisualBasicClassTemplatesTest(
             string templateShortName,
             string langVersion = "",
```
vlada-shubina, line 130: not related to PR, but it would be good to add to those item tests: let's build the project once the item template is added, and see if it's can be built.

vlada-shubina (approved):

src/Tests/dotnet-new.Tests/DotnetClassTemplateTests.cs
YuliiaKovalova: done

KlausLoeffelmann (changes requested):
I have a general issue with the templates, I can only discuss later this week based on important people's availability.

Here is the general issue.

In VB we know the concept of default imports - it was a feature VB had _long_ before C#. They almost feel like implicit Imports, although I guess internally it's a bit different. If Imports are not defined on project level in particular, there is a general set of default imports defined. It's this:

image

So, I'd say as far as I can tell, we don't need in most cases _any_ import statements inside the class, module, etc.

This has also been overlooked for the standard VB templates, which I'd say we'd need to change.

This needs to be clarified, first.

@KathleenDollard, FYI.

KlausLoeffelmann:
Also @merriemcgaw for task prioritization info.

KathleenDollard:
We should test, but the imports should be added in TFMs for several releases. They should not require anything in the project files for defaults. Of course we should check that :-), particularly that the MSBuild stuff is set up to work correctly with .NET Standard.

I am not sure where we landed with the WinForms imports - whether the MSBuild stuff adds imports for those when WinForms is targeted. I am guessing it does not.

KathleenDollard:
> We should test, but the imports should be added in TFMs for several releases. 

Klaus pointed out that this may be confusing. I was trying to say that for the last several versions of the SDK, either since .NET Core 1 or .NET 5, VB has been adding imports via the SDK (referring here to the MSBuild props/targets brought in by the SDK entry in the project file).

KlausLoeffelmann:
Then let's get rid of the `Import` statements for this PR, and I think we're good to go!

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.langVersion=10.0.targetFramework=net6.0.verified/ClassLib.vbproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>ClassLib</RootNamespace>
```
KlausLoeffelmann, line 4: I would suggest no to have `RootNamespace` defined, except there is a specific reason I'd miss here.
It's likely that folks are renaming this project. But then they would need to change the `RootNamespace` in addition.

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.langVersion=10.0.targetFramework=net6.0.verified/ClassLib.vbproj
```diff
@@ -0,0 +1,9 @@
...
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10.0</LangVersion>
```
KlausLoeffelmann, line 6: Also, do we need to have `LangVersion` defined?

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.langVersion=11.0.targetFramework=net7.0.verified/ClassLib.vbproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>ClassLib</RootNamespace>
```
KlausLoeffelmann, line 4: Same as above - both for RootNamespace and Language version.

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.langVersion=9.targetFramework=netstandard2.0.verified/ClassLib.vbproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>ClassLib</RootNamespace>
```
KlausLoeffelmann, line 4: Same for RootNamespace.

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.langVersion=9.targetFramework=netstandard2.0.verified/ClassLib.vbproj
```diff
@@ -0,0 +1,9 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>9</LangVersion>
```
KlausLoeffelmann, line 6: Please note, that the language version for VB are totally different from C#:
https://learn.microsoft.com/en-us/dotnet/visual-basic/language-reference/configure-language-version

I am pretty sure, that we don't want a language version here which equals to the VS 2008 compiler (if I remember the versions correctly.)

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.verified/ClassLib.vbproj
```diff
@@ -0,0 +1,8 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>ClassLib</RootNamespace>
```
KlausLoeffelmann, line 4: Same as mentioned above.

src/Tests/dotnet-new.Tests/DotnetClassTemplateTests.cs
```diff
@@ -121,6 +122,10 @@ await engine.Execute(options)
         [InlineData("interface", "11.0", "net7.0")]
         [InlineData("interface", "10.0", "net6.0")]
         [InlineData("interface", "9", "netstandard2.0")]
+        [InlineData("module")]
+        [InlineData("module", "11.0", "net7.0")]
```
KlausLoeffelmann, line 126: Again, please be aware about the different language versions in VB and C#!

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Module-VisualBasic/.template.config/localize/templatestrings.en.json
```diff
@@ -0,0 +1,5 @@
+{
+  "author": "Microsoft",
+  "name": "Module",
+  "description": "Creates a new Visual Basic module"
```
KlausLoeffelmann, line 4: Do we want to add/replace the original item template descriptions?

image

(I don't have a strong opinion here, I just noticed.)

Also, in other item templates - are we ending the description with a period character?

YuliiaKovalova: From what I can see we don't follow a stable pattern.
@baronfel , should we unite the approach (description with/without period character)?

src/Tests/dotnet-new.Tests/Approvals/DotnetVisualBasicClassTemplatesTest.module.langVersion=10.0.targetFramework=net6.0.verified/ClassLib.vbproj
KlausLoeffelmann: Oh, I see. My bad.
If they come from the current templates, then we should modify them for .NET 8. As @KathleenDollard said: We are not servicing them for .NET 7.

KlausLoeffelmann:
I had built this branch and tested.

I created a new VB project, and then added a VB Module like this:

image

I would have expected the created Module would have be named like the filename, but I ended up with this (Module name derived from project name).

image

To me, that doesn't look correct.
Can we double check that this works as expected? Thanks!

Could we also check the other item templates (also in C#) for correct behavior in this regard?

Thanks!

YuliiaKovalova:
> I had built this branch and tested.
> 
> I created a new VB project, and then added a VB Module like this:
> 
> image
> 
> I had expected the created Module would have been named like the filename, but I ended up with this (Module name derived from project name).
> 
> image
> 
> To me, that doesn't look correct. Can we double check that this works as expected? Thanks!
> 
> Could we also check the other item templates (also in C#) for correct behavior in this regard?
> 
> Thanks!

Hi @KlausLoeffelmann ,

It's a nice catch! This problem was applicable to the module item template only due to my vague understanding of this object.
I have fixed it and covered with the corresponding test. Thank you!!!

KlausLoeffelmann:
One additional issue I found which is confusing:
To the solution of yesterday, I added a structure via this:

image

But then I added up with this.
Why was the structure put into a module?

image

Also, another thought in this context:
Since - in contrast to C# - it is `Structure` in VB rather than `struct`, should we call the template also "Structure" in the VB case?

YuliiaKovalova:
> One additional issue I found which is confusing: To the solution of yesterday, I added a structure via this:
> 
> image
> 
> But then I added up with this. Why was the structure put into a module?
> 
> image
> 
> Also, another thought in this context: Since - in contrast to C# - it is `Structure` in VB rather than `struct`, should we call the template also "Structure" in the VB case?

Hi Klaus, 

Thank you for guiding us. We treated the Module as an equivalent of the namespace in C#. I will remove it.
We can change struct to structure for VB. 
Is it ok to keep it in lowercase (consistent with Interface, Class, Enum, etc)?

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Structure-VisualBasic/.template.config/localize/templatestrings.en.json
```diff
@@ -1,5 +1,5 @@
 {
   "author": "Microsoft",
-  "name": "Struct",
+  "name": "Structure",
```
KlausLoeffelmann, line 3: Should we rename this file to Structure1.vb then?

KlausLoeffelmann:
I tested everything, and I am left with just one issue about Structs/Structures, where I would either need some additional guidance about build/reinstallation of the templates, or there actually is a new naming issue due to our latest changes. I did send you PM!

Thanks,

Klaus

KlausLoeffelmann (approved):

KlausLoeffelmann:
Retested the _Structure_ template item, and everything looks OK now.
We're good to go - thanks everyone!

YuliiaKovalova:
> Retested the _Structure_ template item, and everything looks OK now. We're good to go - thanks everyone!

Thank you Klaus!
You helped us a lot! Now the dotnet/templating team has a reliable contact point for the all Visual Basic related questions 😄

PR #30163, opened 2023-01-27
Adding extra check to identity on template installation

maridematte:
This is a PR to address https://github.com/dotnet/templating/issues/4817.

This changes how identity is resolved when checking for template installation validation. If the template is a path (to a local folder), this will read the `template.json` file of the template being installed and that will be checked against already installed template's identities instead of the path.


src/Cli/Microsoft.TemplateEngine.Cli/TemplatePackageCoordinator.cs
```diff
@@ -203,6 +204,7 @@ internal async Task<NewCommandStatus> EnterInstallFlowAsync(InstallCommandArgs a
                 string[] splitByColons = installArg.Split(new[] { "::" }, StringSplitOptions.RemoveEmptyEntries);
                 string identifier = splitByColons[0];
                 string? version = splitByColons.Length > 1 ? splitByColons[1] : null;
+                // For niget packages this is the package's name and not a path
```
YuliiaKovalova, line 207: typo in nuget

maridematte:
Switching this check to the template engine instead of the SDK

PR #30212, opened 2023-01-31
Fix up 'CurrentTargetFramework' usage error

v-wuzhai:


src/Assets/TestProjects/SimpleDependencies/SimpleDependencies.csproj
```diff
@@ -9,7 +9,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == '$(CurrentTargetFramework)'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp1.1'">
```
marcpopMSFT, line 12: @v-wuzhai will the Target Framework here ever be 1.1 or can we remove this itemgroup entirely?

v-wuzhai: Removing this ItemGroup will cause an error only when TargetFramework is 1.1, so this package is required when TargetFramework is 1.1.  Now our TargetFramework value is $(CurrentTargetFramework) so I think we can delete this ItemGroup.

marcpopMSFT (approved):

PR #29966, opened 2023-01-17
revert part of #29916

kasperk81:
to unblock linux-musl test in installer https://github.com/dotnet/installer/pull/15228#issuecomment-1384737277

PR #30233, opened 2023-01-31
revert part of #29916 (#29966)

marcpopMSFT:
This test will fail on the musl run without this change.


v-wuzhai:
Superceded by #30235

PR #30234, opened 2023-01-31
Fix SupportsTargetFramework for tests - 7.0.2xx

dsplaisted:
Port #30057 to 7.0.2xx


nagilson (approved):
LGTM... but I think we shouldn't merge this yet.

Issue #28370, opened 2022-09-15
dotnet publish profile silently continues even if you give an invalid publish profile path


nagilson:
When using a publish profile, there is no warning or error message for using an incorrect publish profile that doesn't exist.
Current behavior:
`dotnet new console` `dotnet publish -p:PublishProfile=TotallyNonexistentButRadPublishProfileFile -bl`. Will silently publish, binlog shows `PublishProfileImported=false` and `PublishProfile=TotallyNonexistentButRadPublishProfileFile `. 
Expected / Desired Behavior: 
MSBuild produces a warning akin to `A publish profile was provided with the name TotallyNonexistentButRadPublishProfileFile but could not be found in ./Properties/PublishProfiles/. `

HughWarrington:
I would add to the original comment, this should absolutely be a terminal error and not just a warning. According to [these docs](https://learn.microsoft.com/en-us/visualstudio/deployment/building-clickonce-applications-from-the-command-line?view=vs-2022#build-net-clickonce-applications-from-the-command-line):

> For .NET Core 3.1 and .NET 5 and later [...] you **need to** provide an additional property for the publish profile on the MSBuild command line

At the moment, if you specify a non-existent publish profile, the publish 'succeeds' but I have no idea what on earth it's doing, presumably it is using some undocumented defaults?

nagilson:
Closing this as resolved, thanks Vijay!

Issue #26161, opened 2022-06-22
Add a `--profile` option to `dotnet publish` to expose PublishProfiles as first-class concepts.


baronfel:
### Is your feature request related to a problem? Please describe.

Currently, users can define PublishProfiles for use with `dotnet publish`. These profiles are XML Project files that contain properties unique to a specific 

a) publish _mechanism_, like a filesystem folder containing all of your assets, or a build of a Dockerfile), and
b) publish _target_, like the filesystem folder to which to publish, or the external Registry to push the generated Docker image to

The default publish profile is a filesystem publish to the `/publish` directory, and users are likely already quite familiar with the behavior of this, but there are several others included in the SDK, and various tooling in Visual Studio (and other mechanisms) will scaffold out PublishProfiles on your behalf.

To use a specific publish profile, the user has to add the following MSBuild property: `/p:PublishProfile=<Name of profile>`. This is obtuse and easy to typo.

### Describe the solution you'd like
I propose a new string Option be added to the publish Command - `[--profile PUBLISH_PROFILE_NAME]`. This command must take a single argument that is the file name (without extension) of a PublishProfile that exists in `<project root>/Properties/PublishProfiles` OR the name of one of the in-box, default PublishProfiles (which are located [here](https://github.com/baronfel/sdk/tree/main/src/WebSdk/Publish/Targets/PublishProfiles)) . The user-provided value will be forwarded to the Publish target as the value of the MSBuild property `PublishProfile`.  Tab-completion for this property can be quickly determined based on these files, _but_ we should probably filter the list by the value of the `WebPublishMethod` inside the PublishProfile itself - some methods do not work on non-Windows systems.

HughWarrington:
Can I request that this also fails when you provide a publish profile that doesn't exist? Would fix #28370.

PR #30239, opened 2023-02-01
Warn if the publish profile provided is not present in the project

vijayrkn:
Fixes: https://github.com/dotnet/sdk/issues/28370

Fix: 
Adds a warning if the publish profile is provided but is not imported and is not present in the project.
The fix also makes sure that websdk will mark the profile as imported if the default profile is imported.

Why this fix? 
With publish profiles getting more and more popular and most projects now onboarding to this profile concept, it is important that we warn the user if their provided profile is not imported for any reason.

why warning and not an error?
PublishProfile property is just an msbuild property that the sdk handles in a specific way. It would be aggressive to fail the build just because the user passed an arbitrary msbuild property. Also, having shipped this way for the last various versions, it is probably safer to warn than to error in this scenario.

Tests: 
Added tests for both the success and failure scenarios.

@baronfel / @noahfalk - Let me know if you have any concerns with the approach/fix here.

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -888,7 +888,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1195: "}</comment>
   </data>
   <data name="PublishProfileNotPresent" xml:space="preserve">
-    <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name</value>
+    <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name.</value>
     <comment>{StrBegin="NETSDK1198: "}</comment>
   </data>
 </root>
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -887,4 +887,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1195: The SDK does not support ahead-of-time compilation. Set the PublishAot property to false.</value>
     <comment>{StrBegin="NETSDK1195: "}</comment>
   </data>
+  <data name="PublishProfileNotPresent" xml:space="preserve">
+    <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name</value>
+    <comment>{StrBegin="NETSDK1198: "}</comment>
+  </data>
 </root>
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
@@ -1076,11 +1076,12 @@ public static class Program
         [InlineData("invalidProfile", true)]
         [InlineData("invalidProfile.pubxml", true)]
         [InlineData("..\\Properties\\PublishProfiles\\invalidProfile.pubxml", true)]
+        [InlineData("invalidProfile.txt", true)]
         [InlineData("testProfile", false)]
         [InlineData("testProfile.pubxml", false)]
         [InlineData("..\\Properties\\PublishProfiles\\testProfile.pubxml", false)]
         [InlineData("", false)]
-        public void It_warns_with_an_invalid_publish_profile(string publishProfile, bool shouldWarn)
+        public void It_warns_with_an_invalid_publish_profile_NetSdk(string publishProfile, bool shouldWarn)
         {
             var tfm = ToolsetInfo.CurrentTargetFramework;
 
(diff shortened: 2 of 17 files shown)


dsplaisted:
> why warning and not an error?
PublishProfile property is just an msbuild property that the sdk handles in a specific way. It would be aggressive to fail the build just because the user passed an arbitrary msbuild property.

If the SDK uses a property and it has an invalid value, it is appropriate to generate an error and fail the build.  We do this for all sorts of other MSBuild properties.

> Also, having shipped this way for the last various versions, it is probably safer to warn than to error in this scenario.

This could be a reason to generate a warning instead of an error.  There's probably no way to tell how many people are passing invalid publish profiles, so it could be safer to generate a warning.  On the other hand, warnings still can generate build breaks for folks that have warnings as errors enabled.

vijayrkn:
I am marking this as draft since I have identified an issue if a Web SDK based project references a .NET SDK based class lib and the profile passed to it is a default profile, then it could generate un-intended warnings.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -887,4 +887,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1195: The SDK does not support ahead-of-time compilation. Set the PublishAot property to false.</value>
     <comment>{StrBegin="NETSDK1195: "}</comment>
   </data>
+  <data name="PublishProfileNotPresent" xml:space="preserve">
+    <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name</value>
```
baronfel, line 891: At the time this warning is created, could we show the list of known or valid profiles? That would make it more actionable for the end user. Other alternatives:

* list only profiles specific to the project being published, never any 'default' profiles
* don't show the list here, and instead do something like https://github.com/dotnet/sdk/issues/26161, pushing validation into that option instead
* do nothing

baronfel:
It would be good to do _something_ here - this request has come up a couple times before, and in order to hit this case users have to be explicitly specifying a publish profile (meaning they are already customizing their builds, so they'd likely be ok with making an adjustment).

src/Tasks/Common/Resources/Strings.resx
vijayrkn: Comments on various options: 
1. We have seen instances where users project may have more than 100+ profiles. So listing all profiles might not be an option.
2. I think having profile an option can still be an orthogonal feature that we can add. Even in this case, they could make a mistake and the warning could still be useful.

vijayrkn:
> I am marking this as draft since I have identified an issue if a Web SDK based project references a .NET SDK based class lib and the profile passed to it is a default profile, then it could generate un-intended warnings.

The code was working as expected. The publish command is only run once for the project getting published. The reason I saw multiple warnings while testing was when I ran it at the solution level instead of project level. Multiple warnings are expected when the publish is run at the solution level for each project getting published. So, moving the PR out of draft. 

Thanks @dsplaisted for the pointer. I completely missed the fact that I was unknowingly running the command on the solution instead of the project.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -1072,6 +1072,61 @@ public void It_publishes_with_full_path_publish_profile()
...
+        [InlineData("", false)]
+        public void It_warns_with_an_invalid_publish_profile(string publishProfile, bool shouldWarn)
+        {
+            var tfm = ToolsetInfo.CurrentTargetFramework;
+
+            var testProject = new TestProject()
+            {
+                Name = "ConsoleWithPublishProfile",
+                TargetFrameworks = tfm,
+                ProjectSdk = "Microsoft.NET.Sdk;Microsoft.NET.Sdk.Publish",
```
dsplaisted, line 1091: Is it the normal customer scenario here that the project would specify the Publish Sdk?

dsplaisted (approved):
I'd lean towards generating an error message, but generating a warning is better than nothing at all, and it looks like we may add telemetry around this to help us get data about whether we should later change the warning to an error.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
vijayrkn: For the console apps, the publish SDK is not required. But we do add the publish sdk for console projects under specific circumstances like publishing the consoleapp as a WebJob to Azure App service etc... 

In this case the behavior should be same with or w/o the sdk.. I will add a test for that scenario as well.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -1072,6 +1072,61 @@ public void It_publishes_with_full_path_publish_profile()
...
 
+        [Theory]
+        [InlineData("invalidProfile", true)]
+        [InlineData("invalidProfile.pubxml", true)]
+        [InlineData("..\\Properties\\PublishProfiles\\invalidProfile.pubxml", true)]
+        [InlineData("testProfile", false)]
+        [InlineData("testProfile.pubxml", false)]
+        [InlineData("..\\Properties\\PublishProfiles\\testProfile.pubxml", false)]
+        [InlineData("", false)]
+        public void It_warns_with_an_invalid_publish_profile(string publishProfile, bool shouldWarn)
```
nagilson, line 1083: One other thing to consider is that this doesn't look like it'd catch the scenario where you have a file with the corresponding name, but it is not a valid publish profile. Like `dotnet publish -p:PublishProfile=foo.txt`. Maybe we don't need to account for that here.

src/Tasks/Common/Resources/Strings.resx
nagilson, line 891: nit: 
```suggestion
    <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name.</value>
```

nagilson: I do like the option of showing potential profiles.

To circumvent the 100+ issue: We could do something like, show the available profiles that have a 'similar' string to what you entered, such as I gave --profile myPublishProfile, but the real publish profile was `publishprofile`. With that being said, that is probably a p3/p4 issue.

@baronfel Personally I would even like this feature in the SDK, where you type `dotnet built` and it says `did you mean dotnet build?`

nagilson (approved):
Thank you for making this, it's a big improvement in the UX 🎆

src/Tasks/Common/Resources/Strings.resx
vijayrkn: Thanks for catching it. I will fix it.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
vijayrkn: The scenario itself should be handled since we check for the exist check.I will add it to the test as well.

src/Tasks/Common/Resources/Strings.resx
vijayrkn: Since we already have an issue tracking adding support for publishProfile as a top level option - https://github.com/dotnet/sdk/issues/26161, I will mark this conversation as resolved.

PR #30265, opened 2023-02-01
Fix bug with `dotnet publish -p:Foo` 

nagilson:
Just discovered this on my own testing ... since NET 7 there has been a invalid index exception which occurs when you `pack` or `publish` and give an incomplete property key value pair, such as `dotnet publish -p:Foo`. This fixes said error.  In NET 6 it was a no op to do this.

### DIFF
--- a/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
+++ b/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
@@ -282,23 +282,10 @@ namespace Microsoft.DotNet.Cli
 
             foreach (var keyEqValString in globalPropEnumerable)
             {
-                foreach (var keyEqVal in keyEqValString.Split(";"))
+                var propertyPairs = MSBuildPropertyParser.ParseProperties(keyEqValString);
+                foreach (var propertyKeyValue in propertyPairs)
                 {
-                    string[] keyValuePair = keyEqVal.Split("=", 2);
-                    if (keyValuePair.Count() == 0)
-                    {
-                        // The user provided an empty value after ;. e.g. -p:Foo=Bar;
-                        continue;
-                    }
-                    else if (keyValuePair.Count() == 1)
-                    {
-                        // The user provided a property such as -p:Foo without giving a value. Undefine the property.
-                        // Note that MSBuild will fail if given -p:Foo; but we can ignore this situation because of this and let MsBuild fail, or, assuming it comes to parity with ';', behave the same as if ; was not present.
-                        // The same can be said for -p:Foo=bar;-p:Bar=Foo, which is also invalid syntax to MSBuild.
-                        globalProperties[keyValuePair[0]] = "";
-                        continue;
-                    }
-                    globalProperties[keyValuePair[0]] = keyValuePair[1];
+                    globalProperties[propertyKeyValue.key] = propertyKeyValue.value;
                 }
             }
             return globalProperties;
--- a/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
+++ b/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
@@ -283,6 +283,11 @@ namespace Microsoft.DotNet.Cli
             foreach (var keyEqVal in globalPropEnumerable)
             {
                 string[] keyValuePair = keyEqVal.Split("=", 2);
+                if (keyValuePair.Count() != 2)
+                {
+                    // The user provided a property such as -p:Foo without giving a value. Ignore the property.
+                    continue;
+                }
                 globalProperties[keyValuePair[0]] = keyValuePair[1];
             }
             return globalProperties;
--- a/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
+++ b/src/Cli/dotnet/ReleasePropertyProjectLocator.cs
@@ -280,15 +280,26 @@ namespace Microsoft.DotNet.Cli
 
             string[] globalPropEnumerable = _parseResult.GetValue(CommonOptions.PropertiesOption);
 
-            foreach (var keyEqVal in globalPropEnumerable)
+            foreach (var keyEqValString in globalPropEnumerable)
             {
-                string[] keyValuePair = keyEqVal.Split("=", 2);
-                if (keyValuePair.Count() != 2)
+                foreach (var keyEqVal in keyEqValString.Split(";"))
                 {
-                    // The user provided a property such as -p:Foo without giving a value. Ignore the property.
-                    continue;
+                    string[] keyValuePair = keyEqVal.Split("=", 2);
+                    if (keyValuePair.Count() == 0)
+                    {
+                        // The user provided an empty value after ;. e.g. -p:Foo=Bar;
+                        continue;
+                    }
+                    else if (keyValuePair.Count() == 1)
+                    {
+                        // The user provided a property such as -p:Foo without giving a value. Undefine the property.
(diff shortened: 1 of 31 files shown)


dsplaisted:
What happens to a property such as this when it gets to the MSBuild property parsing?

Should we consider this for 7.0 servicing (or at least 7.0.300)?

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -283,6 +283,11 @@ private bool IsValidSlnFilePath(string path)
             foreach (var keyEqVal in globalPropEnumerable)
             {
                 string[] keyValuePair = keyEqVal.Split("=", 2);
+                if (keyValuePair.Count() != 2)
```
kasperk81, line 286: does it work for joined properties?
```
dotnet publish -p:'X=1;Y=2'
```

nagilson:
@dsplaisted  That's a good question... looks like MSBuild undefines it. I will change this code to account for this. I think we should service in 7.0.300.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Thanks, that's also a really good point. It didn't parse those properly either.

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -280,10 +280,26 @@ private bool IsValidSlnFilePath(string path)
...
+                    {
+                        // The user provided an empty value after ;. e.g. -p:Foo=Bar;
+                        continue;
+                    }
+                    else if (keyValuePair.Count() == 1)
+                    {
+                        // The user provided a property such as -p:Foo without giving a value. Undefine the property.
+                        // Note that MSBuild will fail if given -p:Foo; but we can ignore this situation because of this and let MsBuild fail, or, assuming it comes to parity with ';', behave the same as if ; was not present.
+                        // The same can be said for -p:Foo=bar;-p:Bar=Foo, which is also invalid syntax to MSBuild.
+                        globalProperties[keyValuePair[0]] = "";
```
dsplaisted, line 298: Do you need to specifically set this to empty, or can you just skip adding the key to the dictionary?

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
```diff
@@ -280,10 +280,26 @@ private bool IsValidSlnFilePath(string path)
...
+                        continue;
+                    }
+                    globalProperties[keyValuePair[0]] = keyValuePair[1];
```
dsplaisted, line 301: I'd suggest just using if/else logic here and skipping the `continue` statements:

```suggestion
                    if (keyValuePair.Count() == 0)
                    {
                        // The user provided an empty value after ;. e.g. -p:Foo=Bar;
                    }
                    else if (keyValuePair.Count() == 1)
                    {
                        // The user provided a property such as -p:Foo without giving a value. Undefine the property.
                        // Note that MSBuild will fail if given -p:Foo; but we can ignore this situation because of this and let MsBuild fail, or, assuming it comes to parity with ';', behave the same as if ; was not present.
                        // The same can be said for -p:Foo=bar;-p:Bar=Foo, which is also invalid syntax to MSBuild.
                        globalProperties[keyValuePair[0]] = "";
                    }
                    else
                    {
                        globalProperties[keyValuePair[0]] = keyValuePair[1];
                    }
```

dsplaisted (approved):

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: We should. If we skip adding it to the dictionary, it will not mimic Msbuild behavior... for example, if I do -p:_IsPublishing, that will globally undefine, permanently, the _IsPublishing property ... if we dont do the same for these properties we inject into the pre-evaluation, the pre-evaluation can yield inconsistent results with what msbuild would actually do

baronfel: A lot of this parsing seems to overlap with [MSBuildPropertyParser](https://teams.microsoft.com/l/channel/19%3a2db9948434504972b9bc70d3b3ad71e2%40thread.skype/CLI?groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47), which currently is only used when calculating the values to 'forward' to MSBuild. Is there opportunity for reuse here, or potentially even changing the type of the `PropertiesOption` away from `string[]` to a type that more accurately describes the semantics?

nagilson: We should re-use it for sure, initially we used it. I thought it handled the `-p` `--p` `/p` shenanigans for some reason when I initially used it, and when I realized I didn't I went back to using the PropertiesOption because I thought it handled everything. Obviously, my world view was incorrect. 

BTW, that link is to our teams channel, should change and delete the edit history ;)

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: This is a 10/10 suggestion but the code has changed so not doing this now

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -584,6 +585,8 @@ public void It_publishes_on_release_if_PublishRelease_property_set_in_csproj()
         [InlineData("-property:Configuration=Debug")]
         [InlineData("--property:Configuration=Debug")]
         [InlineData("/p:Configuration=Debug")]
+        [InlineData("-p:_IsPublishing=true;Configuration=Debug")]
```
nagilson, line 588: this test used to have the surrounding ' but caused unspeakable disaster (it seems to work but not under test...) 
per @/baronfel
https://nullprogram.com/blog/2022/02/18/
It's the wild west of Windows command line parsing 😬

src/Cli/dotnet/ReleasePropertyProjectLocator.cs
nagilson: Using it again now 🎉

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
```diff
@@ -600,7 +603,7 @@ public void PublishRelease_does_not_override_Configuration_property_across_forma
                });
 
             new DotnetPublishCommand(Log, helloWorldAsset.TestRoot)
-                .Execute(configOpt)
+                .Execute(configOpt,"-bl:C:\\users\\noahgilson\\it_passes_but_fails.binlog")
```
baronfel, line 606: Not sure if you intended this binlog to get emitted as well

baronfel (approved):

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAHelloWorldProject.cs
nagilson: Definitely not, thanks! I meant to force push this earlier and remove it ...

PR #30266, opened 2023-02-01
Introduce `Workload Clean`: Garbage Collection Component

### LINKED ISSUE #19603: Add workload force clean option
Consider adding an option to remove all old workload/pack installation records. Users can run into issues and want this functionality during previews where a workload is renamed and as a result cannot be uninstalled due to being unrecognized by the resolver.

nagilson:
# What's this?
Resolves https://github.com/dotnet/sdk/issues/19603
Doc: https://github.com/dotnet/sdk/pull/30480 
* Adds PART ONE of `dotnet workload clean`: 
(So not the stuff in `2)` shown below.)
* Make `workload list` show all VS workloads regardless of feature bands.

* Workload --info now correctly shows `msi` or `filebased` install type.

* Fixes Incorrect customer facing logging messages.

* Refactor MSI Garbage Collection into separate functions to make it easier to use

* Note that workloads will still appear after `workload clean` --> `workload list`, but not for `workload clean --all`, because we don't remove the workload installation record itself in GC. The old GC also does not remove workload installation records for workloads of uninstalled featurebands. To do this, we would need the manifest or some sort of history ledger to map N-1 packs to a workload id. Or maybe we could include the `workload id` in the workload pack records, but not sure if we want to do that.

* Evidence of VS Workload Behavior:
image

* Evidence of MSI Clean up:
image
image
image

* Shows that MSI clean up works for older feature bands: 
```
C:\Users\WDAGUtilityAccount>dotnet --info
.NET SDK:
 Version:   7.0.103
 Commit:    276c71d299
...

C:\Users\WDAGUtilityAccount>dotnet workload install wasm-tools

The machine has a pending reboot. Installation will continue, but you may need to restart.
...

Successfully installed workload(s) wasm-tools.


C:\Users\WDAGUtilityAccount>dotnet workload list

Installed Workload Id      Manifest Version      Installation Source
--------------------------------------------------------------------
wasm-tools                 7.0.3/7.0.100         SDK 7.0.100
maui-windows               7.0.59/7.0.100        VS 17.4.33403.182

...


--> Then in an 8.0 terminal:

PS C:\Users\WDAGUtilityAccount\sdk> dotnet workload clean --all

--> Then back to 7:

C:\Users\WDAGUtilityAccount>dotnet workload list

Installed Workload Id      Manifest Version      Installation Source
--------------------------------------------------------------------
maui-windows               7.0.59/7.0.100        VS 17.4.33403.182
```

```
[2023-02-09 16:25:41.122] [00001D28] Microsoft.Maui.Resizetizer.Sdk (7.0.59) will not be removed because other dependents remain: VS.{AEF703B8-D2CC-4343-915C-F54A30B90937}.
[2023-02-09 16:25:41.123] [00001D28] Microsoft.Maui.Sdk (6.0.552) will not be removed because other dependents remain: VS.{AEF703B8-D2CC-4343-915C-F54A30B90937}.
[2023-02-09 16:25:41.124] [00001D28] Microsoft.Maui.Sdk (7.0.59) will not be removed because other dependents remain: VS.{AEF703B8-D2CC-4343-915C-F54A30B90937}.
[2023-02-09 16:25:41.124] [00001D28] Microsoft.Maui.Templates.net6 (6.0.552) will not be removed because other dependents remain: VS.{AEF703B8-D2CC-4343-915C-F54A30B90937}.
[2023-02-09 16:25:41.124] [00001D28] Microsoft.Maui.Templates.net7 (7.0.59) will not be removed because other dependents remain: VS.{AEF703B8-D2CC-4343-915C-F54A30B90937}.
```

* Show that VS workloads across versions are displayed:

```
PS C:\Users\source\repos\sdk> dotnet workload clean --all
Workload 'maui-windows (VS 17.4.33213.308, VS 17.4.33205.214)' was not removed because it is installed and managed by Visual Studio. Please uninstall this workload through Visual Studio to fully remove it.
Workload 'maui-ios (VS 17.4.33205.214)' was not removed because it is installed and managed by Visual Studio. Please uninstall this workload through Visual Studio to fully remove it.
Workload 'ios (VS 17.4.33205.214)' was not removed because it is installed and managed by Visual Studio. Please uninstall this workload through Visual Studio to fully remove it.
```

* Updated test plan for MSIs for vendors.

(Note that I added this into my own copy of `main` so it won't go into prod.)

Should
1) Run garbage collection for the SDK feature band under execution
~~2) Restore Manifests to 'fresh' state, to whatever baseline manifests were initially shipped with the current SDK feature band, however, including edits that VS made to said manifest, assuming a ledger is available. (Daniel is working on the ledger.)~~

Add `dotnet workload clean --all`:
Should
1)  Remove ALL CLI installed workloads for the SDK feature band under execution and for all feature bands of the SDK below the current version of the SDK. It should also give a message for VS installed workloads and say that these workloads weren't installed because they are managed by VS.

Note that this is different from running `dotnet workload uninstall` en masse. Workload uninstall needs the manifest to know what has been orphaned.

~~2) Restore manifests to 'fresh' state, like above,  but for all lower SDK feature bands if possible. (They may not have a ledger available.)~~ 

# Not in this PR:

Make `workload list` show all workloads regardless of feature band.


src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -107,7 +107,7 @@ public IEnumerable<WorkloadDownload> GetDownloads(IEnumerable<WorkloadId> worklo
         /// <summary>
         /// Cleans up and removes stale workload packs.
         /// </summary>
-        public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = null)
```
nagilson, line 110: The diff looks gross but it's mostly the same code minus CleanAllPacks, and it has been split into multiple functions to make it more digestible to read.

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (_cleanAll)
+            {
+                _workloadInstaller.GarbageCollectInstalledWorkloadPacks(cleanAllPacks: true);
+
+                if (OperatingSystem.IsWindows())
+                {
+                    InstalledWorkloadsCollection vsWorkloads = new();
+                    VisualStudioWorkloads.GetInstalledWorkloads(_workloadResolver, new SdkFeatureBand(_sdkVersion), vsWorkloads);
+
+                    foreach (var vsWorkload in vsWorkloads.AsEnumerable())
```
nagilson, line 72: We should decide if we want to clean all feature bands, and then if not, do we still list VS workloads from all feature bands, or just the current? My vote is that we list all of them.

src/Tests/dotnet-workload-install.Tests/GivenFileBasedWorkloadInstall.cs
```diff
@@ -337,7 +337,7 @@ public void GivenManagedInstallItDoesNotRemovePacksWithInstallRecords()
                 }
             }
             // Write fake workload install record for 6.0.100
```
nagilson, line 339:
```suggestion
            // Write fake workload install record for 6.0.300
```

src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
```diff
@@ -298,7 +298,7 @@
     <value>Workload updates are available. Run `dotnet workload list` for more information.</value>
   </data>
   <data name="PendingReboot" xml:space="preserve">
-    <value>The machine has a pending reboot. Installation will continue, but you may need to restart.</value>
```
nagilson, line 301: We used to say "Installation" will continue even in an uninstall comand, and now in the clean command...

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -217,46 +171,152 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
+
+        public void EradicateAllWorkloadInstallationRecords(SdkFeatureBand featureBand)
+        {
+            var allFeatureBands = RecordRepository.GetFeatureBandsWithInstallationRecords();
+
+            foreach (SdkFeatureBand potentialBandToClean in allFeatureBands)
+            {
+                if (potentialBandToClean.CompareTo(featureBand) < 1)
+                {
+                    var workloadInstallationRecordIds = RecordRepository.GetInstalledWorkloads(potentialBandToClean);
```
nagilson, line 314: This will delete records from VS as well which we don't want. Need to limit this

nagilson: Ack, I was hoping this would be done before I'm out tomorrow so it'd be reviewable. I think this is the last item remaining though, so pinging for review.

```suggestion
                    var workloadInstallationRecordIds = RecordRepository.GetCLIInstalledWorkloads(potentialBandToClean);
```

When I am back, I will implement `GetCLIInstalledWorkloads` and see if we can determine if the workload has a VS installation source in the registry key, so we don't return it and remove the VS workload installation record for MSI based clean --all.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -307,30 +308,44 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
             {
                 foreach (var packVersionDir in Directory.GetDirectories(packIdDir))
                 {
-                    var bandRecords = Directory.GetFileSystemEntries(packVersionDir);
+                    var bandRecordPaths = Directory.GetFileSystemEntries(packVersionDir);
 
-                    var unneededBandRecords = bandRecords
-                        .Where(recordPath => !installedSdkFeatureBands.Contains(new SdkFeatureBand(Path.GetFileName(recordPath))));
+                    // Mark packs for GC from versions / feature bands of the SDK that are not present on the machine
```
nagilson, line 313:
```suggestion
                    // Mark packs for GC from versions / feature bands of the SDK that are not present on the machine and lower or equal to the current sdk.
```

src/Cli/dotnet/commands/dotnet-workload/install/LocalizableStrings.resx
joeloff: Good catch

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -217,46 +171,152 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
                     else
                     {
                         packsToRemove.Add(packRecord);
+                        Log?.LogMessage($"Removing {packRecord.MsiId} ({packRecord.MsiNuGetVersion}) as no dependents remain.");
```
joeloff, line 174: Maybe change the message to `Adding .... for removal as no dependents remain` since this is not actually removing the pack, but adding it to the list of packs to remove

nagilson:
Marking as draft again to confirm the MSI portion of things is working in all scenarios.  Will ping once this is verified, sorry.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
nagilson: Done, good suggestion. Thanks.

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+                HashSet<string> vsWorkloadUninstallWarnings = new();
+
+                foreach (string sdkVersion in installedSDKVersionsWithPotentialVSRecords)
+                {
+                    try
+                    {
+#pragma warning disable CS8604 // We error in the constructor if the dotnet path is null.
+                        // The below environment expansion should be architecture agnostic for the C:\Program Files path, but only works on windows 7+. https://learn.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN
+                        // In addition, we don't need to worry about when dotnet will be dotnet/x64 (x64 on processor arch of arm64) because MSI install thru VS will only support installs on processor arch, not chip/emulation arch.
+                        string defaultDotnetWinPath = Path.Combine(Environment.ExpandEnvironmentVariables("%ProgramW6432%"), "dotnet");
```
joeloff, line 100: I'd suggest using Environment.SpecialFolder instead of the environment variables. Users can always override these.

Take a look at [this](https://github.com/dotnet/sdk/blob/033003f9fdbd09e33b3b2882c21d338d26bba88e/src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs#L202) for an example

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+                    {
+                        // Limitation: We don't know the dotnetPath of the other feature bands when making the manifestProvider and resolvers.
+                        // It can theoretically be customized, but that is not currently supported for workloads with VS.
+                        Reporter.WriteLine(AnsiColorExtensions.Yellow(string.Format(LocalizableStrings.CannotAnalyzeVSWorkloadBand, sdkVersion, ex)));
+                    }
+                }
+
+                foreach(string warning in vsWorkloadUninstallWarnings)
+                {
+                    Reporter.WriteLine(AnsiColorExtensions.Yellow(warning));
```
joeloff, line 126: Pretty sure you can use `warning.Yellow()`

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -18,6 +18,7 @@
 using Microsoft.DotNet.Workloads.Workload.Install.InstallRecord;
 using System.Text.Json;
 using System.Threading.Tasks;
```
joeloff, line 20: Mind sorting/removing the `using` statements (Ctrl+R+G in VS)

src/Cli/dotnet/commands/dotnet-workload/install/IInstaller.cs
```diff
@@ -20,7 +20,13 @@ internal interface IInstaller : IWorkloadManifestInstaller
 
         void RepairWorkloads(IEnumerable<WorkloadId> workloadIds, SdkFeatureBand sdkFeatureBand, DirectoryPath? offlineCache = null);
 
-        void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = null);
+        void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = null, bool cleanAllPacks = false);
+
+        /// <summary>
+        /// Remove all non-VS workload installation records for the specified feature band, and all records for feature bands lower than featureBand.
+        /// </summary>
+        /// <param name="featureBand">The maximum feature band to remove all workload installation records from.</param>
```
joeloff, line 28: Maybe rename the parameter to `maxFeatureBand` or something that indicates an upper bound.

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -418,7 +418,7 @@ protected string GetMsiLogName(WorkloadPackRecord record, InstallAction action)
         /// Get a list of all MSI based SDK installations that match the current host architecture.
         /// </summary>
         /// <returns>A collection of all the installed SDKs. The collection may be empty if no installed versions are found.</returns>
-        protected IEnumerable<string> GetInstalledSdkVersions()
+        public static IEnumerable<string> GetInstalledSdkVersions()
```
joeloff, line 421: Any reason for changing the access modifier to public? Since you're moving from an instance to a static method, `internal` might be a better option.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
```diff
@@ -83,8 +85,44 @@ public IEnumerable<WorkloadId> GetInstalledWorkloads(SdkFeatureBand sdkFeatureBa
...
+
+        public IEnumerable<WorkloadId> GetInstalledWorkloadsForFeatureBandAndBelow(SdkFeatureBand sdkFeatureBand)
+        {
+            using RegistryKey wrk = _baseKey.OpenSubKey(Path.Combine(BasePath, $"{sdkFeatureBand}"));
+            List<WorkloadId> allWorkloads = new();
+            var sdkBands = wrk?.GetSubKeyNames();
+
+            foreach (var band in sdkBands)
+            {
+                Log?.LogMessage($"Clean --all: Obtaining workload installation records for {BasePath} + {band}.");
```
joeloff, line 99: For MSI based installs we already log the initial command and arguments. If we ever renamed `clean --all`, this text would need to be updated. It might also be confusing if we ever used this code for something else in the future.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
```diff
@@ -83,8 +85,44 @@ public IEnumerable<WorkloadId> GetInstalledWorkloads(SdkFeatureBand sdkFeatureBa
...
+                try
+                {
+                    if (new SdkFeatureBand(band).CompareTo(sdkFeatureBand) < 1)
+                    {
+                        var sdkPath = Path.Combine(BasePath, band);
+                        allWorkloads.AddRange(GetWorkloadInstallationRecordsFromRegistry(_baseKey.OpenSubKey(sdkPath)));
+                    }
+                }
+                catch (Exception ex) when (
```
joeloff, line 108: Do we really only care about these exception types? Seems like any exception raised when trying to convert the version numbers should make us skip the invalid version, log it and move to the next one.

src/RazorSdk/Tool/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "WSL": {
+      "commandName": "WSL2",
```
joeloff, line 4: What's this?

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Parsing;
+using System.Diagnostics;
```
joeloff, line 8: Can this be removed? I don't see any other changes

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -340,6 +353,10 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
                             deletablePack = JsonSerializer.Deserialize(jsonPackInfo, typeof(PackInfo)) as PackInfo;
                         }
                         DeletePack(deletablePack);
+                        foreach (string packRecordToDeletePath in unneededBandRecordPaths)
+                        {
+                            DeletePackInstallationRecord(deletablePack, new SdkFeatureBand(Path.GetFileName(packRecordToDeletePath)));
+                        }
```
dsplaisted, line 359: I believe this is unnecessary, the pack installation records are already deleted in the `File.Delete(unneededRecord);` line.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -348,6 +365,29 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
                 }
             }
+
+            if (cleanAllPacks)
+            {
+                EradicateAllWorkloadInstallationRecords(_sdkFeatureBand);
+            }
+        }
+
+        public void EradicateAllWorkloadInstallationRecords(SdkFeatureBand featureBand)
```
dsplaisted, line 375: Suggest renaming this variable to avoid making it seem like we're only deleting the workload installation records for the specified band.

```suggestion
        public void EradicateAllWorkloadInstallationRecords(SdkFeatureBand currentFeatureBand)
```

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -348,6 +365,29 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
+        {
+            FileBasedInstallationRecordRepository workloadRecordRepository = new(_workloadMetadataDir);
+            var allFeatureBands = workloadRecordRepository.GetFeatureBandsWithInstallationRecords();
+
+            foreach (SdkFeatureBand potentialBandToClean in allFeatureBands)
+            {
+                if (potentialBandToClean.CompareTo(featureBand) < 1)
```
dsplaisted, line 382: As before, maybe we don't need this check at all.

```suggestion
                if (potentialBandToClean.CompareTo(featureBand) < 0)
```

src/Cli/dotnet/commands/dotnet-workload/install/IInstaller.cs
dsplaisted: Or `currentFeatureBand`.  Also for the file-based installer I think we could go ahead and delete all of the records regardless of band, so we might not need the parameter at all in that case.

dsplaisted (commented):
So far I mostly just looked at the file-based installer stuff, not the MSI based installer or the tests.

src/RazorSdk/Tool/Properties/launchSettings.json
nagilson: No clue, getting rid of it.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
nagilson: So we ended up never using this function, even though I tested it. We can keep it as the work has already been done and it might be useful at some point, or just entirely get rid of it.
Thoughts?

#

If we do keep it:
We could catch the following: (But the registry access would've already caused a failure elsewhere, most likely.)
[ObjectDisposedException](https://learn.microsoft.com/en-us/dotnet/api/system.objectdisposedexception?view=net-7.0)
or 
[SecurityException](https://learn.microsoft.com/en-us/dotnet/api/system.security.securityexception?view=net-7.0)
can be thrown by the `OpenSubKey` but the registry validation would've already occured. 

We SHOULD catch the following: (can occur if the feature band contains a character that's an invalid path.)
[ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception?view=net-7.0)

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
nagilson: Thanks, I think it would be even better to just call `GetDotNetHome`

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
nagilson: Good point. I've removed all references to the clean command in the log messages.

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+            
+        }
+
+        public override int Execute()
+        {
+            ExecuteGarbageCollection();
+            return 0;
+        }
+
+        private void ExecuteGarbageCollection()
```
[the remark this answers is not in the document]
nagilson: Initially there was `ExecuteGarbageCollection` and `ExecuteMSIRestoration` or something like that. I still prefer to keep it outside of execute. IMO it makes it clearer what is happening

src/Cli/dotnet/commands/dotnet-workload/clean/LocalizableStrings.resx
```diff
@@ -0,0 +1,137 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="CleanAllOptionDescription" xml:space="preserve">
+    <value>Clean up and uninstall all CLI workloads for every SDK feature band (version) that is less than or equal to the current running SDK.</value>
+  </data>
+  <data name="CommandDescription" xml:space="preserve">
+    <value>Remove orphaned workload components that may be on the machine.</value>
```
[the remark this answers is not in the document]
Eilon: Also instead of "orphaned" can we use a more precise term? What exactly does that mean here? A pack with no matching installer? Or the reverse? Or something else entirely?

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
```diff
@@ -0,0 +1,132 @@
...
+#pragma warning restore CS8604
+
+                        InstalledWorkloadsCollection vsWorkloads = new();
+                        VisualStudioWorkloads.GetInstalledWorkloads(bandedResolver, vsWorkloads, _cleanAll ? null : new SdkFeatureBand(sdkVersion));
+                        foreach (var vsWorkload in vsWorkloads.AsEnumerable())
+                        {
+                            vsWorkloadUninstallWarnings.Add(string.Format(LocalizableStrings.VSWorkloadNotRemoved, $"{vsWorkload.Key} ({vsWorkload.Value})"));
+                        }
+                    }
+                    catch (Exception ex)
```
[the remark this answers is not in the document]
nagilson: Thank you for this suggestion. I have made it catch one specific error type which is WorkloadManifestException. Before if you give a bad path or a path to a different SDK that did have manifests then the manifest reader would fail. I was worried about other things that could happen if the path is wrong and tried to account for that better instead of just catching all exceptions.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
nagilson: I don't think the entire directory can be deleted with just the `File.Delete()`. The existing code may delete the record, and the it may delete the pack folders if no packs remain. But the pack record folders can still remain even if they are empty. If that's intentional or sounds incorrect please let me know.

Before this line of code, which is in this commit: https://github.com/dotnet/sdk/pull/30266/commits/4c22bf897fa34f5673a444ed9561fa49008918df
When running `workload list` the workload would still appear after its been deleted, but after this change it did not appear in list, so I assume it is needed.

nagilson: Im going to reverify this. Maybe its because the installation record for the workload wasn't being cleaned up.

src/Cli/dotnet/commands/dotnet-workload/clean/LocalizableStrings.resx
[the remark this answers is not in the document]
nagilson: Ya. The word "orphaned" is not great.

> Also, should we say "workload packs" instead of "workload components"?

I don't think so. `clean` also cleans pack installation records. `clean --all` also cleans all of that + workload installation records. But I doubt most customers know what those are. Hence `components`.

> A pack with no matching installer? 

No matching installer or no matching workload installation record. Or if clean all, every pack.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallRecords/RegistryWorkloadInstallationRecordRepository.cs
nagilson: Well. https://softwareengineering.stackexchange.com/questions/103031/should-i-remove-unreferenced-code The answer is pretty clear. Let's just remove this.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -307,30 +308,42 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
+                                    (!installedSdkFeatureBands.Contains(featureBand) || cleanAllPacks );
+                            }
+                    );
 
+                    // Mark packs for GC that don't have a corresponding workload installation record (or all packs).
                     var currentBandRecordPath = Path.Combine(packVersionDir, _sdkFeatureBand.ToString());
-                    if (bandRecords.Contains(currentBandRecordPath) && !currentBandInstallRecords.Contains(currentBandRecordPath))
+                    if (
+                        bandRecordPaths.Contains(currentBandRecordPath) && // A pack for this band exists.
+                        new SdkFeatureBand(Path.GetFileName(currentBandRecordPath)).CompareTo(_sdkFeatureBand) < 1 && // We can parse the format of that band (its older than our band or current.)
```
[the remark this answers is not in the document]
nagilson: > CompareTo results should be compared against zero.

I am confused. `band.CompareTo(band) == 0`. It is meant to be if the feature band is equal to or less than. If it is < 0 then it is only doing less than the current band. Do you mean `<=`? 

But sounds good. Let's just do it for every band.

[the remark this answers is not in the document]
dsplaisted: You're right, it should be `<= 0`, and since `CompareTo` returns an integer, `<= 0` is the same as `< 1`.  I was thinking the value might be a `double`.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
nagilson: This code is indeed needed to delete the pack installation record. We also cannot move it out of this location because we don't want to remove the pack record if other bands still have the pack.

dsplaisted: To me it seems like the files will already have been deleted when it gets here.  That means that when `DeletePackInstallationRecord` is called, the `File.Exists` check in that method will be false, so it won't actually run the code that would delete the higher level directories.

Does this make sense?  Am I missing something?

src/Cli/dotnet/commands/dotnet-workload/clean/LocalizableStrings.resx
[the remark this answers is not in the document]
nagilson: I should be more precise with my language. It will only clean the packs of the same SDK type (file based or MSI based if the SDK is in this mode.) And it will tell for VS workloads how to remove them in VS as it can't remove those. The current code also only removes for feature bands at or below the current band, but I'm about to change that limitation.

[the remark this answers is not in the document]
nagilson: Per @baronfel,
`Removes workload components from this SDK feature band that may have been left behind from previous updates and uninstallations.`

`--all` --> `Causes clean to remove all workload components from all SDK versions.`

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
nagilson: Yea, whoops.... Looks like this is a no-op. Thank you

nagilson:
@dsplaisted @joeloff I have responded to all feedback, including making it work for > feature bands. If we would like to revert that part, it is a simple commit revert. Thank you.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -314,19 +314,16 @@ public void EradicateAllWorkloadInstallationRecords(SdkFeatureBand currentFeatur
...
-                {
-                    Log?.LogMessage($"'{potentialBandToClean}' is confirmed to be less than or current to the current band.");
-
-                    var workloadInstallationRecordIds = RecordRepository.GetInstalledWorkloads(potentialBandToClean);
-                    foreach (WorkloadId workloadInstallationRecordId in workloadInstallationRecordIds)
-                    {
-                        Log?.LogMessage($"Workload {workloadInstallationRecordId} for '{potentialBandToClean}' has been marked for deletion.");
-                        RecordRepository.DeleteWorkloadInstallationRecord(workloadInstallationRecordId, potentialBandToClean);
-                    }
+                Log?.LogMessage($"'{potentialBandToClean}' is confirmed to be less than or current to the current band.");
```
joeloff, line 317: Should that be `less than or equal`?

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -348,6 +357,26 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
                 }
             }
+
+            if (cleanAllPacks)
+            {
+                EradicateAllWorkloadInstallationRecords(_sdkFeatureBand);
+            }
+        }
+
+        public void EradicateAllWorkloadInstallationRecords(SdkFeatureBand currentFeatureBand)
```
dsplaisted, line 367: Looks like the parameter isn't used any more so we can just delete it, right?

Also, I'd suggest using "Delete" or "Remove" in the name instead of "Eradicate".

```suggestion
        public void RemoveAllWorkloadInstallationRecords()
```

src/Cli/dotnet/commands/dotnet-workload/install/IInstaller.cs
```diff
@@ -20,7 +20,13 @@ internal interface IInstaller : IWorkloadManifestInstaller
...
         void RepairWorkloads(IEnumerable<WorkloadId> workloadIds, SdkFeatureBand sdkFeatureBand, DirectoryPath? offlineCache = null);
 
-        void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = null);
+        void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = null, bool cleanAllPacks = false);
+
+        /// <summary>
+        /// Remove all non-VS workload installation records for the specified feature band, and all records for feature bands lower than featureBand.
+        /// </summary>
+        /// <param name="currentFeatureBand">The maximum feature band to remove all workload installation records from.</param>
+        void EradicateAllWorkloadInstallationRecords(SdkFeatureBand currentFeatureBand);
```
dsplaisted, line 29: It looks like this method is only called from inside each `IInstaller` implementation.  So I think it doesn't need to be added to the interface, and doesn't need to be public either.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
nagilson: Good point. I should've removed this line.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
nagilson: Thank you! Yeah, that function name was not great.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -307,30 +306,40 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
+                    var jsonPackInfo = File.ReadAllText(unneededBandRecordPaths.First());
+                    foreach (var unneededRecord in unneededBandRecordPaths)
                     {
                         File.Delete(unneededRecord);
                     }
 
-                    if (!bandRecords.Except(unneededBandRecords).Any())
+                    if (!bandRecordPaths.Except(unneededBandRecordPaths).Any())
                     {
                         Directory.Delete(packVersionDir);
```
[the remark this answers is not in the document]
nagilson: I believe this is equivalent to the existing code:

```cs
 if (!Directory.GetFileSystemEntries(packIdDir).Any())
                {
                    Directory.Delete(packIdDir);
                }
```

Do you prefer to get rid of this and move it into this function?

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -307,30 +306,40 @@ public void GarbageCollectInstalledWorkloadPacks(DirectoryPath? offlineCache = n
...
+                    var bandRecordPaths = Directory.GetFileSystemEntries(packVersionDir);
 
-                    var unneededBandRecords = bandRecords
-                        .Where(recordPath => !installedSdkFeatureBands.Contains(new SdkFeatureBand(Path.GetFileName(recordPath))));
+                    // Mark packs for GC from versions / feature bands of the SDK that are not present on the machine.
+                    var unneededBandRecordPaths = bandRecordPaths
+                        .Where(recordPath =>
+                            {
+                                var featureBand = new SdkFeatureBand(Path.GetFileName(recordPath));
+                                return !installedSdkFeatureBands.Contains(featureBand) || cleanAllPacks;
```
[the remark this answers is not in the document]
nagilson: I have changed the suggestion slightly.

```cs
                    var unneededBandRecordPaths = bandPackRecordPaths
                        .Where(recordPath =>
                            {
                                var thisRecordFeatureBand = new SdkFeatureBand(Path.GetFileName(recordPath));

                                // Mark the pack record for garbage collection if we're cleaning all packs.
                                if (cleanAllPacks)
                                {
                                    return true;
                                }

                                // Mark the pack record for garbage collection if the feature band is not installed/has no workloads installed.
                                if (!installedSdkFeatureBands.Contains(thisRecordFeatureBand))
                                {
                                    return true;
                                }

                                // Mark the pack record for garbage collection if the pack has no corresponding workload installation record and has the current SDK feature band.
                                var currentBandRecordPath = Path.Combine(packVersionDir, _sdkFeatureBand.ToString());
                                if (thisRecordFeatureBand.Equals(_sdkFeatureBand) && !currentBandInstallRecords.Contains(currentBandRecordPath))
                                {
                                    return true;
                                }

                                return false;
                            }
                    );
```

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <summary>
+        /// Print VS Workloads with the same machine arch which can't be uninstalled through the SDK CLI to increase user awareness that they must uninstall via VS.
+        /// </summary>
+        private void DisplayUninstallableVSWorkloads()
+        {
+#if !DOT_NET_BUILD_FROM_SOURCE
+            if (OperatingSystem.IsWindows())
+            {
+                // All VS Workloads should have a corresponding MSI based SDK. This means we can pull all of the VS SDK feature bands using MSI/VS related registry keys.
+                var installedSDKVersionsWithPotentialVSRecords = MsiInstallerBase.GetInstalledSdkVersions();
```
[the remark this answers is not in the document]
nagilson: That's true. But I think it's better to show it in both cases. I don't think the average user will be aware of the difference of whether their SDK is in file-based or MSI based mode. And they might be confused as to why things from VS weren't removed if their machine is still in a bad state after cleaning with `--all`. IMO this messaging from us reduces the confusion, but it may be annoying to see the message pop up for every VS workload if you don't intend to mess with them.

Is the main concern about accessing the registry (elevation concerns) in a file-based mode? I (think*) registry reads are generally [permitted](https://superuser.com/questions/308693/if-i-am-not-an-administrator-but-i-want-to-read-the-registry-what-is-easiest-wa#:~:text=Non%2Dadministrators%20have%20the%20same,Open%20the%20Windows%20Start%20Menu) regardless of user-privelleges. Please let me know if you disagree.

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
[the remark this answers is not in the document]
dsplaisted: I think it would go in both places, unless you do more refactoring on the `DeletePackInstallationRecord` method.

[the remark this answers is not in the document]
nagilson: As confirmed in the call, we don't need to do this delete as it would be a no op

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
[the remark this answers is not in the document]
nagilson: Sorry, didn't notice we can just use recordPath and don't need to construct currentBandRecordPath anymore 🤣

src/Cli/dotnet/commands/dotnet-workload/clean/WorkloadCleanCommand.cs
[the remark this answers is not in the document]
nagilson: We decided to limit this to MSI because another concern is that MSI doesnt clean up file based and vice versa and we don't want to overlap those two worlds. Also, custom dotnet directory worlds should ideally not overlap either, you don't want to affect, say, another sdk pinned by a global.json

nagilson:
@dsplaisted I have responded to all of the feedback, thanks for taking a look. I know you still need to look @ the tests.

src/Tests/dotnet.Tests/dotnet-workload-clean/GivenDotnetWorkloadClean.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        [Theory]
+        [InlineData(true, true)]
+        [InlineData(false, true)]
+        [InlineData(true, false)]
+        [InlineData(false, false)]
+        public void GivenWorkloadCleanFileBasedItRemovesPacksAndPackRecords(bool userLocal, bool cleanAll)
+        {
+            // Setup ( I wish we could not copy this everywhere, but these are convenient as locals. )
```
joeloff, line 47: Should probably remove/clean up this comment and then maybe file an issue if you think there's work required here.

nagilson: I don't think there's anything we can easily do, this is only copied two times and its just a lot of local variables that are needed for the test. Technically we could put it in a class but that'd end up being more loc/complexity than it currently is. (I tried it and got rid of it because it was worse.) The comment is basically an admission of 'this sucks, but dont waste time trying to make it better'

joeloff (approved):

src/Tests/dotnet.Tests/dotnet-workload-clean/GivenDotnetWorkloadClean.cs
```diff
@@ -0,0 +1,198 @@
...
+            AssertExtraneousPacksAreRemoved(extraPackPath, extraPackRecordPath);
+            AssertValidPackCountsMatchExpected(installRoot, expectedPackCount: cleanAll ? 0 : 7, expectedPackRecordCount: cleanAll ? 0 : 8);
+
+            AssertAdjacentCommandsStillPass(userProfileDir, dotnetRoot, testDirectory, workloadResolver, nugetDownloader);
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void GivenWorkloadCleanAllFileBasedItCleansOnlyOlderAndCurrentFeatureBands(bool userLocal)
```
dsplaisted, line 79: Is this test name correct?  I think we changed it so that clean all would remove workloads from future bands, and the last line of the test verifies that no packs are installed.  So probably the name should be updated to something like `GivenWorkloadCleanAllFileBasedItCleansOlderAndNewerFeatureBands`

dsplaisted (approved):
Looks good.

nagilson:
@marcpopMSFT  pls merge on red, arcade outage seems to be blocking this pr

marcpopMSFT:
Do we have a link to a known outage we can ping them on?

nagilson:
I believe it's fixed now, so no action needed :)

PR #30345, opened 2023-02-04
ApiCompat refactoring and small features

### LINKED ISSUE #28712: API Compat implements functionality for checking internal API but does not expose it
### Describe the bug
As far as I can tell we don't expose a way to enable internal API checking.  In the old tool this was done with `--respect-internals` which enabled checking of internals for compatibility when the assembly happened to have an  `InternalsVisibleTo` attribute on it.

### To Reproduce
examine parameters for API Compat task / tool

### Further technical details
https://github.com/dotnet/arcade/blob/3d988e17f3317c888c82e21b4b3dd5196dfc4d78/src/Microsoft.DotNet.ApiCompat/src/ApiCompatRunner.cs#L64-L68

https://github.com/dotnet/arcade/blob/3d988e17f3317c888c82e21b4b3dd5196dfc4d78/src/Microsoft.DotNet.ApiCompat/src/Microsoft.DotNet.ApiCompat.Core/Executor.cs#L122-L124

I'm not sure we need this, but we do seem to be carrying a lot of code to implement and test for this functionality.  That could be deleted if we didn't want to support it.

ViktorHofer:
**Refactoring:**
- Use ApiSymbolExtensions filtering (composite + attribute filter)
- Move ApiCompat mapping infrastructure into ApiCompatibility.Mapping namespace and subfolder.
- Move ApiCompat comparer into ApiCompatibility.Comparer namespace and subfolder. Expose AttributeData, TypedConstant and NamedArgument comparers as IEqualityComparers<T> and make them generally usable outside of the AttribuesMustMatch rule.
- Move MapperSettings and RuleSettings into interfaces and make ApiComparer settings inherit them. That guarantees that settings are in sync and allows passing in filters and components from the frontend layer (i.e. when configuring the attribute filter).
- Remove ApiCompatibility.Abstractions namespace in favor of the newly introduced namespaces. Remove the using statement from tests.
- Update xml documentation
- Address typos that are indicated by the repo's spell checker

**Features:**
- Expose "RespectInternals" flag that existed in the legacy ApiCompat tool. Fixes https://github.com/dotnet/sdk/issues/28712

**Bugfix:**
- Avoid a crash in the RuleRunner when both left and right are null. That can happen when different assembly identities are marked as compatible.


ViktorHofer:
cc @andriipatsula as I'm touching ApiSymbolExtensions and GenAPI here as well (just a few files).

andriipatsula:
> cc @andriipatsula as I'm touching ApiSymbolExtensions and GenAPI here as well (just a few files).

The GenAPI changes look good to me.

smasher164 (approved):

Issue #16292, opened 2021-03-10
Update SDK tests to supported TFM's


danmoseley:
Many tests target out of support TFM's and few target the latest TFM.

eg
```
      (_CheckForEolTargetFrameworks target) -> 
        C:\h\w\95ED08C8\p\d\sdk\6.0.100-ci\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp1.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [C:\h\w\95ED08C8\t\dotnetSdkTests\kviqdfge.mx1\It_packs_nond---BFC73927\NetStandardAndNetCoreApp\NetStandardAndNetCoreApp.csproj]
```
```
        [RequiresSpecificFrameworkFact("netcoreapp1.1")]
        public void IfAToolHasNotBeenRestoredForNetCoreApp2_0ItFallsBackToNetCoreApp1_x()
```
```
 334          [InlineData("Major", "netcoreapp3.0", true)]
  335:         [InlineData("Major", "netcoreapp2.0", true)]
```

Random examples pasted above from searching the SDK tree. I assume there will need to be a few cases of obsolete TFM's since customers will try to build projects containing them, but almost all our tests should be testing supported combinations.

marcpopMSFT:
First step is in this PR. https://github.com/dotnet/sdk/pull/19708  Adding a note here to remove DisableImplicitNamespaceImports once it's no longer needed.

marcpopMSFT:
Completed for all TestProjects (or close enough). Dynamic project files are next.

marcpopMSFT:
@v-wuzhai I wanted to check in here. You've done a lot of PRs and I've noticed you've moved on to RIDs. Are there any remaining items to fix for the target framework or are we close enough to call it done? We don't need to be 100% as long as most are updated.

v-wuzhai:
@marcpopMSFT All the files in the TestProjects and Test folders have been updated which include TFM's and RID's, except for the PRs of #26821、#25479 and #22784 which are not been merged.

marcpopMSFT:
Since the vast majority of PRs are now in. I'll close this as we clean up the last few.

PR #14094, opened 2020-10-10
Skip installing SDK 1.0.5 and 1.1.2 on Ubuntu 18+

am11:
SDK v1.x Ubuntu package has 16.04 hardcoded as version, which breaks build on higher LTS versions (18.04, 20.04)

This patch fixes the issue by skipping SDK v1.x installation on unsupported LTS versions, and has no effect on non-Ubuntu Linux distros or other operating systems.

Tested with v16.04 and v20.04.

Fixes #3768

cc @wli3, @wfurt

PR #30379, opened 2023-02-07
fix repo build on non-ubuntu linux

smasher164:
`eng/restore-tools.{sh, ps1}` tries to fetch runtime versions 1.0.5 and 1.1.2 on
non-ubuntu linuxes. Since the web resource for those builds return 404, a
fresh build of the sdk can't succeed on non-ubuntu linuxes. These builds could
be made available again, but since they're manually installed for testing, it's
better to remove them from the restore script.

### DIFF
--- a/eng/restore-toolset.sh
+++ b/eng/restore-toolset.sh
@@ -21,10 +21,6 @@ function InitializeCustomSDKToolset {
 
   InitializeDotNetCli true
   
-  if [[ "$DISTRO" != "ubuntu" || "$MAJOR_VERSION" -le 16 ]]; then
-    InstallDotNetSharedFramework "1.0.5"
-    InstallDotNetSharedFramework "1.1.2"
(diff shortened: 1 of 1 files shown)


ericstj:
Looks like @marcpopMSFT mentioned in https://github.com/dotnet/sdk/pull/14094#issuecomment-708667530 that we should just remove 1.x.  I couldn't find an issue tracking this - @marcpopMSFT is there anything else that should be cleaned up as part of this?

marcpopMSFT:
@ericstj the issue was this one: https://github.com/dotnet/sdk/issues/16292 @v-wuzhai upgraded most of the tests that he could over the past year. There are still some targeting 1.0 but it appears they don't require the runtime to be present to pass as this PR passed.

I think it's safe to merge this.

marcpopMSFT (approved):

Issue #30395, opened 2023-02-07
[NETSDKE2E] With NET8 preview 1 SDK installed, component names display incorrectly when installing workload.


v-ainigao:
Repro steps:

1.Install .NET 8 SDK (8.0.100-preview.1.23106.1) from main branch of https://github.com/dotnet/installer
2.Install workload in CLI
dotnet workload install wasm-tools

Expected Result:
workload can be installed successfully and the output shows correctly.

Actual Result:
component names display incorrectly when installing workload.
![MicrosoftTeams-image (11)](https://user-images.githubusercontent.com/106054229/217171241-ab4aae1c-f4d7-49c0-b839-ea9a3ba181cf.png)

dotnet --info:

.NET SDK:
Version:   8.0.100-preview.1.23106.1
Commit:    0e4c551aeb Runtime Environment:
OS Name:     Windows
OS Version:  10.0.20348
OS Platform: Windows
RID:         win10-x64
Base Path:   C:\Program Files\dotnet\sdk\8.0.100-preview.1.23106.1\ Host:
  Version:      8.0.0-preview.1.23104.3
  Architecture: x64
  Commit:       6507a44770 .NET SDKs installed:
  8.0.100-preview.1.23106.1 [C:\Program Files\dotnet\sdk] .NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.0-preview.1.23105.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.0-preview.1.23104.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 8.0.0-preview.1.23102.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App] Other architectures found:
  x86   [C:\Program Files (x86)\dotnet] Environment variables:
  Not set global.json file:
  Not found Learn more:https://aka.ms/dotnet/info 
  Download .NET:https://aka.ms/dotnet/download

marcpopMSFT:
Can we unhash them to display the right names? Is there a hash to real name mapping somewhere? Should we hide these names?

joeloff:
I'm testing a fix now

WenJunJi01:
This issue is also reproduced on win arm64.
![image](https://user-images.githubusercontent.com/87680983/217497003-d22499bb-b762-47bf-a683-51ecd616f500.png)

[...]

PR #30420, opened 2023-02-08
Log MSI package ID instead of MSI filename

joeloff:
## Description

Workload pack MSIs now all have fixed length names (MD5 hash based on the workload pack type, etc.). When installing these from the CLI the user won't know to which packs the MSIs belong. See #30395 for an example of the output.

The fix for this is to report the pack IDs.


dsplaisted (approved):

nagilson (approved):

joeloff:
Thank you

PR #30476, opened 2023-02-09
GenAPI: update event declaration and reference parameters.

andriipatsula:
Epic: https://github.com/dotnet/arcade/issues/10291

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxGeneratorExtensions.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxGeneratorExtensions.cs
@@ -66,17 +66,13 @@ namespace Microsoft.DotNet.GenAPI
                 }
             }
 
-            if (symbol.Kind == SymbolKind.Event)
+            if (symbol is IEventSymbol eventSymbol && !eventSymbol.IsAbstract)
             {
-                IEventSymbol eventSymbol = (IEventSymbol)symbol;
-                if (!eventSymbol.IsAbstract)
-                {
-                    // adds generation of add & remove accessors for the non abstract events.
-                    return syntaxGenerator.CustomEventDeclaration(eventSymbol.Name,
-                        syntaxGenerator.TypeExpression(eventSymbol.Type),
-                        eventSymbol.DeclaredAccessibility,
-                        DeclarationModifiers.From(eventSymbol));
-                }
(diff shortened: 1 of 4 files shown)


src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxRewriter/BodyBlockCSharpSyntaxRewriter.cs
```diff
@@ -48,7 +48,9 @@ public BodyBlockCSharpSyntaxRewriter(string? exceptionMessage)
             }
 
             string returnType = rs.ReturnType.ToString();
-            if (returnType != "void" && returnType != "System.Void")
+            if (returnType != "void" && returnType != "System.Void" ||
```
ViktorHofer, line 51: This is easy to get/read wrong. Please add braces. I assume it should be `((returnType != "void" && returnType != "System.Void") || X)`

ViktorHofer (approved):

src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxGeneratorExtensions.cs
```diff
@@ -66,6 +66,19 @@ public static SyntaxNode DeclarationExt(this SyntaxGenerator syntaxGenerator, IS
                 }
             }
 
+            if (symbol.Kind == SymbolKind.Event)
```
smasher164, line 69: You can probably replace this `if` with `if (symbol is IEventSymbol eventSymbol)`, and avoid the nested conditional

src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxRewriter/BodyBlockCSharpSyntaxRewriter.cs
andriipatsula: Agree. Thanks

src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxGeneratorExtensions.cs
andriipatsula: Updated:
```C#
if (symbol is IEventSymbol eventSymbol && !eventSymbol.IsAbstract) {...}
```

PR #30486, opened 2023-02-10
Add design proposal for workload history

dsplaisted:


documentation/general/workloads/workload-history.md
```diff
@@ -0,0 +1,44 @@
...
+- Success or failure, and error message on failure
+- Workload state before and after command run
+  - Manifest versions (rollback file)
+  - Installed workloads
+
+Each installation record will be stored as a file in a folder specific to the SDK feature band.  For MSI-based installs, this folder will be `%PROGRAMDATA%\dotnet\workloads\{sdk-band}\history\`.  For File-based installs, it will be `{DOTNET ROOT}\metadata\workloads\{sdk-band}\history\`.
+
+## `dotnet workload history`
+
+The `dotnet workload history` command will show the history of workload installation commands.  The output will be similar to the following format:
```
joeloff, line 24: Should we include the success/failure indicator in the table as well?

documentation/general/workloads/workload-history.md
```diff
@@ -0,0 +1,44 @@
...
+
+## `dotnet workload update --from-history
+
+A `--from-history` option will be added to the `dotnet workload update` command, which will update to a state from the workload history.  This will require one of two new parameters, `--before` or `--after`, to be specified.  By default the command will update the manifests to the versions specified, as well as install or uninstall workloads to match the workloads that were installed at the specified point.  A `--manifests-only` option will be supported which will update the manifests, but not change the current workloads that are installed (this may result in installing or uninstalling packs to match the new manifest versions).
+
+For example:
+
+`dotnet workload update --from-history --before 3 --manifests-only` - This will roll back to the manifests before operation 3 in the workload history.
```
joeloff, line 41: So `before` walks back and selects the first operation that was successful? For example, if you did `--before 8` and 6 and 7 failed, but 5 succeeded, it will select 5? Would `--after 2` select the last known successful state (if 3 and 5 were successful and 4 failed, it would pick 5)?

Forgind: I'd say --before picks the latest successful operation before the id specified, so in your case, it would pick 5, but if 1/2/3/4 were all successful, it would still pick 5. After would pick the first successful one after that, so 3 for your case or 2 if that one was successful. (There has to be some way to say "pick 2 if possible", and after seems more appropriate than before.) Does that sound reasonable to you?

documentation/general/workloads/workload-history.md
```diff
@@ -0,0 +1,44 @@
+# Workload History
+
+When .NET SDK workloads are updated, it is possible that the new versions of the workloads won't work (if we shipped a bug) or won't be available (if the manifest updates are available but the packs themselves are not).  Right now it is hard to fix such a state.  We will add a `dotnet workload history` command that will show a history of the workload installation actions that have been run via the .NET CLI, and add new options on the `dotnet workload update` command that will allow rolling back to a previously installed version of workloads from the history.
+
+## Recording installation actions
+
+When a CLI command is run which modifies the installed workloads or updates the manifests, we will write a record to the workload history log.  The workload commands this applies to include `install`, `update`, `uninstall`, `repair`, and `restore`.  Each workload history record will include the following information:
+
+- Date/time command was run
```
Forgind, line 9: Do we really need all this information? I'm wondering why we need both time run and time completed, I'm wondering if we should be keeping failures _at all_ (given that they don't actually update workload state and we aren't rolling back to them), and I'm a bit iffy on the rollback file/workload state parts, not because those aren't reasonable things to include but because those are potentially _big_, and a file that's intended to persist and never go is a major problem if it grows that fast. We presumably aren't intending to delete it at any point (or what's the point of having the file in the first place?), so I can easily imagine this becoming a problem, especially if we really do include all this information. In particular, imagine if I specify the wrong file by accident as my rollback file. It's 10 GB. I include the whole thing in this file, but the command fails. I don't immediately realize what my mistake was, so I run it 2-3 more times, adding another 20-30 GB to the history file. Do we really want a 40 GB file here? And I would argue that if the user then deletes the file as "why do I have this massive thing in my SDK?", we've failed. I think it should ideally be as streamlined as possible to prevent users from ever hitting that problem unless they run 100 million commands.

documentation/general/workloads/workload-history.md
dsplaisted: No, what I was going for was that `--before n` would give you the state immediately before command n was executed, regardless of whether it succeeded or failed.  If command failed, then ideally it would roll back any changes, and `--before n` and `--after n` would mean the same thing.

Basically each record in the history records the state before and after the command was executed.  The number specified selects which record to use, and whether you specify `--before` or `--after` selects which part of the state in that record to use.  Whether a command was successful or not doesn't affect the logic.

Does this make sense?  If this wasn't obvious from my description, maybe it's also going to be confusing to developers and we should try to find a different way to express these operations.

documentation/general/workloads/workload-history.md
dsplaisted: None of this should be big.  The workload state isn't supposed to include the actual contents of the workload packs or even the manifests.  "Installed workloads" just means the list of names of the workloads that are installed.  To roll back to a given state may still require downloading the workload packages from NuGet feeds, we aren't trying to avoid that.

Time run and time completed lets us know how long the command took to run, which could be useful.  I think we should try to gather all this information as well as other information that we think might be useful.  We'll be writing a separate data file for each command, and I would guess that each file would be 1 KB or less.

I think it's very unlikely that someone would accidentally specify a rollback file that was very large but still parsed correctly as a rollback file.  We don't need to include the rollback file contents verbatim, just what we parsed out of it successfully.  And if it fails to parse, then that's a failure that's early enough that I agree we wouldn't need to write a workload history record.  If we actually start to try downloading something or installing something, I think we should write the record even if the operation ends up failing.

marcpopMSFT:
Is this superseded by #36809 ? @Forgind @dsplaisted

Forgind:
> Is this superseded by #36809 ? @Forgind @dsplaisted

Yes

PR #30539, opened 2023-02-12
GenAPI: add readonly keyword for synthesized fields.

andriipatsula:
Epic: https://github.com/dotnet/arcade/issues/10291

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
@@ -71,16 +71,22 @@ namespace Microsoft.DotNet.GenAPI
         }
 
         // Build dummy field from a type, field name, and attribute list.
-        private static SyntaxNode CreateDummyField(string typ, string fieldName, SyntaxList<AttributeListSyntax> attrs) =>
-            SyntaxFactory.FieldDeclaration(
+        private static SyntaxNode CreateDummyField(string typ, string fieldName, SyntaxList<AttributeListSyntax> attrs, bool isReadonly)
+        {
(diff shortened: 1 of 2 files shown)


src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
```diff
@@ -71,16 +71,22 @@ private static SyntaxList<AttributeListSyntax> FromAttributeData(IEnumerable<Att
         }
 
         // Build dummy field from a type, field name, and attribute list.
-        private static SyntaxNode CreateDummyField(string typ, string fieldName, SyntaxList<AttributeListSyntax> attrs) =>
-            SyntaxFactory.FieldDeclaration(
+        private static SyntaxNode CreateDummyField(string typ, string fieldName, SyntaxList<AttributeListSyntax> attrs, bool isReadonly)
+        {
+            List<SyntaxToken> modifiers = new List<SyntaxToken> { SyntaxFactory.Token(SyntaxKind.PrivateKeyword) };
```
smasher164, line 76: If we're gonna use `readonly` everywhere, then we don't need the `if` statement. We can just pass in
```C#
WithModifiers(SyntaxFactory.TokenList(new[] { SyntaxFactory.Token(SyntaxKind.PrivateKeyword), SyntaxFactory.Token(SyntaxKind.ReadOnlyKeyword) })
```

andriipatsula: I checked the CCI-based GenAPI and it does not always include the `readonly` keyword. But, I believe we could use this keyword in all cases. @ViktorHofer , what do you think?

smasher164 (approved):

PR #30549, opened 2023-02-13
Bump up dependencies and make them fit for Dependabot

GangWang01:
### Problem
The version of Verify.Xunit and Verify.DiffPlex is lower than the ones in dotnet/templating. It causes Detected package downgrade error during building.

### Solution
Keep the version same as dotnet/templating. And make change for utilizing Dependabot.


YuliiaKovalova:
@GangWang01 , I was fixing the same issue here
https://github.com/dotnet/sdk/pull/30247/commits/a78fa47f543d066d9fd4931e58db645929625c68

YuliiaKovalova (approved):

PR #30572, opened 2023-02-13
GenAPI: add workaround for IFieldSymbol.Name that has angle brackets

smasher164:
Roslyn can (surprisingly?) return IFieldSymbols whose names are prefixed with `<T>` type parameters. For now, replace angle brackets with underscores before emitting a synthesized field.

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
@@ -113,17 +113,9 @@ namespace Microsoft.DotNet.GenAPI
                 // Add a dummy field for each generic excluded field
                 foreach(IFieldSymbol genericField in genericTypedFields)
                 {
-                    String fieldName = genericField.Name;
-                    int i = fieldName.LastIndexOf('>');
-
-                    if (i >= 0)
-                    {
-                        fieldName = fieldName.Remove(0, i+1);
-                    }
-
                     yield return CreateDummyField(
                         genericField.Type.ToDisplayString(),
-                        fieldName,
+                        genericField.Name.Replace('<', '_').Replace('>', '_'),
                         FromAttributeData(genericField.GetAttributes()),
                         namedType.IsReadOnly);
                 }
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
@@ -113,9 +113,17 @@ namespace Microsoft.DotNet.GenAPI
                 // Add a dummy field for each generic excluded field
                 foreach(IFieldSymbol genericField in genericTypedFields)
                 {
+                    String fieldName = genericField.Name;
+                    int i = fieldName.LastIndexOf('>');
+
+                    if (i >= 0)
+                    {
+                        fieldName = fieldName.Remove(0, i+1);
+                    }
+
                     yield return CreateDummyField(
                         genericField.Type.ToDisplayString(),
-                        genericField.Name,
+                        fieldName,
                         FromAttributeData(genericField.GetAttributes()),
                         namedType.IsReadOnly);
                 }
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
@@ -88,6 +88,11 @@ namespace Microsoft.DotNet.GenAPI
             return declaration;
         }
 
+        // Sometimes the metadata can contain names that are not valid C# identifiers. For example,
+        // to express a set of type parameters, they can be prefixed with angle brackets.
+        // Normalize them by replacing these special characters with '_'.
+        private static string NormalizeIdentifier(string s) => s.Replace('<', '_').Replace('>', '_');
+
         // SynthesizeDummyFields yields private fields for the namedType, because they can be part of the API contract.
         // - A struct containing a field that is a reference type cannot be used as a reference.
         // - A struct containing nonempty fields needs to be fully initialized. (See "definite assignment" rules)
@@ -115,7 +120,7 @@ namespace Microsoft.DotNet.GenAPI
                 {
                     yield return CreateDummyField(
                         genericField.Type.ToDisplayString(),
-                        genericField.Name.Replace('<', '_').Replace('>', '_'),
+                        NormalizeIdentifier(genericField.Name),
                         FromAttributeData(genericField.GetAttributes()),
                         namedType.IsReadOnly);
                 }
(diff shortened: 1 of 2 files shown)


ViktorHofer:
That sounds like a bug that we should fix directly in roslyn's CodeAnalysis. cc @CyrusNajmabadi

src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
```diff
@@ -115,7 +115,7 @@ public static IEnumerable<SyntaxNode> SynthesizeDummyFields(this INamedTypeSymbo
                 {
                     yield return CreateDummyField(
                         genericField.Type.ToDisplayString(),
-                        genericField.Name,
+                        genericField.Name.Replace('<', '_').Replace('>', '_'),
```
ViktorHofer, line 118: I think you want to use `MetadataName` instead: https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.isymbol.metadataname?view=roslyn-dotnet

andriipatsula: Alternatively, you may try to create a `CSharpSyntaxRewriter` like I did in https://github.com/dotnet/sdk/blob/ac4bf62fff612626e5bca17178d52dabcc915323/src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxRewriter/TypeForwardAttributeCSharpSyntaxRewriter.cs

ViktorHofer: Presumably, syntax rewriters are heavy weight as they are invoked for every single symbol during the traversal.

smasher164: This field name is present in the metadata. Additionally, `ToDisplayString` is not trying to produce a valid identifier.
It's not exactly the fault of Roslyn that it's preserving that information. However, I'm gonna reach out to them to see if there's an alternative way.

That being said, I spoke to Eric about this approach, and he seems alright with it.

CyrusNajmabadi:
> Roslyn can (surprisingly?) return IFieldSymbols whose names are prefixed with <T> type parameters. 

This does not seem surprising to me.  Fields in metadata can have those names.

src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
andriipatsula:
```C#
    public readonly partial struct MethodSignature<TType>
    {
// Error:
        private readonly TType <ReturnType>k__BackingField;
// Error:
        private readonly Collections.Immutable.ImmutableArray<TType> <ParameterTypes>k__BackingField;

        private readonly object _dummy;
        private readonly int _dummyPrimitive;

        public Collections.Immutable.ImmutableArray<TType> ParameterTypes { get { throw null; } }
        public TType ReturnType { get { throw null; } }
    }
```
@smasher164 , one more example of incorrectly generated fields. We may check together the solution

smasher164: @andriipatsula Are you sure you're using this branch to test that scenario? For the same case, I'm seeing fields named `_ReturnType_k__BackingField` and `_ParameterTypes_k__BackingField`.

ericstj: As Akhil mentioned this case is one where the correct thing in the metadata actually contains characters that cannot be represented in C#.  This is because it's not *normally* represented in C# since it's the backing field for the [auto-implemented property](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties).

Since we're replacing those auto properties with throwing method bodies, we don't get the compiler generated backing field, we have to represent one in C#.

Because of this we landed on a 1-way transformation of the member name that's unlikely to have collisions - just strip all unsupported characters and replace them with `_`.

I think for readability I'd pull this out to a method named something like `NormalizeIdentifier`, `EnsureValidIdentifier`, `SanitizeIdentifier` or something like that.

It might be more readable if such an method was driven off of an invalid character list to make it more readable, eg: `
var invalidChars = new [] { '<', '>', ...};`  Not like perf matters here (how many structs with generic fields are we really going to encounter? 😆), but you could probably use a StringBuilder if you needed to do replacement and then use IndexOfAny to walk through the input string, or just return the source if no invalidCharacters were found.

andriipatsula: > @andriipatsula Are you sure you're using this branch to test that scenario? For the same case, I'm seeing fields named `_ReturnType_k__BackingField` and `_ParameterTypes_k__BackingField`.

I took your PR and run the SBRP generator and could confirm that generated fields for the class `MethodSignature<TType>` from the example above, look like:

```C#
    public readonly partial struct MethodSignature<TType>
    {
        private readonly TType _ReturnType_k__BackingField;
        private readonly Collections.Immutable.ImmutableArray<TType> _ParameterTypes_k__BackingField;
// ...
}
```

CyrusNajmabadi: > Since we're replacing those auto properties with throwing method bodies,

why do that?  why not preserve them as autoprops?

smasher164: @ericstj I agree that `NormalizeIdentifier` would be the right way to extract this logic out, but I feel like for this case where I just need to operate on a single identifier with just angle brackets, it might be better to leave it as is, and do that refactoring later.

ericstj: > why do that? why not preserve them as autoprops?

This tool is meant to produce source files that replace method bodies and strip non-public API.  I believe it throws because throwing satisfies C# for any method signature while minimizing method body size.  Here's a reference for it does this, from the old codebase: https://github.com/dotnet/buildtools/pull/1849/files

> agree that NormalizeIdentifier would be the right way to extract this logic out

I'm not feeling strongly about re-implementing the current algorithm (indexof / stringbuilder) -- that's probably premature optimization.  I am feeling strongly about readability and maintainability.  What is the next person who looks at this code going to think when they see this arbitrary string replacement?  Why is it happening? 

I'd be ok with just factoring this single step outside the CreateDummyField line and commenting why it's happening.  I think it could be more self-evident if things were factored in a method but that's your call.

CyrusNajmabadi: > This tool is meant to produce source files that replace method bodies and strip non-public API. I believe it throws because throwing satisfies C# for any method signature while minimizing method body size. Here's a reference for it does this, from the old codebase: https://github.com/dotnet/buildtools/pull/1849/files

Fair.  I'm not really savvy about this domain.  But it did seem to me that preserving auto-props might be the simplest solution that also satisfies your needs.  I guess it depends on what yuo mean by `minimizing method body size`.  If you literally mean 'source code size', then auto-props are def the minimized form.  If yuo mean `IL size`, then i can see that an auto-prop's body might be larger (maybe??) than a throw-only impl.

ViktorHofer (approved):

PR #30573, opened 2023-02-13
Better fix for the OSX issues in installer repo

marcpopMSFT:
Revert #29916 as this is causing linux failures in installer now
Port fix for environmentinfo.
Realize that the fix won't work for 11.0
Modify the fix to work better following the idea of the other fix that almost worked

https://github.com/dotnet/sdk/pull/29916
https://github.com/dotnet/sdk/pull/29083

I manually tested various combinations of 10.12, 10.14, 10.15, 11.0, and 12


nagilson (approved):
Approved, thank you for adding the comment mapping SDK versions to OSX versions

Issue #30624, opened 2023-02-15
Breaking change inside v7.0.200 - MSBuild Preview version used !?


WernerMairl:
### Describe the bug
"dotnet build -o  <folder>" feature (output folder for entire solution build)

![image](https://user-images.githubusercontent.com/10084630/218982850-a374375e-26e0-4bc3-8416-4fe5b5c6bb22.png)

### To Reproduce

 use "dotnet build -o <folder>" with sdk 7.0.200 from today

### Exceptions (if any)
error NETSDK1194: The "--output" option isn't supported when building a solution.

### Further technical details
You can see in the screenshot that a preview version of MSBuild is used/delivered (7.5.0-preview....".
Maybe this is the root cause of the issue!
Because with the SDK from last week (7.0.100) i can workaround the exception!

I hope that helps - assuming this issue may cause a lot of problems...

BR
Werner

WernerMairl:
@KalleOlaviNiemitalo 
Nice fix. but following semantic versioning policy, this may be the wrong delivery :-(

PeterLehmann:
still a breaking change you have not write in the release note

dave-yotta:
No Semver? That's crazy! Do I have to avoid security patches just so that I can avoid breaking the build?

WernerMairl:
all in all... 
not a good idea to deliver this in this way....

we will see how many build pipelines worldwide are broken...

🍿 🍿  🍿

dave-yotta:
So if my runtime is `dotnet/aspnet:7.0` I need to somehow mirror that in an sdk version `dotnet/sdk:7.0.1**`?? There are no valid version strings for this in any supported tools I can think of!

WernerMairl:
In the same time Lufthansa is grounded for IT problems...

🍿 🍹

GregorLamche:
> So if my runtime is `dotnet/aspnet:7.0` I need to somehow mirror that in an sdk version `dotnet/sdk:7.0.1**`?? There are no valid version strings for this supported in any tools I can think of!

One reason why my team stopped using Microsoft docker images (MS is too slow to role out security updates of the base image).

As a work around you can switch to a OS base image, and install the sdk/runtime using its package manager.
With Ubuntu it would look like this:

```
FROM ubuntu:22.04 as base
RUN apt-get update \
    && apt-get install -y dotnet-sdk-7.0=7.0.102-1
```

dave-yotta:
[...]

PR #30660, opened 2023-02-15
Downgrade errors for specifying `--output` for solutions to warnings

dsplaisted:
# Downgrade errors for specifying `--output` for solutions to warnings

## Description

In 7.0.200, we [added an error](https://github.com/dotnet/sdk/pull/29065) when the `--output` option was specified for a solution.  This was because this causes multiple projects to use the same output path, which can lead to hard-to-diagnose issues.  It can also result in build flakiness if there are different projects copying the same files to the output path.

However, it appears that this change broke a lot more people than expected, so we will downgrade this error to a warning.

Additionally, we are going to remove the error/warning entirely from the `dotnet pack` command.  The `--output` parameter for `dotnet pack` only affects the package output path, and since each project should be generating a different .nupkg file, there shouldn't be a risk of file collisions when multiple projects share the same package output path.

## Customer impact

We have [multiple](https://github.com/dotnet/sdk/issues/30624) [reports](https://github.com/dotnet/dotnet-docker/issues/4424) of this breaking CI builds.

## Regression?
* [x]  Yes
* [ ]  No

## Risk

Low

## Verification

* [x]  Manual (required)
* [x]  Automated

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-pack/PackCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-pack/PackCommandParser.cs
@@ -24,7 +24,7 @@ namespace Microsoft.DotNet.Cli
         public static readonly Option<string> OutputOption = new ForwardedOption<string>(new string[] { "-o", "--output" }, LocalizableStrings.CmdOutputDirDescription)
         {
             ArgumentHelpName = LocalizableStrings.CmdOutputDir
-        }.ForwardAsOutputPath("PackageOutputPath");
+        }.ForwardAsSingle(o => $"-property:PackageOutputPath={CommandDirectoryContext.GetFullPath(o)}");
 
         public static readonly Option<bool> NoBuildOption = new ForwardedOption<bool>("--no-build", LocalizableStrings.CmdNoBuildOptionDescription)
             .ForwardAs("-property:NoBuild=true");
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -880,7 +880,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1193: "}</comment>
   </data>
   <data name="CannotHaveSolutionLevelOutputPath" xml:space="preserve">
-    <value>NETSDK1194: The "--output" option isn't supported when building a solution.</value>
+    <value>NETSDK1194: The "--output" option isn't supported when building a solution. Specifying a solution-level output path results in all projects copying outputs to the same directory, which can lead to inconsistent builds.</value>
     <comment>{StrBegin="NETSDK1194: "}</comment>
   </data>
   <data name="AotNotSupported" xml:space="preserve">
--- a/src/Tasks/Common/Resources/xlf/Strings.cs.xlf
+++ b/src/Tasks/Common/Resources/xlf/Strings.cs.xlf
@@ -148,8 +148,8 @@
         <note>{StrBegin="NETSDK1099: "}</note>
       </trans-unit>
       <trans-unit id="CannotHaveSolutionLevelOutputPath">
-        <source>NETSDK1194: The "--output" option isn't supported when building a solution.</source>
-        <target state="translated">NETSDK1194: Možnost „--output“ se při sestavování řešení nepodporuje.</target>
+        <source>NETSDK1194: The "--output" option isn't supported when building a solution. Specifying a solution-level output path results in all projects copying outputs to the same directory, which can lead to inconsistent builds.</source>
+        <target state="needs-review-translation">NETSDK1194: Možnost „--output“ se při sestavování řešení nepodporuje.</target>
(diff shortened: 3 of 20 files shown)


src/Tasks/Common/Resources/Strings.resx
```diff
@@ -880,7 +880,7 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1193: "}</comment>
   </data>
   <data name="CannotHaveSolutionLevelOutputPath" xml:space="preserve">
-    <value>NETSDK1194: The "--output" option isn't supported when building a solution.</value>
+    <value>NETSDK1194: The "--output" option isn't supported when building a solution. Specifying a solution-level output path results in all projects copying outputs to the same directory, which can lead to inconsistent builds.</value>
```
baronfel, line 883: I'm assuming it's ok to ship a servicing fix with error message clarifications without accompanying localization? Do we just merge those when they come in later?

src/Cli/dotnet/commands/dotnet-pack/PackCommandParser.cs
```diff
@@ -24,7 +24,7 @@ internal static class PackCommandParser
         public static readonly Option<string> OutputOption = new ForwardedOption<string>(new string[] { "-o", "--output" }, LocalizableStrings.CmdOutputDirDescription)
         {
             ArgumentHelpName = LocalizableStrings.CmdOutputDir
-        }.ForwardAsOutputPath("PackageOutputPath");
+        }.ForwardAsSingle(o => $"-property:PackageOutputPath={CommandDirectoryContext.GetFullPath(o)}");
```
baronfel, line 27: This is a good catch - I was only looking at the targets-based use of Pack!

baronfel (approved):

src/Tasks/Common/Resources/Strings.resx
dsplaisted: Yes, I believe so.

xt0rted:
Nice to see this reverted and by the looks of it my scenario will be supported again, but will print out a warning.

For context I was using `dotnet publish App.sln --no-build --output ./artifacts` so I could publish all of our projects at once which runs a lot faster than doing each project individually. In a quick test earlier `dotnet publish App.sln --no-build` ran in 2 seconds, while running `dotnet publish /src/*.csproj --no-build` for each project took 20 seconds.

The way I've been publishing our projects is as follows:

**Directory.Build.props**

```xml
<Project>

  <PropertyGroup>
    <IsPublishable>False</IsPublishable>
  </PropertyGroup>

</Project>
```

**Directory.Build.targets**

```xml
<Project>

  <Target Name="_PrepareForPublish" BeforeTargets="PrepareForPublish">
    <!--
      Adjust publish output so it appends the PublishTo to the PublishDir.

      This makes running `dotnet publish -o some/folder` easier since each project
      will be published to its own named folder inside of the PublishDir location.

      The default behavior of publishing to the project's bin folder should be
      preserved if a custom output folder isn't provided.
      -->
    <PropertyGroup Condition="!$(PublishDir.StartsWith('bin'))">
      <!-- Ensure any PublishDir has a trailing slash, so it can be concatenated -->
      <PublishDir Condition="!HasTrailingSlash('$(PublishDir)')">$(PublishDir)\</PublishDir>

      <!--
        Default to the project name which can be overridden inside each project file.

        We're using a custom property instead of the built-in PublishDirName because that
        affects the output location when running `dotnet publish` without an output folder.
        -->
      <PublishTo Condition="'$(PublishTo)' == ''">$(MSBuildProjectName)</PublishTo>

      <PublishDir>$(PublishDir)$(PublishTo)</PublishDir>
    </PropertyGroup>
  </Target>

</Project>
```

Then inside each `.csproj` that I want to publish artifacts for (5 websites, 3 worker services, and 4 console apps) I put the following properties:

```xml

  <PropertyGroup>
    <IsPublishable>True</IsPublishable>
    <PublishTo>account-web</PublishTo>
  </PropertyGroup>

</Project>
```

In our ci workflow we'd then call `dotnet build`, `dotnet test --no-build`, and finally `dotnet publish --no-build --output ./artifacts`. Each published project would be published into the artifacts folder, inside of it's own folder. So the structure would be `/artifacts/account-web`, `/artifacts/public-api`, etc.

For now we can work around it with `dotnet publish --no-build -p:PublishDir=../../artifacts` but it'd be nice if `dotnet publish --output ...` continued to work with solutions as long as it was also using `--no-build`.

baronfel:
@xt0rted nice work - It appears to me that you might be a prime candidate for the root output path property we're discussing introducing as part of https://github.com/dotnet/designs/pull/281 - would you mind reviewing that when you have time and seeing if it would allow you to remove this build customization? one of our goals is to remove the need for folks to do complex build manipulation to get reasonable layouts, and reduce the need to use `--output` overall.

TheConstructor:
https://github.com/dotnet/sdk/issues/30624#issuecomment-1432118204 describes a work-around until the new SDK is released

PR #30700, opened 2023-02-17
[Blazor] Fix `publish --self-contained` for Blazor builds.

javiercn:
* Fixes `dotnet publish --self-contained -rid` for hosted applications.
* Adds tests for `--self-contained` and `--no-self-contained` from the command line.


javiercn:
https://github.com/dotnet/aspnetcore/issues/46711

MackinnonBuck (approved):

javiercn:
A note on:

```
<ProjectReference Include="...Client">
  <GlobalPropertiesToRemove>SelfContained</GlobalPropertiesToRemove>
</ProjectReference>
```

The reason is that the CLI property comes from the command-line and override the local property set in the project/targets file, and without `SelfContained` the build system does not detect the RuntimePack for webassembly during the build, which causes the wasm files not to be included during the build and as a result, they don't get considered during publish.

Issue #30369, opened 2023-02-06
Enable using docker build `--platform` switch (easily)


richlander:
This is a specific aspect of: https://github.com/dotnet/dotnet-docker/issues/4388

We constructed a "hero Dockerfile" that we want to offer users, but it doesn't work, sadly. There are multiple points of incompatibility that we need to resolve. See the other issue for more context.

For the SDK/CLI, it comes down to Docker's and .NET's syntax/symbols for architecture not matching.

Docker archs: 

- `amd64`
- `arm64`
- `arm`

.NET RID archs:

- `x64`
- `arm64`
- `arm`

The archs mostly match across the two systems, except for the first one. We need the RID in `-r linux/am64` to be translated to `linux/x64`. No other change is needed (at least that we've discovered).

We only need this for Linux, for two reasons:

- The ENVs we are using are only offered via Linux and buildkit. Windows doesn't use buildkit.
- Cross-architecture docker isn't yet a scenario for Windows containers.

baronfel:
When adding this mapping, we should cover `-r/--runtime`, and `-a/--arch`, but _not_ values for `RuntimeIdentifier` passed explicitly via MSBuild properties (`-p RuntimeIdentifier=linux-amd64`).

richlander:
I forgot about `-a` for this. That's actually a great point. We should probably switch our examples to `-a` since that's simpler. As long as SDK and runtime match (like Debian and Debian or Alpine and Alpine), everything will work great. Note that I realize that mixing Debian and Ubuntu will also work fine.

richlander:
In fact, if it made the feature simpler, I think we might be happy with support for just `-a`. Is that a good idea?

@mthalman

baronfel:
I think the amount of work difference for `-r` and `-a` is pretty negligible, so we should go ahead and do both. There's already an UX inconsistency between CLI options and MSBuild Properties for `RuntimeIdentifier` so I'd prefer to make sure the CLI options behave as uniformly as possible.

richlander:
I was just playing with this and realized that `-a` doesn't work with `dotnet restore`. 

I'd love for `-a` to be our guidance for Dockerfiles, however, we'd need that to work for `dotnet restore`.

PR #30762, opened 2023-02-21
fix issue 30369 enable docker build platform switch

JL03-Yue:
Issue #30369: 
translate RID in -r <OS>/amd64 to <OS>/x64
translate arch in -a amd64 to x64


dsplaisted:
@JL03-Yue Reading through #30369, it looks like the `-a` and `--arch` parameters are missing from `dotnet restore`.  Can you look at adding those together with this PR?

@richlander As implemented here, this will also translate the architecture from amd64 to x64 for Windows or other RIDs.  It's a lot easier to implement that way.  Does that sound OK?

richlander:
Right. That sounds good. I have a PR that will soon be ready on our container samples that relies on these changes, so we'll have some specific use cases to test with.

baronfel (approved):
Nice, very happy to have this in!

richlander:
What is the easiest way for me to get a build to test this?

So nice to see this!!!

baronfel:
@richlander least-effort is probably going to be to wait a couple days for the codeflow to dotnet/installer, then grab an 8.0 preview installer from there. Fastest turnaround would be stitching your own SDK together by:

* building this repo locally with `./build.[cmd|sh]`
* take the `.\artifacts\bin\redist\Debug\dotnet\sdk\8.0.100-dev\dotnet.dll` that results and copy it over `dotnet.dll` from a different, already-installed 8.0 nightly preview SDK
* ???
* profit

richlander:
I love least effort. I will just try once a day starting Monday and we will see when the change shows up.

Once I have validated it work, let's start talking about the 7.0 back port.

PR #30769, opened 2023-02-22
Replace File.Move with copy/delete

joeloff:
`File.Move` does not retain the inherited ACEs while an explicit `Copy` does, so we manually move the file using a separate copy & delete instead.

### DIFF
--- a/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
+++ b/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
@@ -196,7 +196,11 @@ namespace Microsoft.DotNet.Installer.Windows
             if (!File.Exists(destinationFile))
             {
                 // See https://github.com/dotnet/sdk/issues/28450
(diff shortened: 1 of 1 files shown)


joeloff:
> Is this a bug in move?

@danmoseley Not sure, but there's no documentation that I've been able to find that explains what I'm observing. The target directory has
`O:BAG:S-1-5-21-2047949552-857980807-821054962-513D:P(A;OICI;0x1200a9;;;WD)(A;OICI;FA;;;SY)(A;OICI;FA;;;BA)(A;OICI;0x1200a9;;;BU)`

When the file is moved, it ends up with `P(A;;FA;;;SY)(A;;FA;;;BA)(A;;FA;;;S-1-5-21-2047949552-857980807-821054962-504)` and doesn't inherit the world or bullt-in user SIDs. With Copy I see the expected result. I did find one document that mentioned issues with inherited ACEs. It could also be some weirdenss crossing privilege levels because while the user level process extracts the file into a temporary location, an elevated process moves it into the protected location.

@KalleOlaviNiemitalo thanks, I'll take a look at your suggestion.

PR #30792, opened 2023-02-23
Produce reference assembly when target is .NET Standard

drewnoakes:
In general we want to produce reference assemblies whenever possible, as they improve incremental build performance.

Making them work end-to-end requires support from other components, such as the fast up-to-date check in VS. While support was originally added for SDK-style projects, it was not present for legacy `.csproj` files. Accordingly, `ProduceReferenceAssemblies` was set `true` only when the target framework was `net5.0` or later (and `net7.0` for F#).

VS 17.5 ships the new Build Acceleration feature, which works best when used in conjunction with reference assemblies. A customer reported in https://github.com/dotnet/project-system/issues/8865 that builds in their .NET Standard projects were not accelerated. Investigation shows this is because reference assemblies are not produced for such projects.

This change enables the production of reference assemblies for all (non-F#) projects targeting .NET Standard.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets
@@ -240,6 +240,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


baronfel:
As of F# 7 (shipped with .NET 7) F# supports producing reference assemblies. @vzarytovskii should it also be enabled here for libraries?

vzarytovskii:
> As of F# 7 (shipped with .NET 7) F# supports producing reference assemblies. @vzarytovskii should it also be enabled here for libraries?

We can enable them by default, yeah. It's targeting NET8p2, right?

marcpopMSFT:
@drewnoakes per offline discussion, was the decision to change this in msbuild instead and then update these tests when that flows?

drewnoakes:
@marcpopMSFT yes, the proposed MSBuild change is here: https://github.com/dotnet/msbuild/pull/8571

PR #30832, opened 2023-02-24
Interactive mode proof of concept

maridematte:
Implementing a first very basic pass on interactive mode. Mostly opt-in mechanism and where the interactive mode will be setup. 
Currently it has a lot of hard-coded values  that will stay until we get to the dynamic tree parsing for template inputs.

TODO:
- [x] basic tests
- [ ] basic integration tests
- [ ] localization
- [ ] correct template resolution ( STRONGLY RECOMMENDED )

Missing features:
- asking for template name(?)
- types of parameters
- choices for restricted types 
- validation of entered value
- skipping next parameters
- check if there are any template parameters are already entered
- configuring which parameters to show (required, optional)


src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Microsoft.TemplateEngine.Cli.Commands.InteractiveMode
+{
+    public static class InteractiveMode
+    {
+        public static async Task EnterInteractiveMode(InvocationContext context, Func<InvocationContext, Task> next)
+        {
+            if (context.ParseResult.CommandResult.Command.Name == "new")
+            {
+                TemplateCommand? templateCommand = context.ParseResult.CommandResult.Command as TemplateCommand;
```
vlada-shubina, line 16: this command is never `TemplateCommand`, we should look for `NewCommand` or `InstantiateCommand`

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.CommandLine;
+using System.CommandLine.Invocation;
+
+namespace Microsoft.TemplateEngine.Cli.Commands.InteractiveMode
+{
+    public static class InteractiveMode
+    {
+        public static async Task EnterInteractiveMode(InvocationContext context, Func<InvocationContext, Task> next)
+        {
+            if (context.ParseResult.CommandResult.Command.Name == "new")
```
vlada-shubina, line 14: this is identical to `context.ParseResult.CommandResult.Command is NewCommand` so next statement will never be true.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+using Microsoft.TemplateEngine.Edge.Settings;
+
+namespace Microsoft.TemplateEngine.Cli.Commands.InteractiveMode
+{
+    public static class InteractiveMode
+    {
+        public static async Task EnterInteractiveMode(InvocationContext context, Func<InvocationContext, Task> next)
+        {
+            // TODO: make sure we have access to Instantiate command
+            if (context.ParseResult.CommandResult.Command.Name == "new" || context.ParseResult.CommandResult.Command.Name == "create")
```
vlada-shubina, line 19: `is NewCommand` and `is InstantiateCommand` is more reliable way as in parsing tree might be other commands named `create`.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+            // TODO: find a better way to query for the option
+            // TODO: Better handling of this option when executing
+            bool interactiveOptionValue = false;
+            if (NewCommand.InteractiveTemplateOption is not null)
+            {
+                interactiveOptionValue = parsedArgs.GetValue(NewCommand.InteractiveTemplateOption);
+            }
+
+            var envVariable = Environment.GetEnvironmentVariable("DOTNETNEWINTERACTIVE");
+            bool envVariableValue = Convert.ToBoolean(envVariable);
```
vlada-shubina, line 168: Use Env.GetEnvironmentVariableAsBool instead.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+
+                var enterInteractiveMode = ShouldEnterInteractiveMode(context.ParseResult);
+                if (enterInteractiveMode)
+                {
+                    // TODO: add the default skip key
+                    // TODO: short name validation. right now we are assuming that the provided template exists and is valid
+                    context.Console.WriteLine("Currently under development, not all features implemented yet");
+
+                    // --- Workaround to get the environmental settings at this stage. Firgure this out later
+                    ITemplateEngineHost host = instantiateCommand.Host(context.ParseResult);
```
vlada-shubina, line 35: all of this should be done in `InstantiateCommand` class instead.

the method should return the collection of questions to ask, possible options, and how to add them to current command.

ideally this should be queried each time after the answer is received to
- validate it
- recreate the next questions as they may depend on choices made. Example: if user selects F# language, some of parameters might not be available or have different values.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+                    IEngineEnvironmentSettings settings = new EngineEnvironmentSettings(
+                        host,
+                        environment: environment,
+                        pathInfo: new CliPathInfo(host, environment, null));
+
+                    TemplatePackageManager templatePackageManager = new(settings);
+                    HostSpecificDataLoader? hostSpecificDataLoader = new(settings);
+
+                    IReadOnlyList<ITemplateInfo> templates =
+                        Task.Run(async () => await templatePackageManager.GetTemplatesAsync(default).ConfigureAwait(false)).GetAwaiter().GetResult();
```
vlada-shubina, line 46: why not just `await`? This method is async

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+            // hard coded for now
+            // Check if current environment supports what we want to do (If we need any fancy stuff)
+            return true;
+        }
+
+        internal static TemplateCommand? GetTemplate(
+            InstantiateCommandArgs args,
+            IEngineEnvironmentSettings environmentSettings,
+            IEnumerable<TemplateGroup> templateGroups,
+            TemplatePackageManager templatePackageManager)
```
vlada-shubina, line 189: if calling method is moved to `InstantiateCommand` it should be possible to reuse those methods.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+                        {
+                            Console.Error.WriteLine(LocalizableStrings.GenericWarning, e.Message);
+                        }
+                    }
+                }
+            }
+            return null;
+        }
+
+        private static IEnumerable<CliTemplateInfo> GetAllowedTemplates(TemplateConstraintManager constraintManager, TemplateGroup templateGroup)
```
vlada-shubina, line 222: if calling method is moved to `InstantiateCommand` it should be possible to reuse those methods.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+            }
+            else
+            {
+                instantiateCommand = newCommand.Subcommands.First(command => command.Name == "create") as InstantiateCommand;
+            }
+
+            return instantiateCommand;
+        }
+
+        internal static InstantiateCommandArgs? GetArgs(ParseResult parseResult)
```
vlada-shubina, line 136: This logic is better fit in `NewCommand` \ `InstantiateCommand`.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+                    // Need to get input given by user
+                    string commandLineCall = $"dotnet new {templateCommand.Template.ShortNameList[0]}";
+                    foreach (var param in paramsToAdd)
+                    {
+                        commandLineCall += ($" --{param.Key} {param.Value}");
+                    }
+                    // After everything is done reparse the command line input and execute the result
+
+                    Parser parser = ParserFactory.CreateParser(context.ParseResult.RootCommandResult.Command);
+                    parser.Invoke(commandLineCall);
```
vlada-shubina, line 111: Will this way trigger the same middleware again?

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
```diff
@@ -0,0 +1,253 @@
...
+                            if (paramValue == string.Empty)
+                            {
+                                rightInfoCollected = true;
+                            }
+                        }
+                    }
+
+                    // Hard coded for now bc I cannot find where this information would be in the parser / parserResult
+                    // Need to get input given by user
+                    string commandLineCall = $"dotnet new {templateCommand.Template.ShortNameList[0]}";
```
vlada-shubina, line 103: Consider displaying the final command before running it, so the user can reuse it afterwards.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveAsks.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    internal class InteractiveQuerying
+    {
+        // Tree or List depending on future implementation choices
+        // I do not know how to use generic types T_T
+        private List<UserQuery<string>> parametersTree;
+        private List<string> parametersToNotAsk = new List<string>
+        {
+            "TargetFrameworkOverride",
```
vlada-shubina, line 24: `TargetFrameworkOverride` is hidden. Use `Framework` instead.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
maridematte: Moved things to the `InstantiateCommand` class. At this level of prototype I want to limit the amount of choices that our / parameter evaluations do. I feel like it is beyond this scope and it is valuable to sit down and plan this work and some decisions beforehand.

I have also limited this to strings right now, as I was not sure how to use generic types. I think it is good enough for a rough prototype, but it should definitely be the next step for this.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
maridematte: Method was not async at the beginning, this is just a bit of overlooked code when I made it async!

src/Cli/Microsoft.TemplateEngine.Cli/Commands/InteractiveMode/InteractiveMode.cs
maridematte: If we pass the interactive mode options, and the environmental flag than yes. But I think we might be able to disable both at this point if necessary.

baronfel:
I added a bit of visual flair to this PR so it made nice screenshots, and tweaked the display of choice and boolean parameters. Here's what it looks like with the `web` template:

image

We definitely don't have to treat this layout as the final one, but I wanted things to pop a bit for demos :)

baronfel:
After removing that errant newline: 

image

PR #30840, opened 2023-02-24
Remove scrubbing the warning for furhter observation

GangWang01:
### Problem

### Solution
It's not able to reproduce the problem after running tests many times locally or in the pipeline https://github.com/dotnet/sdk/pull/30732.
I think we can do further observation by removing scrubbing the warning and catch potential issue.


YuliiaKovalova:
let's merge it into the main 
to updates to release/7.0.2xx are accepted

PR #30841, opened 2023-02-24
Remove scrubbing the warning for furhter observation

GangWang01:
### Problem

### Solution
It's not able to reproduce the problem after running tests many times locally or in the pipeline https://github.com/dotnet/sdk/pull/30732.
I think we can do further observation by removing scrubbing the warning and catch potential issue.


YuliiaKovalova (approved):

YuliiaKovalova:
It's luck from very first shot :)

GangWang01:
Finally it's reproduced in main branch. No idea what can make it go into the case. Let me try logging and investigate.

Issue #30170, opened 2023-01-27
.NET 7 SDK's dotnet.exe change the attached terminal code page when its standard output is redirected


Dragnalith:
### Bug Description
dotnet.exe change the code page of the attached terminal to be 65001 (i.e UTF-8) when the standard output of the dotnet.exe process is redirected. (it does not occur with dotnet.exe of .Net 6 SDK)

Illustration:
![image](https://user-images.githubusercontent.com/993556/215121910-5219faca-b009-4ce5-89e3-89a5a42bdf4c.png)
![image](https://user-images.githubusercontent.com/993556/215121958-f899de38-f45d-4298-a9f6-69f154e08820.png)

### Impact Of The Bug

This bug break our build.

The terminal code page is a global setting and changing it impacts how following process will run. In our case, it breaks our build system based on ninja. Indeed, our build system run both MSVC C++ compiler and dotnet.exe, and MSVC C++ compiler depends on the code page to decide on its stdout encoding. Having the MSVC C++ compiler encoding changing randomly in the middle of the build creates issues.

### How To Visualize The Terminal Code Page

To understand the bug it is important to be able to display the current code page of the terminal. I have created simple `ShowConsoleCP.exe` tool to do so which call and display the result of `GetConsoleCP()` and `GetConsoleOutputCP()`'s Win32 API. The tool is available in this [ReproCodePage.zip](https://github.com/dotnet/sdk/files/10519982/ReproCodePage.zip) archive.

### How To Reproduce The Bug

Simple way: you can run the `run_repro.cmd` script of [ReproCodePage.zip](https://github.com/dotnet/sdk/files/10519982/ReproCodePage.zip)

Detailed way:
1. Start a terminal
2. create a new dotnet project. (e.g `dotnet new console HelloWorld`)
3. visualize the current terminal code page (e.g use ShowConsoleCP.exe included in [ReproCodePage.zip](https://github.com/dotnet/sdk/files/10519982/ReproCodePage.zip))
     - If the code page is 65001, set another code page and go back to step 3 (e.g run `chcp 932`)
4. build using `dotnet` and redirect its standard output. (e.g `dotnet build HelloWorld > foo.txt`)
5. Visualize the current terminal code page and noticed it is now 65001 instead of the one before

### Further technical details
- I have been able to reproduce the bug using various process creation api and stdout redirection:
  - using cmd.exe `>` symbol to redirect to file
  - using python and creation the dotnet.exe process with python's subprocess module
[...]

PR #30963, opened 2023-03-02
Restore Encoding for the .NET SDK upon Exit

nagilson:
Fixes https://github.com/dotnet/sdk/issues/30170

- The encoding before the SDK is under execution is saved and then restored upon program exit. This will make it so it doesn't affect the behavior of other programs on the console. This behavior was observed before the encoding fixes for other language support, but now that the encoding is actually being applied correctly, it would occur more often. 

- In addition, we have limited the build number to have the encoding UTF 8 fix only on windows builds that officially support UTF 8. Previous builds of Windows 10 2019 that were earlier, even in 2018, did have the support, but the behavior is technically unsupported and thus 'undefined.' 

- Added environment variable to turn on the and force the encoding even if your OS may not support it officially, or if you don't have read registry access and so we can't tell what build of windows you're on. This is 'breaking' in the sense that previously the encoding would work for those without registry access... maybe we don't want to do that? cc @baronfel for a decision. 

- This is a breaking change in that those who relied on the encoding to be changed by the SDK will no longer be able to rely on this.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
@@ -4,7 +4,9 @@
 using System;
 using System.Globalization;
 using System.Runtime.InteropServices;
+using System.Security;
 using System.Text;
+using Microsoft.Win32;
 
 namespace Microsoft.DotNet.Cli.Utils
 {
@@ -27,7 +29,7 @@ namespace Microsoft.DotNet.Cli.Utils
             if (
                 RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && // Encoding is only an issue on Windows
                 !CultureInfo.CurrentUICulture.TwoLetterISOLanguageName.Equals("en", StringComparison.InvariantCultureIgnoreCase) &&
-                Environment.OSVersion.Version.Major >= 10 // UTF-8 is only officially supported on 10+.
+                CurrentWindowsVersionOfficiallySupportsUTF8Encoding()
                 )
             {
                 Console.OutputEncoding = DefaultMultilingualEncoding;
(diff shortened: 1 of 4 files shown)


baronfel:
> Added environment variable to turn on the and force the encoding even if your OS may not support it officially, or if you don't have read registry access and so we can't tell what build of windows you're on. This is 'breaking' in the sense that previously the encoding would work for those without registry access... maybe we don't want to do that? cc @baronfel for a decision.

This behavior is fine to me - we just need to make sure that the environment variable is documented.

src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
```diff
@@ -97,5 +99,31 @@ private static void SetIfNotAlreadySet(string environmentVariableName, int value
         {
             SetIfNotAlreadySet(environmentVariableName, value.ToString());
         }
+
+        private static bool CurrentWindowsVersionOfficiallySupportsUTF8Encoding()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Environment.OSVersion.Version.Major >= 10) // UTF-8 is only officially supported on 10+.
```
v-wuzhai, line 105: It looks like this judgment is repeated in line 30.

nagilson: This was true in the old code but it has been moved into this function

nagilson: Thanks for taking a look though ;)

src/Cli/dotnet/AutomaticEncodingRestorer.cs
```diff
@@ -0,0 +1,65 @@
...
+
+using System;
+using System.IO;
+using System.Security;
+using System.Text;
+
+namespace Microsoft.DotNet.Cli
+{
+    /// <summary>
+    /// A program can change the encoding of the console which would affect other programs.
```
joeloff, line 13: Are there links to any documentation we can include in the code?

src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
joeloff, line 105: Line 30 already checked that the platform is Windows. Maybe make this a more generic function `PlatformSupportsUtf8Encoding`. For now it only checks Windows, but will simplify the logic in the caller a bit.

nagilson: Thanks, that makes sense and maybe what @v-wuzhai was getting at (sorry.)

I changed it to `CurrentPlatformIsWindowsAndOfficiallySupportsUTF8Encoding` because we don't need to change the encoding on non windows platforms, and that is the reason the check was initially separated into is it windows, and then does this version of windows support the encoding, instead of a function that says if the platform does

src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
```diff
@@ -97,5 +99,31 @@ private static void SetIfNotAlreadySet(string environmentVariableName, int value
...
+            {
+                try
+                {
+                    RegistryKey windowsVersionRegistry = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows NT\CurrentVersion");
```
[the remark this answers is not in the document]
nagilson: Good call, thank you. Forgot that it was a disposable

src/Cli/dotnet/AutomaticEncodingRestorer.cs
nagilson: Unfortunately no.
But here is some evidence that this is true, besides the customer who reported it as an issue:

image

nagilson:
@joeloff Thanks for taking a look, I have responded to all feedback.

joeloff (approved):

nagilson:
Made a breaking change doc as well. Thanks!

marcpopMSFT:
This should have targeted main. Preview 2 is closed and there is no automatic port to main from the preview branches. @nagilson

PR #30973, opened 2023-03-03
[release/7.0.3xx] Add new line to C# project templates with file-scoped namespaces

GangWang01:
Port #30972 to release/7.0.3xx.


YuliiaKovalova:
It looks like I figured out why it fails.

**FeaturesSupport** test implies that templates stay are the same between different versions (e.g. _FeaturesSupport.console.Nullable-True#TopLevel-False#ImplicitUsings-True#FileScopedNs-True#cs#NoLangVer.verified_\MyProject\Program.cs - doesn't say explicitly about the package source of the used template in the test output folder name).
But since we add spacing to 7.0.2xx only, when it runs templates from net6.0, can't find the delimiter and fails.

For item templates, it works correctly, because we introduced them recently.

@JanKrivanek, what would you recommend here?
I suggest adding a target framework to the test folder name and thus splitting the output.
cc @vlada-shubina

YuliiaKovalova (approved):

Issue #30962, opened 2023-03-02
Consistency for `os` and `arch` CLI argument names


baronfel:
### Is your feature request related to a problem? Please describe.

The `os` and `arch` options do not set the `ArgumentHelpName` on their definitions, and so aren't uppercased like the other arguments in the help display. We should add this and localize the argument help names the same as all of the other arguments.

```
08:08:16 ❯ dotnet build --help
Description:
  .NET Builder

Usage:
  dotnet build [<PROJECT | SOLUTION>...] [options]

Arguments:
  <PROJECT | SOLUTION>  The project or solution file to operate on. If a file is not specified, the command will search the current directory for
                        one.

Options:
  --ucr, --use-current-runtime         Use current runtime as the target runtime.
  -f, --framework <FRAMEWORK>          The target framework to build for. The target framework must also be specified in the project file.
  -c, --configuration <CONFIGURATION>  The configuration to use for building the project. The default for most projects is 'Debug'.
  -r, --runtime <RUNTIME_IDENTIFIER>   The target runtime to build for.
  --version-suffix <VERSION_SUFFIX>    Set the value of the $(VersionSuffix) property to use when building the project.
  --no-restore                         Do not restore the project before building.
  --interactive                        Allows the command to stop and wait for user input or action (for example to complete authentication).
  -v, --verbosity <LEVEL>              Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and
                                       diag[nostic].
  --debug
  -o, --output <OUTPUT_DIR>            The output directory to place built artifacts in.
  --no-incremental                     Do not use incremental building.
  --no-dependencies                    Do not build project-to-project references and only build the specified project.
  --nologo                             Do not display the startup banner or the copyright message.
  --sc, --self-contained               Publish the .NET runtime with your application so the runtime doesn't need to be installed on the target
                                       machine.
                                       The default is 'true' if a runtime identifier is specified.
  --no-self-contained                  Publish your application as a framework dependent application. A compatible .NET runtime must be installed
                                       on the target machine to run your application.
  -a, --arch <arch>                    The target architecture.
  --os <os>                            The target operating system.
  --disable-build-servers              Force the command to ignore any persistent build servers.
  -?, -h, --help                       Show command line help.
```

### Describe the solution you'd like
A clear and concise description of what you want to happen. Include any alternative solutions you've considered.

baronfel:
[...]

PR #30985, opened 2023-03-03
add <OS><ARCH> argument help name

JL03-Yue:
#30962 

Add `ArgumentHelpName` for `os` and `arch` and uppercase them as other arguments

### DIFF
--- a/src/Cli/dotnet/CommonLocalizableStrings.resx
+++ b/src/Cli/dotnet/CommonLocalizableStrings.resx
@@ -714,7 +714,7 @@ The default is 'true' if a runtime identifier is specified.</value>
   <data name="ArchArgumentName" xml:space="preserve">
     <value>ARCH</value>
   </data>
-  <data name="OsArgumentName" xml:space="preserve">
+  <data name="OSArgumentName" xml:space="preserve">
     <value>OS</value>
   </data>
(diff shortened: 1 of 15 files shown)


baronfel (approved):
I think the only thing missing is to regenerate the loc files for the non-english languages we support.

src/Cli/dotnet/CommonLocalizableStrings.resx
```diff
@@ -708,4 +708,10 @@ The default is 'true' if a runtime identifier is specified.</value>
   <data name="CorruptSolutionProjectFolderStructure" xml:space="preserve">
     <value>The solution file '{0}' is missing EndProject tags or has invalid child-parent project folder mappings around project GUID: '{1}'. Manually repair the solution or try to open and save it in Visual Studio."</value>
   </data>
-</root>
+  <data name="ArchArgumentName" xml:space="preserve">
+    <value>ARCH</value>
+  </data>
+  <data name="OsArgumentName" xml:space="preserve">
```
kasperk81, line 714: convention for two letter abbr in PascalCase is `OSArgumentName`, `OSDescription`, `TargetOS` etc. starting with three letters, it uses the regular pacal casing `XmlSerializer`

Issue #30784, opened 2023-02-22
`dotnet --info` should also show workload info


Eilon:
## Background

The `dotnet` CLI tool has many command line options for showing info about what is installed:

```
sdk-options:
   ...
  --info            Display .NET information.
  --list-runtimes   Display the installed runtimes.
  --list-sdks       Display the installed SDKs.
  --version         Display .NET SDK version in use.
```

The most detailed of those options is `--info`, which includes the details of all the other top-level options, and looks like this:

```
.NET SDK:
 Version:   7.0.200
 Commit:    534117727b

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.200\

Host:
  Version:      7.0.3
  Architecture: x64
  Commit:       0a2bda10e8

.NET SDKs installed:
  3.1.426 [C:\Program Files\dotnet\sdk]
  6.0.309 [C:\Program Files\dotnet\sdk]
  7.0.200 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  ...
  Microsoft.WindowsDesktop.App 7.0.3 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  x86   [C:\Program Files (x86)\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  Not set

global.json file:
  Not found

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
  https://aka.ms/dotnet/download
```

But wait, there's even more info for workloads! The versions of workloads are fundamental regarding what is installed. Info on it is available via the `dotnet workload list` command:

```
Installed Workload Id      Manifest Version       Installation Source
---------------------------------------------------------------------
maccatalyst                16.2.1035/7.0.100      VS 17.6.33417.168
runtimes-windows           7.0.3/7.0.100          VS 17.6.33417.168
maui-ios                   52.1.0/7.0.100         VS 17.6.33417.168
maui-windows               52.1.0/7.0.100         VS 17.6.33417.168
maui-maccatalyst           52.1.0/7.0.100         VS 17.6.33417.168
[...]

PR #31007, opened 2023-03-06
30784 dotnet   info workload info

JL03-Yue:
#30784 Temporary fix of adding workload info from dotnet/sdk to `dotnet --info`


src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
```diff
@@ -77,6 +83,9 @@ private static int ProcessArgs(ParseResult parseResult)
             if (parseResult.HasOption(InfoOption) && parseResult.RootSubCommandResult() == "workload")
             {
                 ShowWorkloadsInfo(parseResult);
+                IReporter reporter = null;
+                reporter ??= Cli.Utils.Reporter.Output;
+                reporter.WriteLine("");
```
dsplaisted, line 88: I think this does the same thing and is simpler:

```suggestion
                Cli.Utils.Reporter.Output.WriteLine("");
```

dsplaisted (approved):

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
```diff
@@ -29,9 +29,16 @@ public static Command GetCommand()
             return Command;
         }
 
-        internal static void ShowWorkloadsInfo(ParseResult parseResult, IWorkloadInfoHelper workloadInfoHelper = null, IReporter reporter = null)
+        internal static void ShowWorkloadsInfo(ParseResult parseResult = null, IWorkloadInfoHelper workloadInfoHelper = null, IReporter reporter = null)
         {
-            workloadInfoHelper ??= new WorkloadInfoHelper(parseResult.HasOption(SharedOptions.InteractiveOption));
+            if(workloadInfoHelper != null)
+            {
+                workloadInfoHelper ??= new WorkloadInfoHelper(parseResult.HasOption(SharedOptions.InteractiveOption));
```
baronfel, line 36: `parseResult` here could be null, might need to guard this access with a check.

baronfel (approved):
:shipit:

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
JL03-Yue: Does the if condition below this line work as a check?

PR #31023, opened 2023-03-07
Update the Newtonsoft.Json to 13.0.1

v-wuzhai:


mmitche (approved):

marcpopMSFT:
> Why not 13.0.2?

We haven't moved the sdk and tools over to .2 yet. We should probably move that first and then all the tests.

PR #31154, opened 2023-03-13
Split WebAssembly SDK from Blazor SDK

maraf:
- Overall implementation plan in the https://github.com/dotnet/runtime/issues/70762
- This PR doesn't produce any change in the build output (some will come after integration on the runtime side in following PRs).
  - All blazor properties, items and targets should work as before.
- All tests remain unchanged (some parts might be moved after runtime side integration).
- All existing boot config options are generated from wasm boot generator. We will probably move some to blazor config extension after integration with blazor JS side. In the meanwhile new blazor-only flags should go to blazor config extension.
- Wasm SDK will probably go through some clean up later as we will integrate runtime side.

Contributes to https://github.com/dotnet/aspnetcore/issues/47001


src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/GenerateBlazorBootJsonTest.cs
```diff
@@ -91,7 +92,7 @@ public void GroupsResourcesByType()
                         ("FileHash", "my-custom-extensionhash"),
                         ("RelativePath", "my-custom-extension.blz"),
                         ("TargetPath", "_bin/my-custom-extension.blz"),
-                        ("AssetTraitName", "BlazorWebAssemblyResource"),
+                        ("AssetTraitName", "WasmResource"),
```
javiercn, line 95: We need to be careful here because I believe we use this on the service worker. Just making a note here so that I don't forget as I haven't gone through the entire PR

src/WasmSdk/Sdk/Sdk.props
```diff
@@ -0,0 +1,24 @@
...
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
+<Project ToolsVersion="14.0" TreatAsLocalProperty="RuntimeIdentifier">
+  <PropertyGroup>
+    <!-- FIXME: Important to get rid of missing workload -->
+    <UsingMicrosoftNETSdkBlazorWebAssembly>true</UsingMicrosoftNETSdkBlazorWebAssembly>
```
javiercn, line 15: Is this temporary?

src/WasmSdk/Targets/Microsoft.NET.Sdk.WebAssembly.Current.props
```diff
@@ -0,0 +1,55 @@
...
+    <_WasmSdkImportsMicrosoftNETSdkWebProjectSystem Condition="'$(MicrosoftNETSdkWebProjectSystem)' != 'true'">true</_WasmSdkImportsMicrosoftNETSdkWebProjectSystem>
+    <_WasmSdkImportsMicrosoftNETSdkPublish Condition="'$(MicrosoftNETSdkPublish)' != 'true'">true</_WasmSdkImportsMicrosoftNETSdkPublish>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectCapability Include="WebAssembly" />
+  </ItemGroup>
+
+  <Import Sdk="Microsoft.NET.Sdk.StaticWebAssets" Project="Sdk.props" Condition="'$(_WasmSdkImportsMicrosoftNETSdkStaticWebAssets)' == 'true'" />
+  <Import Sdk="Microsoft.NET.Sdk.Web.ProjectSystem" Project="Sdk.props" Condition="'$(_WasmSdkImportsMicrosoftNETSdkWebProjectSystem)' == 'true'" />
```
javiercn, line 52: Do you need to import this? Static web assets should be importing it, is it not?

src/WasmSdk/Tasks/BootJsonData.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        /// <summary>
+        /// "assembly" (.dll) resources
+        /// </summary>
+        public ResourceHashesByNameDictionary assembly { get; set; } = new ResourceHashesByNameDictionary();
+
+        /// <summary>
+        /// "debug" (.pdb) resources
+        /// </summary>
+        [DataMember(EmitDefaultValue = false)]
```
javiercn, line 79: FYI, this can be updated to use `System.Text.Json`. It used datacontractserializer because it predates it, but I did support to add System.Text.Json support to the SDK tasks, so you should be able to use it. (It might be that I only did it in the RazorSDK, but it should be "copiable")

javiercn: Not asking you to change it, just letting you know

src/WasmSdk/Tasks/FileHasher.cs
```diff
@@ -6,7 +6,7 @@
 using System.Security.Cryptography;
 using System.Text;
 
-namespace Microsoft.NET.Sdk.BlazorWebAssembly
+namespace Microsoft.NET.Sdk.WebAssembly
 {
     public static class FileHasher
```
javiercn, line 11: @MackinnonBuck this will move to static web assets as part of https://github.com/dotnet/aspnetcore/issues/47000

src/WasmSdk/Sdk/Sdk.props
maraf: Yes, I'll update the property during the migration to runtime repo once I update the wasm-tools workload manifest

src/WasmSdk/Tasks/BootJsonData.cs
maraf: I would love to do it. I think I struggled on the package version when I tried it last week. So you are saying that I can use version 6.0 like in here https://github.com/dotnet/razor/blob/33e06ed0fd64c8b67b80d464370ee0380f9ab32c/eng/Versions.props#L179, right?

javiercn: https://github.com/dotnet/sdk/commit/9ed247b28cba4501185dad4e546dc9c51e1c5711#diff-8bd75765f34d87f1dcad4a428355ceb7e78e7be3febb15e66bbae735c60ae374R42

javiercn (approved):

PR #31174, opened 2023-03-14
Blazor configuration for StartupMemoryCache and Jiterpreter

maraf:
New MSBuild and blazor.boot.json properties for configuring runtime features
- `(BlazorWebAssembly)StartupMemoryCache` enables mono runtime statup memory caching in the browser cache
- `(BlazorWebAssembly)RuntimeOptions` provides options for mono runtime
- `BlazorWebAssemblyJiterpreter` enables mono runtime jiterpreter
  - The property in replaced with actual runtime options for jiterpreter while generating boot config


javiercn (approved):
Looks good

maraf:
I have replaced jiterpreter boot config flag with runtime options

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -44,6 +44,8 @@ public class GenerateBlazorWebAssemblyBootJson : Task
 
         public string Jiterpreter { get; set; }
 
+        public string RuntimeOptions { get; set; }
```
javiercn, line 47: Hmm, this is coming from a property from what I can tell. If you have multiple values, shouldn't it be better if it comes from an ItemGroup?

I believe it still can be a property if you use `;` as the delimiter and just evaluate it in the itemgroup as Include="$(Prop) and MSBuild will do the parsing for you.

javiercn: @rainersigwald is there a way to go from Property to Item using an alternate character separator?

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -85,8 +91,30 @@ public void WriteBootJson(Stream output, string entryAssemblyName)
...
+                startupMemoryCache = ParseOptionalBool(StartupMemoryCache),
             };
 
+            if (!String.IsNullOrEmpty(RuntimeOptions))
+            {
+                string[] runtimeOptions = RuntimeOptions.Split(' ');
+                result.runtimeOptions = runtimeOptions;
+            }
+
+            if (ParseOptionalBool(Jiterpreter) == false)
```
pavelsavara, line 103: this code assumes that the runtime default is `true`. I would prefer to make it also work in the other direction. 
That is enable it if the runtime default was `false`.

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
maraf: The reason a choosed property is that it has a syntax of command line arguments. So the user types "-a -b -c". But I can change it to item group if that is a preferred way.

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
```diff
@@ -100,17 +102,17 @@ public void WriteBootJson(Stream output, string entryAssemblyName)
...
-                    runtimeOptions.Add("--no-jiterpreter-traces-enabled");
-
-                if (!runtimeOptions.Contains("--jiterpreter-interp-entry-enabled"))
-                    runtimeOptions.Add("--no-jiterpreter-interp-entry-enabled");
-
-                if (!runtimeOptions.Contains("--jiterpreter-jit-call-enabled"))
-                    runtimeOptions.Add("--no-jiterpreter-jit-call-enabled");
+                foreach (var jiterpreterOption in jiterpreterOptions)
+                {
+                    if (jiterpreter == true)
```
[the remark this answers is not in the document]
maraf: It's a `HashSet`

pavelsavara (approved):

src/BlazorWasmSdk/Tasks/GenerateBlazorWebAssemblyBootJson.cs
javiercn: @maraf It is a bit strange to use the command-line syntax for the symbol in MSBuild.

It means that if you ever change the CLI syntax, then you break anybody that had the option setup in the csproj file. Typically, the SDK defines well-known properties for the different options and those get translated later to specific CLI flags/arguments.

I am not asking that you change it, just that you think about it, if you want to be consistent with other areas of the product.

pavelsavara: I prefer opaque style.

I think this is very low level feature and that SDK should not create tight coupling to Mono VM features.
If the customer project already knows enough about Mono runtime options, they are already tightly coupled to it.

Also we are not trying to document the full list of options here. 
We are not making individual options public API here.

maraf: @javiercn I see you point and I would go that way. I asked around whether it's worth to intruducing property per runtime option and the answer I got is that once jiterpreter is stabilized, we won't need these settings. The main reason we introduce this property now is to be able to granularily disable jiterpreter features in following previews.

maraf: @pavelsavara For features that are used quite often, I prefer to have concrete properties, and once we will have these bits of SDK closer to runtime, it would be easy to do.

rainersigwald: > @rainersigwald is there a way to go from Property to Item using an alternate character separator? I know for displaying you can use `@(Item, ', ')` to define an alternative separator, but I don't know a way to do it the other way around.

Sorry, I was out of the office when this was fresh. There's no way to have the MSBuild engine do that for you, just "task takes `string` and splits on its own delimiter".

PR #31201, opened 2023-03-14
Don't add AOT package in restore for < net7

### LINKED ISSUE #30814: PublishAOT property causes build failure when used with earlier TFMS in `TargetFrameworks` property
Repro:

```

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```

Now publish with `~/tmp/dn/dotnet.exe publish -r win-x64 -p:PublishAot=true -f net8.0`

This should work because PublishAot is valid for `net8.0` and we're passing `-f net8.0` to only target that TFM. But it appears at least some logic is being executed for `net6.0` regardless.

This will also fail if you place `PublishAot` in the project file unconditionally and then run `publish --sc -r win-x64 -f net8.0`.

In both of these cases, we should somehow guard against the PublishAot flag being propagated through in unsupported TFMs, or fix framework targeting to run no logic for alternate TFMs when specified with `-f`.

jtschuster:
Fixes https://github.com/dotnet/sdk/issues/30814.

In order to avoid errors for invalid RIDs for AOT during restore, we need to avoid trying to restore a package for ILC when TFM < 7. Then, we need to warn during publish when there are no ILC packages found during restore to make sure we don't still go through with publish with PublishAOT=true and TFM<7.


dsplaisted (approved):

sbomer (commented):
Could you check if the other publish scenarios have the same problem? I think you could get it with PublishReadyToRun (maybe you need to opt into crossgen2) on downlevel TFMs too.

jtschuster:
I forgot to follow up on the PR feedback and let this get stale. The issue was still present in preview 4, and this fix is still necessary. @sbomer @agocke Can you take another look?

sbomer (changes requested):
Thanks for following up on this!

Looking at this a second time, I don't think adding a TFM condition to the task is a good idea. The source of truth for which TFMs are supported by the various scenarios (not just AOT) lives in https://github.com/dotnet/installer/blob/main/src/redist/targets/GenerateBundledVersions.targets, and `ProcessFrameworkReferences` is the current choke point that produces errors for all of these scenarios. I've been leaning on this to improve the error messages (see https://github.com/dotnet/sdk/pull/32943 and https://github.com/dotnet/sdk/pull/33041).

The current restore behavior (restoring for all TFMs) is by design per https://github.com/dotnet/sdk/issues/30814#issuecomment-1442418518, so if we're unhappy with this behavior I think we should be looking at a broader solution that changes the design.

Issue #15254, opened 2021-01-11
When installing a .NET local tool, create the tool manifest if one does not exist


augustoproiete:
When a user tries to install a .NET tool as a local tool on a folder that does not contain the manifest, we get an error `Cannot find a manifest file`.

```powershell
> dotnet tool install --local Cake.Tool --version 0.38.4
```

```
Cannot find a manifest file.
For a list of locations searched, specify the "-d" option before the tool name.
If you intended to install a global tool, add `--global` to the command.
If you would like to create a manifest, use `dotnet new tool-manifest`, usually in the repo root directory.
```

However, if the user tries to update an existing local tool, `dotnet tool` is able to update the existing manifest:

```powershell
dotnet new tool-manifest
dotnet tool install --local Cake.Tool --version 0.38.4
dotnet tool update --local Cake.Tool
```

```
Tool 'cake.tool' was successfully updated from version '0.38.4' to version '0.38.5' (manifest file /Users/augustoproiete/.config/dotnet-tools.json).
```

---

_TL;DR;_ It would be great if `dotnet tool install` automatically created the tool manifest, if one does not exist when installing a local tool, instead of failing with an error.

wli3:
The side of doing that is we could potentially generating manifest file accidentally on unexpected locations. And people could run unexpected tool.

augustoproiete:
> The side of doing that is we could potentially generating manifest file accidentally on unexpected locations. And people could run unexpected tool.

Add a `--force` or `--trust-me` or `--i-said-install-so-just-install-it-damn-it` parameter so users can convince `dotnet tool` that they know what they are doing? :smile:

wli3:
Prefer to https://github.com/NuGet/NuGetGallery/issues/8373#issuecomment-780298879

wli3:
> > The side of doing that is we could potentially generating manifest file accidentally on unexpected locations. And people could run unexpected tool.
> 
> Add a `--force` or `--trust-me` parameter so users can convince `dotnet tool` that they know what they are doing? 😄

The current behavior -- asking the user to run `dotnet new tool-manifest` is about the same level of work comparing with asking the user to add a parameter to `dotnet tool ...` command. I think it is fine adding another option.

augustoproiete:
@wli3 The end goal is to be able to run a command (or set of commands) that **ensures** that a particular dotnet tool, of a particular version, is installed.

[...]

PR #31231, opened 2023-03-15
add --create-manifest-if-needed flag

JL03-Yue:
#15254 

Add `--create-manifest-if-needed` flag
Create tool manifest automatically when installing a new local tool if the flag is on and no manifest exists.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/Constants.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/Constants.cs
@@ -10,12 +10,14 @@ namespace Microsoft.DotNet.Cli.Utils
         public const string DefaultConfiguration = "Debug";
 
         public static readonly string ProjectFileName = "project.json";
+        public static readonly string ManifestFileName = "dotnet-tools.json";
         public static readonly string DotConfigDirectoryName = ".config";
         public static readonly string ExeSuffix =
             RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".exe" : string.Empty;
 
         public static readonly string BinDirectoryName = "bin";
         public static readonly string ObjDirectoryName = "obj";
+        public static readonly string GitDirectoryName = ".git";
 
         public static readonly string MSBUILD_EXE_PATH = "MSBUILD_EXE_PATH";
         public static readonly string MSBuildExtensionsPath = "MSBuildExtensionsPath";
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/Constants.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/Constants.cs
@@ -10,7 +10,7 @@ namespace Microsoft.DotNet.Cli.Utils
         public const string DefaultConfiguration = "Debug";
 
         public static readonly string ProjectFileName = "project.json";
-        public static readonly string ManifestFileName = "dotnet-tools.json";
+        public static readonly string ToolManifestFileName = "dotnet-tools.json";
         public static readonly string DotConfigDirectoryName = ".config";
         public static readonly string ExeSuffix =
             RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".exe" : string.Empty;
--- a/src/Cli/dotnet/ToolManifest/LocalizableStrings.resx
+++ b/src/Cli/dotnet/ToolManifest/LocalizableStrings.resx
@@ -169,4 +169,11 @@ For a list of locations searched, specify the "-d" option before the tool name.<
   <data name="MultipleSamePackageId" xml:space="preserve">
     <value>More than one entry exists for package(s): {0}.</value>
   </data>
+  <data name="ManifestFileContent" xml:space="preserve">
+    <value>{
+  "version": 1,
+  "isRoot": true,
+  "tools": {}
+}</value>
+  </data>
 </root>
--- a/src/Cli/dotnet/ToolManifest/LocalizableStrings.resx
+++ b/src/Cli/dotnet/ToolManifest/LocalizableStrings.resx
@@ -169,11 +169,4 @@ For a list of locations searched, specify the "-d" option before the tool name.<
   <data name="MultipleSamePackageId" xml:space="preserve">
     <value>More than one entry exists for package(s): {0}.</value>
   </data>
-  <data name="ManifestFileContent" xml:space="preserve">
-    <value>{
-  "version": 1,
-  "isRoot": true,
-  "tools": {}
-}</value>
-  </data>
 </root>
--- a/src/Cli/dotnet/ToolManifest/ToolManifestFinder.cs
+++ b/src/Cli/dotnet/ToolManifest/ToolManifestFinder.cs
@@ -212,7 +212,13 @@ namespace Microsoft.DotNet.ToolManifest
 
         private string WriteManifestFile(DirectoryPath folderPath)
         {
-            var manifestFileContent = "{\r\n  \"version\": 1,\r\n  \"isRoot\": true,\r\n  \"tools\": {}\r\n}";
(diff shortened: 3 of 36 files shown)


src/Cli/Microsoft.DotNet.Cli.Utils/Constants.cs
```diff
@@ -10,12 +10,14 @@ public static class Constants
         public const string DefaultConfiguration = "Debug";
 
         public static readonly string ProjectFileName = "project.json";
+        public static readonly string ManifestFileName = "dotnet-tools.json";
```
dsplaisted, line 13: Name suggestion:

```suggestion
        public static readonly string ToolManifestFileName = "dotnet-tools.json";
```

src/Cli/dotnet/ToolManifest/ToolManifestFinder.cs
```diff
@@ -153,7 +155,7 @@ public bool TryFind(ToolCommandName toolCommandName, out ToolManifestPackage too
             }
         }
 
-        public FilePath FindFirst()
+        public FilePath FindFirst(bool createManifestFileOption = false)
```
dsplaisted, line 158: Naming suggestion:

```suggestion
        public FilePath FindFirst(bool createIfNotFound = false)
```

src/Cli/dotnet/ToolManifest/ToolManifestFinder.cs
```diff
@@ -162,12 +164,48 @@ public FilePath FindFirst()
...
-                string.Format(LocalizableStrings.ListOfSearched,
-                    string.Join(Environment.NewLine,
-                        EnumerateDefaultAllPossibleManifests().Select(f => "\t" + f.manifestfile.Value))));
+                    LocalizableStrings.CannotFindAManifestFile,
+                    string.Format(LocalizableStrings.ListOfSearched,
+                        string.Join(Environment.NewLine,
+                            EnumerateDefaultAllPossibleManifests().Select(f => "\t" + f.manifestfile.Value))));
+        }
+
+        private DirectoryPath GetInsertManifestFilePath()
```
dsplaisted, line 182: I'm not sure what a good name for this method would be, but I think this is a bit of an improvement:

```suggestion
        private DirectoryPath GetDirectoryToCreateToolManifest()
```

dsplaisted (commented):
This looks good.  The main thing still needed is to add test coverage.

src/Cli/dotnet/ToolManifest/LocalizableStrings.resx
```diff
@@ -169,4 +169,11 @@ For a list of locations searched, specify the "-d" option before the tool name.<
...
   </data>
-</root>
+  <data name="ManifestFileContent" xml:space="preserve">
+    <value>{
+  "version": 1,
+  "isRoot": true,
+  "tools": {}
+}</value>
+  </data>
+</root>
```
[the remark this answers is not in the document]
JL03-Yue: So can it be set as a member variable in the class?

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
```diff
@@ -38,6 +38,11 @@ internal static class ToolInstallCommandParser
             ArgumentHelpName = LocalizableStrings.FrameworkOptionName
         };
 
+        public static readonly Option<bool> CreateManifestIfNeededOption = new Option<bool>("--create-manifest-if-needed", LocalizableStrings.CreateManifestIfNeededOptionDescription)
+        {
+            ArgumentHelpName = LocalizableStrings.CreateManifestIfNeededOptionName
```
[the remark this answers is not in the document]
baronfel: No - we don't do this for the names of boolean options, only descriptions.

nagilson:
@baronfel Was a final decision reached on this change? I believe we are going to go with what this PR does for now?

<hr>

There is the proposed ` --manifest-file`, but that was not added here to my knowledge.

<hr>

In addition, regarding:
> --create-manifest-if-needed false

Also, one question about the below proposal:
> change the default so that you don't need to specify the --create-manifest-if-needed

I assume that with the below stipulation, this means it's by default IFF the following is true:

> If the --create-manifest-if-needed flag is not specified, then we probably do not want to create a manifest in the current folder

src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
```diff
@@ -232,4 +232,10 @@ If you would like to create a manifest, use `dotnet new tool-manifest`, usually
   <data name="PrereleaseAndVersionAreNotSupportedAtTheSameTime" xml:space="preserve">
     <value>The --prerelease and --version options are not supported in the same command</value>
   </data>
+  <data name="CreateManifestIfNeededOptionDescription" xml:space="preserve">
+    <value>Create tool manifest when installing tools if manifest not exist.</value>
```
dsplaisted, line 236: @baronfel Any suggestions for this option description?  Should we describe how we decide where to put the manifest file?

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs
```diff
@@ -38,6 +38,11 @@ internal static class ToolInstallCommandParser
...
+        };
```
dsplaisted, line 44: Per @baronfel's response, we don't need to specify the ArgumentHelpName here:

```suggestion
        public static readonly Option<bool> CreateManifestIfNeededOption = new Option<bool>("--create-manifest-if-needed", LocalizableStrings.CreateManifestIfNeededOptionDescription);
```

src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
```diff
@@ -232,4 +232,10 @@ If you would like to create a manifest, use `dotnet new tool-manifest`, usually
...
+  </data>
+  <data name="CreateManifestIfNeededOptionName" xml:space="preserve">
+    <value>MANIFEST</value>
+  </data>
```
dsplaisted, line 240: Looks like we don't need this string and can delete it (see comment in ToolInstallCommandParser.cs).

```suggestion
```

dsplaisted (approved):
Looks great!  A few suggestions still.

src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
baronfel: How about "Create a tool manifest if one isn't found during tool installation. For information on how manifests are located, see [HELP URL]".

We should document the search algorithm on https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-local-tool, though. Potentially on the [CannotFindAManifestFile](https://github.com/dotnet/sdk/pull/31231/files/6177c345e89b2357ab290c0d541fe55e209733e0#diff-4288e2d70efc0eba5c7ffec9c84313f69b4111f7611d00345c58442ef96b690aR120-R123) error as well.

dsplaisted: How about https://aka.ms/dotnet-local-tool-create-manifest-if-needed?

baronfel: An aka.ms link would be perfect - folks have been 'namespacing' them using url path segments, though - maybe something like `aka.ms/dotnet/tools/create-manifest-if-needed`?

baronfel:
@JL03-Yue we need to doc this! I just went looking and couldn't find it. I'll make a docs issue and we can work on doing that.

Issue #30872, opened 2023-02-25
How should we represent source-generators / analyzers that are off by default?


ericstj:
When we first added source generators/analyzers to our shared frameworks in .NET 6.0 they were meant to be always on and used like "add-on" features.  New API that was explicitly addressed through some user gesture - like applying an attribute.

Now in .NET 8.0 more generators are coming online that intend to apply some behavior to existing code with no source changes.  The user's source itself does not opt-in to using the generator, but the generator will change the user's code in some way. 

One example of this is: https://github.com/dotnet/runtime/issues/44493
This is meant to automatically replace usage of API that is not linker safe with source generated code that is linker safe.

There may be more in the future if we come up with additional generators that are meant to change the user's code in some meaningful way.

In light of these, we should think about how we want to represent these generators and activate them / disable them.

Today the conventions for generators / analyzers in both nuget packages and the ref-packs are "always on".  There is not a great convention for disabling generators passed to the compiler.  See https://github.com/dotnet/roslyn/issues/55518

One case where this is happening today is 
https://github.com/dotnet/sdk/blob/a76a0ccd0d6ab74dd6ec63002f7fd16085e11659/src/WebSdk/Web/Sdk/Sdk.props#L64-L75
Where ASP.NET doesn't want the generator enabled by default and runs a target to remove it.  This works, but requires targets be added to the SDK for each generator, along with separate handling in the nuget package (in cases of OOB generators).

### Describe the solution you'd like
Some consistent way to represent and interact with optional generators.  Ideally it shouldn't involve running targets.

Potential solutions:
1. A way to represent disabled generators in the ref packs and packages and SDK support for "enabling" them.
2. A way to represent disabled by default generators in the actual generator API, then compiler support for "enabling" them.
3. Some other type of gesture in the source for enabling/disabling.  EG: assembly level attribute / additional data to compiler / etc.
etc

cc @eerhardt @chsienki @dsplaisted @davidfowl @captainsafia @layomia @terrajobst @stephentoub

captainsafia:
I'll chime in with a gotcha around the approach that the ASP.NET generators currently use that @RussKie and I recently ran into, specifically when projects reference the ASP.NET target framework but _don't_ reference the SDK. One example might be test projects that need to reference ASP.NET APIs without being ASP.NET apps themselves. In these cases, there is no way to disable the generator without duplicating the logic that is in the Web SDK in the frameworks targets.

dsplaisted:
[...]

PR #31265, opened 2023-03-16
Move RDG omission to ResolveTargetingPackAssets target

captainsafia:
Related to https://github.com/dotnet/sdk/issues/30872.

Centralize the configuration for off-by-default analyzers/generators into a new build target that runs after `ResolveTargetingPackAssets`. The idea is that we roll out new generators that need to be off-by-default, we would add new items to the `OffByDefaultAnalyzer` item group.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
@@ -427,7 +427,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <ItemGroup>
       <AnalyzerToRemove Include="@(OffByDefaultAnalyzer)"
-                        Path="%(Analyzer.FullPath)"
+                        Path="%(Analyzer.Identity)"
                         Condition="'%(Analyzer.Filename)' == '@(OffByDefaultAnalyzer)'" />
     </ItemGroup>
 
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
@@ -30,7 +30,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ItemGroup>
       <KnownFrameworkReference Remove="Microsoft.Windows.SDK.NET.Ref" />
     </ItemGroup>
-    
+
     <!-- Generate KnownFrameworkReference items for the Windows SDK pack -->
     <CreateWindowsSdkKnownFrameworkReferences
       UseWindowsSDKPreview="$(UseWindowsSDKPreview)"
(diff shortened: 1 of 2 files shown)


captainsafia:
@dsplaisted Thoughts on this? Would be great to get it in for preview3.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
```diff
@@ -418,6 +418,24 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+          Condition="'@(FrameworkReference)' != ''">
+    <ItemGroup>
+      <OffByDefaultAnalyzer Include="Microsoft.AspNetCore.Http.Generators"
+                            IsEnabled="$(EnableRequestDelegateGenerator)"/>
+    </ItemGroup>
+
+    <ItemGroup>
+      <AnalyzerToRemove Include="@(OffByDefaultAnalyzer)"
+                        Path="%(Analyzer.Identity)"
+                        Condition="'%(Analyzer.Filename)' == '@(OffByDefaultAnalyzer)'" />
```
dsplaisted, line 431: There are tests covering this, right?  The join is a bit tricky and I probably wouldn't get it right on the first try.

dsplaisted (approved):
Should this work on analyzers coming from NuGet packages too?  I think I remember that we discussed that it should, but it looks like in this case it will only work on analyzers from targeting packs.

Otherwise, looks good.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets
layomia: FWIW here's the current logic for handling analyzers in NuGet packages in the runtime (to provide the generator to folks using older SDKs) - https://github.com/dotnet/runtime/blob/1e601c5101e1f0573cd03560615aa8f6cb38fa64/eng/packaging.targets#L196-L229

 Is there any concern around these targets conflicting with each other?

captainsafia: > There are tests covering this, right? The join is a bit tricky and I probably wouldn't get it right on the first try.

Yep, there are pre-existing tests for the behavior in [this test class](https://github.com/dotnet/sdk/blob/28ba4ed390c790662da8a89309e311b1ce900f39/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToUseAnalyzers.cs#L26-L60) which is how I validated it worked (definitely not on the first try though 🤪).

> Should this work on analyzers coming from NuGet packages too? 

At least for this PR, I was focused on the case for analyzers derived from targeting packs as described here (https://github.com/dotnet/sdk/issues/30872#issuecomment-1463030722).

> Is there any concern around these targets conflicting with each other?

It doesn't seem like they would conflict, especially because the config generator has not yet been added to the list of `OffByDefaultAnalyzer`s (assuming this PR is merged).

I'm not sure if it is worthwhile to move the targets that are currently in the runtime to the SDK.

dsplaisted: I see, looks good.

dsplaisted:
FYI @marcpopMSFT, this PR should go into preview 3 if possible.

captainsafia:
> FYI @marcpopMSFT, this PR should go into preview 3 if possible.

Is this good to merge now?

There's one reaction in ASP.NET we'd like to get in for preview3 after this. We want to rename our generator assembly from Microsoft.AspNetCore.Http.Generator**s** => Microsoft.AspNetCore.Http.Generator once this change is in.

PR #31301, opened 2023-03-20
GenAPI should use the latest available codeanalysis workspaces package

ViktorHofer:
The roslyn based GenAPI is a new tool that still requires frequent code fixes in the roslyn repository. We want to automatically pick the fixes up when they are available via the shipping Microsoft.CodeAnalysis.Workspaces package.

In addition to that, we currently don't have any ambitions to make the tooling available publicly and therefore are OK with requiring an 8.0 SDK.


andriipatsula:
Do we want to create a maestro subscription between roslyn --> sdk ? Do you see a value in this?

andriipatsula (approved):

PR #31313, opened 2023-03-20
Add AfterPublish target to ClickOnce target dependencies

sujitnayak:
Fixes [AB#1730358](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1730358)

### Issue
Users add custom target to run after AfterPublish expecting to run custom tasks after ClickOnce publish is complete and don't see their custom target invoked.

### Fix
Add AfterPublish msbuild target to list of dependencies invoked during ClickOnce publish.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ClickOnce.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ClickOnce.targets
@@ -63,6 +63,7 @@ Copyright (c) .NET Foundation. All rights reserved.
       _DeploymentGenerateBootstrapper;
       ResolveKeySource;
       _DeploymentSignClickOnceDeployment;
+      AfterPublish
     </ClickOncePublishDependsOn>
   </PropertyGroup>
 


Forgind:
I may be confused about the ordering here. So it sounds like you're observing:
AfterPublish --> CustomerTarget --> ClickOncePublish
but you want:
AfterPublish --> ClickOncePublish --> CustomerTarget

Is that right? Adding DependsOnTargets="AfterPublish" to ClickOncePublish will ensure that AfterPublish runs before ClickOncePublish, but it doesn't necessarily put it after CustomerTarget.

sujitnayak:
> 
Currently AfterPublish target is not being invoked at all.

What we want is for AfterPublish to get invoked after all ClickOnce specific targets have finished. The last ClickOnce target to get invoked is _DeploymentSignClickOnceDeployment. We want AfterPublish to run after that.

ClickOncePublish itself is a pseudo target that does not do anything. So it's fine if it finishes after custom target.


Custom Target will be defined as such:
<Target Name="MyAfterPublish" AfterTargets="AfterPublish">


This is similar to how we do it in ClickOnce for .NET FX:
https://github.com/dotnet/msbuild/blob/a2490dd3f78cce4abc8f9e6f1b5268437332818f/src/Tasks/Microsoft.Common.CurrentVersion.targets#L5846

sujitnayak:
@Forgind do you have any more concerns?

sujitnayak:
@Forgind 
do you any more concerns?

Forgind (approved):

John-Hart (approved):

PR #31388, opened 2023-03-22
Make SdkFeatureBand internal (sometimes)

joeloff:
## Description

This change fixes a performance regression in Visual Studio. MSBuild uses reflection to locate SDK resolvers. It uses [`GetTypeInfo`](https://github.com/dotnet/msbuild/blob/56087a97257fab6a67d110a7b2c4116f707325ee/src/Build/BackEnd/Components/SdkResolution/SdkResolverLoader.cs#L231) that trigger assembly loads. The `SdkFeatureBand` struct is public and uses a private backing field that depends on the `ReleaseVersion` type. This causes an additional load operation for Microsoft.Deployment.DotNet.Releases that trips the module count performance regression.

### DIFF
--- a/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkFeatureBand.cs
+++ b/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkFeatureBand.cs
@@ -7,7 +7,11 @@ using Microsoft.Deployment.DotNet.Releases;
 #nullable disable
 namespace Microsoft.NET.Sdk.WorkloadManifestReader
 {
+#if SDK_FEATURE_BAND_INTERNAL
+    internal struct SdkFeatureBand : IEquatable<SdkFeatureBand>, IComparable<SdkFeatureBand>
+#else
     public struct SdkFeatureBand : IEquatable<SdkFeatureBand>, IComparable<SdkFeatureBand>
+#endif
     {
         private ReleaseVersion _featureBand;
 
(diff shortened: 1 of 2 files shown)


joeloff:
I'm working on figuring out how to verify this without building the full stack and inserting to VS.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkFeatureBand.cs
```diff
@@ -7,7 +7,11 @@
 #nullable disable
 namespace Microsoft.NET.Sdk.WorkloadManifestReader
 {
+#if SDK_FEATURE_BAND_INTERNAL
+    internal struct SdkFeatureBand : IEquatable<SdkFeatureBand>, IComparable<SdkFeatureBand>
+#else
     public struct SdkFeatureBand : IEquatable<SdkFeatureBand>, IComparable<SdkFeatureBand>
+#endif
```
dsplaisted, line 14: What do you think about something like this, to avoid the duplication?

```c#
#if SDK_FEATURE_BAND_INTERNAL
    internal
#else
    public
#endif
    struct SdkFeatureBand : IEquatable<SdkFeatureBand>, IComparable<SdkFeatureBand>
```

dsplaisted (approved):

joeloff:
We resolved this with VS. The specific resolver will be loaded eventually and will pull in the additional dependencies so any fix just pushes the load further down, but cannot eliminate it for the specific test.

PR #31398, opened 2023-03-23
Choose KnownILCompilerPacks based on TFM + RID

kant2002:
Right now appropriate ILCompiler pack selected purely based on TFM. To relatively easy support of NativeAOT LLVM I would like that selection of ILCompiler pack happens on pair TFM + RID.

That allow adding custom ILC (as in case NativeAOT LLVM) using following code in the project file (only)
```
<KnownILCompilerPack Include="Microsoft.DotNet.ILCompiler.LLVM"
   TargetFramework="net7.0" ILCompilerPackNamePattern="runtime.**RID**.Microsoft.DotNet.ILCompiler.LLVM" ILCompilerPackVersion="8.0.0" ILCompilerRuntimeIdentifiers="browser-wasm" />
```
As you see it allows indication of the supported "browser-wasm" RID without disrupting other RID targets. Right now it's possible to have workaround like this
```
<KnownILCompilerPack Update="Microsoft.DotNet.ILCompiler"
   TargetFramework="net7.0" ILCompilerPackNamePattern="runtime.**RID**.Microsoft.DotNet.ILCompiler.LLVM" ILCompilerPackVersion="8.0.0-alpha.1.23172.3" ILCompilerRuntimeIdentifiers="browser-wasm;linux-musl-x64;linux-x64;win-x64;linux-arm;linux-arm64;linux-musl-arm;linux-musl-arm64;osx-arm64;osx-x64;win-arm;win-arm64;win-x86" />
```
but with appearace of .NET 8 which has 2 KnownILCompilerPack entry that become problematic and implicit package referencing logic rely on item identity.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -368,7 +368,7 @@ namespace Microsoft.NET.Build.Tasks
                     return;
                 }
             }
-            
+
             if (AotEnabled)
             {
                 if (!AddToolPack(ToolPackType.ILCompiler, _normalizedTargetFrameworkVersion, packagesToDownload, implicitPackageReferences))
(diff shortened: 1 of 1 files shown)


lewing:
can this be closed now?

kant2002:
Technically I think that was sensible change anyway, but if that bring some complexity to SDK, I have no case for that now.

lewing:
I'm sure it will come up again just trying to keep things tidy at the moment

PR #31421, opened 2023-03-23
Adding locked messages as there is overlocalization in at least 5 of …

marcpopMSFT:
…these translations.

@cristianosuzuki77 is this how you lock multiple substrings in a single localized string?  I see over-localization in 5 of the 26 files affected.


src/Tasks/Common/Resources/xlf/Strings.es.xlf
```diff
@@ -812,8 +812,8 @@ The following are names of parameters or literal values and should not be transl
       </trans-unit>
       <trans-unit id="SelfContainedOptionShouldBeUsedWithRuntime">
         <source>NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.</source>
-        <target state="translated">NETSDK1179: Se requiere una de las opciones "--self-contained" o "--no-self-contained" cuando se usa "--runtime".</target>
-        <note>{StrBegin="NETSDK1179: "}</note>
+        <target state="new">NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.</target>
```
joeloff, line 815: Shouldn't `--runtime` also be locked?

marcpopMSFT: Good catch. I also found a --output in the same file so fixed it as well.

joeloff (approved):

PR #31432, opened 2023-03-24
[release/7.0.4xx] include container tasks and targets directly to `Microsoft.NET.Publish.Sdk`

MichalPavlik:
fixes: https://github.com/dotnet/sdk-container-builds/issues/370

This modification depends on insertion of "Containers SDK" tracked in #31302.

It hooks our targets to enable containers publishing.

TODO: 
- [x] include tasks and targets from `Microsoft.NET.Build.Containers` directly to `Microsoft.NET.Publish.Sdk`.
- [x]  `EnableSDKContainersSupport` to be set tor `true` by default
- [x]  warning should be given when explicit `PackageReference` to  `Microsoft.NET.Build.Containers` exists (however it should be possible to override it)
- [x] remove implementation for 7.0.3xx
- [x] adjust tests
- [ ] documentation - will be covered by https://github.com/dotnet/sdk-container-builds/issues/408


src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.props
```diff
@@ -11,4 +11,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props"
+          Condition="Exists('$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props')" />
```
baronfel, line 15: we need to add a condition here on `EnableSdkContainerSupport` being `true`. We also need to flip `EnableSdkContainerSupport` to `true` as part of this work :)

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
```diff
@@ -146,6 +147,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(_DotNetCLIToolTargetsDir)Microsoft.NET.Sdk.DotNetCLITool.targets" Condition="Exists('$(_DotNetCLIToolTargetsDir)Microsoft.NET.Sdk.DotNetCLITool.targets')"/>
 
+<!--
+  ***********************************************************************************************
+  Import the Containers target
+  ***********************************************************************************************
+ -->
+
+  <Import Project="$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets" Condition="Exists('$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets')" />
```
baronfel, line 156: need a check for `EnableSdkContainerSupport` being `true` here as well.

baronfel:
We also need a check to make sure that a user doesn't use both the built-in support as well as the PackageReference version of the tools. We can't add this check as a `Condition` on the `Import` of the props/targets because MSBuild items (like PackageReference) aren't able to be referenced at property-evaluation time, so instead we'll need to add a 'validation' target, like the pre-existing SDK version check target. This target should check if both `EnableSdkContainerSupport` is `true` and if a `PackageReference` with the `Identity` `Microsoft.Net.Build.Containers` is present. If so, a diagnostic should be raised asking the user to either set `EnableSdkContainerSupport` to `false` (to disable the SDK targets), or to remove the `PackageReference` (to use the SDK targets).

baronfel (approved):
LGTM - one minor rename nit.

Good to have this first step done - now to tackle Console publishing :D

src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -215,6 +187,29 @@
...
+
+    <MSBuild
+      Targets="Publish"
+      Projects="$(RepoRoot)/src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj"
+      Properties="Configuration=$(Configuration);PublishDir=$(OutputPath)/Containers/tasks/net472;TargetFramework=net472"
+      Condition="'$(DotNetBuildFromSource)' != 'true'" />
+    <MSBuild
+      Targets="Publish"
+      Projects="$(RepoRoot)/src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj"
+      Properties="Configuration=$(Configuration);PublishDir=$(OutputPath)/Containers/tasks/net7.0;TargetFramework=net7.0" />
```
ladipro, line 206: Is there a prop to get the `net7.0` from so this line doesn't have to be updated manually/annually?

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.props
```diff
@@ -11,4 +11,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <PropertyGroup>
+    <EnableSdkContainerSupport Condition="'$(EnableSdkContainerSupport)' == ''">true</EnableSdkContainerSupport>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props"
+          Condition="Exists('$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props') AND '$(EnableSdkContainerSupport)' == 'true'" />
```
ladipro, line 19: super-nit: Swap the conditions (prop AND exists) so there's no I/O if SDK container support is disabled?

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
```diff
@@ -146,6 +147,14 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(_DotNetCLIToolTargetsDir)Microsoft.NET.Sdk.DotNetCLITool.targets" Condition="Exists('$(_DotNetCLIToolTargetsDir)Microsoft.NET.Sdk.DotNetCLITool.targets')"/>
 
+<!--
+  ***********************************************************************************************
+  Import the Containers target
+  ***********************************************************************************************
+ -->
+
+  <Import Project="$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets" Condition="Exists('$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets') AND '$(EnableSdkContainerSupport)' == 'true'" />
```
ladipro, line 156: super-nit: Same as in `.props`, tiny bit better to try to short-circuit on the property first.

ladipro (approved):

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.props
```diff
@@ -11,4 +11,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <PropertyGroup>
+    <EnableSdkContainerSupport Condition="'$(EnableSdkContainerSupport)' == ''">true</EnableSdkContainerSupport>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props"
```
vijayrkn, line 18: The layout of the file (to out of the sdk path and find this) is odd but I understand we may not have a better way to deal with it. If there was a sdksroot property or something, we could avoid unintentionally breaking things if we decide to move this file somewhere else.

src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets
```diff
@@ -166,6 +175,15 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
   </Target>
 
+  <Target Name="CheckContainersPackage" AfterTargets="Build" Condition="'$(EnableSdkContainerSupport)' == 'true'">
+    <PropertyGroup>
+      <ContainersPackageIdentity>Microsoft.NET.Build.Containers</ContainersPackageIdentity>
+    </PropertyGroup>
+    <ItemGroup>
+      <ContainersPackage Include="@(PackageReference)" Condition="'%(Identity)' == '$(ContainersPackageIdentity)'"/>
+    </ItemGroup>
+    <Warning Text="Microsoft.NET.Build.Containers package is explicitly referenced. Consider removing the package reference to Microsoft.NET.Build.Containers as it is now part of .NET SDK." Condition="'@(ContainersPackage)' != ''" />
```
vijayrkn, line 185: Should this error message say NuGet package to be clear so that it won't be confused with any other kind of package.

vijayrkn (approved):

PR #31454, opened 2023-03-24
Add support for new output path format

dsplaisted:
This ports the artifacts output support from #29599 to main.  This implements the design in https://github.com/dotnet/designs/pull/281.

In addition to porting the code, it also includes the following:

- De-duplicate `ArtifactsPath` calculation logic by adding a .props file which can be imported from two different places depending on whether there are any artifacts output properties set in Directory.Build.props (or the command line) or not.
- Add telemetry to help us understand how and how much the artifacts output format is used.

Still to do:

- [x] De-duplicate test solution creation code
- [ ] Add hook to MSBuild to inject targets after `Directory.Build.props` import, and use that in SDK instead of duplicate import logic.


javiercn (approved):
Changes from the ASP.NET Core perspective look good. Let us know if you need further input from our side

nagilson (approved):
The port and solution consolidation looks great to me. Thank you for consolidating those. :)

PR #31460, opened 2023-03-27
Update off-by-default analyzers to account for rename

### LINKED ISSUE #47359: React to RDG name change in SDK for preview4
In https://github.com/dotnet/aspnetcore/pull/47334, we renamed the RDG generator project (and assembly name) to make it clearer that it contains only one generator.

Now that the change has flown to the installer, we need to update the off-by-default logic in the SDK to react to this name change:

https://github.com/dotnet/sdk/blob/e3b39b8db045f14193a29726cb193cea8913e0dc/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets#L421-L437

This change will require updating the installer used to the build the SDK repo and updating the target above.

In the meantime, the request delegate generator will be **on by default** for all projects using the preview4 runtime.

captainsafia:
Closes https://github.com/dotnet/aspnetcore/issues/47359.


javiercn:
@ilonatommy is this something that was recently added? `_framework\dotnet.timezones.blat`

The PR is updating the SDK from preview1 to preview4 and we are trying to determine if there is a bug.

From my understanding, it should be ok to update the baselines as this is just an extra file. https://github.com/dotnet/sdk/blob/main/src/RazorSdk/update-test-baselines.ps1

captainsafia:
I updated the baselines in https://github.com/dotnet/sdk/pull/31460/commits/a31a53e6fd51e04b9ebb30a76e126378c59bf94d but happy to revert if that ends up not being necessary.

Locally, I did observe that I was still seeing test failures related to `_framework\dotnet.timezones.blat` after updating the baselines but we'll see if they repro in CI.

javiercn:
@captainsafia I think looking at your commit made things clearer.

I believe @pavelsavara mentioned that this change was coming, but I'll let him confirm.

javiercn:
@captainsafia the "additionally" failing tests might be tests that are not using baselines, it should be fine to just "remove" the assertion on those.

captainsafia:
> @captainsafia the "additionally" failing tests might be tests that are not using baselines, it should be fine to just "remove" the assertion on those.

Quite right! Just pushed https://github.com/dotnet/sdk/pull/31460/commits/b0a0ea46156145624d74794d8c3cc942e5da4e75 with this.

ilonatommy:
> @ilonatommy is this something that was recently added? `_framework\dotnet.timezones.blat`
> 
> The PR is updating the SDK from preview1 to preview4 and we are trying to determine if there is a bug.
> 
> From my understanding, it should be ok to update the baselines as this is just an extra file. https://github.com/dotnet/sdk/blob/main/src/RazorSdk/update-test-baselines.ps1

On the contrary, `dotnet.timezones.blat` was there since a while already. Then, this PR: https://github.com/dotnet/runtime/pull/82250 removed it and moved the data to be embedded into `.wasm` file.

javiercn:
@ilonatommy @maraf just for the sake of completeness, it's correct that the file is being removed by the update, correct?

maraf:
> @ilonatommy @maraf just for the sake of completeness, it's correct that the file is being removed by the update, correct?

Yes, as Ilona mentioned, it is bundled in the wasm file. And we need to fix `BlazorEnableTimeZoneSupport`

maraf:
@radekdoulik Do you have any idea why it picks emscripten 3.1.34?

https://dev.azure.com/dnceng-public/public/_build/results?buildId=218974&view=ms.vss-test-web.build-test-results-tab&runId=4098548&resultId=100001&paneView=attachments

radekdoulik:
> @radekdoulik Do you have any idea why it picks emscripten 3.1.34?

The CI is using docker images with 3.1.34. https://github.com/dotnet/runtime/issues/83986 should help

maraf:
> The CI is using docker images with 3.1.34. [dotnet/runtime#84014](https://github.com/dotnet/runtime/pull/84014) should help

The workload manifests for 3.1.34 are packed in the SDK

lewing:
@dsplaisted here we're picking up emsdk packs that are ahead of what is in runtime.  I think we need the rollback file sooner rather than later.

captainsafia:
> @captainsafia Is it possible to use preview3?

Unfortunately not, this is in reaction to a change that was made to preview4.

lewing:
I think the best move for the moment is to disable the AOT tests while we sort out how to make them work in any sane way here.

lewing:
@captainsafia the current aot errors will go away once https://github.com/dotnet/emsdk/pull/325 lands and publishes to the dotnet8 feed

captainsafia:
> @captainsafia the current aot errors will go away once [dotnet/emsdk#325](https://github.com/dotnet/emsdk/pull/325) lands and publishes to the dotnet8 feed

OK, I can skip the tests in this PR and open a follow-up issue to revert once that change has flowed through.

maraf (approved):
Runtime related changes looks good to me 👍

The `.blat` file could also be removed from this test (but the test is not asserting anything...) https://github.com/dotnet/sdk/blob/712dfa7c9b21acda4cd4f55b02c8cb9c5200ce0e/src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/ComputeBlazorPublishAssetsTest.cs#L60-L75

We will update this test once we have a working solution https://github.com/dotnet/sdk/blob/712dfa7c9b21acda4cd4f55b02c8cb9c5200ce0e/src/Tests/Microsoft.NET.Sdk.BlazorWebAssembly.Tests/WasmBuildIntegrationTest.cs#L236

maraf:
I'm sorry, I didn't realize the PR will merged when I add approval..
TIL about this feature

captainsafia:
> I'm sorry, I didn't realize the PR will merged when I add approval..

Ah yes, I did this intentionally as I was keen to get this change (it breaks the off-by-default support for the RequestDelegateGenerator). 

Should I make a new issue to track [this follow-up](https://github.com/dotnet/sdk/pull/31460#pullrequestreview-1362613219)?

maraf:
> Should I make a new issue to track this follow-up?

@captainsafia it's ok, we'll do it once we have the solution.

PR #31519, opened 2023-03-29
Wasm SDK packed as a nuget package

maraf:
Move bits of Wasm SDK that are expected to change more frequently to the runtime repo.

- Add `Microsoft.NET.Sdk.WebAssembly.Pack` implicit package reference if `UsingMicrosoftNETSdkWebAssembly=true`.
- Provide `_WebAssemblyPropsFile` and `_WebAssemblyTargetsFile` properties to enable overriding props and targets.
- Move tests for Wasm SDK tasks to runtime repo. Leave tests checking boot.json in place at the moment.
- Return some default property values back to Blazor SDK to support .NET 5.

Contributes to https://github.com/dotnet/runtime/issues/81367


src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -688,13 +703,13 @@ private enum ToolPackType
             }
 
             // ILCompiler and ILLink have RID-agnostic build packages that contain MSBuild targets.
-            if (toolPackType is ToolPackType.ILCompiler or ToolPackType.ILLink)
+            if (toolPackType is ToolPackType.ILCompiler or ToolPackType.ILLink or ToolPackType.WebAssemblySdk)
```
lewing, line 706: Not a strong preference here but it seems like
```suggestion
            if (toolPackType is not ToolPackType.Crossgen2)
```
is slightly cleaner

lewing (commented):
This is looking good, what else is left?

maraf:
> This is looking good, what else is left?

lewing:
@rainersigwald @tmat @mthalman I'm having a little trouble following all the flow here but are we expecting the failures here to be resolved once everthing flows? https://github.com/dotnet/sourcelink/issues/1007

tmat:
@lewing I think so.

lewing:
triggered https://github.com/dotnet/sdk/pull/31953 but also pulled the bump in here to verify

lewing (approved):

lewing:
@javiercn @MackinnonBuck any objections?

javiercn (approved):

PR #31545, opened 2023-03-30
GenAPI should not filter out base generic interfaces.

### LINKED ISSUE #31484: GenAPI filters out generic interfaces for implementations
Affected version - `8.0.100-preview.3.23170.2`

Steps to reproduce:
  - generate `System.Collections.Immutable: 6.0.0` using [SBRP](https://github.com/dotnet/source-build-reference-packages/blob/main/generate.sh)
  - build the resulting project (`./build.sh --projects <path_to_csproj> `)

Several classes will be generated incorrectly. For example:
```
    public sealed partial class ImmutableHashSet<T> : IEnumerable, ICollection
    {
        ... 
        bool Generic.ICollection<T>.IsReadOnly { get { throw null; } }
		...
    }
```

with the exception:
```
'ImmutableHashSet<T>.ICollection<T>.IsReadOnly': containing type does not implement interface 'ICollection<T>'
```

where `ICollection` is a public type ([source](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/ICollection.cs#L8)):
```
	public interface ICollection<T> : IEnumerable<T>
	{
		...
	}
```

Adding the generic manually to the interface list eliminates the issue.

CC: @andriipatsula

andriipatsula:
Fixes: https://github.com/dotnet/sdk/issues/31484

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
@@ -177,7 +177,7 @@ namespace Microsoft.DotNet.GenAPI
         /// <param name="symbolFilter">Assembly symbol filter <see cref="ISymbolFilter"/>.</param>
         /// <returns>Boolean</returns>
         public static bool HasInaccessibleTypeArgument(this INamedTypeSymbol namedType, ISymbolFilter symbolFilter)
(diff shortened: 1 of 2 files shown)


ericstj (approved):
LGTM, thank you for the fix!

ViktorHofer (commented):
LGTM. Thanks

Issue #9500, opened 2018-06-13
dotnet tool install should return exit code 0 if tool is already installed


natemcmaster:
I'm trying to write a script that will install global tools if necessary. I would like to just run "dotnet tool install", but this returns exit code 0 if the tool is already present, so I have to first execute `dotnet tool list` and grep the output.

## Expected  behavior
Calling `dotnet tool install` should no-op AND exit code 0 if the tool is already installed

## Actual behavior
Running dotnet tool install twice fails scripts because the second time it runs it will exit code 0
```
PS> & dotnet tool install --tool-path "$(pwd)/.tools" sleet --version 2.3.25 --add-source https://api.nuget.org/v3/index.json
You can invoke the tool using the following command: sleet
Tool 'sleet' (version '2.3.25') was successfully installed.
PS> $lastexitcode
0
PS>  & dotnet tool install --tool-path "$(pwd)/.tools" sleet --version 2.3.25 --add-source https://api.nuget.org/v3/index.json
Tool 'sleet' is already installed.
PS>  $lastexitcode
1
```

The result of this is that I have write more complicate code like this:

```ps1
if (& $dotnet tool list --tool-path "$PSScriptRoot/.tools" | Select-String "sleet") {
    Write-Host -f Yellow 'Skipping install of sleet. It''s already installed'
}
else {
    Invoke-Block { & $dotnet tool install --tool-path "$PSScriptRoot/.tools" sleet --version 2.3.25 --add-source https://api.nuget.org/v3/index.json }
}
```

## Environment data
`dotnet --info` output:
```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.300
 Commit:    adab45bf0c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\dev\aspnet\Universe20\.dotnet\sdk\2.1.300\

Host (useful for support):
  Version: 2.1.0
  Commit:  caa7b7e2ba

.NET Core SDKs installed:
  2.1.300 [C:\dev\aspnet\Universe20\.dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.0 [C:\dev\aspnet\Universe20\.dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #31549, opened 2023-03-30
9500 dotnet tool install: Add automatically tool upgrade and downgrade when specify version

JL03-Yue:
#9500 

Update `dotnet tool install` so that it support automatic upgrade, downgrade, and error-throwing behaviors. 

- If the command specify version 3.0.0 without `--allow-downgrade` flag and version 4.0.0 is installed, it throws out an error.
- If the command specify version 3.0.0 with`--allow-downgrade` flag and version 4.0.0 is installed, downgrade to 3.0.0. Exit code = 0


src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallLocalCommand.cs
```diff
@@ -48,12 +51,70 @@ internal class ToolInstallLocalCommand : CommandBase
...
+            }
+
+            var existingPackage = existingPackageWithPackageId.Single();
+            var toolDownloadedPackage = _toolLocalPackageInstaller.Install(manifestFile);
+
+            InstallLogic(existingPackage, toolDownloadedPackage, manifestFile);
+            return 0;
+        }
+
+        public int InstallLogic(ToolManifestPackage existingPackage, IToolPackage toolDownloadedPackage, FilePath manifestFile)
```
dsplaisted, line 74: I'd suggest renaming this to `InstallToolUpdate` or something similar to make it clearer what it does.

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallLocalCommand.cs
```diff
@@ -48,12 +51,70 @@ internal class ToolInstallLocalCommand : CommandBase
...
+            }
+
+            _localToolsResolverCache.SaveToolPackage(
+                toolDownloadedPackage,
+                _toolLocalPackageInstaller.TargetFrameworkToInstall);
+
+            return 0;
         }
 
         public int Install(FilePath manifestFile)
```
dsplaisted, line 120: This could be renamed to `InstallNewTool` to better describe it and differentiate it from the previous method.

src/Cli/dotnet/commands/dotnet-tool/update/ToolUpdateLocalCommand.cs
```diff
@@ -77,71 +77,29 @@ public override int Execute()
             (FilePath? manifestFileOptional, string warningMessage) =
```
dsplaisted, line 77: It looks like the only difference between this `Execute` method and the one for ToolInstallCommand is that this one may print a warning message.  Can we add that to the ToolInstallCommand and then remove the duplication by having this one call that `Execute` method?

src/Tests/dotnet.Tests/CommandTests/ToolInstallLocalCommandTests.cs
```diff
@@ -316,6 +322,122 @@ public void WhenRunWithPrereleaseAndPackageVersionItShouldSucceed()
...
+                        _packageIdA,
+                        _packageVersionA.ToNormalizedString(),
+                        _manifestFilePath));
+        }
+
+        [Fact]
+        public void GivenFeedVersionIsLowerRunPackageIdItShouldThrow()
+        {
+            GetDefaultTestToolInstallLocalCommand().Execute().Should().Be(0);
+            _mockFeed.Packages.Single().Version = "0.9.0";
```
dsplaisted, line 346: I think this scenario probably shouldn't fail.  If you run `dotnet tool install` and version 1.0.4 is already installed but the highest version in the feed is 0.9.0, then probably it should just leave 1.0.4 installed without failing (it should print a message letting you know it wasn't updated), the same as would happen if 1.0.4 was the highest version on the feed.  What do you think @baronfel?

There should be a separate test where if you have version 1.0.4 installed and the feed has 0.9.0 and 1.0.4 available, if you try to install and specify `--version 0.9.0`, then it should fail.

src/Tests/dotnet.Tests/CommandTests/ToolInstallLocalCommandTests.cs
```diff
@@ -316,6 +322,122 @@ public void WhenRunWithPrereleaseAndPackageVersionItShouldSucceed()
...
+            Action a = () => GetDefaultTestToolInstallLocalCommand().Execute();
+            a.Should().Throw<GracefulException>().And.Message.Should().Contain(string.Format(
+                LocalizableStrings.UpdateToLowerVersion,
+                "0.9.0",
+                _packageVersionA.ToNormalizedString(),
+                _manifestFilePath));
+        }
+
+        [Fact]
+        public void GivenFeedVersionIsLowerWithAllowDowngradeOptionRunPackageIdItShouldUpdateToLowerVersion()
```
dsplaisted, line 358: Similar to the previous comment, I don't think in this case we should downgrade unless you explicitly specified the version to downgrade to.

src/Tests/dotnet.Tests/CommandTests/ToolInstallLocalCommandTests.cs
baronfel: Agree with @dsplaisted here.

JL03-Yue:
Reopened at #35088

src/Cli/dotnet/commands/dotnet-tool/update/ToolUpdateLocalCommand.cs
JL03-Yue: Resolved in #35088

PR #31559, opened 2023-03-30
Move asset compression to the StaticWebAssets SDK

### LINKED ISSUE #46998: Move asset compression to the Static Web Assets SDK from the Blazor Webassembly SDK
Static file compression happens within the Blazor pipeline, we should not require the Blazor SDK to leverage compression of assets.

* The general design should follow an approach similar to the one provided by scoped CSS or JS modules.
* We need to have a `CompressedFile` item group or similar that can be used to declare that the file needs to be compressed.
* We probably want to have a `CompressFilePattern` item group so that we can declare patterns for files we want to compress.
  * Our tasks already support evaluating glob patterns, so this is "easy" to do.
  * We need to consider whether we want to also support "exclude" patterns, as this makes easy to filter out other content that might have already been processed ahead of time. (Images, fonts, etc.)
* We need to have two targets one for compressing during build and one for compressing during publish.
* Any part of the pipeline that requires compression must run before the build/publish compression targets.
* The build and publish compression targets need to consume the list of static web assets up to the point of generation and the `CompressedFile` items, as they'll produce new Static Web Assets representing the final files.
* Static Web Assets representing compressed files contain metadata that correlates them to their uncompressed version, that can be used for locating the original asset.

MackinnonBuck:
## Move asset compression to the StaticWebAssets SDK

This PR moves the asset compression feature from the BlazorWebAssembly SDK  to the StaticWebAssets SDK.

### List of major changes

1. Separated the compression step from the "resolution" step
    * Before this PR, the resolution of assets to compress and the compression of said assets would happen in the same MSBuild target. This PR separates these steps by allowing the specification of a "compression configuration" that the StaticWebAssets SDK picks up to resolve the final compressed `StaticWebAsset` definitions before the actual compression happens.
      * The compression configuration specifies:
        * The compression format
        * The stage to perform the compression
        * Optional include/exclude patterns to match files to be compressed
      * Assets can also be explicitly added to a configuration by adding items to the `AssetsToCompress` item group
      * Compression occurs after all `StaticWebAsset` items have been added
2. Added unit tests for the new `ResolveCompressedAssets` task
3. Copied the existing `BrotliCompress` and `GZipCompress` tasks to `M.NET.Sdk.StaticWebAssets.Tasks` and updated them accordingly
    * This was a copy instead of a move in order to retain backwards compatibility
4. Removed the now unused `ComputeBlazorFilesToCompress` task and its corresponding tests.

Fixes https://github.com/dotnet/aspnetcore/issues/46998


src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -463,6 +476,22 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+        Condition="'%(AssetKind)' != 'Build' and ('%(StaticWebAsset.AssetTraitName)' == 'BlazorWebAssemblyResource' or '%(StaticWebAsset.AssetTraitName)' == 'Culture')">
+        <ConfigurationName>BlazorPublishGzip</ConfigurationName>
+      </AssetToCompress>
+
+      <AssetToCompress
+        Include="@(StaticWebAsset)"
+        Condition="'%(AssetKind)' != 'Build' and ('%(StaticWebAsset.AssetTraitName)' == 'BlazorWebAssemblyResource' or '%(StaticWebAsset.AssetTraitName)' == 'Culture')">
+        <ConfigurationName>BlazorPublishBrotli</ConfigurationName>
+      </AssetToCompress>
+    </ItemGroup>
```
MackinnonBuck, line 492: Maybe we could combine these into one item, something like:
```targets
<AssetToCompress
  Include="@(StaticWebAsset)"
  Condition="'%(AssetKind)' != 'Build' and ('%(StaticWebAsset.AssetTraitName)' == 'BlazorWebAssemblyResource' or '%(StaticWebAsset.AssetTraitName)' == 'Culture')">
  <ConfigurationName>BlazorPublishGzip;BlazorPublishBrotli</ConfigurationName>
</AssetToCompress>
```

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -107,8 +104,55 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <!--
+      Compress referenced binaries using GZip during build. This skips files such as the project's assemblies
+      that change from build to build. Runtime assets contribute to the bulk of the download size. Compressing it
+      has the most benefit while avoiding any ongoing costs to the dev inner loop.
+    -->
+    <CompressionConfiguration Include="BlazorBuildGzip">
+      <Stage>Build</Stage>
+      <Format>gzip</Format>
+      <IncludePattern>_framework\**</IncludePattern>
+      <ExcludePattern>_framework\blazor.boot.json</ExcludePattern>
```
javiercn, line 139: Is there a reason why we need to exclude `_framework\blazor.boot.json`?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: Is there a reason why the patterns don't take care of these?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -107,8 +104,55 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      ComputeBlazorExtensions;
+      _AddPublishBlazorBootJsonToStaticWebAssets;
+      _CollectAssetsToCompress;
+    </ResolveCompressedFilesForPublishDependsOn>
+
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <_BlazorBuildGZipCompressDirectory>$(IntermediateOutputPath)build-gz\</_BlazorBuildGZipCompressDirectory>
+    <_BlazorPublishCompressDirectory>$(IntermediateOutputPath)compress\</_BlazorPublishCompressDirectory>
```
javiercn, line 125: This should just be defined by the compression targets. There is no need for consumers to make a choice on this.

Ideally, all compressed assets end up in `obj\fx\config\compressed\(publish?)` in a flat list of files where the name is a "hash" of the relative path of the asset + the .br or .gz extension.

Move [this](https://github.com/dotnet/sdk/blob/a821467d6ef6f4d3f9c3de8c468041d8c8d5b3b0/src/WasmSdk/Tasks/FileHasher.cs) class around and use it to compute the name (instead of what we currently use, which I think is base64 + some character replacements).

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
```diff
@@ -0,0 +1,266 @@
...
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.Extensions.FileSystemGlobbing;
+
+namespace Microsoft.AspNetCore.StaticWebAssets.Tasks;
+
+public class ResolveCompressedAssets : Task
```
javiercn, line 16: FYI,

This is logic that we can easily unit test.

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        for (var i = 0; i < 8; i++)
+        {
+            var c = hashString[i];
+            builder.Append(s_invalidPathChars.Contains(c) ? '+' : c);
+        }
+
+        builder.Append(extension);
+        return builder.ToString();
+    }
```
javiercn, line 205: Use `FileHasher` instead to produce a name https://github.com/dotnet/sdk/blob/a821467d6ef6f4d3f9c3de8c468041d8c8d5b3b0/src/WasmSdk/Tasks/FileHasher.cs#L12

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
```diff
@@ -0,0 +1,266 @@
...
+
+    private sealed class CompressionConfiguration
+    {
+        private readonly Matcher _matcher = new();
+
+        public string ItemSpec { get; set; }
+
+        public string TargetDirectory { get; }
```
javiercn, line 213: I mentioned elsewhere, but we don't need the consumer to define the target directory.

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
javiercn, line 16: Log pretty much everything as shown below so that we can easily troubleshoot binlogs
https://github.com/dotnet/sdk/blob/main/src/StaticWebAssetsSdk/Tasks/DefineStaticWebAssets.cs#L100

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
MackinnonBuck: Kind of. In this case, the criterion for compressing a file is expressed in terms of its `AssetTraitName` instead of its file name. We could maybe define a very specific pattern that includes the right file names and file extensions to perfectly match this subset (which I did try), but even this won't work if the customer defines their own `BlazorPublishExtension` item that doesn't conform to that pattern.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
MackinnonBuck: It's currently not included in the baselines. Should it be?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
MackinnonBuck: Sounds good. I've changed it so the consumer no longer needs to specify anything. When this PR is closer to its final state, I'll change the output directories to match what you wrote so the baseline updates happen at the end.

MackinnonBuck:
Source build failures seem unrelated. They're happening in `main`, too.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -63,25 +60,58 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <Format>gzip</Format>
+      <TargetDirectory>$(_BlazorPublishCompressDirectory)</TargetDirectory>
+    </CompressionConfiguration>
+    <CompressionConfiguration Include="BlazorPublishBrotli">
+      <Stage>Publish</Stage>
+      <Format>brotli</Format>
+    </CompressionConfiguration>
+
+  </ItemGroup>
+
```
javiercn, line 114: I think this should be inside the wasm SDK, and not inside the Blazor specific bits (since all the runtime functionality moved there)

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -63,25 +60,58 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <CompressionConfiguration Include="BlazorPublishGzip">
+      <Stage>Publish</Stage>
+      <Format>gzip</Format>
+      <TargetDirectory>$(_BlazorPublishCompressDirectory)</TargetDirectory>
+    </CompressionConfiguration>
+    <CompressionConfiguration Include="BlazorPublishBrotli">
+      <Stage>Publish</Stage>
+      <Format>brotli</Format>
+    </CompressionConfiguration>
```
javiercn, line 111: Why do these two items don't have a pattern defined?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -63,25 +60,58 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <Stage>Build</Stage>
+      <Format>gzip</Format>
+      <IncludePattern>_framework\**</IncludePattern>
+      <ExcludePattern>_framework\blazor.boot.json</ExcludePattern>
+    </CompressionConfiguration>
+
+    <CompressionConfiguration Include="BlazorPublishGzip">
+      <Stage>Publish</Stage>
+      <Format>gzip</Format>
+      <TargetDirectory>$(_BlazorPublishCompressDirectory)</TargetDirectory>
```
javiercn, line 106: Is this still used, or did we get rid of it, in favor of a single location controlled by the compression targets

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -169,7 +163,23 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+        <ConfigurationName>BlazorPublishGzip</ConfigurationName>
+      </AssetToCompress>
+
+      <AssetToCompress
+        Include="@(StaticWebAsset)"
+        Condition="'%(AssetKind)' != 'Build' and ('%(StaticWebAsset.AssetTraitName)' == 'WasmResource' or '%(StaticWebAsset.AssetTraitName)' == 'Culture')">
+        <ConfigurationName>BlazorPublishBrotli</ConfigurationName>
+      </AssetToCompress>
+    </ItemGroup>
+  </Target>
```
javiercn, line 182: Could we express this in terms of file patterns?

MackinnonBuck: Maybe. I tried this but found it caused issues in some cases. See https://github.com/dotnet/sdk/pull/31559#discussion_r1154881408

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
MackinnonBuck: Yeah I don't think this does anything anymore, I'll remove it.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
MackinnonBuck: Currently there's no pattern because the assets are defined only explicitly. See https://github.com/dotnet/sdk/pull/31559#discussion_r1154881408

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -169,7 +163,23 @@ Copyright (c) .NET Foundation. All rights reserved.
     <Message Importance="High" Text="$(MSBuildProjectName) (Blazor output) -&gt; $(TargetDir)wwwroot" Condition="'$(CopyBuildOutputToOutputDirectory)' == 'true' and '$(SkipCopyBuildProduct)'!='true'" />
   </Target>
 
-  <!-- Publish starts here -->
+  <Target
```
maraf, line 166: Shoudn't this also be in the Wasm SDK?

src/StaticWebAssetsSdk/Targets/Sdk.StaticWebAssets.CurrentVersion.targets
```diff
@@ -31,7 +31,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <_StaticWebAssetsSdkTasksTFM Condition=" '$(MSBuildRuntimeType)' == 'Core'">net8.0</_StaticWebAssetsSdkTasksTFM>
     <_StaticWebAssetsSdkTasksTFM Condition=" '$(_StaticWebAssetsSdkTasksTFM)' == ''">net472</_StaticWebAssetsSdkTasksTFM>
     <StaticWebAssetsSdkBuildTasksAssembly>$(StaticWebAssetsSdkBuildTasksDirectoryRoot)$(_StaticWebAssetsSdkTasksTFM)\Microsoft.NET.Sdk.StaticWebAssets.Tasks.dll</StaticWebAssetsSdkBuildTasksAssembly>
-    <_StaticWebAssetsSdkToolAssembly>$(StaticWebAssetsSdkDirectoryRoot)..\Microsoft.NET.Sdk.Razor\tools\rzc.dll</_StaticWebAssetsSdkToolAssembly>
+    <_StaticWebAssetsSdkToolAssembly>$(StaticWebAssetsSdkDirectoryRoot)tools\net8.0\Microsoft.NET.Sdk.StaticWebAssets.Tool.dll</_StaticWebAssetsSdkToolAssembly>
```
maraf, line 34: Is it correct to always have net8.0 here?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: We should hopefully not need this target.

I am hoping we can express everything in terms of patterns, which makes things more declarative.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -63,25 +60,58 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   </PropertyGroup>
 
+  <ItemGroup>
+
+    <!--
+      Compress referenced binaries using GZip during build. This skips files such as the project's assemblies
+      that change from build to build. Runtime assets contribute to the bulk of the download size. Compressing it
+      has the most benefit while avoiding any ongoing costs to the dev inner loop.
+    -->
+    <CompressionConfiguration Include="BlazorBuildGzip">
```
javiercn, line 96: For the compression configuration we should have "Build, Publish, All". With "All" meaning that it runs for the pattern during build and publish. For example, we want to gzip all the bits during build, but we also want to gzip them in publish, and there's no value in having to specify two separate configuration entries.

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: I am hoping that we can define the patterns like this on the Wasm SDK

```
  <!-- Wire-up compression -->
    <ItemGroup>

      <CompressionConfiguration Include="WebassemblyGzip">
        <Stage>All</Stage>
        <Format>gzip</Format>
        <IncludePattern>_framework\**</IncludePattern>
      </CompressionConfiguration>

      <CompressionConfiguration Include="WebassemblyBrotli">
        <Stage>Publish</Stage>
        <Format>brotli</Format>
        <IncludePattern>_framework\**</IncludePattern>
      </CompressionConfiguration>

    </ItemGroup>
```

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
javiercn: Were you running into ordering problems?

src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.Compression.targets
```diff
@@ -0,0 +1,184 @@
...
+    <CompressFilesDependsOn>
+      $(CompressFilesDependsOn);
+      CopyStaticWebAssetsToOutputDirectory;
+    </CompressFilesDependsOn>
+    <CompressFilesForPublishDependsOn>$(CompressFilesForPublishDependsOn)</CompressFilesForPublishDependsOn>
+  </PropertyGroup>
+
+  <PropertyGroup>
+
+    <GenerateStaticWebAssetsPublishManifestDependsOn Condition="'$(WasmBuildingForNestedPublish)' != 'true'">
```
javiercn, line 31: We should not have this condition here. Instead, the Wasm SDK needs to disable compression when `WasmBuildingForNestedPublish`. It can probably do so by setting the prop based on `WasmBuildingForNestedPublish` before the import for the Static Web Assets SDK targets

src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.Compression.targets
```diff
@@ -0,0 +1,184 @@
...
+      $(GenerateStaticWebAssetsPublishManifestDependsOn);
+      _CompressFilesForPublish;
+    </GenerateStaticWebAssetsPublishManifestDependsOn>
+
+    <StaticWebAssetsPrepareForRunDependsOn>
```
javiercn, line 36: There is I believe a better hook for this already in the SDK. `GenerateComputedBuildStaticWebAssetsDependsOn`

javiercn: There is one for build and one for publish

src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.Compression.targets
```diff
@@ -0,0 +1,184 @@
...
+      <StaticWebAsset Include="@(_CompressedStaticWebAssetsForPublish->'%(FullPath)')" />
+      <FilesToCompress Remove="@(FilesToCompress)" />
+    </ItemGroup>
+
+  </Target>
+
+  <Target
+    Name="_CompressFiles"
+    DependsOnTargets="$(CompressFilesDependsOn)"
+    BeforeTargets="_SplitStaticWebAssetsByCopyOptions">
```
javiercn, line 124: You should not need this if you use `GenerateComputedBuildStaticWebAssetsDependsOn` I believe

src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.Compression.targets
```diff
@@ -0,0 +1,184 @@
...
+    </ItemGroup>
+
+    <GZipCompress Condition="'@(_GZipCompressedStaticWebAssets)' != ''"
+      FilesToCompress="@(_GZipCompressedStaticWebAssets)" />
+
+    <PropertyGroup Condition="'$(DOTNET_HOST_PATH)' == ''">
+      <_DotNetHostDirectory>$(NetCoreRoot)</_DotNetHostDirectory>
+      <_DotNetHostFileName>dotnet</_DotNetHostFileName>
+      <_DotNetHostFileName Condition="'$(OS)' == 'Windows_NT'">dotnet.exe</_DotNetHostFileName>
+    </PropertyGroup>
```
javiercn, line 138: Not sure where this is defined, but it's duplicated across targets.

Can we try putting it on a global propertygroup (outside of the target) or factor it into a common target?

src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.Compression.targets
```diff
@@ -0,0 +1,184 @@
...
+
+  <Target
+    Name="_ResolveCompressedFilesForPublish"
+    DependsOnTargets="$(ResolveCompressedFilesForPublishDependsOn)">
+
+    <ItemGroup>
+
+      <_CompressedFileIncludePatternForPublish
+        Include="@(CompressionConfiguration)"
+        Condition="'%(Stage)' == 'Publish'" />
```
javiercn, line 94: != 'Build'

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
```diff
@@ -0,0 +1,256 @@
...
+                    "Ignoring asset '{0}' for compression because it is already compressed.",
+                    asset.ItemSpec);
+                continue;
+            }
+
+            var relativePath = asset.GetMetadata("RelativePath");
+
+            foreach (var configuration in compressionConfigurations)
+            {
+                if (configuration.Matches(relativePath))
```
javiercn, line 131: All the include/excludes should be evaluated against all the assets (or we should have a global set of `excludes`) that can be defined globally.

We want to avoid situations where if we said, never to compress a format (for example .woff) we accidentally include it because a separate pattern was too loose and included it.

src/StaticWebAssetsSdk/Targets/Sdk.StaticWebAssets.CurrentVersion.targets
MackinnonBuck: We do something similar in [Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets](https://github.com/dotnet/sdk/blob/0961189a48b0763d7cccba637539c876ed29f43f/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets#L33), where we've just been updating the TFM manually. Not sure why we haven't been parameterizing it. @javiercn what do you think?

javiercn: I think so, since this gets put in the SDK, we don't want to evaluate the condition based on the user target framework (that would be wrong) and we don't have a mechanism to pick it up automagically from the TFM the project is targeting, so I think having this hardcoded is ok.

It will break immediately when we update the repo branding.

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
MackinnonBuck: Sounds good. I've added a `GlobalCompressionExcludePattern` MSBuild property to decide which exclude patterns are "global".

src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.ScopedCss.5_0.targets
```diff
@@ -190,7 +190,7 @@ Integration with static web assets:
   <MakeDir Directories="$(_ScopedCssIntermediatePath)" />
   <RewriteCss
     FilesToTransform="@(_ScopedCss)"
-    ToolAssembly="$(_StaticWebAssetsSdkToolAssembly)"
+    ToolAssembly="$(_StaticWebAssetsSdkRazorToolAssembly)"
```
javiercn, line 193: Is Razor needed for a reason here?

src/StaticWebAssetsSdk/Tasks/ResolveCompressedAssets.cs
```diff
@@ -0,0 +1,309 @@
...
+
+        var fileName = FileHasher.GetFileHash(originalItemSpec) + fileExtension;
+        var outputRelativePath = Path.Combine(targetDirectory, fileName);
+
+        result = new TaskItem(outputRelativePath, asset.CloneCustomMetadata());
+
+        result.SetMetadata("RelativePath", relativePath + fileExtension);
+        result.SetMetadata("OriginalItemSpec", asset.ItemSpec);
+        result.SetMetadata("RelatedAsset", asset.ItemSpec);
+        result.SetMetadata("RelatedAssetOriginalItemSpec", originalItemSpec);
```
javiercn, line 266: I see now, it's because we are putting everything in the same result.

It's ok, but when you add this to the StaticWebAsset ItemGroup use the RemoveMetadata to get rid of this property.

javiercn (approved):
A few minor details, but otherwise looks great!

maraf (approved):

javiercn (approved):

src/StaticWebAssetsSdk/Targets/Sdk.StaticWebAssets.CurrentVersion.targets
```diff
@@ -31,7 +31,8 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <_StaticWebAssetsSdkRazorToolAssembly>$(StaticWebAssetsSdkDirectoryRoot)..\Microsoft.NET.Sdk.Razor\tools\rzc.dll</_StaticWebAssetsSdkRazorToolAssembly>
```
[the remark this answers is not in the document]
MackinnonBuck: https://github.com/dotnet/aspnetcore/issues/47374#issuecomment-1515097057

PR #31590, opened 2023-04-03
Infer analyzer settings from EnableAotAnalyzer

### LINKED ISSUE #31284: EnableAotAnalyzer=true should also default EnableSingleFileAnalyzer=true
There are almost no situations where I want to enable the AOT analyzer, but don't also want the Single File analyzer. We should set `EnableSingleFileAnalyzer=true` by default when someone sets `EnableAotAnalyzer=true`.

In dotnet/aspnetcore we enabled the AOT analyzer, thinking that would be enough to catch NativeAOT warnings. But we didn't remember to enable the SingleFile analyzer, and thus warnings slipped through.

@vitek-karas @agocke @sbomer @MichalStrehovsky

agocke:
Fixes https://github.com/dotnet/sdk/issues/31284


sbomer (approved):
LGTM, thanks!

agocke:
Closing this out since we're gonna go in a different direction

PR #31591, opened 2023-04-04
Disable BinaryFormatter across most .NET 8 project types

GrabYourPitchforks:
Ref: https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md

In .NET 7, we obsoleted (as error) the BinaryFormatter APIs, but existing compiled code could still call it and it would work. There was no runtime block except in certain project types like aspnet, Blazor, and MAUI.

In .NET 8, we're flipping this so that BinaryFormatter is now disabled across all project types except WinForms and WPF, which need a little bit more time to work on removal. If you retarget a .NET 7 console application to .NET 8 and that app was using BinaryFormatter under the covers, calls to BF will now fail at runtime.

Just like in .NET 7, applications can continue to use the `<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>` compat switch to opt back in to restoring BinaryFormatter behavior. Setting that compat switch suppresses _both_ the compile-time error (same as .NET 7) _and_ the .NET 8 runtime disablement.

For more information on the matrix, see the comment in **Microsoft.NET.Sdk.targets** within this PR.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -130,6 +130,44 @@ Copyright (c) .NET Foundation. All rights reserved.
     <IncludeProjectsNotInAssetsFileInDepsFile Condition="'$(IncludeProjectsNotInAssetsFileInDepsFile)' == ''">true</IncludeProjectsNotInAssetsFileInDepsFile>
   </PropertyGroup>
 
+  <!--
+    BinaryFormatter killbitting
+    ===========================
+
+    The EnableUnsafeBinaryFormatterSerialization setting controls both runtime and compile-time behavior.
+    The behavior is slightly different depending on the project type and target runtime.
+
+    If the property is explicitly set to TRUE:
+      - The APIs are obsolete as warning, and calls to the APIs will succeed at runtime.
+    
+    If the property is explicitly set to FALSE:
+      - On .NET 5 & 6, the APIs are obsolete as warning, and calls to the APIs will fail at runtime.
+      - On .NET 7+, the APIs are obsolete as error, and calls to the APIs will fail at runtime.
+
+    If the property is not explicitly TRUE or FALSE:
+      - On .NET 5 & 6, the APIs are obsolete as warning, and calls to the APIs will succeed at runtime.
+      - On .NET 7, the APIs are obsolete as error, but calls to the APIs will succeed at runtime.
+      - On .NET 8+, the APIs are obsolete as error, and calls to the APIs will fail at runtime
+        unless the SDK has opted in to keeping legacy BinaryFormatter behavior around.
+    
+    n.b. The APIs are already marked obsolete (as warning) in .NET 5, so we don't need to special-case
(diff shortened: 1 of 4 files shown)


baronfel:
Paging @vzarytovskii for a matching change in the FSharp SDK targets, which are stored in dotnet/fsharp.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -130,6 +130,44 @@ Copyright (c) .NET Foundation. All rights reserved.
     <IncludeProjectsNotInAssetsFileInDepsFile Condition="'$(IncludeProjectsNotInAssetsFileInDepsFile)' == ''">true</IncludeProjectsNotInAssetsFileInDepsFile>
   </PropertyGroup>
 
+  <!--
+    BinaryFormatter killbitting
```
JeremyKuhne, line 134:
```suggestion
    BinaryFormatter Disabling
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -130,6 +130,44 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    them unless we want to upgrade them to warn-as-error.
+  -->
+
+  <PropertyGroup Condition="'$(EnableUnsafeBinaryFormatterSerialization)' != 'true' AND '$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '7.0'))">
+    <!--
+      Certain project types need to re-enable BinaryFormatter by default; it will still be warn-as-error but will work at runtime.
+      WinForms: enabled through 8.0 (but not after)
+      WPF: enabled through 8.0 (but not after)
+    -->
+    <_ProjectTypeRequiresBinaryFormatter Condition="'$(UseWindowsForms)' == 'true' AND $([MSBuild]::VersionLessThanOrEquals($(TargetFrameworkVersion), '8.0'))">true</_ProjectTypeRequiresBinaryFormatter>
```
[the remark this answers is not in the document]
GrabYourPitchforks: That is correct. However, I would be surprised if unit tests break, since in my experience they tend not to exercise serialization-related code paths. If they call into BF, then they started receiving compilation errors in .NET 7, and they would have mitigated these errors by setting the compat switch either at the project level or in a higher directory's props file. That same compat switch continues to be honored with this .NET 8 change and will reenable all the killbitted code paths.

joeloff (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
jeffhandley: Good catch on this, @JeremyKuhne. @GrabYourPitchforks, I've updated the PR title and description to use "disable" verbiage.

jeffhandley (approved):

PR #31593, opened 2023-04-04
Remove reference to the VS preview flag in the developer guide

drewnoakes:
The "Use previews of the .NET SDK" feature flag no longer seems to be in VS (based on my recent internal build, Version 17.6.0 Preview 3.0 [33529.503.main]).

This removes the section in red:

image


drewnoakes:
Yes, you're right. The flag is there for GA releases.

PR #31597, opened 2023-04-04
[WIP do no review yet] Use MSBuild Server in interactive mode only

rokonec:
Addresses https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1676390

Context
It has been decided that MSBuild server will be used only in interactive mode, i.e. when output of CLI is console, and user is expecting to be in inner dev loop.
So for example in CI/CD pipelines it will not be used as it does bring advantages only in short repeat builds loops.

Changes Made
We consider it as non interactive mode when output is redirected.

Testing
Local, and unit tests...

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
@@ -16,7 +16,8 @@ namespace Microsoft.DotNet.Cli.Utils
     internal class MSBuildForwardingAppWithoutLogging
     {
         private static readonly bool AlwaysExecuteMSBuildOutOfProc = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_RUN_MSBUILD_OUTOFPROC");
-        private static readonly bool UseMSBuildServer = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_USE_MSBUILD_SERVER", false);
+        private static readonly bool UseMSBuildServerEnvVarConfig = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_USE_MSBUILD_SERVER", true)
+            && !Env.GetEnvironmentVariableAsBool("DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER", false);
 
         private const string MSBuildExeName = "MSBuild.dll";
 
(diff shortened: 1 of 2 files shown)


src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -16,7 +16,8 @@ namespace Microsoft.DotNet.Cli.Utils
     internal class MSBuildForwardingAppWithoutLogging
     {
         private static readonly bool AlwaysExecuteMSBuildOutOfProc = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_RUN_MSBUILD_OUTOFPROC");
-        private static readonly bool UseMSBuildServer = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_USE_MSBUILD_SERVER", false);
+        private static readonly bool UseMSBuildServerEnvVarConfig = Env.GetEnvironmentVariableAsBool("DOTNET_CLI_USE_MSBUILD_SERVER", true)
```
Forgind, line 19: Why does this have to be more complicated? That is, why would anyone have both DOTNET_CLI_USE_MSBUILD_SERVER set to true and DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER set to true?

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -63,6 +64,13 @@ public MSBuildForwardingAppWithoutLogging(IEnumerable<string> argsToForward, str
             }
         }
 
+        private bool ShallMSBuildServerBeUSed()
+        {
+            // We want to use MSBuild server only in inner loop development.
+            // We use Console.IsOutputRedirected as simple heuristics detection of possible non-inner-loop usage (Azure CI/CD, GitHub Actions, ...)
+            return UseMSBuildServerEnvVarConfig && !Console.IsOutputRedirected;
```
Forgind, line 71: I'm not a fan of this. If something is clearly broken, and it happens to be related to MSBuild server (not necessarily a bug in the server but perhaps some caching issue), then one fairly natural next step is to redirect the output to a file and look at that file to see what's up. Except that with this change, it might suddenly start working when you output to a file. Then, as a user, I think it was just an ephemeral bug, and I go back to building without redirecting output, and the error pops up again. I think you can see how that would be very confusing.

baronfel: This is very common for applications with an interactive mode. If you as a user want to explicitly dump the ANSI-code-aware output, then you usually have to provide an explicit flag (like `--interactive`). I spoke with @rokonec about also adding a check for this flag (which is already present in the SDK codebase) as an additional explicit signal from the user. The overall heirarchy would be

* explicit user enable via `--interactive` -> use interactive mode
* explicit user disable via `--interactive false` -> do not use interactive mode
* no explicit user signal -> infer based on env var/console output/etc

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
rokonec: Since now DOTNET_CLI_USE_MSBUILD_SERVER is true by dafault, people who want to dissable it migth use DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER . 
Unfortunately we have officially documented that the way to not use MSBuild server is through DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER. We will remove it from docs and will recommend only DOTNET_CLI_USE_MSBUILD_SERVER but Chet mentioned that had respect this env var if people set it for any reason. Maybe we can delete it in year or so, but I am afraid it will stick.

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
rokonec: @Forgind best way is to address such issues is to use /bl. However if there is such behavior it is a bug which needs to be fixed or weird/wrong custom tasks which ought to be fixed as well. So far we are not aware of any such thing, 
We hope people will report this and we will either help them to fix their buggy tasks, give them workaround, and/or fix internal bug.

PR #31626, opened 2023-04-05
Add a JsonSerializerIsReflectionEnabledByDefault property

eiriktsarpalis:
Adds a msbuild property that ties to the feature switch implemented in https://github.com/dotnet/runtime/pull/83844. ~Also disables the property when `PublishAot=true`.~

Remove ASP.NET's feature switch and use JsonSerializerIsReflectionEnabledByDefault instead.


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -20,6 +20,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</_IsTrimmingEnabled>
     <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == ''">false</_IsTrimmingEnabled>
     <DynamicCodeSupport Condition="'$(DynamicCodeSupport)' == '' And '$(PublishAot)' == 'true'">false</DynamicCodeSupport>
+    <JsonSerializerIsReflectionEnabledByDefault Condition="'$(JsonSerializerIsReflectionEnabledByDefault)' == '' And '$(PublishAot)' == 'true'">false</JsonSerializerIsReflectionEnabledByDefault>
```
jkotas, line 23: Reflection fallback is a problem for trimming in general, not just for native AOT. Is that correct?

This may rather be in Microsoft.NET.Sdk.targets. Look for `OpenGenericServiceTrimmability` for a good example.

eiriktsarpalis: Cf. https://github.com/dotnet/runtime/pull/83844#discussion_r1158727295

We think that turning it off on `PublishTrimmed=true` is both a breaking change and undesirable in certain scenaria where reflection happens to work in trimmed apps. OTOH it's a sound default to set for `PublishAot=true`

jkotas: We do not optimize for "happens to work" with the product defaults as a general principle. If a feature does not work by construction with trimming, it should be off by default with trimming.

jkotas: > both a breaking change

It would be a breaking change only if you have a working application that compiles without trimming warning that stopped working after upgrade.

If you have an application with trimming warning that happens to work and the application breaks after upgrade, it is by design. We do not document breaking changes like this - there are a ton of them in each release as we are optimizing the tools and libraries.

vitek-karas: I agree with Jan that by default this should be off. In a way it actually improves the experience. While the app might have worked on 7.0, it's very possible it won't work on 8.0 with the reflection fallback. With this off, the app will provide a consistent experience - consistently fail explaining the problem.

Note though that if we do set it to off by default, we will need to modify Blazor and MAUI to turn it back on by default - as the expectations there are exactly opposite.

MichalStrehovsky: Is this even a different "breaking change" for PublishAot vs PublishTrimmed? Both of these shipped in 7.0.

How much more broken this gets only depends on how many more optimizations we have time for in 8.0 for PublishTrimmed. Things that are generating warnings are things that are likely to be broken by new trimming optimization. It's by design. For example it's possible that we'll start more aggressively trim properties with PublishTrimmed. If we do that, the reflection fallback will get more broken than it was in 7. We would not consider it a breaking change.

eerhardt: https://github.com/dotnet/runtime/issues/84378 is opened to have the discussion about whether this switch should be defaulted to off when `PublishTrimmed=true` or not. A bunch of app model devs are tagged on that issue. It might be best to move this conversation there (I have linked this conversation there so people can read it). I think the most people will see it there.

eerhardt:
I'm marking this PR as "ready for review".

I've removed the defaulting of the feature switch until a decision can be made in https://github.com/dotnet/runtime/issues/84378.

I've removed the temporary ASP.NET JSON feature switch and am using the new one from System.Text.Json. This is needed to unblock ASP.NET work needed in preview4. So I want to get this merged without blocking on how to default the feature switch for all apps.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
eerhardt: Resolving this conversation for now. The feature switch will be defaulted in a future PR, once we can decide on the policy and how to do it.

eerhardt (approved):

jonathanpeppers (approved):
If this doesn't change the default value yet, this shouldn't impact the various MAUI workloads.

eerhardt:
I'm going to merge this by EOD today to unblock ASP.NET. Please let me know if you have any feedback.

vijayrkn (approved):

rolfbjarne (approved):

eiriktsarpalis (commented):
Can't approve my own PR, but final changes LGTM.

eerhardt:
I needed to add back the ASP.NET feature switch until we can move the ASP.NET libraries off of it. Need to do this in a staged manner so we don't regress.

PR #31628, opened 2023-04-05
[WIP] GenAPI: a work around for a generation an explicit interface implementation indexer.

andriipatsula:
This is a work around solution for the https://github.com/dotnet/sdk/issues/31570 issue to unblock the A&D team.

Indexer properties is not resolved correctly for the scenario like:
```C#
public interface IFooList<T>
{
    T this[int index] { get; set; }
}

public struct Bar<T> : IFooList<T>
{
    // indexer №1
    public T this[int index] { get; set; }

    // indexer №2
    T IFooList<T>.this[int index] { get; set; }
}
```

Indexer №1 is generated correctly, but explicit interface implementation indexer (№2) is not. I double checked that the customer passes correct reference assemblies, and it seems, it could be a bug in the Roslyn API.

@ericstj could you please review this work around. Once @ViktorHofer returns from his vacation, I will communicate the problem with him.

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
@@ -222,8 +222,9 @@ namespace Microsoft.DotNet.GenAPI
 
             foreach (ISymbol member in members.Order())
             {
+                IMethodSymbol? method = member as IMethodSymbol;
                 // If the method is ExplicitInterfaceImplementation and is derived from an interface that was filtered out, we must filter out it either.
-                if (member is IMethodSymbol method &&
+                if (method != null &&
                     method.MethodKind == MethodKind.ExplicitInterfaceImplementation &&
                     method.ExplicitInterfaceImplementations.Any(m => !_symbolFilter.Include(m.ContainingSymbol) ||
                         // if explicit interface implementation method has inaccessible type argument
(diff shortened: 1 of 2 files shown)


ericstj:
> This is a work around solution for the https://github.com/dotnet/sdk/issues/31570 issue to unblock the A&D team.

Unblocking can always happen with a partial file definition, right?

Do we know where the root cause of this bug lies?  Is it a problem in SyntaxGenerator?  If we know we should file an issue and link to it in the workaround.

Can you add a test case that covers the scenario?

andriipatsula:
@ericstj Thank you for the feedback. The problem is either in the Roslyn API or in the SBRP (Source-Build-Reference-Package) generator. I agree that the solution is not 100% clear and not in the best code style. I [wrote an algorithm](https://github.com/dotnet/sdk/issues/31570#issuecomment-1498804761) for the @MichaelSimons 's team how to build a local version of the GenAPI from the PR, prepare a nuget package and use it in the SBRP generator. I'll close this PR soon (without merging).

ericstj:
@andriipatsula I'm OK with us taking the workaround, I was just hoping for a test case and an answer to those questions so that we can understand the urgency.  I'll see if I can pick this up and fill in.

ericstj:
I couldn't get this to fail.  The following test passes for me: https://github.com/dotnet/sdk/pull/31678

ericstj:
It wasn't failing because we were testing with source symbols at this point.  If we test wit metadata symbols it works correctly.  I'll resurrect this workaround.

PR #31658, opened 2023-04-06
Revert "[main] Update dependencies from microsoft/vstest"

lewing:
Reverts dotnet/sdk#31645 to unblock https://github.com/dotnet/installer/pull/16005 until the arcade fix in https://github.com/dotnet/arcade-services/pull/2257 flows


lewing:
The arcade fix isn't working yet and runtime needs a new installer so I think we should revert while we sort out the source-build problem

lewing:
@mmitche this needs to be approved

mmitche (approved):

Issue #4277, opened 2019-12-10
Odd option "-lang" in "dotnet new --help"


am11:
## Steps to reproduce

```sh
dotnet new --help
```

## Expected  behavior

should not have `-lang` as a shorthand for `--language`, which is sufficient on its own:

```sh
.. snip ..
  --language   Filters templates based on language and specifies the language of the template to create.
.. snip ..
```

## Actual behavior

has `-lang, --language`, which is a odd one out as the convention is to either have "single character followed by a single hyphen" and/or "full name after double hyphen".

FWIW, gcc and clang use `-x` to specify language:

```sh
// C++
gcc -x c++ -O -Wall -o myapp.exe /dev/stdin <<<"int main(){return 0;}"

// or C

clang -x c -O -Wall -o myapp.exe /dev/stdin <<<"int main(void){return 0;}"
```

which `dotnet-new` can use as well.

## Environment data
`dotnet --info` output:
```sh
.NET Core SDK (reflecting any global.json):
 Version:   3.1.100
 Commit:    cd82f021f4

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.14
 OS Platform: Darwin
 RID:         osx.10.14-x64
 Base Path:   /Users/adeel/.dotnet/sdk/3.1.100/

Host (useful for support):
  Version: 3.1.0
  Commit:  65f04fb6db
```

PR #31739, opened 2023-04-12
Add `-aot` parameter to the console template

MichalStrehovsky:
This is similar to the `api` template in ASP.NET that also allows `-aot` (or `--publish-native-aot`) option (see https://github.com/dotnet/aspnetcore/pull/46064).

We set similar properties:
* `<PublishAot>true</PublishAot>` (this is the obvious one)
* `<InvariantGlobalization>true</InvariantGlobalization>` to get rid of ICU dependency on Linux and make the hello world executable ~20% smaller in general, even outside Linux.
* a commented out `<StripSymbols>true</StripSymbols>`. This is for discoverability. We default this to `false` due to platform conventions, but many users would be fine setting this to `true`. `true` is the .NET convention.

I took the string from ASP.NET. I could have also copied the localized strings but I guess it's better to not step on the toes of the localization team.

Cc @dotnet/ilc-contrib @DamianEdwards


template_feed/Microsoft.DotNet.Common.ProjectTemplates.8.0/content/ConsoleApplication-CSharp/Company.ConsoleApplication1.csproj
```diff
@@ -8,6 +8,13 @@
     <LangVersion Condition="'$(langVersion)' != ''">$(ProjectLanguageVersion)</LangVersion>
     <ImplicitUsings Condition="'$(csharpFeature_ImplicitUsings)' == 'true'">enable</ImplicitUsings>
     <Nullable Condition="'$(csharpFeature_Nullable)' == 'true'">enable</Nullable>
+    <!--#if (NativeAot) -->
+    <PublishAot>true</PublishAot>
+    <InvariantGlobalization>true</InvariantGlobalization>
+
+    <!-- Uncomment below line to make native publish outputs a lot smaller on Linux and macOS -->
+    <!-- <StripSymbols>true</StripSymbols> -->
```
MichalStrehovsky, line 16: I've struggled a bit with the level of detail here. We don't want paragraphs of text that explain native debuggers/profilers/etc may or may not support GNU debuglink and if they don't extra steps are needed to make sense of the resulting binary, etc. Most people will be fine because most tools do support it.

eerhardt: What about making an aka.ms link, and then you can point it to docs that explains it fully?

eerhardt, line 16: I think we should follow the same approach for ASP.NET and Hosting templates. We shouldn't have different strategies for the different templates if there are no technical reasons for it. We should decide here if this is the strategy we want to take, and apply it in all templates.

I also think that this just clutters the template. IMHO - I think `StripSymbols=true` should be the default without specifying it in the .csproj. It is what most normal .NET devs expect.

vitek-karas: >  I think StripSymbols=true should be the default without specifying it in the .csproj. It is what most normal .NET devs expect.

I think this is the crux of it. Lot of feedback I heard was that .NET Devs expect debuggability by default, which would mean we have to leave the symbols there. Another view is what is the typical behavior for the target platform (Linux, containers).

eerhardt: Could we base it on `Release` or not?

DamianEdwards: Setting `<StripSymbols>true</StripSymbols>` doesn't make it undebuggable thought right? The symbols will still be in a file next to the executable and discovered by the debugger. The publish target doesn't exclude the symbols file (at least in my testing) so the developer would need to manually add logic to exclude the symbols file when deploying and hence make the app undebuggable. Having the files be separate makes it clear where the app size is coming from IMO. Certainly now, during previews, is the time to try out ideas like @eerhardt's and assess feedback.

DamianEdwards: Are .NET symbols generally larger than other native stacks, e.g. C++, Go?

agocke: Somewhat, because we tend to have very long names. This could probably be significantly improved, but we have spent zero time on it.

jkotas: In addition to the typical name lengths, another factor is complexity of types used by your program.

Here are a few numbers for a few .NET runtime binaries that are written in C++:  
- libcoreclr.so - 7,054,328
- libcoreclr.so.dbg - 96,091,944 (symbols size is 13x binary size)
- libclrjit.so - 3,416,848
- libclrjit.so.dbg - 27,781,352 (symbol size is 8x binary size)

`dotnet new api -aot` (NET 8 P3):
- todo - 12,361,544
- todo.dbg - 37,794,328 (symbol size is 3x binary size)

You can see that we are not doing that bad compared to C++.

.NET runtime uses some C++ templates, but it is not a modern C++. I would expect that the multiplier for a typical modern C++ program is going to be even higher than the multiplier for libcoreclr.so. C++ STL types tend to be very complicated, and thus have very long names.

golang had no generics until very recently. It means you could not create complicated types in it up until very recently. The naming convention in golang tends to be very terse. Both these factors together mean that the multiplier for golang symbols is expected to be on small side, smaller than what we get.

jkotas: Some more insights:
- golang compresses dwarf symbols by default (see https://github.com/golang/go/issues/26318). It makes their relatively small symbols even smaller. Dwarf symbols - especially the ones with long names - are highly compressible. The compression ratio is 6x for `dotnet new api -aot` template. The downside is that it increases build time and slows down symbol loading in debugger.
- We have prior art with managed (portable) PDBs. They are separate files by default, but one can opt-in into embedding them via `<DebugType>embedded</DebugType>`. The managed embedded portable PDBs are compressed by default as well (it adds to the build time).

I have been mulling over this over last several days. I am leaning towards changing the SDK default to `<StripSymbols>true</StripSymbols>` (ie follow .NET / Windows convention even though it differs from Unix conventions). The prior art with portable PDBs is what made me to flip.

We should consider exposing options for symbol compression and default to compressed symbols, either all the time or at least when the symbols are not stripped.

jkotas (approved):
LGTM.

If we choose to change the StripSymbols defaults, it can be done as a follow up across the board.

template_feed/Microsoft.DotNet.Common.ProjectTemplates.8.0/content/ConsoleApplication-CSharp/Company.ConsoleApplication1.csproj
MichalStrehovsky: If we decide to default `StripSymbols` to true on the SDK side, I have some additional questions:

* .NET developers are used to symbols having the .pdb extension, no matter their underlying representation (managed/portable are very different file formats). Should we also name the native symbol file with the Windows (and by extension .NET) convention? If we don't want people to learn the platform conventions, this follows the principle of least surprise.
* Is there any point in the AOT console template? My main motivation for it was really just about the discoverability of `StripSymbols`. `<PublishAot>true</PublishAot>` doesn't seem to be interesting enough as a differentiator. I've also put `<InvariantGlobalization>true</InvariantGlobalization>` because it gets rid of ICU on Linux and helps discoverability, but that one is basically just getting a ride. Are the things that are left still important enough that we need a template?

DamianEdwards: I think it's still worth adding the option to the template if for nothing else that visibility and discovery that native AOT is now a first-class option.

jkotas: > Should we also name the native symbol file with the Windows (and by extension .NET) convention? 

This would not work well even if we wanted to. Many tools infer the file format from the extension. If we started giving .pdb extension to DWARF files, it would be endless source of breaks.

MichalStrehovsky: We could maybe go with .pdb.dbg/.pdb.dwarf. It's just whether we want people to know these are the debugging symbols, not a vital part of a non-single-file app. We assume .NET developers don't know about the underlying platform. But I guess this would be orthogonal to this PR and at this point it would also be a "breaking" change from 7.0 (so will be the new default though).

@dotnet/templating-engine-maintainers could someone have a look? I think we settled on the non-technical questions and we might tweak this based on the defaults in the coming weeks.

jkotas: > We could maybe go with .pdb.dbg/.pdb.dwarf. It's just whether we want people to know these are the debugging symbols, not a vital part of a non-single-file app.

I do not think we need to go this far. People will need to learn some amount of platform specific details in any case. The problem is that we do not have a good trigger for learning about symbols today. People do not read the manual; they just try to do stuff. We have seen many variants of a story that goes like this: Somebody builds a native AOT binary on Linux for the first time, sees a single giant binary in the publish folder, and concludes that this technology is unusable. I think that having two files - one smaller executable and one large non-executable - would be good enough to trigger the learning even without the .pdb hint. Seeing two files will make people ask why there are two files, whether they need both, etc.

vlada-shubina (commented):
Please consider adding tests to https://github.com/dotnet/sdk/blob/main/src/Tests/dotnet-new.Tests/CommonTemplatesTests.cs

MichalStrehovsky:
@baronfel do you have any opinions? I'd like this to make it into Preview 4 and it's getting close.

baronfel:
LGTM - wanted to make sure the parameters aligned with ASP.NET's template (of course they do).

Any reason why the F# console template can't have this as well?

MichalStrehovsky:
> LGTM - wanted to make sure the parameters aligned with ASP.NET's template (of course they do).
> 
> Any reason why the F# console template can't have this as well?

https://github.com/dotnet/sdk/pull/31739#issuecomment-1504937521 :)

baronfel:
Ah, thanks :) IIRC @KevinRansom is doing some more trimming/aot work right now that might make that disappear. @vzarytovskii want to make a note to check the experience after that lands and update the F# console template with this option as well?

baronfel (approved):
LGTM!

am11:
Thanks @MichalStrehovsky! It would be nice(r) if we use single or double hyphens consistently with `dotnet new`. `-aot` is the odd one out: `dotnet new console --name foo --framework net8.0 -aot` as Unix-y tools [arguments convention](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) suggest: `-aot` == `-a -o -t`. We used to have one such anomaly: https://github.com/dotnet/sdk/issues/4277, now we have two. 🥲

MichalStrehovsky:
> Thanks @MichalStrehovsky! It would be nice(r) if we use single or double hyphens consistently with `dotnet new`. `-aot` is the odd one out: `dotnet new console --name foo --framework net8.0 -aot` as Unix-y tools [arguments convention](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) suggest: `-aot` == `-a -o -t`. We used to have one such anomaly: #4277, now we have two. 🥲

@DamianEdwards what do you think? (For this template I just aligned with https://github.com/dotnet/aspnetcore/pull/46064 and there wasn't much to think about because the precedent was set.)

DamianEdwards:
I'm happy to say from now on we should strive to match the Unix-y tool convention and change the `-aot` alias to be `--aot` on all the templates.

DamianEdwards:
Should we consider ret-conning this across all templates, by adding `--` aliases for any non-single char `-` prefixed parameters in 8.0 and even 7.0.xxx, and consider a deprecation plan (warnings, etc.) for the now non-conforming `-` prefixed parameters?

eerhardt:
> I've not done anything for F# since even a hello world produces both trimming and AOT warnings and it's not the experience we're shooting for with PublishAot.

@MichalStrehovsky - I am able to successfully publish and run an F# "worker" app with AOT with no warnings:

image

What warnings were you getting with a hello world F# app?

eerhardt:
Oh, I just tried it manually, and it appears to be coming from `printfn "Hello from F#"`. It actually completely breaks ILC:

```
EXEC : error : Failed to load type '!0' from assembly 'FSharp.Core, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' [C:\DotNetTest\Net8FSharp\Net8FS
harp.fsproj]
  Internal.TypeSystem.TypeSystemException+TypeLoadException: Failed to load type '!0' from assembly 'FSharp.Core, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f
  5f7f11d50a3a'
     at Internal.TypeSystem.ThrowHelper.ThrowTypeLoadException(ExceptionStringID, String, String) + 0x37
     at System.Reflection.TypeNameParser.GetType(String, ReadOnlySpan`1, String) + 0x15c
     at System.Reflection.TypeNameParser.NamespaceTypeName.ResolveType(TypeNameParser&, String) + 0x3e
     at System.Reflection.TypeNameParser.GenericTypeName.ResolveType(TypeNameParser&, String) + 0x73
     at Internal.TypeSystem.Ecma.CustomAttributeTypeProvider.GetTypeFromSerializedName(String) + 0x71
     at System.Reflection.Metadata.Ecma335.CustomAttributeDecoder`1.DecodeArgument(BlobReader&, CustomAttributeDecoder`1.ArgumentTypeInfo) + 0xb4
     at System.Reflection.Metadata.Ecma335.CustomAttributeDecoder`1.DecodeFixedArguments(BlobReader&, BlobReader&, Int32, BlobReader) + 0x11e
     at System.Reflection.Metadata.Ecma335.CustomAttributeDecoder`1.DecodeValue(EntityHandle, BlobHandle) + 0x39c
     at Internal.TypeSystem.Ecma.MetadataExtensions.<GetDecodedCustomAttributes>d__3.MoveNext() + 0x21a
     at ILCompiler.DependencyAnalysis.DynamicDependencyAttributesOnEntityNode.GetStaticDependencies(NodeFactory) + 0xbf
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependenciesImpl(DependencyNodeCore`1) + 0x44
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependencies(DependencyNodeCore`1) + 0x32
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ProcessMarkStack() + 0xb5
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ComputeMarkedNodes() + 0x4e
     at ILCompiler.ILScanner.ILCompiler.IILScanner.Scan() + 0x14
     at ILCompiler.Program.<Run>g__RunScanner|3_0(Program.<>c__DisplayClass3_0&) + 0x108
     at ILCompiler.Program.Run() + 0x153f
     at ILCompiler.ILCompilerRootCommand.<>c__DisplayClass206_0.<.ctor>b__0(InvocationContext context) + 0x20c
```

I'll remove the F# `-aot` parameter to be consistent everywhere until F# fully supports AOT.

MichalStrehovsky:
> It actually completely breaks ILC:

Heh, F# bug: https://github.com/dotnet/fsharp/issues/15140. Nevertheless, we should not crash. This should be in a try/catch and we should ignore the invalid blobs here.

PR #31743, opened 2023-04-12
Add details command

YuliiaKovalova:
Implements the item: https://github.com/dotnet/templating/issues/6683

Add details command to display more information about a template package.


YuliiaKovalova:
@vlada-shubina, please take a look when you come back.

baronfel:
@KalleOlaviNiemitalo that's actually a great point - since NuGet.org is our source of truth here, we should use licenses.nuget.org instead of spdx. Have to bias towards consistency here.

baronfel:
Here is my sketched up feedback for the display output - note the # are comments on each line of output:

```shell
PS C:\ykovalova_sdk\yk_fork_8\sdk> dotnet new details NUnit3.DotNetNew.Template --version "1.0.0"
   NUnit3.DotNetNew.Template
      Details:  Project and item templates containing basic NUnit 3 Test Project.
      Authors:  Aleksei Kharlov
      Owners:  akharlov, rprouse # each of these should link to the owners page on nuget.org (e.g. https://www.nuget.org/profiles/akharlov)
      License Metadata: # depending on how the license is provided in the package, this should change
         License Url:  https://github.com/halex2005/dotnet-new-nunit/blob/master/LICENSE # clickable link
         License Expression: MIT # link to licenses.nuget.org/MIT per https://learn.microsoft.com/en-us/nuget/nuget-org/licenses.nuget.org#request
      Repository Url:  https://github.com/halex2005/dotnet-new-nunit # clickable link
      Templates:
         NUnit 3 Test Project
            Short Names:  nunit
            Author:  Aleksei Kharlov aka halex2005 (codeofclimber.ru)
            Tags:  Test/NUnit # tags should be split on '/' and each tag should be a clickable link to the NuGet.org query for that tag: https://www.nuget.org/packages?q=Tags%3A%22Test%22
            Languages:  C#, F#, VB # group the languages available for each template
            Details:  A project that contains NUnit tests that can run on .NET Core on Windows, Linux and macOS
         NUnit 3 Test Item
            Short Names:  nunit-test
            Author:  Aleksei Kharlov aka halex2005 (codeofclimber.ru)
            Tags: Test/NUnit
            Languages:  C#, F#, VB
            Details:  A item that contains NUnit tests
```

baronfel:
@KalleOlaviNiemitalo yes, all feedback about clickable links is predicated on the terminal supporting control codes to signal that. This is a muscle that the CLI and MSBuild are learning this release cycle and we're fairly confident it in - though of course there should always be an explicit opt out.

baronfel:
Might just have to be a file:// URI to the nuspec and let it open with whatever file handler the users system has configured for nuspec files.

maridematte:
The current display of details looks like this:

image

I added the table to not feel like a wall of text.

maridematte:
This is the text from the above image:

``` 
NUnit3.DotNetNew.Template
   Details:  Project and item templates containing basic NUnit 3 Test Project.
   Authors:
      Aleksei Kharlov
   Owners:
      https://nuget.org/profiles/akharlov
      https://nuget.org/profiles/rprouse
   License Metadata:
      License Url: https://github.com/halex2005/dotnet-new-nunit/blob/master/LICENSE
      Repository Url: https://github.com/halex2005/dotnet-new-nunit

   Templates:
      Template Name         Short Name  Type     Tags        Language
      --------------------  ----------  -------  ----------  --------
      NUnit 3 Test Project  nunit       project  Test/NUnit  C#,F#,VB
      NUnit 3 Test Item     nunit-test  item     Test/NUnit  C#,F#,VB 
```

src/Cli/Microsoft.TemplateEngine.Cli/TemplatePackageCoordinator.cs
```diff
@@ -368,6 +423,45 @@ private static void InitializeNuGetCredentialService(bool interactive)
...
+            var list = new List<string> { $"{packageIdentity}".Indent(1) };
+
+            AddIfNotNull(list, SymbolStrings.Command_Details_Description_Property, templatePackage?.Description, 2);
+            AddIfNotNull(list, SymbolStrings.Command_Details_Authors_Property, templatePackage?.Authors, 2);
+            AddIfNotNull(list, SymbolStrings.Command_Details_Owners_Property, templatePackage?.Owners, 2);
+            list.Add($"{SymbolStrings.Command_Details_LicenseMetadata_Property}:".Indent(2));
+            AddIfNotNull(list, SymbolStrings.Command_Details_License_Property, templatePackage?.License, 3);
+            AddIfNotNull(list, SymbolStrings.Command_Details_License_Expression_Property, templatePackage?.LicenseExpression, 3);
+            AddIfNotNull(list, SymbolStrings.Command_Details_License_Url_Property, templatePackage?.LicenseUrl?.ToString(), 3);
+            AddIfNotNull(list, SymbolStrings.Command_Details_Repository_Url_Property, templatePackage?.ProjectUrl?.ToString(), 2);
```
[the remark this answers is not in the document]
maridematte: It does not. I think as of right now the commit Id is out of scope as it would need more changes to what we get from the Nuget API.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/SymbolStrings.resx
```diff
@@ -126,6 +126,59 @@
   <data name="Command_Alias_Description" xml:space="preserve">
     <value>Creates or displays defined aliases.</value>
   </data>
+  <data name="Command_Details_Description" xml:space="preserve">
+    <value>
+      Provides details for specified Nuget package.
+      The command checks if the package is installed locally, if it wasn't found, then it searches on the https://www.nuget.org/.</value>
```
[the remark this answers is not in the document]
baronfel: @maridematte/ @YuliiaKovalova for details we should be able to use any configured NuGet source, right? This doesn't require non-NuGet-API metadata like `search` does.

Separately, @KalleOlaviNiemitalo I'm not sure if the NuGet API exposes PSM data or if it's all implementation details. @YuliiaKovalova can you check with that team?

maridematte:
The urls are clickable currently, if you're using windows terminal you can press `ctrl + left click` to navigate to the url.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/SymbolStrings.resx
```diff
@@ -126,6 +126,59 @@
   <data name="Command_Alias_Description" xml:space="preserve">
     <value>Creates or displays defined aliases.</value>
   </data>
+  <data name="Command_Details_Description" xml:space="preserve">
+    <value>
+      Provides details for specified Nuget template package.
+      The command checks if the package is installed locally, if it wasn't found, then it searches on the https://www.nuget.org/.</value>
```
vlada-shubina, line 132:
```suggestion
      Provides the details for specified template package.
      The command checks if the package is installed locally, if it wasn't found, then it searches on the https://www.nuget.org/.</value>
```

src/Cli/Microsoft.TemplateEngine.Cli/Commands/SymbolStrings.resx
```diff
@@ -126,6 +126,59 @@
...
+  </data>
+  <data name="Command_Details_Argument_PackageIdentifier" xml:space="preserve">
+    <value>Nuget package identifier</value>
```
vlada-shubina, line 135:
```suggestion
    <value>NuGet package identifier</value>
```

should it work also for non-NuGet packages?

src/Cli/Microsoft.TemplateEngine.Cli/Commands/SymbolStrings.resx
```diff
@@ -126,6 +126,59 @@
   <data name="Command_Alias_Description" xml:space="preserve">
     <value>Creates or displays defined aliases.</value>
   </data>
+  <data name="Command_Details_Description" xml:space="preserve">
```
vlada-shubina, line 129: This file is intended only for command symbols localizations. All other strings should be moved to `LocalizableStrings`.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/BaseDetailsCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Cli.NuGet;
+using Microsoft.TemplateEngine.Edge.Settings;
+
+namespace Microsoft.TemplateEngine.Cli.Commands
+{
+    internal class BaseDetailsCommand : BaseCommand<DetailsCommandArgs>
```
vlada-shubina, line 12: Base command is not needed in this case. Base commands were used only when we have 2 different commands doing exactly the same.
```suggestion
    internal class DetailsCommand : BaseCommand<DetailsCommandArgs>
```

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/BaseDetailsCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+            Arity = new ArgumentArity(0, 1)
+        };
+
+        protected async override Task<NewCommandStatus> ExecuteAsync(
+            DetailsCommandArgs args,
+            IEngineEnvironmentSettings environmentSettings,
+            TemplatePackageManager templatePackageManager,
+            InvocationContext context)
+        {
+            var packageIdentity = args.NameCriteria ?? string.Empty;
```
vlada-shubina, line 43: `NameCriteria` cannot be null, empty or whitespace. It's better to handle it via argument validator.

src/Cli/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -182,6 +182,13 @@
   <data name="Generic_Info_NoMatchingTemplates" xml:space="preserve">
     <value>No templates found matching: {0}.</value>
   </data>
+  <data name="Generic_Info_NoMatchingTemplatePackage" xml:space="preserve">
+    <value>
+      No template package found matching: {0}.
+      
+      Make sure the specified package exists on the https://www.nuget.org/.
+    </value>
+  </data>
```
vlada-shubina, line 191: I would suggest:
```
No template packages found matching: 'blabla'.

To list installed template packages, run:
   dotnet new uninstall
To search for the template packages on NuGet.org, run:
   dotnet new search --package blabla


```

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/BaseDetailsCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        private static NugetApiManager _nugetApiManager = new NugetApiManager();
+
+        internal BaseDetailsCommand(
+            Func<ParseResult, ITemplateEngineHost> hostBuilder,
+            string commandName)
+            : base(hostBuilder, commandName, SymbolStrings.Command_Details_Description)
+        {
+            AddArgument(NameArgument);
+            AddOption(VersionOption);
```
vlada-shubina, line 22: should we support
- getting details from the feed other than NuGet.org
- interactive auth

@baronfel what do you think?

This would make the command consistent with other usages.

src/Cli/Microsoft.TemplateEngine.Cli/TemplatePackageCoordinator.cs
```diff
@@ -332,6 +379,87 @@ internal async Task<NewCommandStatus> EnterUninstallFlowAsync(UninstallCommandAr
...
+                (localPackage, packageTemplates) = await _templatePackageManager
+                    .GetManagedTemplatePackageAsync(packageIdentity, packageVersion, cancellationToken).ConfigureAwait(false);
+            }
+            catch
+            {
+                Reporter.Output.WriteLine(
+                LocalizableStrings.Generic_Info_NoMatchingTemplatePackage.Bold().Red(),
+                $"{packageIdentity}{(string.IsNullOrWhiteSpace(packageVersion) ? string.Empty : $"::{packageVersion}")}");
+
+                return NewCommandStatus.NotFound;
```
vlada-shubina, line 406: should we try to search instead in this case?

vlada-shubina (commented):
Consider adding unit and integration tests

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/BaseDetailsCommand.cs
baronfel: yes - details, unlike search, should be usable for any package from any feed. It should also accept the interactive flag to allow for authentication.

Apologies for the delay here!

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/DetailsCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+        internal static Argument<string> NameArgument { get; } = new("package-identifier")
+        {
+            Description = LocalizableStrings.DetailsCommand_Argument_PackageIdentifier,
+            Arity = new ArgumentArity(1, 1)
+        };
+
+        internal static Option<string> VersionOption { get; } = new Option<string>(new string[] { "-version", "--version" })
+        {
+            Description = LocalizableStrings.DetailsCommand_Option_Version,
+            Arity = new ArgumentArity(0, 1)
```
YuliiaKovalova, line 35: @vlada-shubina , does this property represent the min/max number of params?

src/Cli/Microsoft.TemplateEngine.Cli/Commands/xlf/SymbolStrings.es.xlf
```diff
@@ -17,6 +17,15 @@
         <target state="translated">Crea o muestra alias definidos.</target>
         <note />
       </trans-unit>
+      <trans-unit id="Command_Details_Description">
+        <source>
+      Provides the details for specified template package.
+      The command checks if the package is installed locally, if it wasn't found, then it searches on the https://www.nuget.org/.</source>
```
YuliiaKovalova, line 23: please change wasn't -> was not

It is a localization requirement

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/DetailsCommand.cs
vlada-shubina: it represents min-max amount of accepted args for the options. For this option it has to be 1,1

src/Cli/Microsoft.TemplateEngine.Cli/Commands/details/DetailsCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+            InvocationContext context)
+        {
+            var templatePackageCoordinator = new TemplatePackageCoordinator(environmentSettings, templatePackageManager);
+
+            NewCommandStatus status = await templatePackageCoordinator.DisplayTemplatePackageMetadata(
+                args,
+                _nugetApiManager,
+                context.GetCancellationToken()).ConfigureAwait(false);
+
+            await CheckTemplatesWithSubCommandName(args, templatePackageManager, context.GetCancellationToken()).ConfigureAwait(false);
```
vlada-shubina, line 55: this might be done in parallel

src/Cli/Microsoft.TemplateEngine.Cli/TemplatePackageCoordinator.cs
```diff
@@ -556,5 +808,367 @@ private async Task<bool> IsTemplateShortNameAsync(string sourceIdentifier, Cance
...
+            NugetApiManager apiManager,
+            string packageIdentifier,
+            string? packageVersion = null,
+            CancellationToken cancellationToken = default)
+        {
+            NugetPackageMetadata? highestVersion = null;
+
+            foreach (PackageSource source in sources)
+            {
+                NugetPackageMetadata? package = await apiManager.GetPackageMetadataAsync(packageIdentifier, packageVersion, source, cancellationToken).ConfigureAwait(false);
```
vlada-shubina, line 1156: potentially this can be done in parallel too, and then will be faster. See the implementation in dotnet/templating:
https://github.com/dotnet/templating/blob/dec5d8684547b8b3347def807d4545de9d309ea0/src/Microsoft.TemplateEngine.Edge/Installers/NuGet/NugetApiPackageManager.cs#L248-L310

YuliiaKovalova:
Looks good to me!

src/Tests/dotnet-new.Tests/DotnetNewDetailsTest.cs
```diff
@@ -21,27 +21,28 @@ public DotnetNewDetailsTest(DiagnosticFixture diagnosisFixture, ITestOutputHelpe
         [Fact]
         public void CanDisplayDetails_RemotePackage()
         {
-            new DotnetNewCommand(_log, "details", "Microsoft.Azure.WebJobs.ItemTemplates")
+            new DotnetNewCommand(_log, "details", "Microsoft.DotNet.Web.ProjectTemplates.5.0")
```
vlada-shubina, line 24: I would recommend to use approval tests here for better verification

YuliiaKovalova:
@maridematte, I would recommend adding in output 3 more fields: Version, Source Feed, and Prefix Reserved flag.

vlada-shubina:
> The current display of details looks like this:
> 
> image
> I added the table to not feel like a wall of text.

this output misses:
- version of the package
- NuGet feed the information is taken from
- indication if prefix is reserved

vlada-shubina:
Tests: i would recommend to cover at least the following scenarios
- the package is not installed
   - the package is from NuGet.org (with / without version, with templates information)
   - the package is from other feed (with / without version, without package information0
- the package is installed
   - the package is from NuGet.org
   - the package is from other feed 
   - the package is installed from local *.nupkg
   - the package is the folder

src/Cli/Microsoft.TemplateEngine.Cli/TemplatePackageCoordinator.cs
```diff
@@ -332,6 +381,198 @@ internal async Task<NewCommandStatus> EnterUninstallFlowAsync(UninstallCommandAr
...
+            return NewCommandStatus.NotFound;
+        }
+
+        internal void DisplayNuGetPackageMetadata(NugetPackageMetadata packageMetadata, IReporter reporter)
+        {
+            reporter.WriteLine($"{packageMetadata.Identity.Id}");
+            WriteIfNotNull(LocalizableStrings.DetailsCommand_Property_Version, packageMetadata.PackageVersion.ToString(), reporter, 1);
+
+            WriteIfNotNull(LocalizableStrings.DetailsCommand_Property_Description, packageMetadata.Description, reporter, 1);
+            WriteIfNotNull(LocalizableStrings.DetailsCommand_Property_SourceFeed, packageMetadata.Source.ToString(), reporter, 1);
```
baronfel, line 477: I don't think we should rely on the `ToString()` member of PackageSource here - it could change. We should be explicit and write the source in the following fashion:

* if source.Name == source.Source, only write the source.Source
* otherwise, write `{source.Name} [{source.Source}]`

This reduces duplication, and insulates us from tostring changes coming from NuGet APIs.

src/Cli/Microsoft.TemplateEngine.Cli/TemplatePackageCoordinator.cs
```diff
@@ -556,5 +805,371 @@ private async Task<bool> IsTemplateShortNameAsync(string sourceIdentifier, Cance
...
+                currentDirectory = Directory.GetCurrentDirectory();
+                ISettings settings = global::NuGet.Configuration.Settings.LoadDefaultSettings(currentDirectory);
+                PackageSourceProvider packageSourceProvider = new PackageSourceProvider(settings);
+                defaultSources = packageSourceProvider.LoadPackageSources().Where(source => source.IsEnabled);
+                if (includeNuGetFeed)
+                {
+                    var nuGetFeed = new PackageSource(NugetOrgFeed);
+                    defaultSources = defaultSources.Append(nuGetFeed);
+                }
+            }
```
baronfel, line 1098: Future expansion: I think if `string packageIdentity` was added to this method, then we could pretty easily add support for PackageSourceMapping to this feature via something like this:

```csharp
PackageSourceMapping packageSourceMapping = PackageSourceMapping.GetPackageSourceMapping(settings);
IReadOnlyList<string> validPackageSources = packageSourceMapping.GetConfiguredPackageSources(packageIdentity);
defaultSources = packageSourceProvider.LoadPackageSources().Where(source => source.IsEnabled && (validPackageSources?.Contains(source.Name) ?? true));
```

We should consider PSM support all-up for templating, though.

baronfel (approved):
There are a few notes about localization of user-facing strings and a couple layout notes that we spoke about offline as well. Overall I like the state of this a lot!

YuliiaKovalova:
[celebrate]     Yuliia Kovalova reacted to your message:
________________________________

PR #31766, opened 2023-04-12
Add IsAotCompatible flag to enable all analyzers and IsTrimmable

### LINKED ISSUE #31284: EnableAotAnalyzer=true should also default EnableSingleFileAnalyzer=true

agocke:
Fixes https://github.com/dotnet/sdk/issues/31284


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -524,6 +524,42 @@ public void Only_Aot_warnings_are_produced_if_EnableAotAnalyzer_is_set(string ta
...
+            // The below relies on the build above being cached so that no warnings are produced
+            // If this fails, it's likely because the build above was not cached
+            var publishCommand = new PublishCommand(Log, Path.Combine(testAsset.TestRoot, testProject.Name));
+            publishCommand
+                .Execute(RuntimeIdentifier)
+                .Should().Pass()
+                .And.NotHaveStdOutContaining("warning IL3050")
+                .And.NotHaveStdOutContaining("warning IL3056")
+                .And.NotHaveStdOutContaining("warning IL2026")
+                .And.NotHaveStdOutContaining("warning IL3002");
```
eerhardt, line 560: IMO - this is kind of weird behavior to test. Is there a reason we are verifying this in the test?

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
```diff
@@ -524,6 +524,42 @@ public void Only_Aot_warnings_are_produced_if_EnableAotAnalyzer_is_set(string ta
...
+                .And.HaveStdOutContaining("warning IL3050")
+                .And.HaveStdOutContaining("warning IL3056")
+                .And.HaveStdOutContaining("warning IL2026")
+                .And.HaveStdOutContaining("warning IL3002");
+
+            var outputDirectory = buildCommand.GetOutputDirectory(targetFramework).FullName;
+            var assemblyPath = Path.Combine(outputDirectory, $"{projectName}.dll");
+
+            // injects the IsTrimmable attribute
+            AssemblyInfo.Get(assemblyPath)["AssemblyMetadataAttribute"].Should().Be("IsTrimmable:True");
```
eerhardt, line 549: I'm not sure why, but this is failing in CI:

```
   Microsoft.NET.Publish.Tests.GivenThatWeWantToPublishAnAotApp.IsAotCompatible_implies_enable_analyzers(targetFramework: "net8.0") [FAIL]
      System.IO.FileNotFoundException : Could not find file 'C:\h\w\A4E408F0\t\dotnetSdkTests\v4xmjh5h.pae\IsAotCompatib---48D17CD7\WarningAppWithAotAnalyzer\bin\Debug\net8.0\WarningAppWithAotAnalyzer.dll'.
      Stack Trace:
           at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
           at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
           at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
           at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
           at System.IO.File.OpenRead(String path)
        /_/src/Tests/Microsoft.NET.TestFramework/AssemblyInfo.cs(57,0): at Microsoft.NET.TestFramework.AssemblyInfo.Get(String assemblyPath)
        /_/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs(549,0): at Microsoft.NET.Publish.Tests.GivenThatWeWantToPublishAnAotApp.IsAotCompatible_implies_enable_analyzers(String targetFramework)
           at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
           at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
      Output:
        > C:\h\w\A4E408F0\p\d\dotnet.exe msbuild /t:Build C:\h\w\A4E408F0\t\dotnetSdkTests\v4xmjh5h.pae\IsAotCompatib---48D17CD7\WarningAppWithAotAnalyzer\WarningAppWithAotAnalyzer.csproj /restore /p:RuntimeIdentifier=win10-x64
```

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
agocke: I asked myself -- what's the difference between `IsAotCompatible` and `PublishAot`? And came up with this. Seems like we should have a test to verify that they are semantically distinct.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
agocke: Yeah, trying to debug. Not having great success with running the test locally.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAnAotApp.cs
eerhardt: > what's the difference between IsAotCompatible and PublishAot? And came up with -- IsAotCompatible just does analysis, it doesn't actually enable trimming.

Maybe it would make more sense to verify this another way then? Using the "does it emit warnings?" as a verification for "did it publish for AOT?" doesn't seem like a direct way to verify it - especially if that rule can fail, like mentioned in the comments.

Is there a better way to check if the app was published for AOT?

agocke: I think checking the warnings is fine. That's also how we verify trimming for the linker. I've changed the code to more reliably detect the warnings, so I'm not worried about caching affecting things now.

sbomer (approved):
LGTM, thank you!

PR #31860, opened 2023-04-18
Disambiguate GenAPI calls to base constructors

### LINKED ISSUE #31655: GenAPI can generate ambiguous call to base constructor
When GenAPI defines arguments to a base constructor, it uses `default!` but this can lead to an ambiguous call when the base type has constructor overloads with the same number of parameters but different types.

An example of this is the generated code for NuGet.Protocol.6.5.0:

```c#
public partial class SourcePackageDependencyInfo : Packaging.Core.PackageDependencyInfo
{
    public SourcePackageDependencyInfo(Packaging.Core.PackageIdentity identity, System.Collections.Generic.IEnumerable<Packaging.Core.PackageDependency> dependencies, bool listed, SourceRepository source, System.Uri downloadUri, string packageHash) : base(default!, default!) { }
```

This leads to the following compiler error:

```
error CS0121: The call is ambiguous between the following methods or properties: 'PackageDependencyInfo.PackageDependencyInfo(PackageIdentity, IEnumerable<PackageDependency>)' and 'PackageDependencyInfo.PackageDependencyInfo(string, NuGetVersion)
```

This can be solved by explicitly casting `default` to the desired type:

```c#
: base((Packaging.Core.PackageIdentity)default!, (System.Collections.Generic.IEnumerable<Packaging.Core.PackageDependency>)default!) { }
```

cc @andriipatsula @ViktorHofer

ericstj:
Fix https://github.com/dotnet/sdk/issues/31655

### DIFF
--- a/src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
+++ b/src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
@@ -1321,6 +1321,53 @@ namespace Microsoft.DotNet.GenAPI.Tests
                     """);
         }
 
+        [Fact(Skip = "https://github.com/dotnet/roslyn/issues/67867")]
+        public void TestBaseTypeWithAmbiguousNonDefaultConstructorsRegression31655()
+        {
+            RunTest(original: """
+                    namespace Foo
+                    {
+                        public class A
+                        {
+                            public A(Id id, System.Collections.Generic.IEnumerable<D> deps) { }
+                            public A(string s, V v) { }
+                        }
+
+                        public class B : A
+                        {
+                            public B() : base(new Id(), new D[0]) {}
+                        }
+
+                        public class Id { }
+
+                        public class D { }
+                    
+                        public class V { }
+                    }
+                    """,
+                expected: """
+                    namespace Foo
+                    {
+                        public partial class A
+                        {
+                            public A(Id id, System.Collections.Generic.IEnumerable<D> deps) { }
+                            public A(string s, V v) { }
+                        }
+
+                        public partial class B : A
+                        {
+                            public B() : base(default(Id), default(System.Collections.Generic.IEnumerable<D>)) {}
+                        }
+
+                        public partial class Id { }
+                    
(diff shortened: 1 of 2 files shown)


ViktorHofer (approved):
Lovely. When we wrote this, we believed that a default without specifying the underlying type is sufficient. We were wrong.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
```diff
@@ -1284,6 +1284,43 @@ public partial class C : B
...
+                        }
+
+                        public partial class B : A
+                        {
+                            public B() : base(default(char)) {}
+                        }
+                    }
+                    """);
+        }
+
```
carlossanlop, line 1323: @ericstj do you mind also adding test cases for the examples provided in the issue? https://github.com/dotnet/sdk/issues/31655

One default is casted to a class type, the other one to a generic type:

```cs
base((Packaging.Core.PackageIdentity)default!, (System.Collections.Generic.IEnumerable<Packaging.Core.PackageDependency>)default!)
```

Want to make sure both work as expected.

ericstj: Sure I can, though I don't think we need to use those NuGet types.  I'll add some user types.

ericstj: Interesting, so I'm adding the type arguments to default correctly, but it's being removed by https://github.com/dotnet/sdk/blob/4ea3e8304c21022bbea72a29154b7602af4f9239/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs#L94

This seems to be a bug in the Reducer in roslyn.

ericstj: Filed https://github.com/dotnet/roslyn/issues/67867 and I will add the regression test with a suppression.

ViktorHofer (approved):

PR #31863, opened 2023-04-18
Fix incorrect MicrosoftSourceLinkBitbucketGitVersion property name

MichaelSimons:
https://github.com/dotnet/sdk/pull/31632 introduced the wrong property name - it is missing the `Git` part.  This is causing a prebuilt in source-build.  Source-build is trying to override this property with the live version but because the property is named incorrectly, it is not overridden.


tmat (approved):
Thanks for fixing.

mmitche (approved):

PR #31868, opened 2023-04-18
Signing:  enable NuGet signature verification by default on Linux

### LINKED ISSUE #11262: Signing:  re-enable signed package verification on Linux and macOS
[Design specification](https://github.com/dotnet/designs/blob/main/accepted/2021/signed-package-verification/re-enable-signed-package-verification.md)
[Technical specification](https://github.com/dotnet/designs/blob/main/accepted/2021/signed-package-verification/re-enable-signed-package-verification-technical.md)

dtivel:
_Note:  wait for NuGet's automated test results for .NET 8 SDK before merging this PR._

Resolves https://github.com/NuGet/Home/issues/11262

This change enables NuGet signed package verification by default on Linux in .NET 8 SDK.  Because [NuGet.Client](https://github.com/NuGet/NuGet.Client) dual-inserts into both .NET 7 and 8 SDK's, NuGet needed a way to enable verification by default only in .NET 8 SDK not .NET 7 SDK.  Since .NET SDK's `main` branch is .NET 8, this change does that by setting NuGet's environment variable `DOTNET_NUGET_SIGNATURE_VERIFICATION` to `true` in all NuGet code paths except when the variable is already set to `false` (because a user explicitly opted out).

The new NuGet warning [NU3042](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3042) should make new failures on Linux resulting from this change actionable.

@nkolev92 informally reviewed and blessed this change before his OOF-ness.

CC @dotnet/nuget-team, @aortiz-msft, @skofman1

### DIFF
--- a/src/Cli/dotnet/NuGetSignatureVerificationEnabler.cs
+++ b/src/Cli/dotnet/NuGetSignatureVerificationEnabler.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.MSBuild;
+
+namespace Microsoft.DotNet.Tools
+{
+    public static class NuGetSignatureVerificationEnabler
+    {
+        private static readonly EnvironmentProvider s_environmentProvider = new();
+
+        internal static readonly string DotNetNuGetSignatureVerification = "DOTNET_NUGET_SIGNATURE_VERIFICATION";
+
+        public static void ConditionallyEnable(ForwardingApp forwardingApp, IEnvironmentProvider? environmentProvider = null)
+        {
+            ArgumentNullException.ThrowIfNull(forwardingApp, nameof(forwardingApp));
+
+            if (!IsLinux())
+            {
+                return;
+            }
+
+            string value = GetSignatureVerificationEnablementValue(environmentProvider);
+
+            forwardingApp.WithEnvironmentVariable(DotNetNuGetSignatureVerification, value);
+        }
+
+        public static void ConditionallyEnable(MSBuildForwardingApp forwardingApp, IEnvironmentProvider? environmentProvider = null)
+        {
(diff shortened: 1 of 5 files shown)


donnie-msft (approved):

src/Cli/dotnet/NuGetSignatureVerificationEnabler.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            ArgumentNullException.ThrowIfNull(forwardingApp, nameof(forwardingApp));
+
+            if (!IsLinux())
+            {
+                return;
+            }
+
+            string value = GetSignatureVerificationEnablementValue(environmentProvider);
+
```
donnie-msft, line 30: extra whitespace

donnie-msft: GH tricked me. Nevermind

src/Cli/dotnet/NuGetSignatureVerificationEnabler.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Microsoft.DotNet.Tools
+{
+    public static class NuGetSignatureVerificationEnabler
+    {
+        private static readonly EnvironmentProvider s_environmentProvider = new();
+
+        internal static readonly string DotNetNuGetSignatureVerification = "DOTNET_NUGET_SIGNATURE_VERIFICATION";
+
+        public static void ConditionallyEnable(ForwardingApp forwardingApp, IEnvironmentProvider? environmentProvider = null)
```
kartheekp-ms, line 20: Is it possible to extract the logic from these two overloaded methods, which are identical in their implementation, and create a new method to contain that logic, which can then be called by both of the overloaded methods?

dtivel: Their implementation is similar but not identical.  Crucially, how you set environment variables for `ForwardingApp` and `MSBuildFowardingApp` instances is different.  See [here](https://github.com/dotnet/sdk/pull/31868/files#diff-8777ab94208f467bed9672d3f15e8d1343514f51dc1771f2049c9cbc7a02e2ffR29) and [here](https://github.com/dotnet/sdk/pull/31868/files#diff-8777ab94208f467bed9672d3f15e8d1343514f51dc1771f2049c9cbc7a02e2ffR45).

jeffkl (approved):

PR #31921, opened 2023-04-21
Remove Unnecessary Check In RazorSourceGenerator Targets

lonitra:
Related: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1803616/

### DIFF
--- a/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.SourceGenerators.targets
+++ b/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.SourceGenerators.targets
@@ -26,7 +26,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     </PropertyGroup>
 
     <ItemGroup>
-      <EditorConfigFiles Include="$(RazorEncConfigFile)" Condition="'$(DesignTimeBuild)' == 'true' AND '$(BuildingInsideVisualStudio)' == 'true'"/>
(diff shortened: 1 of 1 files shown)


davidwengier (approved):

davidwengier:
Guessing @chsienki is the authority on this.

jaredpar (changes requested):
This does not feel like the right change.

jaredpar:
Spent a bit of time digging into this scenario today. The two properties in question here:

- `$(BuildingInsideVisualStudio)` this will be true for any build that originates from Visual Studio: design time or real builds. 
- `$(DesignTimeBuild)` this will be true for design time builds only in SDK style projects.

Given that the condition as stated appears to be redundant given that the latter implies the former. Therefore we can remove this.

jaredpar (approved):

DustinCampbell:
@jaredpar, would you prefer that we close this PR and file an issue to fix the SDK more thoroughly?

jaredpar:
I'm good with either. I took a look around to see if there were any other `and` of these two values cause it seems suspicious. Found one in android but none other here.

tmat:
This broke dotnet-watch Hot Reload. 

`dotnet watch` sets `/p:DesignTimeBuild=true` to build the app, but it does not maintain two kinds of solutions (design-time + compile-time) like VS does. Previously, even design time build run Razor source generator in dotnet watch. Now it does not because this condition was removed.

See https://github.com/aspnet/AspNetCore-ManualTests/issues/2029

PR #31949, opened 2023-04-21
Change the workloads txt file name to KnownWorkloadManifests.txt. 

marcpopMSFT:
This will need to be updated in installer and then the overlay will have to be updated with stage0


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -42,7 +42,7 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
             _sdkRootPath = sdkRootPath;
             _sdkVersionBand = new SdkFeatureBand(sdkVersion);
 
-            var knownManifestIdsFilePath = Path.Combine(_sdkRootPath, "sdk", sdkVersion, "IncludedWorkloadManifests.txt");
+            var knownManifestIdsFilePath = Path.Combine(_sdkRootPath, "sdk", sdkVersion, "KnownWorkloadManifests.txt");
```
dsplaisted, line 45: I think for the time being we should look for both files, so that we don't have to make the change in dotnet/installer exactly when the SDK with this change flows in.

src/Layout/redist/targets/OverlaySdkOnLKG.targets
```diff
@@ -35,7 +41,7 @@
       <OverlaySdkFilesFromStage0 Include="$(_DotNetHiveRoot)/sdk/$(Stage0SdkVersion)/Microsoft.NETCoreSdk.BundledCliTools.props" />
       <OverlaySdkFilesFromStage0 Include="$(_DotNetHiveRoot)/sdk/$(Stage0SdkVersion)/RuntimeIdentifierGraph.json" />
       <OverlaySdkFilesFromStage0 Include="$(_DotNetHiveRoot)/sdk/$(Stage0SdkVersion)/NETCoreSdkRuntimeIdentifierChain.txt" />
-      <OverlaySdkFilesFromStage0 Include="$(Stage0IncludedWorkloadManifestsFile)" Condition="Exists('$(Stage0IncludedWorkloadManifestsFile)')"/>
+      <OverlaySdkFilesFromStage0 Include="$(Stage0KnownWorkloadManifestsFile)" Condition="Exists('$(Stage0KnownWorkloadManifestsFile)')"/>
```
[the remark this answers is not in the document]
marcpopMSFT: My prior solution would have ensured we copied the file in use but I don't have concerns copying both for now.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
marcpopMSFT: Updated but tried to keep the logic as similar as possible. You could have issues if one txt file has some workloads and the other has different ones AND the order they load in matters. I can't imagine when we'd end up in that state. I do try to avoid duplicates at least.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -43,16 +43,28 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
...
                 _knownManifestIdsAndOrder = new Dictionary<string, int>();
-                int lineNumber = 0;
                 foreach (var manifestId in File.ReadAllLines(knownManifestIdsFilePath).Where(l => !string.IsNullOrEmpty(l)))
                 {
                     _knownManifestIdsAndOrder[manifestId] = lineNumber++;
                 }
             }
-
+            var includedManifestIdsFilePath = Path.Combine(_sdkRootPath, "sdk", sdkVersion, "IncludedWorkloadManifests.txt");
+            if (File.Exists(includedManifestIdsFilePath))
```
dsplaisted, line 57: I wouldn't duplicate the logic or ever read both files.  How about something like this:

```c#
if (!File.Exists(knownManifestIdsFilePath))
{
    knownManifestIdsFilePath = Path.Combine(_sdkRootPath, "sdk", sdkVersion, "IncludedWorkloadManifests.txt");
}

if (File.Exists(knownManifestIdsFilePath))
{
    // Existing logic to read the file}
```

marcpopMSFT: Ahh, I thought you meant read both files in case a testing environment ended up with both. I should have recognized that that didn't really make sense to read them both from the same sdk as that would only happen with copying files around.

dsplaisted (approved):

PR #31955, opened 2023-04-21
Use 'artifacts' instead of `.artifacts` for default artifacts path

dsplaisted:
Based on feedback from the .NET 8 Preview 3 support for [artifacts output](https://github.com/dotnet/designs/pull/281), this PR removes the `.` from the default artifacts output path.

It also blocks setting the artifacts path in the project file instead of in Directory.Build.props.  Reasons for this include:

- With the `.` removed, adding a new folder inside the project path would cause problems, since the new folder wouldn't be excluded from the default item includes
- It would be unexpected that simply adding an empty Directory.Build.props file to your repo would move the artifacts path from the project folder up the tree to where the Directory.Build.props file was
- Making it an error now might make it easier to change the heuristic in the future to look for .sln files, which might enable us to turn on the new output path format by default for some future TargetFramework.

Some things I still want to add to this PR:

- [x] Add an error if you set `UseArtifactsOutput` to true on the command line but don't have a Directory.Build.props file.
- [x] Add properties that allow you to opt in to putting the artifacts output in the project folder if you want, and to prevent the project name from being added to the path afterwards.
- [x] Verify that artifacts path is removed from globs

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -900,4 +900,9 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1199: The ArtifactsPath and UseArtifactsOutput properties cannot be set in a project file, due to MSBuild ordering constraints. They must be set in a Directory.Build.props file or from the command line. See https://aka.ms/netsdk1199 for more information.</value>
     <comment>{StrBegin="NETSDK1199: "}</comment>
   </data>
+  <data name="UseArtifactsOutputRequiresDirectoryBuildProps" xml:space="preserve">
+    <value>NETSDK1200: If UseArtifactsPath is set to true and ArtifactsPath is not set, there must be a Directory.Build.props file in order to determine where the artifacts folder should be located.</value>
+    <comment>{StrBegin="NETSDK1200: "}</comment>
+  </data>
+  
 </root>
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -896,4 +896,8 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name.</value>
     <comment>{StrBegin="NETSDK1198: "}</comment>
   </data>
+  <data name="ArtifactsPathCannotBeSetInProject" xml:space="preserve">
+    <value>NETSDK1199: The ArtifactsPath and UseArtifactsOutput properties cannot be set in a project file, due to MSBuild ordering constraints. They must be set in a Directory.Build.props file or from the command line. See https://aka.ms/netsdk1199 for more information.</value>
+    <comment>{StrBegin="NETSDK1199: "}</comment>
+  </data>
 </root>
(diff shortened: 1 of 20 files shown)


baronfel (approved):
Wording of the message and the `.` removal looks great to me :+1:

timheuer:
I feel like we should also PR adding `artifacts` then to default .NET `.gitignore` files across the ecosystem then (similar to how `dist` is on default for JS apps

baronfel:
@timheuer it's already in the gitignore for the .NET CLI [templates](https://github.com/dotnet/sdk/blob/f921fc64463743c72397827926158abfc531c6a6/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Gitignore/.gitignore#L63) and the community-maintained ones at [github/gitignore](https://github.com/github/gitignore/blob/main/VisualStudio.gitignore#L63), so we should be covered.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -896,4 +896,13 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1198: A publish profile with the name '{0}' was not found in the project. Set the PublishProfile property to a valid file name.</value>
     <comment>{StrBegin="NETSDK1198: "}</comment>
   </data>
+  <data name="ArtifactsPathCannotBeSetInProject" xml:space="preserve">
+    <value>NETSDK1199: The ArtifactsPath and UseArtifactsOutput properties cannot be set in a project file, due to MSBuild ordering constraints. They must be set in a Directory.Build.props file or from the command line. See https://aka.ms/netsdk1199 for more information.</value>
```
joeloff, line 900: Do we need to manually lock the ArtifactsPath and UseArtifactsOutput strings inside a `<note />` section for the translators so they cannot translate the property names by accident?

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -74,6 +74,12 @@ Copyright (c) .NET Foundation. All rights reserved.
     <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)' == ''">$(ArtifactsPath)\obj\</BaseIntermediateOutputPath>
   </PropertyGroup>
 
+  <!-- Record whether ArtifactsPath / UseArtifactsOutput was set at this point in evaluation.  We will generate an error if these properties are set
```
joeloff, line 77: Should this be inside the PropertyGroup on line 67 (it has the same condition)?

src/Tasks/Common/Resources/Strings.resx
dsplaisted: Might be a good idea, can you point me to an example of what that looks like.

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
dsplaisted: Good idea, but I will try to do that later since I'd like to get this in before we branch for Preview 4.

src/Tasks/Common/Resources/Strings.resx
nagilson: You can add something like:
`<note>{Locked="dotnet workload search"}</note>`
if `dotnet workload search` is what you want to lock.

This PR has a lot of examples: https://github.com/dotnet/sdk/pull/28553/files

joeloff: In the ENU resx file you should be able to add something like
```xml
<comment>{Locked="ArtifactsPath, UseArtifactsOutput"}</comment>
```
The comments will automatically be added in the XLF as notes for the translators IIRC

nagilson:
I think this is a great change. With that said, this is technically a breaking change even though it only breaks one preview to preview version. Im not sure what our modus operandi here is but IMO we should create a breaking change doc, especially since this is removing a feature, we should do all we can to give clarity as to why we took something away and how to fix it.

Issue #30336, opened 2023-02-03
[Proposal] .NET Tools should default to running on the latest available Runtime


baronfel:
### Is your feature request related to a problem? Please describe.

Currently, .NET applications use a runtimeConfig to determine which runtime they run on. This is something that is determined at build time, but non-tool .NET applications can override this with various flags and environment variables given to the .NET runtime host. The default rollforward policy is `LatestPatch`, which allows an application to run on any runtime of the same major/minor version that is _at least_ as high a version as the application was built against.

.NET (local) Tools are a specific kind of .NET application that are intended to be distributed and managed via SDK commands. They are generally development tools/utilities. We think the semantics of tools are such that tools should be able to run on _any_ .NET runtime equal or greater than the runtime the application originally targeted. This will reduce friction for users adopting new SDKS/Runtimes by not completely blocking their workflows without some explicit proof on incompatibility.

Without this change, tool authors must either explicitly set a more flexible roll-forward policy, or multitarget. Multitargeting has the side effect of increasing the overall size of the tool, and still not solving the issue - a tool that multitargets net6.0 and net7.0 will still not run on net8.0 runtimes.

### Describe the solution you'd like

There are two places we could address this mismatch:

* build time, when a tool's RollForward defaults are baked into the resulting application
* install time, when any apphosts/runtimeConfigs are authored by the SDK for subsequent executions

### Build Time changes

We would change the `RollForward` of all tools from `LatestPatch` to `LatestMajor`. This would cause the desired behavior for all newly-built apps. We would also track a sentinel value that _we_ set the RollForward, so that at install time we could tell if the user explicitly requested a change to the RollForward.

### Install Time changes

We would ensure that tools that did not explicitly override to force a certain RollForward behavior would be written with `LatestMajor` behavior. This helps us plug the existing behavior gap for tools that haven't been authored with an SDK containing these changes.

### Additional context
* https://github.com/dotnet/sdk/issues/26824
* https://github.com/dotnet/command-line-api/issues/2014
* https://github.com/fsprojects/FAKE/pull/2372
* https://github.com/cake-build/cake/blob/develop/src/Shared.msbuild#L4

baronfel:
Pinging @richlander and @vitek-karas beacuse this behavior change would technically limit the flexibility of the runtime to make breaking changes.

vitek-karas:
[...]

PR #31957, opened 2023-04-21
#30336 Default Dotnet Tool to Run on Latest Runtime - Build time change

JL03-Yue:
#30336 NET Tools should default to running on the latest available Runtime 

Build time change: 
Change the `RollForward` of all tools from `LatestPatch` to `Major`, so that tools can be able to run on any .NET runtime equal or greater than the runtime the application originally targeted

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
@@ -42,10 +42,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DotnetCliToolTargetFramework>netcoreapp2.2</DotnetCliToolTargetFramework>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <IncludeBuildOutput Condition=" '$(PackAsTool)' == 'true' ">false</IncludeBuildOutput>
-    <PackageType Condition=" '$(PackAsTool)' == 'true' ">DotnetTool</PackageType>
-    <RuntimeIdentifiers Condition=" '$(PackAsTool)' == 'true' ">$(RuntimeIdentifiers);$(PackAsToolShimRuntimeIdentifiers)</RuntimeIdentifiers>
+  <PropertyGroup Condition=" '$(PackAsTool)' == 'true' ">
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <PackageType>DotnetTool</PackageType>
+    <RuntimeIdentifiers>$(RuntimeIdentifiers);$(PackAsToolShimRuntimeIdentifiers)</RuntimeIdentifiers>
+    <RollForward Condition="'$(RollForward)' == ''">Major</RollForward>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(EnablePreviewFeatures)' == 'true' And '$(IsNetCoreAppTargetingLatestTFM)' == 'true'">
--- a/src/Tests/Microsoft.NET.ToolPack.Tests/GivenThatWeWantToPackAToolProject.cs
+++ b/src/Tests/Microsoft.NET.ToolPack.Tests/GivenThatWeWantToPackAToolProject.cs
@@ -13,8 +13,10 @@ using Xunit.Abstractions;
 using NuGet.Packaging;
 using System.Xml.Linq;
 using System.Runtime.CompilerServices;
+using Microsoft.NET.TestFramework.ProjectConstruction;
 using System;
 using System.Runtime.InteropServices;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.NET.ToolPack.Tests
 {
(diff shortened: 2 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
```diff
@@ -42,10 +42,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DotnetCliToolTargetFramework>netcoreapp2.2</DotnetCliToolTargetFramework>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <IncludeBuildOutput Condition=" '$(PackAsTool)' == 'true' ">false</IncludeBuildOutput>
-    <PackageType Condition=" '$(PackAsTool)' == 'true' ">DotnetTool</PackageType>
-    <RuntimeIdentifiers Condition=" '$(PackAsTool)' == 'true' ">$(RuntimeIdentifiers);$(PackAsToolShimRuntimeIdentifiers)</RuntimeIdentifiers>
+  <PropertyGroup Condition=" '$(PackAsTool)' == 'true' ">
```
nagilson, line 45: Thank you for putting that condition up at the top. It IS much cleaner that way. 😄

src/Tests/Microsoft.NET.ToolPack.Tests/GivenThatWeWantToPackAToolProject.cs
```diff
@@ -300,5 +302,37 @@ public void Given_targetplatform_set_It_should_error()
...
+                .Should()
+                .Pass();
+
+            var outputDirectory = packCommand.GetOutputDirectory(testProject.TargetFrameworks);
+
+            string runtimeConfigFile = Path.Combine(outputDirectory.FullName, testProject.Name + ".runtimeconfig.json");
+            JObject runtimeConfig = ReadRuntimeConfig(runtimeConfigFile);
+            runtimeConfig["runtimeOptions"]["rollForward"].Value<string>()
+                .Should().Be("Major");
+        }
```
nagilson, line 331: nit: add a blank link between these functions

src/Tests/Microsoft.NET.ToolPack.Tests/GivenThatWeWantToPackAToolProject.cs
```diff
@@ -300,5 +302,37 @@ public void Given_targetplatform_set_It_should_error()
...
+        private JObject ReadRuntimeConfig(string runtimeConfigPath)
```
nagilson, line 332: nit: add a blank line between these functions

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
```diff
@@ -42,10 +42,11 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <PackageType>DotnetTool</PackageType>
+    <RuntimeIdentifiers>$(RuntimeIdentifiers);$(PackAsToolShimRuntimeIdentifiers)</RuntimeIdentifiers>
+    <RollForward Condition="'$(RollForward)' == ''">Major</RollForward>
```
nagilson, line 49: Where is the detection logic for `RollFoward`? I dont see this anywhere in the SDK repository. I dont know if `Major` is the same as `LatestMajor`?

src/Tests/Microsoft.NET.ToolPack.Tests/GivenThatWeWantToPackAToolProject.cs
```diff
@@ -300,5 +302,37 @@ public void Given_targetplatform_set_It_should_error()
             result.Should().Fail().And.HaveStdOutContaining("NETSDK1146");
 
         }
+
+        [Fact]
+        public void WhenPackAToolItRollForwardDefaultToMajor()
```
nagilson, line 307:
```suggestion
        public void WhenPackingAToolItDefaultsRollsForwardToMajor()
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
nagilson, line 49: @baronfel Is it possible for us to add some logic to detect if we did do a roll-forward and then errored.. and if so warn the user about the roll forward and tell them to use an older SDK?

nagilson (commented):
The change looks good so far bar a few questions and comments. :) 
We should get a breaking change doc written for this when its merged. Feel free to reach out to me for help on that if you have any questions.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
baronfel: This is not a bad idea. What happens with tools is that 
* we eventually write a `DotnetToolSettings.xml` file that looks like this:

```
<DotNetCliTool Version="1">
  <Commands>
    <Command Name="fsautocomplete" EntryPoint="fsautocomplete.dll" Runner="dotnet" />
  </Commands>
</DotNetCliTool>
```

* the rollforward is set in the runtimeconfig.json to whatever we infer here

We probably need to set some other kind of data in the `DotnetToolSettings.xml` to signal that _we_ opted the user into auto-rollforward so we could give an error here.

nagilson (dismissed):
Welp. That was a misclick.. please ignore 🤣

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets
JL03-Yue: I think `Major` works here for `RollForward`? Can @baronfel confirm this?

baronfel: `Major` is what we want here - `LatestMajor` would _force_ the tool to run on a more recent runtime if one was available, but we want this automatic re-targeting to be a last-chance effort. If a tool was compiled using a TFM of net7.0, it was clearly _intended_ to run on the 7.0 runtime if present, and we should prefer that. A user can signal more braod support by either choosing `LatestMajor` themselves, or multitargeting the tool to target e.g. `net8.0`.

baronfel: I don't think we need to do this (add the data to do error detection) for this PR - I think this behavior is correct and good and what we should have done all along, and a breaking change notice is enough. Virtually every tool author I've spoken to has wondered why this wasn't the default, so I don't think there's a user experience gap that we need to cover here.

If we get feedback from users post-preview5 that this was confusing, we can certainly come back to this though!

nagilson (approved):
Thanks for following up here :)

PR #31980, opened 2023-04-24
Update AOT console template

MichalStrehovsky:
Based on discussions.

* StripSymbols is going to be SDK default.
* `--aot` instead of `-aot` to match Unix conventions.

Cc @dotnet/ilc-contrib @am11 @DamianEdwards @eerhardt


vlada-shubina (approved):

eerhardt (approved):
LGTM. Thanks!

PR #31997, opened 2023-04-24
See what fails if we don't install the old runtimes

marcpopMSFT:


marcpopMSFT:
135 test failures so we can't get rid of this just yet.

PR #32003, opened 2023-04-24
add artifacts path to buildprops template

baronfel:
This adds a flag that users can use to opt into a form of the new SDK artifacts output layout that was introduced in preview 3. Since the 'default' way of enabling this functionality can 'float' the root directory's location as new buildprops are introduced, we think it's safest to pin the artifacts directory location to the directory containing the D.B.p file. Since this gesture is relatively complex to a non-MSBuild user, scaffolding the expression also makes it easier to on-board users to the new functionality.

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/.template.config/dotnetcli.host.json
+++ b/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/.template.config/dotnetcli.host.json
@@ -1,0 +1,12 @@
+{
+  "$schema": "http://json.schemastore.org/dotnetcli.host",
+  "symbolInfo": {
+    "useArtifacts": {
+      "longName": "use-artifacts",
+      "shortName": ""
+    },
+    "inherit": {
+      "shortName": ""
+    }
+  }
+}
--- a/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/Directory.Build.props
+++ b/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/Directory.Build.props
@@ -6,5 +6,9 @@
 <!--#endif-->
   <PropertyGroup>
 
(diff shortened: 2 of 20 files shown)


template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/.template.config/dotnetcli.host.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json.schemastore.org/dotnetcli.host",
+  "symbolInfo": {
+    "useArtifacts": {
+      "longName": "use-artifacts",
+      "shortName": ""
```
baronfel, line 6: the short name for this was `-ua` without this property. instead we disable it entirely.

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/.template.config/dotnetcli.host.json
```diff
@@ -0,0 +1,12 @@
...
+    },
+    "inherit": {
+      "shortName": ""
```
baronfel, line 9: the generated short name was `-id`, which is not POSIX, so we delete it entirely instead.

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/BuildProps/Directory.Build.props
```diff
@@ -6,5 +6,9 @@
 <!--#endif-->
   <PropertyGroup>
 
+<!--#if useArtifacts -->
+  <ArtifactsPath>$(MSBuildThisFileDirectory)artifacts</ArtifactsPath>
```
baronfel, line 10: @dsplaisted to confirm this is the correct expression. I think it is, but safety-first!

dsplaisted (approved):

PR #32045, opened 2023-04-25
Reference ILLink analyzer matching ILLink version

### LINKED ISSUE #84119: ILLink analyzer and code fixer passed twice to the csc
Using very recent SDK 8.0.0-preview.4.23177.1

```console
dotnet new api -aot
dotnet publish /bl
```

Open the `msbuild.binlog`:
image

This means that any warning produced by the ILLink analyzer is reported twice.

sbomer:
Fixes https://github.com/dotnet/runtime/issues/84119.

We were already referencing the analyzer from the ILLink package on net8.0, so there we only needed to remove the bundled analyzer to fix https://github.com/dotnet/runtime/issues/84119.

For older TFMs, we currently use the latest analyzer that's bundled with the SDK. I replaced this with logic to use a packagreference, matching the version of ILLink. Note that with this change, when targeting downlevel TFMs the 8.0 SDK will use the 7.0 analyzer instead of the 8.0 analyzer that we use today.


src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -709,6 +709,15 @@ private enum ToolPackType
                 implicitPackageReferences.Add(buildPackage);
             }
 
+            // Before net8.0, ILLink analyzers shipped in a separate package.
+            // Add the analyzer package with version taken from KnownILLinkPack.
+            if (normalizedTargetFrameworkVersion < new Version(8, 0) && toolPackType is ToolPackType.ILLink)
+            {
+                var analyzerPackage = new TaskItem("Microsoft.NET.ILLink.Analyzers");
+                analyzerPackage.SetMetadata(MetadataKeys.Version, packVersion);
+                implicitPackageReferences.Add(analyzerPackage);
```
vitek-karas, line 718: Just curious: where do we download this from? It seems it's not on NuGet... is the SDK adding some other public feeds for this?

vitek-karas (approved):

eng/Version.Details.xml
```diff
@@ -120,10 +120,6 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>195bf987887ea645a747ff0586440f7920c47b01</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NET.ILLink.Analyzers" Version="8.0.100-1.23067.1">
-      <Uri>https://github.com/dotnet/linker</Uri>
-      <Sha>c790896f128957acd2999208f44f09ae1e826c8c</Sha>
-    </Dependency>
```
dsplaisted, line 126: Should the corresponding version number be removed from `Versions.props`?  Otherwise I believe it will still be there and available for use, but won't ever get updated.

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
dsplaisted: I think it's downloaded from NuGet: https://www.nuget.org/packages/Microsoft.NET.ILLink.Analyzers/

For internal builds the specific version would be available on an internal feed.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -17,7 +17,13 @@ Copyright (c) .NET Foundation. All rights reserved.
     <IsPublishable Condition="'$(IsPublishable)' == ''">true</IsPublishable>
     <!-- PublishAot depends on PublishTrimmed. This must be set early enough for the KnownILLinkPack to be restored. -->
     <PublishTrimmed Condition="'$(PublishTrimmed)' == '' And '$(PublishAot)' == 'true'">true</PublishTrimmed>
-    <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</_IsTrimmingEnabled>
+    <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == '' And (
+        '$(PublishTrimmed)' == 'true' Or
+        '$(PublishSingleFile)' == 'true' Or
```
eerhardt, line 22: Can we rename `_IsTrimmingEnabled`? It doesn't seem like a good name anymore, if we are setting it to `true` when I only set `PublishSingleFile`.

ViktorHofer: dotnet/runtime uses that property already. @sbomer can you please take care of updating it when these changes are being consumed? https://github.com/dotnet/runtime/blob/98dd1e83ba8b17299af9d3d78000f8a9c000a12e/eng/illink.targets#L5C1-L6

mthalman:
@sbomer - This causes a failure when using the SDK containing these changes to build the runtime repo. This is affecting source-build. Can you investigate please? I have this issue logged with details: https://github.com/dotnet/source-build/issues/3490

PR #32100, opened 2023-04-27
Allow opting-in to NativeAOT publishing using runtime packs

akoeplinger:
Contributes to https://github.com/dotnet/runtime/issues/81024

Installer changes are in https://github.com/dotnet/installer/pull/16261

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -48,6 +48,8 @@ namespace Microsoft.NET.Build.Tasks
 
         public bool AotEnabled { get; set; }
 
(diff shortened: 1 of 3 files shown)


dsplaisted:
I don't really follow what's going on here.  Is the `AotUseKnownRuntimePackForTarget` property something you'd expect developers to set to opt in to this new mode?

akoeplinger:
@dsplaisted no, this would almost certainly only be set by the workloads (e.g. iOS/Android) that need the runtime packs. There is a chance that we will switch completely to the new mode even for desktop at which point the property would go away so I'm not particularly concerned with the naming.

dsplaisted (approved):
Looks good so far.  Sounds like changes will be needed to workloads or other logic to opt in to this. Also I'm not sure if the ILCompiler will need to be updated to know to use the runtime pack.

PR #32121, opened 2023-04-28
Remove reference to 'BINDING' in 'BlazorHotReload.js'

### LINKED ISSUE #47878: [blazor] Remove reference to BINDING api in global namespace
The blazor hotreload feature expects BINDING api to be exported to global namespace in the https://github.com/dotnet/sdk/blob/main/src/BuiltInTools/BrowserRefresh/BlazorHotReload.js.

The https://github.com/dotnet/aspnetcore/pull/46693 rewrote the invoke as a JSImport and with that we can call directly the async function `receiveHotReloadAsync`.

With the resolution of this issue we can remove `disableDotnet6Compatibility: false` from https://github.com/maraf/aspnetcore/blob/99a1d2c0cb39cbddb0e771d76c15fa96cae46261/src/Components/Web.JS/src/Platform/Mono/MonoPlatform.ts#L347

MackinnonBuck:
This is the last step required to enable the removal of `disableDotnet6Compatibility: false` from https://github.com/maraf/aspnetcore/blob/99a1d2c0cb39cbddb0e771d76c15fa96cae46261/src/Components/Web.JS/src/Platform/Mono/MonoPlatform.ts#L347.

This change will require a corresponding reaction in dotnet/aspnetcore: https://github.com/dotnet/aspnetcore/pull/47964

Fixes https://github.com/dotnet/aspnetcore/issues/47878

### DIFF
--- a/src/BuiltInTools/BrowserRefresh/BlazorHotReload.js
+++ b/src/BuiltInTools/BrowserRefresh/BlazorHotReload.js
@@ -1,8 +1,4 @@
-export function receiveHotReload() {
-  return BINDING.js_to_mono_obj(new Promise((resolve) => receiveHotReloadAsync().then(resolve(0))));
-}
-
-async function receiveHotReloadAsync() {
+export async function receiveHotReloadAsync() {
   const cache = window.sessionStorage.getItem('blazor-webassembly-cache');
   let headers;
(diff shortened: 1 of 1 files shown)


maraf (approved):
Looks good

PR #32141, opened 2023-04-30
Fix license headers in all source files

tmat:
Applies code fix to update source file headers.
Some files had different wording, some incorrectly referred to Apache license, some had Xamarin copyright, some did not have any license header.


tmat:
@richlander

richlander (approved):

richlander:
Thanks for the fixes.

.editorconfig
```diff
@@ -161,7 +161,7 @@ csharp_space_between_square_brackets = false
 dotnet_code_quality.api_surface = all
 
 # License header
-file_header_template = Copyright (c) .NET Foundation and contributors. All rights reserved.\nLicensed under the MIT license. See LICENSE file in the project root for full license information.\n
+file_header_template = Copyright (c) .NET Foundation and contributors. All rights reserved.\nLicensed under the MIT license. See LICENSE file in the project root for full license information.
```
mavasani, line 164: I see that the diagnostic is marked as a warning: https://github.com/dotnet/sdk/blob/dfae9b7b85f2b693a9a55f8d70b08af01d3ee83f/.editorconfig#L335-L336

Why isn't this enforced on CI? Or are none of the code style rules enforced on build on this repo?

tmat: Not sure. Roslyn's model of running a separate "correctness" CI leg would be nice, but I don't think this repo has it.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/ISuppressionEngine.cs
```diff
@@ -1,5 +1,5 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
ViktorHofer, line 2: Why are making this change? dotnet/runtime uses the existing license header everywhere: https://github.com/dotnet/runtime/blob/5d370f31305012b074f070337ace39ca1b5aedeb/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L1-L2

cc @jkotas

jkotas: "Licensed to the .NET Foundation under one or more agreements." is the correct header to use in .NET Foundation projects.

See https://github.com/dotnet-foundation/foundation/blob/master/guidance/copyright.md#file-headers . 

dotnet/runtime further shortened it to 2 lines, both the make it shorter and to address some problems with the 3 line version. See https://github.com/dotnet/runtime/pull/38732 for rationale.

src/Cli/Microsoft.DotNet.Cli.Sln.Internal/FileManipulation/FileUtil.cs
```diff
@@ -1,28 +1,5 @@
-//
-// FileUtil.cs
-//
-// Author:
-//       Lluis Sanchez Gual <lluis@xamarin.com>
-//
-// Copyright (c) 2015 Xamarin, Inc (http://www.xamarin.com)
```
jkotas, line 7: In general, it is not ok to delete Xamarin and Novell copyrights like this (in dotnet/runtime at least).

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Logging/ISuppressionEngine.cs
tmat: Sounds good. I'll update the editorconfig based on dotnet/runtime repo and run the quick fix again.

src/Cli/Microsoft.DotNet.Cli.Sln.Internal/FileManipulation/FileUtil.cs
tmat: Done.

jkotas (approved):

PR #32145, opened 2023-05-01
Make TryGetInternalDefaultConstructor work without internal metadata

### LINKED ISSUE #32119: GenAPI drops internal ctor which results in compilation errors
Microsoft.DotNet.GenAPI.Task/8.0.100-preview.5.23228.4 drops the following API: https://github.com/dotnet/runtime/blob/e6f9af1361030fca92abe58efed1ea45d2eb552c/src/libraries/System.IO.Packaging/ref/System.IO.Packaging.cs#L180

I verified that the latest changes are part of the used GenAPI version. A repro in the source-build-reference-packages repo:
```sh
./generate.sh --package "System.IO.Packaging,7.0.0" --excludeDependencies
./build.sh
# The above command failed for me at some point so I then built the generated project manually:
.dotnet/dotnet build /home/vihofer/git/source-build-reference-packages/src/referencePackages/src/system.io.packaging/7.0.0/System.IO.Packaging.7.0.0.csproj
```

Parameters used for invoking the GenAPI task:
image

cc @ericstj @andriipatsula

ericstj:
Fixes #32119 

TryGetInternalDefaultConstructor was previously expecting to be able to see an internal constructor in order to generate a non-public default constructor.  This will not be the case if the INamedTypeSymbol came from an assembly imported with MetadataImportOptions.Public (as is the default) or if the linker ran on the assembly and removed the constructor.

This change adds the default constructor when none are present and fixes the method to do nothing for types which do not get default constructors.

Note: existing test cases already cover this scenario but miss the bug since they are all running with an assembly from the same compilation that defines the internal API - so there is no filtering.  I'm planning a separate change which modifies all tests to use an assembly written to a MemoryStream then re-imported, but this is a larger change which I'd like to keep separate.  It will also involve some discussion around when to apply the `MetadataImportOptions`.

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
@@ -195,41 +195,48 @@ namespace Microsoft.DotNet.GenAPI
         /// <param name="symbolFilter">Assembly symbol filter <see cref="ISymbolFilter"/>.</param>
         public static IEnumerable<SyntaxNode> TryGetInternalDefaultConstructor(this INamedTypeSymbol namedType, ISymbolFilter symbolFilter)
         {
-            IEnumerable<IMethodSymbol> constructors = namedType.InstanceConstructors;
-
-            // if non-implicit constructors exist and all are filtered
-            if (constructors.Any(c => !c.IsImplicitlyDeclared) && !constructors.Any(symbolFilter.Include))
+            // only non-static classes can have a default constructor that differs in visibility
+            if (namedType.IsStatic ||
+                namedType.TypeKind != TypeKind.Class)
             {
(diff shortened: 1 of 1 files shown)


src/GenAPI/Microsoft.DotNet.GenAPI/INamedTypeSymbolExtension.cs
```diff
@@ -195,41 +195,48 @@ public static bool HasInaccessibleTypeArgument(this INamedTypeSymbol namedType,
         /// <param name="symbolFilter">Assembly symbol filter <see cref="ISymbolFilter"/>.</param>
         public static IEnumerable<SyntaxNode> TryGetInternalDefaultConstructor(this INamedTypeSymbol namedType, ISymbolFilter symbolFilter)
         {
```
ericstj, line 197: note to reviewers: hide whitespace
image

ViktorHofer (approved):

Issue #32038, opened 2023-04-25
Errors restoring runtime pack


lewing:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=251447
Build error leg or test failing: Microsoft.NET.Pack.Tests.dll.1.WorkItemExecution
Pull request: https://github.com/dotnet/sdk/pull/32029

## Error Message
Fill the error message using [known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json

{
  "ErrorMessage": "The runtime pack for Microsoft.NETCore.App.Runtime.Mono.browser-wasm was not downloaded.",
  "BuildRetry": false,
  "ErrorPattern": "",
  "ExcludeConsoleLog": false
}
```

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[259234](https://dev.azure.com/dnceng-public/public/_build/results?buildId=259234)|dotnet/sdk|[Microsoft.NET.Sdk.BlazorWebAssembly.Tests.dll.2.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=259234&view=ms.vss-test-web.build-test-results-tab&runId=5069456&resultId=103352)|dotnet/sdk#32163|
|[259233](https://dev.azure.com/dnceng-public/public/_build/results?buildId=259233)|dotnet/sdk|[Microsoft.NET.Sdk.BlazorWebAssembly.Tests.dll.2.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=259233&view=ms.vss-test-web.build-test-results-tab&runId=5069360&resultId=103352)|dotnet/sdk#32162|
|[259152](https://dev.azure.com/dnceng-public/public/_build/results?buildId=259152)|dotnet/sdk|[Microsoft.NET.Sdk.BlazorWebAssembly.Tests.dll.2.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=259152&view=ms.vss-test-web.build-test-results-tab&runId=5066922&resultId=103352)|dotnet/sdk#29880|
[...]

PR #32170, opened 2023-05-02
[wasm] Move SelfContained=true from pack to SDK

maraf:
- Based on suggestion from https://github.com/dotnet/sdk/issues/32038#issuecomment-1530685783
- Runtime PR is https://github.com/dotnet/runtime/pull/85636


lewing (approved):

dsplaisted:
Looks good.

Do we plan to also remove the (now redundant) property from the NuGet package?

maraf:
> Looks good.
> 
> Do we plan to also remove the (now redundant) property from the NuGet package?

Yes. It's in the referenced PR in runtime

PR #32213, opened 2023-05-03
Add AssemblyName metadata to references based on FrameworkList.xml

ladipro:
We are adding an optimization to RAR to use metadata passed from the `ResolveTargetingPackAssets` task to eliminate the costly sniffing of framework assemblies: https://github.com/dotnet/msbuild/pull/8688

In this PR I am adding the `AssemblyName` metadatum to the output references for MSBuild to consume. I am also adding test coverage for this and few other metadata RAR is newly depending on.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolveTargetingPackAssetsTask.cs
@@ -34,7 +34,15 @@ namespace Microsoft.NET.Build.Tasks.UnitTests
 
             task.Execute().Should().BeTrue();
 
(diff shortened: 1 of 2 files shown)


ladipro:
@dsplaisted, related to the same RAR optimization that made me create this PR, is it safe to assume that all assemblies returned from `ResolveTargetingPackAssets` have runtime version `v4.0.30319`? MSBuild is reading the version to determine if the assembly is a winmd file, and being able to shortcircuit for framework files would further eliminate some I/O.

https://github.com/dotnet/msbuild/blob/768cad16718650f120dd42acfcc9cf5c4dec555f/src/Tasks/AssemblyDependency/AssemblyInformation.cs#L231-L268

dsplaisted (approved):

ladipro:
Winmd files have runtime versions similar to `WindowsRuntime 1.4`. However, funnily enough, the .NET Core implementation of .NET metadata reader in MSBuild appears to have a bug and it doesn't return the string if it doesn't start with `v`. Consequently the check for winmd has been broken so I think we're safe to assume that no such files are distributed with the SDK.

Issue #32165, opened 2023-05-01
GenAPI emits default constructors when reading from metadata


ericstj:
### Describe the bug
GenAPI emits default constructors when they are not needed.

### To Reproduce
```c#
public partial class Class1 { }
```
Compile this and point GenAPI to it.

Expect: no constructor emitted since it isn't required.
Actual:
```c#
    public partial class Class1
    {
        public Class1() { }
    }
```

### Further technical details
The problem appears to be that when reading from metadata the roslyn API doesn't set the `IsImplicitlyDeclared` flag.  This could be because it cannot determine if the default constructor was implicit or explicit from metadata alone.

If we wanted to fix this, we might be able to filter the constructor ourselves, if we see that it is parameter-less, has expected visibility (public, protected for abstract classes, etc), and calls the default base constructor.

ViktorHofer:
Should this issue have been closed with https://github.com/dotnet/sdk/pull/32276?

ericstj:
No the tool still emits the constructor when reading from metadata.  I think we may just want to keep that behavior though.

ViktorHofer:
Agreed and we just recently talked about this. Let's close the issue as not planned.

Issue #31196, opened 2023-03-14
GenAPI: __arglist is not preserved as a parameter


smasher164:
Discovered in https://github.com/dotnet/sdk/pull/30932, the following example
```C#
namespace A
{
    public partial class B
    {
        public void Foo(__arglist) {}
    }
}
```
will not result in `__arglist` printed in the output.

ViktorHofer:
Low priority. Moving into backlog.

PR #32276, opened 2023-05-05
Fix GenAPI when reading assemblies from metadata

### LINKED ISSUE #32152: GenAPI emits compiler generated members when internals are included
### Describe the bug
When running GenAPI and specifying  `--include-visible-outside` attributes defined by the compiler are emitted.

### To Reproduce
```
dotnet tool install Microsoft.DotNet.GenAPI.Tool --prerelease
dotnet genapi --assembly "C:\Program Files\dotnet\packs\Microsoft.NETCore.App.Ref\7.0.5\ref\net7.0\System.Linq.dll" --include-visible-outside
```

Expect: only authored user defined types appear
Actual: compiler generated types appear.

Sample undesired output:
```C#
namespace <global namespace>
{
    internal partial class <Module>
    {
    }
}
namespace Microsoft.CodeAnalysis
{
    [System.Runtime.CompilerServices.CompilerGeneratedAttribute]
    [Embedded]
    internal sealed partial class EmbeddedAttribute : System.Attribute
    {
        public EmbeddedAttribute() { }
    }
}
namespace System.Runtime.CompilerServices
{
    [CompilerServices.CompilerGeneratedAttribute]
    [Embedded]
    [System.AttributeUsageAttribute]
    internal sealed partial class NullableAttribute : System.Attribute
    {
        public readonly System.Byte[] NullableFlags;
        public NullableAttribute(System.Byte value) { }

        public NullableAttribute(System.Byte[] value) { }
    }

    [CompilerServices.CompilerGeneratedAttribute]
    [Embedded]
    [System.AttributeUsageAttribute]
    internal sealed partial class NullableContextAttribute : System.Attribute
    {
        public readonly System.Byte Flag;
        public NullableContextAttribute(System.Byte value) { }
    }

    [CompilerServices.CompilerGeneratedAttribute]
    [Embedded]
    [System.AttributeUsageAttribute]
    internal sealed partial class NullablePublicOnlyAttribute : System.Attribute
    {
        public readonly System.Boolean IncludesInternals;
        public NullablePublicOnlyAttribute(System.Boolean value) { }
    }

    [CompilerServices.CompilerGeneratedAttribute]
    [Embedded]
    [System.AttributeUsageAttribute]
    internal sealed partial class RefSafetyRulesAttribute : System.Attribute
    {
        public readonly System.Int32 Version;
        public RefSafetyRulesAttribute(System.Int32 value) { }
    }
}
```

### Further technical details
I discovered this when looking at https://github.com/dotnet/sdk/issues/32119#issuecomment-1529073358.  I modified the GenAPI tests to emit an assembly to memory and reload it (rather than operating on the IAssemblySymbol from the original compilation) and I found every test that included internals would fail since these compiler generated types were emitted.

ericstj:
Fix: #32152 #32165 
Related: https://github.com/dotnet/sdk/issues/32119

GenAPI was testing with symbols from source code, which contain more information than those which are read from metadata.

When reading from metadata, IsImplicitlyDeclared cannot always be determined, compiler generated types and members are present, records look like regular types, etc.

This fixes the tests to read from metadata instead of source, since running GenAPI on source symbols isn't something the tool will ever do.

It also fixes an issue where Roslyn's default is to only read a subset of all metadata - MetadataImportOptions.Public which is a mix of public API and some internal API that the compiler deems relevant.  We need more metadata when generating internal API, or doing API compat against internal API.


src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoaderFactory.cs
```diff
@@ -8,7 +8,14 @@ namespace Microsoft.DotNet.ApiSymbolExtensions
     /// </summary>
     public sealed class AssemblySymbolLoaderFactory : IAssemblySymbolLoaderFactory
     {
+        public AssemblySymbolLoaderFactory(bool includeInternals = false)
+        {
+            IncludeInternals = includeInternals;
+        }
+
+        public bool IncludeInternals { get; }
```
ViktorHofer, line 16: Please add this to the interface with triple slash documentation and then use inheritdoc here.

src/GenAPI/Microsoft.DotNet.GenAPI/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+            // Add test case for this branch when that is fixed.
+            if (method.IsVararg != otherMethod.IsVararg)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public static bool IsImplicitDefaultConstructor(this IMethodSymbol method, ISymbolFilter symbolFilter)
```
ViktorHofer, line 66: As this is public API, please add triple slash documentation.

src/GenAPI/Microsoft.DotNet.GenAPI/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+            // these are the only two accessibilities given to the implicit default constructor, regardless of containing type accessibility.
+            if (method.ContainingType.IsAbstract ?
+                    method.DeclaredAccessibility != Accessibility.Protected :
+                    method.DeclaredAccessibility != Accessibility.Public)
+            {
+                return false;
+            }
+
+            // if the type defines any other instance constructors the default constructor cannot be implicit 
+            if (method.ContainingType.InstanceConstructors.Where(c => symbolFilter.Include(c) && !c.Parameters.IsEmpty).Any())
```
ViktorHofer, line 84:
```suggestion
            if (method.ContainingType.InstanceConstructors.Any(c => symbolFilter.Include(c) && !c.Parameters.IsEmpty))
```

src/GenAPI/Microsoft.DotNet.GenAPI/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (method.MethodKind != MethodKind.Constructor || !method.Parameters.IsEmpty)
+            {
+                return false;
+            }
+
+            // see https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#15115-default-constructors
+            // these are the only two accessibilities given to the implicit default constructor, regardless of containing type accessibility.
+            if (method.ContainingType.IsAbstract ?
+                    method.DeclaredAccessibility != Accessibility.Protected :
+                    method.DeclaredAccessibility != Accessibility.Public)
```
ViktorHofer, line 78: nit: indentation doesn't look right.

src/GenAPI/Microsoft.DotNet.GenAPI/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            // if the type defines any other instance constructors the default constructor cannot be implicit 
+            if (method.ContainingType.InstanceConstructors.Where(c => symbolFilter.Include(c) && !c.Parameters.IsEmpty).Any())
+            {
+                return false;
+            }
+
+            // if the base type does not have a parameterless constructor the default constructor cannot be implicit
+            var baseType = method.ContainingType.BaseType;
```
ViktorHofer, line 90: Please avoid var on the left hand side here and use the concrete type instead.

src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoaderFactory.cs
```diff
@@ -8,7 +8,14 @@ namespace Microsoft.DotNet.ApiSymbolExtensions
     /// </summary>
     public sealed class AssemblySymbolLoaderFactory : IAssemblySymbolLoaderFactory
     {
+        public AssemblySymbolLoaderFactory(bool includeInternals = false)
```
ViktorHofer, line 11: Please add triple slash documentation for the introduced ctor.

src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoader.cs
```diff
@@ -39,10 +39,11 @@ public class AssemblySymbolLoader : IAssemblySymbolLoader
         public const string AssemblyReferenceNotFoundErrorCode = "CP1002";
 
         /// <inheritdoc />
-        public AssemblySymbolLoader(bool resolveAssemblyReferences = false)
+        public AssemblySymbolLoader(bool resolveAssemblyReferences = false, bool includeInternals = false)
```
ViktorHofer, line 42: Please remove the inheritdoc above (which is wrong) and document the ctor with its parameters?

src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoaderFactory.cs
ericstj: I'll fix the doc issues for the API I'm touching but this codebase is not consistent with docs.  Many public APIs are missing docs and these projects aren't even set up to produce doc files.  If you wish to keep this project documented then docs XML should be enabled so that any missing or out-of-sync docs will show up as compiler warnings that devs can resolve independently.  That should save both of us significant time in PR review.

src/GenAPI/Microsoft.DotNet.GenAPI/IMethodSymbolExtensions.cs
ericstj: I intentionally indented here to show these were the two operands to the ternary operator above.  It made it more readable to me (and Format Document didn't correct it), but I'm fine to remove it.

src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoaderFactory.cs
ViktorHofer: Thank you. Yes, we should turn on documentation file generation and make sure that CS1591 isn't suppressed.

ViktorHofer (approved):

src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoaderFactory.cs
ericstj: There is already an issue.  I was just moving this.

src/GenAPI/Microsoft.DotNet.GenAPI/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,105 @@
...
+            // compare parameter types
+            for (int i = 0; i < method.Parameters.Length; i++)
+            {
+                if (!method.Parameters[i].Type.Equals(otherMethod.Parameters[i].Type, SymbolEqualityComparer.Default))
+                {
+                    return false;
+                }
+            }
+
+            // TODO: GenAPI does not currently preserve __arglist as a parameter.
```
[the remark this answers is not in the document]
ericstj: https://github.com/dotnet/sdk/issues/31196  -- note this code was just moved in this PR.

PR #32302, opened 2023-05-05
Remove template runtimeconfig

### LINKED ISSUE #32300: Can't run SDK tests in VS
### Describe the bug
When loading up the SDK sln and trying to execute tests from test explorer I see an error:
```
Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Testhost process for source(s) 'C:\src\dotnet\sdk\artifacts\bin\Microsoft.DotNet.ApiSymbolExtensions.Tests\Debug\net8.0\Microsoft.DotNet.ApiSymbolExtensions.Tests.dll' exited with error: It's invalid to use both `rollForward` and one of `rollForwardOnNoCandidateFx` or `applyPatches` in the same runtime config.
Invalid runtimeconfig.json [C:\src\dotnet\sdk\artifacts\bin\Microsoft.DotNet.ApiSymbolExtensions.Tests\Debug\net8.0\Microsoft.DotNet.ApiSymbolExtensions.Tests.runtimeconfig.json] [C:\src\dotnet\sdk\artifacts\bin\Microsoft.DotNet.ApiSymbolExtensions.Tests\Debug\net8.0\Microsoft.DotNet.ApiSymbolExtensions.Tests.runtimeconfig.dev.json]
. Please check the diagnostic logs for more information.
```

### To Reproduce
Build repo.
Run eng\dogfood.cmd
Open a sln, or slnf, build and try to run tests.

This regression seems to be caused by https://github.com/dotnet/sdk/pull/31957 @JL03-Yue @nagilson 

The `rollforward` value comes from 
https://github.com/dotnet/sdk/blob/003a147e2dfc932efc430ff60f788114f0966510/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets#L49

And for some reason most of the tests set PackAsTool:
https://github.com/dotnet/sdk/blob/003a147e2dfc932efc430ff60f788114f0966510/src/Tests/Directory.Build.targets#L8-L10

Because they are Exes
https://github.com/search?q=repo%3Adotnet%2Fsdk+OutputType+path%3Asrc%2FTests+language%3AXML&type=code&l=XML

Here's where the other rollforward value comes from:
https://github.com/dotnet/sdk/blob/003a147e2dfc932efc430ff60f788114f0966510/src/Tests/Common/runtimeconfig.template.json#L2

I can workaround the bug by deleting that, but not sure what the right fix is.  @dsplaisted

ericstj:
Fix #32300 

https://github.com/dotnet/sdk/pull/31957 made it so all tools will set rollforward, which clashes with the setting added https://github.com/dotnet/sdk/commit/c5c5011e035c2358f92a012b7c6882f0d1a5f403

This clash was causing the runtime to fail to start.  I'm not sure why that doesn't happen in CI - perhaps in that case SDK is not building the tests with the latest targets?  If so, we might need to make a slightly different change (by ensuring `RollForward` is set in `Tests/Directory.Build.targets` until we get an SDK with that change.

### DIFF
--- a/src/Tests/Common/runtimeconfig.template.json
+++ b/src/Tests/Common/runtimeconfig.template.json
@@ -1,3 +1,0 @@
-{
-    "rollForwardOnNoCandidateFx": 2
-}
--- a/src/Tests/Directory.Build.targets
+++ b/src/Tests/Directory.Build.targets
@@ -13,8 +13,6 @@
 
     <!-- Put packages for tests in subfolder so we don't try to sign them -->
     <PackageOutputPath>$(PackageOutputPath)tests\</PackageOutputPath>
-
-    <UserRuntimeConfig>$(MSBuildThisFileDirectory)Common\runtimeconfig.template.json</UserRuntimeConfig>
   </PropertyGroup>
   
   <ItemGroup Condition="'$(IsTestProject)' == 'true' AND '$(OutputType)' == 'Exe' AND '$(TargetFrameworkIdentifier)' == '.NETCoreApp'">


dsplaisted:
Thanks @ericstj.

@baronfel @JL03-Yue This looks like a break that we should add to the tool RollForward breaking change document.  Before we had the `RollForward` property at all, you could make an app roll forward by creating your own `runtimeconfig.template.json` with `rollForwardOnNoCandidateFx` set, as we did in https://github.com/dotnet/sdk/commit/c5c5011e035c2358f92a012b7c6882f0d1a5f403.

If you've done that, when we start setting `RollForward` to true, you'll end up with an invalid runtimeconfig and the app won't run.

dsplaisted (approved):
Thanks a bunch for catching this and investigating and fixing it.

baronfel:
Is it plausible to see if the user has a config and defer to it (or the specific rollforward value in it) if present?

PR #32305, opened 2023-05-05
Make GeneratePublishDependencyFile incremental

### LINKED ISSUE #28661: `dotnet publish` updates `project.deps.json` for a project with a `PackageReference` with `PrivateAssets="All"` even when nothing has changed
### Describe the bug
`dotnet publish` will update the `project.deps.json` file for a project with a `PackageReference` with `PrivateAssets="All"` even when none of the input files have changed.

### To Reproduce
1. `dotnet new console --name test`
2. `cd test`
3. `dotnet add package Newtonsoft.Json`
4. Add `PrivateAssets="All"` to the `PackageReference`
5. `dotnet publish`
6. Wait a minute
7. Run `dotnet publish` again

The second call to `dotnet publish` has updated the `test.deps.json` file in the `publish` directory (notice that the timestamp of the file changes), even though there was no change to the project or any of its inputs. When removing `PrivateAssets="All"` the `.deps.json` file is only updated when there is an actual change (as would be expected).

### Further technical details

<details><summary>dotnet --info</summary>
<p>

```
dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.110
 Commit:    ce0a42998a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.110\

global.json file:
  Not found

Host:
  Version:      6.0.10
  Architecture: x64
  Commit:       5a400c212a

.NET SDKs installed:
  5.0.403 [C:\Program Files\dotnet\sdk]
  5.0.404 [C:\Program Files\dotnet\sdk]
  5.0.413 [C:\Program Files\dotnet\sdk]
  6.0.110 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.17 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.18 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.21 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.22 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.17 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.18 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.21 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.22 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.21 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.15 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.16 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.18 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.21 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.22 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.21 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.21 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.22 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.29 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.30 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.12 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.13 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.10 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```
</p>
</details>

sbomer:
Fixes https://github.com/dotnet/sdk/issues/28661 by making GeneratePublishDependencyFile incremental. The incremental inputs are the same as those used for GenerateBuildDependencyFile.


dsplaisted:
The `It_compresses_single_file_as_directed` test is failing, it looks like because the `EnableCompressionInSingleFile` changes something which is not accounted for in the inputs and outputs that are declared for the GeneratePublishDependencyFile target.

Probably every property / file that is read in the target should be part of the inputs.  However, you can't declare properties as inputs to a target since they don't have a corresponding timestamp to let you know if they changed.  What we do in this type of situation is to create a separate target which hashes all of the inputs and writes them to a file with `WriteOnlyWhenDifferent="True"`, so the timestamp only gets updated when the properties change.  You can see an example of this with the `_GenerateRuntimeConfigurationFilesInputCache` target.

dsplaisted:
I think this change as is would be a regression.  Previously if you changed the value of `EnableCompressionInSingleFile` on the command line, it would always regenerate the deps file.  Now if you only change that value it won't regenerate it and the output will be wrong.

There are probably other properties that would be similarly broken, this was just the one that we happened to have test coverage for.

It sounds like there are also other properties that are already broken in this way.  However I wouldn't want to introduce new broken behavior without having a way to be sure that the impact would be minimal.

We should discuss (probably in the MSBuild / CLI partner sync) whether we should fix the underlying issue.  If not, then we may need to add a bunch of explicit input hashing targets to fix the incremental behavior with command-line specified global properties.

agocke:
> We should discuss (probably in the MSBuild / CLI partner sync) whether we should fix the underlying issue

In particular, I think we need a broad design solution to this problem. Given that this is pervasive in the system, the right message for users, right now, is that you should not expect incrementality of properties.

dsplaisted:
A target can be:

- Not incremental - it always runs for each build that requires it, whether inputs have changed or not
- Correctly incremental - it only runs if inputs have changed
- Incorrectly incremental - Sometimes it doesn't run even if inputs have changed

There are targets which are already incorrectly incremental, and I think it's OK to not fix them until we have better MSBuild incrementality support.  But I don't think we should change targets from "not incremental" to "incorrectly incremental".  If we want to add incrementality to the target, I think we should try to make it correct.

So for this PR, if we do want to make GeneratePublishDependencyFile incremental, I think we should do the hashing to make it correct.

agocke:
@dsplaisted Makes sense. Is there a public writeup of what we need to do here for best practice?

dsplaisted:
> @dsplaisted Makes sense. Is there a public writeup of what we need to do here for best practice?

Not yet.  @rainersigwald has an action item to write that up

dsplaisted (approved):

Issue #32332, opened 2023-05-08
Include known TFM values in NETSDK1129 error when publishing a multi-targeted project


baronfel:
### Is your feature request related to a problem? Please describe.

When a user publishes a multitargeted project without a TFM provided (usually via the `-f` CLI option to `dotnet publish`) they get the following error:

```
The 'Publish' target is not supported without specifying a target framework. The current project targets multiple frameworks, you must specify the framework for the published application.
```

This doesn't tell the user what TFMs are valid, so the user has to then go check the project file/Directory.Build.props/etc.

### Describe the solution you'd like
This error message should emit `$(TargetFrameworks)` in some way so that users know which TFMs are available to publish for:

```
The 'Publish' target is not supported without specifying a target framework. The current project targets multiple frameworks, you must specify one of the following frameworks in order to publish: net7.0, net8.0.
```

PR #32333, opened 2023-05-08
add known TFMs to multitarget-publish error message

baronfel:
Draft fix for #32332 that includes TargetFrameworks values as a comma-separated list in the reported error message.

It looks like this:

image


nagilson (approved):
Thanks I love it

PR #32371, opened 2023-05-09
Handle Uri-less Realms during basic auth, and fix auth lookups for hosts with ports specified

baronfel:
Fixes https://github.com/dotnet/sdk-container-builds/issues/314


baronfel:
@vlada-shubina the test I made relies on docker-compose (for easier local testing), but compose isn't available on the ubuntu runner. Do you know what we can do to request that? or should I just do the long `docker` command that is the equivalent of the compose file?

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -46,25 +46,66 @@ private static bool TryParseAuthenticationInfo(HttpResponseMessage msg, [NotNull
...
+                _ => false
+            };
+            return result;
+        }
+        return false;
 
-            if (keyValues.TryGetValue("realm", out string? realm) && keyValues.TryGetValue("service", out string? service))
+        static bool TryParseBearerAuthInfo(Dictionary<string, string> authValues, [NotNullWhen(true)] out AuthInfo? authInfo) {
+            if (authValues.TryGetValue("realm", out string? realm) && authValues.TryGetValue("service", out string? service))
             {
```
tmds, line 63:
```suggestion
            if (authValues.TryGetValue("realm", out string? realm))
            {
                string? service = null;
                keyValues.TryGetValue("service", out service);
```

tmds: This fixes an issue with the OpenShift internal registry which doesn't include a `service`.

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -46,25 +46,66 @@ private static bool TryParseAuthenticationInfo(HttpResponseMessage msg, [NotNull
...
-        return false;
+        static bool TryParseBasicAuthInfo(Dictionary<string, string> authValues, Uri requestUri, RegistryName registry, [NotNullWhen(true)] out AuthInfo? authInfo) {
+            if (authValues.TryGetValue("realm", out string? realm))
+            {
+                if (string.IsNullOrWhiteSpace(realm) || !Uri.IsWellFormedUriString(realm, UriKind.Absolute))
+                {
+                    realm = $"{requestUri.Scheme}://{registry}";
+                }
+                string? scope = null;
+                authInfo = new AuthInfo(new Uri(realm), string.Empty, scope);
```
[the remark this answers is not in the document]
tmds: > This still treats the realm as a URI if specified

Right, it should be treated as an opaque string.

> so, I'd expect a tuple (uri.GetLeftPart(UriPartial.Authority), realm) as a key.

You mean as a key for the `AuthHeaderCache`?

What do you think about getting rid of the `static AuthHeaderCache` and instead storing the `AuthenticationHeaderValue` directly in the `AuthHandshakeMessageHandler` as a field?

tmds:
https://github.com/dotnet/sdk/pull/33753 supersedes this PR, and targets the main branch.

baronfel:
Closing in favor of #33753

Issue #30802, opened 2023-02-23
[blazor][net7] dotnet..js file name is missing version info


pavelsavara:
### Actual behavior
The file name of `dotnet..8wdpaewqyh.js` generated by blazor app is missing version info.

![image](https://user-images.githubusercontent.com/271576/220920177-e201acb6-9763-43b0-9ffa-bf3beb2ffb96.png)

### Expected behavior
dotnet.7.0.3.pe05f8nz0f.js

### To Reproduce
- SDK 7.0.103
- install wasm-tools 7.0.3
- create simpla blazor wasm app from template
- add `<WasmBuildNative>true</WasmBuildNative>` to client project file
- `dotnet run -bl -c Release`

### Notes
`DotNetJsVersion` is used to build the name in `_ResolveBlazorWasmOutputs`, but it is set much later in the `ProcessPublishFilesForBlazor`

pavelsavara:
cc @lewing @maraf @TanayParikh @javiercn

mkArtakMSFT:
Ping @pavelsavara

pavelsavara:
It doesn't reproduce for you ?

pavelsavara:
I tested again with 7.0.400-preview.23224.17 and wasm-tools 7.0.5/7.0.100 (SDK 7.0.400-preview.23224).

`dotnet run -c Release` will indeed produce wrong file name `dotnet..o2qge3ran4.js`.

![image](https://user-images.githubusercontent.com/271576/234294719-70d4ff09-c5ee-4be4-8e35-11f3d45a88af.png)

![image](https://user-images.githubusercontent.com/271576/234294867-0ac0455d-d5cc-4c89-aa9c-e635072920c1.png)

It works well with `publish`
![image](https://user-images.githubusercontent.com/271576/234292338-4924d506-eb8b-4675-af7e-1a437c1a57f8.png)

lewing:
is this with an 8.0.1xx sdk?

pavelsavara:
no, this is Net7

pavelsavara:
it's just bad file name but the app works fine

PR #32447, opened 2023-05-11
[release/7.0.4xx] Fix fingerprinting of relinked dotnet.js during build

maraf:
- dotnet.js from runtime pack is replaced by relinked version
- reading `NuGetPackageVersion` metadata doesn't provide a value
- instead we read the version from runtime pack file and than pass it to the task

Closes https://github.com/dotnet/sdk/issues/30802


javiercn (approved):

marcpopMSFT:
@maraf Yes. 7.0.3xx is now in servicing and any changes to that branch require additional approval. 7.0.4xx is fully open for checkin so the simplest action is to retarget to 7.0.4xx and not go through the additional approval process.

Issue #32236, opened 2023-05-03
Support for SxS workload manifests


marcpopMSFT:
This is the initial work in arcade and the resolver to support SxS workload manifests
https://github.com/dotnet/designs/pull/294

### Arcade
- [x] Change workload manifests to be SxS (currently they upgrade)
- [x] Change workload manifests to install in a specific version folder

### Workload resolver (overlap with workload set version)
  - [x] Look under the new path
  - [x] Read the workload set version
  - [x] Read the rollback file version when installed from rollback - https://github.com/dotnet/sdk/pull/33643
  - [x] Support for global.json workload set version (sync with runtime on maybe adding built in support)
    - [x] sync on serialization effort within global.json code 
    - [x] Error if trying to pin to a baseline workload set version (as those packs likely aren't available)
  - [x] TBD do we resolve the latest manifests available or the latest with packs installed <-- yes for now
  - [x] Make sure we don't explode if we both new and old install locations for manifests

### Workload CLI
  - [x] #34032
    - [x] file based install of SxS manifests
    - [x] garbage collection for file based
  - [x] msi-based garbage collection
  - [x] #34280
  
### Miscellaneous
- [ ] Engage Tizen on the planned changes to workloads and how it will work for their workload

PR #32471, opened 2023-05-12
Support side by side workload manifests in SDK resolver

dsplaisted:
This is the first part of updating the SDK resolver to support side-by-side workload manifests (#32236).

It supports reading workload manifests from versioned subfolders, ie `dotnet\sdk-manifests\<Feature Band>\<Manifest ID>\<Manifest Version>`.

In this PR, it will choose the latest installed manifest.  Subsequent changes will support reading workload set versions and using those versions of the manifests.  The start of code to support this by taking in a list of requested manifest versions is in this PR, but that code is not used yet.


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -58,25 +62,40 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
...
+                    _manifestRoots = new[] { userManifestsRoot, dotnetManifestRoot };
+                }
+                else
+                {
+                    _manifestRoots = new[] { dotnetManifestRoot };
+                }
             }
 
             var manifestDirectoryEnvironmentVariable = getEnvironmentVariable(EnvironmentVariableNames.WORKLOAD_MANIFEST_ROOTS);
             if (manifestDirectoryEnvironmentVariable != null)
```
joeloff, line 80: If both `WORKLOAD_MANIFEST_ROOTS` and `WORKLOAD_MANIFEST_IGNORE_DEFAULT_ROOTS` the former will win. Should we consider that an invalid configuration, or should it include both roots?

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -86,21 +88,80 @@ var sdkDirectoryWorkloadManifestProvider
...
+                .BeEmpty();
+        }
+
+        [Fact]
+        public void ItReturnsLatestManifestVersion()
+        {
+            Initialize();
 
-            a.Should().Throw<FileNotFoundException>();
+            CreateMockManifest(_manifestRoot, "5.0.100", "ios", "11.0.1", true);
```
joeloff, line 110: Just to safe, should we have a test that use a mixture of stable/non-stable versions?

```cs
[InlineData("11.0.1", "11.0.2", ...)]
[InlineData("11.0.1", "11.0.2-preview", ...)]
```

and one with SxS feature bands, e.g. `5.0.100-preview3\ios\11.0.2` and `5.0.100-preview4\ios\11.02`?

joeloff (approved):

PR #32538, opened 2023-05-15
Enable EventSourceSupport in ASP.NET Native AOT apps

### LINKED ISSUE #48247: Default EventSourceSupport to true in ASP.NET Native AOT apps
EventSource wasn't supported when publishing an app for NativeAOT in .NET 7. This led to the `EventSourceSupport` feature switch to being [disabled by default in AOT](https://github.com/dotnet/runtime/pull/76000).

In .NET 8 we are getting EventSource to work in Native AOT apps - https://github.com/dotnet/runtime/issues/79241. However, the `EventSourceSupport` feature switch is still disabled, so users won't be able to use things like `dotnet monitor` or `dotnet counters` unless they explicitly opt-in.

We should default `EventSourceSupport=true` in ASP.NET Native AOT'd applications.

This will add size to the default apps (the estimate is around 1 MB), however the tradeoff is worth it, since EventSource is a very common mechanism used for getting information about a running app.

cc @DamianEdwards @davidfowl @noahfalk @agocke @LakshanF @jkotas @MichalStrehovsky

eerhardt:
Fix https://github.com/dotnet/aspnetcore/issues/48247

### DIFF
--- a/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
+++ b/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
@@ -46,18 +46,19 @@ Copyright (c) .NET Foundation. All rights reserved.
     <EnableUnsafeBinaryFormatterSerialization Condition="'$(EnableUnsafeBinaryFormatterSerialization)' == ''">false</EnableUnsafeBinaryFormatterSerialization>
   </PropertyGroup>
 
+  <!-- Default settings for trimmed and aot'd apps. -->
   <PropertyGroup Condition="'$(PublishTrimmed)' == 'true' Or '$(PublishAot)' == 'true'">
-    <!-- Runtime feature defaults to trim unnecessary code -->
     <JsonSerializerIsReflectionEnabledByDefault Condition="'$(JsonSerializerIsReflectionEnabledByDefault)' == ''">false</JsonSerializerIsReflectionEnabledByDefault>
   </PropertyGroup>
 
-  <!--
-    Enable the RequestDelegateGenerator by default for NativeAot apps.
-    This will be enabled for all PublishTrimmed apps once the RequestDelegateGenerator supports all MinimalApi features.
-   -->
+  <!-- Default settings for aot'd apps. -->
   <PropertyGroup Condition="'$(PublishAot)' == 'true'">
+    <!-- Enable the built-in source generators. -->
     <EnableRequestDelegateGenerator Condition="'$(EnableRequestDelegateGenerator)' == ''">true</EnableRequestDelegateGenerator>
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
+
+    <!-- Enable EventSource support because it is disabled by default in aot'd apps. -->
+    <EventSourceSupport Condition="'$(EventSourceSupport)' == ''">true</EventSourceSupport>
   </PropertyGroup>
 
   <!--
--- a/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
+++ b/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
@@ -56,9 +56,6 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Enable the built-in source generators. -->
     <EnableRequestDelegateGenerator Condition="'$(EnableRequestDelegateGenerator)' == ''">true</EnableRequestDelegateGenerator>
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
-
-    <!-- Enable EventSource support because it is disabled by default in aot'd apps. -->
-    <EventSourceSupport Condition="'$(EventSourceSupport)' == ''">true</EventSourceSupport>
   </PropertyGroup>
 
   <!--
--- a/src/WebSdk/Web/Targets/Sdk.Server.targets
+++ b/src/WebSdk/Web/Targets/Sdk.Server.targets
@@ -14,6 +14,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <!-- We can't add it here because the OutputPath and other msbuild properties are not evaluated.-->
   <!--<Import Sdk="Microsoft.NET.Sdk.Publish" Project="ImportPublishProfile.targets" />-->
 
+  <Import Sdk="Microsoft.NET.Sdk.Web.ProjectSystem" Project="..\targets\Microsoft.NET.Sdk.Web.BeforeCommon.targets" />
+
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
 
   <PropertyGroup>
--- a/sdk.sln
+++ b/sdk.sln
@@ -139,6 +139,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Sdk", "Sdk", "{0C6854E8-01E
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Targets", "Targets", "{9446E46C-61A6-4DC0-975F-CF22B8A596A5}"
 	ProjectSection(SolutionItems) = preProject
+		src\WebSdk\ProjectSystem\Targets\Microsoft.NET.Sdk.Web.BeforeCommon.targets = src\WebSdk\ProjectSystem\Targets\Microsoft.NET.Sdk.Web.BeforeCommon.targets
 		src\WebSdk\ProjectSystem\Targets\Microsoft.NET.Sdk.Web.DefaultItems.props = src\WebSdk\ProjectSystem\Targets\Microsoft.NET.Sdk.Web.DefaultItems.props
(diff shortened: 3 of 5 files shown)


src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
```diff
@@ -46,18 +46,19 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-    This will be enabled for all PublishTrimmed apps once the RequestDelegateGenerator supports all MinimalApi features.
-   -->
+  <!-- Default settings for aot'd apps. -->
   <PropertyGroup Condition="'$(PublishAot)' == 'true'">
+    <!-- Enable the built-in source generators. -->
     <EnableRequestDelegateGenerator Condition="'$(EnableRequestDelegateGenerator)' == ''">true</EnableRequestDelegateGenerator>
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
+
+    <!-- Enable EventSource support because it is disabled by default in aot'd apps. -->
+    <EventSourceSupport Condition="'$(EventSourceSupport)' == ''">true</EventSourceSupport>
```
eerhardt, line 61: Unfortunately this is too late. The NativeAOT targets have already defaulted this to `false` by now.

MichalStrehovsky: Would somewhere around here work better?

https://github.com/dotnet/sdk/blob/4b18a3ca87269fce6b915586de2e7e7ae5d80e48/src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets#L97-L99

agocke:
@LakshanF Can you take a look?

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
eerhardt: Unfortunately that is still too late. (It is even later than where I put it above).

https://github.com/dotnet/sdk/blob/d86a7414a36eab675c7a92cba10c81c262a9c4e3/src/WebSdk/Web/Targets/Sdk.Server.targets#L29-L31

Publish gets imported after Web.ProjectSystem.

eerhardt:
@javiercn - You've done a decent amount of SDK work. Can you take a look? I'm not 100% sure I'm doing this right. Or what would be a better way.

MichalStrehovsky:
> @javiercn - You've done a decent amount of SDK work. Can you take a look? I'm not 100% sure I'm doing this right. Or what would be a better way.

Just throwing one option:

https://github.com/dotnet/sdk/blob/ce5131990f41575a7f2c63b7e839116ea2123a1e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L157-L169

We could introduce a `_ProjectTypeRequiresEventSourceSupport` property that, if set to true, would prevent defaulting this to `false`.

eerhardt:
> We could introduce a _ProjectTypeRequiresEventSourceSupport property that, if set to true, would prevent defaulting this to false.

If we are considering changing the NativeAOT msbuild logic, another option would be to not "default" the `EventSourceSupport` property to `false` until a Target runs. This is similar to how `TrimMode` gets defaulted in `PrepareForILLink`:

https://github.com/dotnet/runtime/blob/b066a16f3af015be3f457430aeddfd7f24514c0a/src/tools/illink/src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets#L181-L215

Doing it this way means we don't have to worry about ordering before/after files get imported.

We could move this line:

https://github.com/dotnet/runtime/blob/6f97718bc344ef347c7a8b2e90ccff9dacdc5ed5/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L45

to be inside

https://github.com/dotnet/runtime/blob/b066a16f3af015be3f457430aeddfd7f24514c0a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L127-L128

MichalStrehovsky:
We need to set it before it's used - the use is not in a target:

https://github.com/dotnet/sdk/blob/ce5131990f41575a7f2c63b7e839116ea2123a1e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L472-L475

eerhardt:
> We need to set it before it's used - the use is not in a target:

Yeah, good point. I forgot about that part.

> We could introduce a _ProjectTypeRequiresEventSourceSupport property that, if set to true, would prevent defaulting this to false.

The naming isn't great since ASP.NET doesn't "require" event source support. We just want it by default. I also don't like using `_` properties across layers like this since they are supposed to be "private".

I'm fine with getting some logic to run before the NET.SDK's logic (what I've done in the current PR). So unless there is strong pushback to the current approach, my suggestion is to use this approach.

agocke (approved):
This looks good to me. I think there's an open question of whether or not we want to refactor the Native AOT property setting as well. Right now almost all the feature switch logic is in one place in the SDK -- it's a little unfortunate that Native AOT has its own setting inside the Native AOT targets. Now that Native AOT is officially supported, there may be a cleaner end-to-end solution here.

MichalStrehovsky:
> Now that Native AOT is officially supported, there may be a cleaner end-to-end solution here.

Agreed. The SDK can become a little bit more aware of `PublishAot`. The NativeAOT targets need to be looked at holistically. But that's not a task for this PR.

agocke:
@eerhardt OK to merge?

eerhardt:
I was hoping to get a review from someone in ASP.NET.

@captainsafia @javiercn @amcasey @mitchdenny @halter73 - any thoughts?

amcasey:
> I was hoping to get a review from someone in ASP.NET.
> 
> @captainsafia @javiercn @amcasey @mitchdenny @halter73 - any thoughts?

What were you looking for feedback on?  I thought the idea of re-enabling `EventSourceSupport` had been discussed at length in the various AOT forums.

amcasey:
I guess my one question would be, how does this interact with the metrics work that @JamesNK is doing?  Does it make it possible to consume the new metrics in AOT'd apps?

src/WebSdk/Web/Targets/Sdk.Server.targets
```diff
@@ -14,6 +14,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <!-- We can't add it here because the OutputPath and other msbuild properties are not evaluated.-->
   <!--<Import Sdk="Microsoft.NET.Sdk.Publish" Project="ImportPublishProfile.targets" />-->
 
+  <Import Sdk="Microsoft.NET.Sdk.Web.ProjectSystem" Project="..\targets\Microsoft.NET.Sdk.Web.BeforeCommon.targets" />
```
halter73, line 17: Out of curiosity, why a new target? Why not do this inline like AddRazorSupportForMvc?

halter73 (approved):
LGTM

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
```diff
@@ -46,16 +46,14 @@ Copyright (c) .NET Foundation. All rights reserved.
...
 
+  <!-- Default settings for trimmed and aot'd apps. -->
   <PropertyGroup Condition="'$(PublishTrimmed)' == 'true' Or '$(PublishAot)' == 'true'">
-    <!-- Runtime feature defaults to trim unnecessary code -->
     <JsonSerializerIsReflectionEnabledByDefault Condition="'$(JsonSerializerIsReflectionEnabledByDefault)' == ''">false</JsonSerializerIsReflectionEnabledByDefault>
   </PropertyGroup>
 
-  <!--
-    Enable the RequestDelegateGenerator by default for NativeAot apps.
-    This will be enabled for all PublishTrimmed apps once the RequestDelegateGenerator supports all MinimalApi features.
```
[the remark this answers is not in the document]
eerhardt: I think it is not a valuable comment going forward. If we want to enable it for PublishTrimmed apps, we should have an issue tracking it.

https://github.com/dotnet/aspnetcore/issues/48416

src/WebSdk/Web/Targets/Sdk.Server.targets
eerhardt: It is setting up some "infrastructure" for a place where we can add more settings that need to be defined between the .csproj and Microsoft.NET.Sdk's targets. A lot of the value is defining intent from the name of the file (BeforeCommon.targets). We have the same thing in the NET.Sdk here:

https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets

This way everything doesn't get piled into `Sdk.targets`.

javiercn (approved):
LGTM

eerhardt:
> What were you looking for feedback on?

The new MSBuild structure here. Specifically how to inject a `BeforeCommon.targets` file that is imported between the user's .csproj and Microsoft.NET.Sdk's .targets file.

>  how does this interact with the metrics work that @JamesNK is doing? Does it make it possible to consume the new metrics in AOT'd apps?

Metrics can get pumped through EventSource so external tools can see them. See https://github.com/dotnet/runtime/blob/7a119156cbe958c5c34098434e84010526fac5f1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/MetricsEventSource.cs#L14-L44. This change means that consumers outside of an AOT'd app (for example dotnet counters) can consume these new metrics now too.

PR #32567, opened 2023-05-16
workload from rollback file for tests

pavelsavara:


maraf:
> Is this covered by this change: https://github.com/dotnet/sdk/pull/32107

That is only for AOT jobs, right?
Pavel is trying it here also for non-AOT, so that it gets the latest runtime pack.

maraf:
Closing in favor of https://github.com/dotnet/sdk/pull/32745

PR #32594, opened 2023-05-17
Containers/CreateNewImage: don't change the user to root when ContainerUser is empty.

tmds:
When a rootless base image is used without specifying a ContainerUser, the image user was changed to User="". This causes the application image to run as the 'root' user.

@baronfel ptal.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
@@ -80,7 +80,7 @@ public static class ContainerBuilder
             // ports are validated by System.CommandLine API
             imageBuilder.ExposePort(number, type);
         }
-        if (containerUser is { } user)
+        if (containerUser is { Length: > 0 } user)
         {
             imageBuilder.SetUser(user);
         }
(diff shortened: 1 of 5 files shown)


tmds:
This doesn't affect the Microsoft rootless base images because the `ContainerUser` gets set here:

https://github.com/dotnet/sdk/blob/1e3e940c5663d3a6eb0f896c213e8fd202eab7fb/src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets#L107-L112

tmds:
> the Microsoft rootless base images

Correction: none of the Microsoft base images are rootless. The user is expected to add the `USER ` directive (like the sdk tooling is doing here).

baronfel:
Interesting - this makes sense to me. Can you think of a way to test? An end-to-end test with specific properties would let you load and check the manifest's User value.

tmds:
I'll look into adding a test for this next week.

baronfel (changes requested):
Requesting changes so it doesn't get merged without a test :)

tmds:
@baronfel I've added a test that verifies `CreateNewImage` doesn't replace the rootless user from a base image. ptal.

baronfel (approved):
Thanks for adding a test to keep this green going forward :+1:

PR #32606, opened 2023-05-17
Containers/Layer: trim x-bit based on filesystem permissions on Unix.

tmds:
@baronfel ptal.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/Layer.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Layer.cs
@@ -7,6 +7,7 @@ using System.Globalization;
 using System.IO.Compression;
 using System.Security.Cryptography;
 using System.Text;
+using System.IO.Enumeration;
 
 namespace Microsoft.NET.Build.Containers;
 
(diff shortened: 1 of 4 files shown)


src/Containers/Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -44,111 +45,78 @@ public static Layer FromDescriptor(Descriptor descriptor)
     }
 
     public static Layer FromDirectory(string directory, string containerPath, bool isWindowsLayer)
```
tmds, line 47: I've removed the `FromFiles` method. The implementation enumerates recursively over both directories and files using `FileSystemEnumerable` which doesn't work for `FromFiles`.

Using `FileSystemEnumerable` we will encounter parent directories before their children which gives a straight-forward way to add the parent directory entries to the archive.

src/Containers/Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -44,111 +45,78 @@ public static Layer FromDescriptor(Descriptor descriptor)
...
-
-    public static Layer FromFiles(IEnumerable<(string path, string containerPath)> fileList, bool isWindowsLayer)
     {
         long fileSize;
         Span<byte> hash = stackalloc byte[SHA256.HashSizeInBytes];
         Span<byte> uncompressedHash = stackalloc byte[SHA256.HashSizeInBytes];
 
-        // this factory helps us creating the Tar entries with the right attributes
-        PaxTarEntry CreateTarEntry(TarEntryType entryType, string containerPath)
+        // Docker treats a COPY instruction that copies to a path like `/app` by
```
tmds, line 53: I moved the operations that operate on the `containerPath` together to the start of the method.

src/Containers/Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -44,111 +45,78 @@ public static Layer FromDescriptor(Descriptor descriptor)
...
-            }
-
-            return containerPath;
-        }
-
-        // Ensures that all directory entries for the given segments are created within the tar.
-        var directoryEntries = new HashSet<string>();
-        void EnsureDirectoryEntries(TarWriter tar,
-            IReadOnlyList<string> filePathSegments)
+        var entryAttributes = new Dictionary<string, string>();
```
tmds, line 75: We can allocate this once and pass it to every entry.

src/Containers/Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -44,111 +45,78 @@ public static Layer FromDescriptor(Descriptor descriptor)
...
-
-                        writer.WriteEntry(entry);
+                        WriteTarEntryForFile(writer, item.file, item.containerPath, entryAttributes);
                     }
 
                     // Windows layers need a Hives folder, we do not need to create any Registry Hive deltas inside
                     if (isWindowsLayer)
                     {
-                        var entry = CreateTarEntry(TarEntryType.Directory, "Hives/");
+                        var entry = new PaxTarEntry(TarEntryType.Directory, "Hives/", entryAttributes);
```
tmds, line 119: I guess the trailing slash may not be needed here, but it doesn't cause any harm either.

src/Containers/Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -163,6 +131,43 @@ string SanitizeContainerPath(string containerPath)
...
+                    };
+                    writer.WriteEntry(entry);
+                }
+
+                static UnixFileMode DetermineFileMode(FileSystemInfo file)
+                {
+                    const UnixFileMode nonExecuteMode = UnixFileMode.UserRead | UnixFileMode.UserWrite |
+                                                        UnixFileMode.GroupRead |
+                                                        UnixFileMode.OtherRead;
+                    const UnixFileMode executeMode = nonExecuteMode | UnixFileMode.UserExecute | UnixFileMode.GroupExecute | UnixFileMode.OtherExecute;
```
tmds, line 164: When we add a property to control these permissions in https://github.com/dotnet/sdk-container-builds/issues/399.
That property could control `nonExecuteMode`, and we can derive `executeMode` from it by adding x-bit where there are r-bits.

(This matches the note in https://github.com/dotnet/sdk-container-builds/issues/399#issuecomment-1489777710).

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/LayerEndToEndTests.cs
```diff
@@ -42,7 +42,7 @@ public void SingleFileInFolder()
         VerifyDescriptorInfo(l);
 
         var allEntries = LoadAllTarEntries(l.BackingFile);
-        Assert.True(allEntries.TryGetValue("app/", out var appEntry) && appEntry.EntryType == TarEntryType.Directory, "Missing app directory entry");
+        Assert.True(allEntries.TryGetValue("app", out var appEntry) && appEntry.EntryType == TarEntryType.Directory, "Missing app directory entry");
```
tmds, line 45: Zip entries have trailing slashes to denote they are directory entries, but tar files don't need this suffix.
I removed it in the implementation, and updated the test accordingly.

baronfel:
Thanks for the detailed explanation. Thanks to the tests I have confidence this is fine for Linux containers - however I'll want to run the commented-out Windows tests on my box tomorrow.

baronfel:
I've been having trouble getting the SDK to build on my Windows machine so I haven't been able to validate this.

baronfel (approved):
I was able to test Windows containers, and am satisfied. Thank you @tmds!

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
```diff
@@ -405,16 +405,16 @@ public async Task CanPackageForAllSupportedContainerRIDs(string dockerPlatform,
...
             DockerRegistryManager.RuntimeBaseImage,
-            DockerRegistryManager.Net8PreviewImageTag,
+            isWin ? DockerRegistryManager.Net8PreviewWindowsSpecificImageTag : DockerRegistryManager.Net8PreviewImageTag,
             rid,
             ToolsetUtils.GetRuntimeGraphFilePath(),
             cancellationToken: default).ConfigureAwait(false);
         Assert.NotNull(imageBuilder);
 
-        Layer l = Layer.FromDirectory(publishDirectory, "/app", false);
+        Layer l = Layer.FromDirectory(publishDirectory, isWin ? "C:\\app" : "/app", isWin);
```
baronfel, line 417: This has been broken for a long time :)

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
```diff
@@ -405,16 +405,16 @@ public async Task CanPackageForAllSupportedContainerRIDs(string dockerPlatform,
 
         // Build the image
         Registry registry = new(ContainerHelpers.TryExpandRegistryToUri(DockerRegistryManager.BaseImageSource));
-
+        var isWin = rid.StartsWith("win");
         ImageBuilder? imageBuilder = await registry.GetImageManifestAsync(
             DockerRegistryManager.RuntimeBaseImage,
-            DockerRegistryManager.Net8PreviewImageTag,
+            isWin ? DockerRegistryManager.Net8PreviewWindowsSpecificImageTag : DockerRegistryManager.Net8PreviewImageTag,
```
baronfel, line 411: This is required because the dotnet containers have changed to no longer include windows images in the manifest list.

PR #32685, opened 2023-05-19
add two artifacts-output-related-telemetry properties to emitted events

baronfel:
For the Artifacts output path feature we added two new telemetry data points
* was the feature enabled, and
* what mode was the feature using (explicit, implicit, project-local, etc)

Those data points were being collected, but filtered out because we missed another filtering step. This PR

* adds the data points to the filter, then
* fixes a typo in the already-collected `UseAppHost` telemetry value, and
* finally converts the entire `targetframeworkeval` event to use the new declarative hashing feature of the `AllowEmptyTelemetry` task so that we don't need the additional filter in the MSBuild telemetry logger. This change is the most impactful, and we should discuss it.
* I also flipped the defaults to `opt-in` hashing for the telemetry task. This involved annotating the existing usages with Hash=false to keep behavior the same.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
+++ b/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
@@ -27,6 +27,8 @@ namespace Microsoft.DotNet.Tools.MSBuild
         internal const string SelfContainedTelemetryPropertyKey = "SelfContained";
         internal const string UseApphostTelemetryPropertyKey = "UseApphost";
         internal const string OutputTypeTelemetryPropertyKey = "OutputType";
+        internal const string UseArtifactsOutputPropertyKey = "UseArtifactsOutput";
+        internal const string ArtifactsPathLocationTypePropertyKey = "ArtifactsPathLocationType";
 
         public MSBuildLogger()
         {
@@ -95,7 +97,9 @@ namespace Microsoft.DotNet.Tools.MSBuild
                     RuntimeIdentifierTelemetryPropertyKey,
                     SelfContainedTelemetryPropertyKey,
                     UseApphostTelemetryPropertyKey,
-                    OutputTypeTelemetryPropertyKey
+                    OutputTypeTelemetryPropertyKey,
+                    UseArtifactsOutputPropertyKey,
+                    ArtifactsPathLocationTypePropertyKey
                 })
                 {
                     if (args.Properties.TryGetValue(key, out string value))
--- a/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
+++ b/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
@@ -22,14 +22,6 @@ namespace Microsoft.DotNet.Tools.MSBuild
         internal const string PublishPropertiesTelemetryEventName = "PublishProperties";
         internal const string ReadyToRunTelemetryEventName = "ReadyToRun";
 
-        internal const string TargetFrameworkVersionTelemetryPropertyKey = "TargetFrameworkVersion";
-        internal const string RuntimeIdentifierTelemetryPropertyKey = "RuntimeIdentifier";
-        internal const string SelfContainedTelemetryPropertyKey = "SelfContained";
-        internal const string UseApphostTelemetryPropertyKey = "UseApphost";
-        internal const string OutputTypeTelemetryPropertyKey = "OutputType";
-        internal const string UseArtifactsOutputPropertyKey = "UseArtifactsOutput";
-        internal const string ArtifactsPathLocationTypePropertyKey = "ArtifactsPathLocationType";
-
         public MSBuildLogger()
         {
             try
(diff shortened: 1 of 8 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets
```diff
@@ -122,13 +122,13 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <TFTelemetry Include="TargetFrameworkVersion" Value="$([MSBuild]::Escape('$(TargetFrameworkMoniker)'))" Hash="true" />
+      <TFTelemetry Include="RuntimeIdentifier" Value="$(RuntimeIdentifier)" Hash="true" />
       <TFTelemetry Include="SelfContained" Value="$(SelfContained)" />
-      <TFTelemetry Include="UseApphost" Value="$(UseApphost)" />
+      <TFTelemetry Include="UseAppHost" Value="$(UseAppHost)" />
       <TFTelemetry Include="OutputType" Value="$(OutputType)" />
-      <TFTelemetry Include="UseArtifactsOutput" Value="$(UseArtifactsOutput)" />
-      <TFTelemetry Include="ArtifactsPathLocationType" Value="$(_ArtifactsPathLocationType)" />
+      <TFTelemetry Include="UseArtifactsOutput" Value="$(UseArtifactsOutput)"  />
+      <TFTelemetry Include="ArtifactsPathLocationType" Value="$(_ArtifactsPathLocationType)" Hash="true"/>
```
baronfel, line 131: We can opt a data point into hashing with `Hash="true"` here. The valid values for SelfContained, UseAppHost, OutputType, and UseArtifactsOutput are well-known, but being MSBuild they could be anything. I think that means that we should defensively hash them as well. @dsplaisted what are your thoughts?

src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
```diff
@@ -93,23 +87,7 @@ internal static void FormatAndSend(ITelemetry telemetry, TelemetryEventArgs args
...
                 var newEventName = $"msbuild/{TargetFrameworkTelemetryEventName}";
-                Dictionary<string, string> maskedProperties = new Dictionary<string, string>();
-
-                foreach (var key in new[] {
-                    TargetFrameworkVersionTelemetryPropertyKey,
-                    RuntimeIdentifierTelemetryPropertyKey,
-                    SelfContainedTelemetryPropertyKey,
-                    UseApphostTelemetryPropertyKey,
-                    OutputTypeTelemetryPropertyKey
-                })
```
baronfel, line 104: This is the filter we missed before - having this meant we both needed to collect data points declaratively in MSBuild code, but then also sync the fields back here. This seems fragile, so I think we should move to the built-in hashing I added a bit ago.

baronfel: Devil's Advocate - @dsplaisted wondered if there was a potential for new data points to 'leak' out unexpectedly without this filter. I think this risk is fairly small, as we are already constraining the properties in MSBuild props, but we should consider this in more detail.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets
```diff
@@ -122,13 +122,13 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Target Name="_CollectTargetFrameworkForTelemetry" AfterTargets="_CheckForUnsupportedTargetFramework">
     <ItemGroup>
-      <TFTelemetry Include="TargetFrameworkVersion" Value="$([MSBuild]::Escape('$(TargetFrameworkMoniker)'))" />
-      <TFTelemetry Include="RuntimeIdentifier" Value="$(RuntimeIdentifier)" />
+      <TFTelemetry Include="TargetFrameworkVersion" Value="$([MSBuild]::Escape('$(TargetFrameworkMoniker)'))" Hash="true" />
+      <TFTelemetry Include="RuntimeIdentifier" Value="$(RuntimeIdentifier)" Hash="true" />
       <TFTelemetry Include="SelfContained" Value="$(SelfContained)" />
-      <TFTelemetry Include="UseApphost" Value="$(UseApphost)" />
+      <TFTelemetry Include="UseAppHost" Value="$(UseAppHost)" />
```
baronfel, line 128: Here's the typo - every other usage in the SDK is looking for `UseAppHost`, not `UseApphost`.

src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs
baronfel: We mitigated this risk by defaulting all collected data points to hashing by default. Individual data points must set the `Hash=false` metadata to disable this behavior.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets
baronfel: We decided to hash everything by default.

baronfel:
Closing in favor of more targeted fix.

dsplaisted:
@baronfel Should we bring back some of the changes here for .NET 9?

baronfel:
We should attempt to I believe - it's a safer, more declarative way of doing the telemetry logging, and it prevents the kinds of gaps we had in our telemetry for ArtifactsPath.

PR #32687, opened 2023-05-19
Update SDK in global.json to upgrade WebAssembly tests

maraf:
- Update global.json to point to newer SDK
- Update Blazor tests to expect webcil on by default
- Regenerate baselines for the rest of the Blazor tests
- Temporal condition in Blazor SDK to turn webcil off for .NET < 8. I'll update Wasm SDK in the runtime repo and than send another PR to remove it from here


javiercn (approved):
Looks great!

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -51,6 +51,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DebuggerSupport Condition="'$(DebuggerSupport)' == '' and '$(Configuration)' != 'Debug'">false</DebuggerSupport>
     <BlazorCacheBootResources Condition="'$(BlazorCacheBootResources)' == ''">true</BlazorCacheBootResources>
     <WasmFingerprintDotnetJs>true</WasmFingerprintDotnetJs>
+    
+    <_TargetingNET80OrLater>$([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', '8.0'))</_TargetingNET80OrLater>
+    <WasmEnableWebcil Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' or '$(_TargetingNET80OrLater)' != 'true'">false</WasmEnableWebcil>
```
javiercn, line 56: You need to check if the user has explicitly set `WasmEnableWebcil`, as otherwise it can't be overridden

lewing:
@marcpopMSFT this has the changes that would be needed with the workload update patch

PR #32688, opened 2023-05-19
[Blazor] Support loading the Blazor WebAssembly SDK when targeting net8.0-browser

javiercn:
* As part of Blazor united, the app project targets the web SDK and multitargets to the net8.0 and net8.0-browser TFMs.

```xml

  <PropertyGroup>
    <TargetFrameworks>net8.0;net8.0-browser1.0</TargetFrameworks>
  </PropertyGroup>

  <ItemGroup Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' == 'browser'">
    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="8.0.0-preview.4.23260.4" />
  </ItemGroup>

</Project>
```

For this to work, we need to update the Web SDK to import the Blazor SDK instead of the other SDKs when targeting net8.0-browser.

### DIFF
--- a/src/WebSdk/Web/Sdk/Sdk.props
+++ b/src/WebSdk/Web/Sdk/Sdk.props
@@ -12,9 +12,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <!-- Imports the .NET, Razor, and other web related SDKs on projects that target server scenarios. -->
-  <Import Project="../Targets/Sdk.Server.props" Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' != 'browser'" />
+  <Import Project="..\Targets\Sdk.Server.props" Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' != 'browser'" />
 
   <!-- Delegates to the Blazor Webassembly SDK on projects that target browser scenarios -->
-  <Import Project="../Targets/Sdk.Browser.props" Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' == 'browser'" />
+  <Import Project="..\Targets\Sdk.Browser.props" Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' == 'browser'" />
 
 </Project>
--- a/src/WebSdk/Web/Sdk/Sdk.props
+++ b/src/WebSdk/Web/Sdk/Sdk.props
@@ -13,7 +13,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!--
       Indicate to other targets that Microsoft.NET.Sdk.Web is being used.
     -->
-  <PropertyGroup>
+  <PropertyGroup Condition="'$([MSBuild]::GetTargetPlatformIdentifier($TargetFramework))' != 'browser'">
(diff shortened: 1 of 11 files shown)


src/WebSdk/Web/Sdk/Sdk.props
```diff
@@ -23,19 +23,21 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DebugSymbols Condition="'$(DebugSymbols)' == ''">true</DebugSymbols>
   </PropertyGroup>
 
-  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
+  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' != 'browser'" />
```
vijayrkn, line 26: Do you want to store the value of the condition in a property so that the condition check in each of these will look simpler `[MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' != 'browser'`?

src/WebSdk/Web/Sdk/Sdk.props
```diff
@@ -23,19 +23,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
 
-  <Import Sdk="Microsoft.NET.Sdk.Razor" Project="Sdk.props" />
+  <Import Sdk="Microsoft.NET.Sdk.BlazorWebAssembly" Project="Sdk.props" Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' == 'browser'" />
 
   <Import Sdk="Microsoft.NET.Sdk.Web.ProjectSystem" Project="Sdk.props" />
 
   <Import Sdk="Microsoft.NET.Sdk.Publish" Project="Sdk.props" />
 
-  <ItemGroup Condition="'$(DisableImplicitFrameworkReferences)' != 'true' And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' >= '3.0'">
+  <ItemGroup Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' != 'browser' And '$(DisableImplicitFrameworkReferences)' != 'true' And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(_TargetFrameworkVersionWithoutV)' >= '3.0'">
```
vijayrkn, line 36: Do you want to store the value of the condition in a property so that the condition check in each of these will look simpler [MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' != 'browser'?

javiercn: @vijayrkn I was thinking of breaking the contents into two separate files so that way we only check the condition once, and we don't have to litter it over each individual statement. Will that work for you?

vijayrkn: Ya, that would make it much easier to read/understand what gets imported/conditioned based on the TFM.

MackinnonBuck (approved):
Looks good to me after @vijayrkn's feedback gets addressed

src/WebSdk/Web/Sdk/Sdk.props
javiercn: Ok, I've updated the PR and hopefully will go through the build.

vijayrkn (approved):

vijayrkn:
> @vijayrkn do you mind taking another look and making sure you are happy with the results?

Thanks for making this change. This makes it much easier to differentiate the 2 cases.

PR #32698, opened 2023-05-19
Update MSBuildLocator version

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/3356

Source-build only builds once version of MSBuildLocator and that is 1.5.5 (used by format and roslyn).  It would be nice if we could centralize on a single version so that we can avoid having to carry around reference packages for multiple versions.

I am not sure who are the correct folks to code review this.  @nagilson - you were the last one to update this version.


nagilson (approved):
This should be good! 

I also talked to @rainersigwald very briefly about this -- @YuliiaKovalova will be the new MSBuild Locator expert for future reference though she is just onboarding to it :)

PR #32716, opened 2023-05-22
Containers.IntegrationTests: support running with podman.

tmds:
@baronfel ptal.

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ContainerCli.cs
+++ b/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ContainerCli.cs
@@ -1,0 +1,73 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Linq;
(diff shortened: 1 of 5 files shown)


baronfel (approved):
Straightforward, mechanical change, thanks for making us more compatible.

Is it correct to assume that we can rely on you for validation of podman support going forward? We don't have it available on our CI pipelines I believe.

tmds:
> Is it correct to assume that we can rely on you for validation of podman support going forward? We don't have it available on our CI pipelines I believe.

I haven't done this yet. If possible, it would be nice if we can add it to the `sdk` CI pipelines by using a Fedora/RHEL/CentOS/Alma/... base image. If that is not feasible, I'll definitely look into setting up an internal CI to ensure `podman` keeps working.

baronfel:
> using a Fedora/RHEL/CentOS/Alma/... base image

This is an interesting idea, especially since .NET supports Fedora explicitly in the [OS support matrix](https://github.com/dotnet/core/blob/main/release-notes/7.0/supported-os.md). @marcpopMSFT thoughts on adding any leg(s) for the distros mentioned?

PR #32751, opened 2023-05-23
Containers: rename "Docker daemon" to "local registry".

tmds:
The container tooling supports working with the local registry provided by the the 'docker' cli.

The implementation and messages refer to this as the "Docker/local daemon". This changes to use "local registry" instead. This makes the naming appropriate also for a daemonless implementation like 'podman'.

@baronfel ptal.

This PR is about renaming things. I will make a separate PR that is specifically about adding 'podman' support.


baronfel (approved):
This renaming LTGM - thanks for breaking this into two steps!

baronfel:
One note @tmds - right now this is targeting .NET 8 - did you want your changes for podman support to be in 7.0.4xx (so available around the August time frame?). Most of your other changes have been targeting that branch.

tmds:
It's fine if this is only part of the .NET 8 SDK. That SDK can then still be used to target .NET 7.

PR #32766, opened 2023-05-23
Read workload sets in workload manifest reader

dsplaisted:
Following up on #32471, this is the next part of updating the workload resolver to support workload sets (#32236).

This PR adds support for reading workload sets from `dotnet\sdk-manifests\<Feature Band>\workloadsets\<Workload Set Version>`.  With this PR, the latest available workload set for the current feature band will be used.  Later we will add support for controlling the workload set via global.json and other mechanisms.


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -15,22 +15,24 @@ namespace Microsoft.NET.Sdk.WorkloadManifestReader
 {
```
nagilson, line 15: I think we should add a nullable enable directive here?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -118,7 +124,7 @@ public IEnumerable<ReadableWorkloadManifest> GetManifests()
         public IEnumerable<string> GetManifestDirectories()
```
nagilson, line 124: Should we consider adding the feature band as a parameter to this function? I think that might break the concept of some feature bands not touching others, so perhaps not. But for cleaning these manifests for unused bands I feel like we'll eventually add that.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -206,7 +215,8 @@ void ProbeDirectory(string manifestDirectory)
         private (string? id, string? manifestDirectory) ResolveManifestDirectory(string manifestDirectory)
         {
             string manifestId = Path.GetFileName(manifestDirectory);
-            if (_outdatedManifestIds.Contains(manifestId))
+            if (_outdatedManifestIds.Contains(manifestId) ||
+                manifestId.Equals(WorkloadSetsFolderName, StringComparison.OrdinalIgnoreCase))
```
nagilson, line 219: When/why does this get called with the `WorkloadSetsFolderName`? Is this just a safety measure / defensive programming?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -206,7 +215,8 @@ void ProbeDirectory(string manifestDirectory)
         private (string? id, string? manifestDirectory) ResolveManifestDirectory(string manifestDirectory)
```
nagilson, line 215: Thanks for adding a good docstring here :)

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -292,6 +302,47 @@ private bool IsManifestIdOutdated(string workloadManifestDir)
             return _outdatedManifestIds.Contains(manifestId);
         }
 
+        private Dictionary<string, WorkloadSet> GetAvailableWorkloadSets()
+        {
+            Dictionary<string, WorkloadSet> availableWorkloadSets = new Dictionary<string, WorkloadSet>();
+
+            foreach (var manifestRoot in _manifestRoots.Reverse())
```
nagilson, line 309: Are we doing the .Reverse so that the dotnet directory goes first before the user home directory when picking the newest available workload set?

It looks like we do that above as well. May we just re-order the manifestRoots?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -292,6 +302,47 @@ private bool IsManifestIdOutdated(string workloadManifestDir)
...
+                            if (workloadSet == null)
+                            {
+                                workloadSet = newWorkloadSet;
+                            }
+                            else
+                            {
+                                //  If there are multiple workloadset.json files, merge them
+                                foreach (var kvp in newWorkloadSet.ManifestVersions)
+                                {
+                                    workloadSet.ManifestVersions.Add(kvp.Key, kvp.Value);
```
nagilson, line 330: I think it might be possible for a duplicate to be added here if it was on the disk, would that be a problem? Should we do any validation here?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -292,6 +302,47 @@ private bool IsManifestIdOutdated(string workloadManifestDir)
             return _outdatedManifestIds.Contains(manifestId);
         }
 
+        private Dictionary<string, WorkloadSet> GetAvailableWorkloadSets()
```
nagilson, line 305: Does 'available' mean that the workload set is on the disk and of the same feature band of the running SDK? It could be beneficial to add a docstring for this.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadSet.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static WorkloadSet FromDictionaryForJson(IDictionary<string, string> dictionary, SdkFeatureBand defaultFeatureBand)
+        {
+            var manifestVersions = dictionary
+                .Select(manifest =>
+                {
+                    ManifestVersion manifestVersion;
+                    SdkFeatureBand manifestFeatureBand;
+                    var parts = manifest.Value.Split('/');
+
+                    string manifestVersionString = (parts[0]);
```
nagilson, line 47:
```suggestion
                    string manifestVersionString = parts[0];
```

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadSet.cs
```diff
@@ -0,0 +1,97 @@
...
+#endif
+
+
+namespace Microsoft.NET.Sdk.WorkloadManifestReader
+{
+    public class WorkloadSet
+    {
+        public Dictionary<ManifestId, (ManifestVersion Version, SdkFeatureBand FeatureBand)> ManifestVersions = new Dictionary<ManifestId, (ManifestVersion version, SdkFeatureBand featureBand)>();
+
+        //  TODO: Generate version from hash of manifest versions if not otherwise set
```
nagilson, line 27: Is this TODO still needed? It looks like currently it's never being used and only added based on the manifest folder name.

nagilson (commented):
I'm excited the resolver work is getting done so quickly! I haven't looked at the tests yet or tried it out which I will do in a bit, but here are my comments regarding the code.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/Strings.resx
```diff
@@ -192,4 +192,10 @@
   <data name="SpecifiedManifestNotFound" xml:space="preserve">
     <value>Specified workload manifest was not found: {0}</value>
   </data>
+  <data name="SpecifiedWorkloadSetNotFound" xml:space="preserve">
+    <value>Specified workload set was not found: {0}</value>
```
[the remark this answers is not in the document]
dsplaisted: This should probably change in subsequent PRs.  Right now the workload set version is an optional argument to the constructor which is never actually used.  The next PR will probably involve changing the code to read it from global.json or a "workload install state" folder.  When we know where the value is coming from we can probably generate a more helpful message.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: I don't think we need it specifically in the file, I think it's set at the project level.  This file is compiled into 4 different projects, and I got errors when I didn't have the nullability declarations correct, so it's already enabled for at least some of the projects compiling it.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: Right now it's not needed, the only external use of this method is for the resolver to find the `WorkloadManifest.targets` files in the manifests.

This manifest provider is specific to an SDK version, so if we need the data for other feature bands we might need to create a separate instance for each of them.  We can deal with those changes when we need to make them for clean.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: The `workloadsets` folder is a sibling to the workload manifest folders, but should be ignored when looking for workload manifests, as it stores different information.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: It will throw an exception:

> System.ArgumentException : An item with the same key has already been added. Key: ios

I've added a test for this.

The exception is not very helpful but this is a scenario that may never be used.  I expect all the workload sets we create to just have one json file.  The reason for supporting multiple is to leave the opportunity for something like Tizen to augment a workload set with versions for additional manifests.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadSet.cs
dsplaisted: I expect it to be used in the future.  We will want to report the current workload set version from `dotnet workload --version`, and show the old and new workload set versions when updating workloads.

dsplaisted:
@nagilson Thanks, I've responded to the feedback and added a few more changes.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
nagilson: Thanks, that makes sense and I agree with that. I meant to ask since it _should_ be ignored, in what contexts would this function be called with a `manifestDirectory` that has the foldername `workloadsets`? Because we should never do that 'ideally,' right? Do we expect that to ever happen, or is this a defensive mechanism against it? Either way, seems OK, but I wanted to know.

dsplaisted: This method will actually get called for the `workloadsets` folder.  It needs to get filtered out somewhere, and this method seemed like the best place to do it because then the filtering doesn't need to be duplicated in the multiple places that call this method.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -167,6 +167,241 @@ var sdkDirectoryWorkloadManifestProvider
...
+            var sdkDirectoryWorkloadManifestProvider
+                = new SdkDirectoryWorkloadManifestProvider(sdkRootPath: _fakeDotnetRootDirectory, sdkVersion: "8.0.200", userProfileDir: null);
+
+            GetManifestContents(sdkDirectoryWorkloadManifestProvider)
+                .Should()
+                .BeEquivalentTo("ios: 12.0.1/8.0.200");
+        }
+
+        [Fact]
+        public void ItUsesLatestManifestThatIsNotInWorkloadSet()
```
nagilson, line 234:
```suggestion
        public void ItUsesLatestManifestForWorkloadsNotInWorkloadSet()
```

The old name is a bit ambiguous --> e.g. it can imply that a newer manifest should be used for a workload in the set even if that manifest is not in the set, which is not true.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
nagilson: Ah yes, I see how it happens now. Thank you :)

nagilson (approved):
LGTM, thanks for responding to my questions!

PR #32783, opened 2023-05-24
Containers: add support for using 'podman' without a system 'docker' alias.

tmds:
@baronfel this builds on top of https://github.com/dotnet/sdk/pull/32751, which deals with renaming things.
We can look at this PR once we've reviewed and merged the other PR. That will clear up the diff here.


src/Containers/Microsoft.NET.Build.Containers/Resources/Strings.resx
```diff
@@ -158,7 +158,7 @@
     <comment>{StrBegin="CONTAINER3002: "}</comment>
   </data>
   <data name="DockerProcessCreationFailed" xml:space="preserve">
-    <value>CONTAINER3001: Failed creating docker process.</value>
+    <value>CONTAINER3001: Failed creating {0} process.</value>
     <comment>{StrBegin="CONTAINER3001: "}</comment>
```
baronfel, line 162: Can you also add a comment describing what `{0}` is here for the localization team? Something like "{0} is the name of the command we failed to run, usually 'docker' or 'podman'".

src/Tests/Microsoft.NET.Build.Containers.UnitTests/DockerAvailableUtils.cs
```diff
@@ -7,26 +7,47 @@ namespace Microsoft.NET.Build.Containers.UnitTests;
...
+    public static readonly bool IsAvailable;
+    public static readonly string? Command;
+
+    static DockerCliStatus()
+    {
+        DockerCli cli = new(Console.WriteLine);
+        IsAvailable = cli.IsAvailable();
+        Command = cli.GetCommand();
     }
 }
```
baronfel: neat!

src/Containers/Microsoft.NET.Build.Containers/LocalDaemons/DockerCli.cs
```diff
@@ -56,32 +78,43 @@ public async Task LoadAsync(BuiltImage image, ImageReference sourceReference, Im
...
+                    else
+                    {
+                        // we have errors, turn them into a string and log them
+                        string messages = string.Join(Environment.NewLine, errorProperty.EnumerateArray());
+                        logger($"The daemon server reported errors: {messages}");
+                        return false;
+                    }
+                }
+                case PodmanCommand:
+                    return true;
```
tmds, line 115: I will add a "podman version" invocation here to ensure the tool is installed when this method returns true.

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.Interface.cs
```diff
@@ -48,7 +48,6 @@ partial class CreateNewImage
     /// <summary>
     /// The kind of local registry to use, if any.
     /// </summary>
-    [Required]
```
tmds, line 51: I missed removing this attribute in https://github.com/dotnet/sdk/pull/32751 which causes the IntegrationTests to fail.

tmds:
@baronfel can you take another look at this PR?

baronfel:
@tmds this looks solid - I want to do some testing myself, and I'd also like to add a podman-using run to baronfel/sdk-container-demo before merging this.

baronfel:
This is working in sdk-container-builds using the package built from this PR: https://github.com/baronfel/sdk-container-demo/actions/runs/5134715407/jobs/9239111546

Ignore the other failures - they are because of a runtime mismatch on the Regex Source Generator, actual users won't hit this error, nor will users using the nightly package from the nightly feeds.

I was able to prove that podman alone knew about the push by checking image listings for podman and docker - if podman were just an alias to docker then docker would also list the new image, but here we can see that is not the case:

image

baronfel (approved):
:shipit: Thank you for making our support for podman more resilient!

tmds:
@baronfel can you merge this?

PR #32905, opened 2023-05-30
Rename GenAPI IncludeVisibleOutsideOfAssembly

### LINKED ISSUE #32258: GenAPI IncludeVisibleOutsideOfAssembly --include-visible-outside is misnamed
### Describe the bug
"Visible outside of assembly" typically means public API, however this flag for GenAPI is indicating to include internal API.

A better name might be `IncludeInternals (--include-internals)` which is similar to the APICompat flag `--respect-internals`.

### To Reproduce
`dotnet genapi -?`

Before making this change I'd like to confirm that no-one is using the existing flag.

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/32258

There's only a single GenAPI consumer yet which is SBRP which doesn't use this option. "IncludeVisibleOutsideOfAssembly" -> "RespectInternals" (which is already used and exposed by APICompat).

Also change "includeInternals" to "includeInternalSymbols" which is used in most cases already.

This consolidates frontends to use the "respectInternals" name while the backend libraries use the "includeInternalSymbols" name.

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI.Task/build/Microsoft.DotNet.GenAPI.Task.targets
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI.Task/build/Microsoft.DotNet.GenAPI.Task.targets
@@ -1,5 +1,6 @@
 <!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. -->
 <Project>
+
   <PropertyGroup>
     <DotNetGenAPITaskAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tools\net472\Microsoft.DotNet.GenAPI.Task.dll</DotNetGenAPITaskAssembly>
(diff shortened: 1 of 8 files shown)


src/GenAPI/Microsoft.DotNet.GenAPI.Task/build/Microsoft.DotNet.GenAPI.Task.targets
```diff
@@ -49,9 +50,10 @@
       ExceptionMessage="$(GenAPIExceptionMessage)"
       ExcludeApiFiles="@(GenAPIExcludeApiList)"
       ExcludeAttributesFiles="@(GenAPIExcludeAttributesList)"
-      IncludeVisibleOutsideOfAssembly="$(GenAPIIncludeVisibleOutsideOfAssembly)"
+      RespectInternals="$(GenAPIRespectInternals)"
```
andriipatsula, line 53: Please, remember to update the SBRP generator.

andriipatsula (approved):

src/GenAPI/Microsoft.DotNet.GenAPI.Task/build/Microsoft.DotNet.GenAPI.Task.targets
ViktorHofer: We don't use `GenAPIIncludeVisibleOutsideOfAssembly` in the SBRP generator.

PR #32919, opened 2023-05-31
Containers: include the registry name when pushing an image.

tmds:
The full name can then easily be copy-pasted.

@baronfel ptal.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/Registry.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Registry.cs
@@ -57,12 +57,13 @@ internal sealed class Registry
 
     public Registry(Uri baseUri)
     {
+        RegistryName = DeriveRegistryName(baseUri);
         BaseUri = baseUri;
+        // "docker.io" is not a real registry. Replace the uri to refer to an actual registry.
         if (BaseUri.Host == ContainerHelpers.DockerRegistryAlias)
         {
             BaseUri = new UriBuilder(BaseUri.ToString()) { Host = DockerHubRegistry1 }.Uri;
         }
-        RegistryName = DeriveRegistryName(baseUri);
         _client = CreateClient();
     }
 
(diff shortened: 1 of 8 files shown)


src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -130,7 +130,7 @@ internal async Task<bool> ExecuteAsync(CancellationToken cancellationToken)
                             destinationImageReference,
                             message => SafeLog(message),
                             cancellationToken).ConfigureAwait(false);
-                        SafeLog("Pushed container '{0}' to registry '{1}'", destinationImageReference.RepositoryAndTag, OutputRegistry);
+                        SafeLog("Pushed image '{0}' to registry '{1}'", destinationImageReference, OutputRegistry);
```
tmds, line 133: This PR is about including the image in this log message, so it can be easily copy-pasted, for example when you want to do a `docker run` next.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -55,8 +57,13 @@ internal sealed class Registry
...
     public Registry(Uri baseUri)
     {
-        BaseUri = baseUri;
         RegistryName = DeriveRegistryName(baseUri);
+        BaseUri = baseUri;
+        // "docker.io" is not a real registry. Replace the uri to refer to an actual registry.
+        if (BaseUri.Host == ContainerHelpers.DockerRegistryAlias)
+        {
+            BaseUri = new UriBuilder(BaseUri.ToString()) { Host = DockerHubRegistry1 }.Uri;
+        }
```
tmds, line 66: I've moved the mapping of the `docker.io` alias here, so all log messages contain `docker.io` (when that is what the user specified).

baronfel (approved):
This is a good, and very minor, change. This makes it so that the full name of the image is printed, which is very useful. Good thinking!

Issue #18676, opened 2021-07-01
[Package Validation] Baseline package validation should run with references


safern:
Currently we are running API Compat without references, causing typeforwards to not be resolved or some other rules might be limited by this.

cc: @ericstj @Anipik

Anipik:
@safern do we have that support available in api compat ?

safern:
> @safern do we have that support available in api compat ?

Yes we do. You just specify, `resolveAssemblyReferences` to true when creating the `AssemblySymbolLoader`: https://github.com/dotnet/sdk/blob/faf81a525654c706d1fdcc36fce228f24df959ab/src/Compatibility/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs#L30

And then you can specify search directories for the loader to find references on:
https://github.com/dotnet/sdk/blob/faf81a525654c706d1fdcc36fce228f24df959ab/src/Compatibility/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs#L43

https://github.com/dotnet/sdk/blob/faf81a525654c706d1fdcc36fce228f24df959ab/src/Compatibility/Microsoft.DotNet.ApiCompatibility/AssemblySymbolLoader.cs#L50

ericstj:
My suggestion in the past was just to use the current set of references as an approximation.  It's not perfect and it can miss breaking changes in the references, but those are not really the things under test.

Fetching the real references for a past packages *correctly* would require a whole separate restore + resolve references as an isolated project.  I don't see this as a practical tradeoff we'd make for the common case.  I think we should permit it at the task level if someone wanted to do this themselves (for instance by having a separate project and writing some targets) but in the default case I think we can just get by using existing references as an approximation.

Issue #23533, opened 2022-01-21
Renable packaging tests disabled for the helix merge


marcpopMSFT:
A handful of ~7 tests had to be disabled from the packaging tests in the PR to move to helix:
https://github.com/dotnet/sdk/pull/23381

This issue is to track reenabling them.

PR #32930, opened 2023-05-31
[APICompat - PackageValidation] Enable baseline assembly references

### LINKED ISSUE #32922: ApiCompat flags introduction of a base exception type for exceptions previously inheriting from System.Exception
### Describe the bug
ApiCompat check flags introduction of exception hierarchy:

was:
`<Our Exceptions> --> System.Exception`
proposed:
`<Our Exceptions> --> Microsoft.Build.BackEnd.BuildExceptionBase --> System.Exception`

error:
```
Microsoft.NET.ApiCompat.ValidatePackage.targets(39,5): error CP0007: (NETCORE_ENGINEERING_TELEMETRY=Build) Type 'Microsoft.Build.Exceptions.BuildAbortedException' does not inherit from base type 'System.Exception' on ref/net472/Microsoft.Build.dll but it does on [Baseline] ref/net472/Microsoft.Build.dll
```

More details on the scenario: https://github.com/dotnet/msbuild/pull/8779#issuecomment-1568311132
Build with failure: https://dev.azure.com/dnceng-public/public/_build/results?buildId=287532&view=results

The new base was introduced in assembly `Microsoft.Build.Framework`. The exceptions in the same assembly that underwent same refactors are not flaged. The change is flaged only in assembly `Microsoft.Build`. However `Microsoft.Build` references `Microsoft.Build.Framework` and the dependency is properly captured in the packages as well:

image

### To Reproduce
Introduce new base exception type that is defined in the referenced assembly 

### Further technical details
- sdk `7.0.203` (https://github.com/dotnet/msbuild/blob/main/global.json)

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/32922, https://github.com/dotnet/sdk/issues/18676, https://github.com/dotnet/sdk/issues/31271

If assembly references aren't provided for the baseline package via the MSBuild task or the CLI frontend, use the current package's assembly references as an approximation.

As current assembly references are grouped by TFMs, use the nearest compatible TFM for which assembly references are provided. This guarantees that assembly references from current are used even if TFMs don't match.

@ericstj what do you think about the assembly reference change related to picking the nearest compatible TFM? Do we want a baseline package with a `lib/net7.0/a.dll` asset to use the assembly references from the current package `lib/net8.0/a.dll` asset? Note that the current assembly symbol loader ignored versions.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
+++ b/src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using Microsoft.DotNet.ApiCompatibility;
 using Microsoft.DotNet.ApiCompatibility.Logging;
@@ -25,19 +26,13 @@ namespace Microsoft.DotNet.PackageValidation
             Package leftPackage,
             Package? rightPackage = null)
         {
+            Debug.Assert(leftContentItems.Count > 0);
+            Debug.Assert(rightContentItems.Count > 0);
+
             // Don't enqueue duplicate items (if no right package is supplied and items match)
-            if (rightPackage == null && ContentItemCollectionEquals(leftContentItems, rightContentItems))
+            if (rightPackage is null && ContentItemCollectionEquals(leftContentItems, rightContentItems))
             {
                 return;
-            }
-
-            MetadataInformation[] left = new MetadataInformation[leftContentItems.Count];
-            for (int leftIndex = 0; leftIndex < leftContentItems.Count; leftIndex++)
-            {
-                left[leftIndex] = GetMetadataInformation(log,
-                    leftPackage,
-                    leftContentItems[leftIndex],
-                    options.IsBaselineComparison ? Resources.Baseline + " " + leftContentItems[leftIndex].Path : null);
             }
 
             MetadataInformation[] right = new MetadataInformation[rightContentItems.Count];
@@ -48,30 +43,46 @@ namespace Microsoft.DotNet.PackageValidation
                     rightContentItems[rightIndex]);
             }
 
+            MetadataInformation[] left = new MetadataInformation[leftContentItems.Count];
+            for (int leftIndex = 0; leftIndex < leftContentItems.Count; leftIndex++)
+            {
+                left[leftIndex] = GetMetadataInformation(log,
+                    leftPackage,
+                    leftContentItems[leftIndex],
+                    displayString: options.IsBaselineComparison ? Resources.Baseline + " " + leftContentItems[leftIndex].Path : null,
+                    // Use the assembly references from the right package if the left package doesn't provide them.
+                    assemblyReferences: rightPackage is not null ? right[right.Length > leftIndex ? leftIndex : 0].References : null);
+            }
+
             apiCompatRunner.EnqueueWorkItem(new ApiCompatRunnerWorkItem(left, options, right));
         }
 
         private static MetadataInformation GetMetadataInformation(ISuppressableLog log,
             Package package,
             ContentItem item,
-            string? displayString = null)
+            string? displayString = null,
+            IEnumerable<string>? assemblyReferences = null)
         {
             displayString ??= item.Path;
-            string[]? assemblyReferences = null;
 
             if (item.Properties.TryGetValue("tfm", out object? tfmObj))
             {
-                // Find the nearest set of assembly references for the package item (assembly).
-                NuGetFramework targetFramework = ((NuGetFramework)tfmObj);
-                assemblyReferences = package.FindBestAssemblyReferencesForFramework(targetFramework);
+                string targetFramework = ((NuGetFramework)tfmObj).GetShortFolderName();
 
-                if (package.AssemblyReferences != null && assemblyReferences is null)
+                if (package.AssemblyReferences != null)
                 {
-                    log.LogWarning(new Suppression(DiagnosticIds.SearchDirectoriesNotFoundForTfm) { Target = displayString },
-                        DiagnosticIds.SearchDirectoriesNotFoundForTfm,
-                        string.Format(Resources.MissingSearchDirectory,
-                            targetFramework.GetShortFolderName(),
-                            displayString));
+                    if (package.AssemblyReferences.TryGetValue(targetFramework, out string[]? assemblyReferencesRaw))
+                    {
+                        assemblyReferences = assemblyReferencesRaw;
+                    }
+                    else
+                    {
+                        log.LogWarning(new Suppression(DiagnosticIds.SearchDirectoriesNotFoundForTfm) { Target = displayString },
+                           DiagnosticIds.SearchDirectoriesNotFoundForTfm,
+                           string.Format(Resources.MissingSearchDirectory,
+                               targetFramework,
+                               displayString));
+                    }
                 }
             }
 
--- a/src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
+++ b/src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
@@ -61,14 +61,16 @@ namespace Microsoft.DotNet.PackageValidation
 
             if (item.Properties.TryGetValue("tfm", out object? tfmObj))
             {
-                string targetFramework = ((NuGetFramework)tfmObj).GetShortFolderName();
+                // Find the nearest set of assembly references for the package item (assembly).
+                NuGetFramework targetFramework = ((NuGetFramework)tfmObj);
+                assemblyReferences = package.FindBestAssemblyReferencesForFramework(targetFramework);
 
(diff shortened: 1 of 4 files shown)


ericstj:
> @ericstj what do you think about the assembly reference change related to picking the nearest compatible TFM? 

What I was imagining we'd do was use the right-hand-side dependencies when the left-hand-side dependencies were missing.  This was what we did previously with API Compat.  

Nearest might be OK - so long as you're going forward and not back (I always get mixed up with NuGet's terms around this), since the nearest forward _should_ be compatible and _should_ be a parallel operation to the compatibility tuples we're creating between frameworks.

ericstj (commented):
Can we add a test case for this somehow?

ViktorHofer:
> What I was imagining we'd do was use the right-hand-side dependencies when the left-hand-side dependencies were missing. This was what we did previously with API Compat.

Yes, that's what this PR is doing.

> Nearest might be OK - so long as you're going forward and not back (I always get mixed up with NuGet's terms around this), since the nearest forward should be compatible and should be a parallel operation to the compatibility tuples we're creating between frameworks.

I'm slightly confused what you refer to with going back. The current package (right hand side) usually uses the same or a newer TFM and presumably rarely an older TFM. In my example above, would we want to allow the following behavior or not (?):

> Do we want a baseline package with a lib/net7.0/a.dll asset to use the assembly references from the current package lib/net8.0/a.dll asset?

> Can we add a test case for this somehow?

Non trivial as many of our package validation integration tests are currently disabled because of https://github.com/dotnet/sdk/issues/23533. I'm unsure if I can come up with a new test without deeply looking and with that fixing the test issue. I did run my changes on top off the msbuild change and the error went away. I also verified in the debugger that the assembly references are considered on the left hand side - as expected.

ericstj:
> Yes, that's what this PR is doing.

Not quite, it's computing which references to use independently based on the TFM of each side.  I meant something more specific.  Given a tuple comparing {A} to {B, b-references} use b-references for A -- without any calculation involving TFM at all.  Its a bit mechanical and leaves less up to chance.  For example, in this method:
https://github.com/dotnet/sdk/blob/e0a643139aa4ec61f8953f36ff3cd0988bbbce51/src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs#LL20C28-L20C57
All the content items on the right would have the same TFM and you could pass that down to `GetMetadataInformation` and use it in case the left-side TFM didn't exist in `package.AssemblyReferences`.  Maybe as a "fallbackReferenceFramework" parameter?

> would we want to allow the following behavior or not

Yeah, we'd want that behavior.  I'm agreeing with you -- I'm just asking to double check that `NuGetFrameworkUtility.GetNearest` is doing the right thing here.  I can never remember which direction that API searches.  Have a look at this tool to illustrate what we *don't* want to do: https://nugettools.azurewebsites.net/6.4.0/get-nearest-framework?project=net7.0&package=net6.0%0D%0Anet8.0.  Regardless, if you can avoid doing an extra calculation here I think that would bet better.

ViktorHofer:
> Not quite, it's computing which references to use independently based on the TFM of each side. I meant something more specific. Given a tuple comparing {A} to {B, b-references} use b-references for A -- without any calculation involving TFM at all. Its a bit mechanical and leaves less up to chance.

What makes this hard with the current architecture is that the component that loads symbols and compares left against rights does have a 1:n relationship. A single left is often compared against multiple rights. As an example:

| Package | TFMS |
| --- | --- |
| Baseline | netstandard2.0 |
| Current | netstandard2.0, net7.0 |

APICompat / PackageValidation validates the `[Baseline] netstandard2.0` asset against the `netstandard2.0` and `net7.0` assets. That design is implemented in all `ApiCompatibility.dll` components: `ApiCompatRunner`, `ApiComparer`, `RuleRunner` except for the rules themselves which operate on symbols via a 1:1 relationship.
While in this example, it's clear which reference assemblies to pick (as there's a direct match), it's less clear in the following case:

| Package | TFMS |
| --- | --- |
| Baseline | netstandard2.0 |
| Current | net6.0 |

In that scenario, the customer decides to drop the netstandard2.0 asset but still expects APICompat to validate that the API between the netstandard2.0 and the net6.0 isn't changed in an incompatible way. It's not entirely clear to me if we want to pass the `net6.0` reference assemblies from the current asset to the `[Baseline] netstandard2.0` asset.

ViktorHofer:
I found a place that enqueues package validation assemblies with an 1:1 mapping (TFM to TFM) which allows to re-use the assembly references from the right. Let's see what CI says. I re-ran the tests on top of the msbuild PR and the errors disappeared.

ericstj:
> It's not entirely clear to me if we want to pass the net6.0 reference assemblies from the current asset to the [Baseline] netstandard2.0 asset.

Think about API Compat baseline check as doing an experiment on framework supported by the latest package.  It's doing a _what if_ check between the previous version and the new version on a supported framework.  Imagine all the references as the "control" in that experiment.  So API Compat is answering the question of _what will break if I updated the baseline package to the latest package in a project targeting a specific framework_?  The references in that case are not the variable under test, so treating them like a control is reasonable -- it does mean we might miss breaking changes in those references but we're already excluding them from the comparison rules so IMO it's an acceptable heuristic.

src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
```diff
@@ -48,28 +43,46 @@ internal static class ApiCompatRunnerExtensions
...
 
+            MetadataInformation[] left = new MetadataInformation[leftContentItems.Count];
+            for (int leftIndex = 0; leftIndex < leftContentItems.Count; leftIndex++)
+            {
+                left[leftIndex] = GetMetadataInformation(log,
+                    leftPackage,
+                    leftContentItems[leftIndex],
+                    displayString: options.IsBaselineComparison ? Resources.Baseline + " " + leftContentItems[leftIndex].Path : null,
+                    // Use the assembly references from the right package if the left package doesn't provide them.
+                    assemblyReferences: rightPackage is not null ? right[right.Length > leftIndex ? leftIndex : 0].References : null);
```
ericstj, line 54: Are the references actually specific to some index here?  I thought this index was iterating through multiple content items from the package, multiple assemblies in the same folder, so I would expect all the references to be the same for a given call of this method.

Ultimately it looks like a work item loads all assemblies in the same loader ([CSharpCompilation](https://github.com/dotnet/sdk/blob/5f6b9be62ba8952410d2b181fe5a2ac3fee4c03a/src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoader.cs#L29)), so associating references with individual content items is probably unnecessary overhead:
https://github.com/dotnet/sdk/blob/5f6b9be62ba8952410d2b181fe5a2ac3fee4c03a/src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Runner/ApiCompatRunner.cs#LL123C31-L123C31

ViktorHofer:
With these changes, I do see quite a lot of errors in a runtime allconfigurations build. I just verified that at least one of them is legit and due to https://github.com/dotnet/runtime/pull/75235.

I need to look at the other ones now.

ericstj (commented):
This looks better, just one small suggestion.

src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
ViktorHofer: It's unnecessary, yes but I felt like keeping the index in sync (as in most cases left and right count is in sync) makes more sense than just always defaulting to the first item.

ericstj: Fair, but it was confusing to me trying to understand the relationship between the indexes of these two lists.  I was trying to understand why we would want to align them but would also be OK with just substituting the first.  Perhaps a future refactoring might move references up to the work-item instead of associate with the individual files.

ericstj (approved):
Change seems OK.  My feedback is more around a future refactoring.  Thank you for fixing this, by the way.  I think this fix will help quite a bit.

src/ApiCompat/Microsoft.DotNet.PackageValidation/ApiCompatRunnerExtensions.cs
ViktorHofer: I will remove that indexing as you are right that it's unnecessary overhead and adds complexity.

> Ultimately it looks like a work item loads all assemblies in the same loader ([CSharpCompilation](https://github.com/dotnet/sdk/blob/5f6b9be62ba8952410d2b181fe5a2ac3fee4c03a/src/Microsoft.DotNet.ApiSymbolExtensions/AssemblySymbolLoader.cs#L29)), so associating references with individual content items is probably unnecessary overhead:

Just to clarify, a workitem doesn't share the compilation context as this would otherwise result in reference assembly clashes. The individual elements (lefts and rights) each create a compilation context and load the assemblies into that container. Not relevant for this discussion but this domain has some perf optimization potentials. I.e. the assembly symbol loader could respect assembly versions (and public key tokens) and cache the loaded assembly symbols.

ViktorHofer:
All the errors that appeared in runtime with this change are correct. There is one odd case though:
- APIs switching from the public `ObsoleteAttribute` to the internal polyfill one (for netstandard2.0). That is flagged as an attribute removal (as the public attribute is removed and APICompat filters out the internal one).

PR #32943, opened 2023-05-31
Improve errors about unsupported aot scenarios

sbomer:
This improves the errors when there's no `KnownILCompilerPack` to include more detailed information. There are now separate errors for an unsupported host RID, unsupported target RID, and unsupported TFM (the existing warning).

Partial fix for https://github.com/dotnet/runtime/issues/84583. The other piece is in https://github.com/dotnet/installer/pull/16545.


jkotas (approved):

MichalStrehovsky (approved):
Thank you!

Today I learned that "assembly" in Czech is "sestavení" and "AOT" is "kompilace s předstihem", but we don't translate "runtime".

PR #32950, opened 2023-06-01
Enable interceptors compiler feature when RDG is enabled

captainsafia:
The interceptors compiler feature will be hidden behind a compiler feature flag. To smooth out the user experience, we want to automatically enable this feature flag to projects that are using the RequestDelegateGenerator.

Although the interceptors feature is not in, it's necessary to get this code change in first so that we can flow a version of the SDK that enables this feature flag into the repo once the interceptors implementation lands.

### DIFF
--- a/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
+++ b/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
@@ -58,6 +58,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
   </PropertyGroup>
 
+  <!-- Enable the interceptors compiler feature by default for projects that use the RequestDelegateGenerator. -->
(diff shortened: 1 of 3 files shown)


RikkiGibson (approved):
LGTM, I'll rename the feature on the compiler side accordingly.

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
```diff
@@ -58,6 +58,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
   </PropertyGroup>
 
+  <!-- Enable the interceptors compiler feature by default for projects that use the RequestDelegateGenerator. -->
+  <PropertyGroup Condition="'$(EnableRequestDelegateGenerator)' == 'true'">
```
eerhardt, line 62: @layomia - how do you plan on handling `EnableConfigurationBindingGenerator`? Should we optimistically add this now?

```suggestion
  <PropertyGroup Condition="'$(EnableRequestDelegateGenerator)' == 'true' Or '$(EnableConfigurationBindingGenerator)' == 'true'">
```

eerhardt (approved):
LGTM. We can decide separately if we should update the Config Binder generator.

PR #32952, opened 2023-06-01
Explicitly enable BF for WinForms and WPF in Microsoft.NET.Sdk.targets

### LINKED ISSUE #32941: BinaryFormatter is disabled in WinForms or WPF projects in preview6
### Describe the bug
The intent behind this change - https://github.com/dotnet/sdk/commit/ccfcbe828cda905f1f995f63f26485dd9a7a7dd5 was that projects with these properties
```
    <UseWindowsForms>true</UseWindowsForms>
    <UseWPF>true</UseWPF>
```
Have BinaryFormatter.Serialize/Deserialize methods enabled in .NET8. However in SDK Preview6, we are getting a `NotSupportedException` 
### To Reproduce

1.	Install the [latest SDK](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdotnet%2Finstaller%23installers-and-binaries&data=05%7C01%7CTatiana.Solyanik%40microsoft.com%7C813bfb40a3cb4b89886708db621395ba%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638211608480348989%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=bb6tzFSBYYLdP%2FIGGwZkE5LV%2Fz0TwURa7H8CLSwBJy0%3D&reserved=0)
2.	Create a winforms application targeting .NET 8 in the VS (app is here [https://github.com/Tanya-Solyanik/Repro](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FTanya-Solyanik%2FRepro&data=05%7C01%7CTatiana.Solyanik%40microsoft.com%7C813bfb40a3cb4b89886708db621395ba%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638211608480348989%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=LfZq%2FblGC0ldwEvVhsOC7FmXKsMcUl8mSmxmJvtVktI%3D&reserved=0))
3.	Add `<NoWarn>SYSLIB0011</NoWarn>` to the project file
4.	Add a call to BinaryFormatter.Serialize()
5.	Run that code
 
**Result**: _System.NotSupportedException: BinaryFormatter serialization and deserialization are disabled within this application. See [https://aka.ms/binaryformatter](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Faka.ms%2Fbinaryformatter&data=05%7C01%7CTatiana.Solyanik%40microsoft.com%7C813bfb40a3cb4b89886708db621395ba%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638211608480348989%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=%2BHG2WuS1NzS5%2FDUJqWSsPaQtBACgCG9HNi513bfu1eY%3D&reserved=0) for more information.
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Serialize(Stream serializationStream, Object graph)
   at WinFormsApp29.Form1.button1_Click(Object sender, EventArgs e) in C:\Users\tanyaso\source\repos\WinFormsApp29\WinFormsApp29\Form1.cs:line 17_

### Further technical details
- Repros both in the VS and on the command line, as long as SDK preview 6.23280 is installed
- If I add a switch
```
            AppContext.SetSwitch(
                "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization",
                isEnabled: true);
```
or 
`<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>`
BinaryFormatter serializes successfully.

GrabYourPitchforks:
Resolves https://github.com/dotnet/sdk/issues/32941.

The _\*.runtimeconfig.json_ files are being picked up from the shared framework builds and included as part of the .NET 8 installers. Since the shared framework is itself not a WinForms or WPF application, it results in "explicitly disable BF" being hardcoded into the config file, which is then distributed to customers' machines via the installer. This is resulting in BF disablement for _all_ project types starting with Preview 5, even though that was not our intent.

The proposed fix here special-cases WinForms and WPF within the .targets file where we made the initial change, and it forces those project types to re-enable BF if the developer has not set a different value in the .csproj. One downside of this change is that if a WinForms or WPF application is published against .NET 8, the "enable BF" policy will be carried within that app's _runtimeconfig.json_ file, which means that if the app is rolled forward it won't honor the .NET 9 "disabled absolutely everywhere" default setting once that change is checked in. But after talking with some folks here, that seems like an acceptable risk to unblock this scenario.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -165,6 +165,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- controls warn as error -->
     <_BinaryFormatterObsoleteAsError>true</_BinaryFormatterObsoleteAsError>
     <!-- controls runtime behavior (AppContext & trimming) -->
+    <EnableUnsafeBinaryFormatterSerialization Condition="'$(EnableUnsafeBinaryFormatterSerialization)' == '' AND '$(_ProjectTypeRequiresBinaryFormatter)' == 'true'">true</EnableUnsafeBinaryFormatterSerialization>
     <EnableUnsafeBinaryFormatterSerialization Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '8.0')) AND '$(_ProjectTypeRequiresBinaryFormatter)' != 'true'">false</EnableUnsafeBinaryFormatterSerialization>
   </PropertyGroup>
 


hoyosjs (approved):

GrabYourPitchforks:
The root cause of the BF issue seems to be https://github.com/dotnet/sdk/issues/32969. This PR can still serve to unblock https://github.com/dotnet/sdk/issues/32941 temporarily. However, if https://github.com/dotnet/sdk/issues/32969 is fixed, then this PR should be reverted.

GrabYourPitchforks:
I've retargeted the PR into the Preview5 branch, per Chet's request.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -165,6 +165,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- controls warn as error -->
     <_BinaryFormatterObsoleteAsError>true</_BinaryFormatterObsoleteAsError>
     <!-- controls runtime behavior (AppContext & trimming) -->
+    <EnableUnsafeBinaryFormatterSerialization Condition="'$(EnableUnsafeBinaryFormatterSerialization)' == '' AND '$(_ProjectTypeRequiresBinaryFormatter)' == 'true'">true</EnableUnsafeBinaryFormatterSerialization>
```
marcpopMSFT, line 168: At this point, EnableUnsafeBinaryFormatterSerialization is known to not be true. It could be empty, false, or something else.  You'll only set it to true if it was empty before not if it was explicitly set to false. Confirming that that's what you want.

Also, this will only run on 7.0+ so confirming you only want this change to apply there

Tanya-Solyanik: Winforms/WPF application might want to disable binary formatter explicitly by setting this property to false, we shouldn't override it.

marcpopMSFT (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
GrabYourPitchforks: > At this point, EnableUnsafeBinaryFormatterSerialization is known to not be true. It could be empty, false, or something else. You'll only set it to true if it was empty before not if it was explicitly set to false. Confirming that that's what you want.

Ideally I don't want this line to be present _at all_. It's only a workaround for https://github.com/dotnet/sdk/issues/32969. Once that fix is in, I'm reverting this change, since I don't want applications to be carrying the "true" flag in their own runtimeconfig.json files unless they've explicitly set the property themselves as true.

But like Tanya said, for now, we need to support people having explicitly written "true" or "false" in config and we need to honor those explicit settings.

> Also, this will only run on 7.0+ so confirming you only want this change to apply there

Correct. 7.0 WinForms apps are not affected.

PR #32981, opened 2023-06-01
Enable error when using BinaryFormatter in GenerateResource

Forgind:
When GenerateResourceWarnOnBinaryFormatterUse is true and GenerateResource executes and ultimately uses BinaryFormatter, it throws a warning. This is to enable that warning when building from the SDK.

@dsplaisted, is this the best way to set that property? I noticed that Microsoft.NET.Sdk.props is imported in things like Sdk.props and the Razor SDK; adding Sourcelink.props (I think) adds C++ scenarios...what scenarios am I missing, and is there a single central location that would hit everything?

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.SourceLink.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.SourceLink.props
@@ -18,6 +18,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!-- Suppress implicit SourceLink inclusion if any Microsoft.SourceLink package is referenced. -->
     <SuppressImplicitGitSourceLink Condition="'$(PkgMicrosoft_SourceLink_Common)' != ''">true</SuppressImplicitGitSourceLink>
+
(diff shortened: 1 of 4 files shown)


JanKrivanek:
I have a lack ow knowledge here - what will happen if one tries to `dotnet build` app targeting Full FW? We probably do not want to enable this check there by default - do we?

dsplaisted:
Is there something that you can refer to giving some more background on this change?

Does it need to be set in a .props file?  If you set it in Microsoft.NET.Sdk.targets, I think it would cover C++ as well as other cases.

Should we include tests for using this property?  Should you be able to turn it off?

JanKrivanek:
Related: https://github.com/dotnet/msbuild/issues/8827
We want to get idea of BinaryFormatter usages - internal and external - via `GenerateResource` task. This is needed to:
* Properly plan the replacement/deprecation
* To properly suggest the viable deprecation timelines (net8 vs net9)

rainersigwald (changes requested):
This should be in targets and it should be enabled only when targeting .NET 8+.

cc @merriemcgaw @JeremyKuhne

rainersigwald (approved):

PR #33005, opened 2023-06-02
Do not fail dotnet watch on workspace errors

### LINKED ISSUE #28998: .Net7 dotnet watch regression: Failed to create MSBuildWorkspace for *.fsproj
I'm running BlazorServer app which has a dependency on F# project. Previously with .Net6 there was no problem running applying HotReload and some of stuff was changing instantly (for example applying a css class to blazor markup).

With .Net 7 I'm seeing an error "Cannot_open_project_0_because_the_file_extension_1_is_not_associated_with_a_language" error because only csproj and vbproj are supported.

Now it does full rebuild even if only C# parts are changed. Before it worked fine.

To repro:
Create an empty blazor server project and reference an F# project with some data. Run dotnet watch. Change something trivial like css/markup.

Expected: hot reload works and page is updated immediately
Result: hot reload fails with this error and full rebuild is issued

tmat:
Report all workspace diagnostics only in `--verbose` mode. Any significant issues with incorrectly formed projects would be reported by `dotnet build` before we load the workspace.

Fixes https://github.com/dotnet/sdk/issues/28998

### DIFF
--- a/src/BuiltInTools/dotnet-watch/HotReload/CompilationWorkspaceProvider.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/CompilationWorkspaceProvider.cs
@@ -41,14 +41,8 @@ namespace Microsoft.DotNet.Watcher.Tools
 
                 workspace.WorkspaceFailed += (_sender, diag) =>
                 {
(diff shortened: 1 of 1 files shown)


tlmii (approved):
LGTM

En3Tho:
@tmat Big thanks!

PR #33019, opened 2023-06-03
[Containers] Fix registry operations that use chunks and various resiliency enhancements

baronfel:
AWS publishes that used the chunk mechanisms were failing because AWS ECR has a constraint that isn't mentioned in the HTTP API Spec for container registries - they _require_ all chunked uploads to be at least 5MB in size, except the last chunk in a chunked upload.  Users would use the chunked upload when layers got too big for ECR to accept in a single atomic push - ECR's limit for atomic layer upload size (as well as individual chunk size) is 10MB.  Our previous upload strategy was either-or, so you either could atomic upload layers < 10MB, or you couldn't upload anything at all (because you would try to use the chunked upload and it would always fail).

All of this together made me realize that our current upload strategy wasn't working - we needed to consult implementations in other client libraries and see how they approached the problem of layer uploads. Most seem to use the following strategy:
* attempt to have the destination 'mount' the layer from another source registry,
* if that doesn't work then attempt an atomic upload, 
* if that doesn't work fallback to a chunked upload

I have implemented this in this PR - we now have a fallback, and in addition to that we now make use of headers that registries can (optionally) send when an upload operation is begun to determine what our chunk size should be, and even if that registry will support the atomic-PUT mechanism (if it doesn't we jump right to chunking).

To help discover all of this I used Fiddler as well as usage logs from the various cloud providers, as well as adding trace-level logging to this library. I've plumbed the trace messages back into the binlogs as part of this, but they will not surface to the user in normal operation.

A series of test publishes to the various registries can be seen [here](https://github.com/baronfel/sdk-container-demo/actions/runs/5172500858) - for the first time in a very long time we are green!

This _may_ also help with speed issues that users were experiencing, because we optimistically try the atomic upload now, with a safe fallback. For Azure specifically we're still using chunk sizes of 64k (they don't give us any size details in the HTTP responses), so if the atomic upload doesn't work that won't change anything for folks, but more people should have _more_ success.

---

Additional changes:
- fixes the authentication cache logic: previous logic was not cutting the parameters in uri
- introduces logging based on `Microsoft.Logging.Extensions` and ready implementation from dotnet/templating
- makes logic in `Registry.cs` unit-testable and adds the tests.

---

Partially addresses https://github.com/dotnet/sdk-container-builds/issues/424:
- atomic upload to public ECR works
- atomic upload to private ECR works
- chunked upload to private ECR works
- chunked upload to public ECR doesn't work (`"None of the provided layer digests match the calculated layer digest"` error)

fixes https://github.com/dotnet/sdk-container-builds/issues/418 by removing the usage of Cookies in teh base HttpClientHandler.


src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -447,32 +581,91 @@ private async Task<UriBuilder> StartUploadSessionAsync(string repository, string
             throw new ApplicationException(errorMessage);
         }
         cancellationToken.ThrowIfCancellationRequested();
-        return GetNextLocation(pushResponse);
+        var chunkSize = ParseRangeAmount(pushResponse) ?? ParseOCIChunkMinSizeAmount(pushResponse);
```
baronfel, line 584: We should start here. When starting an upload, registries can send `Range` or `OCI-Chunk-Min-Length` headers to tell us what chunk size they can handle. AWS for example sends a range of `0-10485760` (so 10MB) so we know what their upper-bound chunk size is. We do not however get a min-size from them.

Other registries also send one of these headers, but return invalid or useless values for them, so we have to validate that the values are > 0 in order to be used elsewhere. 

All of this data is now captured in a 'UploadInformation' structure that callers can mine for data for the next steps of the operation.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -447,32 +581,91 @@ private async Task<UriBuilder> StartUploadSessionAsync(string repository, string
...
-    private static async Task FinishUploadSessionAsync(string digest, HttpClient client, UriBuilder uploadUri, CancellationToken cancellationToken)
+    /// <summary>
+    /// Uploads the contents of the stream, either in an atomic PUT or in chunks.
+    /// </summary>
+    /// <remarks>
+    /// Generally, we try to upload the blob in an atomic PUT. If the registry does not support this, we fall back to chunked uploads.
+    /// If the atomic PUT fails, we fall back to chunked uploads.
+    /// If the registry provides a chunk max and that max is less than the content length, then we use chunked uploads.
+    /// </remarks>
+    private async Task<UploadFinalizeInformation> UploadBlobContentsAsync(string repository, string digest, Stream contents, HttpClient client, UploadInformation uploadInfo, Action<LogMessage> logProgressMessage, CancellationToken cancellationToken)
```
baronfel, line 633: Now that we've started an upload, we actually perform the upload. This method got _a lot_ smarter. It now tries to do an atomic upload if available - and it seems to be pretty widely available. If the registry actually defined a chunk size, and that chunk size is > than the layer to upload, we feel pretty confident that we can support an atomic upload.

However we also need to recover from faults and switch to a chunked upload for resiliency.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -339,68 +343,186 @@ public async Task<string> DownloadBlobAsync(string repository, Descriptor descri
...
-    private async Task<UriBuilder> UploadBlobChunkedAsync(string repository, string digest, Stream contents, HttpClient client, UriBuilder uploadUri, CancellationToken cancellationToken)
+    /// <summary>
+    /// Captures the data needed to finalize an upload
+    /// </summary>
+    private record UploadFinalizeInformation(UriBuilder uploadUri, string digest);
+
+    /// <summary>
+    /// Upload a blob to the registry using chunks. The chunks are uploaded sequentially, and the size of the chunks is determined by the UploadInformation derived from the 'Start' call
+    /// </summary>
+    private async Task<UploadFinalizeInformation> UploadBlobChunkedAsync(string repository, Stream contents, HttpClient client, UploadInformation uploadInfo, Action<LogMessage> logProgressMessage, CancellationToken cancellationToken)
```
baronfel, line 365: This method changed a lot, but I refactored it to hopefully be more clear.  It's a state machine, so I've modeled it as a series of states based on the status code of the response we get back from uploading each chunk.

The counts for retry logic are very much negotiable - I started with amounts from golang clients that I researched for background here.

Essentially we start a processing loop:
* while there are chunks to send
  * read a chunk
  * send the chunk
  * if ok, loop (there's a quirk here for AWS)
  * if error < retryMax, reset our read state and try again, incrementing our counter
  * finally error out

each of these decisions are a separate method so they can be inspected as individuals.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -339,68 +343,186 @@ public async Task<string> DownloadBlobAsync(string repository, Descriptor descri
...
+            // failures are retried up to retryMax times,
+            // or else we fail the whole operation.
+            (patchResponse.StatusCode switch
+            {
+                HttpStatusCode.Accepted => ProcessSuccessfulChunk(patchResponse),
+                // known quirk - AWS ECR sends 201 Created instead of 202 Accepted
+                HttpStatusCode.Created when IsAmazonECRRegistry => ProcessSuccessfulChunk(patchResponse, ignoreRange: true),
+                _ when retryCount < retryMax => await CheckReadAmountAndRetry(patchUri, cancellationToken).ConfigureAwait(false),
+                _ => await LogError(patchResponse, patchUri, cancellationToken).ConfigureAwait(false)
+            })();
```
baronfel, line 399: Please excuse this odd usage of switch expressions - they cannot currently return void, so all branches are returning `Action` and I'm immediately invoking them. This is a pretty widely-used hack around current C# language limitations on switch expressions. Yes, I've complained to Kathleen :)

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -339,68 +343,186 @@ public async Task<string> DownloadBlobAsync(string repository, Descriptor descri
...
-        var localUploadUri = new UriBuilder(uploadUri.Uri);
-        localUploadUri.Query += $"&digest={Uri.EscapeDataString(digest)}";
+        Uri patchUri = uploadInfo.uploadUri.Uri;
+        contents.Seek(0, SeekOrigin.Begin);
 
-        // TODO: this chunking is super tiny and probably not necessary; what does the docker client do
-        //       and can we be smarter?
-
-        byte[] chunkBackingStore = new byte[MaxChunkSizeBytes];
+        using IncrementalHash hash = IncrementalHash.CreateHash(HashAlgorithmName.SHA256);
```
baronfel, line 371: I added incremental hash generation for chunks because at one point I thought hash calculation was getting modified. Trace logging proved this wasn't so - this can probably safely be removed at this point.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -339,68 +343,186 @@ public async Task<string> DownloadBlobAsync(string repository, Descriptor descri
...
-                throw new ApplicationException(errorMessage);
+                var amountSent = ignoreRange ? null : ParseRangeAmount(response);
+                patchUri = GetNextLocation(response).Uri;
+                if (amountSent is not null)
+                {
+                    chunkStart = amountSent.Value + 1;
+                    // what the server actually read/processed may not align with the bytesRead from the content originally,
+                    // so seek to that position to ensure no gaps
+                    contents.Seek(chunkStart, SeekOrigin.Begin);
+                }
```
baronfel, line 443: This is a pretty big change that's required for correctness - it's possible for registries to only partially accept a chunk, so we need to read their Range headers if present to verify how 'far' in the upload _they_ think they are. If there's a discrepancy, we need to update our read position to that location or else the server will send validation errors.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -339,68 +343,186 @@ public async Task<string> DownloadBlobAsync(string repository, Descriptor descri
...
+                {
+                    // reset back to previous chunk so we can retry
+                    contents.Seek(chunkStart, SeekOrigin.Begin);
+                }
+                else
+                {
+                    UpdateStateFromResponse(getResponse, false);
+                }
+                return () => { };
+            }
```
baronfel, line 474: This is new, and also helps with resiliency - when we retry we might be in a state where the response from the server doesn't have position information (Range headers, etc). In that case, we can issue a GET to the upload URI to get a 'status' response that _does_ have Range headers, and use that to ensure we're in sync. This is why `UpdateStateFromResponse` is a separate method - so we can ensure the metadata is always acted upon the same way.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -526,9 +727,15 @@ private HttpClient GetClient()
 
     private HttpClient CreateClient()
     {
-        HttpMessageHandler clientHandler = new AuthHandshakeMessageHandler(new SocketsHttpHandler() { PooledConnectionLifetime = TimeSpan.FromMilliseconds(10 /* total guess */) });
-
-        if(IsAmazonECRRegistry)
+        HttpMessageHandler clientHandler = new AuthHandshakeMessageHandler(new SocketsHttpHandler()
+        {
+            PooledConnectionLifetime = TimeSpan.FromMilliseconds(10 /* total guess */),
+            // disabling cookies prevents CSRF tokens from being sent - some servers send these and
```
baronfel, line 733: This is a fix for Harbor registry specifically (though it makes sense for all registries). Other golang based clients disable cookies so that UI-only cookies like CSRF tokens don't impact communication, so we should as well.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -8,26 +8,29 @@
...
 
+public enum ProgressMessageLevel {
+    Info,
+    Trace
+}
+
+public record struct LogMessage (ProgressMessageLevel level, string messageFormat, object[] formatArgs) {
+    public static LogMessage Info(string messageFormat, params object[] formatArgs) => new LogMessage(ProgressMessageLevel.Info, messageFormat,  formatArgs);
+    public static LogMessage Trace(string messageFormat, params object[] formatArgs) => new LogMessage(ProgressMessageLevel.Trace, messageFormat,  formatArgs);
+};
```
baronfel, line 26: Finally, a note about logs - I added trace-level logs for the low-level details that I don't think users need to see, but that can be useful when debugging what's gone on.  I also needed a way to expose a 'level' for these logs, so that we could map that onto MSBuild Information severity (either High or Low). Eventually some kind of Microsoft.Extensions.Logging layer might be useful here to prevent this kind of yet-another-abstraction, but it was simple to do here.

baronfel (commented):
I've left an ordered walkthrough of how to read through the changes. They are virtually all driven by Registry.cs, and the primary entry point is `StartUploadSessionAsync`.

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -2,11 +2,22 @@
...
 
 using Microsoft.NET.Build.Containers.Resources;
+using System.Text;
 
 namespace Microsoft.NET.Build.Containers;
 
 public static class ContainerBuilder
 {
+
+    private static void WriteLogMessage(LogMessage message) {
```
donJoseLuis, line 12: very minor: private methods should follow public members.

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -2,11 +2,22 @@
...
+        if (message.level is ProgressMessageLevel.Info) {
+            var builder = new StringBuilder("Containerize: ");
+            builder.AppendFormat(message.messageFormat, message.formatArgs);
+            Console.WriteLine(builder.ToString());
+        }
+        // TODO: dropping Trace level messages - this should be ok for the net472 VS experience. CI systems will get them in binlogs.
```
donJoseLuis, line 18: minor: trace level should be managed globally via configuration.

Should we adopt this common scheme?
https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -2,11 +2,22 @@
...
 using Microsoft.NET.Build.Containers.Resources;
+using System.Text;
 
 namespace Microsoft.NET.Build.Containers;
 
 public static class ContainerBuilder
 {
+
+    private static void WriteLogMessage(LogMessage message) {
+        if (message.level is ProgressMessageLevel.Info) {
```
donJoseLuis, line 13: minor: we should probably move to the `ILogger` scheme.

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -91,7 +102,7 @@ public static class ContainerBuilder
         {
             if (isDaemonPush)
             {
-                LocalDocker localDaemon = GetLocalDaemon(localContainerDaemon,Console.WriteLine);
+                LocalDocker localDaemon = GetLocalDaemon(localContainerDaemon, WriteLogMessage);
```
donJoseLuis, line 105: minor: passing and `ILogger` or `ILoggerFactory` will allow recipients to decide the level of the logged message.

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -134,7 +145,7 @@ public static class ContainerBuilder
         return 0;
     }
 
-    private static LocalDocker GetLocalDaemon(string localDaemonType, Action<string> logger)
+    private static LocalDocker GetLocalDaemon(string localDaemonType, Action<LogMessage> logger)
```
donJoseLuis, line 148: minor: `ILogger` makes this signature more c-sharpie

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -235,6 +235,20 @@ private static void SetEnvironmentVariables(ImageBuilder img, ITaskItem[] envVar
     private void SafeLog(string message, params object[] formatParams) {
         if(BuildEngine != null) Log.LogMessage(MessageImportance.High, message, formatParams);
     }
+    private void SafeLog(LogMessage message) {
```
donJoseLuis, line 238: minor: consider pattern matching.

https://learn.microsoft.com/en-US/dotnet/csharp/language-reference/operators/switch-expression

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
```diff
@@ -65,7 +67,7 @@ public async Task ApiEndToEndWithRegistryPushAndPull()
         var sourceReference = new ImageReference(registry, DockerRegistryManager.BaseImage, DockerRegistryManager.Net6ImageTag);
         var destinationReference = new ImageReference(registry, NewImageName(), "latest");
 
-        await registry.PushAsync(builtImage, sourceReference, destinationReference, Console.WriteLine, cancellationToken: default).ConfigureAwait(false);
+        await registry.PushAsync(builtImage, sourceReference, destinationReference, WriteMessage, cancellationToken: default).ConfigureAwait(false);
```
donJoseLuis, line 70: minor: why is an actual cancellation token not passed?

src/Tests/Microsoft.NET.Build.Containers.UnitTests/DockerDaemonAvailableUtils.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.NET.Build.Containers.UnitTests;
 
 public class DockerDaemonAvailableTheoryAttribute : TheoryAttribute
 {
-    private static bool IsDaemonAvailable = new LocalDocker(Console.WriteLine).IsAvailable();
+    private static bool IsDaemonAvailable = new LocalDocker(message => Console.WriteLine($"{message.level}: {String.Format(message.messageFormat, message.formatArgs)}")).IsAvailable();
```
donJoseLuis, line 10: minor: the `ILogger` scheme allows for different providers, Console included.

donJoseLuis (commented):
some minor comments

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
baronfel: similar to the switch expression state machine in Registry.cs, switch expressions cannot return 'void' in the C# language right now. So I could either do the same immediately-executed-Action hack, or use the old-style switch here.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
baronfel: mostly because this is in testing code only, and pushing to the local dockerized registry is quite fast. cancellation is never a problem in tests, the main problem that arises is if the local docker registry started at all or not!

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -339,68 +343,186 @@ public async Task<string> DownloadBlobAsync(string repository, Descriptor descri
...
-            await UploadBlobAsync(repository, digest, contents, cancellationToken).ConfigureAwait(false);
+            await UploadBlobAsync(repository, digest, contents, logProgressMessage, cancellationToken).ConfigureAwait(false);
         }
     }
 
-    private async Task<UriBuilder> UploadBlobChunkedAsync(string repository, string digest, Stream contents, HttpClient client, UriBuilder uploadUri, CancellationToken cancellationToken)
+    /// <summary>
+    /// Captures the data needed to finalize an upload
+    /// </summary>
+    private record UploadFinalizeInformation(UriBuilder uploadUri, string digest);
```
vlada-shubina, line 360: this member is not ordered correctly (should follow methods).

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
vlada-shubina: Will it be removed then? or moved to DEBUG at least?

vlada-shubina (dismissed):
New changes in `Registry` should be tested better. Please consider mocking client and verifying possible scenarios of retry, chunking etc.

src/Tests/Microsoft.NET.Build.Containers.UnitTests/Microsoft.NET.Build.Containers.UnitTests.csproj
```diff
@@ -13,6 +13,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Containers\Microsoft.NET.Build.Containers\Microsoft.NET.Build.Containers.csproj" />
     <ProjectReference Include="..\Microsoft.NET.TestFramework\Microsoft.NET.TestFramework.csproj" />
+    <PackageReference Include="Moq" Version="4.18.4" />
```
vlada-shubina, line 16: please use globally declared version (from Versions.props instead)

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
vlada-shubina: the method was removed in favor of `ILogger` implementation

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
vlada-shubina: the method was removed in favor of ILogger implementation

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
vlada-shubina: Changed in favor of `ILogger` implementation

src/Tests/Microsoft.NET.Build.Containers.UnitTests/DockerDaemonAvailableUtils.cs
vlada-shubina: Changed in favor of `ILogger` implementation

src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultBlobOperations.cs
```diff
@@ -0,0 +1,34 @@
...
+        Client = client;
+        Upload = new DefaultBlobUploadOperations(Client);
+    }
+
+    public IBlobUploadOperations Upload { get; }
+
+    public async Task<HttpResponseMessage> GetAsync(string repositoryName, string digest, CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        using var request = new HttpRequestMessage(HttpMethod.Get, new Uri($"/v2/{repositoryName}/blobs/{digest}"));
```
vlada-shubina, line 22: @baronfel there is a problem with this code now, previously this line was
`using var request = new HttpRequestMessage(HttpMethod.Get, new Uri(BaseUri, $"/v2/{repositoryName}/manifests/{reference}"));`

Now, `BaseUri` is missing. There are number of tests failing because of that.  Could you please take a look?

Otherwise, I will start on this tomorrow.

baronfel:
@vlada-shubina the new logging has prefixes that stand out:

image

I removed the category headers to return to the previous layout for these cases, but other than that we are green again now.

I've added a few tests around asserting our behavior in chunked/non-chunked scenarios that give me more confidence.

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -166,6 +166,11 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
             request.Headers.Authorization = cachedAuthentication;
         }
 
+        if (request.Method == HttpMethod.Patch && request.Content?.Headers.ContentLength > 0)
+        {
+            request.Headers.AcceptEncoding.Add(new StringWithQualityHeaderValue("gzip"));
```
baronfel, line 171: should we set this on the atomic-upload request specifically instead of in the auth handler?

vlada-shubina: it was fast way to do it.  It is also needed for chunks, though no way to prove now. 
Since we are using single `HttpClient` everywhere, we need to move to `SendAsync` to be able to set the headers, it needs a bit of rework.

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -56,10 +60,10 @@ public static class ContainerBuilder
         }
         if (imageBuilder is null)
         {
-            Console.WriteLine(Resource.GetString(nameof(Strings.BaseImageNotFound)), sourceImageReference.RepositoryAndTag, containerRuntimeIdentifier);
+            Console.WriteLine(DiagnosticMessage.ErrorFromResourceWithCode(nameof(Strings.BaseImageNotFound), sourceImageReference.RepositoryAndTag, containerRuntimeIdentifier));
             return 1;
         }
-        Console.WriteLine("Containerize: building image '{0}' with tags {1} on top of base image {2}", imageName, string.Join(",", imageName), sourceImageReference);
+        logger.LogInformation("Building image '{0}' with tags {1} on top of base image {2}", imageName, string.Join(",", imageName), sourceImageReference);
```
vlada-shubina, line 66: All log messages of levels error, warning, info should be localized.

src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultBlobUploadOperations.cs
```diff
@@ -0,0 +1,93 @@
...
+        if (finalizeResponse.StatusCode != HttpStatusCode.Created)
+        {
+            var headers = finalizeResponse.Headers.ToString();
+            var detail = await finalizeResponse.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+            string errorMessage = Resource.FormatString(nameof(Strings.BlobUploadFailed), $"PUT {putUri}", finalizeResponse.StatusCode, headers + Environment.NewLine + detail);
+            throw new ApplicationException(errorMessage);
+        }
+    }
+
+    public async Task<bool> TryMount(string destinationRepository, string sourceRepository, string digest)
```
vlada-shubina, line 87: This method should accept `cancellationToken` and be `Async`.

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
vlada-shubina: Changed this

vlada-shubina (dismissed):

vlada-shubina:
@baronfel does it really address https://github.com/dotnet/sdk-container-builds/issues/418? I see Harbor leg is still red

baronfel:
The demo Harbor instance gets wiped on a daily basis and so I have to go in periodically and recreate the test registry. It's not a high priority to keep permanently green as a result IMO, but the fix for the issue we had with Harbor was disabling cookies on our HttpClient.

vlada-shubina:
@baronfel @MichalPavlik 

i found couple of issues with the PR:
case: chunked upload / public AWS ECR
- Patch request always returns Range: 0-10485759, `UpdateStateFromResponse` sets next position as `10485760`, therefore upload is stuck on 2nd chunk and never progresses.
- also i found no evidence that GET `https://public.ecr.aws/v2/<repo>/blobs/uploads/<id>` works for AWS ECR, it always returns `NotFound` even though 1st chunk is successfully uploaded.

I haven't tried other registries for `CheckReadAmountAndRetry` correctness yet.

Please take a look if you have time.

baronfel:
We should consider merging this even with the outstanding chunked/public AWS known issue, because this represents a big improvement in performance, usability, and has bugfixes for authentication.

vlada-shubina:
We agreed to split this PR to smaller chunks for the ease of review.

baronfel:
going to go ahead and close this one since we've split it up into reviewable chunks.

PR #33037, opened 2023-06-05
Add support for starting the application through CMD (instead of ENTRYPOINT).

### LINKED ISSUE #398: Support using the base image ENTRYPOINT and starting the application through CMD
Some base images (like those provided by Red Hat) have a container `ENTRYPOINT` that performs some initialization before running the `CMD`.

The current logic supports figuring out the `ENTRYPOINT` so it starts the .NET application.

Can the logic be extended so it is possible to choose to start the application using `CMD`?

cc @baronfel @richlander @omajid @aslicerh

tmds:
The new ContainerAppCommand and ContainerAppCommandArgs properties are used to hold the command line to start the application.

The ContainerAppCommandInstruction controls how the command line is emitted.

The default behavior is to emit an ENTRYPOINT when the base image has no ENTRYPOINT, and a CMD when the base image has an ENTRYPOINT (to use the base image ENTRYPOINT).

This also adds a ContainerCmd property, which can be used to directly set the image CMD.

Fixes https://github.com/dotnet/sdk-container-builds/issues/398.


tmds:
@baronfel this is a WIP, it outlines most of the implementation. I want to share what I have so we can do an early feedback round. ptal.

src/Containers/Microsoft.NET.Build.Containers/Resources/Strings.Designer.cs
```diff
@@ -447,6 +447,15 @@ internal class Strings {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized string similar to CONTAINER2006: Unknown AppCommandInstruction &apos;{0}&apos;. Valid instructions are {1}..
+        /// </summary>
+        internal static string UnknownAppCommandInstruction {
```
tmds, line 453: I've added this manually again.

Maybe it only works on Windows?

@baronfel when you run the top-level build script on Windows, does it update this file with new strings?

tmds:
> Sorry, finally getting to this. Will take a look this afternoon and give feedback.

Take your time. I got just got it in a state that is good for a review.

I'll extend the unit test with some more cases, I don't have any other changes in mind currently.

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -240,4 +242,108 @@ public void Dispose()
...
+                }
+            }
+            else
+            {
+                // Default to use an Entrypoint.
+                // If the base image defines an ENTRYPOINT, print a warning.
+                if (baseImageEntrypoint.Length > 0)
+                {
+                    Log.LogWarningWithCodeFromResources(nameof(Strings.BaseEntrypointOverwritten));
+                }
```
tmds, line 301: Rather than change between `Entrypoint` and `Cmd` based on the presence of a base entrypoint, for deterministic behavior, I've opted to prefer `Entrypoint` and print a warning if the base image defines an entrypoint already (as is the case with Red Hat images).

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.Interface.cs
```diff
@@ -87,6 +87,26 @@ partial class CreateNewImage
     /// </summary>
     public ITaskItem[] EntrypointArgs { get; set; }
 
+    /// <summary>
+    /// Default arguments passed. These can be overridden by the user when the container is created.
+    /// </summary>
+    public ITaskItem[] DefaultArgs { get; set; }
```
tmds, line 93: This roughly corresponds to what https://github.com/dotnet/sdk-container-builds/issues/11 called `ContainerCommand`.

tmds:
> I'll extend the unit test with some more cases, I don't have any other changes in mind currently.

@baronfel I've just extended the test. If you look at it, it should describe well how these different properties combine together to determine the application image Entrypoint and Cmd.

baronfel:
Thanks @tmds - this LGTM overall, but I'm going to wait to merge until Friday. We're going over a batch of the open Containers PRs for detailed review then and I want to make sure the rest of the team is on board with the changes.

baronfel (approved):
Thanks for this - and especially thanks for the suite of tests illustrating the different scenarios.

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -240,4 +242,108 @@ public void Dispose()
     {
         _cancellationTokenSource.Dispose();
     }
+
+    internal (string[] entrypoint, string[] cmd) DetermineEntrypointAndCmd(string[]? baseImageEntrypoint)
```
baronfel, line 246: we need to replicate this logic in the net472 version of the command - the validations there are only looking at EntryPoint/EntrypointArgs and so are breaking after this PR.

tmds: Where is the net472 version?

baronfel: It's the sibling CreateNewImageToolTask.cs file next to CreateNewImage.cs - this task invokes the `containerize` exe with CLI variables, but does validation of those variables in its `Execute` method so that users get good MSBuild diagnostics.

The ToolTask and the 'normal' task share the same Interface.fs partial class, but are conditionally compiled for net472 or net8.0, respectively.

tmds: Got it. Shall I look into making the necessary changes?

baronfel: That would be awesome! I _think_ it would be a matter of

* pushing this entrypoint-detection method down into the .Interface.cs class
* Using it in both CreateNewImage tasks separately
* Emitting the new CLI args for the entrypoint args/command/etc in the ToolTask's 'GenerateCommandLine' method
* A brief look at emitting nice diagnostics from the ToolTask

tmds: I'll look into it. Thanks for the outline.

PR #33043, opened 2023-06-05
add GarbageCollectionAdaptationMode msbuild property

### LINKED ISSUE #87028: GC Dynamic adaptation to app sizes: Support configuration using MSBuild property
Enable MSBuild property similar to ServerGC to configure DOTNET_GCDynamicAdaptation so it could be specified in the template. The setting is not currently supported in runtime.config so that is the first step to get this working before enabling support in the SDK.

mangod9:
Adds a msbuild property to set GarbageCollectionAdaptationMode 

Fixes https://github.com/dotnet/runtime/issues/87028


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -486,6 +486,10 @@ Copyright (c) .NET Foundation. All rights reserved.
                                     Condition="'$(ServerGarbageCollection)' != ''"
                                     Value="$(ServerGarbageCollection)" />
 
+    <RuntimeHostConfigurationOption Include="System.GC.DynamicAdaptationMode"
+                                    Condition="'$(GarbageCollectionAdaptationMode)' != ''"
```
mangod9, line 490: we should determine a better name here, have made it similar to `ServerGarbageCollection` used above.

cshung: What is our policy to determine which public configuration property gets exposed through MSBuild or not? We have some other properties that might make sense to expose here (e.g. GCConserveMem) while some other might not (e.g. RetainVM)

mangod9: yeah not quite sure about the history around this, given the naming differences I am guessing different teams ended up implementing this. We could inventory what needs to be added and happy to add them so we get some consistent naming here.

Maoni0: since this kind of tasks were not historically done by the GC team, my guess is what used to be in the hosting startup flags were made available in the msbuild options which meant Concurrent GC, Server GC and RetainVM (which was an important thing in the 32-bit days but no longer relevant in 64-bit).

agreed that we should do an inventory but that shouldn't delay the checkin of this change. perhaps open an issue on the inventory part?

Maoni0 (approved):
this looks reasonable to me but of course I have no experience at all with this :) so I can't tell if anything else needs to be changed or not.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
mangod9: yeah sounds good, I can create an issue to inventory settings. Once @mrsharm validates that the new runtimeconfig settings are working, I can merge this in. I have validated that the msbuild property adds the value withing runtimeconfig.

mangod9:
This csproj settings generate the runtimeconfig below:

```
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ServerGarbageCollection>true</ServerGarbageCollection>
    <GarbageCollectionAdaptationMode>1</GarbageCollectionAdaptationMode>
  </PropertyGroup>
```

```
{
  "runtimeOptions": {
    "tfm": "net8.0",
    "framework": {
      "version": "8.0.0-preview.6.23301.8"
    },
    "configProperties": {
      "System.GC.Server": true,
      "System.GC.DynamicAdaptationMode": 1,
      "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": false
    }
  }
}
```

mrsharm:
@mangod9, just tested this out and tl;dr: LGTM! 

With the following in the csproj:

```
   <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ServerGarbageCollection>true</ServerGarbageCollection>
    <GarbageCollectionAdaptationMode>1</GarbageCollectionAdaptationMode>
  </PropertyGroup>
```

confirmed that the runtimeconfig.json contained the following:

```
{
  "runtimeOptions": {
    "tfm": "net8.0",
    "framework": {
      "version": "8.0.0-preview.6.23306.10"
    },
    "configProperties": {
      "System.GC.Server": true,
      "System.GC.DynamicAdaptationMode": 1,
      "System.Reflection.Metadata.MetadataUpdater.IsSupported": false,
      "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": false
    }
  }
}
```

And the configuration API outputs the appropriate values:

```
ServerGC: True
ConcurrentGC: True
RetainVM: False
NoAffinitize: False
GCCpuGroup: False
GCLargePages: False
HeapCount: 20
GCHeapAffinitizeMask: 0
GCHeapAffinitizeRanges:
GCHighMemPercent: 0
GCHeapHardLimit: 0
GCHeapHardLimitPercent: 0
GCHeapHardLimitSOH: 0
GCHeapHardLimitLOH: 0
GCHeapHardLimitPOH: 0
GCHeapHardLimitSOHPercent: 0
GCHeapHardLimitLOHPercent: 0
GCHeapHardLimitPOHPercent: 0
GCConserveMem: 0
GCName:
**GCDynamicAdaptationMode: 1**
```

As an aside, that HeapCount needs to be adjusted based on the new value. 

The sample code to test this as follows:

```csharp

foreach (var kvp in System.GC.GetConfigurationVariables())
{
    Console.WriteLine($"{kvp.Key}: {kvp.Value}");
}
```

mrsharm (approved):

mrsharm (approved):

mangod9:
cool thanks for validating. assume you have also validated that the setting takes effect and heaps are dynamic. Would be good to add some functional test for this in runtime repo.

eerhardt:
> <GarbageCollectionAdaptationMode>1</GarbageCollectionAdaptationMode>

I wonder if we want actual names here. I was imagining either:

```xml
<GarbageCollectionAdaptationMode>Adaptive</GarbageCollectionAdaptationMode>
```

or 

```xml
<GarbageCollectionAdaptationMode>true</GarbageCollectionAdaptationMode>
```

Using `1` isn't consistent with the rest of MSBuild properties.

mangod9:
@eerhardt, the "mode" values were discussed in this PR conversation: https://github.com/dotnet/runtime/pull/87034/files/3b59d0a13185ee835a03a0a8b5d4a013e058cdc8#r1213780969

PR #33045, opened 2023-06-05
Don't run helix tests if the build fails

marcpopMSFT:
I noticed there was a PR where the build failed but we still tried to run helix tests. That doesn't make sense to me as it's just wasting resources. This was for Windows NT, Full Framework, Ubuntu, and Darwin legs that I saw.

See an example here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=295957&view=results

image


nagilson (approved):
Makes sense.. good change, thank you!

PR #33102, opened 2023-06-08
Fix duplicate error code

v-wuzhai:


src/Tasks/Common/Resources/Strings.resx
```diff
@@ -234,8 +234,8 @@
     <comment>{StrBegin="NETSDK1031: "}</comment>
   </data>
   <data name="CannotDirectlyReferenceMicrosoftNetCompilersToolsetFramework" xml:space="preserve">
-    <value>NETSDK1080: The Microsoft.Net.Compilers.Toolset.Framework package should not be set directly. Set the property 'BuildWithNetFrameworkHostedCompiler' to 'true' instead if you need it.</value>
-    <comment>{StrBegin="NETSDK1080: "}{Locked="Microsoft.Net.Compilers.Toolset.Framework"}{Locked="BuildWithNetFrameworkHostedCompiler"}</comment>
+    <value>NETSDK1203: The Microsoft.Net.Compilers.Toolset.Framework package should not be set directly. Set the property 'BuildWithNetFrameworkHostedCompiler' to 'true' instead if you need it.</value>
+    <comment>{StrBegin="NETSDK1203: "}{Locked="Microsoft.Net.Compilers.Toolset.Framework"}{Locked="BuildWithNetFrameworkHostedCompiler"}</comment>
```
marcpopMSFT, line 238: Thanks for catching this. We should consider some sort of automation to prevent this. Can you file an issue requesting us add that check (not sure how to add that check but I'll consult with others)?

Two notes:
1. Can you move the value into the spot in the resx file in line with the number (ie at the end). We don't do a great job of keeping these in order but I think it'll help avoid duplicates in the future if we try to make that a policy
2. 7.0.4xx also has this duplicate problem but there's already a 1203. I think we should make this 1205 as that's available in 4xx so the flow up to 4xx will update things correctly there (as we want it to be the same error in all branches for documentation)

marcpopMSFT (approved):

marcpopMSFT:
Looks good. Thanks for identifying this and getting it fixed!

PR #33132, opened 2023-06-09
Default container image tags to latest instead of versions

### LINKED ISSUE #458: Change default image tag from $(Version) to latest
To ease the getting-started flow, we should infer the `latest` tag instead of the `$(Version)` tag for generated images when no explicit `ContainerImageTag` is provided by the user. This will make it easier to round-trip test built images in container tech like Docker and Podman. Neither Jib nor ko force the same kind of default as we've done.

baronfel:
This is to make dev inner-loop and onboarding experiences better.

Closes dotnet/sdk-container-builds#458.


baronfel:
Reset this pr to target main because it is a breaking change in container tag inference, so we should not break that mid-cycle

MichalPavlik (approved):

baronfel:
Thanks for review @MichalPavlik - I'll do the docs update for this between now and the preview6 release in a few weeks.

PR #33146, opened 2023-06-09
Revert "[main] Update dependencies from dotnet/command-line-api (#29131)"

baronfel:
This reverts commit f8831a640e1876655e44dbf87a6e2f74f9c308a2.

We're reverting this because it blocks installer codeflow. We need to sychronize the changes with other repos that codeflow S.CL so that flow is blocked as little as possible.


src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
```diff
@@ -22,7 +22,7 @@ internal static class ValidatePackage
             bool enableRuleAttributesMustMatch,
             string[]? excludeAttributesFiles,
             bool enableRuleCannotChangeParameterName,
-            string? packagePath,
+            string packagePath,
```
ViktorHofer, line 25: This doesn't look right.

marcpopMSFT: I think this is a strict revert of the prior change.

marcpopMSFT (approved):

MichaelSimons (approved):

src/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
marcpopMSFT: Sorry, this is currently blocking codeflow to installer and we don't want to delay further. It'll come back with the revert of the revert.

PR #33155, opened 2023-06-09
Add test for catching duplicate error codes

### LINKED ISSUE #33133: Duplicate error code found
In 7.0.3、7.0.4、main branch src/Tasks/Common/Resources/Strings. resx file found repetitive error code NETSDK1080, I submitted a [pr](https://github.com/dotnet/sdk/pull/33102) to repair it for the time being.

 Problems such as duplicate error codes can be difficult to avoid due to lack of checks. In order to solve this problem, we may need a strategy to check for duplicate code.

cc @marcpopMSFT @dotnet/domestic-cat

marcpopMSFT:
The risk of this test is if two PRs go in in parallel, they could pass in PR but then fail afterwards.  The other risk is if we add a duplicate in a downlevel branch that doesn't have the test.

I'm open to other ways of catching this but I've seen this happen a few times now so it felt like it was worth adding the test. This will fail until https://github.com/dotnet/sdk/pull/33102 is merged and flows.

Fixes #33133

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantAGoodErrorExperience.cs
+++ b/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantAGoodErrorExperience.cs
@@ -1,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.NET.TestFramework;
+using Xunit;
+using Xunit.Abstractions;
+using Microsoft.NET.Build.Tasks;
+using System.Reflection;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.NET.Build.Tests
+{
+    public class GivenWeWantAGoodErrorExperience : SdkTest
+    {
+        public GivenWeWantAGoodErrorExperience(ITestOutputHelper log) : base(log)
+        {
+        }
+
+        [Fact]
+        public void ConfirmNoDuplicateErrorCodes()
+        {
+            Type type = typeof(Strings);
+            var resourceStrings = type.GetRuntimeProperties();
+            List<string> listOfErrorCode = new List<string>();
+            foreach (var resource in resourceStrings )
+            {
+                if ( resource.PropertyType == typeof(string))
+                {
+                    string errorCode = Strings.GetResourceString(resource.Name).Substring(0, 10);
+                    if (errorCode.Contains("NETSDK"))
+                    {
+                        listOfErrorCode.Add(errorCode);
+                    }
+                }
+            }
+            var anyDuplicate = listOfErrorCode.GroupBy(x => x).Where(g => g.Count() > 1);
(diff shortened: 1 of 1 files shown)


src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantAGoodErrorExperience.cs
```diff
@@ -0,0 +1,42 @@
...
+            var resourceStrings = type.GetRuntimeProperties();
+            List<string> listOfErrorCode = new List<string>();
+            foreach (var resource in resourceStrings )
+            {
+                if ( resource.PropertyType == typeof(string))
+                {
+                    string errorCode = Strings.GetResourceString(resource.Name).Substring(0, 10);
+                    if (errorCode.Contains("NETSDK"))
+                    {
+                        listOfErrorCode.Add(errorCode);
```
Forgind, line 34: If you make listOfErrorCodes a HashSet, you can say:
```suggestion
                        Assert.True(listOfErrorCode.Add(errorCode), $"Duplicate error code found: {errorCode}");
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantAGoodErrorExperience.cs
```diff
@@ -0,0 +1,42 @@
...
+        public void ConfirmNoDuplicateErrorCodes()
+        {
+            Type type = typeof(Strings);
+            var resourceStrings = type.GetRuntimeProperties();
+            List<string> listOfErrorCode = new List<string>();
+            foreach (var resource in resourceStrings )
+            {
+                if ( resource.PropertyType == typeof(string))
+                {
+                    string errorCode = Strings.GetResourceString(resource.Name).Substring(0, 10);
```
Forgind, line 31: Doesn't this throw if you have a string of length < 10? Maybe just check errorCode.StartsWith("NETSDK") then from there assume it's at least 10 characters long and add characters 6-10?

Forgind (commented):
I like this plan. I want to check that typeof(Strings).GetRuntimeProperties() returns all the errors and not just the ones in one assembly, for instance. (I haven't even checked if there are multiple strings.resx files in the SDK.)

marcpopMSFT:
Failed with the expected error as there is currently a duplicate. Will pass once https://github.com/dotnet/sdk/pull/33156 is merged.

dsplaisted:
This is already supposed to be covered in `GivenThatWeHaveErrorCodes` in Microsoft.NET.Build.Tasks.UnitTests.  I think the problem is that those unit tests aren't running in CI.

marcpopMSFT:
> This is already supposed to be covered in `GivenThatWeHaveErrorCodes` in Microsoft.NET.Build.Tasks.UnitTests. I think the problem is that those unit tests aren't running in CI.

I guess I could go look at the work item I have to get those working...

marcpopMSFT:
Ok, there are 13 failing unit tests. I'll put it onto my list to investigate those and then see if I can get it running in helix (think I've done it but would have to push the branch to find out).

marcpopMSFT:
Covered by this

PR #33163, opened 2023-06-10
Add a concept of image family to container base image inference

### LINKED ISSUE #401: Support `ContainerBaseImageTagSuffix` for only modifying the suffix of inferred image tags
May relate to #301.

Currently many containers have some sort of OS assumption baked into their 'default' tags, and the present various other decisions (OS, libc, etc) as sub-tags related to that base tag. Take the dotnet/aspnet image for example:
image

you've got:
* the default debian bookworm image
* alpine
* alpine 3.17 (same alpine version, just more explicit)
* slim debian bookworm
* ubuntu jammy
* slim ubuntu jammy

It would be great if we could keep most of the inference in place but allow users to switch OS/libc/etc with a suffix switch like `ContainerBaseImageTagSuffix`. The logic would be

* If inferring a base image at all (meaning the user didn't provide a ContainerBaseImage)
* and the ContainerBaseImageTagSuffix is present
* ContainerBaseImage should be set to the inferred `_ContainerBaseRegistry` followed by the inferred `_ContainerBaseImageName`, with the inferred `_ContainerBaseImageTag` and finally the `ContainerBaseImageTagSuffix`:

```xml
<ContainerBaseImage Condition="'$(_ContainerIsUsingMicrosoftDefaultImages)' == 'true'">$(_ContainerBaseRegistry)/$(_ContainerBaseImageName):$(_ContainerBaseImageTag)$(ContainerBaseImageTagSuffix)</ContainerBaseImage>
```

This works great for the MS-supplied/inferred images, but it wouldn't work so well in the general case (when a user wants to set a base image and separately manage the tag suffix), therefore if the user _isn't_ using our inference AND there is a ContainerBaseImage and ContainerBaseImageTagSuffix, then the two should be concat'd together.

baronfel:
This allows for easily selecting a group of related images (like chiseled or alpine) without having to 'eject' from inference overall. Right now we don't set this value anywhere, but further work might infer various properties and opt into chiseled ubuntu containers, for example. Or detect a musl RID and flip over to the alpine family of images.

Closes https://github.com/dotnet/sdk-container-builds/issues/401

This is pretty crucial for making it easier for users to opt into Chiseled containers for example - with this, a user could add

```xml
<ContainerFamily>jammy-chiseled</ContainerFamily>
```

to their project file and use the new, slim containers.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
@@ -8,6 +8,8 @@ internal static class KnownStrings
     public static class Properties
     {
         public static readonly string ContainerBaseImage = nameof(ContainerBaseImage);
+        public static readonly string ContainerFamily = nameof(ContainerFamily);
+        public static readonly string _ContainerBaseImageTag = nameof(_ContainerBaseImageTag);
         public static readonly string ContainerRegistry = nameof(ContainerRegistry);
         /// <summary>Note that this is deprecated in favor of <see cref="ContainerRepository"/></summary>
         public static readonly string ContainerImageName = nameof(ContainerImageName);
@@ -22,6 +24,8 @@ internal static class KnownStrings
         public static readonly string ContainerPort = nameof(ContainerPort);
         public static readonly string ContainerEnvironmentVariable = nameof(ContainerEnvironmentVariable);
 
+        public static readonly string ComputeContainerBaseImage = nameof(ComputeContainerBaseImage);
+        public static readonly string _ComputeContainerBaseImageTag = nameof(_ComputeContainerBaseImageTag);
         public static readonly string ComputeContainerConfig = nameof(ComputeContainerConfig);
         public static readonly string AssemblyName = nameof(AssemblyName);
         public static readonly string ContainerBaseRegistry = nameof(ContainerBaseRegistry);
--- a/src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
@@ -25,6 +25,8 @@ public sealed class ComputeDotnetBaseImageTag : Microsoft.Build.Utilities.Task
     [Required]
     public string TargetFrameworkVersion { get; set; }
 
+    public string ContainerFamily { get; set; }
+
     [Output]
     public string? ComputedBaseImageTag { get; private set; }
 
@@ -32,6 +34,7 @@ public sealed class ComputeDotnetBaseImageTag : Microsoft.Build.Utilities.Task
     {
         SdkVersion = "";
         TargetFrameworkVersion = "";
+        ContainerFamily = "";
     }
 
     public override bool Execute()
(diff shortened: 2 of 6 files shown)


baronfel:
Sorry to those pinged for review - I rebased from 7.0.4xx to main and so you all got notified.

richlander:
This is a great feature. Strong support for it.

It's also just as needed for:

```xml
<ContainerFamily>cbl-mariner2.0</ContainerFamily>
```

and

```xml
<ContainerFamily>cbl-mariner2.0-distroless</ContainerFamily>
```

src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
```diff
@@ -25,33 +25,40 @@ public sealed class ComputeDotnetBaseImageTag : Microsoft.Build.Utilities.Task
...
             ComputedBaseImageTag = ComputeVersionInternal(version, tfm);
-            return true;
         }
         else
         {
             Log.LogError(Resources.Strings.InvalidSdkVersion, SdkVersion);
-            return false;
+            return !Log.HasLoggedErrors;
+        }
+        if (!string.IsNullOrEmpty(ContainerFamily))
```
donJoseLuis, line 55: very minor: ensure you selected the proper test...

https://stackoverflow.com/questions/6976597/string-isnulloremptystring-vs-string-isnullorwhitespacestring

src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
```diff
@@ -25,33 +25,40 @@ public sealed class ComputeDotnetBaseImageTag : Microsoft.Build.Utilities.Task
...
+        {
+            // for the inferred image tags, 'family' aka 'flavor' comes after the 'version' portion (including any preview/rc segments).
+            // so it's safe to just append here
+            ComputedBaseImageTag += $"-{ContainerFamily}";
         }
+        return true;
```
donJoseLuis, line 61: confirm that this should return true even if parsin failed.

donJoseLuis (approved):
minor comments

src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
```diff
@@ -8,6 +8,8 @@ internal static class KnownStrings
     public static class Properties
     {
         public static readonly string ContainerBaseImage = nameof(ContainerBaseImage);
+        public static readonly string ContainerFamily = nameof(ContainerFamily);
+        public static readonly string _ContainerBaseImageTag = nameof(_ContainerBaseImageTag);
```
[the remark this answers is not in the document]
baronfel: I did this purposefully to mimic the name of the target we're testing - is that an allowable deviation?

src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
baronfel: Good catch, fixed just now.

src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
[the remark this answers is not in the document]
ladipro: I vote for this being allowable. It's not exposed publicly (public field in a public class nested in an **internal** class).

ladipro (approved):

baronfel:
Thanks for the review, folks!

richlander:
Nice. Will this be in RC1?

baronfel:
Should be - I believe cutoff is next Monday, so we're well under that.

PR #33166, opened 2023-06-10
[Containers] Enable multi-RID and multi-project containerization

baronfel:
Closes part of https://github.com/dotnet/sdk-container-builds/issues/87

This surfaces the containerization targets in another target that orchestrates the creation of zero or more containers.

When called at the solution level, so-called 'traversal' targets are created for each project, and each project's version of this new target is called in parallel.

For each project, there are a few multi-targeting scenarios that have to be addressed:

## How do you containerize a project with a single TFM and no RIDs (but potentially a single RID)?

This is the situation we have today - you just call the existing target.

## How do you containerize with multiple TFMs?

Easy, you ask each TFM to create its own container by calling the existing target once for each TFM.

## How do you containerize with multiple RIDs (and a single TFM)?

This is the novel thing, since we don't have a concept of multi-RID publishing buttoned down. Based on following patterns from MAUI, we can in parallel build the project with a different set of properties, and have those separate builds create containers for each architecture-specific variant of the container. We take the opportunity here to uniquely tag the containers with architecture-specific information, or else the container image names might overlap. Doing this tagging correctly required mapping .NET RIDs to Docker platform monikers, which was pulled out into a new Task for reuse, and also served to remove the need to rely on the RID graph from the SDK.

## Future work

This doesn't create manifest lists, which would 'wrap' each of the architecture-specific images in a wrapper that to external users looks like the same image regardless of architecture used.

## Use Cases

Microservices - now you can containerize an entire solution at once. Using `dotnet build /t:Containerize && docker-compose up` you can build new `latest` images of each of the projects, and if your Compose file is set up to use `image:` instead of `build:` then you are all ready to go!

## TODOs
* [x] re-introduce (limited) RID graph based selection based on constrained RID graphs (not the open-ended model of today) to enable compatibility with source-built SDKs
* [x] consider removing the multi-RID/multi-TFM publishing for now, and instead focus on the core scenario of enabling solution-level publish for all compatible projects
  * I don't think we can remove multi-TFM and multi-RID publishing if we want to support solution-level publish, because at a solution level you can't really control the TFM/RID selection for each and every project in the solution. So it'll have to stay in _some_ fashion, and we'll just have to evolve it responsibly.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/PlatformMapping.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/PlatformMapping.cs
@@ -1,0 +1,94 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.NET.Build.Containers;
+
+/// <summary>
+/// Handles mapping between .NET RIDs and Docker platform names/structures
+/// </summary>
+public static class PlatformMapping {
+
+    public static bool TryGetRidForDockerPlatform(string golangPlatform, bool isMuslBased, [NotNullWhen(true)] out string? runtimeIdentifier) {
+        runtimeIdentifier = null;
+
+        runtimeIdentifier = golangPlatform.Split('/') switch {
+            ["linux", "amd64"] when isMuslBased => "linux-musl-x64",
+            ["linux", "amd64"] => "linux-x64",
+
+            ["linux", "amd64", var _amd64Version] when isMuslBased => "linux-musl-x64",
+            ["linux", "amd64", var _amd64Version] => "linux-x64",
+
+            ["linux", "arm64"] when isMuslBased => "linux-musl-arm64",
+            ["linux", "arm64"] => "linux-arm64",
+
+            ["linux", "arm64", var _arm64Version ] when isMuslBased => "linux-musl-arm64",
+            ["linux", "arm64", var _arm64Version ] => "linux-arm64",
+
+            ["linux", "arm" ] or ["linux", "arm", "v7" ] when isMuslBased => "linux-musl-arm",
+            ["linux", "arm" ] or ["linux", "arm", "v7" ] => "linux-arm",
+
+            ["linux", "arm", "v6" ] when isMuslBased => "linux-musl-armv6",
+            ["linux", "arm", "v6" ]   => "linux-armv6",
+
+            ["linux", "riscv64" ] when isMuslBased => "linux-musl-riscv64",
+            ["linux", "riscv64" ] => "linux-riscv64",
+
+            ["linux", "ppc64le" ] when isMuslBased => "linux-musl-ppc64le",
+            ["linux", "ppc64le" ] => "linux-ppc64le",
+
+            ["linux", "s390x" ] when isMuslBased => "linux-musl-s390x",
+            ["linux", "s390x" ] => "linux-s390x",
+
+            ["linux", "386" ] when isMuslBased => "linux-musl-x86",
+            ["linux", "386" ] => "linux-x86",
+
+            ["windows", "amd64"] => "win-x64",
+            ["windows", "arm64"] => "win-arm64",
+
+            _ => null // other golang platforms are not supported
+        };
+        return runtimeIdentifier != null;
+    }
+
+    public static bool TryGetDockerPlatformForRid(string runtimeIdentifier, [NotNullWhen(true)] out string? dockerPlatform) {
+        dockerPlatform = null;
+
+        runtimeIdentifier = runtimeIdentifier.Replace("-musl-", "-"); // we lose musl information in the docker platform name
+
+        dockerPlatform = runtimeIdentifier.Split('-') switch {
+            ["linux", "x64"] => "linux/amd64",
+            ["linux", "arm64"] => "linux/arm64",
+            ["linux", "arm"] => "linux/arm/v7",
+            ["linux", "armv6"] => "linux/arm/v6",
+            ["linux", "riscv64"] => "linux/riscv64",
+            ["linux", "ppc64le"] => "linux/ppc64le",
+            ["linux", "s390x"] => "linux/s390x",
+            ["linux", "x86"] => "linux/386",
+            ["win", "x64"] => "windows/amd64",
+            ["win", "arm64"] => "windows/arm64",
+            _ => null
+        };
+        return dockerPlatform != null;
+    }
+
+    public static bool TryGetDockerImageTagForRid(string runtimeIdentifier, [NotNullWhen(true)] out string? dockerPlatformTag) {
+        dockerPlatformTag = null;
+
+        runtimeIdentifier = runtimeIdentifier.Replace("-musl-", "-"); // we lose musl information in the docker platform name
+
+        dockerPlatformTag = runtimeIdentifier.Split('-') switch {
+            ["linux", "x64"] => "amd64",
+            ["linux", "arm64"] => "arm64v8",
+            ["linux", "arm"] => "arm32v7",
+            ["linux", "armv6"] => "arm32v6",
+            ["linux", "riscv64"] => "riscv64",
+            ["linux", "ppc64le"] => "ppc64le",
+            ["linux", "s390x"] => "s390x",
+            ["linux", "x86"] => "386",
+            _ => null // deliberately not trying to make tag names for windows containers
+        };
+        return dockerPlatformTag != null;
+    }
+}
--- a/src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
@@ -192,35 +192,14 @@ internal sealed class Registry
         return null;
     }
 
-    private static string? CreateRidForPlatform(PlatformInformation platform)
+    private static string? CreateRidForPlatform(PlatformInformation platform, bool isMuslBased = false)
     {
-        // we only support linux and windows containers explicitly, so anything else we should skip past.
-        var osPart = platform.os switch
-        {
-            "linux" => "linux",
-            "windows" => "win",
-            _ => null
-        };
-        // TODO: this part needs a lot of work, the RID graph isn't super precise here and version numbers (especially on windows) are _whack_
-        // TODO: we _may_ need OS-specific version parsing. Need to do more research on what the field looks like across more manifest lists.
-        var versionPart = platform.version?.Split('.') switch
-        {
-            [var major, .. ] => major,
-            _ => null
-        };
-        var platformPart = platform.architecture switch
-        {
-            "amd64" => "x64",
-            "x386" => "x86",
-            "arm" => $"arm{(platform.variant != "v7" ? platform.variant : "")}",
-            "arm64" => "arm64",
-            "ppc64le" => "ppc64le",
-            "s390x" => "s390x",
-            _ => null
-        };
-
-        if (osPart is null || platformPart is null) return null;
-        return $"{osPart}{versionPart ?? ""}-{platformPart}";
+        var dockerPlatformString = $"{platform.os}/{platform.architecture}{(platform.variant is null ? "" : $"/{platform.variant}")}";
+        if (PlatformMapping.TryGetRidForDockerPlatform(dockerPlatformString, isMuslBased: isMuslBased, out string? rid)) {
+            return rid;
+        } else {
+            return null;
+        }
     }
 
     private static RuntimeGraph GetRuntimeGraphForDotNet(string ridGraphPath) => JsonRuntimeFormat.ReadRuntimeGraph(ridGraphPath);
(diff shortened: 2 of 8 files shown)


src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -216,11 +217,69 @@
...
+            but this is also nice and explicit.
+     -->
+
+    <!-- TFMs but no TF -> multitarget, making image for each TFM -->
+    <ItemGroup Condition="'$(TargetFrameworks)' != ''
+                          and '$(TargetFramework)' == ''" >
+      <_TFMItems Include="$(TargetFrameworks)" />
+      <_SingleContainerPublish Include="$(MSBuildProjectFullPath)"
+        AdditionalProperties="TargetFramework=%(_TFMItems.Identity);
+                              ContainerImageTag=$([MSBuild]::GetTargetFrameworkVersion('%(_TFMItems.Identity)', 2))" />
```
baronfel, line 256: Discussion: we need to disambiguate tags here because version/latest doesn't make unique container/tag pairs. I've gone with TFMs here. Is this useful, or should we sidestep/ignore/error the multi-TFM containerization case?

src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -216,11 +217,69 @@
...
+                                      and '$(RuntimeIdentifiers)' != ''"
+                            RuntimeIdentifiers="@(_RIDItems)">
+      <Output TaskParameter="ModifiedRuntimeIdentifiers" ItemName="_RIDItemsWithDockerPlatformTag" />
+    </MapRidToDockerPlatform>
+    <ItemGroup Condition="'$(TargetFramework)' != ''
+                          and '$(RuntimeIdentifiers)' != ''">
+      <_SingleContainerPublish Include="$(MSBuildProjectFullPath)"
+          AdditionalProperties="ContainerRuntimeIdentifier=%(_RIDItemsWithDockerPlatformTag.Identity);
+                                RuntimeIdentifier=%(_RIDItemsWithDockerPlatformTag.Identity);
+                                ContainerImageTag=$(Version)-%(_RIDItemsWithDockerPlatformTag.DockerPlatformTag);" />
```
baronfel, line 280: Here's an interesting consequence of moving to `latest` for image tag inference as a default. Before, `$(Version)` was used as the default tag, which meant we had something useful to hang a platform-specific bit off of. For example, `1.0.0-amd64` would make sense as a container image tag.

Now that we default to latest, you would see things like `latest-amd64`, which I have never seen as a pattern or practice before.  What I would _like_ to do is just set something like `ContainerImageTagSuffix` here and flow `Version` along, but I think there's even a bug with that - `$(Version)` at the point that this target is run may not be calculated yet. For example my OSS projects generate Version in a target that's hooked up to `PrepareForBuild` as a dependency. To me, this means we need to find a way to inject this on a per-image basis. I'm not sure how to tackle that one. Love to have any ideas.

src/Containers/Microsoft.NET.Build.Containers/PlatformMapping.cs
```diff
@@ -0,0 +1,94 @@
...
+            ["linux", "ppc64le" ] => "linux-ppc64le",
+
+            ["linux", "s390x" ] when isMuslBased => "linux-musl-s390x",
+            ["linux", "s390x" ] => "linux-s390x",
+
+            ["linux", "386" ] when isMuslBased => "linux-musl-x86",
+            ["linux", "386" ] => "linux-x86",
+
+            ["windows", "amd64"] => "win-x64",
+            ["windows", "arm64"] => "win-arm64",
```
baronfel, line 48: this is the overlap between our rid graph (which may change but is generally very static for these 'core' architectures) and the golang platform/arch combinations.

golang has a number of these that .NET doesn't support - but since we're focused on building _.NET_ applications we don't need complete support here.

src/Containers/Microsoft.NET.Build.Containers/PlatformMapping.cs
```diff
@@ -0,0 +1,94 @@
...
+            ["linux", "arm"] => "arm32v7",
+            ["linux", "armv6"] => "arm32v6",
+            ["linux", "riscv64"] => "riscv64",
+            ["linux", "ppc64le"] => "ppc64le",
+            ["linux", "s390x"] => "s390x",
+            ["linux", "x86"] => "386",
+            _ => null // deliberately not trying to make tag names for windows containers
+        };
+        return dockerPlatformTag != null;
+    }
```
baronfel, line 93: There's a slight different in format for the tags you'll see in the wild for docker architectures - there's an assumption of linux, first of all, so you drop that portion and then your just take the right hand side, with small differences for explicitly specifying versions in a couple places.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -189,25 +188,22 @@ private async Task<ImageBuilder> ReadSingleImageAsync(string repositoryName, Man
...
         cancellationToken.ThrowIfCancellationRequested();
-        var runtimeGraph = GetRuntimeGraphForDotNet(runtimeIdentifierGraphPath);
-        var (ridDict, graphForManifestList) = ConstructRuntimeGraphForManifestList(manifestList, runtimeGraph);
-        var bestManifestRid = CheckIfRidExistsInGraph(graphForManifestList, ridDict.Keys, runtimeIdentifier);
-        if (bestManifestRid is null) {
-            throw new BaseImageNotFoundException(runtimeIdentifier, repositoryName, reference, graphForManifestList.Runtimes.Keys);
-        }
-        var matchingManifest = ridDict[bestManifestRid];
-        var manifestResponse = await GetManifestAsync(repositoryName, matchingManifest.digest, cancellationToken).ConfigureAwait(false);
+        IReadOnlyDictionary<string, PlatformSpecificManifest> ridDict = ConstructRuntimeGraphForManifestList(manifestList);
```
baronfel, line 194: Because we have a more regimented way to map RIDs, and because we as a platform are going away from a flexible RID graph, we can remove our use of the graph in favor of using the new mapper to 
a) identify RIDs supported by the platform-specific manifests, and
b) find the RID that matches best

tmds: > Because we have a more regimented way to map RIDs, and because we as a platform are going away from a flexible RID graph, we can remove our use of the graph in favor of using the new mapper to
a) identify RIDs supported by the platform-specific manifests, and
b) find the RID that matches best

For non-portable rids used by source-build SDKs we still want the mapping to work from the non-portable rid to the portable rid (which then corresponds to an image platform os+arch).

That requires using the bundled runtime graph file.

baronfel: Good point, the conversation internally about the need for _some_ kind of graph did change over the past couple days. So a portion of this will likely come back, just with a different source graph.

src/Containers/Microsoft.NET.Build.Containers/PlatformMapping.cs
```diff
@@ -0,0 +1,94 @@
...
+            ["linux", "arm", "v6" ]   => "linux-armv6",
+
+            ["linux", "riscv64" ] when isMuslBased => "linux-musl-riscv64",
+            ["linux", "riscv64" ] => "linux-riscv64",
+
+            ["linux", "ppc64le" ] when isMuslBased => "linux-musl-ppc64le",
+            ["linux", "ppc64le" ] => "linux-ppc64le",
+
+            ["linux", "s390x" ] when isMuslBased => "linux-musl-s390x",
+            ["linux", "s390x" ] => "linux-s390x",
```
baronfel, line 42: @tmds this should cover the same as your PR, right? so I've inadvertently solved the issue you saw?

tmds: Yes, thank you!

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
tmds: I was trying this out today, and the current code wasn't able to map a non-portable rid to a portable rid.
I haven't checked yet if the pebkac, but anyway could we use this function here:

https://github.com/dotnet/sdk/blob/741b910599db7f6a1a409fe70937af0f76caa437/src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs#L83-L84

Given a runtime graph, the target runtimeIdentifier, and the rids that correspond to the image platforms it will tell you what the best rid is.

baronfel: The only concern I have there is inflating the dependencies of this tooling - would need to check that packaging still works correctly. But I think the tests @vlada-shubina added around package layout would help us detect any problems.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -233,58 +229,28 @@ private async Task<HttpResponseMessage> GetBlobAsync(string repositoryName, stri
...
             }
         }
-
-        var graph = new RuntimeGraph(runtimeDescriptionSet);
-        return (ridDict, graph);
+        return ridDict;
     }
 
-    private static string? CreateRidForPlatform(PlatformInformation platform)
+    private static string? CreateRidForPlatform(PlatformInformation platform, bool isMuslBased = false)
```
tmds, line 244: Maybe this `isMuslBased` can also be removed when the rid graph is used again?
(The graph maps `linux-musl-x64` to `linux-x64` which corresponds to an image platform.)

tmds: You can use the `NETCoreSdkPortableRuntimeIdentifier` property and check if it starts with `linux-musl-`.

tmds: Ignore that. This is the target rid.

Either it's `linux-musl-{arch}` as the portable rid.
Or if it's not portable, you can run it through the runtime graph and see if matches `linux-musl`.

baronfel: Yeah, that's a big part of it. I was planning to leave it for a different PR to do a full treatment, it just felt wrong to not acknowledge the concept in this PR

tmds: I don't think the registry needs to know about `isMuslBased`.

Like a non-portable rid, the runtime graph will map `linux-musl` to the most compatible rid provided by a container image (which is: `linux`).

tmds: The problem is that images don't describe if they are musl based.

The sdk could know for some images (in particular: the Microsoft ones) that they are musl-based based on their name.

baronfel: Yeah, that's the kind of determination I was thinking about hooking up later. For the MS images we know a family called `alpine` exists so if the user asked for a musl RID we could auto-insert that. Anything else we probably can't rely on - we just have to accept that a user knows what they are doing if the eject out of inference.

tmds: Yes. That could be handled by the `ComputeDotnetBaseImageTag` Task if we pass it the `RuntimeIdentifier`.

src/Containers/Microsoft.NET.Build.Containers/Registry.cs
baronfel: Resolving this conversation as I just merged this enhancement in another PR you've sent.

src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets
baronfel: @rainersigwald hate to ping, but I had a design question that I'd love your input on if you have a moment.

baronfel:
Re-targeted to main for 8.0.100

src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
```diff
@@ -192,35 +192,14 @@ private async Task<ImageBuilder> ReadSingleImageAsync(string repositoryName, Man
...
-            "arm64" => "arm64",
-            "ppc64le" => "ppc64le",
-            "s390x" => "s390x",
-            _ => null
-        };
-
-        if (osPart is null || platformPart is null) return null;
-        return $"{osPart}{versionPart ?? ""}-{platformPart}";
+        var dockerPlatformString = $"{platform.os}/{platform.architecture}{(platform.variant is null ? "" : $"/{platform.variant}")}";
+        if (PlatformMapping.TryGetRidForDockerPlatform(dockerPlatformString, isMuslBased: isMuslBased, out string? rid)) {
```
tmds, line 198: The result of this method is used to pick the proper base image for the target rid.
This base image is the same one for musl-based and glibc-based image rids.

We can remove the `isMuslBased` argument here because it doesn't affect the result.

src/Containers/Microsoft.NET.Build.Containers/PlatformMapping.cs
```diff
@@ -0,0 +1,94 @@
...
+        dockerPlatformTag = runtimeIdentifier.Split('-') switch {
+            ["linux", "x64"] => "amd64",
+            ["linux", "arm64"] => "arm64v8",
+            ["linux", "arm"] => "arm32v7",
+            ["linux", "armv6"] => "arm32v6",
+            ["linux", "riscv64"] => "riscv64",
+            ["linux", "ppc64le"] => "ppc64le",
+            ["linux", "s390x"] => "s390x",
+            ["linux", "x86"] => "386",
+            _ => null // deliberately not trying to make tag names for windows containers
```
tmds, line 90: Can you add support for non-portable rids to this method?
It could be to return `null` for anything that starts with `win`, and return the currently mapped values based on ending with the arch suffix.

An alternative to map from rids could be to map from docker platforms.

baronfel:
Closing as this has drifted quite a bit.

PR #33189, opened 2023-06-12
Swap the 1205 and 1206 error codes 

marcpopMSFT:
3xx and 4xx already had the com…piler framework error at 1205. This stays consistent with those branches.


marcpopMSFT:
@elinor-fung was there already documentation created around NonPortableRuntimeIdentifierDetected? Mind swapping the error numbers? We had CannotDirectlyReferenceMicrosoftNetCompilersToolsetFramework go into 7.0.3xx as 1205 late last week as that was the highest number at the time and we want to be consistent between releases so the documentation is consistent.

Forgind (approved):

elinor-fung (approved):
No doc referencing the error number. The doc link just points the breaking change notice right now. Looks good to me.

PR #33195, opened 2023-06-13
[release/6.0.4xx] Fix error code to maintain consistency

### LINKED ISSUE #33196: Fix error code to maintain consistency
Refer #33189, I noticed that there were still other inconsistent error codes. We want to unify those inconsistent error codes to facilitate the consistency of different branch documents.

@marcpopMSFT

v-wuzhai:
Fixes #33196

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -833,16 +833,16 @@ You may need to build the project on another operating system or architecture, o
     <comment>{StrBegin="NETSDK1182: "}</comment>
   </data>
   <data name="TargetingPackNotRestored_TransitiveDisabled" xml:space="preserve">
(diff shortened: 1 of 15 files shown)


marcpopMSFT (approved):

v-wuzhai:
> @v-wuzhai Thanks for catching these and getting them fixed. I'm not sure how to have a test for this in the future. Mind prepping a PR to update the primary doc list as well: https://github.com/dotnet/docs/blob/main/docs/core/tools/sdk-errors/index.md

Hi @marcpopMSFT, thanks for your reminder, looks like the documentation is good, so we don't need to update the documentation.

marcpopMSFT:
To clarify, I meant adding the new error codes to that document as it's a bit out of date. We should probably do that with each quarterly release to make sure the document stays aligned with regular updates.  We created that main page so it'd be easier for customers to search and so any customer hitting F1 (navigate to help) in VS would have a landing page.

PR #33225, opened 2023-06-14
Fix R2R map generation on OSX

trylek:
As @AArnott noticed, Crossgen2 targeting OSX doesn't produce any form of symbol information. I believe it should produce the R2R map (as the only other option - PDB - is Windows-specific and unsupported on non-Windows OSes) but that was blocked in the SDK by upfront checks for the Linux targeting OS. I propose removing these checks so that the R2R map logic can apply for all non-Windows OSes.

Thanks

Tomas

/cc @dotnet/crossgen-contrib

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
@@ -78,18 +78,6 @@ namespace Microsoft.NET.Build.Tasks
             }
         }
 
-        private bool IsTargetLinux
-        {
-            get
-            {
-                // Crossgen2 V6 and above always has TargetOS metadata available
-                if (ReadyToRunUseCrossgen2 && !string.IsNullOrEmpty(Crossgen2Tool.GetMetadata(MetadataKeys.TargetOS))) 
-                    return Crossgen2Tool.GetMetadata(MetadataKeys.TargetOS) == "linux";
-                else
-                    return RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
-            }
-        }
-
         protected override void ExecuteCore()
         {
             if (ReadyToRunUseCrossgen2)
@@ -170,7 +158,7 @@ namespace Microsoft.NET.Build.Tasks
                         outputPDBImageRelativePath = Path.ChangeExtension(outputR2RImageRelativePath, "ni.pdb");
                         crossgen1CreatePDBCommand = $"/CreatePDB \"{Path.GetDirectoryName(outputPDBImage)}\"";
                     }
-                    else if (IsTargetLinux)
+                    else
                     {
                         string perfmapExtension;
                         if (ReadyToRunUseCrossgen2 && !_crossgen2IsVersion5 && _perfmapFormatVersion >= 1)
@@ -275,7 +263,7 @@ namespace Microsoft.NET.Build.Tasks
                         compositePDBImage = Path.ChangeExtension(compositeR2RImage, ".ni.pdb");
                         compositePDBRelativePath = Path.ChangeExtension(compositeR2RImageRelativePath, ".ni.pdb");
                     }
-                    else if (IsTargetLinux)
+                    else
                     {
                         string perfmapExtension = (_perfmapFormatVersion >= 1 ? ".ni.r2rmap" : ".ni.{composite}.map");
                         compositePDBImage = Path.ChangeExtension(compositeR2RImage, perfmapExtension);
--- a/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
@@ -281,7 +281,7 @@ namespace Microsoft.NET.Build.Tasks
                             compositePDBRelativePath = Path.ChangeExtension(compositeR2RImageRelativePath, ".ni.pdb");
                         }
                     }
-                    else if (!_crossgen2IsVersion5 || IsTargetLinux)
+                    else if ((ReadyToRunUseCrossgen2 && !_crossgen2IsVersion5) || IsTargetLinux)
                     {
                         string perfmapExtension = (_perfmapFormatVersion >= 1 ? ".ni.r2rmap" : ".ni.{composite}.map");
                         compositePDBImage = Path.ChangeExtension(compositeR2RImage, perfmapExtension);
--- a/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
@@ -173,7 +173,7 @@ namespace Microsoft.NET.Build.Tasks
                             crossgen1CreatePDBCommand = $"/CreatePDB \"{Path.GetDirectoryName(outputPDBImage)}\"";
                         }
                     }
-                    else if (!_crossgen2IsVersion5 || IsTargetLinux)
+                    else if ((ReadyToRunUseCrossgen2 && !_crossgen2IsVersion5) || IsTargetLinux)
                     {
                         string perfmapExtension;
                         if (ReadyToRunUseCrossgen2 && !_crossgen2IsVersion5 && _perfmapFormatVersion >= 1)
--- a/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
@@ -152,11 +152,14 @@ namespace Microsoft.NET.Build.Tasks
 
                 if (EmitSymbols)
                 {
-                    if (IsTargetWindows && hasValidDiaSymReaderLib)
+                    if (IsTargetWindows)
                     {
-                        outputPDBImage = Path.ChangeExtension(outputR2RImage, "ni.pdb");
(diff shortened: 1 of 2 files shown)


mangod9 (approved):

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
```diff
@@ -170,7 +158,7 @@ protected override void ExecuteCore()
                         outputPDBImageRelativePath = Path.ChangeExtension(outputR2RImageRelativePath, "ni.pdb");
                         crossgen1CreatePDBCommand = $"/CreatePDB \"{Path.GetDirectoryName(outputPDBImage)}\"";
                     }
-                    else if (IsTargetLinux)
+                    else
```
jkotas, line 161: I do not think there is any tooling on Windows that is capable of consuming the perfmap files.

It probably does not make sense to produce the perfmap file when we are on Windows and we do not have good diasymreader.

trylek: OK, sounds reasonable. I initially thought it might be useful for various experimental purposes but you're probably right that it's better to keep the logic simple and consistent. I have modified my change based on your feedback in the 2nd commit.

AArnott (approved):

AArnott:
BTW, the docs suggest r2rmap only applies to linux. No mention of mac. [ReadyToRun deployment overview - .NET | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/core/deploying/ready-to-run#symbol-generation-for-use-with-profilers). 
Maybe they should be updated as well (separate to this PR).

hoyosjs:
@trylek @AArnott is there tooling you are thinking of for macOS? Or is this for adhoc. There's no native tooling that I know.

AArnott:
I am not aware of any tooling for r2rmap files at all. But we're trying to get a good symbol story together for our .NET 7 processes on all OSs, and as r2rmap files are the best we get, we noticed that mac had no story at all, which is what drew attention to this.

trylek:
Thanks @AArnott for pointing out the doc, I guess I'll just need to figure out how to update it.

AArnott:
@trylek the source for that doc is here: https://github.com/dotnet/docs/blob/main/docs/core/deploying/ready-to-run.md

AntonLapounov:
Before .NET 6 we used the old crossgen for generating symbols.  However, the macOS version of the old crossgen does not support the `/CreatePerfMap` switch, which leads to test failures.

It seems we have not updated R2R tests to target recent .NET releases.  For instance, the test below does not target anything higher than .NET 5.  You might want to add `[InlineData(ToolsetInfo.CurrentTargetFramework]` there to actually test your changes.  Other tests should also be updated in a separate change.
https://github.com/dotnet/sdk/blob/482495096f715aad609908dc9e96e61c137ec673/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishReadyToRun.cs#L106-L109

trylek:
Hi @AntonLapounov! Great to hear from you after a while. Thanks for your feedback but I'm not sure if I'm fully grasping it - the section you quoted does include the test variant for <code>[InlineData(ToolsetInfo.CurrentTargetFramework]</code> that should be <code>net8.0</code> in the current SDK. Having said that, I fully agree that we should add <code>net6.0</code> and <code>net7.0</code> as explicit "older .NET versions" to all these tests - I agree that such a change will be easier to make separately.

AntonLapounov:
Oh, you are correct.  Not sure how I quoted that line and thought it was missing at the same time.  Probably I was looking at the log output and did not find this test executed for .NET 8 (which, I guess, is expected as we log failures only).

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
```diff
@@ -164,13 +164,16 @@ protected override void ExecuteCore()
...
-                        crossgen1CreatePDBCommand = $"/CreatePDB \"{Path.GetDirectoryName(outputPDBImage)}\"";
+                        if (hasValidDiaSymReaderLib)
+                        {
+                            outputPDBImage = Path.ChangeExtension(outputR2RImage, "ni.pdb");
+                            outputPDBImageRelativePath = Path.ChangeExtension(outputR2RImageRelativePath, "ni.pdb");
+                            crossgen1CreatePDBCommand = $"/CreatePDB \"{Path.GetDirectoryName(outputPDBImage)}\"";
+                        }
                     }
-                    else if (IsTargetLinux)
+                    else if (!_crossgen2IsVersion5 || IsTargetLinux)
```
AntonLapounov, line 176: Should this be
```suggestion
                    else if (ReadyToRunUseCrossgen2 && !_crossgen2IsVersion5 || IsTargetLinux)
```
as we cannot generate symbols on macOS using the old cressgen?

trylek: Hmm, how come it doesn't hit the condition <code>!_crossgen2IsVersion5</code>?

trylek: I think I'm starting to get it, let me try it out...

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishReadyToRun.cs
```diff
@@ -411,9 +406,9 @@ private static bool IsTargetOsLinux(string runtimeIdentifier)
             else
                 publishDirectory.Should().NotHaveFile("System.Private.CoreLib.dll");
 
-            if (emitNativeSymbols && !IsTargetOsOsX(testProject.RuntimeIdentifier))
+            NuGetFramework framework = NuGetFramework.Parse(targetFramework);
+            if (emitNativeSymbols && (!IsTargetOsOsX(runtimeIdentifier) || framework.Version.Major >= 6))
```
AntonLapounov, line 410:
```suggestion
            if (emitNativeSymbols && (!IsTargetOsOsX(testProject.RuntimeIdentifier) || framework.Version.Major >= 6))
```
The passed `runtimeIdentifier` may be `null`, in which case it is inferred.

AntonLapounov (approved):

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
AntonLapounov: If we are using the old crossgen on macOS, `_crossgen2IsVersion5` is `false`, we execute this branch and use the unsupported `/CreatePerfMap` switch.

trylek: Thanks Anton for explaining, I have modified the source file based on your suggestion in the 9th commit.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishReadyToRun.cs
trylek: Fixed in 9th commit, thanks Anton for pointing that out!

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
```diff
@@ -270,12 +273,15 @@ protected override void ExecuteCore()
...
-                        compositePDBImage = Path.ChangeExtension(compositeR2RImage, ".ni.pdb");
-                        compositePDBRelativePath = Path.ChangeExtension(compositeR2RImageRelativePath, ".ni.pdb");
+                        if (hasValidDiaSymReaderLib)
+                        {
+                            compositePDBImage = Path.ChangeExtension(compositeR2RImage, ".ni.pdb");
+                            compositePDBRelativePath = Path.ChangeExtension(compositeR2RImageRelativePath, ".ni.pdb");
+                        }
                     }
-                    else if (IsTargetLinux)
+                    else if ((ReadyToRunUseCrossgen2 && !_crossgen2IsVersion5) || IsTargetLinux)
```
AntonLapounov, line 284: This code path is already under `if (Crossgen2Composite)`, so checking `ReadyToRunUseCrossgen2` is optional here.  You changed it in iteration 9 instead of the other place.

trylek: OK, thanks for the heads-up. I believe that the composite build mode was only made official in .NET 6 so that we shouldn't actually need any extra check in the else block, I'll make the change in an extra commit.

trylek:
Thanks to Anton's insightful feedback I've been finally able to achieve a green CI run, merging in.

src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs
AntonLapounov: Sorry for the late reply.  On one hand, we default to the composite mode for .NET 5 if `PublishReadyToRunUseCrossgen2` is set: https://github.com/dotnet/sdk/blob/f5632207158c5800208d3482f5ed163e7dc54932/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.CrossGen.targets#L15-L16 On the other hand, generating the composite PDB in this method is guarded by the check at line 291, which requires targeting .NET 6 or higher: https://github.com/dotnet/sdk/blob/f5632207158c5800208d3482f5ed163e7dc54932/src/Tasks/Microsoft.NET.Build.Tasks/PrepareForReadyToRunCompilation.cs#L284-L291 That means your change is correct even for the non-official composite mode in .NET 5.

PR #33227, opened 2023-06-14
Allow MSBuild to use BinaryFormatter

rainersigwald:
With EnableUnsafeBinaryFormatterSerialization=false, MSBuild can crash
instead of passing along an error, dotnet/msbuild#8786. Temporarily
opt back into the old behavior while fixing this in MSBuild.

### DIFF
--- a/src/Layout/redist/targets/GenerateLayout.targets
+++ b/src/Layout/redist/targets/GenerateLayout.targets
@@ -438,6 +438,9 @@
     <PropertyGroup>
       <ReplacementPattern>"version": ".*"</ReplacementPattern>
       <ReplacementString>"version": "$(MicrosoftNETCoreAppRuntimePackageVersion)"</ReplacementString>
+
+      <DisabledBinaryFormatterPattern>"System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": false</DisabledBinaryFormatterPattern>
+      <EnableBinaryFormatterString>"System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": true</EnableBinaryFormatterString>
     </PropertyGroup>
     <ItemGroup>
       <!-- Exclude F# from retargeting: https://github.com/dotnet/toolset/issues/2866 -->
@@ -452,6 +455,14 @@
       DestinationFiles="@(ToolRuntimeConfigPath)"
       ReplacementPatterns="$(ReplacementPattern)"
       ReplacementStrings="$(ReplacementString)" />
+
+    <!-- Work around https://github.com/dotnet/msbuild/issues/6215 by opting back into net7 behavior -->
+    <ReplaceFileContents
+      InputFiles="@(ToolRuntimeConfigPath)"
+      DestinationFiles="@(ToolRuntimeConfigPath)"
+      ReplacementPatterns="$(DisabledBinaryFormatterPattern)"
+      ReplacementStrings="$(EnableBinaryFormatterString)"
+      Condition=" '%(ToolRuntimeConfigPath.FileName)' == 'MSBuild.runtimeconfig'" />
 
     <Move
       SourceFiles="@(MSBuild15Items)"


rainersigwald:
Crashes caused by the BinaryFormatter changes have been problematic for ASP.NET, so let's get an SDK out that fixes them since https://github.com/dotnet/msbuild/pull/8779 has been a bit more difficult than we expected.

donJoseLuis (approved):

src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -453,6 +456,14 @@
       ReplacementPatterns="$(ReplacementPattern)"
       ReplacementStrings="$(ReplacementString)" />
 
+    <!-- Work around https://github.com/dotnet/msbuild/issues/6215 by opting back into net7 behavior -->
+    <ReplaceFileContents
+      InputFiles="@(ToolRuntimeConfigPath)"
+      DestinationFiles="@(ToolRuntimeConfigPath)"
+      ReplacementPatterns="$(DisabledBinaryFormatterPattern)"
```
JanKrivanek, line 463: I'm failing to find where this is being populated by the original value that we are replacing - is the key allways expected to be present?

src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -453,6 +456,14 @@
...
+      ReplacementStrings="$(EnableBinaryFormatterString)"
+      Condition=" '%(ToolRuntimeConfigPath.FileName)' == 'MSBuild.runtimeconfig'" />
```
JanKrivanek, line 465: Shouldn't this be `MSBuild.runtimeconfig.json`?

JanKrivanek (approved):
Let's ship this!

As soon as my questions are irrelevant - I have no objections

src/Layout/redist/targets/GenerateLayout.targets
JanKrivanek: Thank you @KalleOlaviNiemitalo!

src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -438,6 +438,9 @@
     <PropertyGroup>
       <ReplacementPattern>"version": ".*"</ReplacementPattern>
       <ReplacementString>"version": "$(MicrosoftNETCoreAppRuntimePackageVersion)"</ReplacementString>
+
+      <DisabledBinaryFormatterPattern>"System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": false</DisabledBinaryFormatterPattern>
+      <EnableBinaryFormatterString>"System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": true</EnableBinaryFormatterString>
```
Forgind, line 443: Do true and false need to be quoted?

Forgind (approved):

src/Layout/redist/targets/GenerateLayout.targets
baronfel: correct - true/false are literals in javascript and don't need quote-wrapping

GrabYourPitchforks:
I'm not quite sure if there's something special about the runtimeconfig.json logic here, but _in theory_ all you need to do is put `<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>` in your .csproj and it will emit the correct runtimeconfig.json on your behalf. Is that not a viable workaround here?

rainersigwald:
@GrabYourPitchforks this is the moral equivalent of that. Things that ship _as tools in the SDK_ are more convoluted because they're retargeted to the runtime that the SDK will ship with, and one side effect of that was inheriting the latest/greatest .NET 8 rules.

GrabYourPitchforks (approved):

GrabYourPitchforks:
I should also point out that the SDK's runtimeconfig.json including an explicit _false_ value by default is an error. See also: https://github.com/dotnet/sdk/issues/32969

Hopefully that will also make life a little less painful for you once that is addressed.

PR #33228, opened 2023-06-14
Fix duplicate NETSDK1183

### LINKED ISSUE #33196: Fix error code to maintain consistency

v-wuzhai:
Fixes #33196

The NETSDK1183 of the main branch is different from other branches.

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -835,10 +835,6 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1182: Targeting .NET 6.0 or higher in Visual Studio 2019 is not supported.</value>
     <comment>{StrBegin="NETSDK1182: "}</comment>
   </data>
-  <data name="AotUnsupportedTargetFramework" xml:space="preserve">
-    <value>NETSDK1183: Ahead-of-time compilation is not supported for the target framework.</value>
-    <comment>{StrBegin="NETSDK1183: "}</comment>
-  </data>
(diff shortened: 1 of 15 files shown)


marcpopMSFT:
This message was changed in https://github.com/dotnet/sdk/pull/32943 from Sven. Looks like he took one error and split into three and reused the existing error code. I think this PR is correct to take as then documentation on 7.0 SDKs can include the old 1183 error and documentation in 8.0 will have the new error under a different number but I want @sbomer to review in case there is any existing documentation that refers to these errors (new or old).

sbomer (approved):
LGTM - the only docs I'm aware of reference the old text of `NETSDK1183` (https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/).

marcpopMSFT:
Thanks. That should work then as the doc will be for the older version of the SDK. @tdykstra how did you want to mark NETSDK errors that have been removed in later SDKs in that document?

tdykstra:
> @tdykstra how did you want to mark NETSDK errors that have been removed in later SDKs in that document?

In that document we could do something like this:

| SDK Message Number | Message            |
|--------------------|--------------------|
|NETSDK1183|**.NET 7 only** - Unable to optimize assemblies for Ahead of time compilation: a valid runtime package was not found. Either set the PublishAot property to false, or use a supported runtime identifier when publishing. When targeting .NET 7 or higher, make sure to restore packages with the PublishAot property set to true.|

PR #33229, opened 2023-06-14
Don't set UseApphost or SelfContained to true for PublishAot

### LINKED ISSUE #79575: Publishing as NativeAOT should not depend on AppHost
We currently need to build apphost because for some reason PublishAot ends up needing it in some of our test legs:

https://github.com/dotnet/runtime/blob/cdb9f5f179db22745fbb2e878460662c0b21643e/eng/pipelines/runtime.yml#L206

Without this line, the CI fails with:

```
Could not copy the file "/__w/1/s/artifacts/bin/trimmingTests/projects/System.Diagnostics.DiagnosticSource.NativeAotTests/DiagnosticSourceEventSourceTests/linux-x64/obj/Release/net8.0/linux-x64/apphost" because it was not found.
```

This is probably not just a test issue, but also a product issue. Historically the NativeAOT targets were just an addon to the SDK targets that hacked into the SDK (we e.g. delete outputs the SDK produces in publish and replace them with NativeAOTd stuff).

The SDK probably needs to become more aware of what NativeAOT does and suppress some behaviors.

MichalStrehovsky:
Fixes dotnet/runtime#79575.

`UseAppHost` doesn't make any sense for `PublishAot`, and neither do the behaviors around `SelfContained` - if `SelfContained` is true, the bin directory gets populated with a ton of garbage files we don't need during publish, and the publish process starts to have opinions about whether UseAppHost=true should be specified.

Cc @dotnet/ilc-contrib @ViktorHofer

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -69,8 +69,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                              $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '8.0')) and
                               (
                                   '$(PublishTrimmed)' == 'true' or
-                                  '$(PublishSingleFile)' == 'true' or
-                                  '$(PublishAot)' == 'true'
+                                  '$(PublishSingleFile)' == 'true'
                               )">
     <PublishSelfContained>true</PublishSelfContained>
   </PropertyGroup>
@@ -173,6 +172,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                               $([MSBuild]::VersionLessThan($(_TargetFrameworkVersionWithoutV), '8.0'))">true</SelfContained>
     
     <SelfContained Condition="'$(SelfContained)' == ''">false</SelfContained>
+    <SelfContained Condition="'$(PublishAot)' == 'true' and '$(_IsPublishing)' == 'true'">false</SelfContained>
     <_RuntimeIdentifierUsesAppHost Condition="$(RuntimeIdentifier.StartsWith('ios')) or $(RuntimeIdentifier.StartsWith('tvos')) or $(RuntimeIdentifier.StartsWith('maccatalyst')) or $(RuntimeIdentifier.StartsWith('android')) or $(RuntimeIdentifier.StartsWith('browser'))">false</_RuntimeIdentifierUsesAppHost>
     <_RuntimeIdentifierUsesAppHost Condition="'$(_RuntimeIdentifierUsesAppHost)' == ''">true</_RuntimeIdentifierUsesAppHost>
     <UseAppHost Condition="'$(UseAppHost)' == '' and
@@ -181,6 +181,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                             ('$(RuntimeIdentifier)' != '' and '$(_TargetFrameworkVersionWithoutV)' >= '2.1') or
                             '$(_TargetFrameworkVersionWithoutV)' >= '3.0')">true</UseAppHost>
     <UseAppHost Condition="'$(UseAppHost)' == ''">false</UseAppHost>
+    <UseAppHost Condition="'$(PublishAot)' == 'true' and '$(_IsPublishing)' == 'true'">false</UseAppHost>
   </PropertyGroup>
 
   <!-- Only use the default apphost if building without a RID and without a deps file path (used by GenerateDeps.proj for CLI tools). -->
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
```diff
@@ -173,6 +172,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                               $([MSBuild]::VersionLessThan($(_TargetFrameworkVersionWithoutV), '8.0'))">true</SelfContained>
     
     <SelfContained Condition="'$(SelfContained)' == ''">false</SelfContained>
+    <SelfContained Condition="'$(PublishAot)' == 'true' and '$(_IsPublishing)' == 'true'">false</SelfContained>
```
vitek-karas, line 175: Technically we should set `PublishSelfContained=true` as the app will be self-contained and some things might change behavior based on that. But that setting is also broken :-(
I just wonder if setting it to false will break something somewhere because it will now think we're publishing an FDD app.

vitek-karas (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
MichalStrehovsky: The problem is that we're not doing either - I couldn't find a single behavior of `SelfContained=true` in the SDK that looks desirable for PublishAot. I don't think FDD behaviors are desirable either but I don't see any of them getting in the way the way SelfContained is so FDD just looks like a path of least resistance. The nice side effect of setting SelfContained to false is that publish is becoming faster because we no longer copy 100 files to the bin directory that Defender needs to scan.

ViktorHofer:
Just making sure that you saw @jkotas comment in the tracking issue: https://github.com/dotnet/runtime/issues/79575#issuecomment-1348757248

> We do want to keep using app host for dotnet run or F5 in Visual Studio. I do not think we want to change the global default to false when PublishAot is set.

MichalStrehovsky:
> Just making sure that you saw @jkotas comment in the tracking issue: [dotnet/runtime#79575 (comment)](https://github.com/dotnet/runtime/issues/79575#issuecomment-1348757248)
> 
> > We do want to keep using app host for dotnet run or F5 in Visual Studio. I do not think we want to change the global default to false when PublishAot is set.

That's what `_IsPublishing` is supposed to guard against.

steveisok:
I understand the reason for the change, but nativeaot is self contained.  Are we doing something out of order in the build?

agocke:
+ @dsplaisted @nagilson

MichalStrehovsky:
> I understand the reason for the change, but nativeaot is self contained. Are we doing something out of order in the build?

The problem is what SelfContained does in the Sdk - and we don't want pretty much any of those behaviors.

I still want to explore an alternative approach (allow SelfContained=true and just suppress the useless behaviors. If so just passing `--self-contained` to publish with aot doesn't error out. I just didn't have time to come back to this yet.

nagilson:
_IsPublishing will never activate in VS and only when doing dotnet publish in the SDK, is this the behavior you had expected?

cc @richlander for heads up of this PR to revert SelfContained implication for PublishAot

MichalStrehovsky:
The problem this is trying to fix is that if someone takes a project and just wants to publish it as native aot, we'll end up downloading apphost and downloading the entire runtime pack, copy all of that into bin (as part of Build) and then do the actual native aot publish that doesn't need any of this because we don't publish the result of Build. So we just copy around a 100 MB of garbage.

It is wasteful but it wouldn't be the first wasteful thing the SDK is doing. Which is also why I didn't get back to this yet because maybe it doesn't move the needle enough.

I am ambivalent to fixing this. The current behavior makes publish slower and less efficient, but a lot of things in the SDK are already slow and inefficient. I'm honestly leaning towards just closing this and won't fixing the issue, especially given https://github.com/dotnet/sdk/pull/33229#issuecomment-1652190567 it's not possible for this optimization to kick in within VS.

richlander:
Seems like a good direction to me.

AppHost and SelfContained are CoreCLR concepts.

akoeplinger:
I _think_ we're currently relying on the SelfContained publish behavior regarding runtime packs in the NativeAOT iOS scenario. I do agree that a lot of this is wasteful but I'd be more comfortable if we look into this for .NET 9.

richlander:
The SDK has a different bar. A middle ground is to defer it to 8.0.200.

akoeplinger:
I'm pretty sure it'd need corresponding workload changes in dotnet/runtime though.

MichalStrehovsky:
I now learned some new things while investigating https://github.com/dotnet/runtime/pull/95496 and I no longer want to pursue this.

PR #33266, opened 2023-06-15
Containers: Registry: support pulling from quay.io.

tmds:
quay.io returns 'application/vnd.oci.image.manifest.v1+json' which we can treat the same as 'application/vnd.docker.distribution.manifest.v2+json'.

@baronfel ptal.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/Registry.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Registry.cs
@@ -17,6 +17,7 @@ namespace Microsoft.NET.Build.Containers;
 internal sealed class Registry
 {
     private const string DockerManifestV2 = "application/vnd.docker.distribution.manifest.v2+json";
+    private const string OciManifestV1 = "application/vnd.oci.image.manifest.v1+json"; // https://containers.gitbook.io/build-containers-the-hard-way/#registry-format-oci-image-manifest
     private const string DockerManifestListV2 = "application/vnd.docker.distribution.manifest.list.v2+json";
     private const string DockerContainerV1 = "application/vnd.docker.container.image.v1+json";
     private const string DockerHubRegistry1 = "registry-1.docker.io";
@@ -157,7 +158,7 @@ internal sealed class Registry
 
         return initialManifestResponse.Content.Headers.ContentType?.MediaType switch
         {
-            DockerManifestV2 => await ReadSingleImageAsync(
+            DockerManifestV2 or OciManifestV1 => await ReadSingleImageAsync(
                 repositoryName,
                 await initialManifestResponse.Content.ReadFromJsonAsync<ManifestV2>(cancellationToken: cancellationToken).ConfigureAwait(false),
                 cancellationToken).ConfigureAwait(false),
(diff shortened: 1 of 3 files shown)


baronfel:
This also impacts pulls directly from docker hub, at least for the images I tried. We'd been talking about this for future work, but the new schemas have differences from the 'docker' models that we'd want to properly model to ensure we don't miss any data. For this media type, here's the relevant section from the [media types](https://github.com/opencontainers/image-spec/blob/main/media-types.md#compatibility-matrix) docs page:

> # application/vnd.oci.image.manifest.v1+json
>
>Similar/related schema
>
>* [application/vnd.docker.distribution.manifest.v2+json](https://github.com/distribution/distribution/blob/master/docs/spec/manifest-v2-2.md#image-manifest-field-descriptions)
>    * .annotations: only present in OCI
>    * .config.annotations: only present in OCI
>    * .config.urls: only present in OCI
>    * .[]layers.annotations: only present in OCI

Practically this means that we would need to be able to read the `annotations` at minimum from the image manifest and forward them along as `Labels` on the older model that we current use.

Does Quay serve [image indexes](https://github.com/opencontainers/image-spec/blob/main/media-types.md#applicationvndociimageindexv1json) ever? 

I think we should be looking at modeling these types more explicitly and having the 'domain' layer paper over the presentation differences (e.g. keep the user thinking of ContainerLabel but express those as annotations if required by the underlying media type).

tmds:
> Practically this means that we would need to be able to read the annotations at minimum from the image manifest and forward them along as Labels on the older model that we current use.

@baronfel after reading through the annotations at https://github.com/opencontainers/image-spec/blob/main/annotations.md it seems most are meant to reflect the image itself and using the annotations of the base image and including them in some way for the application image is not appropriate. Do you agree?

If/when we make the container tooling support pushing `application/vnd.oci.image.manifest.v1+json`, adding the `org.opencontainers.image.base.{base,name}` annotations provides a mechanism for obtaining base image annotations.

baronfel:
After reading in detail, I agree with your conclusion 👍

tmds:
@baronfel I don't think there are changed needed to the implementation.
Can you take a look at the test I've included? Are there other registries that I could add to that test?
I have not included `docker.io` because I assume it may cause CI failures due to hitting rate limits.

MichalPavlik (approved):

jeffpardy:
@baronfel I was curious if this would be included in [release/7.0.4xx]  or would that need a separate PR? This unblocks my usage for using JFrog Artifactory which requires the oci manifest v1.

baronfel:
No, this was targeted at main branch and so currently is only being shipped in the 8.0.100 previews. I think this one in particular is shipping in preview 6.  We could potentially backport it to 7.0.400, but the time frame for that is getting very short.

jeffpardy:
Makes sense, I'd be happy to put up a PR. Maybe after #33536 is merged since it refactors `Registry.cs`?

baronfel:
@jeffpardy could you go ahead and prep it, tag it with the `Area: Containers` label, and make sure `@dotnet/sdk-container-builds-maintainers` is tagged for review? Time is getting tight (we close for 7.0.400 at the end of the week) and that would really help us line it up :)

PR #33282, opened 2023-06-15
Do not look in root for .`config\dotnet-tools.json ` manifest for windows

JL03-Yue:
Targeting branch 8.0.0

When use this command in Windows system. The process will search for tool information in the .config\dotnet-tools.json file from the bottom up and download it.

For Windows, the program does not check the root folder for the tool manifest if the Environment Variable `CHECK_MANIFEST_IN_ROOT" is not set.

For Linux/MacOS, the program checks the root folder for the tool manifest.


GrabYourPitchforks (approved):
Looks like a straightforward port of the approved 7.0 change.

GrabYourPitchforks:
Marc told me offline that this change already came in via an automated PR https://github.com/dotnet/sdk/pull/33290. That would explain the conflicts you're seeing.

marcpopMSFT:
Post-merge resolution looks like the only change left in this PR is white space so I think we're good to close it as that confirms the codeflow in 33290 brought in the fix.

Issue #33241, opened 2023-06-14
Secrets Management in different Environments


DanielHabenicht:
### Is your feature request related to a problem? Please describe.

Secrets management in Development environments is unfulfilling. The Basic configuration is helpful, but is not user friendly when handling multiple environments and bigger teams, as it lacks:  

- Quickly switching between different secrets for different environments
- finding currently configured secrets (of course there is `dotnet user-secrets list`, but one has to remember this command instead of visually seeing the `.env` file in the repository`) 
- misalignment with "industry standards" (if you could call `.env` files this)
- Easy sharing of secrets between developers (where was the file again to quickly share with my coworker?)

### Describe the solution you'd like

There are multiple ways to relief some development pain: 

1. Allowing another flag for the current cli: (Apart from the `--configuration`), e.g. `--environment`
Basically this would be the equivalent to the different `appsettings.json` files. 
If the application would read `appsettings.Development.json` it would also read `secrets.Development.json` and the CLI options would write to the respective file. 
This would allow to easily switch between "Staging" and "Development" Secrets (basically allowing each launch Profile to potentially have different Secrets).
Please note that the current `--configuration` implementation does not allow this, as I might want to Debug with the secrets from the Staging Environment or test the Release build with the local Development Secrets. 

2. Make the dotnet eco system understand `.env` files. 
Instead of writing extra documentation about how to handle secrets with the dotnet project or searching the internet for this [guide](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-7.0&tabs=windows) `.env` files could be used as they are a defacto standard known to many development tools. 
Some ideas on how to add them: 
   - Add `launchSettings.json` support: 
   ```jsonc
   {
    "profiles": {
      "EnvironmentsSample": {
        "commandName": "Project",
        "dotnetRunMessages": true,
        "launchBrowser": true,
        "applicationUrl": "https://localhost:7152;http://localhost:5105",
        "environmentVariables": {
          "ASPNETCORE_ENVIRONMENT": "Development"
        },
       // Add another Option for reading .env Files
       "environmentFiles": [".env", "development.env"]
      },
   }
   ```
   - Add `dotnet` cli support:
     For `--configuration=Development` the `.env` file could be loaded automatically. 
     For manually adding it with other configurations adding the option `--env-file .env` would be great. 

### Additional context
[...]

PR #33305, opened 2023-06-15
Update .gitignore to ignore .env files

baronfel:
dotenv is a very common convention in other ecosystems for allowing user-specific env var values without doing crazy custom configurations. We're talking about things in https://github.com/dotnet/sdk/issues/33241 that might lead to .NET apps doing more with dotenv, so let's get ahead of the game and pre-emptively don't commit these files.

A longer-term change would sync this with github/gitignore, but wanted to get this flag in the ground first.


MiYanni (approved):
FYI, there was discussion about supporting .env natively in VS not too long ago. Issue is here: https://github.com/dotnet/project-system/issues/8943

danmoseley:
Should other nearby repos have a similar PR?

YuliiaKovalova:
@baronfel, Gang has fixed test failure

PR #33309, opened 2023-06-15
Enable dynamic adaptive GC by default in Web SDK when PublishAot==true

### LINKED ISSUE #33298: Enable dynamic adaptive GC by default in Web SDK when PublishAot==true
The new DATAS GC mode should be made the default in the Web SDK when `<PuslibhAot>true</PublishAot>`. The property to enable DATAS is `<GarbageCollectionAdaptationMode>1</GarbageCollectionAdaptationMode>`. The existing property default to enable Server GC needs to remain, i.e. both `<ServerGarbageCollection>true</ServerGarbageCollection>` and `<GarbageCollectionAdaptationMode>1</GarbageCollectionAdaptationMode>` should be set when `<PuslibhAot>true</PublishAot>` is set.

eerhardt:
Fix #33298

cc @mangod9 @Maoni0

### DIFF
--- a/src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
+++ b/src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
@@ -61,7 +61,7 @@ namespace Microsoft.NET.Sdk.Web.Tests
 
             var testProject = CreateTestProjectForILLinkTesting(targetFramework, projectName);
             testProject.RecordProperties("NETCoreSdkPortableRuntimeIdentifier");
-            testProject.AdditionalProperties["PublishAOT"] = "true";
+            testProject.AdditionalProperties["PublishAot"] = "true";
             testProject.AdditionalProperties["UseCurrentRuntimeIdentifier"] = "true";
             testProject.PropertiesToRecord.Add("PublishTrimmed");
             testProject.PropertiesToRecord.Add("TrimMode");
@@ -83,6 +83,7 @@ namespace Microsoft.NET.Sdk.Web.Tests
 
             responseFileContents.Should().Contain("--feature:System.Text.Json.JsonSerializer.IsReflectionEnabledByDefault=false");
             responseFileContents.Should().Contain("--feature:System.Diagnostics.Tracing.EventSource.IsSupported=true");
+            responseFileContents.Should().Contain("--runtimeknob:System.GC.DynamicAdaptationMode=1");
             File.Exists(Path.Combine(outputDirectory, "web.config")).Should().BeFalse();
         }
 
--- a/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
+++ b/src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
@@ -57,7 +57,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <EnableRequestDelegateGenerator Condition="'$(EnableRequestDelegateGenerator)' == ''">true</EnableRequestDelegateGenerator>
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
 
-    <!-- Enable System.GC.DynamicAdaptationMode by default for aot'd apps to get a GC that dynamically adapts to the application size. -->
+    <!-- Enable System.GC.DynamicAdaptationMode by default for aot'd apps to get a GC that dynamically adapts to the application's memory use. -->
     <GarbageCollectionAdaptationMode Condition="'$(GarbageCollectionAdaptationMode)' == '' And '$(ServerGarbageCollection)' == 'true'">1</GarbageCollectionAdaptationMode>
   </PropertyGroup>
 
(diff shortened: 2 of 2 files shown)


eerhardt:
FYI @agocke @vitek-karas @MichalStrehovsky @jkotas for awareness.

src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
```diff
@@ -83,6 +83,7 @@ public void TrimmingOptions_Are_Defaulted_Correctly_On_Aot_Apps(string targetFra
 
             responseFileContents.Should().Contain("--feature:System.Text.Json.JsonSerializer.IsReflectionEnabledByDefault=false");
             responseFileContents.Should().Contain("--feature:System.Diagnostics.Tracing.EventSource.IsSupported=true");
+            responseFileContents.Should().Contain("--runtimeknob:System.GC.DynamicAdaptationMode=1");
```
DamianEdwards, line 86: it's literally called `--runtimeknob`? 😆

eerhardt: https://github.com/dotnet/runtime/blob/3afe6bd73788bd941dc06e4160616dc100bbcc43/src/coreclr/tools/aot/ILCompiler/ILCompilerRootCommand.cs#L95-L96

```C#
        public Option<string[]> RuntimeKnobs { get; } =
            new(new[] { "--runtimeknob" }, Array.Empty<string>, "Runtime knobs to set");
```

😆

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
```diff
@@ -56,6 +56,9 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!-- Enable the built-in source generators. -->
     <EnableRequestDelegateGenerator Condition="'$(EnableRequestDelegateGenerator)' == ''">true</EnableRequestDelegateGenerator>
     <EnableConfigurationBindingGenerator Condition="'$(EnableConfigurationBindingGenerator)' == ''">true</EnableConfigurationBindingGenerator>
+
+    <!-- Enable System.GC.DynamicAdaptationMode by default for aot'd apps to get a GC that dynamically adapts to the application size. -->
```
DamianEdwards, line 60:
```suggestion
    <!-- Enable System.GC.DynamicAdaptationMode by default for aot'd apps to get a GC that dynamically adapts to the application memory use. -->
```

DamianEdwards (approved):
Only the one comment nit

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.targets
```diff
@@ -56,6 +56,9 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <GarbageCollectionAdaptationMode Condition="'$(GarbageCollectionAdaptationMode)' == '' And '$(ServerGarbageCollection)' == 'true'">1</GarbageCollectionAdaptationMode>
```
mangod9, line 61: should this set it to `$(GarbageCollectionAdaptationMode)'` instead of `1`?

DamianEdwards: No it's saying it will only set it to 1 if it's not already set.

mangod9: ah, ok. Makes sense.

mangod9 (approved):

src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
MichalStrehovsky: In CoreCLR lingo, the settings one can set through environment variables are "config values" (e.g. `DOTNET_gcServer=1`). The values in runtimeconfig.json are "knob values" (e.g. `System.GC.Server=true`).

I can't say I'm thrilled to see the existence/name/contents of the RSP file interpreted outside of the dotnet/runtime repo though.

vijayrkn (approved):

src/Tests/Microsoft.NET.Sdk.Web.Tests/PublishTests.cs
eerhardt: > I can't say I'm thrilled to see the existence/name/contents of the RSP file interpreted outside of the dotnet/runtime repo though.

Agreed this isn't ideal. But I can't really think of a better way to ensure the properties are flown correctly. Running the app, giving it load, and inspecting its memory usage would be the "real" way to test it. But that's too heavy handed.

The good news is that this is limited to a very small subset of tests (namely 1) to ensure we have the right defaulting logic in place. If we ever change these ilc command line argument names, it is simple to fix it here.

MichalStrehovsky: > Agreed this isn't ideal. But I can't really think of a better way to ensure the properties are flown correctly. Running the app, giving it load, and inspecting its memory usage would be the "real" way to test it. But that's too heavy handed.

If we're allowed to run the app, we also inspect AppContext (these also flow to AppContext), or there might be a GC API we can query directly. Similarly for the feature switches.

I'm not saying we need to do this because digging into RSP also works and this code is already written, but maybe for next time.

PR #33322, opened 2023-06-16
Fix workload cache security descriptors

joeloff:
# Description

Fixes the security descriptors for the workload cache under `ProgramData`. This supersedes #30769. Object inheritance on the cache root and subfolders ensures consistent results across multiple account contexts.

The fix addresses a regression that impacted running workload commands inside Windows Sandbox when trying to assign the Domain Users SID. The regression was the result of attempting to fix another issue related to multiple users on the same machine running `dotnet workload` commands. The initial fix assigned the Domain Users SID that does not always exist.

The desired descriptor should provide full access to administrators and SYSTEM while providing read and execute permissions to built-in users and Everyone. Additionally for files inside the cache, the individual descriptors in the DACL should be inherited from the parent directory (the `ID` flag should be present in the ACE).

# Testing

A manual test application with documentation is included in this change. See the test documentation for additional details on the scenarios it verifies.

# Sample results using the manual test application

The sample below is from running inside Windows Sandbox
image

Running the same test as `nt authority\system` (local SYSTEM account)

image

### DIFF
--- a/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
+++ b/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
@@ -172,7 +172,7 @@ namespace Microsoft.DotNet.Installer.Windows
                 string packageDirectory = GetPackageDirectory(packageId, packageVersion);
 
                 // Delete the package directory and create a new one. If all the files were properly
-                // cached, the client would not request this action. 
+                // cached, the client would not request this action.
                 if (Directory.Exists(packageDirectory))
                 {
                     Directory.Delete(packageDirectory, recursive: true);
--- a/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
+++ b/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
@@ -92,48 +92,13 @@ namespace Microsoft.DotNet.Installer.Windows
         /// and files to inherit access control entries. 
         /// </summary>
         /// <param name="path">The path of the directory to create.</param>
-        /// <param name="resetAccessRules">When <see langword="true"/>, all existing access rules are removed before
-        /// setting up new rules.
-        /// </param>
-        /// <param name="log">The underlying setup log to use.</param>
-        public static void CreateSecureDirectory(string path, bool resetAccessRules = false, ISetupLogger log = null)
+        public static void CreateSecureDirectory(string path)
         {
             if (!Directory.Exists(path))
             {
-                // Create the directory with the desired access rights if it doesn't exist. We don't use
-                // Directory.CreateDirectory as that can cause problems with inherited ACEs from
-                // the parent folders as access rules become cumulative. The acccess rules for the directory should all have
-                // object and container inheritance set to ensure access rights are inherited when creating files.
                 DirectorySecurity ds = new();
                 SetDirectoryAccessRules(ds);
                 ds.CreateDirectory(path);
-            }
-            else if (resetAccessRules)
-            {
-                // Best effort to reset the DACL and remove unnecessary ACEs. This will
-                // help to address some issues that may arise from older .NET versions that created
-                // the cache with permissions that were too restrictive.
-                try
-                {
-                    DirectorySecurity ds = new(path, AccessControlSections.Access);
-
-                    ds.SetAccessRuleProtection(isProtected: true, preserveInheritance: false);
-
-                    foreach (FileSystemAccessRule ace in ds.GetAccessRules(includeExplicit: true,
-                        includeInherited: true, typeof(NTAccount)))
-                    {
-                        ds.PurgeAccessRules(ace.IdentityReference);
-                    }
-
-                    SetDirectoryAccessRules(ds);
-                    DirectoryInfo di = new DirectoryInfo(path);
-                    di.SetAccessControl(ds);
-                }
-                catch (Exception ex)
-                {
-                    // Record the error, but don't fail the operation if we couldn't reset the access rules.
-                    log?.LogMessage($"Failed to reset access rules for '{path}'. Error: {ex.Message}");
-                }
             }
         }
 
@@ -154,10 +119,6 @@ namespace Microsoft.DotNet.Installer.Windows
 
             if (IsElevated)
             {
-                // Create the root directory. If the directory already exists then we'll try to
-                // reset the DACL to conform to our expectations.
-                CreateSecureDirectory(PackageCacheRoot, resetAccessRules: true, Log);
-
                 string packageDirectory = GetPackageDirectory(packageId, packageVersion);
 
                 // Delete the package directory and create a new one that's secure. If all the files were properly
@@ -209,8 +170,12 @@ namespace Microsoft.DotNet.Installer.Windows
         {
             if (!File.Exists(destinationFile))
             {
-                FileAccessRetrier.RetryOnMoveAccessFailure(() => File.Move(sourceFile, destinationFile));
-
+                FileAccessRetrier.RetryOnMoveAccessFailure(() =>
+                {
+                    // Moving the file preserves the owner SID and fails to inherit the WD ACE.
+                    File.Copy(sourceFile, destinationFile, overwrite: true);
+                    File.Delete(sourceFile);
+                });
                 log?.LogMessage($"Moved '{sourceFile}' to '{destinationFile}'");
 
                 FileInfo fi = new(destinationFile);
@@ -237,8 +202,7 @@ namespace Microsoft.DotNet.Installer.Windows
             string packageCacheDirectory = GetPackageDirectory(packageId, packageVersion);
             payload = default;
 
-            string msiPath;
-            if (!TryGetMsiPathFromPackageData(packageCacheDirectory, out msiPath, out string manifestPath))
+            if (!TryGetMsiPathFromPackageData(packageCacheDirectory, out string msiPath, out string manifestPath))
             {
                 return false;
             }
--- a/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
+++ b/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
@@ -117,7 +117,7 @@ namespace Microsoft.DotNet.Installer.Windows
                 {
                     DirectorySecurity ds = new(path, AccessControlSections.Access);
 
-                    ds.SetAccessRuleProtection(true, false);
+                    ds.SetAccessRuleProtection(isProtected: true, preserveInheritance: false);
 
                     foreach (FileSystemAccessRule ace in ds.GetAccessRules(includeExplicit: true,
                         includeInherited: true, typeof(NTAccount)))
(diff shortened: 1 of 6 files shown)


src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -88,41 +88,63 @@ internal class MsiPackageCache : InstallerBase
...
+        /// <param name="log">The underlying setup log to use.</param>
+        public static void CreateSecureDirectory(string path, bool resetAccessRules = false, ISetupLogger log = null)
         {
             if (!Directory.Exists(path))
             {
-                CreateSecureDirectory(Directory.GetParent(path).FullName);
+                // Create the directory with the desired access rights if it doesn't exist. We don't use
+                // Directory.CreateDirectory as that can cause problems with inherited ACEs from
+                // the parent folders as access rules become cumulative. The acccess rules for the directory should all have
+                // object and container inheritance set to ensure access rights are inherited when creating files.
```
MiYanni, line 106: Quick Q: Is the comment wrapping width some editorconfig setting or is it a VS thing? Or is it arbitrary?

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -143,12 +165,14 @@ public void CachePayload(string packageId, string packageVersion, string manifes
...
+                // Create the root directory. If the directory already exists then we'll try to
+                // reset the DACL to conform to our expectations.
+                CreateSecureDirectory(PackageCacheRoot, resetAccessRules: true, Log);
 
                 string packageDirectory = GetPackageDirectory(packageId, packageVersion);
 
-                // Delete the directory and create a new one that's secure. If the files were properly
-                // cached, the client won't request this action.
+                // Delete the package directory and create a new one. If all the files were properly
+                // cached, the client would not request this action.
```
MiYanni, line 175: nit: end of line space

MiYanni (approved):
Awesome test harness utility. Is there a plan to run it in CI? If so, what's the strategy for that?

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
joeloff: Arbitrary. I sometimes have to fight my 80x25/43 instinct.

joeloff:
> Awesome test harness utility. Is there a plan to run it in CI? If so, what's the strategy for that?

We've avoided all the tests where UAC gets involved because there's no easy way to automate that and running with admin privileges is unlikely to be the common case. For this particular situation normal users going through UAC, multiple users and Sandbox are the interesting cases. Shared users might be doable if we were to create a random account during the test, then using `runas`. Doing this safely would probably end up with storing access tokens in a vault, which means users that clone the repo won't be able to run the tests.

Definitely something to consider if we expand these types of tests I think.

src/Tests/SDDLTests/README.md
```diff
@@ -0,0 +1,81 @@
...
+`artifacts\sdk-build.env.bat` followed by `dotnet artifacts\SDDLTests\SDDLTests.dll`.
+
+The test will attempt to clean up the test assets from previous runs before
+recreating them.
+
+## Test Expectations
+
+The test will verify the security descriptor for 3 objects.
+- The root directory
+of the cahce in `ProgramData\SDDLTest`. This is the equivalent of `ProgramData\dotnet` used by
```
nagilson, line 20:
```suggestion
of the cache in `ProgramData\SDDLTest`. This is the equivalent of `ProgramData\dotnet` used by
```

src/Tests/SDDLTests/README.md
```diff
@@ -0,0 +1,81 @@
+# Security Descriptor Tests for Windows
```
nagilson, line 1: I love this document, it is super helpful. Thank you for including it.

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
joeloff: Thanks, fixing

src/Tests/SDDLTests/README.md
joeloff: Fixed in new iteration

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -88,41 +88,63 @@ internal class MsiPackageCache : InstallerBase
...
         {
             if (!Directory.Exists(path))
             {
-                CreateSecureDirectory(Directory.GetParent(path).FullName);
+                // Create the directory with the desired access rights if it doesn't exist. We don't use
+                // Directory.CreateDirectory as that can cause problems with inherited ACEs from
+                // the parent folders as access rules become cumulative. The acccess rules for the directory should all have
+                // object and container inheritance set to ensure access rights are inherited when creating files.
+                DirectorySecurity ds = new();
+                ds.SetOwner(s_AdministratorsSid);
```
nagilson, line 108: It seems like line 108-113 and line 133-138 is the same, I would suggest making a separate private function such as `SetAccessRules()`?

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -88,41 +88,63 @@ internal class MsiPackageCache : InstallerBase
...
+        /// setting up new rules.
+        /// </param>
+        /// <param name="log">The underlying setup log to use.</param>
+        public static void CreateSecureDirectory(string path, bool resetAccessRules = false, ISetupLogger log = null)
         {
             if (!Directory.Exists(path))
             {
-                CreateSecureDirectory(Directory.GetParent(path).FullName);
+                // Create the directory with the desired access rights if it doesn't exist. We don't use
+                // Directory.CreateDirectory as that can cause problems with inherited ACEs from
```
Forgind, line 104: I'm a bit confused here. If we inherit the parent folder's access rules, then reset them, that should be the equivalent of creating it anew, right?

Regardless, it would be nice to unify as much fo the DirectorySecurity modifications as possible 🙂

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -88,41 +88,63 @@ internal class MsiPackageCache : InstallerBase
...
+            {
+                // Best effort to reset the DACL and remove unnecessary ACEs. This will
+                // help to address some issues that may arise from older .NET versions that created
+                // the cache with permissions that were too restrictive.
+                try
+                {
+                    DirectorySecurity ds = new(path, AccessControlSections.Access);
 
-                DirectorySecurity directorySecurity = new();
+                    ds.SetAccessRuleProtection(true, false);
```
Forgind, line 120: nit:
```suggestion
                    ds.SetAccessRuleProtection(isProtected: true, preserveInheritance: false);
```

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -191,23 +215,24 @@ public string GetPackageDirectory(string packageId, string packageVersion)
...
         /// <param name="sourceFile">The source file to move.</param>
         /// <param name="destinationFile">The destination where the source file will be moved.</param>
-        private void MoveAndSecureFile(string sourceFile, string destinationFile)
+        /// <param name="log">The underlying setup log to use.</param>
+        public static void MoveAndSecureFile(string sourceFile, string destinationFile, ISetupLogger log = null)
         {
             if (!File.Exists(destinationFile))
             {
-                // See https://github.com/dotnet/sdk/issues/28450
+
```
Forgind, line 212: nit: blank line

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
```diff
@@ -143,12 +165,14 @@ public void CachePayload(string packageId, string packageVersion, string manifes
...
-                CreateRootDirectory();
+                // Create the root directory. If the directory already exists then we'll try to
+                // reset the DACL to conform to our expectations.
+                CreateSecureDirectory(PackageCacheRoot, resetAccessRules: true, Log);
 
                 string packageDirectory = GetPackageDirectory(packageId, packageVersion);
 
-                // Delete the directory and create a new one that's secure. If the files were properly
-                // cached, the client won't request this action.
+                // Delete the package directory and create a new one. If all the files were properly
```
Forgind, line 163: Did you remove "that's secure" because that isn't important for the MSI payload?

nagilson (commented):
The change itself looks good to me bar the nits, I still need to look at the tests. I remember how awesome this would be to have when I created workload clean 😄 Thank you!

You had mentioned execute permissions are given to everyone, I don't think that's changed. Is there anything of particular concern about that? I think it's fine to allow execution of workload commands, and I think even if admin was required then it would still be necessary?

Ideally there's a security descriptor expert we can have look at this, but I think it's unlikely? There really wasn't much that actually changed in terms of the security of it so it's probably OK.

src/Tests/SDDLTests/Program.cs
```diff
@@ -0,0 +1,323 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.DotNet.Installer.Windows;
+using Xunit;
+
+namespace SDDLTests
+{
+    /// <summary>
+    /// Console application containing a manual test to verify security descriptors used by workloads when caching packages.
+    /// </summary>
+    [SupportedOSPlatform("windows")]
```
Forgind, line 22: I glanced at this only briefly, and sorry for a newb question 🙂

Why does this have to be windows-specific? I noticed a few things that seemed to be the windows flavor of a more general concept (like "IsAdministrator" using windows-specific APIs), but I think this should ideally work across platforms, which means it should ideally be tested across platforms, right?

nagilson: I believe that's because it's for the MSI Package cache, and MSIs are a windows only concept

Forgind: Follow-up: What's MSI? I thought it was Microsoft Installer, presumably for the SDK, which is cross-platform, so I think I have a broken link in my brain.

nagilson (approved):

joeloff:
> The change itself looks good to me bar the nits, I still need to look at the tests. I remember how awesome this would be to have when I created workload clean 😄 Thank you!
> 
> You had mentioned execute permissions are given to everyone, I don't think that's changed. Is there anything of particular concern about that? I think it's fine to allow execution of workload commands, and I think even if admin was required then it would still be necessary?
> 
> Ideally there's a security descriptor expert we can have look at this, but I think it's unlikely? There really wasn't much that actually changed in terms of the security of it so it's probably OK.

With regards to the execution, the problem we want to address is around machines where multiple users have access. Let's say user A and B have access to the same machine. If A installs a workload from the CLI, the workload pack MSI will be cached under ProgramData. The next time the packages are needed, we want to avoid downloading them, verifying signatures, etc. If user B logs on and tries to perform a workload operation that needs those packages, they should have the ability to read/execute the MSIs in the global cache. Previously when the files were moved, they would end up with the user's SID that performed the operation. When another user tries to access them, they end up receiving an access denied error.

src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs
joeloff: I've removed the cleanup code, so things are much simpler now. Even if we modified the top-level permissions, it won't translate through to all the child objects.

Forgind (approved):

PR #33348, opened 2023-06-16
Normalize line endings in GenAPI consistently

### LINKED ISSUE #33289: [SBRP] Generated artifacts have a mixture of line endings
1. delete ./src/referencePackages/src/microsoft.bcl.asyncinterfaces/1.1.0
2. run ./generate.sh -x -p microsoft.bcl.asyncinterfaces,1.1.0
3. Inspect the line endings used across the generated artifacts

Expected Results:
1. Artifacts would use the same line ending formats
2. line ending format would be the default for the environment (e.g. Linux/Windows).

Actual Results in a Linux environment:
1. csproj - LF
2. nuspec - CRLF
3. netstandard 2.0 code - CRLF
4. netstandard 2.1 code - LF

ViktorHofer:
Normalize line endings for the license header and the assembly attributes.

Fixes https://github.com/dotnet/sdk/issues/33289

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
@@ -80,8 +80,7 @@ namespace Microsoft.DotNet.GenAPI
             SyntaxNode compilationUnit = _syntaxGenerator.CompilationUnit(namespaceSyntaxNodes)
                 .WithAdditionalAnnotations(Formatter.Annotation, Simplifier.Annotation)
                 .Rewrite(new TypeDeclarationCSharpSyntaxRewriter())
-                .Rewrite(new BodyBlockCSharpSyntaxRewriter(_exceptionMessage))
-                .NormalizeWhitespace();
+                .Rewrite(new BodyBlockCSharpSyntaxRewriter(_exceptionMessage));
 
             if (_includeAssemblyAttributes)
             {
@@ -89,6 +88,7 @@ namespace Microsoft.DotNet.GenAPI
             }
 
             compilationUnit = GenerateForwardedTypeAssemblyAttributes(assemblySymbol, compilationUnit);
+            compilationUnit = compilationUnit.NormalizeWhitespace(eol: Environment.NewLine);
 
             Document document = project.AddDocument(assemblySymbol.Name, compilationUnit);
             document = Simplifier.ReduceAsync(document).Result;
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
@@ -4,6 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+#if !NET
+using System.Text.RegularExpressions;
+#endif
 using Microsoft.CodeAnalysis;
 using Microsoft.DotNet.ApiSymbolExtensions;
 using Microsoft.DotNet.ApiSymbolExtensions.Filtering;
@@ -198,9 +201,17 @@ namespace Microsoft.DotNet.GenAPI
 
             """;
 
-            return !string.IsNullOrEmpty(headerFile) ?
+            string header = !string.IsNullOrEmpty(headerFile) ?
                 File.ReadAllText(headerFile) :
                 defaultFileHeader;
+
+#if NET
+            header = header.ReplaceLineEndings();
+#else
+            header = Regex.Replace(header, @"\r\n|\n\r|\n|\r", Environment.NewLine);
+#endif
+
+            return header;
         }
     }
 }


MichaelSimons (approved):

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -198,9 +201,17 @@ private static string ReadHeaderFile(string? headerFile)
 
             """;
 
-            return !string.IsNullOrEmpty(headerFile) ?
+            string header = !string.IsNullOrEmpty(headerFile) ?
                 File.ReadAllText(headerFile) :
                 defaultFileHeader;
+
+#if NET
+            header = header.ReplaceLineEndings();
```
carlossanlop, line 209: Why isn't `ReplaceLineEndings` enough for all cases?

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -80,15 +80,15 @@ public void WriteAssembly(IAssemblySymbol assemblySymbol)
             SyntaxNode compilationUnit = _syntaxGenerator.CompilationUnit(namespaceSyntaxNodes)
                 .WithAdditionalAnnotations(Formatter.Annotation, Simplifier.Annotation)
                 .Rewrite(new TypeDeclarationCSharpSyntaxRewriter())
-                .Rewrite(new BodyBlockCSharpSyntaxRewriter(_exceptionMessage))
-                .NormalizeWhitespace();
```
carlossanlop, line 84: Why did you decide to move this call to line 83? Wouldn't adding the `eol: Environment.NewLine` argument here be enough?

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -198,9 +201,17 @@ private static string ReadHeaderFile(string? headerFile)
...
+#else
+            header = Regex.Replace(header, @"\r\n|\n\r|\n|\r", Environment.NewLine);
```
carlossanlop, line 211: I'd be surprised if `\n\r` is ever found in the wild. I think `@"\r\n|\r|\n"` is enough, it would cover that edge case.

carlossanlop (commented):
Left some questions.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -4,6 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+#if !NET
```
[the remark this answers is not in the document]
ViktorHofer: NET means .NETCoreApp. !NET means .NET Framework or .NET Standard. GenAPI targets both .NETCoreApp and .NET Framework.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
ViktorHofer: ReplaceLineEndings isn't available on .NET Framework, it's an API added in .NET 6.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
ViktorHofer: Because the CompilationUnit is mutated below. The `NormalizeWhitespace` invocation shouldn't happen before the syntax tree is finale.

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
ViktorHofer: I saw others doing this kind of replacement so I thought it doesn't harm. This code path is .NET Framework only so I'm less concerned.

carlossanlop (approved):

src/GenAPI/Microsoft.DotNet.GenAPI/GenAPIApp.cs
```diff
@@ -198,9 +201,17 @@ private static string ReadHeaderFile(string? headerFile)
 
             """;
 
-            return !string.IsNullOrEmpty(headerFile) ?
+            string header = !string.IsNullOrEmpty(headerFile) ?
```
ericstj, line 204: This looks like the best minimal change.  Something we might consider in the future is making CSharpFileBuilder responsible for adding the header as syntax, then this could be handled by NormalizeWhitespace.

ViktorHofer: I tried that initially but couldn't get the syntax tree accept arbitary input. The license header file's content might contain single, multi-line comments or something totally different.

PR #33441, opened 2023-06-20
Support setting workload set versions in global.json

dsplaisted:
Implements part of https://github.com/dotnet/designs/pull/294

We may want to add some end-to-end tests.


src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
```diff
@@ -43,7 +43,7 @@ internal class WorkloadUninstallCommand : WorkloadCommandBase
             userProfileDir = userProfileDir ?? CliFolderPathCalculator.DotnetUserProfileFolderPath;
             _sdkVersion = WorkloadOptionsExtensions.GetValidatedSdkVersion(parseResult.GetValueForOption(WorkloadUninstallCommandParser.VersionOption), version, dotnetPath, userProfileDir, true);
 
-            var workloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(dotnetPath, _sdkVersion.ToString(), userProfileDir);
+            var workloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(dotnetPath, _sdkVersion.ToString(), userProfileDir, SdkDirectoryWorkloadManifestProvider.GetGlobalJsonPath(Environment.CurrentDirectory));
```
joeloff, line 46: Can this be moved into a base constructor? Feels like a lot of repetition.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -399,7 +422,105 @@ public void ItThrowsExceptionIfWorkloadSetJsonFilesHaveDuplicateManifests()
...
+        public void ItFailsIfWorkloadSetFromGlobalJsonIsNotInstalled()
+        {
+            Initialize("8.0.200");
+
+            string? globalJsonPath = Path.Combine(_testDirectory, "global.json");
+            File.WriteAllText(globalJsonPath, """
+            {
+                "sdk": {
+                    "version": "8.0.200",
+                    "workloadversion": "8.0.201"
```
joeloff, line 480: The global.json reader looks for `workloadVersion`. Shouldn't this through an exception because the property is all lower case?

dsplaisted: It's doing case-insensitive comparison.  I'm not sure whether global.json is supposed to be case-sensitive or not, but I followed the pattern from the workload manifest parsing code, which is case-insensitive.

joeloff: I went over the global.json docs. It doesn't mention anything from what I could tell. I'm fine if it's case-insensitive

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
joeloff: Taking a look...

joeloff: Looks consistent to me. All the default behavior appears to be on the factory side.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
joeloff: As I mentioned to you in chat, the runtime treats global.json properties as case sensitive. We should probably be consistent with the runtime.

joeloff (approved):

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/Strings.resx
```diff
@@ -192,8 +192,8 @@
   <data name="SpecifiedManifestNotFound" xml:space="preserve">
     <value>Specified workload manifest was not found: {0}</value>
   </data>
-  <data name="SpecifiedWorkloadSetNotFound" xml:space="preserve">
-    <value>Specified workload set was not found: {0}</value>
+  <data name="WorkloadVersionFromGlobalJsonNotFound" xml:space="preserve">
+    <value>Workload version {0}, which was specified in {1}, was not found</value>
```
nagilson, line 196: Do we want users to run restore here, once restore is aware of the global.json changes? Is that why you haven't added that to the error yet? It would be best to give the user some action to take.

src/Microsoft.DotNet.TemplateLocator/TemplateLocator.cs
```diff
@@ -54,7 +54,12 @@ public TemplateLocator()
                     nameof(dotnetRootPath));
             }
 
-            _workloadManifestProvider ??= new SdkDirectoryWorkloadManifestProvider(dotnetRootPath, sdkVersion, userProfileDir);
+            //  Will the current directory correspond to the folder we are creating a project in?  If we need
```
nagilson, line 57: Was this resolved? If not : is there an issue with this marked, so we can make sure we don't lose track of this TODO?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -344,5 +349,20 @@ public string GetSdkFeatureBand()
         {
             return _sdkVersionBand.ToString();
         }
+
+        public static string? GetGlobalJsonPath(string? globalJsonStartDir)
+        {
+            string? directory = globalJsonStartDir;
```
nagilson, line 355: How does the SDK handle multiple global jsons in the directory? Do files closer to the end of the path you're executing in override the values that could've come from global jsons higher up the directories, or does it only look at the top level global json like we do here?

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
```diff
@@ -39,15 +39,25 @@ internal class WorkloadUninstallCommand : WorkloadCommandBase
...
+                UseInstalledSdkVersionForResolver = true
+            };
+
+            var creationResult = WorkloadResolverFactory.Create(creationParameters);
+
+            _sdkVersion = creationResult.SdkVersion;
 
             var sdkFeatureBand = new SdkFeatureBand(_sdkVersion);
-            _workloadInstaller = WorkloadInstallerFactory.GetWorkloadInstaller(Reporter, sdkFeatureBand, workloadResolver, Verbosity, userProfileDir, VerifySignatures, PackageDownloader, dotnetPath);
+            _workloadInstaller = WorkloadInstallerFactory.GetWorkloadInstaller(Reporter, sdkFeatureBand, workloadResolver, Verbosity, userProfileDir, VerifySignatures, PackageDownloader, creationResult.DotnetPath);
```
nagilson, line 60: Why do we not set the workloadResolver here? Probably OK, but its something I noticed we did in the other files but didn't do here.

src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
```diff
@@ -63,22 +61,32 @@ internal abstract class InstallingWorkloadCommand : WorkloadCommandBase
...
                 new PackageSourceLocation(string.IsNullOrEmpty(configOption) ? null : new FilePath(configOption), sourceFeedOverrides: sourceOption);
 
-            var sdkWorkloadManifestProvider = new SdkDirectoryWorkloadManifestProvider(_dotnetPath, _installedSdkVersion.ToString(), userProfileDir);
-            _workloadResolver = workloadResolver ?? WorkloadResolver.Create(sdkWorkloadManifestProvider, _dotnetPath, _installedSdkVersion.ToString(), _userProfileDir);
+            var creationParameters = new WorkloadResolverFactory.CreationParameters()
+            {
+                DotnetPath = dotnetDir,
+                UserProfileDir = userProfileDir,
+                GlobalJsonStartDir = null,
+                VersionFromOption = parseResult.GetValueForOption(InstallingWorkloadCommandParser.VersionOption),
```
nagilson, line 76: What kind of version is this versionFromOption? The SDK Version? Workload version? Workload set version? I would suggest making the variable name a little more specific

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -399,7 +422,105 @@ public void ItThrowsExceptionIfWorkloadSetJsonFilesHaveDuplicateManifests()
                 """);
 
             Assert.Throws<ArgumentException>(() =>
-                new SdkDirectoryWorkloadManifestProvider(sdkRootPath: _fakeDotnetRootDirectory, sdkVersion: "8.0.200", userProfileDir: null));
+                new SdkDirectoryWorkloadManifestProvider(sdkRootPath: _fakeDotnetRootDirectory, sdkVersion: "8.0.200", userProfileDir: null, globalJsonPath: null));
+        }
+
+        [Fact]
```
nagilson, line 428: Thank you for providing well written tests as always 😄

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -26,13 +26,12 @@ public class SdkDirectoryWorkloadManifestProvider : IWorkloadManifestProvider
 
         private readonly WorkloadSet? _workloadSet;
 
-        public SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVersion, string? userProfileDir, string? requestedWorkloadSet = null)
-            : this(sdkRootPath, sdkVersion, Environment.GetEnvironmentVariable, userProfileDir, requestedWorkloadSet)
+        public SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVersion, string? userProfileDir, string? globalJsonPath)
```
nagilson, line 29: A lot of places seem to call this thing without a globalJsonPath... is it a good idea to make it an optional parameter with a null default instead?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.GlobalJsonReader.cs
```diff
@@ -0,0 +1,180 @@
...
+                                            if (string.Equals("workloadVersion", sdkPropName, StringComparison.OrdinalIgnoreCase))
+                                            {
+                                                workloadVersion = ReadString(ref reader);
+                                            }
+                                            else
+                                            {
+                                                ConsumeValue(ref reader);
+                                            }
+                                            break;
+                                        case JsonTokenType.EndObject:
```
nagilson, line 78: Would this cause a failure if someone was using the `msbuild-sdks` global.json option? It seems that would be an object, so we may need to handle the `JsonToken.TypeStartObject`

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
dsplaisted: It's hard for me to see a downside for being case insensitive.  What would the would that break?  On the other hand, if we are case sensitive, some people will probably get the casing wrong and be confused as to why the setting isn't taking effect.

baronfel: We can mitigate that somewhat by updating the [Schemastore json schema for global.json](https://json.schemastore.org/global.json) with this field - this enables editors to flag this error.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: Only one global.json ever applies, ie it starts in a folder and goes up the tree looking for global.json, and stops the search once it finds it.  It's not like NuGet.config where it will keep looking and merge the data from multiple files.

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
dsplaisted: No, I'm pretty sure this would have broken the uninstall command.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/Strings.resx
dsplaisted: I just hadn't thought of adding it when I was updating the string.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: I think the places that pass null for this are all in the tests.  Since I was adding a new parameter I wanted to make sure to update all the code that calls it to account for the new parameter.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.GlobalJsonReader.cs
dsplaisted: The StartObject token would need to come after a PropertyName token, and should be handled by `ConsumeValue`.  The tests include global.json files with `msbuild-sdks` in them.  I'm not an expert at json parsing but I copied the patterns from the existing workload manifest parser, so I think the global.json parser is probably correct.

PR #33444, opened 2023-06-20
[Blazor] Embed browser static web assets into non browser TFM

javiercn:
# Unified project system

The unified project system enables a project to multi-target net-x.y and the new net-x.y-browser TFMs simultaneously. In the way we intend this to work, the netx.y-browser TFM will produce a set of webassembly files that will be consumed by the netx.y target framework as a set of static web assets.

```mermaid
flowchart
    subgraph BlazorWebApp
    A["netx.y"]
    B["netx.y-browser"]
    end

    subgraph E["netx.y"]
    C["Server"]
    end    
    subgraph F["netx.y-browser"]
    D["Client"]
    end
    E --> F
```

However, this implies new challenges, since instead of 2 projects, we have one. Given this, there are certain things we need to take into consideration:
* Transitive dependencies. When there is a project or package reference that is only conditionally included in the browser TFM, we need to include the assets for that dependency in the parent too.
* Common assets. For example, those in the wwwroot folder. In this case, we need to discard those from the browser TFM in favor of the ones from the parent TFM.
* Generated files, like scoped CSS files. Can they be platform specific? If so, how do we handle them?
* Build ordering. There is a dependency between the two TFMs, but the build is not aware of it, so we need to reorganize the build to make sure that the dependency is reflected in the build order.

## Design

There are two main goals we need to achieve:
* Update the build to ensure that the dependencies between TFMs are reflected during the build and publish processes.
  * This involves the inner builds as well as the outer builds. An MSBuild invocation can start on the outer build, when for example "dotnet run" is invoked, or on the inner build, when for example "dotnet publish -f xxx" is invoked for that TFM.
* Collect the assets for any dependent (source) TFM and resolve any conflicts that might appear with assets in the embedding (target) TFM.

### Definitions

We establish a dependency between two TFMs by using an element in an itemgroup described as follows:
```xml
<StaticWebAssetsEmbeddingConfiguration Include="netx.y">
  <Platform>browser</Platform>
<StaticWebAssetsEmbeddingConfiguration>
```
This lets the project indicate that it wants to embedd the assets for the netx.y-browser TFM in the netx.y TFM.

#### Outer builds

Given a set of dependencies between TFMs, crosstargeting builds query each TFM for their "embedding configuration" and their "embedding rules". This is because the rules can be conditionally defined only for a given TFM. In the example above, the rule that the SDK actually includes is as follows:
```
```xml
<StaticWebAssetsEmbeddingConfiguration 
  Include="$(TargetFramework)" 
  Condition="'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' == '' and 
              $([MSBuild]::VersionGreaterThanOrEquals(8.0, $([MSBuild]::GetTargetFrameworkVersion($(TargetFramework)))))">
  <Platform>browser</Platform>
<StaticWebAssetsEmbeddingConfiguration>
```

The rule above says that if the TFM is netx.y and the platform is not browser, then include the netx.y-browser assets, but its only defined in the inner build for the netx.y tfm. This enables the SDK to define the rule in a way that it doesn't have to update on every version and it allows updating the TFM versions independently if needed.

Once we've retrieved all the embedding configurations and rules, we can compute the dependency graph and remove any embedded TFM from the list of TFMs that need to be build on the outer build. This is because the embedding TFM will retrieve the assets from the embedded TFM and that will cause the embedded TFM to be built.

#### Inner builds

Each inner build determines the available configurations by querying the outer build for the embedding configurations. With those configurations and the rules defined for the current target framework, it determines which other configurations it needs to embedd.

With that information, as part of resolving the static web assets inputs, it calls each configuration to retrieve the embedded assets for build and publish. Once those assets are made available to the rest of the build, the last process remains to "disambiguate" conflicting assets.

### Conflict resolution

For each asset that gets embedded, there are several possible actions we can take, depending on the outcome that we want. This largely depends on the type of asset and the role it plays among similar assets in the embedding TFM.
* Discovered assets: These are assets that are discovered by the build, like the wwwroot folder. In this case, we want to discard the assets from the embedded TFM in favor of the ones from the embedding TFM.
* Shared assets from dependencies: These are assets that are shared by the embedding TFM and the embedded TFM. In this case, we want to discard the assets from the embedded TFM in favor of the ones from the embedding TFM.
* Unique assets from dependencies: These are assets that are unique to the embedded TFM. In this case, we want to keep the assets from the embedded TFM.
* Generated assets: These are assets that are generated by the build, like scoped CSS files. In this case, we want to discard some, and let others participate in the build. For example, with scoped css files:
  * We want to discard the app bundle from the embedded TFM in favor of the one from the embedding TFM.
  * We want to keep the scoped css file bundle from the embedded TFM and let it get merged into the app bundle for the embedding TFM.

The decission to make in each case is highly dependent on the feature and the assets involved. To this effect, assets will now include two additional properties:
* AssetMergeSource: This property indicates the configuration that is responsible for including this asset into the embedding TFM, and serves as a way to differentiate embedded assets from assets provided by the project.
* AssetMergeBehavior: This property indicates the behavior that should be taken when merging this asset into the embedding TFM. This is a string that can be one of the following values:
  * Discard: This asset should not be taken into account when merging assets into an embedding TFM.
  * PreferTarget: This asset should be discarded if there is a conflict with an asset from the embedding TFM.
  * PreferSource: This asset should be used instead of the asset from the embedding TFM if there is a conflict.
  * Fail: This asset should cause the build to fail if there is a conflict with an asset from the embedding TFM.

### Features impacted by embeddable assets
* JS modules: Fail if there are multiple modules defined for different TFMs simultaneously.
* Scoped CSS: Merge the scoped css files from the embedded TFM into the app bundle for the embedding TFM. Isolation happens at the file level, not the TFM level.
* JS initializers: Fail if there is an initializer defined for different TFMs simultaneously.

## Packaged assets

Until now, there is no support for multitargeting in static web assets. In multitargeting builds, during the packing process we select the assets for the TFM that doesn't define a platform, as the representative assets for the package. Since now we support multitargeting on the unified project system, we need to update the packing process to support multitargeting.

* We will include the list of assets for each TFM in the package in their respective props files.
* We will put TFM specific assets into a folder named after the TFM.
* We will update our props files to include the TFM specific assets.

## Protocol

The protocol for embedding assets is as follows:
* `ResolveStaticWebAssetsCrossTargetingConfiguration`: This is called from the inner build on the outer build to retrieve all available embedding configurations. With this and the rules defined for the current TFM, the inner build determines which TFMs it needs to embed.
* `GetStaticWebAssetsCrosstargetingProjectConfiguration`: This is called from the outer build on the inner build to retrieve the embedding configuration for a given TFM.
* `ResolveEmbeddedProjectsStaticWebAssets`: This is called from the inner build to retrieve the assets for a given embedding configuration. The targets to run and the properties to pass are defined in the embedding configuration that was retrieved in the invocation to `ResolveStaticWebAssetsCrossTargetingConfiguration`.
* `ResolveEmbeddedConfigurationsReferencedProjects`: This retrieves the list of project references for a given embedding configuration. This allows us to retrieve the assets for projects that are only referenced in the embedding TFM.
* `ResolveEmbeddedConfigurationsPackageReferences`: This retrieves the list of package assets defined for a given configuration. This allows us to retrieve the assets for packages that are only referenced in the embedding TFM or assets that are specific to the TFM.

With this information, we extend the general static web assets process to include the assets from the embedded TFMs into the final ouput.

Conflicts between assets are resolved in the same way for project, referenced projects and referenced package assets that produce conflicts.

### Multitaget build

```mermaid
sequenceDiagram
OuterBuild-->>EmbeddingTfm: ResolveStaticWebAssetEmbeddingRules, GetStaticWebAssetsCrossTargetingConfiguration
OuterBuild-->>EmbeddedTfm: ResolveStaticWebAssetEmbeddingRules, GetStaticWebAssetsCrossTargetingConfiguration
OuterBuild-->>OuterBuild: UpdateCrossTargetingBuildProjects
OuterBuild-->>EmbeddingTfm: DispatchToInnerBuilds
EmbeddingTfm-->>EmbeddingTfm: ResolveStaticWebAssetsCrossTargetingConfiguration
EmbeddingTfm-->>OuterBuild: GetStaticWebAssetsCrossTargetingProjectConfiguration
OuterBuild-->>EmbeddingTfm: ResolveStaticWebAssetEmbeddingRules, GetStaticWebAssetsCrossTargetingConfiguration
OuterBuild-->>EmbeddedTfm: ResolveStaticWebAssetEmbeddingRules, GetStaticWebAssetsCrossTargetingConfiguration
EmbeddingTfm-->>EmbeddingTfm: ResolveEmbeddedProjectsStaticWebAssets
EmbeddingTfm-->>EmbeddedTfm: Build, GetCurrentProjectEmbeddedBuildStaticWebAssetItems
EmbeddingTfm-->>EmbeddingTfm: MergeStaticWebAssets
```

### Publish for net8.0

```mermaid
sequenceDiagram
EmbeddingTfm-->>EmbeddingTfm: ResolveStaticWebAssetsCrossTargetingConfiguration
EmbeddingTfm-->>OuterBuild: GetStaticWebAssetsCrossTargetingProjectConfiguration
OuterBuild-->>EmbeddingTfm: ResolveStaticWebAssetEmbeddingRules, GetStaticWebAssetsCrossTargetingConfiguration
OuterBuild-->>EmbeddedTfm: ResolveStaticWebAssetEmbeddingRules, GetStaticWebAssetsCrossTargetingConfiguration
EmbeddingTfm-->>EmbeddingTfm: ResolveEmbeddedProjectsStaticWebAssets
EmbeddingTfm-->>EmbeddedTfm: Publish, GetCurrentProjectEmbeddedBuildStaticWebAssetItems
EmbeddingTfm-->>EmbeddingTfm: MergeStaticWebAssets
```


SteveSandersonMS (approved):
We discussed this offline so I'm approving based on that conversation and what I can understand through a surface-level look. The core concepts definitely sound reasonable, though realistically I wouldn't be able to predict if the MSBuild logic here might be missing edge cases, and there are still other scenarios to be handled in follow-up PRs.

@MackinnonBuck I understand you also have context on this so please do still have a look through it if you're inclined!

MackinnonBuck (approved):

PR #33470, opened 2023-06-21
Prevent built-in artifacts output functionality from conflicting with Microsoft.Build.Artifacts SDK

### LINKED ISSUE #33052: MSBuild property ArtifactsPath is overloaded; used by both .Net 8 and Microsoft.Build.Artifacts
### Describe the bug

When building an existing project that uses the [MSBuild SDK Microsoft.Build.Artifacts](https://github.com/microsoft/MSBuildSdks/tree/main/src/Artifacts) with the .Net 8 SDK, we get an unexpected build error:

```
error NETSDK1199: The ArtifactsPath and UseArtifactsOutput properties cannot be set in a project file, due to MSBuild ordering constraints. They must be set in a Directory.Build.props file or from the command line. See https://aka.ms/netsdk1199 for more information.
```

It looks like the MSBuild variable `ArtifactsPath` is used both by Microsoft.Build.Artifacts and the .Net 8 build files, but with different meanings.

### To Reproduce

The Artifacts SDK is used by the MSBuild SDK project itself, which means we can use that repo to reproduce the problem:

````
git clone https://github.com/microsoft/MSBuildSdks.git
cd MSBuildSdks
dotnet build
````

This works without issues when using the .Net 7 SDK but breaks with .Net 8 SDK. At first glance this looks like a downstream problem, the MSBuild SDK needs to be updated for .Net 8, but this will have a cascading effect with .Net 8 being a build-breaking change across many first- and third-party repos which might warrant a bit of extra care and coordination.

### Further technical details
```
> dotnet --info
.NET SDK:
 Version:   8.0.100-preview.4.23260.5
 Commit:    2268e7b15c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.23471
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.100-preview.4.23260.5\

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.0-preview.4.23259.5
  Architecture: x64
  Commit:       84a3d0e37e

.NET SDKs installed:
  7.0.400-preview.23225.8 [C:\Program Files\dotnet\sdk]
  8.0.100-preview.4.23260.5 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-preview.4.23260.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.4.23259.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.0-preview.4.23260.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  Not set
```

### PR DESCRIPTION
**Pull request by dsplaisted**
Fixes #33052

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultArtifactsPath.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultArtifactsPath.props
@@ -20,7 +20,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
   
   <!-- Setting ArtifactsPath automatically opts in to the artifacts output format -->
-  <PropertyGroup Condition="'$(ArtifactsPath)' != ''">
+  <PropertyGroup Condition="'$(ArtifactsPath)' != '' And '$(UsingMicrosoftArtifactsSdk)' != 'true'">
     <UseArtifactsOutput Condition="'$(UseArtifactsOutput)' == ''">true</UseArtifactsOutput>
     <IncludeProjectNameInArtifactsPaths Condition="'$(IncludeProjectNameInArtifactsPaths)' == ''">true</IncludeProjectNameInArtifactsPaths>
     <_ArtifactsPathLocationType>ExplicitlySpecified</_ArtifactsPathLocationType>
--- a/src/Assets/TestProjects/ArtifactsSdkTest/ArtifactsSdkTest.sln
+++ b/src/Assets/TestProjects/ArtifactsSdkTest/ArtifactsSdkTest.sln
@@ -1,0 +1,28 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PackageReference", "PackageReference\PackageReference.csproj", "{78885455-FD34-4D8A-BB3D-D5DDB1AA2A4F}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MSBuildSdk", "MSBuildSdk\MSBuildSdk.csproj", "{6BFF7261-0DFB-4C37-954F-DBD11213FDEF}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{78885455-FD34-4D8A-BB3D-D5DDB1AA2A4F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{78885455-FD34-4D8A-BB3D-D5DDB1AA2A4F}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{78885455-FD34-4D8A-BB3D-D5DDB1AA2A4F}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{78885455-FD34-4D8A-BB3D-D5DDB1AA2A4F}.Release|Any CPU.Build.0 = Release|Any CPU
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Review by joeloff — APPROVED** [review]
Looks good. Should we expect to generalize the logic for other SDKs in the future?

File: src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultArtifactsPath.props [hunk 1]
```diff
@@ -20,7 +20,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
   
   <!-- Setting ArtifactsPath automatically opts in to the artifacts output format -->
-  <PropertyGroup Condition="'$(ArtifactsPath)' != ''">
+  <PropertyGroup Condition="'$(ArtifactsPath)' != '' And '$(UsingMicrosoftArtifactsSdk)' != 'true'">
```
jeffkl, line 23: When a user references Microsoft.Build.Artifacts as a package, its `.props` is imported by `.nuget.g.props` which is imported after this.  So I don't think `UsingMicrosoftArtifactsSdk` will be set yet since this is imported immediately after `Directory.Build.props`.

jeffkl:
I've also created a pull request to disable Microsoft.Build.Artifacts if the .NET SDK's built-in functionality is enabled: https://github.com/microsoft/MSBuildSdks/pull/458

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultArtifactsPath.props
dsplaisted: This is only imported early on if `UseArtifactsOutput` or `ArtifactsPath` are set in Directory.Build.props: https://github.com/dotnet/sdk/blob/9d157dca94879daaf6f2d7ea7f8243546bbd2799/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props#L64-L65

If they are set later on (ie in the project file), then this file will be imported later, during the .targets evaluation: https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets#L42-L43

So if the Microsoft.Build.Artifacts NuGet package is being used and `ArtifactsPath` is set in Directory.Build.props, then the built-in SDK artifacts functionality will be used.  If `ArtifactsPath` is set in the project file (as seems to be suggested in the Microsoft.Build.Artifacts documentation), then the Microsoft.Build.Artifacts logic will be used.

Does this sound OK?  Do you know if anyone using Microsoft.Build.Artifacts is setting `ArtifactsPath` in Directory.Build.props?  IE, something like:

```xml
<Project>
  <PropertyGroup>
    <BaseArtifactsPath>$(MSBuildThisFileDirectory)artifacts</BaseArtifactsPath>
    <ArtifactsPath>$(BaseArtifactsPath)\$(MSBuildProjectName)</ArtifactsPath>
  </PropertyGroup>
</Project>
```

dsplaisted:
> Looks good. Should we expect to generalize the logic for other SDKs in the future?

No, this is because we added a specific property that was already in use by this specific SDK.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultArtifactsPath.props
dsplaisted: @jeffkl I went ahead and merged this because we wanted to get the fix in before the .NET 8 Preview 6 branch,  Let me know what you think about this and we can follow up with another change if need be.

PR #33471, opened 2023-06-21
When Hot Reload change fails to apply in the browser send full error …

tmat:
…message to the client so that it can be logged

### DIFF
--- a/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
+++ b/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
@@ -43,7 +43,8 @@ setTimeout(async function () {
       const payload = JSON.parse(message.data);
       const action = {
         'UpdateStaticFile': () => updateStaticFile(payload.path),
-        'BlazorHotReloadDeltav1': () => applyBlazorDeltas(payload.sharedSecret, payload.deltas),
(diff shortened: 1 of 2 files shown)


src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
```diff
@@ -43,7 +43,8 @@ setTimeout(async function () {
       const payload = JSON.parse(message.data);
       const action = {
         'UpdateStaticFile': () => updateStaticFile(payload.path),
-        'BlazorHotReloadDeltav1': () => applyBlazorDeltas(payload.sharedSecret, payload.deltas),
+        'BlazorHotReloadDeltav1': () => applyBlazorDeltas(payload.sharedSecret, payload.deltas, false),
+        'BlazorHotReloadDeltav2': () => applyBlazorDeltas(payload.sharedSecret, payload.deltas, true),
```
tmat, line 47: I preserved backward compat here, but I'm not sure what the versioning requirements actually here. Can this script be used across multiple versions of dotnet-watch? If not, why have version in the action name at all?

tlmii: I had a similar thought. I guess I could imagine a scenario where the injection occurs from one version of the SDK and dotnet-watch is run from another (particularly with local copies of the SDK like we have with arcade)? But it seems like you'd have to jump through some hoops to make that happen.

tlmii (approved):
LGTM

PR #33504, opened 2023-06-22
[8.0.1xx-preview6] Fix manifest tests

radical:
Error:
``
Expected manifest OriginalItemSpec of obj\Debug\${Tfm}\webcil\blazorwasm.wasm but found ${ProjectPath}\blazorwasm\obj\Debug\${Tfm}\webcil\blazorwasm.wasm
```

This is failing because the "found" path is fully expanded now, which
changed in https://github.com/dotnet/runtime/pull/87830 .

Update the paths to include the `${ProjectPath}`.


radical:
cc @steveisok

PR #33508, opened 2023-06-22
[main] fix typo

caenguidanos:


danmoseley:
@marcpopMSFT we don't take changes into localized files here, do we ? what's the correct procedure

marcpopMSFT:
Correct, the actual localization lives in a separate ADO repo and gets mirrored here so if we fix here, it'll just get overwritten later. Filed this issue to track not just the es overlocalization but also the zh-hans overlocalization:
https://ceapex.visualstudio.com/CEINTL/_workitems/edit/860186

marcpopMSFT:
The baseline string has now been updated and this is resolved: https://github.com/dotnet/sdk/pull/33735/files

PR #33576, opened 2023-06-26
adjust item templates to langVersion changes in Roslyn

YuliiaKovalova:
## Problem 

## Solution
After removing the evaluation of the default LangVersion, it was decided to adjust item templates and introduce a default value for the lang version in them.


JanKrivanek (approved):
Looks good to go!

PR #33586, opened 2023-06-26
Move wait and result calls to the thread pool to avoid deadlocks.

### LINKED ISSUE #23850: Reduce the use of .Result/.Wait()/.GetResult() in the codebase

As part of updating the dotnet/templating dependency, commit https://github.com/dotnet/sdk/pull/23677/commits/ce3896bde7dc97dc906723181f8eb0d7e5b5bd41 was needed to reduce the frequency of thread deadlocks in the test suite. These deadlocks seem to have come from blocking threads on results though the various means available in the threading APIs.  We should remove these blocking waits as much as possible.

We should reduce/remove the use of .Result/.Wait()/.GetResult() in the core SDK as much as possible. There are 49 uses of these APIs in the `dotnet` project and `Microsoft.DotNet.CLi.Utils` projects, with the remainder coming from Blazor, WebSdk, and other built-in tools that can come at a later date. The command-handler code pathways should be fully asynchronous, since the System.CommandLine handler model supports task-returning handlers.

marcpopMSFT:
Fixes #23850. While it doesn't actually make any of the result or wait calls async, it runs them on a separate thread so as to not lock up the main thread.

@baronfel , I haven't finished the GetResult changes as there are lots of ways to fix those and I wanted to talk with you first. Making more calls async is probably the ideal approach but is a lot more churn as that would flow through a lot of different calls. Let's chat.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/Command.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/Command.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
 using NuGet.Frameworks;
 
 namespace Microsoft.DotNet.Cli.Utils
@@ -69,8 +70,8 @@ namespace Microsoft.DotNet.Cli.Utils
                 var taskErr = _stdErr?.BeginRead(_process.StandardError);
                 _process.WaitForExit();
 
-                taskOut?.Wait();
-                taskErr?.Wait();
+                Task.Run(() => taskOut)?.Wait();
+                Task.Run(() => taskErr)?.Wait();
             }
 
             var exitCode = _process.ExitCode;
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/Command.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/Command.cs
@@ -70,8 +70,8 @@ namespace Microsoft.DotNet.Cli.Utils
                 var taskErr = _stdErr?.BeginRead(_process.StandardError);
                 _process.WaitForExit();
 
-                Task.Run(() => taskOut)?.Wait();
-                Task.Run(() => taskErr)?.Wait();
+                Task.Run(() => taskOut?.Wait());
+                Task.Run(() => taskErr?.Wait());
             }
 
             var exitCode = _process.ExitCode;
--- a/src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.TabCompletion.cs
+++ b/src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.TabCompletion.cs
@@ -159,7 +159,7 @@ namespace Microsoft.TemplateEngine.Cli.Commands
                 if (constraintEvaluationTask.IsCompletedSuccessfully)
                 {
                     //return only allowed templates
-                    return constraintEvaluationTask.Result;
+                    return Task.Run(() => constraintEvaluationTask).Result;
                 }
                 //if evaluation task fails, all the templates in a group are considered as allowed.
                 //in case the template may not be run, it will fail during instantiation.
--- a/src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/TemplateCommand.cs
+++ b/src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/TemplateCommand.cs
@@ -213,7 +213,7 @@ namespace Microsoft.TemplateEngine.Cli.Commands
                 }
             }
 
-            return instantiateTask.Result;
+            return Task.Run(() => instantiateTask).Result;
         }
 
         private void DisplayConstraintResults(IReadOnlyList<TemplateConstraintResult> constraintResults, TemplateCommandArgs templateArgs)
--- a/src/Cli/dotnet/CommandFactory/CommandResolution/MSBuildProject.cs
+++ b/src/Cli/dotnet/CommandFactory/CommandResolution/MSBuildProject.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.Build.Evaluation;
 using Microsoft.DotNet.Cli.Utils;
 using NuGet.Frameworks;
@@ -159,8 +160,8 @@ namespace Microsoft.DotNet.CommandFactory
             var lockFilePath = GetLockFilePathFromProjectLockFileProperty() ??
                 GetLockFilePathFromIntermediateBaseOutputPath();
 
-            return new LockFileFormat()
-                .ReadWithLock(lockFilePath)
+            return Task.Run(() => new LockFileFormat()
+                .ReadWithLock(lockFilePath))
                 .Result;
         }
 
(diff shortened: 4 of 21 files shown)


src/Cli/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -69,8 +70,8 @@ public CommandResult Execute(Action<Process> processStarted)
                 var taskErr = _stdErr?.BeginRead(_process.StandardError);
                 _process.WaitForExit();
 
-                taskOut?.Wait();
-                taskErr?.Wait();
+                Task.Run(() => taskOut)?.Wait();
+                Task.Run(() => taskErr)?.Wait();
```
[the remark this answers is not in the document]
marcpopMSFT: You're right. I was just blindly pasting to match what was in the linked commit but I think that one is wrong too: https://github.com/dotnet/sdk/commit/ce3896bde7dc97dc906723181f8eb0d7e5b5bd41 I need to sync with baronfel to see if he remembers the context and goals of that issue.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.TabCompletion.cs
```diff
@@ -159,7 +159,7 @@ private static IEnumerable<CliTemplateInfo> GetAllowedTemplates(TemplateConstrai
                 if (constraintEvaluationTask.IsCompletedSuccessfully)
                 {
                     //return only allowed templates
-                    return constraintEvaluationTask.Result;
+                    return Task.Run(() => constraintEvaluationTask).Result;
```
zivkan, line 162: Considering three lines up it already checks `containtEvaluationTask.IsCompletedSuccessfully` is true, what this code does now is block the current thread while a new task is scheduled on the thread pool. Once the new task is executed, it is a no-op, and signals back to the original thread that it's ready to continue. Then the operating system has to context switch that thread back into a CPU core.

Basically, this `Task.Run` does nothing more than create memory allocations and add latency, since the task is already complete.

src/Cli/dotnet/Telemetry/Telemetry.cs
```diff
@@ -112,7 +112,7 @@ public void Flush()
                 return;
             }
 
-            _trackEventTask.Wait();
+            Task.Run(() => _trackEventTask.Wait());
```
zivkan, line 115: Since this is in the `Flush` method, this feels incorrect. Since `Task.Run`'s task isn't waited, the `Flush` method will return before `_trackEventTask` has finished, and therefore the telemetry might not yet have been flushed. If the process terminates before the `_trackEventTask` completes, there will be data loss (unless the telemetry is backed by a file that is read on next process startup?)

src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/TemplateCommand.cs
```diff
@@ -213,7 +213,7 @@ internal async Task<NewCommandStatus> InvokeAsync(ParseResult parseResult, Cance
                 }
             }
 
-            return instantiateTask.Result;
+            return Task.Run(() => instantiateTask).Result;
```
zivkan, line 216: Similar to the comment on `InstantiateCommand.TabCompletion.cs`, the task here was already awaited on line 178. I don't believe this `Task.Run` does anything.

src/Cli/dotnet/CommandFactory/CommandResolution/MSBuildProject.cs
```diff
@@ -181,8 +182,8 @@ public bool TryGetLockFile(out LockFile lockFile)
                 return false;
             }
 
-            lockFile = new LockFileFormat()
-                .ReadWithLock(lockFilePath)
+            lockFile = Task.Run(() => new LockFileFormat()
+                .ReadWithLock(lockFilePath))
                 .Result;
```
zivkan, line 187: Unless there's something about subtle about async in .NET I'm not aware of, I don't understand what this achieves.

the `dotnet` CLI is a console app, and console apps don't have synchronization contexts. Although, this method is using `.Result`, so I'm not sure that synchronization contexts and thread affinity even matters.

Anyway, Task.Run will schedule a task to run on the thread pool. Once a thread pool thread eventually picks up the work, it will execute the `LockFileFormat.ReadWithLock` method, then signal that the task is complete. Then the thread blocked on `.Result` will need to be scheduled by the OS onto a CPU core, and it will continue its work.

I don't understand what theoretical deadlock this might avoid. If all the thread pool threads are blocked on synchronous APIs, it will spin up new threads. Even if console apps had a SynchronizationContext on a single thread, I believe it only avoids deadlocks when the entire callstack uses async. A single `.Result` like this can still deadlock as soon as another methods needs to switch back to the synchronized thread.

src/Cli/dotnet/Installer/Windows/PipeStreamMessageDispatcherBase.cs
```diff
@@ -65,7 +65,7 @@ public void Connect()
                 if (_pipeStream is NamedPipeServerStream)
                 {
                     Task connectTask = Task.Factory.StartNew(() => ((NamedPipeServerStream)_pipeStream).WaitForConnection());
-                    connectTask.Wait(ConnectionTimeout);
+                    Task.Run(() => connectTask.Wait(ConnectionTimeout));
 
                     if (!connectTask.IsCompleted)
```
zivkan, line 70: This is a bug. The `Task.Run` will return immediately, before the `conntectTask.Wait` is even executed, so there's effectively zero chance that `connectTask` is complete. I didn't look at the test results, but I would't be surprised if this is a cause of a lot of the test failures.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -91,8 +92,8 @@ public void WriteAssembly(IAssemblySymbol assemblySymbol)
             compilationUnit = compilationUnit.NormalizeWhitespace(eol: Environment.NewLine);
 
             Document document = project.AddDocument(assemblySymbol.Name, compilationUnit);
-            document = Simplifier.ReduceAsync(document).Result;
-            document = Formatter.FormatAsync(document, DefineFormattingOptions()).Result;
+            document = Task.Run(() => Simplifier.ReduceAsync(document)).Result;
+            document = Task.Run(() => Formatter.FormatAsync(document, DefineFormattingOptions())).Result;
 
             document.GetSyntaxRootAsync().Result!
```
ericstj, line 98: Looks like you missed this one.   @ViktorHofer I wonder if we should make this method Async instead.

ViktorHofer: And the outer method (in GenAPIApp) that calls this method, as well? And then what about the outer outer outer method?...

ericstj: Probably only a couple layers to this (Main|Execute & GenAPIApp.Run) -- if the entrypoint supports async it could be plumbed all the way, if not we can decide at what layer to do stop.  I was just suggesting that _maybe_ here is not the right layer - we could make this async and use await and then decide at a higher layer (if at all).  My recollection is that the cost of a Task.Run is a lot higher than await -- especially if the work completes synchronously.  I don't think msbuild tasks can have async Execute so that's one point to stop, console apps can have async main though.

marcpopMSFT:
Tried doing a blind find/replace per the originally filed issue but looks like this will require more deliberate investigation of every location. Closing for now.

PR #33619, opened 2023-06-27
Fix printed version for APICompat.Tool and GenAPI.Tool

ViktorHofer:


ViktorHofer:
@carlossanlop can you please review? This prints the app's version instead of the OS's version.

ericstj (approved):
This looks reasonable.

It's a bit annoying that this is state-of-the-art for getting the version of the app.  That's a significant runtime cost to read that attribute at runtime.  I wonder if we could source-gen an internal constant in the assembly to avoid such a cost 🤔 That's what we did on NetFx - we'd codegen a type called `ThisAssembly` in the global namespace.

ViktorHofer:
Yeah I agree, I also find it weird that we need to retrieve the version at runtime instead of at compile time.

ericstj:
So long as no-one removes the AssemblyFileVersion attribute it will work with aot.  We could enable the compile time option with some infrastructure either in the repo or in the SDK.  That's what's currently generating the attribute: https://github.com/dotnet/sdk/blob/c24c13d485cb61db6a1d5db3cf106031dd75f830/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets#L190

The GetFileVersion API is still useful for getting version of other files, but for version of the same file we could have a better option.

PR #33643, opened 2023-06-28
Use install state in workload manifest provider

dsplaisted:
This PR contains a bug fix and a feature.

# Bug Fix

## Description

Fixes issue where when reading side-by-side workload manifests, the Manifest ID would be reported incorrectly.

## Customer Impact

If this change doesn't go into 7.0.4xx / VS 17.7, then VS 17.7 and its version of MSBuild will not work properly with future builds of the .NET 8 SDK (once we switch to side-by-side workload manifests).  The behavior will either be that the resolver will crash, preventing builds from succeeding, or that it won't successfully load any workloads.

## Regression?

- [ ] Yes
- [X] No - This was a bug introduced in a new feature (in #32471), and doesn't occur if the new feature (side-by-side workload manifests) isn't used.

## Risk

- [ ] High
- [ ] Medium
- [X] Low

## Verification

- [X] Manual
- [X] Automated

# Feature

## Description

Implements last part of resolver logic for https://github.com/dotnet/designs/pull/294, which is to honor workload install state.  This will support commands which install a different workload set version, for example `dotnet workload update --version 8.0.101`.

## Customer Impact

If this change doesn't go into 7.0.4xx / VS 17.7, then in some cases where a command is run that selects a specific workload set version, that would not be honored by Visual Studio 17.7 and its version of MSBuild.

A workaround would be to specify the desired workload set version in a global.json file.

## Regression?

- [ ] Yes
- [X] No

## Risk

- [ ] High
- [ ] Medium
- [X] Low

## Verification

- [ ] Manual
- [X] Automated


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadInstallType.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var installType = GetWorkloadInstallType(sdkFeatureBand, dotnetDir);
+
+            if (installType == InstallType.FileBased)
+            {
+                return Path.Combine(dotnetDir, "metadata", "workloads", sdkFeatureBand.ToString(), "InstallState");
+            }
+            else if (installType == InstallType.Msi)
+            {
+                return Path.Combine(Environment.GetEnvironmentVariable("ProgramData")!, "dotnet", "workloads", sdkFeatureBand.ToString(), "InstallState");
```
joeloff, line 60: Who creates this file?

Forgind:
Any chance you could isolate the changes that were just moving things from one class to another from the changes that had substantive code impact? The line count is high, but a lot of the early parts especially were just moving code 🙂

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadInstallType.cs
dsplaisted: The dotnet CLI will create it in certain cases, like when you run certain versions of `dotnet workload update`.  See here: https://github.com/dotnet/designs/blob/simplify-workload-versioning/accepted/2023/simplify-workload-versioning-implementation.md#rolling-back-and-pinning-workload-versions

joeloff: Suggest using the GetSpecialFolder APIs, (see https://github.com/dotnet/sdk/blob/587fe23396c91efa6be825a6c8eb4aabe5787fe2/src/Cli/dotnet/Installer/Windows/MsiPackageCache.cs#L86). That calls into SHGetSpecialFolderPath AFAIK. Environment variables can always be modified by the user and are less reliable

dsplaisted: @joeloff Thanks, fixed

dsplaisted:
> Any chance you could isolate the changes that were just moving things from one class to another from the changes that had substantive code impact? The line count is high, but a lot of the early parts especially were just moving code 🙂

@Forgind I've rebased the code moves into separate commits.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.InstallStateReader.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        class InstallState
+        {
+            public string? WorkloadSetVersion { get; set; }
+            public WorkloadSet? Manifests { get; set; }
+        }
+
+        static class InstallStateReader
+        {
+            public static InstallState ReadInstallState(string installStatePath)
```
Forgind, line 33: This looks very similar to GetWorkloadVersionFromGlobalJson. Can we combine the two methods?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -102,6 +104,23 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
                     }
                 }
             }
+            else
```
Forgind, line 107: (This is really a comment for line 96, but) GetrWorkloadVersionFromGlobalJson already seems to check if string.IsNullOrEmpty(globalJsonPath) and just returns null in that case, so we should be able to skip the outer if.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -179,7 +198,28 @@ void ProbeDirectory(string manifestDirectory)
...
+            //  Load manifests from install state
+            if (_manifestsFromInstallState != null)
+            {
+                foreach (var kvp in _manifestsFromInstallState.ManifestVersions)
+                {
+                    var manifestSpecifier = new ManifestSpecifier(kvp.Key, kvp.Value.Version, kvp.Value.FeatureBand);
+                    var manifestDirectory = GetManifestDirectoryFromSpecifier(manifestSpecifier);
+                    if (manifestDirectory == null)
+                    {
+                        throw new FileNotFoundException(string.Format(Strings.ManifestFromInstallStateNotFound, manifestSpecifier.ToString(), _installStateFilePath));
```
Forgind, line 220: I'm wondering if this throw is too aggressive...there are several manifests in this folder, and we only care about one, right? So if an irrelevant manifest is broken, do we actually care? Or are these already scoped down to just the manifests we care about?

Forgind (approved):

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/xlf/Strings.cs.xlf
```diff
@@ -77,6 +77,16 @@
         <target state="translated">Závislost manifestu úlohy {0} verze {1} je nižší než verze {2} požadovaná manifestem {3} [{4}].</target>
         <note />
       </trans-unit>
+      <trans-unit id="ManifestFromInstallStateNotFound">
+        <source>Workload manifest {0}, which was specified in {1}, was not found. Running "dotnet workload repair" may resolve this.</source>
+        <target state="new">Workload manifest {0}, which was specified in {1}, was not found. Running "dotnet workload repair" may resolve this.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="ManifestFromWorkloadSetNotFound">
+        <source>Workload manifest {0} from workload version {1} was not installed. Running "dotnet workload repair" may resolve this.</source>
```
joeloff, line 86: Should we lock the `dotnet workload repair` string, just in case it ends up getting translated?

joeloff (approved):

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/xlf/Strings.cs.xlf
dsplaisted: @joeloff Yes, what's the syntax for locking it?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: Good catch!  This was actually a bug where if you had a global.json that didn't specify a workload version, it would ignore the install state.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: It's already scoped down.  `_manifestsFromInstallState.ManifestVersions` is only the manifests that were specified in the `manifests` section of global.json, so we need to load each of those.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.InstallStateReader.cs
dsplaisted: I've refactored this so most of the common code is a method called in `JsonReader` now.

Forgind (approved):

dsplaisted:
@jaredpar [This commit](https://github.com/dotnet/sdk/pull/33643/commits/8ce7dcadf23c51bf53c93416bc37b54acd3bf7ad) seems to have caused the Full Framework CI leg to time out.  Is returning a `ref struct` from a method not supported or handled differently on .NET Framework?

PR #33682, opened 2023-06-29
Fix "Registry is not supported on this platform" with mono

iii-i:
Running `dotnet --info` with mono fails with:

[ERROR] FATAL UNHANDLED EXCEPTION: System.TypeInitializationException: The type initializer for 'Microsoft.Win32.Registry' threw an exception.
 ---> System.PlatformNotSupportedException: Registry is not supported on this platform.
   at Microsoft.Win32.RegistryKey.OpenBaseKeyCore(RegistryHive hKey, RegistryView view)
   at Microsoft.Win32.RegistryKey.OpenBaseKey(RegistryHive hKey, RegistryView view)
   at Microsoft.Win32.Registry..cctor()
   --- End of inner exception stack trace ---
   at Microsoft.DotNet.Cli.Utils.UILanguageOverride.Setup()
   at Microsoft.DotNet.Cli.Program.InitializeProcess()
   at Microsoft.DotNet.Cli.Program.Main(String[] args)

The culprit is the CurrentPlatformIsWindowsAndOfficiallySupportsUTF8Encoding() method: even though it wraps the usage of the Microsoft.Win32.Registry class in an "if" statement, the spec allows loading this class when JITing the method where it's used, leading to the failure above.

Fix by moving the Microsoft.Win32.Registry usage into a separate method. Add Debug.Assert() in order to pacify the checker.

Fixes: b6ca751324b8 ("[wip] restore encoding and only use on win that support utf 8 oficially")

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/UILanguageOverride.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.InteropServices;
 using System.Security;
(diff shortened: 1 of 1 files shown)


marcpopMSFT (approved):

marcpopMSFT:
Thank you for your contribution.

iii-i:
Can this be backported to .NET 7? The issue occurs there as well.

PR #33720, opened 2023-06-30
Fix CmdOption name and aliases

adamsitnik:
The `EntrypointArgsOption` option is already using `--entrypointargs` as it's name

https://github.com/dotnet/sdk/blob/6483a69e9a26df4f0fd598a523f6f09fab9e1c10/src/Containers/containerize/ContainerizeCommand.cs#L70


baronfel (approved):
Thanks for catching this

PR #33748, opened 2023-07-01
23_07_01-PullRequestToCompare

lucianoCircelli:
09:45
Using your self-hosted runner in a new workflow 
runs-on: self-hosted


lucianoCircelli:
23_07_01 12:39
I am not in my fork home, There are several stacks. It is above my understanding why runtimes can leave a custom serialization.
My excuse if a stress comes by tactlessness, I thought that pull request was for my workflow apparently, I have writes that goes everywhere.

PR #33756, opened 2023-07-03
Default to RazorLangVersion 8.0 in .NET 8.0

javiercn:


SteveSandersonMS (approved):

javiercn:
@SteveSandersonMS I think this needs a deeper look, since it appears to be causing a lot of test failures.

SteveSandersonMS:
Weird. It's giving errors like:

```
C:\h\w\B30A09EA\t\dotnetSdkTests\bo524oql.o30\Publish_WithI---129F25F8\Microsoft.NET.Sdk.Razor.SourceGenerators\Microsoft.NET.Sdk.Razor.SourceGenerators.RazorSourceGenerator\App_razor.g.cs(32,23):
error CS1061: 'RenderTreeBuilder' does not contain a definition for 'AddComponentParameter' and no accessible extension method 'AddComponentParameter' accepting a first argument of type 'RenderTreeBuilder' could be found (are you missing a using directive or an assembly reference?) 
```

... when of course that is exactly the API we added to support this. And it's been there for months now, so this can't be a dependency flow issue.

SteveSandersonMS:
My guess is the problem is something like this:

https://github.com/dotnet/sdk/blob/b72a60297b55c19c866af6aafc5fca52355931dd/src/Assets/TestProjects/BlazorWasmWithLibrary/blazorwasm/blazorwasm.csproj#L8-L11

The test projects are hardcoded to use old framework packages, regardless of the SDK or TFM. Are we expected to support this? If so we'll have to only set the `RazorLangVersion` to 8.0 if you're referencing an 8.0+ version of `Microsoft.AspNetCore.Components`. Of course it would be simpler still just to update the test projects to 8.0 and say people can only use `net8.0` with the 8.0 packages.

Either way, we should at least update the test projects, as it doesn't make sense that we're still testing only 5.0 projects.

javiercn:
This seems to require more work than just turning on the language version on the SDK (as tests need to be updated), so will leave this to https://github.com/dotnet/aspnetcore/issues/49167 to handle

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
```diff
@@ -44,6 +44,17 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <!-- Resolve the TFM-specific attributes conditionally. -->
   <Choose>
+      <When Condition=" '$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND $([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', '8.0')) ">
+        <PropertyGroup>
+          <_TargetingNETCoreApp30OrLater>true</_TargetingNETCoreApp30OrLater>
+          <_TargetingNET50OrLater>true</_TargetingNET50OrLater>
+          <_TargetingNET60OrLater>true</_TargetingNET60OrLater>
+          <_TargetingNET70OrLater>true</_TargetingNET70OrLater>
+          <_TargetingNET80OrLater>true</_TargetingNET80OrLater>
```
jaredpar, line 53: What are all these lines intended to do? There is no such equivalent for roslyn so I don't know what this solves for razor.

mkArtakMSFT:
@jjonescz now that the linked PR is merged, should this be simply reopened?

jjonescz:
@mkArtakMSFT yes, but we probably first need https://github.com/dotnet/sdk/pull/34057 so the Razor change flows here

javiercn:
@jjonescz @jaredpar @chsienki good to merge?

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
jjonescz: The 5.0 is used here:

https://github.com/dotnet/sdk/blob/c5bbadb84904b9e6f66b34247680893eaba4dc10/src/RazorSdk/Targets/Microsoft.NET.Sdk.Razor.Compilation.targets#L337-L353

The 6.0 is *seemingly* used here:

https://github.com/dotnet/sdk/blob/c5bbadb84904b9e6f66b34247680893eaba4dc10/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.5_0.targets#L439

But it's actually redefined in the same file, so it probably doesn't come from here.

Didn't find any usages of the others.

jjonescz (approved):

src/RazorSdk/Targets/Sdk.Razor.CurrentVersion.targets
javiercn: Normally we used to define these variables every time we added a new version, since there were features keyed out of them.

PR #33783, opened 2023-07-05
Switch the VM image of the macOS

v-wuzhai:


v-wuzhai:
@marcpopMSFT looks like macOS-latest refers to macOS-12 and not macOS-13 from [this document](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software).

chcosta:
@ilyas1974 , thoughts?

marcpopMSFT:
Ping @ilyas1974 on any guidance here if which image we should use.

marcpopMSFT:
@ilyas1974 thoughts?

dougbu:
not speaking for @ilyas1974 but my opinion is _it depends_&hellip; one advantage of using `vmImage: macOS-13` is you're getting the latest current OSX version and that may pick up some new and better features. one disadvantage is you're getting the latest **current** OSX version and `macOS-latest` could easily move to OSX 14 in the future. either way, Azure has some policy somewhere about updating `macOS-latest` and you might just want to wait for them

note we don't have much to do w/ hosted pool images beyond helping to file tickets when something goes South

ilyas1974:
sorry - to many balls in the air. When I chatted with the team that supports those systems, I was told that it would be at least a month before they had infrastructure in place to support the new operating system. At this point, should you need to utilize macOS 13 for your workloads, you are welcome to use the helix queues we have - osx.13.amd64.open and osx.13.arm64.open.

premun:
FWIW, `macos-latest` is 12 as per
https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software

marcpopMSFT:
maybe just leave it as is as it's fine on 12. The only reason we moved some to 13 is that there wasn't enough infrastructure on branding day to support all branches on 12.

Issue #31134, opened 2023-03-10
Download tool packages from NuGet feeds directly, instead of implicitly restoring


baronfel:
### Is your feature request related to a problem? Please describe.

Currently to install a .NET SDK tool we acquire the nuget package by restoring a temporary project. This has a number of side effects, which often show up as flaky restore errors because MSBuild concepts like Directory.Build.props, or other heirarchical notions, pollute the restore.

### Describe the solution you'd like
We should query NuGet APIs for explicit versions of tool packages to download, and download them via the NuGet API (of course handling authentication, NuGet.config files, etc in a consistent manner as `Restore` does).

baronfel:
@nkolev92 any insight into @KalleOlaviNiemitalo's question? To me, for tools specifically, we should be able to

* query NuGet for versions of the Package ID the user requested (adding the implied constraint of PackageType == DotnetToolPackage or whatever the named constant is)
* use the version the user requested (if any) to pick the latest matching version of that Package Id
* download and extract that NuGet package and continue about our day

nkolev92:
Tools depends on restore for assets and compatibility selection. 
By calling restore, you're also getting all the user configuration, handling of multiple feeds, plus things like source mapping are respected. 

I think it's a lot of code that you'd have to maintain separately that duplicates a lot of what restore does today.

baronfel:
When you say assets and compatibility - for tools we _know_ what the TFM, architecture, and OS platform should be - the ones from the currently-running host and SDK installation. Isn't that enough to serve as inputs into the selection algorithm?

nkolev92:
The SDK does know everything it needs to, but restore is not necessarily written like a library, but rather a tool, which is why I think maintaining multiple workflows could potentially be costly.

In the past, downloading was not done on the SDK side either, so there's some API knowledge on that end, so maybe those concerns are lesser now.

marcpopMSFT:
Merged!

PR #33835, opened 2023-07-06
Download tool packages from NuGet feeds directly, instead of implicitly restoring

JL03-Yue:
PR for issue https://github.com/dotnet/sdk/issues/31134

Currently, when installing a dotnet tool package, the code restores a temporary project using `toolPackageInstaller`. This PR updated the methods handling tools to use NuGet API when installing, updating, restoring, and uninstalling global and local tools.

This PR also changes the tests that test the installing, updating, restoring, and uninstalling global and local tools. It also controls the verbosity output from NuGet API when downloading tools.


src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,244 @@
...
+            string verbosity = null,
+            bool isGlobalTool = false
+            )
+        {
+            _toolDownloadDir = isGlobalTool ? _globalToolStageDir : _localToolDownloadDir;
+            var assetFileDirectory = isGlobalTool ? _globalToolStageDir : _localToolAssetDir;
+            _nugetPackageDownloader = new NuGetPackageDownloader.NuGetPackageDownloader(_toolDownloadDir);
+
+            NuGetVersion version = DownloadAndExtractPackage(packageLocation, packageId, _nugetPackageDownloader, _toolDownloadDir.Value).GetAwaiter().GetResult();
+            CreateAssetFiles(packageId, version, _toolDownloadDir, assetFileDirectory);
```
zivkan, line 66: Is the assets file actually used?  If all you need is the `DotnetToolSettings.xml` file, you could reduce a lot of complexity, and improve runtime performance, by searching `Directory.EnumerateFiles`, using `DotnetToolSettings.xml` as the search pattern, then using a regex to figure out all TFMs and RIDs, then finally use NuGet's APIs only to select the best/closest TFM and RID.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,266 @@
...
+                    // Delete the root if it is empty
+                    if (Directory.Exists(packageRootDirectory.Value) &&
+                        !Directory.EnumerateFileSystemEntries(packageRootDirectory.Value).Any())
+                    {
+                        Directory.Delete(packageRootDirectory.Value, false);
+                    }
+                });
+        }
+
+        private static void AddToolsAssets(
```
dsplaisted, line 110: Add comments to these methods that the implementation was copied from the `LockFileUtils` class in the NuGet code.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,266 @@
...
+            var filesInPackage = await _nugetPackageDownloader.ExtractPackageAsync(packagePath, new DirectoryPath(nupkgDir));
+
+            if (Directory.Exists(packagePath))
+            {
+                throw new ToolPackageException(
+                    string.Format(
+                        CommonLocalizableStrings.ToolPackageConflictPackageId,
+                        packageId,
+                        version.ToNormalizedString()));
+            }
```
dsplaisted, line 207: Shouldn't this be before the package is extracted if it's going to do anything?

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,244 @@
...
+            //  Create NuGetv3LocalRepository
+            NuGetv3LocalRepository localRepository = new(nugetLocalRepository.Value);
+            var package = localRepository.FindPackage(packageId.ToString(), version);
+
+            var collection = new ContentItemCollection();
+            collection.Load(package.Files);
+
+            //  Create criteria
+            var managedCriteria = new List<SelectionCriteria>(1);
+            var currentTargetFramework = NuGetFramework.Parse("net8.0");
```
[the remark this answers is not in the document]
dsplaisted: Yes, this needs to be dynamic.  Probably we want something like this:

```suggestion
            // Use major.minor version of currently running version of .NET
            var currentTargetFramework = new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, new Version(Environment.Version.Major, Environment.Version.Minor));
```

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
```diff
@@ -17,7 +19,8 @@
 namespace Microsoft.DotNet.Tools.Tool.Install
 {
     internal delegate IShellShimRepository CreateShellShimRepository(string appHostSourceDirectory, DirectoryPath? nonGlobalLocation = null);
-    internal delegate (IToolPackageStore, IToolPackageStoreQuery, IToolPackageInstaller) CreateToolPackageStoresAndInstaller(
+    
+    internal delegate (IToolPackageStore, IToolPackageStoreQuery, IToolPackageDownloader) CreateToolPackageStoresAndDownloader(
```
dsplaisted, line 23: Can you also delete the `IToolPackageInstaller` interface and everything that uses it?  I think by now anything that uses it should be dead code.

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallLocalInstaller.cs
```diff
@@ -70,13 +68,11 @@ public IToolPackage Install(FilePath manifestFile)
...
             {
-                IToolPackage toolDownloadedPackage =
-                    _toolPackageInstaller.InstallPackageToExternalManagedLocation(
+                IToolPackage toolDownloadedPackage = _toolPackageDownloader.InstallPackageAsync(
                         new PackageLocation(
                             nugetConfig: configFile,
                             additionalFeeds: _sources,
-                            // Fix https://github.com/dotnet/sdk/issues/23135
-                            rootConfigDirectory: manifestFile.GetDirectoryPath().GetParentPath()),
+                            rootConfigDirectory: manifestFile.GetDirectoryPath()),
```
dsplaisted, line 75: What's the reason for changing this path here?

src/Cli/dotnet/ToolPackage/IToolPackageDownloader.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Threading.Tasks;
+using Microsoft.DotNet.ToolPackage;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Cli.ToolPackage
+{
+    internal interface IToolPackageDownloader
+    {
+        IToolPackage InstallPackageAsync(PackageLocation packageLocation, PackageId packageId,
```
dsplaisted, line 12: If this has `Async` at the end of its name, then it should return `Task<IToolPackage>`, and the implementation should use `await` instead of `.GetAwaiter().Result`.  Then synchronous methods that call it would use `.GetAwaiter().GetResult()`.

Alternatively, you could change the name to `InstallPackage`.

dsplaisted (commented):
Great job getting this working!

Some of this is pretty complicated to review, I took a first pass and added my comments, but I will probably have to look at it again later.

Thanks!

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,266 @@
...
+        protected readonly DirectoryPath _localToolAssetDir;
+        
+
+        public ToolPackageDownloader(
+            IToolPackageStore store
+        )
+        {
+            _toolPackageStore = store ?? throw new ArgumentNullException(nameof(store)); ;
+            _globalToolStageDir = _toolPackageStore.GetRandomStagingDirectory();
+            _localToolDownloadDir = new DirectoryPath(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "nuget", "package"));
```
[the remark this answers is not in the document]
JL03-Yue: Yes, it is the same global package folder.

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
JL03-Yue: Sure. I wonder if it's a good idea to submit another PR for deleting `IToolPackageInstaller` related code. I'm worried if the change is involved in too many files that it will affect other code

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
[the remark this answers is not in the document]
zivkan: [`Settings.LoadDefaultSettings(solutionDirectory)`](https://github.com/NuGet/NuGet.Client/blob/ee7b9525640d9cee3c5108154b13982ad68743df/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L252) from NuGet.Configuration.dll will let you get an `ISettings` instance.

Then you can use [`SettingsUtility.GetGlobalPackagesFolder(settings)`](https://github.com/NuGet/NuGet.Client/blob/ee7b9525640d9cee3c5108154b13982ad68743df/src/NuGet.Core/NuGet.Configuration/Utility/SettingsUtility.cs#L242)

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
dsplaisted: I think we should try doing it in this PR.  There may be a lot of code to delete, but I think all of it is now dead code (ie won't be used), and code deletions like that are easy to review.

nkolev92:
This implementation *must* consider [PackageSourceMapping](https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping). We're moving away from an implementation that considers it, we must not lose that. 

I'll review in more detail later.

dsplaisted:
> This implementation _must_ consider [PackageSourceMapping](https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping). We're moving away from an implementation that considers it, we must not lose that.
> 
> I'll review in more detail later.

@nkolev92 @zivkan We're using NuGet APIs to download a package.  The code for that is [here](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs).  Is there a way to respect package source mapping when calling those APIs?

nkolev92:
Yeah, I'll link examples later when I review it.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        private static async Task<NuGetVersion> DownloadAndExtractPackage(
+            PackageLocation packageLocation,
+            PackageId packageId,
+            INuGetPackageDownloader _nugetPackageDownloader,
+            string hashPathLocation
+            )
+        {
+            var packageSourceLocation = new PackageSourceLocation(packageLocation.NugetConfig, packageLocation.RootConfigDirectory, null, packageLocation.AdditionalFeeds);
+            var packagePath = await _nugetPackageDownloader.DownloadPackageAsync(packageId, null, packageSourceLocation);
```
nkolev92, line 200: I see that the downloading is hidden deep in this package downloader. 

https://github.com/JL03-Yue/sdk/blob/e6ce77b8ea041a5a775a850b8ee5325e7f60626b/src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs#L313.

When dealing with a settings object, you can create a source mapping object. 

https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs#L66

Then anytime *before* you access a source, you'd check https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/src/NuGet.Core/NuGet.Configuration/PackageSourceMapping/PackageSourceMapping.cs#L38. 

PSM promises that there's no calls made to unmapped sources, so looking up a version on an unmapped source is not supposed to be done. 

https://github.com/NuGet/NuGet.Client/blob/4d815bf71aa1ce5fe8d9d9da6573303d7093c2f1/src/NuGet.Core/NuGet.DependencyResolver.Core/ResolverUtility.cs#L49

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs#L61

Restore itself also does lots of logging when PSM is used and when it is not. 
For example, when a source is *not* used, that's useful debugging information.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+            // string localToolPath = null
+        )
+        {
+            _toolPackageStore = store ?? throw new ArgumentNullException(nameof(store)); ;
+            _globalToolStageDir = _toolPackageStore.GetRandomStagingDirectory();
+            /*string localToolDownloadDir = localToolPath
+                ?? (Environment.GetEnvironmentVariable("NUGET_PACKAGES"))
+                ?? (Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "nuget", "package"));*/
+            string localToolDownloadDir = (Environment.GetEnvironmentVariable("NUGET_PACKAGES"))
+                ?? (Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "nuget", "package"));
```
zivkan, line 69: Are you avoiding the `globalPackageDirectory` value in nuget.config intentionally?  Currently with the restore-based tool install, it will consider nuget.config, meaning this would be a change in behaviour. Maybe that's acceptable, I don't know, but if I

I previously provided a comment with the APIs needed to get the global packages folder in the same way that NuGet does for all our other experiences: https://github.com/dotnet/sdk/pull/33835#discussion_r1295679650

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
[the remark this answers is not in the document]
JL03-Yue: @zivkan Thank you! What does solution directory refer to in this case?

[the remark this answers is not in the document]
dsplaisted: @JL03-Yue In this case I think it should just be the current directory, ie the directory the command was run from.

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -306,6 +308,18 @@ private IEnumerable<PackageSource> LoadNuGetSources(PackageSourceLocation packag
             PackageSourceProvider packageSourceProvider = new PackageSourceProvider(settings);
             defaultSources = packageSourceProvider.LoadPackageSources().Where(source => source.IsEnabled);
 
+            PackageSourceMapping packageSourceMapping = PackageSourceMapping.GetPackageSourceMapping(settings);
```
dsplaisted, line 311: The NuGetPackageDownloader is also used to download workload NuGet packages.  We don't yet want to force package source mapping on those downloads.  So we should add a constructor argument or a property to this class so that package source mapping can be enabled for tool installs but not workload package downloads.

@nkolev92 @zivkan FYI on this, we don't want to enable PSM for workload package downloads this late in the .NET 8 release, as we think it would be disruptive.  Do you think we should try to enable it for workload packages in a later release?

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -306,6 +308,18 @@ private IEnumerable<PackageSource> LoadNuGetSources(PackageSourceLocation packag
...
+
+            // filter package patterns if enabled            
+            if (packageSourceMapping?.IsEnabled == true)
+            {
+                IReadOnlyList<string> sources = packageSourceMapping.GetConfiguredPackageSources(packageId.ToString());
+
+                if (sources.Count == 0)
+                {
+                    throw new NuGetPackageInstallerException("No NuGet sources are defined or enabled");
```
dsplaisted, line 320: This error message should be a localized string.  The message should probably be something like:

> Package Source Mapping is enabled but no feeds matched the specified packed ID: {0}

@nkolev92 @zivkan @baronfel Thoughts on what this message should say?

We should also add tests for this failure case, as well as some basic other scenarios when package source mapping is enabled.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+        protected readonly DirectoryPath _localToolAssetDir;
+
+        public static readonly CliOption<VerbosityOptions> VerbosityOption = CommonOptions.VerbosityOption;
+
+        protected readonly string _runtimeJsonPath;
+
+        public ToolPackageDownloader(
+            IToolPackageStore store,
+            string runtimeJsonPathTest = null
+            // string localToolPath = null
```
dsplaisted, line 60: Delete commented out code, and I'd suggest renaming `runtimeJsonPathTest` to something like `runtimeJsonPathForTests` or `runtimeJsonPath`.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
dsplaisted: Yes, please use the NuGet API to get the global packages folder.  The settings should be created the same way as they are created in NuGetPackageDownloader (they are where the feeds and package source mapping are defined).

dsplaisted (commented):
Looking pretty good, great job!

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
zivkan: The purpose of package source mapping is to give customers control over supply chain security concerns. So, I'd say yes.

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
zivkan: The NuGet.Client repo has package source mapping enabled, and when I try `dotnet tool install -g zivkan.doesnotexist`, I get this output, with the current restore-based tool install:

> C:\Users\zivkan\AppData\Local\Temp\d0755727-b84b-4758-b239-72d92b6c2d9c\restore.csproj : error NU1100: Unable to resolv
e 'zivkan.doesnotexist (>= 0.0.0)' for 'net8.0'. PackageSourceMapping is enabled, the following source(s) were not cons
idered: dotnet-eng, dotnet-public, myget-legacy@Local, nuget-build, vside.
> C:\Users\zivkan\AppData\Local\Temp\d0755727-b84b-4758-b239-72d92b6c2d9c\restore.csproj : error NU1100: Unable to resolv
e 'zivkan.doesnotexist (>= 0.0.0)' for 'net8.0/any'. PackageSourceMapping is enabled, the following source(s) were not
considered: dotnet-eng, dotnet-public, myget-legacy@Local, nuget-build, vside.
> The tool package could not be restored.
> Tool 'zivkan.doesnotexist' failed to install. This failure may have been caused by:
> 
> * You are attempting to install a preview release and did not use the --version option to specify the version.
> * A package by this name was found, but it was not a .NET tool.
> * The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
> * You mistyped the name of the tool.
> 
> For more reasons, including package naming enforcement, visit https://aka.ms/failure-installing-tool

I don't know if you want to keep the NU1100 error code. [Our docs for NU1100](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1100#scenario-2) already tell customers to check their package source mapping config in nuget.config. But if you don't use more or less the exact same message, then maybe it'll be confusing to customers who ctrl-f on the page to find snippets of their error message, without actually reading the docs.

Anyway, my point is that NuGet adds " PackageSourceMapping is enabled, the following source(s) were not cons
idered: dotnet-eng, dotnet-public, myget-legacy@Local, nuget-build, vside." to the end of the existing message. So, whatever message you're planning to use for `dotnet tool install -g ToolDoesNotExist`, you can append this message to the end when package source mapping is enabled.

baronfel: I don't think that we should attempt to re-use the NuGet.Client message directly unless it is a publicly accessible resource that we could include via code. I just checked and it's an `internal static` so I wouldn't recommend that route. Instead, we should explain the problem and guide the user to the existing package source mapping documentation in this case:

> Package Source Mapping is enabled for this project but no package sources were configured for the package '{0}'.  See the documentation for Package Source Mapping at https://learn.microsoft.com/nuget/consume-packages/package-source-mapping#enable-by-manually-editing-nugetconfig for more details.

What context will this string be shown in? Will it be part of a larger message? 

### Separate topic, Chet going off the rails a bit now 

I think we can be much more authoritative with our guidance for users when the install fails now that we're using the NuGet APIs directly. Let's take the 4 options we currently show during package installation failures that @zivkan showed us above:

> You are attempting to install a preview release and did not use the --version option to specify the version.

We should be able to detect this now - if we're querying for versions of a particular package we can check if all of the versions are prereleases and if so tell the user that this is 100% the problem.

> A package by this name was found, but it was not a .NET tool.

Similarly, we can now detect this and say 100% that this is the problem

> The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.

We should also get exceptions directly from NuGet that are connectivity related and so could be more direct.

> You mistyped the name of the tool.

this is the only one that I think we couldn't detect - we could try a fallback package name search and return the top N from the configured package sources though in an attempt to be more helpful?

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
JL03-Yue: @dsplaisted Sounds good. Are there properties that can be used to differentiate a tool from workloads? Any examples would be appreciated:)

src/Cli/dotnet/NugetPackageDownloader/LocalizableStrings.resx
```diff
@@ -138,4 +138,7 @@
   <data name="FailedToValidatePackageSigning" xml:space="preserve">
     <value>Failed to validate package signing.</value>
   </data>
-</root>
+  <data name="FailedToGetPackageUnderPackageSourceMapping" xml:space="preserve">
+    <value>Package Source Mapping is enabled but no feeds matched the specified packed ID: {0}. See the documentation for Package Source Mapping at https://learn.microsoft.com/nuget/consume-packages/package-source-mapping#enable-by-manually-editing-nugetconfig for more details.</value>
```
MiYanni, line 142: It would be worth making this an aka.ms link instead. Let me know if you need help doing that.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,306 @@
...
+        public static readonly CliOption<VerbosityOptions> VerbosityOption = CommonOptions.VerbosityOption;
+
+        protected readonly string _runtimeJsonPath;
+
+        public ToolPackageDownloader(
+            IToolPackageStore store,
+            string runtimeJsonPathForTests = null
+        )
+        {
+            _toolPackageStore = store ?? throw new ArgumentNullException(nameof(store)); ;
```
MiYanni, line 62: nit: extra ` ;`

MiYanni (approved):
Looks great! 💪

src/Cli/dotnet/NugetPackageDownloader/LocalizableStrings.resx
```diff
@@ -138,4 +138,7 @@
   <data name="FailedToValidatePackageSigning" xml:space="preserve">
     <value>Failed to validate package signing.</value>
   </data>
+  <data name="FailedToGetPackageUnderPackageSourceMapping" xml:space="preserve">
+    <value>Package Source Mapping is enabled but no feeds matched the specified packed ID: {0}. See the documentation for Package Source Mapping at aka.ms/nuget-package-source-mapping for more details.</value>
```
dsplaisted, line 142:
```suggestion
    <value>Package Source Mapping is enabled but no feeds matched the specified packed ID: {0}. See the documentation for Package Source Mapping at https://aka.ms/nuget-package-source-mapping for more details.</value>
```

baronfel: Also a nit (on mobile at the moment): the phrase 'packed ID' should probably be 'package ID', right?

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -47,7 +50,8 @@ internal class NuGetPackageDownloader : INuGetPackageDownloader
             IReporter reporter = null,
             RestoreActionConfig restoreActionConfig = null,
             Func<IEnumerable<Task>> timer = null,
-            bool verifySignatures = false)
+            bool verifySignatures = false,
```
joeloff, line 53: I'm guessing we never actually require the tool packages to be signed, but how do users trust them? Probably something to discuss for another PR - I'd be curious to know whether there was a threat model for dotnet tools.

src/Cli/dotnet/NugetPackageDownloader/LocalizableStrings.resx
joeloff: Also, a missing comma before the "but"

joeloff: And maybe lock the URL in the localization comments. Vendors should know not to translate, but locked strings are supposed to fail validation tests on the vendor side.

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
baronfel: Nuget has commands for managing the nuget trust store - see 'dotnet nuget trust ---help' for examples of that.

src/Tests/dotnet.Tests/CommandTests/ToolInstallLocalCommandTests.cs
zivkan, line 1: I had a quick look at this file, and I don't see any tests for package source mapping (fair enough, NuGet implemented the feature long after `dotnet tool install` was first created)

There are two scenarios I think are worth testing, and possibly a third that to be considered:

1. the package (tool) requested doesn't have any mapping. Does error message tell customer to check their package source mapping config?
2. The package (tool) requested has a mapping (for example, they have `pattern="*"` for the feed `contoso-private`), but the package is not found in that feed. Will the error message tell them that their nuget.org feed was not considered and they should check their package source mapping config?
3. nuget.config has two package sources, and both package sources have the same package id & version, but different contents. Will `dotnet tool install` correctly use the package source mapping config to download the expected package.

JL03-Yue: Thank you! It's added in 4d15bc4

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
nkolev92: There are multiple parts to signature verification. 
1. Does a package have a signature, is it valid?
2. Do I trust this package's signer. 

The `dotnet nuget trust `  command handles the latter part. It basically allows you to "limit" the number of signers you trust by virtue of only allowing certain author/repositories/certificates.

For #1, NuGet *always* verifies the signature regardless of what's enabled through nuget trust. The verification happens on windows, and @dtivel recently did the work to enable that verification on linux.

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -306,6 +311,18 @@ private IEnumerable<PackageSource> LoadNuGetSources(PackageSourceLocation packag
             PackageSourceProvider packageSourceProvider = new PackageSourceProvider(settings);
             defaultSources = packageSourceProvider.LoadPackageSources().Where(source => source.IsEnabled);
 
+            packageSourceMapping = packageSourceMapping ?? PackageSourceMapping.GetPackageSourceMapping(settings);
+
+            // filter package patterns if enabled            
+            if (_isNuGetTool && packageSourceMapping?.IsEnabled == true)
+            {
+                IReadOnlyList<string> sources = packageSourceMapping.GetConfiguredPackageSources(packageId.ToString());
```
nkolev92, line 319: So I use we get the configured package sources, but I don't them being used anywhere. 

Am I missing something?

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+            }
+
+            // Extract the package
+            var nupkgDir = Path.Combine(hashPathLocation, packageId.ToString(), version.ToString());
+            var filesInPackage = await _nugetPackageDownloader.ExtractPackageAsync(packagePath, new DirectoryPath(nupkgDir));
+
+            return version;
+        }
+
+        private static void CreateAssetFiles(
```
nkolev92, line 255: I'm sure @zivkan brought it up somewhere, but creating an assets file is probably not the best long term approach. Since you're already doing a custom installation, are there are plans to move away from the assets files?

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+            // Use major.minor version of currently running version of .NET
+            var currentTargetFramework = new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, new Version(Environment.Version.Major, Environment.Version.Minor));
+
+            var standardCriteria = conventions.Criteria.ForFrameworkAndRuntime(
+                currentTargetFramework,
+                RuntimeInformation.RuntimeIdentifier);
+            managedCriteria.Add(standardCriteria);
+
+            //  Create asset file
+            if (lockFileLib.PackageType.Contains(PackageType.DotnetTool))
```
nkolev92, line 295: This is the only place where we check the `DotnetTool` type, how does it manifest when a package without the correct type is attempted to be installed?

nkolev92: I think it'd be great to add a test for this. 

Currently all of this validation happens NuGet side and that's where the tests are.

We have https://github.com/NuGet/NuGet.Client/blob/40648c3cb3de37cf62b194accdc8e7117b36245f/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetToolTests.cs I think it'd be g reat to just make sure each of the appropriate tests there has an equivalent in this repo.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
zivkan: https://github.com/dotnet/sdk/pull/33835#discussion_r1282977739

Yes, if the only file that matters is `DotnetToolSettings.xml`, then that can be found with much less complexity via `Directory.EnumerateFiles`.

aortiz-msft:
Given that installing a tool is an explicit action, I'm open to downloading tools outside SourceMapping scope.

nkolev92:
> @nkolev92 @dsplaisted - Given that installing a tool is an explicit action, I'm open to downloading tools outside SourceMapping scope. Thoughts?

That's not always the case. Tools are downloaded in the repo scope. https://learn.microsoft.com/en-us/dotnet/core/tools/local-tools-how-to-use

These have a manifest and are repo bound. 

+ regardless of the scope, tools still run the risk of indeterminism that source mapping solves.

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
JL03-Yue: Thanks for the comment! I believe the `sources` are referred in line 321 when counting if `sources.Count == 0`

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
JL03-Yue: to be follow up in #35283

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
JL03-Yue: Thanks for the comment! To be follow up in #35283

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
JL03-Yue: Thanks for the comment! To be follow up in #35283

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
JL03-Yue: Thank you for the comment~To be follow up in 35289

heng-liu:
NuGet warns for http source since NuGet 6.3 (https://devblogs.microsoft.com/nuget/https-everywhere/)
Before this change, dotnet tool install will run implicit restore so NuGet is able to warn customer about http source when installing dotnet tool.
However, after this change, there will be no warning for http source when installing dotnet tool. 
Hi @JonDouglas @aortiz-msft @nkolev92 @zivkan do you have any concerns about this?

nkolev92:
> NuGet warns for http source since NuGet 6.3 (https://devblogs.microsoft.com/nuget/https-everywhere/) Before this change, dotnet tool install will run implicit restore so NuGet is able to warn customer about http source when installing dotnet tool. However, after this change, there will be no warning for http source when installing dotnet tool. Hi @JonDouglas @aortiz-msft @nkolev92 @zivkan do you have any concerns about this?

That's a great catch! I think that should be added as well.

JonDouglas:
We want to have a consistent experience everywhere we can. Since most of that is implemented at restore, could we somehow provide a central API or similar piping for NuGet APIs as well? That way it is rather easy for people to contribute these types of changes without having to know about all the [NuGet best security practices](https://learn.microsoft.com/en-us/nuget/concepts/security-best-practices) and years of work going into restore security?

nkolev92:
> We want to have a consistent experience everywhere we can. Since most of that is implemented at restore, could we somehow provide a central API or similar piping for NuGet APIs as well? That way it is rather easy for people to contribute these types of changes without having to know about all the [NuGet best security practices](https://learn.microsoft.com/en-us/nuget/concepts/security-best-practices) and years of work going into restore security?

http warnings, PSM, signature verification are concerns at different parts of restore, so they're not really implemented at the same level.

Adding a central API end up being something that's not used anywhere within the NuGet code itself, so we wouldn't know how to best author it.
After some discussions (maybe @zivkan was a part of that?), downloading was implemented on the SDK side (plus time constraints).

@zivkan mentioned this in a side-chat, but I think this is showing that we need to add some tests testing NuGet scenarios within in this repo, regardless of how stuff gets implemented :)

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+            ISettings settings = Settings.LoadDefaultSettings(Directory.GetCurrentDirectory());
+            _localToolDownloadDir = new DirectoryPath(SettingsUtility.GetGlobalPackagesFolder(settings));
+            
+            _localToolAssetDir = new DirectoryPath(PathUtilities.CreateTempSubdirectory());
+            _runtimeJsonPath = runtimeJsonPathForTests ?? Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "RuntimeIdentifierGraph.json");
+        }
+
+        public IToolPackage InstallPackage(PackageLocation packageLocation, PackageId packageId,
+            VersionRange versionRange = null,
+            string targetFramework = null,
```
dsplaisted, line 78: This should be passed through to the `CreateAssetFile` method.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+            NuGetv3LocalRepository localRepository = new(nugetLocalRepository.Value);
+            var package = localRepository.FindPackage(packageId.ToString(), version);
+
+            var collection = new ContentItemCollection();
+            collection.Load(package.Files);
+
+            //  Create criteria
+            var managedCriteria = new List<SelectionCriteria>(1);
+            // Use major.minor version of currently running version of .NET
+            var currentTargetFramework = new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, new Version(Environment.Version.Major, Environment.Version.Minor));
```
dsplaisted, line 287: We should take the target framework to use as a parameter to this method.  If it's not set then we should set it to the default value.

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallLocalInstaller.cs
```diff
@@ -33,20 +36,15 @@ internal class ToolInstallLocalInstaller
...
-                _toolPackageInstaller = toolPackageStoresAndInstaller.installer;
-            }
-            else
-            {
-                _toolPackageInstaller = toolPackageInstaller;
-            }
-
+            (IToolPackageStore store,
+                IToolPackageStoreQuery,
+                IToolPackageDownloader installer) toolPackageStoresAndDownloader
```
dsplaisted, line 41: Nit: rename this to `downloader`.

```suggestion
                IToolPackageDownloader downloader) toolPackageStoresAndDownloader
```

dsplaisted (commented):
I still want to take a look at the tests, but I think I've reviewed everything else in detail now.

Let's take a look at the feedback together.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,311 @@
...
+                                              verbosity == VerbosityOptions.detailed.ToString() ||
+                                              verbosity == VerbosityOptions.diag.ToString() ||
+                                              verbosity == VerbosityOptions.diagnostic.ToString()))
+                    {
+                        nugetLogger = new NuGetConsoleLogger();
+                    }
+                    _toolDownloadDir = isGlobalTool ? _globalToolStageDir : _localToolDownloadDir;
+                    var assetFileDirectory = isGlobalTool ? _globalToolStageDir : _localToolAssetDir;
+                    _nugetPackageDownloader = new NuGetPackageDownloader.NuGetPackageDownloader(_toolDownloadDir, verboseLogger: nugetLogger, isNuGetTool: true);
+                    rollbackDirectory = _toolDownloadDir.Value;
```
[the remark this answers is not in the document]
JL03-Yue: TBD: tests

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
```diff
@@ -60,7 +63,8 @@ internal class ToolInstallGlobalOrToolPathCommand : CommandBase
             _toolPath = parseResult.GetValue(ToolAppliedOption.ToolPathOption);
             _architectureOption = parseResult.GetValue(ToolInstallCommandParser.ArchitectureOption);
 
-            _createToolPackageStoresAndInstaller = createToolPackageStoreAndInstaller ?? ToolPackageFactory.CreateToolPackageStoresAndInstaller;
+            _createToolPackageStoresAndDownloader = createToolPackageStoreAndDownloader ?? ToolPackageFactory.CreateToolPackageStoresAndDownloader;
```
dsplaisted, line 66: Take a look at `ParseResultExtension.VersionRange`.  That has code to get the latest version of a package if it's not specified, which I think we can delete.

dsplaisted (commented):
I've finished reviewing the tests now too.  Some of the feedback here we've already discussed live.

I also think we should add tests for some scenarios where the package has already been downloaded or the tool has already been installed in another local tools folder.

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -306,6 +397,19 @@ private IEnumerable<PackageSource> LoadNuGetSources(PackageSourceLocation packag
...
+
+                if (sources.Count == 0)
+                {
+                    throw new NuGetPackageInstallerException(string.Format(LocalizableStrings.FailedToGetPackageUnderPackageSourceMapping, packageId));
+                }
+                defaultSources = defaultSources.Where(source => sources.Contains(source.Name));
```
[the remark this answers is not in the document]
JL03-Yue: To be change defaultSource.Count == 0 error messages and update the test

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,319 @@
...
+                        VerbosityOptions verbosityOption;
+                        Enum.TryParse(verbosity, out verbosityOption);
+                        if (verbosityOption.IsDetailedOrDiagnostic())
+                        {
+                            nugetLogger = new NuGetConsoleLogger();
+                        }
+                    }
+
+                    var versionString = versionRange?.OriginalString ?? "*";
+                    versionRange = VersionRange.Parse(versionString);
```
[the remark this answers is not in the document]
JL03-Yue: The `versionRange` inside `NugetPackageDownloader.GetBestPackageVersionAsync` doesn't seem to be nullable, and either way it does not affecting the outcome of the code :)

src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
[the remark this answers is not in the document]
JL03-Yue: Tests updated at `src/Tests/Microsoft.DotNet.PackageInstall.Tests/NuGetPackageInstallerTests.cs` line 169-210

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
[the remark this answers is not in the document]
JL03-Yue: Tests added `src/Tests/Microsoft.DotNet.PackageInstall.Tests/ToolPackageDownloaderTests.cs` line 361 - 450

src/Cli/dotnet/NugetPackageDownloader/LocalizableStrings.resx
```diff
@@ -138,4 +138,10 @@
   <data name="FailedToValidatePackageSigning" xml:space="preserve">
     <value>Failed to validate package signing.</value>
   </data>
-</root>
+  <data name="FailedToFindSourceUnderPackageSourceMapping" xml:space="preserve">
+    <value>Package Source Mapping is enabled, but no source found under the specified package ID: {0}. See the documentation for Package Source Mapping at https://aka.ms/nuget-package-source-mapping for more details.</value>
+  </data>
+  <data name="FailedToMapSourceUnderPackageSourceMapping" xml:space="preserve">
+    <value>Package Source Mapping is enabled, but no source mapped under the specified package ID: {0}. See the documentation for Package Source Mapping at https://aka.ms/nuget-package-source-mapping for more details.</value>
+  </data>
```
dsplaisted, line 146: Suggested text edits.  For the second one you'll need to pass in an additional string format parameter, :

```suggestion
  <data name="FailedToFindSourceUnderPackageSourceMapping" xml:space="preserve">
    <value>Package Source Mapping is enabled, but no sources were configured for the specified package ID: {0}. See the documentation for Package Source Mapping at https://aka.ms/nuget-package-source-mapping for more details.</value>
  </data>
  <data name="FailedToMapSourceUnderPackageSourceMapping" xml:space="preserve">
    <value>Package Source Mapping is enabled, but the sources ({1}) mapped for the specified package ID {0} were not configured. See the documentation for Package Source Mapping at https://aka.ms/nuget-package-source-mapping for more details.</value>
  </data>
```


@baronfel @nkolev92 @zivkan Any other suggestions on the error message wording?

src/Tests/Microsoft.DotNet.PackageInstall.Tests/ToolPackageDownloaderTests.cs
```diff
@@ -443,70 +322,142 @@ public void GivenAEmptySourceAndNugetConfigInstallSucceeds(bool testMockBehavior
...
                 {
-                    installer.InstallPackage(new PackageLocation(additionalFeeds: new[] { source }),
+                    downloader.InstallPackage(new PackageLocation(additionalFeeds: new[] { source }),
                         packageId: TestPackageId,
+                        verbosity: TestVerbosity,
                         versionRange: VersionRange.Parse(TestPackageVersion),
                         targetFramework: _testTargetframework);
 
-                    FailedStepAfterSuccessRestore();
+                    FailedStepAfterSuccessDownload();
```
dsplaisted, line 430: I think it would be a good idea to assert that `localToolDownloadDir` exists before throwing the exception.  That would help us be sure that we're checking the right path.

dsplaisted (approved):
Great job!  We're almost there, just a few remaining comments.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,330 @@
...
+                    var assetFileDirectory = isGlobalTool ? _globalToolStageDir : _localToolAssetDir;
+                    var nugetPackageDownloader = new NuGetPackageDownloader.NuGetPackageDownloader(toolDownloadDir, verboseLogger: nugetLogger, isNuGetTool: true);
+
+                    var packageSourceLocation = new PackageSourceLocation(packageLocation.NugetConfig, packageLocation.RootConfigDirectory, null, packageLocation.AdditionalFeeds);
+                    NuGetVersion packageVersion = nugetPackageDownloader.GetBestPackageVersionAsync(packageId, versionRange, packageSourceLocation).GetAwaiter().GetResult();
+
+                    rollbackDirectory = isGlobalTool ? toolDownloadDir.Value: Path.Combine(toolDownloadDir.Value, packageId.ToString(), packageVersion.ToString());  
+
+                    NuGetv3LocalRepository nugetPackageRootDirectory = new(Path.Combine(_toolPackageStore.GetRootPackageDirectory(packageId).ToString().Trim('"'), packageVersion.ToString()));
+                    var globalPackage = nugetPackageRootDirectory.FindPackage(packageId.ToString(), packageVersion);
```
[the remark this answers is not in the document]
JL03-Yue: The returned `_toolPackageStore.GetRootPackageDirectory(packageId).ToString()` is wrapped in double quote. This caused some errors in `Path.Combine`, and after debugging, I found `.Trim()` can solve the problem.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -0,0 +1,330 @@
...
+                    var toolDownloadDir = isGlobalTool ? _globalToolStageDir : _localToolDownloadDir;
+                    var assetFileDirectory = isGlobalTool ? _globalToolStageDir : _localToolAssetDir;
+                    var nugetPackageDownloader = new NuGetPackageDownloader.NuGetPackageDownloader(toolDownloadDir, verboseLogger: nugetLogger, isNuGetTool: true);
+
+                    var packageSourceLocation = new PackageSourceLocation(packageLocation.NugetConfig, packageLocation.RootConfigDirectory, null, packageLocation.AdditionalFeeds);
+                    NuGetVersion packageVersion = nugetPackageDownloader.GetBestPackageVersionAsync(packageId, versionRange, packageSourceLocation).GetAwaiter().GetResult();
+
+                    rollbackDirectory = isGlobalTool ? toolDownloadDir.Value: Path.Combine(toolDownloadDir.Value, packageId.ToString(), packageVersion.ToString());  
+
+                    NuGetv3LocalRepository nugetPackageRootDirectory = new(Path.Combine(_toolPackageStore.GetRootPackageDirectory(packageId).ToString().Trim('"'), packageVersion.ToString()));
```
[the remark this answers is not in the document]
JL03-Yue: This works for local tools. For global tools, the previous implementation generate the path as `toolDirectory`/`packageId`/`packageVersion`/`packageId`/`packageVersion`/tool content, and thus we need to look down one more level. This took up some time in debugging

src/Cli/dotnet/NugetPackageDownloader/LocalizableStrings.resx
baronfel: This is probably a bit complex of an ask - do we know the path to the NuGet.config file that defines the mapping? My guess is somewhere between 'yes, but...' and  'no' because the config files are hierarchical. Could we instead point to the _closest_ config file with package source mapping configuration items here? We tell the users that we couldn't find any mappings for a specific package (great!) and where to learn more about it (awesome!) but not where to make the changes (womp womp :cry:).

The rest of the message is wonderful - good set-up with describing the problem and guiding the user to where information about a solution is.

JL03-Yue: Thanks! This is a really great idea! It is definitely worth to follow up on this proposal later as it really give specific guidance for customers.😀 Currently, I'd love to update the rest of error message to include the source name which are not configured.

PR #33863, opened 2023-07-07
Handle when the target graph is TF/RID to pull out the TF piece

marcpopMSFT:
This was being reported as an issue by vendors against a specific test application where 2.1/win-x64 was being passed through to nuget and throwing an exception.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
@@ -1520,7 +1520,8 @@ namespace Microsoft.NET.Build.Tasks
                         log.LibraryId == package.Name &&
                         log.TargetGraphs.Any(tg =>
                         {
-                            var parsedTargetGraph = NuGetFramework.Parse(tg);
+                            var parts = tg.Split(Path.AltDirectorySeparatorChar);
+                            var parsedTargetGraph = NuGetFramework.Parse(parts[0]);
                             var alias = _lockFile.PackageSpec.TargetFrameworks
                                 .FirstOrDefault(tf => tf.FrameworkName == parsedTargetGraph)
                                 ?.TargetAlias ?? tg;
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
```diff
@@ -1520,7 +1520,8 @@ string GetPackageDiagnosticLevel(LockFileLibrary package)
                         log.LibraryId == package.Name &&
                         log.TargetGraphs.Any(tg =>
                         {
-                            var parsedTargetGraph = NuGetFramework.Parse(tg);
+                            var parts = tg.Split(Path.AltDirectorySeparatorChar);
```
nkolev92, line 1523: Maybe use LockFile.DirectorySeparatorChar directly, since that's what NuGet uses.

marcpopMSFT: Good point. Didn't check if that was actually available to use.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildANetCoreApp.cs
```diff
@@ -800,6 +801,28 @@ public void ItHasNoPackageReferences()
...
+            testProject.PackageReferences.Add(new TestPackageReference("Nuget.Common","6.5.7"));
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProject, testProject.Name);
+
+            var buildCommand = new BuildCommand(testAsset);
+
+            buildCommand
+                .Execute()
+                .Should()
+                .Pass();
```
dsplaisted, line 823: It's probably worth checking that the correct NU1603 warinng is generated

```suggestion
                .Pass()
                .And
                .HaveStdOutContaining("NU1603");
```

dsplaisted (approved):

Issue #32458, opened 2023-05-11
GenAPI produces TupleElementNamesAttribute reference in addition to the tuple itself


mthalman:
When using GenAPI on the Microsoft.CodeAnalysis.Common.4.5.0, it produces the following code snippet for the `IncrementalGeneratorRunStep` class:

```C#
[System.Runtime.CompilerServices.TupleElementNames(new[] { "Source", "OutputIndex" })]
public System.Collections.Immutable.ImmutableArray<(IncrementalGeneratorRunStep Source, int OutputIndex)> Inputs { get { throw null; } }
```

This produces the following compiler error:

```
error CS8138: Cannot reference 'System.Runtime.CompilerServices.TupleElementNamesAttribute' explicitly. Use the tuple syntax to define tuple names.
```

Removal of the generated attribute corrects the problem.

Related: https://github.com/dotnet/arcade/issues/2028

ViktorHofer:
@ericstj should this be closed now?

ViktorHofer:
Fixed via https://github.com/dotnet/sdk/pull/33888

PR #33888, opened 2023-07-10
Ensure GenAPI doesn't emit reserved attributes

### LINKED ISSUE #33853: GenAPI emits nullable attributes
### Describe the bug
Now that the Nullable attributes are defined publicly in the framework instead of internally in the user assembly they are no longer excluded by the visibility filter.  We shouldn't emit these attributes, but instead emit the nullable annotations on the C# syntax.

See https://github.com/dotnet/sdk/pull/33841 for examples of the unwanted nullable attribu tes.

### To Reproduce
Compile an assembly that uses nullable annotations targeting a recent build of .NET 8.0 with https://github.com/dotnet/runtime/commit/d39e0c467dd614b2cade2f29fea93f9530ce6326, then run genapi on the resulting assembly.  

Expect: public API that matches what was defined.
Actual: public API with explicitly references to NullableAttribute, NullableContext, etc.

### Exceptions (if any)
None

ericstj:
Fixes https://github.com/dotnet/sdk/issues/33853, https://github.com/dotnet/sdk/issues/32458

These attributes should only be applied by the compiler and should be excluded from GenAPI.  SyntaxGenerator is responsible for understanding them and applying different syntax -- and I found a few places where it needs to be updated:
https://github.com/dotnet/roslyn/issues/68957
https://github.com/dotnet/roslyn/issues/68959

This is important to fix #33853 since that will appear as a regression for folks using GenAPI against libraries built against .NET 8.0 preview 7 or later.

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxNodeExtensions.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxNodeExtensions.cs
@@ -1,9 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-using System.Collections;
-using System.Collections.Generic;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Editing;
@@ -30,7 +27,7 @@ namespace Microsoft.DotNet.GenAPI
                     continue;
                 }
 
-                if (IsReservedAttribute(attribute.AttributeClass))
+                if (attribute.IsReserved())
                 {
                     continue;
                 }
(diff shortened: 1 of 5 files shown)


src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxNodeExtensions.cs
```diff
@@ -26,10 +29,45 @@ internal static class SyntaxNodeExtensions
...
+                "TupleElementNamesAttribute",
+                "NullableAttribute",
+                "NullableContextAttribute",
+                "NullablePublicOnlyAttribute",
+                "NativeIntegerAttribute",
+                "ExtensionAttribute",
+                "RequiredMemberAttribute",
+                "ScopedRefAttribute",
+                "RefSafetyRulesAttribute"
+            };
```
ViktorHofer, line 58: I assume that we today don't have any "reserved" assembly attributes that should be filtered out? Asking as this code is only for "member" attributes.

src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxNodeExtensions.cs
```diff
@@ -26,10 +29,45 @@ internal static class SyntaxNodeExtensions
...
+
+        /// <summary>
+        /// Determines if an attribute is a reserved attribute class -- these are attributes that may
+        /// only be applied by the compiler and are an error to be applied by the user in source.
```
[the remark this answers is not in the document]
ericstj: Good suggestion.  There is plenty of internal API that does this, but not public.  I filed a request: https://github.com/dotnet/roslyn/issues/68968

src/GenAPI/Microsoft.DotNet.GenAPI/SyntaxNodeExtensions.cs
ericstj: There are some that apply to the module, but not the assembly.  I'll add this to the assembly attribute case for consistency though.

andriipatsula (approved):

ViktorHofer (approved):

PR #33935, opened 2023-07-12
First-run: Minor code adjustments

MiYanni:
## Summary
Just a random set of adjustments/cleanup prior to digging into the first-run workload experience changes.

- Set var usage to have no suggestions
  - The team doesn't seem to have an actual preference based on the code I've looked at thus far. So, just turned off the suggested usage.
- Added `PerformanceMeasurement` to replace manual StopWatch creation
- Removed extraneous method usage and other minor cleanup to `DotnetFirstTimeUseConfigurer`
  - Removed `cliFallbackFolderPath` constructor parameter as it was unused
- Other minor code cleanup/adjustments

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
@@ -48,7 +48,7 @@ namespace Microsoft.DotNet.Cli.Utils
         }
     }
 
-    public class PerformanceMeasurement : IDisposable
+    public sealed class PerformanceMeasurement : IDisposable
     {
         private readonly Stopwatch _timer;
         private readonly Dictionary<string, double> _data;
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 
 namespace Microsoft.DotNet.Cli.Utils
@@ -45,6 +46,28 @@ namespace Microsoft.DotNet.Cli.Utils
 
             EntryPosted += Handler;
         }
+    }
+
+    public class PerformanceMeasurement : IDisposable
+    {
+        private readonly Stopwatch _timer;
+        private readonly Dictionary<string, double> _data;
+        private readonly string _name;
+
+        public PerformanceMeasurement(Dictionary<string, double> data, string name)
+        {
+            // Measurement is a no-op if we don't have a dictionary to store the entry.
+            if (data == null)
+            {
+                return;
+            }
+
+            _data = data;
+            _name = name;
+            _timer = Stopwatch.StartNew();
+        }
+
+        public void Dispose() => _data?.Add(_name, _timer.Elapsed.TotalMilliseconds);
     }
 
     public class BlockFilter : ITelemetryFilter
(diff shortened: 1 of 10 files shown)


src/Cli/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
MiYanni, line 1: Likely want to do a side-by-side to look at this file changes. No logic changes should have occurred.

src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
```diff
@@ -47,6 +48,28 @@ void Handler(object sender, InstrumentationEventArgs eventArgs)
...
+            {
+                return;
+            }
+
+            _data = data;
+            _name = name;
+            _timer = Stopwatch.StartNew();
+        }
+
+        public void Dispose() => _data?.Add(_name, _timer.Elapsed.TotalMilliseconds);
```
joeloff, line 70: Should this call GC.SuppressFinalize() and Dispose(disposing)?

MiYanni: I'm pretty sure that only applies when you have a finalizer, right? Like, if I had a `~PerformanceMeasurement()` method. I'll double check though to make sure.

MiYanni: Discussed offline and sealed the class to avoid any potential issues.

src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
```diff
@@ -47,6 +48,28 @@ void Handler(object sender, InstrumentationEventArgs eventArgs)
...
```
marcpopMSFT, line 70: Question, does the runtime trigger dispose immediately at the end of the using block or does it do it lazily? In other words, will this change end up adding (likely very small amounts of) time to the perf tracking because we're relying on an implicit dispose rather than an explicit elapsed time recording as it was before.

src/Cli/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
```diff
@@ -41,152 +39,83 @@ public class DotnetFirstTimeUseConfigurer
...
-                }
 
-                _firstTimeUseNoticeSentinel.CreateIfNotExists();
-                _performanceMeasurements?.Add("FirstTimeUseNotice Time", beforeFirstTimeUseNotice.Elapsed.TotalMilliseconds);
+                    _firstTimeUseNoticeSentinel.CreateIfNotExists();
+                }
             }
 
-            if (ShouldGenerateAspNetCertificate())
+            if (CanGenAspNetCert())
```
marcpopMSFT, line 83: Nit: CanGenerateAspNETCertificate (I prefer the written out version personally and we've done that most everywhere else in this file as far as I can tell)

marcpopMSFT (approved):

src/Cli/Microsoft.DotNet.Cli.Utils/TelemetryEventEntry.cs
MiYanni: That was the exact question I asked both Jacques and Smit. It will immediately run `Dispose` when exiting the scope of the `using`. So, statement order is:
- Last statement inside the `using`
- Dispose
- First statement outside the `using`

It is deterministic. We used this pattern heavily in our C# code generator on a previous team. You can see an [example here](https://github.com/Azure/autorest.csharp/blob/0ee0276021228fb9a03ed359cd8a621fefd36bb0/src/AutoRest.CSharp/LowLevel/Generation/LowLevelClientWriter.cs#L259-L272) and throughout that file (and many of the files for that project).

src/Cli/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
MiYanni: I've updated this method name. Good call.

Issue #27857, opened 2022-09-13
[APICompat] Document newly added features


ViktorHofer:
TODO:
- [x] Document the msbuild properties that are expected to be set and are passed onto the `ValidatePackage` task.
- [x] Document the `ValidateAssemblies` task and its parameters
- [x] Document the Microsoft.DotNet.ApiCompat.Tool global tool and its parameters
- [x] Document the added rules // cc @smasher164
- [x] Consider grouping the existing PackageValidation documentation under a new APICompat section

ViktorHofer:
The target date for this is October 10th, to make sure that documentation is in place when RC2 releases publicly.

ericstj:
Dropping some notes: https://github.com/dotnet/docs/tree/main/docs/fundamentals/package-validation is a place to put the docs.

carlossanlop:
I opened https://github.com/dotnet/docs/issues/36235 so the Docs Content Team can update the page.

gewarren:
Can you give me any more info that I can use to make these documentation updates - e.g. what are the new MSBuild properties?

For the diagnostic IDs, I'm not seeing any [here](https://github.com/dotnet/sdk/tree/main/src/Compatibility/ApiCompat) that haven't already been [documented](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids).

ViktorHofer:
Sure. All the properties/items that don't start with an underscore are "public" and should be documented:
- https://github.com/dotnet/sdk/blob/5cea8bc6d610c62f57320599954c6eeae52c077e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets#L28-L30
- https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets#L39-L59

We shouldn't document the following ones:
- `RunApiCompat`
- `RuntimeIdentifierGraphPath`

We should mention that individual diagnostic ids can be suppressed via suppression files or via `NoWarn` either per package or globally.

gewarren:
I think this issue can be closed now.

PR #33954, opened 2023-07-12
Align EnableStrictModeForCompatibleTfms default between frontends

### LINKED ISSUE #33011: Should APICompat PackageValidation's EnableStrictModeForCompatibleTfms default to true?
Follow-up on https://github.com/dotnet/sdk/pull/32964#discussion_r1214609786

https://github.com/dotnet/sdk/blob/0991d35e4e52d25816a9656a1f471e2fa7438de7/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets#L48

APICompat PackageValidation's `EnableStrictModeForCompatibleTfms` msbuild property currently defaults to true in the MSBuild frontend. We should discuss if we actually want that.

It was introduced with https://github.com/dotnet/sdk/pull/18714#discussion_r663448216

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/33011

Make sure that the CLI and MSBuild frontend have the same default value for the
EnableStrictModeForCompatibleTfms/--enable-strict-mode-for-compatible-tfms option.

Also bring the description of the parameters in sync again.

### DIFF
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs
@@ -364,12 +364,12 @@ namespace Microsoft.DotNet.ApiCompat.Tool
 
         private static Dictionary<NuGetFramework, IEnumerable<string>>? ParsePackageAssemblyReferenceArgument(ArgumentResult argumentResult)
         {
-            const string invalidPackageAssemblyReferenceFormatMessage = "Invalid package assembly reference format {TargetFrameworkMoniker(+TargetPlatformMoniker)=assembly1,assembly2,assembly3,...}";
+            const string invalidPackageAssemblyReferenceFormatMessage = "Invalid package assembly reference format {TargetFrameworkMoniker(+TargetPlatformMoniker)|assembly1,assembly2,assembly3,...}";
 
             Dictionary<NuGetFramework, IEnumerable<string>> packageAssemblyReferencesDict = new(argumentResult.Tokens.Count);
             foreach (var token in argumentResult.Tokens)
             {
-                string[] parts = token.Value.Split(';');
+                string[] parts = token.Value.Split('|');
                 if (parts.Length != 2)
                 {
                     argumentResult.AddError(invalidPackageAssemblyReferenceFormatMessage);
--- a/src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs
+++ b/src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs
@@ -225,7 +225,8 @@ namespace Microsoft.DotNet.ApiCompat.Tool
             };
             CliOption<bool> enableStrictModeForCompatibleTfmsOption = new("--enable-strict-mode-for-compatible-tfms")
             {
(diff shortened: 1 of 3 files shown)


src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs
```diff
@@ -368,7 +369,7 @@ private static (string CaptureGroupPattern, string ReplacementString)[]? ParseTr
             Dictionary<NuGetFramework, IEnumerable<string>> packageAssemblyReferencesDict = new(argumentResult.Tokens.Count);
             foreach (var token in argumentResult.Tokens)
             {
-                string[] parts = token.Value.Split('=');
+                string[] parts = token.Value.Split(';');
```
ViktorHofer, line 372: That fixes a bug introduced with https://github.com/dotnet/sdk/commit/b8c07fee1b150d46e5c4229cf1f0d81da20393a5. The expected input changed from the TFM simple name to the moniker, i.e. `net7.0` became `.NETCoreApp,Version=v7.0`. We can't use the `=` character as a delimiter here anymore as that's used by the moniker string. The semicolon seems safe to use.

Overall I don't expect many people to use the package validation option in the CLI frontend and even less often to pass in package assembly references. Because of that, I'm not worried about the weird argument syntax.

carlossanlop: The `;` character is normally used to indicate the end of a command and to start concatenating another command in the CLI.

How about using `,` instead?

carlossanlop: Also what's the name of the argument to which the user would pass the list of tokens? This separator should be documented.

ViktorHofer: > This separator should be documented.

We are way behind documentation. We didn't document the previous delimiter either. That's tracked via https://github.com/dotnet/sdk/issues/27857

> The ; character is normally used to indicate the end of a command and to start concatenating another command in the CLI.

👍 

> Also what's the name of the argument to which the user would pass the list of tokens?

`--package-assembly-references`

> How about using , instead?

`,` is an expected value in the `TargetFrameworkMoniker` property: `.NETCoreApp,Version=v7.0`.

carlossanlop: I see.

If you do choose to pick `;` as the separator, then the whole argument needs to be enclosed with double quotes to prevent the CLI to think you're separating commands with it. I think you could test this in PowerShell and confirm if my suspicion is correct.

The docs can wait for now.

Maybe another separator character could be `&`.

carlossanlop (approved):
Signing-off but please double check that `;` does not cause trouble. And if it does, please use another char.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs
ViktorHofer: Changed to `|` which doesn't have a special meaning when surrounded with other characters. Tested it with surrounding quotes and without.

PR #33996, opened 2023-07-13
Add workload version subcommand Fixes #33912

### LINKED ISSUE #33912: Add `dotnet workload --version` command, display workload version other "info" commands
We should add a `dotnet workload --version` command, which will output the workload version (AKA workload set version).  This command will only output the version and nothing else, similar to `dotnet --version`.

Additionally, the `dotnet --info` and `dotnet workload --info` commands should be updated to display the workload version.  From [the design](https://github.com/dotnet/designs/pull/294), this would look like this:

```
> dotnet --info
.NET SDK:
 Version:                 8.0.201
 Commit:                  <commit>
 Workload version:    8.0.201

<further information>

> dotnet workload --info
 Workload version:  8.0.201
 [wasm-tools]
   Installation Source: SDK 8.0.100-preview.4
   Manifest Version:    8.0.0-preview.4.23181.9/8.0.100-preview.4
   Manifest Path:       C:\git\dotnet-sdk-main\.dotnet\sdk-manifests\8.0.100-preview.4\microsoft.net.workload.mono.toolchain.current\WorkloadManifest.json
   Install Type:        FileBased
```

## Implementation

Currently, the `SdkDirectoryWorkloadManifestProvider` class is responsible for selecting the workload version to use.  Probably it (and possibly `IWorkloadManifestProvider`) should add a method that returns the workload version.

When no workload set is installed, a workload version should still be displayed.  Per the design:

> In this case, when a workload set version is needed, the .NET SDK will create a version using the .NET SDK feature band, the pre-release specifier `manifests`, and a hash of the manifest IDs and versions.  For example, `8.0.200-manifests.9e7f4b93`.

Forgind:
Fixes https://github.com/dotnet/sdk/issues/33912

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
@@ -3,6 +3,12 @@
 
 using System.CommandLine;
 using System.CommandLine.Parsing;
+using System.IO;
+using System.Linq;
+using Microsoft.Deployment.DotNet.Releases;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Configurer;
+using Microsoft.DotNet.Workloads.Workload.Install;
 using Microsoft.DotNet.Workloads.Workload.List;
 using Microsoft.NET.Sdk.WorkloadManifestReader;
 using Microsoft.TemplateEngine.Cli.Commands;
@@ -22,22 +28,31 @@ namespace Microsoft.DotNet.Cli
             Description = CommonStrings.WorkloadInfoDescription
         };
 
+        public static readonly CliOption<bool> VersionOption = new("--version")
+        {
+            Description = CommonStrings.WorkloadVersionDescription
+        };
+
         public static CliCommand GetCommand()
         {
             Command.Options.Add(InfoOption);
+            Command.Options.Add(VersionOption);
             return Command;
+        }
+
+        internal static void ShowWorkloadsVersion(ParseResult parseResult)
+        {
+            var workloadManifestProvider =
+                new SdkDirectoryWorkloadManifestProvider(Path.GetDirectoryName(Environment.ProcessPath),
+                Cli.Utils.Product.Version,
+                CliFolderPathCalculator.DotnetUserProfileFolderPath, SdkDirectoryWorkloadManifestProvider.GetGlobalJsonPath(Environment.CurrentDirectory));
+
+            Cli.Utils.Reporter.Output.WriteLine(workloadManifestProvider.GetWorkloadVersion());
         }
 
         internal static void ShowWorkloadsInfo(ParseResult parseResult = null, IWorkloadInfoHelper workloadInfoHelper = null, IReporter reporter = null, string dotnetDir = null)
         {
-            if (workloadInfoHelper != null)
-            {
-                workloadInfoHelper ??= new WorkloadInfoHelper(parseResult != null ? parseResult.HasOption(SharedOptions.InteractiveOption) : false);
-            }
-            else
-            {
-                workloadInfoHelper ??= new WorkloadInfoHelper(false);
-            }
+            workloadInfoHelper ??= new WorkloadInfoHelper(parseResult != null ? parseResult.HasOption(SharedOptions.InteractiveOption) : false);
             IEnumerable<WorkloadId> installedList = workloadInfoHelper.InstalledSdkWorkloadIds;
             InstalledWorkloadsCollection installedWorkloads = workloadInfoHelper.AddInstalledVsWorkloads(installedList);
             reporter ??= Cli.Utils.Reporter.Output;
@@ -82,7 +97,13 @@ namespace Microsoft.DotNet.Cli
             if (parseResult.HasOption(InfoOption) && parseResult.RootSubCommandResult() == "workload")
             {
                 ShowWorkloadsInfo(parseResult);
-                Cli.Utils.Reporter.Output.WriteLine("");
+                Cli.Utils.Reporter.Output.WriteLine(string.Empty);
+                return 0;
+            }
+            else if (parseResult.HasOption(VersionOption) && parseResult.RootSubCommandResult() == "workload")
+            {
+                ShowWorkloadsVersion(parseResult);
+                Cli.Utils.Reporter.Output.WriteLine(string.Empty);
                 return 0;
             }
             return parseResult.HandleMissingCommand();
@@ -104,7 +125,7 @@ namespace Microsoft.DotNet.Cli
 
             command.Validators.Add(commandResult =>
             {
-                if (commandResult.GetResult(InfoOption) is null && !commandResult.Children.Any(child => child is CommandResult))
+                if (commandResult.GetResult(InfoOption) is null && commandResult.GetResult(VersionOption) is null && !commandResult.Children.Any(child => child is System.CommandLine.Parsing.CommandResult))
                 {
                     commandResult.AddError(Tools.CommonLocalizableStrings.RequiredCommandNotPassed);
                 }
--- a/src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
@@ -40,17 +40,17 @@ namespace Microsoft.DotNet.Cli
             return Command;
         }
 
-        internal static void ShowWorkloadsVersion(ParseResult parseResult)
+        internal static string GetWorkloadsVersion()
         {
             var workloadManifestProvider =
                 new SdkDirectoryWorkloadManifestProvider(Path.GetDirectoryName(Environment.ProcessPath),
                 Cli.Utils.Product.Version,
                 CliFolderPathCalculator.DotnetUserProfileFolderPath, SdkDirectoryWorkloadManifestProvider.GetGlobalJsonPath(Environment.CurrentDirectory));
 
-            Cli.Utils.Reporter.Output.WriteLine(workloadManifestProvider.GetWorkloadVersion());
+            return workloadManifestProvider.GetWorkloadVersion();
         }
 
-        internal static void ShowWorkloadsInfo(ParseResult parseResult = null, IWorkloadInfoHelper workloadInfoHelper = null, IReporter reporter = null, string dotnetDir = null)
+        internal static void ShowWorkloadsInfo(ParseResult parseResult = null, WorkloadInfoHelper workloadInfoHelper = null, IReporter reporter = null, string dotnetDir = null)
         {
             workloadInfoHelper ??= new WorkloadInfoHelper(parseResult != null ? parseResult.HasOption(SharedOptions.InteractiveOption) : false);
             IEnumerable<WorkloadId> installedList = workloadInfoHelper.InstalledSdkWorkloadIds;
@@ -74,7 +74,7 @@ namespace Microsoft.DotNet.Cli
                 const int align = 10;
                 const string separator = "   ";
 
-                reporter.WriteLine($" {'[' + workload.Key + ']'}");
+                reporter.WriteLine($" [{workload.Key}]");
 
                 reporter.Write($"{separator}{CommonStrings.WorkloadSourceColumn}:");
                 reporter.WriteLine($" {workload.Value,align}");
@@ -96,14 +96,24 @@ namespace Microsoft.DotNet.Cli
         {
             if (parseResult.HasOption(InfoOption) && parseResult.RootSubCommandResult() == "workload")
             {
-                ShowWorkloadsInfo(parseResult);
-                Cli.Utils.Reporter.Output.WriteLine(string.Empty);
+                WorkloadInfoHelper workloadInfoHelper = new(parseResult.HasOption(SharedOptions.InteractiveOption));
+                if (!workloadInfoHelper.InstalledSdkWorkloadIds.Any())
+                {
+                    Reporter.Output.WriteLine(CommonStrings.NoWorkloadsInstalledInfoWarning);
+                    Reporter.Output.WriteLine(string.Empty);
+                    return 0;
+                }
+
+                Reporter.Output.WriteLine($" Workload version: {workloadInfoHelper.ManifestProvider.GetWorkloadVersion()}");
+
+                ShowWorkloadsInfo(parseResult, workloadInfoHelper: workloadInfoHelper);
+                Reporter.Output.WriteLine(string.Empty);
                 return 0;
             }
             else if (parseResult.HasOption(VersionOption) && parseResult.RootSubCommandResult() == "workload")
             {
-                ShowWorkloadsVersion(parseResult);
-                Cli.Utils.Reporter.Output.WriteLine(string.Empty);
+                Reporter.Output.WriteLine(GetWorkloadsVersion());
+                Reporter.Output.WriteLine(string.Empty);
                 return 0;
             }
             return parseResult.HandleMissingCommand();
--- a/src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
@@ -40,14 +40,11 @@ namespace Microsoft.DotNet.Cli
             return Command;
         }
 
-        internal static string GetWorkloadsVersion()
+        internal static string GetWorkloadsVersion(WorkloadInfoHelper workloadInfoHelper = null)
         {
-            var workloadManifestProvider =
-                new SdkDirectoryWorkloadManifestProvider(Path.GetDirectoryName(Environment.ProcessPath),
-                Cli.Utils.Product.Version,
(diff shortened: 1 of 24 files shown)


src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -150,7 +175,7 @@ void ProbeDirectory(string manifestDirectory, string featureBand)
                 (string? id, string? finalManifestDirectory) = ResolveManifestDirectory(manifestDirectory);
                 if (id != null && finalManifestDirectory != null)
                 {
-                    AddManifest(id, finalManifestDirectory, featureBand);
+                    AddManifest(id, finalManifestDirectory, featureBand, Path.GetFileName(manifestDirectory));
```
dsplaisted, line 178: If the workload manifest is not side-by-side, then this won't pass the correct manifest version.  In that case you would need to read the actual manifest file to get the version.

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandParser.cs
```diff
@@ -86,8 +98,24 @@ private static int ProcessArgs(ParseResult parseResult)
...
-                Cli.Utils.Reporter.Output.WriteLine("");
+                WorkloadInfoHelper workloadInfoHelper = new(parseResult.HasOption(SharedOptions.InteractiveOption));
+                if (!workloadInfoHelper.InstalledSdkWorkloadIds.Any())
+                {
+                    Reporter.Output.WriteLine(CommonStrings.NoWorkloadsInstalledInfoWarning);
+                    Reporter.Output.WriteLine(string.Empty);
+                    return 0;
+                }
+
+                Reporter.Output.WriteLine($" Workload version: {workloadInfoHelper.ManifestProvider.GetWorkloadVersion()}");
```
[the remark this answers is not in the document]
Forgind: I did it this way specifically because I thought we shouldn't display a workload version if no workloads are installed 🙂 I can undo that.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -273,13 +297,13 @@ void ProbeDirectory(string manifestDirectory, string featureBand)
             //  Assume that if there are any versioned subfolders, they are higher manifest versions than a workload manifest directly in the specified folder, if it exists
             if (manifestVersionDirectories.Any())
             {
-                return (manifestId, manifestVersionDirectories.First().directory);
+                return (manifestId, manifestVersionDirectories.First().directory, manifestVersionDirectories.First().version);
             }
             else if (File.Exists(Path.Combine(manifestDirectory, "WorkloadManifest.json")))
             {
-                return (manifestId, manifestDirectory);
+                return (manifestId, manifestDirectory, null);
```
dsplaisted, line 304: I think returning null for the version here will cause problems when we try to compute the hash for the workload set version.  So we probably need to actually read the manifest to get the version.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -273,13 +297,15 @@ void ProbeDirectory(string manifestDirectory, string featureBand)
             //  Assume that if there are any versioned subfolders, they are higher manifest versions than a workload manifest directly in the specified folder, if it exists
             if (manifestVersionDirectories.Any())
             {
-                return (manifestId, manifestVersionDirectories.First().directory);
+                return (manifestId, manifestVersionDirectories.First().directory, manifestVersionDirectories.First().version);
             }
             else if (File.Exists(Path.Combine(manifestDirectory, "WorkloadManifest.json")))
             {
-                return (manifestId, manifestDirectory);
+                var manifestPath = Path.Combine(manifestVersionDirectories.First().directory, "WorkloadManifest.json");
```
Forgind, line 304:
```suggestion
                var manifestPath = Path.Combine(manifestDirectory, "WorkloadManifest.json");
```

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
Forgind: > I think returning null for the version here will cause problems when we try to compute the hash for the workload set version. So we probably need to actually read the manifest to get the version.

This turned out to be a disaster when it came to tests. Tests are making invalid WorkloadManifest.json files, and when I try to parse them for a version, it crashes.

So the obvious good solution to that is to fix the tests, right? I should change the tests so they write valid WorkloadManifest.json files.

Except there are several problems. First, that's a horrible mess just from a "format a string in C#" perspective. That's made worse by the fact that this is in a number of tests, so I can't correct it in one place and then have everything else build off of that—I need to correct both the place that creates the WorkloadManifest.json file and the place that assumes it's present and correct. Second, the previous format assumed certain parameters would be present in the WorkloadManifest.json file—and that's no longer true. The format has changed, which means I can't include the same information that was previously there. Third, it occurred to me that that problem might be true going forward as well; that is, future developers may decide to change the WorkloadManifest.json format, in which case they'd have to change all these tests all over again.

I wasted several hours on this today. My eventual resolution was to throw up my hands and just put in a try/catch statement instead. I think that should work and should make the tests happy. It's not exactly ideal, since I think we should notice if there's a problem and throw an error, but I think it's a fairly good solution as far as solutions go, and much easier from an implementation perspective than what I was trying to do.

Let me know if you disagree or see a better path forward.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -121,12 +122,34 @@ internal SdkDirectoryWorkloadManifestProvider(string sdkRootPath, string sdkVers
...
+                return _workloadSet?.Version!;
+            }
+
+            using (SHA256 sha256Hash = SHA256.Create())
+            {
+                byte[] bytes = sha256Hash.ComputeHash(Encoding.UTF8.GetBytes(string.Join(";",
+                            GetManifests().OrderBy(m => m.ManifestId).Select(m => $"{m.ManifestId}.{m.ManifestFeatureBand}.{m.ManifestVersion}").ToArray()
+                        )));
+                StringBuilder sb = new();
+                for (int b = 0; b < 4 && b < bytes.Length; b++)
```
MiYanni, line 138: What does the `< 4` indicate here?

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
```diff
@@ -273,13 +297,22 @@ void ProbeDirectory(string manifestDirectory, string featureBand)
...
+                var manifestPath = Path.Combine(manifestDirectory, "WorkloadManifest.json");
+                try
+                {
+                    var manifestContents = WorkloadManifestReader.ReadWorkloadManifest(manifestId, File.OpenRead(manifestPath), manifestPath);
+                    return (manifestId, manifestDirectory, new ReleaseVersion(manifestContents.Version));
+                }
+                catch
+                {
+                    return (manifestId, manifestDirectory, null);
+                }
```
MiYanni, line 313: nit: Makes it more obvious that the `catch` doesn't do anything.
```suggestion
                ReleaseVersion releaseVersion = null;
                try
                {
                    var manifestContents = WorkloadManifestReader.ReadWorkloadManifest(manifestId, File.OpenRead(manifestPath), manifestPath);
                    releaseVersion = new ReleaseVersion(manifestContents.Version);
                }
                catch { }

                return (manifestId, manifestDirectory, releaseVersion);
```

MiYanni (approved):

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
Forgind: I should probably add a comment.

This is intended to be part of a version, and we do want it to be unique, hence the hash, but we don't want the version to be too long, so this truncates it.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
Forgind: This is an interesting suggestion...on the one hand, it's longer and less attractive to me. On the other hand, you're right that it makes it more clear that the catch block doesn't do anything. I think clarity is more important, and it makes it easy for me to add a comment in the catch block to explain this.

Thanks for the suggestion!

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
MiYanni: Wait, so if I'm reading this correctly, you take all the manifests, create strings of the version numbers via ManifestId, ManifestFeatureBand, and ManifestVersion. You join these strings via semi-colon, encode them as bytes, and compute a hash. Then, you truncate the hash of 256bits (64 bytes) down to only 4 bytes. It seems like a lot of work done to calculate this unique hash to then simply make it not unique by truncating it. Am I making any sense? 😅

Forgind: Yep! And that's a great question!

Before getting to any math, to reiterate our primary goals, we want the version to be short, almost certainly unique, and communicate information about what you have installed.

Hashes work great for precisely indicating what you have installed while being short. The version is included so you have some sense of what's going on. Hashes are great (in theory, at least) since they choose a uniformly random number that's always the same if you pass in the same input.

In particular, the first bits of the has is also uniformly random under that constraint. There's always some possibility of a hash collision, and this increases that possibility. It's very, very low if we were to take the same hash. Truncating it to just the first four bytes, the chance of any two versions colliding is 2^-32, which is almost negligible still. From the birthday paradox, you can infer that the chance that some pair will collide is much greater than you'd think, but since it also incorporates the major/minor version in plain text, there aren't that many versions that we're likely to be comparing that might collide. As an example, if there are 1000 versions that might collide, even with this truncated version, the probability that no pair will collide is:
https://www.wolframalpha.com/input?i=%28%28%282%5E32%29%21%29%2F%28%28%282%5E32%29-1000%29%21%29%29+%2F+%28%282%5E32%29%5E1000%29

So I can almost guarantee they still won't collide.

On the other hand, if we were to change the 4 to a 3, it would be:
https://www.wolframalpha.com/input?i=%28%28%282%5E24%29%21%29%2F%28%28%282%5E24-1000%29%21%29%29+%2F+%28%282%5E24%29%5E1000%29

That's still unlikely, but I'd feel a lot less confident saying they _won't_ collide.

Does that help?

MiYanni: From my past experience with Git commit hashes (which may or may not be applicable here), the standard for 'short' hashes is 7 characters. Even then, I'd have conflicts with the same short commit hash applying to more-than-one commit. I did some quick research, and found a hashing process that fits your intended goals. It is used to create similar IDs to what YouTube uses for videos. It is called Sqids: https://sqids.org/dotnet

If you don't want to do something like that, I'm fine with the solution here. Just pointing out other options and experiences.

dsplaisted: Just to clarify, this is taking 4 bytes, which ends up as 8 hexadecimal characters.  So the hash it's generating is 8 characters long.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
dsplaisted: Can you add a comment explaining that the catch is there for tests that don't write valid manifests?

I'm not sure what you mean about the manifest format changing.  I don't believe it has changed significantly since it shipped.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -31,6 +31,48 @@ void Initialize(string featureBand = "5.0.100", [CallerMemberName] string? testN
...
+            {
+                string[] manifests = sdkDirectoryWorkloadManifestProvider.GetManifests().OrderBy(m => m.ManifestId).Select(m => $"{m.ManifestId}.{m.ManifestFeatureBand}.{m.ManifestVersion}").ToArray();
+                manifests.Length.Should().Be(1);
+                manifests.Should().Contain("android.8.0.200.33.0.2-rc.1");
+                sdkDirectoryWorkloadManifestProvider.GetWorkloadVersion().Should().Be("8.0.200-manifests.4ba11739");
+            }
+
+            Directory.Delete(Path.Combine(_manifestRoot, "8.0.100"), recursive: true);
+            Directory.Delete(Path.Combine(_manifestRoot, "8.0.200"), recursive: true);
+            Directory.Delete(Path.Combine(_manifestRoot, "8.0.200-rc.2"), recursive: true);
```
dsplaisted, line 73: It shouldn't be necessary to delete anything here, each test should use a test-specific directory.

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -31,6 +31,48 @@ void Initialize(string featureBand = "5.0.100", [CallerMemberName] string? testN
             Directory.CreateDirectory(_manifestVersionBandDirectory);
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ItShouldReturnTheWorkloadVersion(bool useWorkloadSet)
+        {
+            Initialize();
```
dsplaisted, line 39:
```suggestion
            Initialize(identifier: useWorkloadSet.ToString());
```

src/Tests/Microsoft.NET.Sdk.WorkloadManifestReader.Tests/SdkDirectoryWorkloadManifestProviderTests.cs
```diff
@@ -31,6 +31,48 @@ void Initialize(string featureBand = "5.0.100", [CallerMemberName] string? testN
...
+
+            CreateMockManifest(_manifestRoot, "8.0.100", "ios", "11.0.2", true);
+            CreateMockManifest(_manifestRoot, "8.0.200", "android", "33.0.2-rc.1", true);
+            CreateMockManifest(_manifestRoot, "8.0.200-rc.2", "maui", "15.0.1-rc.456", true);
+
+            if (useWorkloadSet)
+            {
+                CreateMockWorkloadSet(_manifestRoot, "8.0.200", "8.0.200", @"
```
dsplaisted, line 47: I think we should change the workload set version to something that is different than the feature band version (such as 8.0.202), to help verify the right version is being returned.

dsplaisted (approved):
This can go into 8.0.200.  Since we don't have workload sets in 8.0.100, I don't think there's a whole lot of value to this until we add them.

src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs
MiYanni: > Just to clarify, this is taking 4 bytes, which ends up as 8 hexadecimal characters. So the hash it's generating is 8 characters long.

Right. When I said '7 characters' for commit hashes, I meant 7 unique values. I believe each is in the alphanumeric range, so 26 x 2 (lower and upper case alphas) + 10 (zero to nine), which is 62. If 1 byte is represented as 2 character, that means 2 characters are holding 256 unique values (2^8 is a byte). But if you have 2 alphanumeric characters, that would be 62^2 which is 3844 unique combinations. This is **not** my area of expertise, but as long as the solution is vetted and makes sense for this scenario, I'm fine with it. Again, this conversation was because the fact that `< 4` seemed suspicious to me. 😋

PR #34063, opened 2023-07-18
Fix error code NETSDK1129

v-wuzhai:


MiYanni:
The change for this string was merged into `main` here: https://github.com/dotnet/sdk/pull/32333

@baronfel Is that string change needed for the other branches?

baronfel:
I'm confused, why are we reverting the new message in this PR?

marcpopMSFT:
It's ok to have different values for error messages in main than other branches as long the intended purpose of the error is the same if the code is the same. It potentially causes issues in our documentation as we generally don't list all possible text for a particular code. I think Jason was trying to get the messages to align rather than just the code so that the documentation would align as well but I think we just make the documents match whatever the latest text is:
https://github.com/dotnet/sdk/pull/34059#issuecomment-1639813365

I think this can be closed and I think the merge to main looks good to me.

PR #34069, opened 2023-07-18
Revert "[main] Update dependencies from nuget/nuget.client (#33913)"

sbomer:
This reverts commit 1690ca1121fd3a18441b48849c7ef0ec19e7eb43.
This should work around a nuget regression https://github.com/NuGet/Home/issues/12757, which is blocking AOT scenarios as in https://github.com/dotnet/runtime/issues/89040.


RussKie:
Is this ok to merge?

eerhardt:
@sbomer - have you verified that this revert will fix the issue?

What happens when we need the next NuGet upgrade? Won't the issue show up again?

eerhardt (approved):

PR #34072, opened 2023-07-18
Add EnableGeneratedComInterfaceComImportInterop property to control the source-generated/built-in COM interop interaction feature switch.

jkoritzinsky:
This change provides a nice property to control the opt-in feature implemented in https://github.com/dotnet/runtime/pull/87583.

We will use this property in the source-generated COM interop analyzers, so we will also make it CompilerVisible.


AaronRobinsonMSFT (approved):

sbomer (approved):
Does the feature need to be disabled by default when trimming?

PR #34077, opened 2023-07-18
Add separate warnings for settings that depend on illink pack

### LINKED ISSUE #3175: Define UX for `IsTrimmable` in `netstandard2.1` libraries
Context: https://github.com/dotnet/aspnetcore/pull/45879#issuecomment-1372517140

The change to use a TFM-versioned linker (https://github.com/dotnet/sdk/pull/29441) means that the illink targets aren't available for `netcoreapp2.1` (or any TFM earlier than 3.0 when trimming was introduced). Those targets include some shared logic (analyzer support, and adding an assembly-level attribute for `IsTrimmable`).

This is the configuration we shipped in 7.0:
When building a `netstandard2.1` app, `IsTrimmable` turns on the assembly-level attribute. This doesn't enable the analyzer by default, but it can be turned on by `EnableTrimAnalyzer`. However, the framework ref assemblies aren't annotated in 2.1, so warnings will be incomplete, and library authors will need to compile in the analysis attributes if they want to annotate their own libraries (like we did in dotnet/runtime).

Marking existing APIs as trim-incompatible in a new TFM is effectively introducing a breaking change for trimmed apps/libraries.  Our recommendation so far to address this is to multi-target libraries to get the latest analysis warnings. We probably need to make this more explicit in the docs. Even so, it seems easy to fall into this trap by just setting `IsTrimmable` in a netstandard library that's single-targeted.

The question is: what is the intended UX for this scenario?

sbomer:
This splits the error added in https://github.com/dotnet/sdk/pull/33041 into separate warnings/errors specific to the various options that depend on the ILLink pack. These warnings/errors are produced when those options are set for an unsupported target framework.

The PublishTrimmed error text reuses the message that used to be produced as a warning when targeting < netcoreapp3.0. This error is produced only when PublishAot is not set (there's a more specific error if it is).

Previously, we would produce errors when attempting to use the analyzers on unsupported TFMs. This has been changed to a warning in the case of a library (for example IsTrimmable, rather than PublishTrimmed). See comments in ProcessFrameworkReferences for details.

This also fixes an issue where ProcessFrameworkReferences was not running for netcoreapp2.0 because there were no FrameworkReferences defined for that TFM. This adds an extra condition to let ProcessFrameworkReferences run if any settings depend on the ILLink pack, so that it can produce proper warnings/errors.

Fixes https://github.com/dotnet/linker/issues/3175


src/Tasks/Common/Resources/Strings.resx
```diff
@@ -937,5 +939,14 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1209: The current Visual Studio version does not support targeting {0} {1}.  Either target {0} {2} or lower, or use Visual Studio version {3} or higher</value>
     <comment>{StrBegin="NETSDK1209: "}</comment>
   </data>
-  <!-- The latest message added is UnsupportedVisualStudioVersion. Please update this value with each PR to catch parallel PRs both adding a new message -->
+  <data name="IsAotCompatibleUnsupported" xml:space="preserve">
+    <value>NETSDK1210: IsAotCompatible is not supported for the target framework. Consider multi-targeting to a supported framework to enable ahead-of-time compilation analysis, and set IsAotCompatible only for the supported frameworks. For example:
+&lt;IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))"&gt;true&lt;/IsAotCompatible&gt;</value>
```
vitek-karas, line 944: AOT is really only supported from 7 and above - so the example should use `net7.0`.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -937,5 +939,14 @@ You may need to build the project on another operating system or architecture, o
...
+    <comment>{StrBegin="NETSDK1210: "}</comment>
+  </data>
+  <data name="EnableSingleFileAnalyzerUnsupported" xml:space="preserve">
+    <value>NETSDK1211: EnableSingleFileAnalyzer is not supported for the target framework.</value>
```
vitek-karas, line 948: Maybe also add the example with the conditional syntax - since we don't have an `IsSingleFileCompatible` the `EnableSingleFileAnalyzer` effectively acts like it - so the guidance should now apply to it.

vitek-karas (approved):
Looks good to me - just minor comments

PR #34093, opened 2023-07-19
Enable MSBuild TerminalLogger by default

rainersigwald:
Opt all MSBuild invocations into TerminalLogger in auto mode. This should not affect any CI invocations, because auto
will fall back to the console logger when it detects redirected output.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
@@ -45,7 +45,7 @@ namespace Microsoft.DotNet.Cli.Utils
             };
 
         private readonly IEnumerable<string> _msbuildRequiredParameters =
-            new List<string> { "-maxcpucount", "-verbosity:m" };
+            new List<string> { "-maxcpucount", "-verbosity:m", "-terminallogger:auto" };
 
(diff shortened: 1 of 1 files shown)


rainersigwald:
Is this the right place to put this enablement? Do we need to have a disable option beyond "use `-tl:false` on the command line"?

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -45,7 +45,7 @@ internal class MSBuildForwardingAppWithoutLogging
             };
 
         private readonly IEnumerable<string> _msbuildRequiredParameters =
-            new List<string> { "-maxcpucount", "-verbosity:m" };
+            new List<string> { "-maxcpucount", "-verbosity:m", "-terminallogger:auto" };
```
baronfel, line 48: if the environment variable is used to disable TL, does that take precedence? what about if a user specifies `-tl:off` manually - what's the precedence for required args here vs user-supplied arguments here vs env vars?

rainersigwald: The environment variable is lower precedence than the command line, so it would be completely ignored if we put this change in as-is today.

marcpopMSFT:
superceded by https://github.com/dotnet/sdk/pull/34606

PR #34132, opened 2023-07-20
Revert "#30336 Default Dotnet Tool to Run on Latest Runtime - Build time change"

marcpopMSFT:
Reverts dotnet/sdk#31957

Per offline discussion, there was some discomfort with the original change as it might imply a change to our compat bar that was not intended. We're going to discuss alternative solutions. 

Options:
1. Warn the customer at install and tell them to install the runtime
2. Add a --roll-forward flag during install that allows tool consumers to make the decision for themselves
3. Other?

We'll continue to discuss options in https://github.com/dotnet/sdk/issues/30336. We should make sure we remove the breaking change notice as well.


JL03-Yue (approved):

JL03-Yue:
Approved, following up with breaking change documentation reversion

Issue #598, opened 2017-01-06
Copy Package Reference assemblies to output directory for .NET Standard libraries


Snipx:
Hi,
I am using VS2017RC to create a .NET Standard library.
When I do that and build the binaries, the package reference assemblies (System.Linq.dll, System.Net.dll etc etc) are not copied to the output path.
I can't seem to find any configurations for whether or not to copy the dependencies to output neither in VS2017RC nor via the .csproj file directly.

Is there any way to achieve copying of all the package references to output?
I specifically need this to run unit tests via NUnit.

I have [asked](https://github.com/Microsoft/msbuild/issues/1515) this question to MSBuild team, but they redirected me to you. They also mentioned a workaround of creating a dummy executable so that the dependencies are copied, but I wonder if there is a more elegant solution to this.

eerhardt:
@Snipx - have you tried using the new `dotnet test` command to run your unit tests?  That verb doesn't need the assemblies in the output directory, as they are loaded from the NuGet cache, just like during `dotnet run`.  I would recommend trying this out, and seeing how it works for you.

Snipx:
I have tried `dotnet test`, no success so far, the tests are not executed I'm getting `Starting test execution, please wait...` message and that's all. I have submitted a request to the NUnit team, but I'm not sure the problem is on their side.

Meanwhile, @eerhardt, do you happen to have an idea on what might be wrong?

I have described my problem [here](https://github.com/nunit/dotnet-test-nunit/issues/98)

Snipx:
Closing the issue. The NUnit team explained that `dotnet test` doesn't currently work for .NET Core after moving to `.csproj` back from json.

PR #34143, opened 2023-07-20
Fix package verification for some valid packages on windows

### LINKED ISSUE #33928: Certificate trust list needs to be backported to 6.0.3xx (and maybe 6.0.1xx)
### Describe the bug

.NET SDK 6.0.4xx introduced a new embedded certificate trust list. Older SDK versions (including LTS 6.0.3xx and 6.0.1xx) rely primarily on the certificate trust list provided by the operating system (which of course varies from OS to OS).

There are two bugs related to this functionality:

1. The new certificate trust list includes the Microsoft Time Stamping service certificate trust, which does not exist as a trust in any OS. Therefore, attempting to verify a signed package using this certificate fails on all operating systems unless the user upgrades to a new enough 6.0.4xx SDK. This affects new packages signed by the new .NET Foundation code signature process, since it uses the Microsoft Time Stamping service. (Users will experience this as [NU3003](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3003))
2. The certificate used for repository countersignature in nuget.org has shifted from Symantec to DigiCert, related to the [feature discussion for the new certificate trust list](https://github.com/dotnet/designs/blob/main/accepted/2021/signed-package-verification/re-enable-signed-package-verification.md#high-level-approach). Unfortunately, some OSes (like Ubuntu 22.04) have already revoked their trust for the Symantec root certificate. This affects older packages countersigned by NuGet before the move to DigiCert. (Users will probably experience this as [NU3028](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028))

In my opinion, all LTS versions need this functionality, not just 6.0.4xx.

#### Microsoft Time Stamping service

1. Download xunit package version 2.5.0: https://www.nuget.org/api/v2/package/xunit/2.5.0
2. Run `dotnet nuget verify --all xunit.2.5.0.nupkg`

Here are the results with 6.0.314 (latest as of this writing):

And here are the results with 6.0.411 (latest as of this writing):

Example images are shown from Linux (as it was the easiest way to quickly move between versions), but this is also reproducible with Windows.

#### Symantec Time Stamping service

On a non-Windows machine:

1. Download xunit package version 2.4.2: https://www.nuget.org/api/v2/package/xunit/2.4.2
2. Run `dotnet nuget verify --all -v normal xunit.2.4.2.nupkg`

Here are the results with 6.0.314 on Linux:

And here are the results with 6.0.411 on Linux:

image

marcpopMSFT:
Partially fixes https://github.com/dotnet/sdk/issues/33928

## Description
Parsing failure: The Microsoft timestamping service does something unusual but within standards. There were a couple of bugs in System.Security.Cryptography.Pkcs where .NET failed to parse these attribute certificates, but those bugs were fixed, and the fixes were backported to the 6.0 runtime. However, some .NET 6.0 SDK feature bands didn't pick up the newer version of System.Security.Cryptography.Pkcs when executing msbuild or nuget.  

Currently in 3xx and 1xx, nuget builds against the 5.0.0 verison if this package and msbuild against 4.7.0 which means the 5.0.0 version gets loaded. In 6.0.4xx, msbuild moved forward to the 6.0.1 version if this package

## Customer Impact
On 6.0.3xx and 6.0.1xx, if you add certain packages to your project that are correctly signed, you'll get `error: NU3003: The package signature is invalid or cannot be verified on this platform.`  You'll also get this error if you run dotnet nuget verify against the package.

An example that shows this is Xunit version 2.5.0 and could impact any package signed with the .NET Foundation signing service after a recent change.

The change to the signing service merged on February 27, 2023 ([Make ACS the default timestamping service by dtivel · Pull Request #598 · dotnet/sign (github.com)](https://github.com/dotnet/sign/pull/598)).  The change which led to the parsing error was [simply defaulting to the Microsoft timestamping service](https://github.com/dotnet/sign/pull/598/files#diff-5b2caeaf35c92a62c3a106e3177262f94ac9eb82a701c5983fd296a0d75db8eaR37).  If a Sign CLI user used a different timestamping service, there probably wouldn't be parsing errors.

## Regression?
* [ ] Yes
* [x] No

## Risk
* [ ] High
* [ ] Medium
* [x] Low

## Verification

* [x] manual
* [ ] automated tests

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -25,7 +25,7 @@
     <NewtonsoftJsonPackageVersion>13.0.1</NewtonsoftJsonPackageVersion>
     <NewtonsoftJsonVersion>$(NewtonsoftJsonPackageVersion)</NewtonsoftJsonVersion>
     <SystemDiagnosticsFileVersionInfoVersion>4.0.0</SystemDiagnosticsFileVersionInfoVersion>
-    <SystemReflectionMetadataVersion>6.0.0</SystemReflectionMetadataVersion>
+    <SystemReflectionMetadataVersion>5.0.0</SystemReflectionMetadataVersion>
     <SystemSecurityCryptographyPkcsPackageVersion>6.0.4</SystemSecurityCryptographyPkcsPackageVersion>
     <MicrosoftDotNetSignToolVersion>6.0.0-beta.23361.3</MicrosoftDotNetSignToolVersion>
     <MicrosoftWebXdtPackageVersion>3.1.0</MicrosoftWebXdtPackageVersion>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -25,7 +25,8 @@
     <NewtonsoftJsonPackageVersion>13.0.1</NewtonsoftJsonPackageVersion>
     <NewtonsoftJsonVersion>$(NewtonsoftJsonPackageVersion)</NewtonsoftJsonVersion>
     <SystemDiagnosticsFileVersionInfoVersion>4.0.0</SystemDiagnosticsFileVersionInfoVersion>
-    <SystemReflectionMetadataVersion>5.0.0</SystemReflectionMetadataVersion>
+    <SystemReflectionMetadataVersion>6.0.0</SystemReflectionMetadataVersion>
+    <SystemSecurityCryptographyPkcsPackageVersion>6.0.4</SystemSecurityCryptographyPkcsPackageVersion>
     <MicrosoftDotNetSignToolVersion>6.0.0-beta.23361.3</MicrosoftDotNetSignToolVersion>
     <MicrosoftWebXdtPackageVersion>3.1.0</MicrosoftWebXdtPackageVersion>
     <SystemSecurityCryptographyProtectedDataPackageVersion>6.0.0</SystemSecurityCryptographyProtectedDataPackageVersion>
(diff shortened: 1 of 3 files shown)


dtivel (approved):

bradwilson:
I disagree that this _**fully**_ fixes #33928. It will _**partially**_ fix it (on Windows only).

eng/Versions.props
```diff
@@ -25,7 +25,8 @@
     <NewtonsoftJsonPackageVersion>13.0.1</NewtonsoftJsonPackageVersion>
     <NewtonsoftJsonVersion>$(NewtonsoftJsonPackageVersion)</NewtonsoftJsonVersion>
     <SystemDiagnosticsFileVersionInfoVersion>4.0.0</SystemDiagnosticsFileVersionInfoVersion>
-    <SystemReflectionMetadataVersion>5.0.0</SystemReflectionMetadataVersion>
+    <SystemReflectionMetadataVersion>6.0.0</SystemReflectionMetadataVersion>
```
dtivel, line 28: @marcpopMSFT, is this bump necessary?

marcpopMSFT: CC @rainersigwald who included that in the change. I assume there is a dependency between the two that required udpating both.

rainersigwald: My initial commit https://github.com/dotnet/sdk/commit/2ecd6a7fea7d33adb0f9da2901351d8823dade94 didn't have it, I think this must have come in a cherry-pick merge conflict resolution at some point.

marcpopMSFT: weird, I don't recall seeing merge conflicts. Did I have local uncommitted changes maybe that I forgot about?

PR #34154, opened 2023-07-21
Parse User and Port Container Metadata from special environment variables

### LINKED ISSUE #478: Create a ContainerPort item for the ASPNETCORE_URLs environment variables if set
ASP.NET has a few environment variables that control which ports/urls are used for hosting the application

* ASPNETCORE_URLS
* ASPNETCORE_HTTP_PORTS
* ASPNETCORE_HTTPS_PORTS

We should check for these in the environment block of an application, and attempt to define `ExposedPort` metadata for any ports found there. `ASPNETCORE_HTTP_PORTS` and `ASPNETCORE_HTTPS_PORTS` have a simple syntax of semicolon-delimited port numbers, but `ASPNETCORE_URLS` has a format of semicolon-delimited `{scheme}://{address}:{port}` items, where `address` is one of:
* `localhost`
* an IP address
* `*`

Creating these ports allows other tooling like the VSCode Docker tooling to automatically detect and expose ports when users run our containers.

### LINKED ISSUE #479: .NET 8 preview 6 containers do not run as app user
I have an ASP.NET Core application targeting .NET 8 with a container image built using the `DefaultContainer` profile.

Since updating to preview 6 of the SDK I've noticed that the application no longer runs as the app user. I initially thought that it was an [issue with the base image](https://github.com/dotnet/dotnet-docker/issues/4766) but I've found that I can only replicate it when building the container image using the SDK publish profile.

I think it's related to [this change](https://github.com/dotnet/dotnet-docker/pull/4715) to the app user UID in the base image as there is still an [open issue here](https://github.com/dotnet/sdk-container-builds/issues/395) about the UID being hard-coded in places instead of using the `APP_UID` variable.

The workaround is to explicity run the container as UID 1654.

baronfel:
Fixes https://github.com/dotnet/sdk-container-builds/issues/478
Fixes https://github.com/dotnet/sdk-container-builds/issues/479

User Id and Ports can be set explicitly (through user-specified MSBuild properties), but also implicitly via container environment variables. These can even be set by the base images that users use.

When we encounter these environment variables, we should apply them to the container configuration.


src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -112,8 +112,6 @@
     <!-- We only set a default user when the base image is Microsoft-authored, and we're targeting a version of those images that supports a nonroot user -->
     <PropertyGroup Label="ContainerUser Assignment" Condition="$(_ContainerIsUsingMicrosoftDefaultImages) and $(_ContainerIsTargetingNet8TFM) and '$(ContainerUser)' == ''">
       <ContainerUser Condition="$(_ContainerIsTargetingWindows)">ContainerUser</ContainerUser>
-      <!-- For Linux, the MS images have an 'app' user set to user id 64198 but some container runtimes work better with user ids instead of names, so we defer to that -->
```
baronfel, line 115: note that the linux images have an app user set, so we no longer need to hardcode this.

baronfel (commented):
A few notes about the implementation

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -113,10 +113,15 @@ public static class ContainerBuilder
...
             imageBuilder.ExposePort(number, type);
         }
+
         if (containerUser is { Length: > 0 } user)
         {
             imageBuilder.SetUser(user);
         }
+
+        ContainerHelpers.AssignUserFromEnvironment(imageBuilder, logger);
+        ContainerHelpers.AssignPortsFromEnvironment(imageBuilder, logger);
```
vlada-shubina, line 123: as discussed on call, consider moving those to `ImageBuilder` to avoid code (and therefore test duplication) in `ContainerBuilder` and `CreateNewImage`.

src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -281,4 +282,86 @@ internal static bool IsValidEnvironmentVariable(string envVar)
...
+    /// <summary>
+    /// ASP.NET can have urls/ports set via three environment variables - if we see any of them we should create ExposedPorts for them
+    /// to ensure tooling can automatically create port mappings.
+    /// </summary>
+    internal static void AssignPortsFromEnvironment(ImageBuilder imageBuilder, ILogger logger)
+    {
+        // asp.net images control port bindings via three environment variables. we should check for those variables and ensure that ports are created for them
+        if (imageBuilder.EnvironmentVariables.TryGetValue(KnownStrings.EnvironmentVariables.ASPNETCORE_HTTP_PORTS, out string? httpPorts))
+        {
+            logger.LogTrace("Setting ports from ASPNETCORE_HTTP_PORTS environment variable");
```
vlada-shubina, line 321: as discussed, it would be good to document this logic and env var used in the reference.

src/Containers/Microsoft.NET.Build.Containers/ImageBuilder.cs
```diff
@@ -67,6 +69,8 @@ internal void AddLayer(Layer l)
         _baseImageConfig.AddLayer(l);
     }
 
+    internal ReadOnlyDictionary<string, string> EnvironmentVariables => _baseImageConfig.EnvironmentVariables;
```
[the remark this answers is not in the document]
vlada-shubina: If the logic is moved to `ImageBuilder`, this may not be needed.

src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj
```diff
@@ -64,13 +64,15 @@
     <Compile Remove="net472Definitions.cs" />
     <Compile Remove="VSHostObject.cs" />
   </ItemGroup>
+
```
vlada-shubina, line 67: nit: this whitespace change is not relevant

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/CreateNewImageTests.cs
```diff
@@ -72,6 +72,79 @@ public void CreateNewImage_Baseline()
...
+        task.BaseRegistry = "mcr.microsoft.com";
+        task.BaseImageName = "dotnet/runtime";
+        task.BaseImageTag = "8.0-preview";
+
+        // set up a bunch of variations on the aspnetcore urls
+        task.ContainerEnvironmentVariables = new[] {
+            new TaskItem("ASPNETCORE_URLS", new Dictionary<string, string> { { "Value", "https://*:12345;http://+:1234;http://localhost:123;http://1.2.3.4:12" } }) ,
+            new TaskItem("ASPNETCORE_HTTP_PORTS", new Dictionary<string, string> { { "Value", "999;666" } }),
+            new TaskItem("ASPNETCORE_HTTPS_PORTS", new Dictionary<string, string> { { "Value", "456;789" } })
+        };
```
vlada-shubina, line 111: consider moving those as datasets instead to have more granular failure in case the logic is broken (however it might be worse test run time wise).

baronfel: Moving the logic to ImageBuilder will make it easier to verify this, so moving to datasets wouldn't turn into a huge performance hit. I'll dig more into that :+1:

baronfel: I was able to do this and added to the ImageBuilderTests in the UnitTest project

src/Containers/Microsoft.NET.Build.Containers/ImageBuilder.cs
```diff
@@ -32,6 +37,10 @@ internal ImageBuilder(ManifestV2 manifest, ImageConfig baseImageConfig)
     /// </summary>
     internal BuiltImage Build()
     {
+        // before we build, we need to make sure that any image customizations occur
+        AssignUserFromEnvironment();
+        AssignPortsFromEnvironment();
```
baronfel, line 42: Here's the change @vlada-shubina requested - we assign these values immediately before construction, so that no matter how you `.Build()` an Image, you get the value applied. This means we don't need code in both ContainerBuilder and CreateNewImage

src/Containers/Microsoft.NET.Build.Containers/ImageConfig.cs
```diff
@@ -80,7 +84,7 @@ internal string BuildConfig()
         {
             newConfig["Labels"] = CreateLabelMap();
         }
-        if (_environmentVariables.Any())
+        if (_environmentVariables.Count != 0)
```
baronfel, line 87: analyzer flagged this for performance/correctness so I did it while I was here

baronfel:
The test failures are in C++/CLI test projects and have failed in other PRs, so shouldn't be considered as part of this.

vlada-shubina (approved):
looks good, couple of minor comments inline

src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
```diff
@@ -52,4 +52,12 @@ public static class ErrorCodes
         public static readonly string CONTAINER4003 = nameof(CONTAINER4003);
         public static readonly string CONTAINER4004 = nameof(CONTAINER4004);
     }
+
+    public static class EnvironmentVariables
```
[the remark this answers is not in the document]
baronfel: great idea, done!

PR #34178, opened 2023-07-21
Configurable wwwroot for StaticWebAssets

maraf:
The goal of this PR is to be able to override both input and output folder of static web assets. We need such configuration for WebAssembly template targeting nodejs/v8/etc. The `wwwroot` is not correct name in this context.

- Add new property `StaticWebAssetRootPath` (default `wwwroot`) allowing to configure target folder for static web assets 
- Add new item `StaticWebAssetSourcePath` (default `wwwroot`) allowing to configure source folder(s) static web assets
- Usage example in test asset `src\Assets\TestProjects\RazorAppWithP2PReferenceWithSwaPaths`
- Add test with complex Razor project
- Produce error when `StaticWebAssetRootPath` is not wwwroot for Blazor Wasm project


src/Assets/TestProjects/RazorAppWithP2PReferenceWithSwaPaths/AnotherClassLib/AnotherClassLib.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <!-- We don't want to run build server when not running as tests. -->
+    <UseRazorBuildServer>false</UseRazorBuildServer>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\ClassLibrary\ClassLibrary.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+</Project>
```
javiercn, line 17: Can we avoid adding an extra test project for this?

src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.6_0.targets
```diff
@@ -66,6 +66,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   <!-- Wire-up static web assets -->
   <PropertyGroup>
     <ResolveStaticWebAssetsInputsDependsOn>
+      _EnsureStaticWebAssetRootPath;
```
javiercn, line 69: I'm fine if we don't block people from changing this. But if we do, this should probably be in ResolveStaticWebAssetsConfiguration

src/StaticWebAssetsSdk/Sdk/Sdk.StaticWebAssets.StaticAssets.ProjectSystem.props
```diff
@@ -22,15 +30,15 @@ Copyright (c) .NET Foundation. All rights reserved.
...
     <DefaultItemExcludes>$(DefaultItemExcludes);**\bower_components\**;bower_components\**</DefaultItemExcludes>
-    <DefaultWebContentItemExcludes>$(DefaultWebContentItemExcludes);wwwroot\**</DefaultWebContentItemExcludes>
+    <DefaultWebContentItemExcludes>$(DefaultWebContentItemExcludes);@(StaticWebAssetSourcePath->'%(Identity)\**')</DefaultWebContentItemExcludes>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(EnableDefaultItems)' == 'true' And '$(EnableDefaultContentItems)' == 'true' ">
-    <!-- Publish everything under wwwroot, all JSON files, all config files and all Razor files -->
-    <Content Include="wwwroot\**" ExcludeFromSingleFile="true" CopyToPublishDirectory="PreserveNewest" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <!-- Publish everything under @(StaticWebAssetSourcePath), all JSON files, all config files and all Razor files -->
+    <Content Include="%(StaticWebAssetSourcePath->Identity)\**" ExcludeFromSingleFile="true" CopyToPublishDirectory="PreserveNewest" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
```
javiercn, line 38: Why can't this just be `$(StaticWebAssetRootPath)\**`

src/StaticWebAssetsSdk/Sdk/Sdk.StaticWebAssets.StaticAssets.ProjectSystem.props
```diff
@@ -11,6 +11,14 @@ Copyright (c) .NET Foundation. All rights reserved.
...
 <Project ToolsVersion="14.0">
 
+  <!-- Input and output configuration -->
+  <PropertyGroup>
+    <StaticWebAssetRootPath Condition="'$(StaticWebAssetRootPath)' == ''">wwwroot</StaticWebAssetRootPath>
+  </PropertyGroup>
+  <ItemGroup Condition="@(StaticWebAssetSourcePath->Count()) == 0">
+    <StaticWebAssetSourcePath Include="wwwroot" />
+  </ItemGroup>
+
```
javiercn, line 21: Do we actually need to make these input sources configurable? I would rather avoid this, as they are web specific. If people want to follow different conventions, that can be done in the project file directly.

src/StaticWebAssetsSdk/Tasks/DiscoverStaticWebAssets.cs
```diff
@@ -17,7 +17,7 @@ public class DiscoverStaticWebAssets : Task
         public ITaskItem[] Candidates { get; set; }
 
         [Required]
-        public string Pattern { get; set; }
+        public ITaskItem[] Patterns { get; set; }
```
javiercn, line 20: Can we avoid taking multiple patterns here? This can be called via MSBuild batching from outside instead, and simplifies things.

src/StaticWebAssetsSdk/Sdk/Sdk.StaticWebAssets.StaticAssets.ProjectSystem.props
maraf: That was my previous approach. With the current approach (which doesn't work yet :/ ) I'm trying to be able to have a nuget package with `public` folder, but target app with `client` folder. So that libraries/packages could be shared (if the actual code make sense to be shared). 

Does it make sense to you? 

Unfortunatelly this style of msbuild globing doesn't work outside of a target

src/StaticWebAssetsSdk/Sdk/Sdk.StaticWebAssets.StaticAssets.ProjectSystem.props
maraf: Intention described in the comment above..

PR #34225, opened 2023-07-24
Remove unused usings

MiYanni:
## Summary

Removed unused usings. This was automatically done for the entire solution via the Visual Studio feature.

We also have this editorconfig setting, but we currently set it to `warning`. I don't feel it is necessary to set this to `error`, though.

```powershell
# Remove unnecessary using directives
dotnet_diagnostic.IDE0005.severity = warning
```


dsplaisted:
I'm not sure I'm in favor of this.  I don't like it when I'm editing a file and basic namespaces such as System.Collections.Generic or System.Linq aren't included.  I think recent versions of VS handle this better, so maybe it's not as much of an issue now.  Or if it is, we could add those namespaces as global usings.

nagilson:
The changes overall look good to me, I went through every file and it seems correct :)
I like this idea, and specifically adding the ide warning. We could really use more code style analysis checks that are static and don't require human review. If we didn't remove the requires, then we'd need to add a bunch of warning disable messages in the files.

With that said, I see Daniel's point. I think some of the time VS will autogenerate or import it for you but I have had times when it gets confused because it doesn't know im trying to use a linq expression. Let's talk as a team maybe during standup to see what we can decide on together.

MiYanni:
> I'm not sure I'm in favor of this. I don't like it when I'm editing a file and basic namespaces such as System.Collections.Generic or System.Linq aren't included. I think recent versions of VS handle this better, so maybe it's not as much of an issue now. Or if it is, we could add those namespaces as global usings.

Generally, I don't think that has been a concern for a while... AFAIK, intellisense via Roslyn metadata understands *most* of the runtime's methods. So here, I don't have `System.Linq` imported and when I type `ToList` and press enter, it just adds it automatically.

Personally, I used Resharper before coming to Microsoft which did this via Alt+Enter, but I think VS has been natively doing this for a while now...? After some quick research. it showing the autocompletion for other non-imported methods seems to be this feature:

image

And it adding the using is from this option:

image

> The changes overall look good to me, I went through every file and it seems correct :) I like this idea, and specifically adding the ide warning. We could really use more code style analysis checks that are static and don't require human review. If we didn't remove the requires, then we'd need to add a bunch of warning disable messages in the files.
> 
> With that said, I see Daniel's point. I think some of the time VS will autogenerate or import it for you but I have had times when it gets confused because it doesn't know im trying to use a linq expression. Let's talk as a team maybe during standup to see what we can decide on together.

The warning is already enabled for the repo. It isn't a compiler warning, but an IDE warning which manifests as all unused usings being **greyed out**. This is why I created the PR in the first place as the amount of unused usings was in many files I was opening.

image

When you hover over it, it shows you the warning and allows you to take action against it.

image

marcpopMSFT (approved):

PR #34264, opened 2023-07-26
Add global and implicit usings

MiYanni:
## FOR REVIEWERS
Review each change **per commit**. All files is too large for GitHub to render.

## Summary
After the previous [remove unused usings PR](https://github.com/dotnet/sdk/pull/34225), it was suggested that we also enable both [ImplicitUsings](https://rehansaeed.com/the-problem-with-csharp-10-implicit-usings/#what-are-implicit-using-statements) and add some [global usings](https://devblogs.microsoft.com/dotnet/welcome-to-csharp-10/) for common namespaces we use in the repo. So, I've done that in this PR.

## Process
- Created a script to collate `using` statements from the repo and counts of those statements
- Add `<ImplicitUsings>enable</ImplicitUsings>`, investigate build issues, and implement solutions for those issues
- Determine which global usings (based on script output) that either apply to the entire repo or only for test projects
- Add root/test global usings, investigate build issues, and implement solutions for those issues
- Clean all unused usings after the tasks above were completed

## Script for repo `usings`
Here's the script I created to collect all using statements and determine the counts of those statements:
```powershell
$prefix = 'using '
$files = Get-ChildItem -Path *.cs -Recurse
$results = $files | Get-Content | ForEach-Object { if($_.StartsWith($prefix)){ $_.Substring($prefix.Length, ($_.Length - $prefix.Length - 1)) } } | Group-Object | Sort-Object Count -Descending | ForEach-Object { "$($_.Count), $($_.Name)" }
$results | Out-File -FilePath '.\Usings.txt'
```

To keep things simple, I only looked at namespaces that had 100 uses or more:
<details>
  <summary>See list of top usings</summary>

  ```powershell
1474, System
848, System.Collections.Generic
791, System.IO
657, System.Linq
575, Xunit
451, FluentAssertions
434, Xunit.Abstractions
400, Microsoft.NET.TestFramework
354, Microsoft.DotNet.Cli.Utils
325, Microsoft.NET.TestFramework.Assertions
322, Microsoft.NET.TestFramework.Commands
268, System.CommandLine
221, Microsoft.Build.Framework
207, System.Threading.Tasks
199, System.Text
167, Newtonsoft.Json
147, System.Runtime.InteropServices
145, Microsoft.DotNet.Cli
141, System.Xml.Linq
140, Microsoft.Build.Utilities
139, System.Diagnostics
130, Microsoft.NET.TestFramework.ProjectConstruction
122, System.Threading
115, Microsoft.Extensions.EnvironmentAbstractions
112, System.Reflection
109, Microsoft.TemplateEngine.Abstractions
106, System.Runtime.CompilerServices
  ```
</details>

I eliminated the ones that are part of the `<ImplicitUsings>enable</ImplicitUsings>`, which are:
```csharp
// <auto-generated/>
global using global::System;
```

I tried various high-usage namespaces and systematically determined which ones made sense based on the number of errors provided when they were added.

## Per Commit Info

This I tested at first by adding the usings generated by hand and determined which namespaces caused issues. Determined that the only one that was causing issues was the `System.Threading.Tasks`. Determined it was fixable by adjusting a few projects and like 4 files manually.

I originally had `Moq` on this list of usings for the `Tests` folder, but the usage wasn't enough and caused more issues than it was worth. The result set seemed to be logical for a majority of the projects, as most of them are actual unit test projects. There are some projects in the Tests folder that are utility projects for testing, so those needed some removal of those included usings as they don't reference those packages.

This is where I realized that any project from the TemplateEngine team has the `unused usings` setting set to always error. So, all of those projects needed to have their unused usings removed. I also tried added these global usings, but they ultimately don't make sense in most cases:
```xml
<Using Include="System.CommandLine" />
<Using Include="Microsoft.Build.Framework" />
<Using Include="Microsoft.Build.Utilities" />
```

I'm sure there are specific projects these would work well to have, but I didn't that too deep into it. I wanted this PR to handle a majority of usings that make sense to have in the repo.

This took more work than expected. I believe I missed many files on my first PR based on how the feature works in VS. It seems to have a limit (not known to the user) for the number of files it can adjust at a time. There's also no mechanism to multi-select in the UI presented. Therefore, I had to use the 'uncheck all' on the files presented, and then *manually press space bar and down arrow* on an arbitrary amount of files. I did this like 15 times or more. I eventually got every file cleaned up and double checked to make sure.


nohwnd (approved):

nagilson (approved):

dsplaisted (approved):
Awesome!  Adding `Microsoft.NET.TestFramework.Assertions` as a global using will make writing new tests easier, that's one that I always have to add manually to test classes (it happened to me today!).

For the types where aliases are needed to disambiguate, is it possible to use the original name?  IE to alias `Microsoft.Build.Utilities.Task` to just `Task` instead of to `MSBuildTask`.  If both Task types aren't being used in the same code that seems a bit better to me and would result in less code change.

MiYanni:
> Awesome! Adding `Microsoft.NET.TestFramework.Assertions` as a global using will make writing new tests easier, that's one that I always have to add manually to test classes (it happened to me today!).
> 
> For the types where aliases are needed to disambiguate, is it possible to use the original name? IE to alias `Microsoft.Build.Utilities.Task` to just `Task` instead of to `MSBuildTask`. If both Task types aren't being used in the same code that seems a bit better to me and would result in less code change.

Sure thing. The reason I used different alias names is, in the past when I've experienced name clashes, I felt like the guidance was to make the alias name for the *more specific type* have the name more... specified. Lol. Like, `Microsoft.Build.Utilities.Task` is a more specific kind of task than `System.Threading.Tasks.Task`. So, if for some reason you needed to use both aliases in the same file, then you can just copy/paste your already existing aliases, and the code would look the same throughout the codebase. It happened a lot when making adapters between 2-like things but from different libraries.

MiYanni:
## 5th: ee471d60180fbfb8a447ad4ae32a474aefade8ca
> Changed aliases to default names. Adjusted 3 files that had name conflicts for Task by fully qualifying.

Based on suggestion to use the normal class names instead of disambiguated ones. Here's a table for before/after:

File Count | Prev Alias | New Alias
----- | ---------- | ---------
58    | MSBuildTask | Task
1     | MSBuildTaskItem | TaskItem
1     | TETestLoggerFactory | TestLoggerFactory
2     | TESharedTestOutputHelper | SharedTestOutputHelper
2     | CLIBuildCommand | BuildCommand
2     | CLIPublishCommand | PublishCommand
1     | CLICleanCommand | CleanCommand
1     | CLIMSBuildCommand | MSBuildCommand
1     | CLIPackCommand | PackCommand
1     | CLIRestoreCommand | RestoreCommand
1     | CLITestCommand | TestCommand

There were 3 files that had conflicts. To resolve them, since they were very simple, I just fully qualified the usage of `System.Threading.Tasks.Task` in those cases:
- ZipDeploymentStatus.cs
- KuduDeploy.cs
- HttpResponseMessageForStatusCode.cs

drewnoakes:
> Created a script to collate using statements from the repo and counts of those statements

Shameless plug for https://github.com/drewnoakes/amusing

MiYanni:
> > Created a script to collate using statements from the repo and counts of those statements
> 
> Shameless plug for https://github.com/drewnoakes/amusing
> 
> Running it on `sdk.sln` after this PR gives:
> ...

Nice! There could definitely be some more trimming there, but that is *way* better than the close to 1000 most of the top usings prior were. I tried several of these new top namespaces before making this PR, but some of them are complicated as they're namespaces to projects in the solution itself. I'd have to be a bit more... *particular* about how I'd approach those. For now, I'm satisfied with this first blush at some cleanup. I'll likely do more things later, but other priorities come first.

PR #34266, opened 2023-07-26
Change the hard code of newtonsoft package version

v-wuzhai:
Use [NewtonsoftJsonPackageVersion](https://github.com/dotnet/sdk/blob/aa4e1e2176bca02d087e98e935fb4c99ab34c7a2/eng/Versions.props#L31C24-L31C24) to replace the newtonsoft package version.


src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
```diff
@@ -121,6 +124,27 @@ public TestAsset UpdateProjProperty(string propertyName, string variableName, st
...
+
+            return WithProjectChanges(project =>
+            {
+                var ns = project.Root.Name.Namespace;
+                foreach (var PropertyName in PropertyNames)
+                {
+                    var packageReferencesToUpdate =
+                        project.Root.Descendants(ns + PropertyName)
+                            .Where(p => p.Attribute("Version") != null && p.Attribute("Version").Value.Equals($"$({targetName})", StringComparison.OrdinalIgnoreCase));
+                    foreach (var packageReference in packageReferencesToUpdate)
```
marcpopMSFT, line 140: Can there be more than one newtonsoft package reference? Is this for test solutions that have multiple projects?

src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
```diff
@@ -121,6 +124,27 @@ public TestAsset UpdateProjProperty(string propertyName, string variableName, st
             }, targetValue);
         }
 
+        public TestAsset ToReplaceTheNewtonsoftJsonPackageVersionVariable()
```
marcpopMSFT, line 127: nit: I'd remove the To in front so the action is the first word
```suggestion
        public TestAsset ReplaceTheNewtonsoftJsonPackageVersionVariable()
```

src/Tests/Microsoft.NET.TestFramework/Microsoft.NET.TestFramework.csproj
```diff
@@ -20,6 +20,13 @@
     <TargetFrameworks>$(SdkTargetFramework);net472</TargetFrameworks>
   </PropertyGroup>
 
+  <ItemGroup>
```
marcpopMSFT, line 23: Clever way of passing through the newtonsoft version the SDK builds with to the tests. I can't think of a reason why we'd want this to be different between the tests and product. @dsplaisted , any concerns with this approach?

src/Tests/Microsoft.NET.TestFramework/TestAsset.cs
v-wuzhai: - Because we just replace $(NewtonsoftJsonPackageVersion) parameters, so it can have more than one newtonsoft package reference.

- Yes, this can apply to multiple projects. Because we find $(NewtonsoftJsonPackageVersion) parameters in all '.proj' and '.xml' files.

dsplaisted (approved):

PR #34286, opened 2023-07-26
[Containers] validate that the normalized repository name is a meaningful name

baronfel:
## Description

When inferring a container name based on the project name the tooling attempts to normalize the project name and remove any invalid characters from the normalized name. This is to ensure that the generated images are usable by all container registries. However, we didn't cover the case where the entire input string was invalid characters - in this case we should fail fast and notify the user that they need to provide an explicit value for the ContainerRepository.

This fixes AzDO work item [1858645](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1858645).

## Customer Impact

GB18030 customers (and other customers whose entire image name post-normalization was comprised of normalized characters) would error during image push with a useless and confusing error.

## Regression
**No** - this validation has been in place since our initial releases

## Risk
**Low** - we have a lot of testing around name validation, and it just got both stricter and more clear when erroring

## Testing

Manual and automated - we explicitly cover two cases we didn't before:
* all invalid container repository name characters
* first name character is a non-latin alphanumeric character

This should flow to 7.0.4xx and main to cover the other in-support SDK versions that shipped the containers feature.

### DIFF
--- a/src/Assets/TestProjects/NetCoreCsharpAppReferenceCppCliLib/CSConsoleApp/CSConsoleApp.csproj
+++ b/src/Assets/TestProjects/NetCoreCsharpAppReferenceCppCliLib/CSConsoleApp/CSConsoleApp.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>$(CurrentTargetFramework)</TargetFramework>
+    <TargetFramework>$(CurrentTargetFramework)-windows</TargetFramework>
     <Platforms>x64</Platforms>
     <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
--- a/src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -3,6 +3,7 @@
 
 #if NETFRAMEWORK
 using System;
+using System.Linq;
 #endif
 #if NET
 using System.Diagnostics.CodeAnalysis;
@@ -279,7 +280,13 @@ public static class ContainerHelpers
                 throw new ArgumentException(Resources.Resource.GetString(nameof(Strings.InvalidImageName)));
             }
             var loweredImageName = containerImageName.ToLowerInvariant();
-            normalizedImageName = imageNameCharacters.Replace(loweredImageName, "-");
+            var potentialNormalizedName = imageNameCharacters.Replace(loweredImageName, "-");
+            if (potentialNormalizedName.All(c => c == '-'))
+            {
+                // The name was normalized to all dashes, so there was nothing recoverable. We should throw.
+                throw new ArgumentException(Resources.Resource.GetString(nameof(Strings.InvalidImageName)));
+            }
+            normalizedImageName = potentialNormalizedName;
             return false;
         }
     }
--- a/src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -8,6 +8,7 @@ using System.Linq;
 #if NET
 using System.Diagnostics.CodeAnalysis;
 #endif
+using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.NET.Build.Containers.Resources;
 
@@ -31,13 +32,6 @@ public static class ContainerHelpers
     /// See <see href="https://github.com/distribution/distribution/blob/78b9c98c5c31c30d74f9acb7d96f98552f2cf78f/reference/normalize.go">normalize.go</see>.
     /// </summary>
     internal const string DefaultRegistry = "docker.io";
-
-    /// <summary>
-    /// Matches if the string is not lowercase or numeric, or ., _, or -.
-    /// </summary>
-    /// <remarks>Technically the period should be allowed as well, but due to inconsistent support between cloud providers we're removing it.</remarks>
-    private static Regex imageNameCharacters = new Regex(@"[^a-z0-9_\-/]");
-
 
     /// <summary>
     /// The enum contains possible error reasons during port parsing using <see cref="TryParsePort(string, out Port?, out ParsePortError?)"/> or <see cref="TryParsePort(string?, string?, out Port?, out ParsePortError?)"/>.
@@ -262,6 +256,8 @@ public static class ContainerHelpers
         return true;
     }
 
+
+
     /// <summary>
     /// Checks if a given container image name adheres to the image name spec. If not, and recoverable, then normalizes invalid characters.
     /// </summary>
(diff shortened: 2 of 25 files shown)


src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -279,7 +280,13 @@ internal static bool IsValidEnvironmentVariable(string envVar)
                 throw new ArgumentException(Resources.Resource.GetString(nameof(Strings.InvalidImageName)));
             }
             var loweredImageName = containerImageName.ToLowerInvariant();
-            normalizedImageName = imageNameCharacters.Replace(loweredImageName, "-");
+            var potentialNormalizedName = imageNameCharacters.Replace(loweredImageName, "-");
+            if (potentialNormalizedName.All(c => c == '-'))
```
vlada-shubina, line 284: should we also check if first char is not a dash? error message says "The first character of the image name must be a lowercase letter or a digit"

src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
```diff
@@ -40,6 +40,7 @@ public static class ErrorCodes
         public static readonly string CONTAINER2009 = nameof(CONTAINER2009);
         public static readonly string CONTAINER2010 = nameof(CONTAINER2010);
         public static readonly string CONTAINER2012 = nameof(CONTAINER2012);
+        public static readonly string CONTAINER2014 = nameof(CONTAINER2014);
```
vlada-shubina, line 43: this code is not defined, is it needed?

src/Containers/Microsoft.NET.Build.Containers/Resources/Strings.resx
```diff
@@ -193,7 +193,7 @@
     <comment>{StrBegin="CONTAINER2015: "}</comment>
   </data>
   <data name="InvalidImageName" xml:space="preserve">
-    <value>CONTAINER2005: The first character of the image name must be a lowercase letter or a digit.</value>
+    <value>CONTAINER2005: The first character of the image name must be a lowercase letter or a digit and all characters in the name must be an alphanumeric character, -, /, or _.</value>
```
vlada-shubina, line 196: imo, this error message doesn't resolve the issue well.
- _The first character of the image name must be a lowercase letter or a digit._ is enough reason to fail "----"
- "all characters in the name must be an alphanumeric character, -, /, or _" does not fail "-----"
- if we have differentiate this case, would this wording will be better - "none of provided chars in the image name are supported as the image name" (and have different code possibly)

vlada-shubina (commented):
Couple of comments inline.
Imo existing error message should suffice, just logic is not completely correct.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ParseContainerPropertiesTests.cs
```diff
@@ -100,4 +100,21 @@ public void CanOnlySupplyOneOfTagAndTags()
         Assert.True(logs.Errors.Count > 0);
         Assert.Equal(logs.Errors[0].Code, ErrorCodes.CONTAINER2008);
     }
+
+    [Fact]
```
vlada-shubina, line 104: As it, this should be 
```suggestion
    [DockerDaemonAvailableFact]
```

src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: I think this check is covered by 278-281 above - what do you think?

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ParseContainerPropertiesTests.cs
baronfel: That's a great point - I was looking at the ComputeContainerConfig target (and the ParseContainerProperties task) and none of them require Docker at all so I figured this Fact would be enough. I can try checking what other tests don't require it in this file.

src/Containers/Microsoft.NET.Build.Containers/KnownStrings.cs
baronfel: This is defined - it's the InvalidContainerImageName error that we throw in the validation, and I wanted a constant to refer to it by in the test I added.

Having said that, if we tweak the message to be more descriptive then a new code would be necessary.

src/Containers/Microsoft.NET.Build.Containers/Resources/Strings.resx
baronfel: Yeah, I agree - I went with the existing message for implementation speed, but users would be better served with a new message and code

baronfel: I added a new message (with the same code) that we use for the new error case:

```
CONTAINER2005: The inferred image name '{0}' contains entirely invalid characters. The valid characters for an image name are alphanumeric characters, -, /, or _, and the image name must start with an alphanumeric character.
```

src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -264,23 +265,45 @@ internal static bool IsValidEnvironmentVariable(string envVar)
     /// <summary>
     /// Checks if a given container image name adheres to the image name spec. If not, and recoverable, then normalizes invalid characters.
     /// </summary>
-    internal static bool NormalizeImageName(string containerImageName,
-                                         [NotNullWhen(false)] out string? normalizedImageName)
+    internal static (string? normalizedImageName, (string, object[])? normalizationWarning, (string, object[])? normalizationError) NormalizeImageName(string containerImageName)
```
baronfel, line 268: I changed this method to return warning/error information to the caller, because the wrapping we were doing in ArgumentException was being incorrectly unwrapped in the ParseContainerProperties task, leading to not-very-useful error messages.

src/Containers/Microsoft.NET.Build.Containers/Tasks/ParseContainerProperties.cs
```diff
@@ -141,22 +141,19 @@ public override bool Execute()
...
+        if (normalizationWarning is (string warningMessageKey, object[] warningParams))
+        {
+            Log.LogMessageFromResources(warningMessageKey, warningParams);
         }
-        catch (ArgumentException)
+
+        if (normalizationError is (string errorMessageKey, object[] errorParams))
         {
-            Log.LogErrorWithCodeFromResources(nameof(Strings.InvalidContainerImageName), nameof(ContainerImageName), ContainerImageName);
+            Log.LogErrorWithCodeFromResources(errorMessageKey, errorParams);
```
baronfel, line 156: Instead of incorrectly forwarding the error message to the user via the ArgumentException, we now just let the Normalize call tell us what to do with the messages directly.

Warnings are allowed to continue execution, while errors might end up stopping the execution.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ParseContainerPropertiesTests.cs
```diff
@@ -63,7 +63,7 @@ public void RegexCatchesInvalidContainerNames()
         using var _ = d;
         var instance = project.CreateProjectInstance(global::Microsoft.Build.Execution.ProjectInstanceSettings.None);
         Assert.True(instance.Build(new[]{ComputeContainerConfig}, new [] { logs }, null, out var outputs));
-        Assert.Contains(logs.Messages, m => m.Message?.Contains("'ContainerImageName' was not a valid container image name, it was normalized to 'dotnet-testimage'") == true);
+        Assert.Contains(logs.Messages, m => m.Message?.Contains("'dotnet testimage' was not a valid container image name, it was normalized to 'dotnet-testimage'") == true);
```
baronfel, line 66: Fixed a bug here - the intent was to show the actual value that was invalid, now we do so

src/Containers/Microsoft.NET.Build.Containers/ContainerHelpers.cs
vlada-shubina: regex on line 286 might replace the first character to a dash - will it be an issue?

baronfel: I don't think the regex will replace the first character ever, because the check on 278 (`!Char.IsLetterOrDigit`) is stronger than the regex, so if that check fails we should never hit the regex case, and if that check passes it should also pass the regex.

vlada-shubina: `Char.IsLetterOrDigit` supports any Unicode character, whereas regex has only latin letters.

So
```
var imageName = "㓳";
var isLetter = Char.IsLetterOrDigit(imageName[0]); //true
var potentialNormalizedName = imageNameCharacters.Replace(imageName, "-"); //"-"
```

baronfel: You're exactly correct - so I should move the Char.IsLetterOrDigit check to a Latin check regex instead and then what I _thought_ was happening would actually happen. great catch.

vlada-shubina: Using similar check `(c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')` may be more lightweight.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
```diff
@@ -26,13 +27,13 @@ public EndToEndTests(ITestOutputHelper testOutput)
 
     public static string NewImageName([CallerMemberName] string callerMemberName = "")
```
vlada-shubina, line 28: nit
```suggestion
    public static string GetNewImageName([CallerMemberName] string callerMemberName = "")
```

vlada-shubina (approved):
imo, the logging strategy makes the introduced logic more complicated then it should.

For servicing, I would recommend to go with minimum changes needed, i.e. put the method directly to task class and perform logging in line.

For .NET 8, consider implementing logging as in `CreateNewImage` task.
https://github.com/dotnet/sdk/blob/dff005cfa5943f677858ec3d6e29b27574e25a12/src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs#L42-L44
This allows logging via `ILogger` from anywhere.

Also, please put the PR to `main` before requesting servicing.

src/Assets/TestProjects/NetCoreCsharpAppReferenceCppCliLib/CSConsoleApp/CSConsoleApp.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>$(CurrentTargetFramework)</TargetFramework>
+    <TargetFramework>$(CurrentTargetFramework)-windows</TargetFramework>
```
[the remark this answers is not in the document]
baronfel: We have to take this change - the CI runners changed configuration and now include a version of the C++ tooling that requires building for windows - this hit a bunch of other PRs, but hadn't hit the 7.0.300 branch because we don't have any active codeflow to this branch. This PR was just the unlucky one that triggered a build showing the issue :)

I've taken the change directly from the fix commits on the other branches, so I could make that case to tactics if necessary.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
baronfel: I'd rather not do the rename here to prevent additional churn if that's ok.

baronfel:
I'm not sure I can move NormalizeImageName to ParseContainerProperties easily - it's also used as a standlone utility in tests, so I either:
* make the tests create ParseContainerProperties (with MSBuild log support/detection) as well, or
* move to ILogger in servicing, which seems like a large change as well.

I do agree that in 8 we should move to ILogger here and consider consolidation in ParseContainerProperties.

PR #34289, opened 2023-07-26
Fix the end of build install of workloads to always refresh the rollback file

marcpopMSFT:
It was failing if you had previously built with the same SDK but prior preview runtime as it wasn't updating the file. This conflicted with https://github.com/dotnet/sdk/pull/32882 in that I started getting exceptions from that build task because I had a preview 5 TestRollback.json but the task was looking for rc1 versions.


MiYanni (approved):
Checked [the doc on WriteLinesToFile](https://learn.microsoft.com/en-us/visualstudio/msbuild/writelinestofile-task?view=vs-2022) just to be sure. Looks good.

PR #34322, opened 2023-07-29
First-run workload experience

### LINKED ISSUE #31872: Improve the first-run experience for workloads - detect common version mismatch scenarios and tell the user the actions to perform to resolve them
### Summary
Currently, upgrading the .NET SDK for either a patch version upgrade or feature band upgrade will overwrite the manifests for emsdk and runtime. There are reasons that runtime wants to update the baseline manifest with each feature band release including staying in-sync with the runtime pack version and security.

The side effect of this is that a customer upgrading their SDK (or worse, getting upgraded by Microsoft Update) will find their workloads are no longer working.

A suggestion for improving this experience for the customer is on first run of the SDK, we detect the customer is in this situation, and try to resolve it for them.

**Requirements**
- Only run the first run after a new SDK install
- Should only run once even if the command fails or is cancelled
- Needs to be cancellable (can it be interactive?)
- Workloads state
  - Detect workload packs have been installed for this feature band (future, detect from a prior feature band)
  - Detect that the current manifests don't match the packs installed
  - ~~Detect that workloads were **not** installed using a rollback file~~
- Don't run this in CI
- Trigger a workload update or install trying to get the current state working between manifests and packs
- This should run anywhere were the SDK would install to a common location (windows, macos, maybe linux)

MiYanni:
Fixes: https://github.com/dotnet/sdk/issues/31872

## Summary
There are certain situations where the user of the CLI might have updated the SDK version which brings in updated workload manifests within the same feature band. The user had workloads installed previously. However, they won't have the updated packages for the workloads that apply to the new manifests.

This PR makes it so that when the first-run experience for an SDK version activates, it (conditionally) runs the process for checking the state of the installed workloads, similar to the process done for the `workload update` command. For the purposes of this feature, I've called it "workload integrity check" since (conceptually) makes sure you're in a 'good' state with the workloads desired.

## Key Aspects
- The workload integrity check only runs when the `.dotnetFirstUseSentinel` for the proper SDK version is not present on-disk.
- The `.dotnetFirstUseSentinel` is created prior to running the workload integrity check. This means that if, for any reason, the workload check doesn't work or run, it won't try to do it again.
- The workload integrity check is wrapped in a try/catch so any failures will not hinder the CLI command from executing.
  - *Update*: Added a log message on failure to advise the user to run `dotnet workload update` if they want the actual error details.
- The workload integrity check only runs if there were workloads previously installed for the feature band.
- The workload integrity check will not output the NuGet Package Manager logging information (it is quite extensive and unnecessary in most cases).
- There is a new environment variable called `DOTNET_SKIP_WORKLOAD_INTEGRITY_CHECK`. ~This is set to `false` by default. If set to `true`, the workload integrity check will be skipped.~
  - *Update*: If the environment variable is not present, the default value is set to `true` (meaning, skip the check) if *running in CI*. Otherwise, `false` is the default.
- ~If the first-run command is a workload command, the workload integrity check is skipped. This avoids an open file conflict with the workload log file.~
  - *Update*: Fixed this by updating the workload log filename to include milliseconds to avoid the "file being open" conflict.
- If the first-run command is ran in a CI environment, the workload integrity check is skipped. The CI detection for this is the same as what used for telemetry.
- Currently, user cancellation is not implemented. A new issue will be opened to outline cancellation implementation.

## Testing Notes
- Tested this both with MSI and file-based installs.
- Tested certain aspects by deleting installed workload packs and/or overriding workload manifests with previous versions.
- Tested that workload integrity check failures (via improper manifest version dependencies) are properly caught and ignored by the try/catch.
- Tested that the first-run command being a `workload` command is properly skipping the workload integrity check.
- Tested that if a CI-based environment variable is set, the workload integrity check is skipped.

## Example Output
image

## Other
- PR contains some minor unrelated changes such as whitespace or comments in other files. These were done during development and investigation, and do not include functional changes.


src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -15,6 +15,7 @@
 
 namespace Microsoft.DotNet.Cli.NuGetPackageDownloader
 {
+    // TODO: Never name a class the same name as the namespace. Update either for easier type resolution.
```
MiYanni, line 18: Added this TODO for (*foreshadowing*) future cleanup purposes.

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandBase.cs
```diff
@@ -123,17 +123,18 @@ protected bool VerifySignatures
         /// <param name="parseResult">The results of parsing the command line.</param>
         /// <returns><see langword="true"/> if signatures of packages and installers should be verified.</returns>
         /// <exception cref="GracefulException" />
-        private static bool ShouldVerifySignatures(ParseResult parseResult)
+        private static bool ShouldVerifySignatures(ParseResult parseResult) =>
+            ShouldVerifySignatures(parseResult.GetValue(WorkloadInstallCommandParser.SkipSignCheckOption));
+
+        public static bool ShouldVerifySignatures(bool skipSignCheck = false)
```
MiYanni, line 129: Method needed to be public for workload integrity check.

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandBase.cs
```diff
@@ -123,17 +123,18 @@ protected bool VerifySignatures
...
         {
             if (!SignCheck.IsDotNetSigned())
             {
                 // Can't enforce anything if we already allowed an unsigned dotnet to be installed.
                 return false;
             }
 
-            bool skipSignCheck = parseResult.GetValue(WorkloadInstallCommandParser.SkipSignCheckOption);
```
MiYanni, line 134: This value was unused.

src/Cli/dotnet/commands/dotnet-workload/WorkloadIntegrityChecker.cs
```diff
@@ -0,0 +1,53 @@
...
+                verifySignatures,
+                packageDownloader,
+                creationResult.DotnetPath);
+            var repository = installer.GetWorkloadInstallationRecordRepository();
+            var installedWorkloads = repository.GetInstalledWorkloads(sdkFeatureBand);
+
+            if (installedWorkloads.Any())
+            {
+                reporter.WriteLine(LocalizableStrings.WorkloadIntegrityCheck);
+                CliTransaction.RunNew(context => installer.InstallWorkloads(installedWorkloads, sdkFeatureBand, context));
```
dsplaisted, line 48: I think it may be better to first check if there are any missing workload packs, and only if that's the case try to install them.  The difference would be that we could print a message saying that workload packs are missing and that they are being installed.  This may be important for msi-based installs, as when it tries to install the workload packs it should pop up an elevation prompt, and a clearer message might help people understand what it's for.

dsplaisted (commented):
Is it currently not possible to cancel the integrity check at all?  Or if you try to cancel, does it cancel the whole command instead of just the integrity check (which seems OK to me)?

src/Cli/dotnet/commands/dotnet-workload/WorkloadIntegrityChecker.cs
MiYanni: I tried to keep the changes simplistic since I wasn't sure if we'd be changing anything else related to workload installs. So, install workloads currently does this kind of check:
image

The check is within the transaction context. Not sure if that is necessary?? Like, we could just pre-filter the packs before creating the context. Looks like it is only done for message purposes. I didn't want to change this stuff too much as testing the first-run scenario is *really* time consuming.

If we want to change the messaging, should there really be different messaging between this new process and that of `workload update`/`workload install`? Testing the CLI command is fairly straight-forward compared to what this PR is fixing.

MiYanni:
> Is it currently not possible to cancel the integrity check at all? Or if you try to cancel, does it cancel the whole command instead of just the integrity check (which seems OK to me)?

Cancelling cancels the entire command (including the workload integrity check). It just immediately stops doing anything. Here, I cancel as soon as it starts trying to install the workload pack.

CancelWorkloadFirstRun

I'll be creating an issue for supporting cancellation in our use of S.CL in the codebase. For this PR however, as soon as the command is cancelled, it will not reattempt to run the workload integrity check again since the first-run sentinel file exists, which is the desired behavior.

src/Cli/dotnet/Program.cs
```diff
@@ -182,11 +180,20 @@ internal static int ProcessArgs(string[] args, TimeSpan startupTime, ITelemetry
                         isDotnetBeingInvokedFromNativeInstaller = true;
                     }
 
+                    // Override the workload integrity check if the command being executed is a workload command or the command is being ran in CI.
+                    // A workload command causes issues as the command opens a log and the integrity check will get a "The process cannot access the file because it is being used by another process." error for the log file.
+                    // Basically, they're both trying to write to the same log at the same time.
```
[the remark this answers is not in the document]
MiYanni: I'd have to dig deeper into what is actually happening. This is the exception I was seeing:
```
Unhandled exception: System.IO.IOException: The process cannot access the file 'C:\Users\LocalAdmin\AppData\Local\Temp\Microsoft.NET.Workload_8956_20230728_154248.log' because it is being used by another process.
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.StreamWriter.ValidateArgsAndOpenPath(String path, Boolean append, Encoding encoding, Int32 bufferSize)
   at System.IO.File.CreateText(String path)
   at Microsoft.DotNet.Installer.Windows.TimestampedFileLogger..ctor(String path, Int32 flushThreshold, String[] logPipeNames) in C:\Code\sdk\src\Cli\dotnet\Installer\Windows\TimestampedFileLogger.cs:line 70
   at Microsoft.DotNet.Installer.Windows.TimestampedFileLogger..ctor(String path) in C:\Code\sdk\src\Cli\dotnet\Installer\Windows\TimestampedFileLogger.cs:line 56
   at Microsoft.DotNet.Workloads.Workload.Install.NetSdkMsiInstallerClient.Create(Boolean verifySignatures, SdkFeatureBand sdkFeatureBand, IWorkloadResolver workloadResolver, INuGetPackageDownloader nugetPackageDownloader, VerbosityOptions verbosity, PackageSourceLocation packageSourceLocation, IReporter reporter, String tempDirPath, RestoreActionConfig restoreActionConfig) in C:\Code\sdk\src\Cli\dotnet\commands\dotnet-workload\install\NetSdkMsiInstallerClient.cs:line 1033
   at Microsoft.DotNet.Workloads.Workload.Install.WorkloadInstallerFactory.GetWorkloadInstaller(IReporter reporter, SdkFeatureBand sdkFeatureBand, IWorkloadResolver workloadResolver, VerbosityOptions verbosity, String userProfileDir, Boolean verifySignatures, INuGetPackageDownloader nugetPackageDownloader, String dotnetDir, String tempDirPath, PackageSourceLocation packageSourceLocation, RestoreActionConfig restoreActionConfig, Boolean elevationRequired) in C:\Code\sdk\src\Cli\dotnet\commands\dotnet-workload\install\WorkloadInstallerFactory.cs:line 40
   at Microsoft.DotNet.Workloads.Workload.List.WorkloadInfoHelper..ctor(Boolean isInteractive, VerbosityOptions verbosity, String targetSdkVersion, Nullable`1 verifySignatures, IReporter reporter, IWorkloadInstallationRecordRepository workloadRecordRepo, String currentSdkVersion, String dotnetDir, String userProfileDir, IWorkloadResolver workloadResolver) in C:\Code\sdk\src\Cli\dotnet\commands\dotnet-workload\WorkloadInfoHelper.cs:line 57
   at Microsoft.DotNet.Cli.WorkloadCommandParser.ShowWorkloadsInfo(ParseResult parseResult, IWorkloadInfoHelper workloadInfoHelper, IReporter reporter, String dotnetDir) in C:\Code\sdk\src\Cli\dotnet\commands\dotnet-workload\WorkloadCommandParser.cs:line 44
   at Microsoft.DotNet.Cli.WorkloadCommandParser.ProcessArgs(ParseResult parseResult) in C:\Code\sdk\src\Cli\dotnet\commands\dotnet-workload\WorkloadCommandParser.cs:line 89
   at System.CommandLine.Invocation.AnonymousCliAction.Invoke(ParseResult parseResult)
   at System.CommandLine.Invocation.InvocationPipeline.Invoke(ParseResult parseResult)
   at System.CommandLine.ParseResult.Invoke()
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, TimeSpan startupTime, ITelemetry telemetryClient) in C:\Code\sdk\src\Cli\dotnet\Program.cs:line 230
```

When I told Marc that this was happening, he said to just make the workload integrity check not run on `workload` commands, which was a simple fix. I could investigate why this is happening if we don't want this restriction.

src/Cli/dotnet/Program.cs
```diff
@@ -324,6 +333,18 @@ private static void ReportDotnetHomeUsage(IEnvironmentProvider provider)
...
+
+            if (isFirstTimeUse && !dotnetFirstRunConfiguration.SkipWorkloadIntegrityCheck)
+            {
+                try
+                {
+                    WorkloadIntegrityChecker.RunFirstUseCheck(reporter);
+                }
+                catch (Exception)
+                {
+                    // If the workload check fails for any reason, we want to eat the failure and continue running the command.
```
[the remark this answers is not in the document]
MiYanni: I can add a generic message about the workload integrity check failing. And if they want to investigate, they can run `workload update` which will likely show whatever error was encountered. The workload integrity check isn't supposed to give friction for the actual executed command, but an output entry shouldn't be problematic.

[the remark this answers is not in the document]
MiYanni: I went ahead and added a message when it fails. If you do what the message says (run `dotnet workload update`), you get the actual exception thrown by workloads. To test this, I put a mismatched set of workload manifests so the version present wouldn't work for the dependency's version requirement.

WorkloadFirstRunErrorMessage

src/Cli/dotnet/Program.cs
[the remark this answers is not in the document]
MiYanni: Just did some testing:
- This issue only happens on MSI installs. Only MSI installs write to a Workload log.
- It only happens if the workload integrity check has no actions. Meaning, it doesn't have any update required to the current packs for the current workloads.
- I can't debug this easily since I can only really test it on my VM.

From my VM:
image

It says "Checking the state of installed workloads..." which means it goes to run the transaction context and run the workload install command. Then, you see the series of dashes after that. It means that it executed the line after the transaction context command to install the workloads. What I think is happening is the file handle doesn't get closed by the time it goes to run the actual user-initiated workload command. Here's the contents of the log:

```
[2023-08-01 18:35:04.636] [00001854] === Logging started ===
[2023-08-01 18:35:04.645] [00001854] Executing: "C:\Program Files\dotnet\dotnet.exe" workload --info, PID: 6228, PPID: 10152
[2023-08-01 18:35:04.648] [00001854] IsElevated: False
[2023-08-01 18:35:04.648] [00001854] Is64BitProcess: True
[2023-08-01 18:35:04.648] [00001854] RebootPending: False
[2023-08-01 18:35:04.648] [00001854] ProcessorArchitecture: amd64
[2023-08-01 18:35:04.648] [00001854] HostArchitecture: x64
[2023-08-01 18:35:04.648] [00001854] SdkDirectory: 8.0.100-dev
[2023-08-01 18:35:04.648] [00001854] VerifySignatures: False
[2023-08-01 18:35:04.648] [00001854] SDK feature band: 8.0.100
[2023-08-01 18:35:04.730] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.WebAssembly.Sdk.Msi.x64\8.0.0-rc.1.23401.6\44747ed41fef19d2ba4cc28ac56d942a-x64.msi.
[2023-08-01 18:35:04.736] [00001854] DetectPackage: ProductCode: {60F864A3-1FA5-48FD-8CBB-E073F887C5CB}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.737] [00001854] PlanPackage: Begin, name: 44747ed41fef19d2ba4cc28ac56d942a-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.737] [00001854] PlanPackage: Completed, name: 44747ed41fef19d2ba4cc28ac56d942a-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.738] [00001854] Dependent already exists, Microsoft.NET.Runtime.WebAssembly.Sdk,8.0.0-rc.1.23401.6,x64 won't be modified.
[2023-08-01 18:35:04.738] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NETCore.App.Runtime.Mono.browser-wasm.Msi.x64\8.0.0-rc.1.23401.6\4eb8bfe2e6d1d2837dd6a9d407ca4172-x64.msi.
[2023-08-01 18:35:04.740] [00001854] DetectPackage: ProductCode: {BEFC0286-8A8C-4E38-8A2C-CE93B9469588}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.740] [00001854] PlanPackage: Begin, name: 4eb8bfe2e6d1d2837dd6a9d407ca4172-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.740] [00001854] PlanPackage: Completed, name: 4eb8bfe2e6d1d2837dd6a9d407ca4172-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.740] [00001854] Dependent already exists, Microsoft.NETCore.App.Runtime.Mono.browser-wasm,8.0.0-rc.1.23401.6,x64 won't be modified.
[2023-08-01 18:35:04.740] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NETCore.App.Runtime.AOT.win-x64.Cross.browser-wasm.Msi.x64\8.0.0-rc.1.23401.6\6f11f8e1278b4606cdea4d37f03c75f8-x64.msi.
[2023-08-01 18:35:04.741] [00001854] DetectPackage: ProductCode: {09608F57-AD19-49BC-B82B-14FCD1AB9738}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.741] [00001854] PlanPackage: Begin, name: 6f11f8e1278b4606cdea4d37f03c75f8-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.741] [00001854] PlanPackage: Completed, name: 6f11f8e1278b4606cdea4d37f03c75f8-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.741] [00001854] Dependent already exists, Microsoft.NETCore.App.Runtime.AOT.win-x64.Cross.browser-wasm,8.0.0-rc.1.23401.6,x64 won't be modified.
[2023-08-01 18:35:04.741] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.MonoAOTCompiler.Task.Msi.x64\8.0.0-rc.1.23401.6\1ebcf0bca573aa00669797be2ceae2fb-x64.msi.
[2023-08-01 18:35:04.741] [00001854] DetectPackage: ProductCode: {E274BD4E-CBF0-4C73-A9B5-5D45B761035E}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.741] [00001854] PlanPackage: Begin, name: 1ebcf0bca573aa00669797be2ceae2fb-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.741] [00001854] PlanPackage: Completed, name: 1ebcf0bca573aa00669797be2ceae2fb-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.741] [00001854] Dependent already exists, Microsoft.NET.Runtime.MonoAOTCompiler.Task,8.0.0-rc.1.23401.6,x64 won't be modified.
[2023-08-01 18:35:04.742] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.MonoTargets.Sdk.Msi.x64\8.0.0-rc.1.23401.6\0381cfbd02545a8408ba755236227998-x64.msi.
[2023-08-01 18:35:04.742] [00001854] DetectPackage: ProductCode: {1497849D-7D6F-4A8E-BE3C-1B20C77F2803}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.742] [00001854] PlanPackage: Begin, name: 0381cfbd02545a8408ba755236227998-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.742] [00001854] PlanPackage: Completed, name: 0381cfbd02545a8408ba755236227998-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.742] [00001854] Dependent already exists, Microsoft.NET.Runtime.MonoTargets.Sdk,8.0.0-rc.1.23401.6,x64 won't be modified.
[2023-08-01 18:35:04.743] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.Emscripten.3.1.34.Node.win-x64.Msi.x64\8.0.0-rc.1.23401.2\10985655e1321d8d25fc1a33c6e2b778-x64.msi.
[2023-08-01 18:35:04.743] [00001854] DetectPackage: ProductCode: {CEB643FB-D959-4989-8C6F-437F403969B9}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.743] [00001854] PlanPackage: Begin, name: 10985655e1321d8d25fc1a33c6e2b778-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.743] [00001854] PlanPackage: Completed, name: 10985655e1321d8d25fc1a33c6e2b778-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.743] [00001854] Dependent already exists, Microsoft.NET.Runtime.Emscripten.3.1.34.Node.win-x64,8.0.0-rc.1.23401.2,x64 won't be modified.
[2023-08-01 18:35:04.744] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.Emscripten.3.1.34.Python.win-x64.Msi.x64\8.0.0-rc.1.23401.2\051c42dcc98abb694951425546aa48fb-x64.msi.
[2023-08-01 18:35:04.744] [00001854] DetectPackage: ProductCode: {8318E480-F90B-46AB-AC29-716CD2173E1D}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.744] [00001854] PlanPackage: Begin, name: 051c42dcc98abb694951425546aa48fb-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.744] [00001854] PlanPackage: Completed, name: 051c42dcc98abb694951425546aa48fb-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.744] [00001854] Dependent already exists, Microsoft.NET.Runtime.Emscripten.3.1.34.Python.win-x64,8.0.0-rc.1.23401.2,x64 won't be modified.
[2023-08-01 18:35:04.744] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.Emscripten.3.1.34.Cache.win-x64.Msi.x64\8.0.0-rc.1.23401.2\78f803e1a4705194e08195e66582d3a1-x64.msi.
[2023-08-01 18:35:04.745] [00001854] DetectPackage: ProductCode: {62F606A8-A5F7-4EFE-BACC-19F6CEDC44A8}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.745] [00001854] PlanPackage: Begin, name: 78f803e1a4705194e08195e66582d3a1-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.745] [00001854] PlanPackage: Completed, name: 78f803e1a4705194e08195e66582d3a1-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.745] [00001854] Dependent already exists, Microsoft.NET.Runtime.Emscripten.3.1.34.Cache.win-x64,8.0.0-rc.1.23401.2,x64 won't be modified.
[2023-08-01 18:35:04.745] [00001854] Skipping signature verification for C:\ProgramData\dotnet\workloads\Microsoft.NET.Runtime.Emscripten.3.1.34.Sdk.win-x64.Msi.x64\8.0.0-rc.1.23401.2\8fcf89b4af176276e3b857689815bf17-x64.msi.
[2023-08-01 18:35:04.746] [00001854] DetectPackage: ProductCode: {464A062D-166A-4356-9B91-4AE27795E25B}, version: 8.0.0.0, state: Present.
[2023-08-01 18:35:04.746] [00001854] PlanPackage: Begin, name: 8fcf89b4af176276e3b857689815bf17-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install.
[2023-08-01 18:35:04.746] [00001854] PlanPackage: Completed, name: 8fcf89b4af176276e3b857689815bf17-x64.msi, version: 8.0.0.0, state: Present, installed version: 8.0.0.0, requested: Install, planned: None.
[2023-08-01 18:35:04.746] [00001854] Dependent already exists, Microsoft.NET.Runtime.Emscripten.3.1.34.Sdk.win-x64,8.0.0-rc.1.23401.2,x64 won't be modified.
[2023-08-01 18:35:04.810] [00001854] Shutting down
[2023-08-01 18:35:04.810] [00001854] Shutdown completed.
[2023-08-01 18:35:04.810] [00001854] Restart required: False
[2023-08-01 18:35:04.810] [00001854] === Logging ended ===
```

You can see that it starts up and shuts down in under 200 ms. So, I think that is the problem. However, I have no idea how to solve that easily.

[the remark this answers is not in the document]
MiYanni: Looked over the situation with Jacques. He recommended that I simply apply the milliseconds to the log name. I did that and the issue is effectively worked around. Looks like this now when it tries to open the same log within the same second (like the error above suggests):

image

I think this is sufficient. If we get any reports with the file being locked, we can look into a better solution. At least we won't have a command restriction now for running the integrity check.

src/Cli/dotnet/commands/dotnet-workload/WorkloadCommandBase.cs
joeloff: how is this unused? It's used on line 137

MiYanni: I think I flubbed in my comment here. I was likely thinking of something else I reverted. The value is used, but passed in now as a parameter. Line 129 and then the same assignment here in line 127.

src/Cli/dotnet/commands/dotnet-workload/WorkloadIntegrityChecker.cs
MiYanni: I think maybe doing a separate endeavor (different PR) to clean up the workload logging would be appropriate. I've already pointed out in one of the workload meeting chats about how when workloads are already up-to-date, we output 3 lines per pack, which seems a bit noisy.

image

I'll make a separate issue for this work.

MiYanni: Created issue to handle this work: https://github.com/dotnet/sdk/issues/34417

dsplaisted: Reducing the spew for updates would be great, but for the first run check, I think it's best if we print out nothing if everything is up to date, not even a "Checking the state of installed workloads..." message.  Then if there are updates needed, we could print something like "Installing updated workload packs".

To do so, we could add a `HasMissingWorkloadPacks` method to `IInstaller`, and do a bit of refactoring to the installer implementations so that they share code between the new method and the `InstallWorkloads` method.

dsplaisted (commented):
I think it's probably feasible to write some automated tests for this.  It may not be the simplest thing, but could be better than having to test everything manually.

In the automated tests we could probably test WorkloadIntegrityChecker directly, instead of also worrying about the first use sentinel.  It would need to support passing in mocked versions of the various classes it uses.  Other workload commands such as `WorkloadUpdateCommand` and the corresponding `GivenDotnetWorkloadUpdate` tests show how this can be done.

src/Cli/dotnet/Program.cs
```diff
@@ -182,11 +179,18 @@ internal static int ProcessArgs(string[] args, TimeSpan startupTime, ITelemetry
                         isDotnetBeingInvokedFromNativeInstaller = true;
                     }
 
+                    // Override the workload integrity check if the command is being ran in CI.
+                    if (new CIEnvironmentDetectorForTelemetry().IsCIEnvironment())
```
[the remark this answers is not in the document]
MiYanni: I've set it so that the environment variable (if present) is honored first. If the environment variable is not present, the default value is set to `true` (meaning, skip the check) if running in CI. Otherwise, false.

MiYanni:
> I think it's probably feasible to write some automated tests for this. It may not be the simplest thing, but could be better than having to test everything manually.
> 
> In the automated tests we could probably test WorkloadIntegrityChecker directly, instead of also worrying about the first use sentinel. It would need to support passing in mocked versions of the various classes it uses. Other workload commands such as `WorkloadUpdateCommand` and the corresponding `GivenDotnetWorkloadUpdate` tests show how this can be done.

I talked with Marc and one approach that seems reasonable is to update the test scenarios we provide to our vendor for workloads to include a test for this first-run functionality. When I was manually testing, one of the main scenarios was for MSI-based installs which we don't have unit tests for currently.

I looked into both `GivenDotnetWorkloadUpdate.cs` and `GivenFileBasedWorkloadInstall.cs`. These tests cover the constituent parts of what the `RunFirstUseCheck` in `WorkloadIntegrityChecker` does. Meaning, I've not created 'new' functionality that needs tested separately. In those tests, they construct a real `FileBasedInstaller` but provide mocks of certain parts, such as `INuGetPackageDownloader` and `IWorkloadManifestProvider`. I've underlined the real components in green (`#00FF00`) and the mocked components in orange (`#FFA500`):

image

The intent of the `RunFirstUseCheck` was to just make the minimum amount of code to call `InstallWorkloads` which defaults as many of those components that make sense. So, all of those parts *should* have tests already. The tests above the code in the screenshot just call `InstallWorkloads` with a canned set of values.

The other call that is made in `RunFirstUseCheck` is `GetInstalledWorkloads` which also has a series of tests for it across a few different parts, such as `WorkloadsInfoProviderTests` and the `dotnet workload list` tests.

src/Cli/dotnet/commands/dotnet-workload/WorkloadIntegrityChecker.cs
MiYanni: I believe that effort would go into the https://github.com/dotnet/sdk/issues/34417 issue since it requires extracting out the logic for handling workload installation/messaging as to not be within the command logic directly. Right now, everything is tightly coupled to the concept of running as a CLI command.

That kind of 'tailored' experience could be unit tested when it is refactored, but I'm not planning on refactoring workload installation logic to simply update the text displayed for the initial implementation of this feature. Upon refactoring, we could add some tests within `GivenADotnetFirstTimeUseConfigurer.cs` and `GivenADotnetFirstTimeUseConfigurerWithStateSetup.cs` to account for the appropriate messaging.

src/Cli/dotnet/commands/dotnet-workload/LocalizableStrings.resx
```diff
@@ -153,4 +153,11 @@
   <data name="WorkloadInfoDescription" xml:space="preserve">
     <value>Display information about installed workloads.</value>
   </data>
-</root>
+  <data name="WorkloadIntegrityCheck" xml:space="preserve">
+    <value>Checking the state of installed workloads...</value>
+  </data>
+  <data name="WorkloadIntegrityCheckError" xml:space="preserve">
+    <value>An issue was encountered when verifying workloads. For more information, run "dotnet workload update".</value>
```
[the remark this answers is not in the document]
MiYanni: Certainly there will be scenarios in which `dotnet workload update` won't show the exact same issue encountered in first-run. However, the point of the first-run check is to validate that the workloads are in a good state. If it fails, it shouldn't impede a user's command. If they see this message in the first command they run and the first command is, say, `dotnet build` and their build fails, it gives them another avenue to investigate. `dotnet workload update` *could* fix their scenario, potentially. This first-run check is just a ~band-aid~ knee-pad to help avoid *potential* pain for the user.

One idea is to have error codes for the CLI that go to articles about potential reasons for the errors and possible workarounds. Not sure if we have that many CLI-specific problems were we need that kind of error system, though.

src/Cli/dotnet/commands/dotnet-workload/WorkloadIntegrityChecker.cs
joeloff: I'd be ok with a separate cleanup PR since it will likely need to touch code that's unrelated to the first run experience and possibly have some test fallout as well.

src/Cli/dotnet/Program.cs
```diff
@@ -162,14 +162,13 @@ internal static int ProcessArgs(string[] args, TimeSpan startupTime, ITelemetry
...
-                        environmentProvider.GetEnvironmentVariableAsBool("DOTNET_GENERATE_ASPNET_CERTIFICATE", defaultValue: true);
-                    bool telemetryOptout =
-                      environmentProvider.GetEnvironmentVariableAsBool(EnvironmentVariableNames.TELEMETRY_OPTOUT, defaultValue: CompileOptions.TelemetryOptOutDefault);
-                    bool addGlobalToolsToPath =
-                        environmentProvider.GetEnvironmentVariableAsBool("DOTNET_ADD_GLOBAL_TOOLS_TO_PATH", defaultValue: true);
-                    bool nologo =
-                        environmentProvider.GetEnvironmentVariableAsBool("DOTNET_NOLOGO", defaultValue: false);
+                    bool generateAspNetCertificate = environmentProvider.GetEnvironmentVariableAsBool("DOTNET_GENERATE_ASPNET_CERTIFICATE", defaultValue: true);
+                    bool telemetryOptout = environmentProvider.GetEnvironmentVariableAsBool(EnvironmentVariableNames.TELEMETRY_OPTOUT, defaultValue: CompileOptions.TelemetryOptOutDefault);
+                    bool addGlobalToolsToPath = environmentProvider.GetEnvironmentVariableAsBool("DOTNET_ADD_GLOBAL_TOOLS_TO_PATH", defaultValue: true);
```
joeloff, line 167: Do we want to add these values to the EnvironmentVariableNames class so we don't have the strings everywhere?

src/Cli/dotnet/commands/dotnet-workload/LocalizableStrings.resx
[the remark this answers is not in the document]
joeloff: Should we remove the "when" or replace with "while". The string reads odd to me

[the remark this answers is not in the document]
joeloff: And can you lock the `dotnet workload update` string? You can inlcude a comment in the resx, like `{Locked="dotnet workload update"}`

joeloff (approved):
LGTM, question though, do we need to retarget against release/8.0.1xx?

src/Cli/dotnet/commands/dotnet-workload/LocalizableStrings.resx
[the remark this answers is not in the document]
MiYanni: I've removed the `when` and added the formatted `{Locked}` comment to the resource.

src/Cli/dotnet/Program.cs
MiYanni: I've added these values to `EnvironmentVariableNames` and referenced them here.

MiYanni:
> LGTM, question though, do we need to retarget against release/8.0.1xx?

Talked to Marc and he said to also put it into 8.0.1xx. We're trying to see if we can get that backport action working.

PR #34388, opened 2023-08-02
Add implementation and test for the DisableRuntimeMarshalling attribute.

jlaanstra:
Add DisableRuntimeMarshalling assembly attribute and test coverage. This enables an easy way to disable runtime marshalling for all projects in a solution via MSBuild and fits in with the general pattern of setting assembly attributes via the project file.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
@@ -127,8 +127,6 @@ Copyright (c) .NET Foundation. All rights reserved.
       </AssemblyAttribute>
       <AssemblyAttribute Include="System.Runtime.Versioning.RequiresPreviewFeaturesAttribute" Condition="'$(EnablePreviewFeatures)' == 'true' and '$(GenerateRequiresPreviewFeaturesAttribute)' == 'true'">
       </AssemblyAttribute>
-      <AssemblyAttribute Include="System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute" Condition="'$(DisableRuntimeMarshalling)' == 'true' and '$(GenerateDisableRuntimeMarshallingAttribute)' == 'true'">
-      </AssemblyAttribute>
     </ItemGroup>
 
     <ItemGroup Condition="'$(GenerateTargetPlatformAttribute)' == 'true'
@@ -153,6 +151,15 @@ Copyright (c) .NET Foundation. All rights reserved.
       <AssemblyAttribute Include="System.Runtime.Versioning.SupportedOSPlatformAttribute"
                          Condition="$([MSBuild]::VersionEquals($(SupportedOSPlatformVersion), '0.0'))" >
         <_Parameter1>$(TargetPlatformIdentifier)</_Parameter1>
+      </AssemblyAttribute>
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(GenerateDisableRuntimeMarshallingAttribute)' == 'true'
+                         and '$(DisableRuntimeMarshalling)' == 'true'
+                         and '$(TargetPlatformIdentifier)' != ''
+                         and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                         and $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '7.0'))">
+      <AssemblyAttribute Include="System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute">
       </AssemblyAttribute>
     </ItemGroup>
   </Target>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
@@ -156,7 +156,6 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <ItemGroup Condition="'$(GenerateDisableRuntimeMarshallingAttribute)' == 'true'
                          and '$(DisableRuntimeMarshalling)' == 'true'
-                         and '$(TargetPlatformIdentifier)' != ''
                          and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
                          and $([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '7.0'))">
       <AssemblyAttribute Include="System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute">
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
@@ -41,6 +41,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <GenerateRequiresPreviewFeaturesAttribute Condition="'$(GenerateRequiresPreviewFeaturesAttribute)' == '' and '$(IsNetCoreAppTargetingLatestTFM)' == 'true'">true</GenerateRequiresPreviewFeaturesAttribute>
     <GenerateTargetPlatformAttribute Condition="'$(GenerateTargetPlatformAttribute)' == ''">true</GenerateTargetPlatformAttribute>
     <GenerateSupportedOSPlatformAttribute Condition="'$(GenerateSupportedOSPlatformAttribute)' == ''">true</GenerateSupportedOSPlatformAttribute>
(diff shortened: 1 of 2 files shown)


AaronRobinsonMSFT:
Thanks @jlaanstra. Do we have a location for these MSBuild properties to be documented?

jlaanstra:
@AaronRobinsonMSFT we do here https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#runtime-configuration-properties

jlaanstra:
@AaronRobinsonMSFT Done for the docs PR. I left some todos to be filled out. The test seems to be unhappy, possibly AssemblyInfo.Get isn't reading this attribute. Seems like there is no test coverage for the similar EnablePreviewFeatures attribute?

AaronRobinsonMSFT:
Unsure on how the SDK repo works actually. Perhaps @dsplaisted or @marcpopMSFT can provide some insight into the issue.

jkoritzinsky:
Once this PR is in, we should file a follow-up issue in dotnet/runtime to update our "Add DisabledRuntimeMarshalling attribute" code fixer to set the project property (if we can).

marcpopMSFT:
> Unsure on how the SDK repo works actually. Perhaps @dsplaisted or @marcpopMSFT can provide some insight into the issue.
Does anyone really?

More seriously, the test is setting a bunch of assembly attributes and then opting out of them one by one. This new attribute only shows up as a disablement. So probably a similar but separate test would make more sense for this as it's not similar behavior to the other attributes in the same time.

jlaanstra:
@marcpopMSFT Added separate tests. @AaronRobinsonMSFT do we need to account for targeting TFM that doesn't have this attribute?

AaronRobinsonMSFT:
> do we need to account for targeting TFM that doesn't have this attribute?

Yes. If the targeted TFM doesn't have the attribute, the `GenerateDisableRuntimeMarshallingAttribute` should be set to `false`. I don't work much on the SDK, but I think that is the pattern for these properties. @marcpopMSFT Is there a downlevel test to check this?

@jkoritzinsky also has experience in this domain so perhaps he can shed some light here.

jlaanstra:
> 

Looks like there is a pre-existing pattern for this that I can easily use. Just need to know the minimum required TFM.

jkoritzinsky:
The minimum TFM is net7.0.

jlaanstra:
Done. Should be good to go now.

AaronRobinsonMSFT (commented):
LGTM. I defer to the SDK team for final sign off.

/cc @marcpopMSFT @dsplaisted

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -464,6 +464,104 @@ public void It_doesnt_includes_requires_preview_features()
...
+        [Theory]
+        [InlineData(true, true, "net6.0", false)]
+        [InlineData(true, true, "net7.0", true)]
+        [InlineData(true, true, ToolsetInfo.CurrentTargetFramework, true)]
+        [InlineData(true, false, ToolsetInfo.CurrentTargetFramework, true)]
+        [InlineData(false, false, ToolsetInfo.CurrentTargetFramework, true)]
+        public void TestDisableRuntimeMarshalling(bool disableRuntimeMarshalling, bool generateDisableRuntimeMarshallingAttribute, string targetFramework, bool shouldHaveAttribute)
+        {
+            var testAsset = _testAssetsManager
+                .CopyTestAsset("HelloWorld", identifier: $"{disableRuntimeMarshalling}${generateDisableRuntimeMarshallingAttribute}${targetFramework}")
```
dsplaisted, line 476: I'd suggest using the `TestProject` class here instead of `CopyTestAsset`.  That lets you set various properties for the project to be created, which means you usually don't have to manipulate the XML with `WithProjectChanges`.  Here's an example of how this is done: https://github.com/dotnet/sdk/blob/f051b536cc12190488231f3a889df44214c1bc2e/src/Tests/Microsoft.NET.Build.Tests/AppHostTests.cs#L354-L365

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -464,6 +464,104 @@ public void It_doesnt_includes_requires_preview_features()
             Assert.False(contains);
         }
 
+        [Theory]
+        [InlineData(true, true, "net6.0", false)]
+        [InlineData(true, true, "net7.0", true)]
+        [InlineData(true, true, ToolsetInfo.CurrentTargetFramework, true)]
+        [InlineData(true, false, ToolsetInfo.CurrentTargetFramework, true)]
+        [InlineData(false, false, ToolsetInfo.CurrentTargetFramework, true)]
+        public void TestDisableRuntimeMarshalling(bool disableRuntimeMarshalling, bool generateDisableRuntimeMarshallingAttribute, string targetFramework, bool shouldHaveAttribute)
```
dsplaisted, line 473: In most cases the `GenerateDisableRuntimeMarshallingAttribute` property won't be explicitly set, so I think it would be clearer to use a nullable bool value for the parameter, and to leave it null when the test project shouldn't explicitly set it.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
```diff
@@ -152,6 +153,15 @@ Copyright (c) .NET Foundation. All rights reserved.
         <_Parameter1>$(TargetPlatformIdentifier)</_Parameter1>
       </AssemblyAttribute>
     </ItemGroup>
+
+    <ItemGroup Condition="'$(GenerateDisableRuntimeMarshallingAttribute)' == 'true'
+                         and '$(DisableRuntimeMarshalling)' == 'true'
+                         and '$(TargetPlatformIdentifier)' != ''
```
dsplaisted, line 159: Is the `TargetPlatformIdentifier` clause supposed to be in here?  It seems like that means the attribute won't be written on `net7.0` for example, the TargetFramework would need to be something like `net7.0-windows` for it to be written.

There are some test failures in CI, maybe this is why they are failing.

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -464,6 +464,104 @@ public void It_doesnt_includes_requires_preview_features()
...
+                {
+                    // The assembly level attribute is generated only for .NET 7 and newer
+                    Assert.False(contains);
+                }
+            }
+
+            if (!generateDisableRuntimeMarshallingAttribute)
+            {
+                Assert.False(contains);
+            }
```
dsplaisted, line 527: I think you express whether the attribute should be generated in the theory parameters, so you can simplify the logic by only checking that value to see what you should assert.

```suggestion
            if (shouldHaveAttribute)
            {
                Assert.True(contains);
            }
            else
            {
                // The assembly level attribute is generated only for .NET 7 and newer
                Assert.False(contains);
            }
```

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -464,6 +464,104 @@ public void It_doesnt_includes_requires_preview_features()
...
+        }
+
+        [Fact]
+        public void It_doesnt_includes_disable_runtime_marshalling()
```
dsplaisted, line 531: I think if you switch the previous test theory to use `bool?` for the parameter types, then this scenario can be expressed as an instance of that theory (ie `false, null, "net7.0", false`).

dsplaisted (commented):
Looks good in general.

I had some suggestions for the tests, and the test failures need to be investigated and fixed.

jlaanstra:
@dsplaisted should look much better now, thanks for the suggestions!

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs
```diff
@@ -464,6 +464,68 @@ public void It_doesnt_includes_requires_preview_features()
...
+                TargetFrameworks = targetFramework,
+                IsExe = true,
+            };
+            testProject.AdditionalProperties["DisableRuntimeMarshalling"] = disableRuntimeMarshalling.ToString();
+            if (generateDisableRuntimeMarshallingAttribute.HasValue)
+            {
+                testProject.AdditionalProperties["GenerateDisableRuntimeMarshallingAttribute"] = generateDisableRuntimeMarshallingAttribute.Value.ToString();
+            }
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProject);
```
dsplaisted, line 497: This should make each instance of the theory use a different test execution folder, and fix the test failures in CI:

```suggestion
            var testAsset = _testAssetsManager.CreateTestProject(testProject, identifier: disableRuntimeMarshalling.ToString() + "_" + generateDisableRuntimeMarshallingAttribute + "_" + targetFramework);
```

dsplaisted (approved):

PR #34390, opened 2023-08-02
Ensure GenAPI replaces init method body

ericstj:
I noticed that we weren't emitting syntax that would compile for init properties.


ViktorHofer (approved):

carlossanlop (approved):

andriipatsula (approved):
Thanks

PR #34412, opened 2023-08-02
Add feature switch to control XmlResolver.IsNetworkingEnabledByDefault

### LINKED ISSUE #89877: Add an msbuild property that toggles the `XmlResolver.IsNetworkingEnabledByDefault` feature switch
> Reopening to track SDK and docs work (feel free to close if this is tracked elsewhere and sorry for the noise in that case). This would be hard/impossible to use without that:
>
>* We should add a property people can use to opt out. The necessary change on the SDK side will look something like this: https://github.com/dotnet/sdk/blob/7059317381a130ea7ee5e48a76b5c28a7bb7d919/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L490-L493
>* We should document the new property. That looks like this: https://github.com/dotnet/docs/pull/36080

_Originally posted by @MichalStrehovsky in https://github.com/dotnet/runtime/issues/83495#issuecomment-1641460783_

MichalStrehovsky:
Fixes dotnet/runtime#89877.

### DIFF
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -26,7 +26,7 @@
     <CustomResourceTypesSupport>false</CustomResourceTypesSupport>
     <UseSystemResourceKeys>true</UseSystemResourceKeys>
     <DynamicCodeSupport>true</DynamicCodeSupport>
-    <XmlResolverNetworkingSupport>false</XmlResolverNetworkingSupport>
+    <XmlResolverIsNetworkingEnabledByDefault>false</XmlResolverIsNetworkingEnabledByDefault>
     <JsonSerializerIsReflectionEnabledByDefault>false</JsonSerializerIsReflectionEnabledByDefault>
     <BuiltInComInteropSupport>false</BuiltInComInteropSupport>
     <_EnableConsumingManagedCodeFromNativeHosting>false</_EnableConsumingManagedCodeFromNativeHosting>
--- a/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
+++ b/src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
@@ -26,6 +26,7 @@
     <CustomResourceTypesSupport>false</CustomResourceTypesSupport>
     <UseSystemResourceKeys>true</UseSystemResourceKeys>
     <DynamicCodeSupport>true</DynamicCodeSupport>
+    <XmlResolverNetworkingSupport>false</XmlResolverNetworkingSupport>
     <JsonSerializerIsReflectionEnabledByDefault>false</JsonSerializerIsReflectionEnabledByDefault>
     <BuiltInComInteropSupport>false</BuiltInComInteropSupport>
     <_EnableConsumingManagedCodeFromNativeHosting>false</_EnableConsumingManagedCodeFromNativeHosting>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -626,6 +626,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RuntimeHostConfigurationOption Include="System.Threading.ThreadPool.MaxThreads"
                                     Condition="'$(ThreadPoolMaxThreads)' != ''"
                                     Value="$(ThreadPoolMaxThreads)" />
+
+    <RuntimeHostConfigurationOption Include="System.Xml.XmlResolver.IsNetworkingEnabledByDefault"
+                                    Condition="'$(XmlResolverNetworkingSupport)' != ''"
+                                    Value="$(XmlResolverNetworkingSupport)"
+                                    Trim="true" />
   </ItemGroup>
 
   <!--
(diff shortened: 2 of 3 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -626,6 +626,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RuntimeHostConfigurationOption Include="System.Threading.ThreadPool.MaxThreads"
                                     Condition="'$(ThreadPoolMaxThreads)' != ''"
                                     Value="$(ThreadPoolMaxThreads)" />
+
+    <RuntimeHostConfigurationOption Include="System.Xml.XmlResolver.IsNetworkingEnabledByDefault"
```
eiriktsarpalis, line 630: The STJ property controlling the `System.Text.Json.JsonSerializer.IsReflectionEnabledByDefault` property was named `JsonSerializerIsReflectionEnabledByDefault`. Echoing that naming convention, shouldn't this also be named `XmlResolverIsNetworkingEnabledByDefault`?

Typically this needs to go through API review but maybe we can fast-track this. cc @eerhardt @bartonjs @terrajobst

MichalStrehovsky: We've not done API review on these or they would be consistent.

The `XXXSupport` is a prevalent theme for things that disable support for something

`DebuggerSupport`, `EventSourceSupport`, `HttpActivityPropagationSupport`, `MetadataUpdaterSupport`, `NullabilityInfoContextSupport`, `CustomResourceTypesSupport`, `DynamicCodeSupport`, `BuiltInComInteropSupport`,...

eiriktsarpalis (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
eerhardt: One difference between the existing XXXSupoort flags and the JsonSerializerIsReflectionEnabledByDefault flag is that the XXXSupport flags completely remove the ability for doing XXX. Where the JsonSerializer is still capable of doing reflection, it just won’t do it by default when the flag is set. Thus the verbose naming (it’s trying to be correct/precise).

I don’t know exactly which case the new XML flag falls into, but if the XmlResolver can still use networking even if the flag is set, I wouldn’t name it XXXSupport. Because it can still do it, even though the flag is set.

MichalStrehovsky: @Jozkee @krwq do you know which side `IsNetworkingEnabledByDefault` is on so we can finalize the name?

eiriktsarpalis: I'm not familiar with the code, but looking at [this change](https://github.com/dotnet/runtime/pull/84169/files#diff-965a740bf9db42c616b767e453dc3051d09e03dc2fe8cf797c5b7e38ef7f9cbfR626-R630) suggests that it falls on the same boat as the STJ switch. It picks the `XmlUrlResolver` implementation if the switch is turned on, which is a public class. The behavior of `XmlUrlResolver` itself doesn't seem to be affected by the feature switch.

src/Assets/TestProjects/KitchenSink/TestApp/TestApp.csproj
```diff
@@ -26,6 +26,7 @@
     <CustomResourceTypesSupport>false</CustomResourceTypesSupport>
     <UseSystemResourceKeys>true</UseSystemResourceKeys>
     <DynamicCodeSupport>true</DynamicCodeSupport>
+    <XmlResolverNetworkingSupport>false</XmlResolverNetworkingSupport>
```
MichalStrehovsky, line 29:
```suggestion
    <XmlResolverIsNetworkingEnabledByDefault>false</XmlResolverIsNetworkingEnabledByDefault>
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -626,6 +626,11 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+                                    Condition="'$(XmlResolverNetworkingSupport)' != ''"
+                                    Value="$(XmlResolverNetworkingSupport)"
```
MichalStrehovsky, line 632:
```suggestion
                                    Condition="'$(XmlResolverIsNetworkingEnabledByDefault)' != ''"
                                    Value="$(XmlResolverIsNetworkingEnabledByDefault)"
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
MichalStrehovsky: @eiriktsarpalis @eerhardt if this updated name looks good, could you please merge? I'm off work through end of August. It would be nice to have this in 8.0.

eiriktsarpalis: Thanks!

PR #34419, opened 2023-08-03
Allow PublishSingleFile and PublishAot specified simultaneously

MichalStrehovsky:
We made native AOT targets error out if these are specified at the same time due to bad failure mode. But running through this, it appears to be mostly harmless except for this task.

So proposing taking this and rolling back https://github.com/dotnet/runtime/pull/73995. But no hard feelings if people feel otherwise.

We might want to eventually set `PublishSingleFile` when `PublishAot` is set, same way we do for `PublishTrimmed`. Some of the SDK behaviors look relevant. Others don't. I don't have a good sense on this one.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
@@ -952,7 +952,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <UsingTask TaskName="GenerateBundle" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <Target Name="GenerateSingleFileBundle"
-          Condition="'$(PublishSingleFile)' == 'true'"
+          Condition="'$(PublishSingleFile)' == 'true' and '$(PublishAot)' != 'true'"
           DependsOnTargets="_ComputeFilesToBundle;PrepareForBundle;_GenerateSingleFileBundleInputCache"
           Inputs="@(FilesToBundle);$(_GenerateSingleFileBundlePropertyInputsCache)"
           Outputs="$(PublishedSingleFilePath)">


richlander:
From user standpoint, those two are intended to be leaf node / user facing outputs. With this proposal `PublishSingleFile` would be both a final output and a characteristic. That would be unfortunate. Instead, it would be better if any single file isms were shared with native aot within the appropriate tasks.

dsplaisted:
Looking at https://github.com/dotnet/runtime/pull/73995, I noticed that the errors generated in Microsoft.NETCore.Native.Publish.targets are not localized.  Ideally they would be, and possibly it would be good for them to have a NETSDK error code too.

MichalStrehovsky:
> From user standpoint, those two are intended to be leaf node / user facing outputs

Is your suggestion to keep erroring out?

People can end up with both in the csproj if they want single file on platforms that don't support native aot (maybe just windows x86 at this point). If it's an error, they need to both enable aot and disable single file.

FWIW, we do allow PublishTrimmed to be enabled for PublishAot (in fact, we enable it by default and forbid users from setting it to false). People do get a single file out of PublishAot too. Whether it's done by gluing assemblies to an existing EXE, or through native compilation feels like an implementation detail.

richlander:
Now that I think about it more, I don't care much. Do whatever is pragmatic.

agocke:
Change LGTM, just need a test

agocke:
Ah, I think I misunderstood the suggested flow here. Is the idea that we

1. Revert the change in runtime
2. Flow the change to the SDK here
3. Take this change and fix the tests at the same time?

Or where you intending to check this in first, to make it work, then do 1-2 later?

MichalStrehovsky:
I wanted to fix the problem first before removing the blocking in nativeaot targets so that we don't produce sdks that fail with inscrutable errors while waiting for the sdk fix. Getting prs approved often takes a long time.

agocke (approved):
LGTM, failures look unrelated

PR #34429, opened 2023-08-03
Add back _RequiresILLinkPack override

sbomer:
https://github.com/dotnet/sdk/pull/34077 removed the ability to download the ILLink pack by setting `_RequiresILLinkPack`.  This is needed for macios scenarios where the ILLink pack should be restored, but it's not known ahead-of-time whether trimming will be enabled. See https://github.com/xamarin/xamarin-macios/pull/17227 for more context.

This adds back the ability to set `_RequiresILLinkPack` to download the ILLink pack even if none of the other relevant settings are enabled.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -41,6 +41,8 @@ namespace Microsoft.NET.Build.Tasks
 
         public bool PublishAot { get; set; }
 
+        public bool RequiresILLinkPack { get; set; }
+
         public bool IsAotCompatible { get; set; }
 
         public bool EnableAotAnalyzer { get; set; }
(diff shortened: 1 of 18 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -38,6 +30,20 @@ Copyright (c) .NET Foundation. All rights reserved.
     </ResolvedFileToPublish>
   </ItemDefinitionGroup>
 
+  <Target Name="_ComputeRequiresILLinkPack" BeforeTargets="ProcessFrameworkReferences">
+    <!-- Keep this in sync with requiresILLinkPack in ProcessFrameworkReferences.cs. 
+         Must be set late enough to see the inferred value of EnableSingleFileAnalyzer. -->
```
rolfbjarne, line 35: I believe this comment is outdated now, since there's no duplicate logic in ProcessFrameworkReference anymore.

rolfbjarne (approved):
Thanks!

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -426,16 +428,11 @@ protected override void ExecuteCore()
...
             }
 
-            // Keep this in sync with _RequiresILLinkPack in Microsoft.NET.Publish.targets.
-            bool requiresILLinkPack = PublishAot
-                || IsAotCompatible || EnableAotAnalyzer
-                || PublishTrimmed
-                || IsTrimmable || EnableTrimAnalyzer
-                || EnableSingleFileAnalyzer;
-            if (requiresILLinkPack)
+            if (RequiresILLinkPack)
```
dsplaisted, line 431: Would it make sense to generate the errors outside of the `ProcessFrameworkReferences` task?  Then I tihnk we could remove all of these additional task parameters that were added.

dsplaisted (approved):

src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
dsplaisted: Thanks for looking at it.  If it seems like there's not an obvious way to simplify it well, then we can leave it as it is.

PR #34277, opened 2023-07-26
Revert "Allow ResolveReadyToRunCompilers to be run under FreeBSD"

jkotas:
Reverts dotnet/sdk#31581

Fixes https://github.com/dotnet/source-build/issues/3568

PR #31581, opened 2023-04-03
Allow ResolveReadyToRunCompilers to be run under FreeBSD

sec:
This should enable almost patchless build of runtime under FreeBSD, with R2R. @Thefrank

PR #34430, opened 2023-08-03
Rework of Allow ResolveReadyToRunCompilers to be run under FreeBSD

Thefrank:
Rework of https://github.com/dotnet/sdk/pull/31581 as it got reverted in https://github.com/dotnet/sdk/pull/34277 due to https://github.com/dotnet/source-build/issues/3568

CC: @jkotas @sec @mthalman

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
@@ -203,10 +203,6 @@ namespace Microsoft.NET.Build.Tasks
                 {
                     portablePlatform = "osx";
                 }
-                else if (RuntimeInformation.IsOSPlatform(OSPlatform.FreeBSD))
-                {
-                    portablePlatform = "freebsd";
-                }
             }
 
             targetOS = portablePlatform switch
@@ -370,9 +366,8 @@ namespace Microsoft.NET.Build.Tasks
             }
             else
             {
-                // Generic Unix-like: linux, freebsd, and others. Does not check for arch
-                _crossgenTool.ToolPath = Path.Combine(_crossgenTool.PackagePath, "tools", "crossgen");
-                _crossgenTool.ClrJitPath = Path.Combine(_crossgenTool.PackagePath, "runtimes", _targetRuntimeIdentifier, "native", "libclrjit.so");
+                // Unknown platform
+                return false;
             }
 
             return File.Exists(_crossgenTool.ToolPath) && File.Exists(_crossgenTool.ClrJitPath);
@@ -386,11 +381,6 @@ namespace Microsoft.NET.Build.Tasks
             {
                 toolFileName = "crossgen2.exe";
                 v5_clrJitFileNamePattern = "clrjit-{0}.dll";
-            }
-            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
-            {
-                toolFileName = "crossgen2";
-                v5_clrJitFileNamePattern = "libclrjit-{0}.so";
             }
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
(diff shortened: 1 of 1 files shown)


Thefrank:
This tries to be as low-impact as possible by explicitly adding freebsd items and allowing for "Generic Unix-like" case instead of failing out as a compromise for portability. 

Finally, runtime and sdk are quite out of sync on this file even without this change. How should this be handled?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
```diff
@@ -393,8 +399,9 @@ private bool GetCrossgen2ComponentsPaths(bool version5)
             }
             else
             {
-                // Unknown platform
-                return false;
+                // Generic Unix-like: linux, freebsd, and others. Does not check for arch
```
jkotas, line 402: Could you please delete `RuntimeInformation.IsOSPlatform(OSPlatform.Linux)` above?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
```diff
@@ -203,6 +203,10 @@ private bool GetCrossgen2TargetOS(out string targetOS)
                 {
                     portablePlatform = "osx";
                 }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.FreeBSD))
```
jkotas, line 206: This whole section is very confusing. It maps Windows to Linux, and Linux to Windows. It suggests that it is not really used. Is it really necessary to add FreeBSD here?

I think it would be best to handle `portablePlatform == null` as generic Unix.

Thefrank: It was added https://github.com/dotnet/sdk/commit/71e047e2adbbd8a714e4a3df9c82f08b86ba148c as a part of source-build so I assumed that it was used and the removal in the prior attempt to add in FreeBSD was the cause of the source-build (centos) issue. 

`OSPlatform` does contain `FreeBSD` but only for [netcore 3+](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.osplatform.freebsd?view=net-7.0#system-runtime-interopservices-osplatform-freebsd). This looks like it causes the build to run into the same issue when trying to add FreeBSD in other areas (like runtime).

Not sure how to approach the rest of this block. Would it be best to just have `portablePlatform = "unix";` for `portablePlatform == null && _targetRuntimeIdentifier == _hostRuntimeIdentifier` instead of the `if` statements? A different case for just `portablePlatform == null` -> `portablePlatform = "unix";` and leave the current `if` alone?

src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
```diff
@@ -381,20 +382,16 @@ private bool GetCrossgen2ComponentsPaths(bool version5)
...
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
                 toolFileName = "crossgen2";
                 v5_clrJitFileNamePattern = "libclrjit-{0}.dylib";
             }
             else
             {
-                // Unknown platform
-                return false;
+                // Generic Unix-like: linux, freebsd, and others. Does not check for arch
```
jkotas, line 392:
```suggestion
                // Generic Unix-like: linux, freebsd, and others.
```

src/Tasks/Microsoft.NET.Build.Tasks/ResolveReadyToRunCompilers.cs
jkotas: Is the PR in the current state fixing the source build problem for FreeBSD? If it does, it looks good to me.

I think there is opportunity to make this work for all Unix-like OSes out of the box, but it does not look easy. We can leave it for later.

Thefrank: Yes. As-is the issues are fixed for FreeBSD

jkotas (approved):
Thanks

PR #34433, opened 2023-08-03
Support Records in GenAPI

### LINKED ISSUE #32196: GenAPI cannot decode records when reading from metadata
### Describe the bug
GenAPI does not emit records correctly.

### To Reproduce
```c#
namepsace ns;
public record class Foo();
```
Compile this and point GenAPI to it.

Expect: Same output

Actual: 
```c#
namespace ns
{
    public record Foo
    {
        public Foo() { }
        protected Foo(Foo original) { }
        protected virtual System.Type EqualityContract { get { throw null; } }
        public virtual Foo <Clone>$() { throw null; }
        public virtual System.Boolean? Equals(Foo? other) { throw null; }
        public override System.Boolean? Equals(System.Object? obj) { throw null; }
        public override System.Int32 GetHashCode() { throw null; }
        public static System.Boolean? operator ==(Foo? left, Foo? right) { throw null; }
        public static System.Boolean? operator !=(Foo? left, Foo? right) { throw null; }
        protected virtual System.Boolean PrintMembers(System.Text.StringBuilder builder) { throw null; }
        public override System.String ToString() { throw null; }
    }
}
```

Related: https://github.com/dotnet/sdk/issues/32165

AFAICT there isn't anything preserved in metadata to indicate that a type is a record.  Perhaps we could use a heuristic -- look for all the members https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/records.

ericstj:
Fixes #32196 

Class records are mostly handled and regenerated with minimal compiler boilerplate.

Struct records show all the compiler generated boilerplate as I don't try to discern them from regular structs.  We could do more here - by recognizing all the compiler generated members and reducing, but it's not necessary for compat parity.


src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
```diff
@@ -154,16 +155,367 @@ public partial struct PublicStruct
...
+                """);
+        }
+
+        [Fact]
+        public void TestRecordDeclaration()
+        {
+            RunTest(original: """
+                namespace Foo
+                {
+
```
carlossanlop, line 213: Probably unimportant, but do we expect GenAPI to care about extra spaces? Both here and before the end of the namespace.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
```diff
@@ -131,6 +131,7 @@ record struct RecordStruct { }
                     public readonly ref struct PublicReadonlyRefStruct { }
                 }
                 """,
+                // It's expected that record structs are expanded - they're effectively syntactic sugar that is not persisted in metadata
```
carlossanlop, line 134: Do you mean the extra newline between the brackets?

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
```diff
@@ -154,16 +155,367 @@ public partial struct PublicStruct
                     {
                     }
 
-                    internal readonly record struct ReadonlyRecordStruct : System.IEquatable<ReadonlyRecordStruct>
+                    internal readonly partial struct ReadonlyRecordStruct : System.IEquatable<ReadonlyRecordStruct>
```
carlossanlop, line 158: I assume this is expected, but why are the compiler-generated APIs added in a separate partial type and not included in the same one above?

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
```diff
@@ -10,6 +10,10 @@
     <NoWarn>$(NoWarn);RS1024</NoWarn>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
+    <Compile Include="$(RepoRoot)src\Common\NullableAttributes.cs" LinkBase="Common" />
```
carlossanlop, line 14: For my own education, what's this file's purpose?

carlossanlop (approved):
LGTM, just have questions.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -174,6 +174,18 @@ private SyntaxNode Visit(SyntaxNode namedTypeNode, INamedTypeSymbol namedType)
...
+            if (namedType.IsRecord && namedType.TryGetRecordConstructor(out IMethodSymbol? recordConstructor))
+            {
+                // find the record parameters, so that we can exclude properties
+                var recordParameteterNames = recordConstructor.Parameters.Select(p => p.Name).ToHashSet();
+
+                // exclude compiler generated members, record constructor, and record properties
+                members = members.Where(member => !member.IsCompilerGenerated() &&
+                    !SymbolEqualityComparer.Default.Equals(member, recordConstructor) &&
+                    !recordParameteterNames.Contains(member.Name));
+            }
```
ViktorHofer, line 187: Shouldn't that be a filter similar to the ImplicitSymbolFilter: https://github.com/dotnet/sdk/blob/main/src/GenAPI/Microsoft.DotNet.GenAPI/Filtering/ImplicitSymbolFilter.cs?

src/GenAPI/Microsoft.DotNet.GenAPI/ISymbolExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// Determines if a symbol was generated by the compiler by checking for the presense of the CompilerGeneratedAttribute.
+        /// </summary>
+        /// <param name="symbol">Symbol to check</param>
+        /// <returns>True if the attribute was found.</returns>
+        public static bool IsCompilerGenerated(this ISymbol symbol)
+        {
+            return symbol.GetAttributes().Any(attribute =>
+                attribute?.AttributeClass?.Name == "CompilerGeneratedAttribute" &&
+                attribute.AttributeClass.ContainingNamespace.ToDisplayString().Equals("System.Runtime.CompilerServices", StringComparison.Ordinal));
+        }
```
ViktorHofer, line 21: I think this might make sense to move into ApiSymbolExtensions' SymbolExtensions. If you want to keep it here, please remove the `I` prefix from the class name.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
ericstj: I mean all the compiler generated members listed for these types.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
ericstj: I don't think so - the problem is we can't always reproduce the compiler syntax that will cause it to generate this API.  Good example is `struct record`.  Another example is a source generator that uses this attribute and exposes public API.

src/ApiCompat/Microsoft.DotNet.ApiCompatibility/Microsoft.DotNet.ApiCompatibility.csproj
ericstj: Nullable attributes used for using nullable features on older frameworks.

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
ericstj: This is what the `record` syntax produces.  For structs we cannot identify them, so we just reproduce the members the compiler produced.  See https://github.com/dotnet/roslyn/issues/69326

src/Tests/Microsoft.DotNet.GenAPI.Tests/CSharpFileBuilderTests.cs
ericstj: It doesn't this is just input that gets compiled to an assembly.  Gen API never sees it.  I'll clean up extra new lines.

src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
ericstj: I think I see what you mean.  I gave a try to refactoring it so that it's efficient enough to do for every member.  Have a look and see if you think that's better.

ViktorHofer (approved):
While this is slightly more costly now, I think it fits better into the existing filtering story. If we would like to mitigate the cost of recurring calls to `symbol.ContainingType.TryGetRecordConstructor`, we could add caching to the filter.

ericstj:
Looks like I broke things a bit when moving to the filter.  Fixing ...

ViktorHofer (approved):

ericstj:
If we wanted to better handle struct records we could implement our own heuristic for detecting them from metadata - essentially if we see the entire set of methods that could be replaced with `record` syntax - do so.  It's a bit more expensive than what we did here for classes but not impossible.

PR #34459, opened 2023-08-04
don't copy over the .dotnet dependencies if they are already there

marcpopMSFT:
I found that this sped up my local build from 55s to 47s


marcpopMSFT:
@baronfel your changes have resulted in a lot of test failures:
`No templates or subcommands found matching: 'classlib'.`

baronfel:
seems like! let's just revert that entirely and go with your thing. I might tinker more over the weekend to see where the changes need to happen.

baronfel (approved):

PR #34460, opened 2023-08-04
Add better error messages when we can't fetch manifests

baronfel:
Users often get confusing error messages when they use invalid container repositories/tags and/or cannot authenticate to those registries. This PR adds specific exceptions when we detect those cases and logs specific MSBuild errors (either through the Task directly or via the CLI expected error format) when these errors occur.

Current experience with an unknown repository:
image

After this change, with an unknown repository:
image

This should hopefully lead to more actionable errors for users - right now we can't tell at a glance if the issue is that we inferred an incorrect base image, or if the user typo'd, or any other failure scenario.

Similarly, authentication issues with the destination registries currently:
image

And after:
image


baronfel:
retargeted to 8.0.1xx now that the branch is open and main has advanced to .NET 9

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -57,12 +57,30 @@ public static class ContainerBuilder
...
+                return 1;
+            }
+            catch (UnableToAccessRepositoryException)
+            {
+                logger.LogError(Resource.FormatString(nameof(Strings.UnableToAccessRepository), baseImageName, registry.RegistryName));
+                return 1;
+            }
+            catch (ContainerHttpException e)
+            {
+                logger.LogError(e.Message);
```
rainersigwald, line 81:
```suggestion
                logger.LogErrorFromException(e);
```

? If you don't think the stack is important, or add `, showStackTrace: true` for more transparency (at the cost of ugliness).

rainersigwald: Oh wait, is this not guaranteed to be a concrete `TaskLoggerHelper`? Hm. Then maybe consider cloning `LogErrorFromException` to make sure it's actionable.

src/Containers/Microsoft.NET.Build.Containers/Exceptions/UnableToAccessRepositoryException.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Microsoft.NET.Build.Containers;
+
+internal sealed class UnableToAccessRepositoryException : Exception
+{
+    public UnableToAccessRepositoryException(string registry)
+        : base(string.Format("Unable to access the registry '{0}'. Please confirm your credentials are correct and that you have access to this registry.", registry))
+    {
+    }
+    public UnableToAccessRepositoryException(string registry, string repository)
+        : base(string.Format("Unable to access the repository '{0}' in the registry '{1}'. Please confirm your credentials are correct and that you have access to this repository and registry.", repository, registry))
```
rainersigwald, line 13: HTTP error useful to surface in this message?

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -159,6 +177,13 @@ internal async Task<bool> ExecuteAsync(CancellationToken cancellationToken)
                 GeneratedArchiveOutputPath = archive.ArchiveOutputPath;
             }
         }
+        catch (ContainerHttpException e)
+        {
+            if (BuildEngine != null)
```
rainersigwald, line 182: What's this condition about and why isn't it needed for the other error types?

rainersigwald (approved):

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
baronfel: Right, it's not a TLH here. This is also called from the VS codepath, where we already have some issues with error reporting, etc. None of this logging was in the canonical format before this PR, and it also doesn't flow through the VS Publish dialog correctly even if it is in the canonical format.

src/Containers/Microsoft.NET.Build.Containers/Exceptions/UnableToAccessRepositoryException.cs
baronfel: I don't think so - it's always stemming from a `401 Unauthorized`, and none of the repositories I tested with included anything meaningful in the response bodies for this error code. Cloud providers are stingy with auth failure details in bodies - they want you to go to their logging infrastructures to learn more to prevent leaking anything.

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
baronfel: This was a miss - this class is tested without a full engine/test harness and so it doesn't have the engine/logging infra. I'll add it here for consistency.

PR #34485, opened 2023-08-07
Add ProjectCapability for Native AOT properties.

melytc:
The `NativeAOTProperties` capability is being used in the project system side to control the visibility of the Native AOT properties introduced recently to .NET 8. The idea is that, in general, we want to set this capability to true so that we show these properties for projects targeting .NET 8 or higher. The exceptions at the moment would be class libraries, WPF and WinForms projects, which are filtered using the property condition. Other project types, like MAUI and Web projects, can disable this capability by setting the `ShowNativeAOTProperties` property to false in their props file.

Unit tests are on their way 🙂

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1263,11 +1263,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
   <PropertyGroup>
-    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
-                           and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
-                           and '$(_TargetFrameworkVersionWithoutV)' >= '8.0'">true</ShowNativeAOTProperties>
-    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
-                             and ('$(UseWPF)' == 'true' or '$(UseWindowsForms)' == 'true')">false</ShowNativeAOTProperties>
+    <ShowNativeAOTProperties Condition="('$(UseWPF)' == 'false' or '$(UseWPF)' == '')
+                             and ('$(UseWindowsForms)' == 'false' or '$(UseWindowsForms)' == '')
+                             and '$(ShowNativeAOTProperties)' == ''
+                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                             and '$(_TargetFrameworkVersionWithoutV)' >= '8.0'">true</ShowNativeAOTProperties>
   </PropertyGroup>
 
   <!-- Reference Manager capabilities -->
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1261,14 +1261,13 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
   </ItemGroup>
 
-  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for class libraries, WPF, WinForms projects -->
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
   <PropertyGroup>
     <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
                            and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
                            and '$(_TargetFrameworkVersionWithoutV)' >= '8.0'">true</ShowNativeAOTProperties>
-    <ShowNativeAOTProperties Condition="'$(UseWPF)' == 'true'
-                             or '$(UseWindowsForms)' == 'true'
-                             or '$(OutputType)' == 'Library'">false</ShowNativeAOTProperties>
+    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
+                             and ('$(UseWPF)' == 'true' or '$(UseWindowsForms)' == 'true')">false</ShowNativeAOTProperties>
   </PropertyGroup>
 
   <!-- Reference Manager capabilities -->
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1256,6 +1256,21 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="CrossPlatformExecutable" />
   </ItemGroup>
 
+  <!-- Native AOT compilation -->
+  <ItemGroup>
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
+  </ItemGroup>
+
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1256,6 +1256,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <ItemGroup>
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
+  </ItemGroup>
+
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for class libraries, WPF, WinForms projects -->
+  <PropertyGroup>
+    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
+                           and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                           and '$(_TargetFrameworkVersionWithoutV)' >= '8.0'">true</ShowNativeAOTProperties>
+    <ShowNativeAOTProperties Condition="'$(UseWPF)' == 'true'
```
eerhardt, line 1269: Do we want to check `'$(ShowNativeAOTProperties)' == ''`?

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1256,6 +1256,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+                             or '$(UseWindowsForms)' == 'true'
+                             or '$(OutputType)' == 'Library'">false</ShowNativeAOTProperties>
```
eerhardt, line 1271: For `'$(OutputType)' == 'Library'` I believe you can publish a library for AOT and it makes a self-contained native library. See https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net7#build-native-libraries

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
melytc: I'd say yes, we want to check if the property has not been set; so it defaults to true.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
melytc: You're right, removing on next commit as we do want it to be on for class libraries.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
drewnoakes: I think the logic here is that the second `ShowNativeAOTProperties` needs to be able to override the first, so checking `'$(ShowNativeAOTProperties)' == ''` on the second instance would break that.

eerhardt: > I think the logic here is that the second `ShowNativeAOTProperties` needs to be able to override the first, so checking `'$(ShowNativeAOTProperties)' == ''` on the second instance would break that.

Then flip the logic around so the `UseWPF` and `UseWindowsForms` checks happen first.

melytc: Got it, thanks for the clarification.

drewnoakes (commented):
Do we need the `ShowNativeAOTProperties` property here? Would it be enough to move the condition from that property to the `ProjectCapability` directly, to avoid having the property altogether? Or is the idea that other files may want to tweak that property during evaluation?

EDIT You explained this above:

> Other project types, like MAUI and Web projects, can disable this capability by setting the ShowNativeAOTProperties property to false in their props file.

drewnoakes (approved):

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1256,6 +1256,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   </ItemGroup>
 
+  <!-- Native AOT compilation -->
+  <ItemGroup>
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
+  </ItemGroup>
+
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
+  <PropertyGroup>
+    <ShowNativeAOTProperties Condition="('$(UseWPF)' == 'false' or '$(UseWPF)' == '')
```
marcpopMSFT, line 1266: you could probably do != true for both of these to simplify the logic some

melytc:
I pushed some changes I had, but I intend to work on this next week; thanks for the patience!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1256,6 +1256,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
+  </ItemGroup>
+
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
+  <PropertyGroup>
+    <ShowNativeAOTProperties Condition="'$(UseWPF)' != 'true'
+                             and '$(UseWindowsForms)' != 'true'
+                             and '$(ShowNativeAOTProperties)' == ''
+                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                             and '$(_TargetFrameworkVersionWithoutV)' >= '8.0'">true</ShowNativeAOTProperties>
```
eerhardt, line 1270:
```suggestion
                             and $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">true</ShowNativeAOTProperties>
```

Is an easier way to make this check.

eerhardt: what are user-defined framework aliases?

eerhardt: ok, if that's the case, then we will need to fix a few places across the stack that are already using this check.

https://github.com/dotnet/sdk/blob/c9330fb79615cfb6b6b9ab399a496944be62ea95/src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets#L97-L100

https://github.com/dotnet/sdk/blob/c9330fb79615cfb6b6b9ab399a496944be62ea95/src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets#L15

https://github.com/dotnet/runtime/blob/3c3f49958f93edba4e899f292a876bd47fdbeca6/src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets#L14-L21

eerhardt: cc @captainsafia @vlada-shubina @baronfel @rainersigwald @layomia

ViktorHofer: In addition to what was already mentioned by @KalleOlaviNiemitalo, evaluation performance is slightly better with the property comparison instead of calling into the `IsTargetFrameworkCompatible` msbuild function.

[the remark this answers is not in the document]
melytc: took this feedback on the latest PR: https://github.com/dotnet/sdk/pull/35710 thanks!

melytc:
Closing in favor of https://github.com/dotnet/sdk/pull/35710 which will be inserted into 8.0.1xx.

PR #34488, opened 2023-08-07
Exempt version from workload set

Forgind:
"Version" is included in the baseline manifest in the json, but it is not a workload ID. The SDK is parsing it as such.


Forgind:
For the error message, I'm thinking of just including the format, i.e., `(ID: version/feature band) `. What do you think?

The reason this adds clarity is that the ID was "version" in the wrong case, which made it start with `workload manifest version: <version>`, which made total sense to me but was indicating a pretty clear failure if you knew that wasn't what the string should be.

PR #34585, opened 2023-08-11
Update our TFM support and error messaging for 17.7

marcpopMSFT:
Add error when targeting net8.0 in 17.7 and remove net8.0 from the supported dropdown list.


rainersigwald (approved):

baronfel (approved):
LGTM per the previous iterations of this check. Conceptual question about the warning check, though - isn't a big problem _using_ the 8 SDK at all? Not just targeting the .NET 8 TFM? I'd expect e.g. the .NET 8 Workload Resolver to not work in any VS less than 17.7, even if a user was using that SDK to target lower levels.

So I wonder if the check we do here shouldn't be on _TargetFrameworkVersionWithoutV and instead on the SDK version itself?

marcpopMSFT:
> LGTM per the previous iterations of this check. Conceptual question about the warning check, though - isn't a big problem _using_ the 8 SDK at all? Not just targeting the .NET 8 TFM? I'd expect e.g. the .NET 8 Workload Resolver to not work in any VS less than 17.7, even if a user was using that SDK to target lower levels.
> 
> So I wonder if the check we do here shouldn't be on _TargetFrameworkVersionWithoutV and instead on the SDK version itself?
Because of the lag in pipelines/actions upgrading their VS versions on day 0, we've made an effort to try to enable this in the past couple of releases. Specifically, our guidance is that we don't support targeting 8 but we do support targeting downlevel. You are correct though that occasionally things slip through or get broken that we missed so it's best if you don't (hence why we update the minimum msbuild in the .2xx versions). Here's the official guidance.
https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs#targeting-and-support-rules

Make sense?

PR #34587, opened 2023-08-12
Fix redist.csproj build failure on SLN rebuild in VS

MiYanni:
## Summary
Currently, the SDK repo will build properly (all projects succeed) when using `build.cmd`. However, if you open the `sdk.sln` and hit *Rebuild Solution*, the build will have 1 project that fails. That project is `redist.csproj`.

## Resolution

I spent some time digging and I don't pretend to know how the entire build of this project works. However, after a bunch of digging in the binlogs, it only fails for these 2 ProjectReferences:

image

These 2 projects, `Microsoft.NET.Build.Extensions.Tasks` and `Microsoft.NET.Build.Tasks`, build completely fine independently. It is only in the `redist.csproj` project that causes an issue. Now, I don't know why this is only an issue for VS and not for `build.cmd`, but I looked at how they are referenced:

https://github.com/dotnet/sdk/blob/0622981c6d82bda0dab39e38218ef4b5fa760612/src/Layout/redist/redist.csproj#L53-L54

The `SkipGetTargetFrameworkProperties` stood out to me. [Looking at the documentation](https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2022#projectreference), this forces it to not set the `TargetFramework` value.

Now, the error image above says about not finding the `GetTargetPath` target. That comes from the `Microsoft.Common.CurrentVersion.targets file.

https://github.com/dotnet/msbuild/blob/76a6ec27c2fbeee0145c20eec91e17a5490b6c86/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2153-L2169

When rebuilding the build system for the .NET Project System repo, I ran into a similar issue like this when doing SDK-style VSIX projects. I remembered it was because when a `TargetFramework` is not resolved, it doesn't know how to load the subsequent `.target` files since they are framework-specific in folder structure. For example, we use this one in ***amd64***:

> C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\\***amd64***\Microsoft.Common.CurrentVersion.targets

Similarly, there are ones for `CSharp` language if you don't specify the language. In the .NET Project System situation, I forced both the framework and language so certain things could resolve properly. (Note: It doesn't work this way in that repo anymore, but it used to prior to [this PR](https://github.com/dotnet/project-system/pull/8964) from April).

To fix this, I simply removed the `SkipGetTargetFrameworkProperties` for the ProjectReference. I'm not sure why it was set in the first place...? Someone with more knowledge than me might be able to provide some insight.

After this change, the build succeeded. However, there was a warning from the incremental build system:

image

This is showing that warning because it uses build outputs to compare for the "state of the world" before/after build (using things such as timestamps on the files). Took a bunch more digging in the binlogs to figure out why the file wasn't there. I even recorded a video with the folder open side-by-side with the VS log to figure out when `redist.dll` exists (because it does) but then gets deleted.

~After locating when/why that happens, I tried to determine the reason it gets deleted (to see if I'd affect the build by not deleting it). I didn't see anything obvious (such as it getting packaged inappropriately), so I simply avoided deleting the file. The `BundledToolProjects` items are only used in the `GenerateCliRuntimeConfigurationFiles` target in `GenerateLayout.targets`, so I simply modify the item list prior to the Delete being called.~ I split the non-list items into a separate Delete call since those calls get duplicated per item from the item lists. You can see the duplication highlighted here in `#FF00FF`:

image

So, a very minor optimization, but I figured I would point it out. ~With the `redist.dll` remaining after building `redist.csproj`, now there is no longer an incremental build warning. So, we get everything looking green in the end. 🥳~

image

*Edit*: Instead to keeping `redist.dll` in the output folder, I've disabled the fast up-to-date check for the `redist.csproj`. This check drives the warning about the project being out-of-date. Thus, we no longer get this warning and the `redist.dll` can be successfully deleted without issue.

### DIFF
--- a/src/Layout/redist/redist.csproj
+++ b/src/Layout/redist/redist.csproj
@@ -50,8 +50,8 @@
     <ProjectReference Include="..\tool_msbuild\tool_msbuild.csproj" />
     <ProjectReference Include="..\tool_nuget\tool_nuget.csproj" />
     <ProjectReference Include="..\..\Cli\dotnet\dotnet.csproj" />
-    <ProjectReference Include="..\..\Tasks\Microsoft.NET.Build.Tasks\Microsoft.NET.Build.Tasks.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" />
-    <ProjectReference Include="..\..\Tasks\Microsoft.NET.Build.Extensions.Tasks\Microsoft.NET.Build.Extensions.Tasks.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" />
+    <ProjectReference Include="..\..\Tasks\Microsoft.NET.Build.Tasks\Microsoft.NET.Build.Tasks.csproj" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\Tasks\Microsoft.NET.Build.Extensions.Tasks\Microsoft.NET.Build.Extensions.Tasks.csproj" ReferenceOutputAssembly="false" />
     <ProjectReference Include="..\..\Resolvers\Microsoft.NET.Sdk.WorkloadMSBuildSdkResolver\Microsoft.NET.Sdk.WorkloadMSBuildSdkResolver.csproj" />
     <ProjectReference Include="$(RepoRoot)src\BuiltInTools\dotnet-watch\dotnet-watch.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Private="false" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


src/Layout/redist/redist.csproj
joeloff, line 1: You could try using `<Project Sdk="Microsoft.Build.NoTargets">`

You'll have to pull it in through the global.json in the root as an msbuild SDK: ` "Microsoft.Build.NoTargets": "3.7.0"` and see if that makes things easier with the redist.dll

MiYanni: I attempted to do this but the build had issues because `.deps.json` was no longer being generated for that project. Changes that attempted it are [here](https://github.com/MiYanni/sdk/commit/cde00ebf0da03bed979adee1a1474d16a9bb706b). Instead, I've kept the code that removes `redist.dll` from the output and disabled the fast up-to-date check so we don't get the incremental build warning.

joeloff (approved):

PR #34618, opened 2023-08-14
When transforming web.config check if target framework is .NET Framework.

AraHaan:
This will avoid the aspNetCore elements from being added to the web.config when the user targets .NET Framework in an SDK style project and as such avoid a possible runtime error when IIS tries to use the AspNetCore module from the web.config inside of a .NET Framework targeted application and fail. The risk from this should be minimal because of the fact that non-SDK style projects do the same exact thing and they run as-is.

Fixes: https://github.com/dotnet/sdk/issues/53736.

### DIFF
--- a/src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
+++ b/src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
@@ -51,10 +51,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         Condition="'$(_IsAspNetCoreProject)' == 'true'
         And '$(IsTransformWebConfigDisabled)' != 'true'
         And '$(IsWebConfigTransformDisabled)' != 'true'
+        And '$(IsAspNetCoreInWebConfigDisabled)' != 'true'
         And '$(PublishIISAssets)' != 'false'"
         TargetPath="$(TargetPath)"
         PublishDir="$(PublishIntermediateOutputPath)"
-        TargetFrameworkIdentifier="$(TargetFrameworkIdentifier)"
         UseAppHost="$(_UseAppHost)"
         ExecutableExtension="$(_ExecutableExtension)"
         IsAzure="$(_TransformWebConfigForAzure)"
--- a/src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
+++ b/src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
@@ -54,6 +54,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         And '$(PublishIISAssets)' != 'false'"
         TargetPath="$(TargetPath)"
         PublishDir="$(PublishIntermediateOutputPath)"
+        TargetFrameworkIdentifier="$(TargetFrameworkIdentifier)"
         UseAppHost="$(_UseAppHost)"
         ExecutableExtension="$(_ExecutableExtension)"
         IsAzure="$(_TransformWebConfigForAzure)"
--- a/src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
+++ b/src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
@@ -26,6 +26,15 @@ namespace Microsoft.NET.Sdk.Publish.Tasks
         public bool UseAppHost { get; set; }
 
         /// <summary>
+        /// The Target Framework Identifier.
+        /// <br>.NET Framework = ".NETFramework"</br>
+        /// <br>.NET Standard = ".NETStandard"</br>
+        /// <br>.NET Core/.NET 5+ = ".NETCoreApp"</br>
+        /// </summary>
+        [Required]
+        public string TargetFrameworkIdentifier { get; set; }
+
+        /// <summary>
         /// [optional] Transform is targeted for Azure
         /// </summary>
         /// <returns></returns>
@@ -121,7 +130,16 @@ namespace Microsoft.NET.Sdk.Publish.Tasks
             }
 
             string outputFile = Path.GetFileName(TargetPath);
-            XDocument transformedConfig = WebConfigTransform.Transform(
+
+            // For ASP.NET Framework web applications do not add any thing mentioning ASP.NET Core in it's Web.config file.
+            // It could break it at runtime when it tries to use the ASP.NET Core Module at runtime with it's version of ASP.NET
+            // (and as such the other ASP.NET Module for .NET Framework).
+            // Also it's because I expect the files between the web.config on non-SDK style and SDK style to be exact copies
+            // provided that it still targets .NET Framework.
+            // This will capture all apps targeting .NET Framework and skip adding the 2 aspNetCore elements to the web.config
+            // but it is worth it to not have breakage at all in web.config.
+            // Fixes: https://github.com/dotnet/sdk/issues/34595
+            XDocument transformedConfig = TargetFrameworkIdentifier != ".NETFramework" ? WebConfigTransform.Transform(
                 webConfigXml,
                 outputFile,
                 IsAzure,
@@ -130,7 +148,7 @@ namespace Microsoft.NET.Sdk.Publish.Tasks
                 AspNetCoreModuleName,
                 AspNetCoreHostingModel,
                 EnvironmentName,
-                ProjectFullPath);
+                ProjectFullPath) : webConfigXml;
 
             // Telemetry
             transformedConfig = WebConfigTelemetry.AddTelemetry(transformedConfig, ProjectGuid, IgnoreProjectGuid, SolutionPath, ProjectFullPath);
--- a/src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
+++ b/src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
@@ -26,15 +26,6 @@ namespace Microsoft.NET.Sdk.Publish.Tasks
         public bool UseAppHost { get; set; }
 
         /// <summary>
-        /// The Target Framework Identifier.
-        /// <br>.NET Framework = ".NETFramework"</br>
-        /// <br>.NET Standard = ".NETStandard"</br>
-        /// <br>.NET Core/.NET 5+ = ".NETCoreApp"</br>
-        /// </summary>
-        [Required]
-        public string TargetFrameworkIdentifier { get; set; }
-
-        /// <summary>
         /// [optional] Transform is targeted for Azure
         /// </summary>
         /// <returns></returns>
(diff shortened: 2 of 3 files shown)


src/WebSdk/Publish/Tasks/WebConfigTransform.cs
```diff
@@ -98,12 +98,6 @@ private static void TransformAspNetCore(XElement aspNetCoreElement, string appNa
                 appPath = Path.ChangeExtension(appPath, !string.IsNullOrWhiteSpace(extension) ? extension : null);
                 aspNetCoreElement.SetAttributeValue("processPath", appPath);
             }
-            // For Apps targeting .NET Framework, the extension is always exe. RID is not set for .NETFramework apps with PlatformType set to AnyCPU.
```
AraHaan, line 101: This was not always true, non-SDK style .NET Framework MVC applications compile with *.dll's, *.pdb's, *.config's ext files by default only. Those projects when migrated in place (Target Framework wise) to SDK style would refuse to build when set to Exe because they lack a Main and therefore would still need to be set to dll in order to build and run correctly.

Due to that and the fact that my changed code would prevent this from ever being called under .NET Framework, this made this check unneeded hence the cleanup of it.

AraHaan:
Sorry guys, but this had to happen as I need transforms to happen to transform ``Web.Debug.config`` and ``Web.Release.config`` and while disabling that is an option, it would open up security issues due to it not having debug disabled when building views in my application which is not acceptable. This means that for me and others this is the only acceptable change.

AraHaan:
I have verified locally that this works, the only thing missing however is the project guid on the last line but that is not important at all and is acceptable for me.

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
```diff
@@ -121,16 +130,27 @@ public override bool Execute()
...
+            // For ASP.NET Framework web applications do not add any thing mentioning ASP.NET Core in it's Web.config file.
+            // It could break it at runtime when it tries to use the ASP.NET Core Module at runtime with it's version of ASP.NET
+            // (and as such the other ASP.NET Module for .NET Framework).
+            // Also it's because I expect the files between the web.config on non-SDK style and SDK style to be exact copies
+            // provided that it still targets .NET Framework.
+            // This will capture all apps targeting .NET Framework and skip adding the 2 aspNetCore elements to the web.config
+            // but it is worth it to not have breakage at all in web.config.
+            // Fixes: https://github.com/dotnet/sdk/issues/34595
+            XDocument transformedConfig = TargetFrameworkIdentifier != ".NETFramework" ||
+                (string.Equals(Path.GetExtension(Path.Combine(".", outputFile).Replace("/", "\\")), ".exe", StringComparison.OrdinalIgnoreCase) && !UseAppHost)
```
AraHaan, line 143: Hopefully this or case is correct in this.

AraHaan:
cc: @marcpopMSFT

marcpopMSFT:
Pinging @vijayrkn as his team manages the websdk and would need to review this.

Rick-Anderson:
@captainsafia  @lewing who can review this PR?

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
```diff
@@ -25,6 +25,15 @@ public class TransformWebConfig : Task
...
         public bool UseAppHost { get; set; }
 
+        /// <summary>
+        /// The Target Framework Identifier.
+        /// <br>.NET Framework = ".NETFramework"</br>
+        /// <br>.NET Standard = ".NETStandard"</br>
+        /// <br>.NET Core/.NET 5+ = ".NETCoreApp"</br>
+        /// </summary>
+        [Required]
+        public string TargetFrameworkIdentifier { get; set; }
```
vijayrkn, line 35: Adding a new required parameter halfway through the release is risky if the assembly versions of the task assembly hasn't changed.

src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
```diff
@@ -54,6 +54,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         And '$(PublishIISAssets)' != 'false'"
         TargetPath="$(TargetPath)"
         PublishDir="$(PublishIntermediateOutputPath)"
+        TargetFrameworkIdentifier="$(TargetFrameworkIdentifier)"
```
vijayrkn, line 57: This whole task/target is only applicable for ASP.NET Core projects. If the intent here is to not run the target for .NET Framework converted SDK projects, then I would add the logic directly in the target here and not change anything in the task.

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
```diff
@@ -121,16 +130,27 @@ public override bool Execute()
...
+
+            // For ASP.NET Framework web applications do not add any thing mentioning ASP.NET Core in it's Web.config file.
+            // It could break it at runtime when it tries to use the ASP.NET Core Module at runtime with it's version of ASP.NET
+            // (and as such the other ASP.NET Module for .NET Framework).
+            // Also it's because I expect the files between the web.config on non-SDK style and SDK style to be exact copies
+            // provided that it still targets .NET Framework.
+            // This will capture all apps targeting .NET Framework and skip adding the 2 aspNetCore elements to the web.config
+            // but it is worth it to not have breakage at all in web.config.
+            // Fixes: https://github.com/dotnet/sdk/issues/34595
+            XDocument transformedConfig = TargetFrameworkIdentifier != ".NETFramework" ||
```
vijayrkn, line 142: This check is also incorrect because ASP.NET still supports running ASP.NET Core on full framework. Adding this check will potentially cause many valid scenarios to fail.

vijayrkn (commented):
I would really like to understand what scenario this PR is trying to fix. If the intent here is to not run the transform, the recommended way to do it would be to set this msbuild property: 
`<IsWebConfigTransformDisabled>true</IsWebConfigTransformDisabled>`. Making the below change is very risky and would potentially cause additional failures on other valid scenarios.

vijayrkn:
@marcpopMSFT @Rick-Anderson - Thanks for the ping. I have reviewed and updated the comments on the PR.

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
AraHaan: Well the issue with ``IsWebConfigTransformDisabled`` is that then it would no longer transform ``Web.Debug.config`` and ``Web.Release.config`` which is unacceptable and is a security risk when doing a Release config publish and it not removing ``debug`` from ``compilation`` inside of the Web.config on the published copy.

Also this required parameter should be ok as the .NET SDK sets it by default anyways for all supported Target Frameworks in it. The risk should be minimal with that.

vijayrkn: The targets that run for .NET Framework projects are different from this. Can you please share a sample project (that has the issue) so that I can understand the issue better?

AraHaan: Can do.

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
```diff
@@ -25,6 +25,15 @@ public class TransformWebConfig : Task
...
 
+        /// <summary>
+        /// The Target Framework Identifier.
+        /// <br>.NET Framework = ".NETFramework"</br>
+        /// <br>.NET Standard = ".NETStandard"</br>
+        /// <br>.NET Core/.NET 5+ = ".NETCoreApp"</br>
+        /// </summary>
+        [Required]
+        public string TargetFrameworkIdentifier { get; set; }
+
```
AraHaan, line 36:
```suggestion
```

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
```diff
@@ -121,16 +130,27 @@ public override bool Execute()
...
+                ? WebConfigTransform.Transform(
+                    webConfigXml,
+                    outputFile,
+                    IsAzure,
+                    UseAppHost,
+                    ExecutableExtension,
+                    AspNetCoreModuleName,
+                    AspNetCoreHostingModel,
+                    EnvironmentName,
+                    ProjectFullPath) : webConfigXml;
```
AraHaan, line 153:
```suggestion
            XDocument transformedConfig = WebConfigTransform.Transform(
                webConfigXml,
                outputFile,
                IsAzure,
                UseAppHost,
                ExecutableExtension,
                AspNetCoreModuleName,
                AspNetCoreHostingModel,
                EnvironmentName,
                ProjectFullPath);
```

src/WebSdk/Publish/Tasks/WebConfigTransform.cs
```diff
@@ -98,7 +98,7 @@ private static void TransformAspNetCore(XElement aspNetCoreElement, string appNa
...
+            // For some Apps targeting .NET Framework, the extension is always exe. RID is not set for .NETFramework apps with PlatformType set to AnyCPU.
```
AraHaan, line 101:
```suggestion
            // For Apps targeting .NET Framework, the extension is always exe. RID is not set for .NETFramework apps with PlatformType set to AnyCPU.
```

src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
AraHaan, line 57:
```suggestion
```

src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets
```diff
@@ -54,6 +54,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         And '$(PublishIISAssets)' != 'false'"
```
AraHaan, line 54:
```suggestion
        And '$(IsAspNetCoreInWebConfigDisabled)' != 'true'
        And '$(PublishIISAssets)' != 'false'"
```

AraHaan (commented):
Perhaps this might be better then, ``IsAspNetCoreInWebConfigDisabled`` That would act like ``IsTransformWebConfigDisabled`` or ``IsWebConfigTransformDisabled`` but the user would not have to hack around transforming ``Web.$(Configuration).config`` by setting ``RunXdt`` themselves (which I consider to be a nasty hack.

src/WebSdk/Publish/Tasks/Tasks/TransformWebConfig.cs
vijayrkn: IsWebConfigTransformDisabled  is used by both ASP.NET Core and the full framework projects. Can you try setting the other property `IsTransformWebConfigDisabled` to false. This is only used by ASP.NET Core and it should allow the other web.config transforms to run in the full framework project.

AraHaan: When I tried that it looks like it sets ``RunXdt`` to false.

```xml
      <_WebConfigTransformCompleted Condition="'$(_IsAspNetCoreProject)' == 'true'
                                    And '$(IsTransformWebConfigDisabled)' != 'true'
                                    And '$(IsWebConfigTransformDisabled)' != 'true'
                                    And '$(PublishIISAssets)' != 'false'">true</_WebConfigTransformCompleted>
      <_WebConfigTransformCompleted Condition="'$(_WebConfigTransformCompleted)' == ''">false</_WebConfigTransformCompleted>
      <RunXdt Condition="'$(RunXdt)' == ''">$(_WebConfigTransformCompleted)</RunXdt>
```
As such I done a new one 2 hours ago which is ``IsAspNetCoreInWebConfigDisabled`` which would not affect that then.

RobbyJensen:
Is there by chance any update on this? I am having the same problem where it's adding the AspNetCore stuff to my web.config on .Net Framework and would love this fix for it.

AraHaan:
While this is a working fix for this, there seems to be another issue with ASP.NET .NET Framework applications on the SDK, and it is the issue where somehow the SDK fails to copy the roslyn compile files to bin\roslyn on the publish outputs with the direct publish outputs being in the bin folder and not at the root of the publish directory when targeting .NET Framework. As such I published a special MSBuild SDK for the .NET SDK that has a target to fix this after the Publish target finishes to work around that issue for now. Also with this fix the GUID at the bottom of the web.config after publish will be missing, however that part at the bottom of the file is nothing major really.

vijayrkn:
Can you please let me know how you are using this target for ASP.NET .NET framework applications? Are you manually updating the csproj to use the SDK style?

AraHaan:
> Can you please let me know how you are using this target for ASP.NET .NET framework applications? Are you manually updating the csproj to use the SDK style?

I used the migration tool that I then told it to do an in place upgrade where it migrated it to SDK style without changing the target framework.

PR #34640, opened 2023-08-15
Fix runtime pack selection for `dotnet build` with PublishAot=true and PublishAotUsingRuntimePack=true

filipnavara:
When using `dotnet build` instead of `dotnet publish` we are not really running NativeAOT compilation. We are just running a regular build with additional trimming analyzers and options. Thus we are not supposed to use the NativeAOT runtime pack.


rolfbjarne (approved):

steveisok (approved):

steveisok:
This needs to go into rc1 as well. I'm not sure what the rules are in sdk for that.

baronfel:
This technically went into .net 9 (main tired into this today). You'll need to cherry pick PRs to release/8.0.1xx (if missing rc1 is ok but it must be in 8.0.100) and release/8.0.1xx-rc1 (if it must be in rc1 as well).

steveisok:
> This technically went into .net 9 (main turned into this today). You'll need to cherry pick PRs to release/8.0.1xx (if missing rc1 is ok but it must be in 8.0.100) and release/8.0.1xx-rc1 (if it must be in rc1 as well).

Did Marc add `/backport` support? I tried it, but doesn't look like it's happening.

baronfel:
Looks like there are a couple kinks to work out still 😅

PR #34642, opened 2023-08-15
Update onelocbuild to 8.0.1xx

marcpopMSFT:


eng/build.yml
```diff
@@ -15,13 +15,13 @@ parameters:
   timeoutInMinutes: 180
 
 jobs:
-- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), eq(variables['Build.SourceBranch'], 'refs/heads/release/7.0.4xx'), not(contains(parameters.agentOs, 'TemplateEngine'))) }}:
+- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), eq(variables['Build.SourceBranch'], 'refs/heads/release/8.0.1xx'), not(contains(parameters.agentOs, 'TemplateEngine'))) }}:
```
MiYanni, line 18: The branch you targeting is `refs/heads/release/7.0.4xx`. Why change it to `8.0.1xx`?

MiYanni (commented):
I'm a bit confused here. Do the old branches (7.0.4xx) require changes for using the latest OneLocBuild? The template itself exists in the branch. I'd have to dig into the template to see how it uses that `MirrorBranch` variable.

eng/build.yml
marcpopMSFT: We only localize a single branch at a time. I switched the branch to 8.0.1xx on the loc side but without this change we'll still be running the onelocbuild in the 7.0.4xx branch. Potentially, I could have just deleted this entire section but by updating to 8.0.1xx, it'll flow smoothly in inter-branch codeflow (it's already been changed in the 8.0.1xx branch)

MiYanni (approved):

PR #34659, opened 2023-08-16
Ran 'dotnet format' on sdk.sln

MiYanni:
## Summary
When working on other PRs, I keep encountering basic formatting issues (extra whitespace, indentation, using ordering, etc.). Since we've already set these rules up in our `.editorconfig` file, all I did was run `dotnet format` which executes the rules set in the `.editorconfig`. I did no adjustment to the rules.

If some of the changes are not preferred, we need to adjust the rules in our `.editorconfig` to match the preferred style/formatting. Any suggestions welcome.

> **Note**
> These kinds of code formatting inconsistencies don't bother most devs. However, to me, they stick out like a ***sore thumb*** and distract me from doing meaningful changes and are speedbumps in my learning process. I'm just trying to do my due diligence and keep the repo clean and friendly for all devs.


dsplaisted (approved):
I'm not about to review all of the changes, but sounds good to me.

Should we do something to keep this up-to-date so inconsistencies don't creep in?

MiYanni:
> Should we do something to keep this up-to-date so inconsistencies don't creep in?

Absolutely. The problem is, even in 2023, there's not a standard way for .NET repos to automatically apply these formatting rules. There are various ways to run them, but I want something where it ***IS NOT*** a system that bugs you if you've written your code with improper formatting or causes errors on-build locally. I've worked in repos like that and my response is *always*, "I don't care what the formatting is, just apply it for me!". My goals for it would be:
- It applies formatting when building via `build.cmd`
- It applies formatting when using Visual Studio (on-build or otherwise)
- (Stretch goal) It applies formatting when saving the document in Visual Studio
- (Stretch goal) If you create a PR where you have not built the repo (which does happen from time to time), the build will error if the formatting is changed via running the build in CI.

There are various ways to accomplish each of these things. Looking for anyone with advice here on how to do this. CC @baronfel @joeloff @AArnott

baronfel:
I agree with your goals, but I would also add in 'Visual Studio Code' as well as VS - especially with DevContainers VSCode usage is very drop-in for contributors.

baronfel:
The FSharp team has incorporated a formatting check into their pipeline, as well as an action that can be triggered `(/fantomas)` to have the pipeline apply the formatting. @vzarytovskii would you mind linking to that?

Evangelink:
Hey @mavasani, do you know if there is dome tooling around the auto apply format rules on build? Or some guideline around it?

vzarytovskii:
> The FSharp team has incorporated a formatting check into their pipeline, as well as an action that can be triggered `(/fantomas)` to have the pipeline apply the formatting. @vzarytovskii would you mind linking to that?

Yeah, the action is here: https://github.com/dotnet/fsharp/blob/main/.github/workflows/commands.yml

Check is there in azure pipelines yaml file as well.

mavasani:
There is already a well supported way to enforce first party code style rules, including formatting, for C# and VB projects: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-7#enable-on-build.

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0055

You can either set the MSBuild property [EnforceCodeStyleInBuild](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild) OR add an explicit NuGet package reference to https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.CodeStyle

mavasani:
> Hey @mavasani, do you know if there is dome tooling around the auto apply format rules on build? Or some guideline around it?

Tagging @JoeRobich @sharwell

sharwell:
I wouldn't trust the code fixes to not make any breaking changes, and it's impossible to review a PR that applies all changes at once. I would never merge a PR that applied more than one rule at a time.

JoeRobich:
> For example, setting Code Cleanup in Visual Studio is something that a user has to do on their own, and the repo cannot enforce it.

Wouldn't it be nice if you could add settings for this to your .editorconfig.

MiYanni:
> There is already a well supported way to enforce first party code style rules, including formatting, for C# and VB projects on command line builds: [learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-7#enable-on-build](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-7#enable-on-build).
> 
> [learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0055](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0055)
> 
> You can either set the MSBuild property [EnforceCodeStyleInBuild](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild) OR add an explicit NuGet package reference to [nuget.org/packages/Microsoft.CodeAnalysis.CSharp.CodeStyle](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.CodeStyle)

@mavasani This looks to be *enforcement* and not *application*. Looking at the docs for `EnforceCodeStyleInBuild`:
> All code style rules that are configured to be warnings or errors will execute on build and report violations.

I'm looking for a solution that **applies** the rules. Telling me that my style/formatting is wrong is *what I want to avoid*. **Applying** the style/formatting of the repo to the code I write is what is desired. I've done this before at other companies via third-parties tools (*cough* [CodeMaid](https://www.codemaid.net/) *cough*) but I'm looking for what is the best way to do that here. I'm not looking to add friction (more build warnings/errors); I'm looking to reduce it (via automatic formatting/application of rules).

JoeRobich:
> I'm looking for a solution that **applies** the rules.

Not that I am recommending it, but with the C# VS Code extension using O#, you could add the following to your settings.

```json
"editor.codeActionsOnSave": {
  "source.fixAll.csharp": true
}
```
It may work again once FixAll is added to Roslyn LSP.

baronfel:
@JoeRobich does dotnet-format have a concept of like syntax-only changes? formatting, etc? Because that's what I'd expect to auto-format. Refactorings should be left on a manual basis.

JoeRobich:
> does dotnet-format have a concept of like syntax-only changes?

@baronfel Yes. You can run `dotnet format whitespace` which will run the Roslyn formatter. You can also use the `--folder` option to speed things up by avoiding design-time builds to populate the workspace.

AArnott:
The solution I've lived with for years is a repo that enforces warning-free builds in PR/CI, and style issues produce warnings. That way users can iterate in their inner-loop without worrying about style, but VS flags them on it as they go and they know they need to fix it before checking in. Automating it on Save can be _great_, as I live that life in VS Code for Typescript repos.
The P0 condition for any of this of course is for the style rules to be codified in the repo itself so that every editor/formatter can respect the same set.

baronfel:
Here's a basic tasks.json file (for use in VSCode) that would trigger whitespace formatting only (due to perf mostly):

```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "type": "shell",
      "command": "dotnet",
      "args": [
        "format",
        "whitespace",
        ".",
        "--folder",
        "--exclude",
        "./Assets"
      ],
      "options": {
        "cwd": "${workspaceFolder}/src"
      },
      "label": "Format all code",
      "problemMatcher": []
    }
  ]
}
```

I couldn't find a way to call format just for a single file. perf-wise it's long-running-enough that I wouldn't trigger it on save, though.

JoeRobich:
> perf-wise it's long-running-enough that I wouldn't trigger it on save

For VS Code you can just take advantage of setting `"editor.formatOnSave": true` in your settings. With the C# extension installed (whether O# or Roslyn LSP powered) the Roslyn formatter will be run against the saved file.

baronfel:
Perfect @JoeRobich!

JoeRobich:
The O# powered extension also supports `"omnisharp.organizeImportsOnFormat": true` which isn't part of the Roslyn formatter. Would have to ask @dibarbet about Roslyn LSP support for that option.

marcpopMSFT:
Weighing in on the formatting question, I would not support build errors related to style/formatting. I would support format on save behavior. I'm aware that we risk customer PRs not matching our formatting as there's no way to enforce format on save in the repo that's a no effort solution for developers.

MiYanni:
I've added an issue for doing the work on some kind of automatic mechanism for running the style/formatting: https://github.com/dotnet/sdk/issues/34810

PR #34660, opened 2023-08-16
Allow ArtifactsPath telemetry to flow

baronfel:
https://github.com/dotnet/sdk/pull/32685 tried to do too much and got stuck in testing - this just explicitly adds and hashes the two ArtifactsPath-related properties we added back in preview 3, so that we can actually see them in telemetry.


marcpopMSFT (approved):

baronfel:
Closing because this made it to RC2.

Issue #14279, opened 2020-10-22
when rebuild a new winformsApp by using targetFramework .Net5.0, it shows error that is showed in comment below.


liyanligt:
when rebuild a new winformsApp by using targetFramework .Net5.0, it shows error that "	NETSDK1136	The target platform must be set to Windows (usually by including '-windows' in the TargetFramework property) when using Windows Forms or WPF, or referencing projects or packages that do so.	WindowsFormsApp1	C:\Program Files\dotnet\sdk\5.0.100-rc.2.20479.15\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets".
When I changed the targetFramework to .Net3.1, the error disappeared.

The infomation of sdk and runtime is below:
Visual Studio support
Visual Studio 2019 (v16.8, Preview 4)
Included runtimes
.NET Runtime 5.0.0-rc.2.20475.5
ASP.NET Core Runtime 5.0.0-rc.2.20475.5
.NET Desktop Runtime 5.0.0-rc.2.20475.5

Do you have found this qustion ? and  how to solve this problem? Thank you very much!

dbulic:
See https://developercommunity.visualstudio.com/content/problem/1218104/cannot-compile-wpf-applications-targeting-net-5.html

wli3:
+1 on @dbulic 's like. And the error message is correct. Just follow the error message's instruction.

liyanligt:
Thank you very much , it has been solved by changing targetframework to net5.0-windows。
![image](https://user-images.githubusercontent.com/53420736/97260311-a686b700-1857-11eb-907a-7686dcc43b17.png)
https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md

peterdrier:
> Thank you very much , it has been solved by changing targetframework to net5.0-windows。

This isn't actually a solution, as you're fundamentally changing the target of the application to Windows Only.  If you are trying to do cross platform work, it still doesn't work.

AraHaan:
> 
> 
> > Thank you very much , it has been solved by changing targetframework to net5.0-windows。
> 
> This isn't actually a solution, as you're fundamentally changing the target of the application to Windows Only. If you are trying to do cross platform work, it still doesn't work.

I also got a case where I make a copy of a library with usage of windows forms and a copy without by optionally defining ``WITH_WINFORMS`` and have also changed mine recently to support .net 5 by doing the following:

[...]

PR #34666, opened 2023-08-16
Fix in Microsoft.NET.Sdk.DefaultItems.targets for .DS_Store causing build failures on MacOSX

moljac:
Context:

.DS_Store file is internal MacOSX file created by Finder and it is irrelevant for .NET builds, but can interfere with .NET builds.

Examples of build failures (dotnet/maui issues):

- https://github.com/dotnet/maui/issues/13452
  Resizetizer is not working if you get a .DS_Store file inside you Fonts files folder #13452
- https://github.com/dotnet/maui/issues/14279
  Microsoft.Maui.Resizetizer.targets breaks build on macOS #14279

Cause is globbing patterns which include `.DS_Store` on MacOSX:

https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj#L48 
https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj#L52

The workarounds in MAUI repo was:

1. to add `Exclude = "Resources\Images\.DS_Store" to MauiImage`
   ```
   <MauiImage Include="Resources\Images\*" Exclude = "Resources\Images\.DS_Store" />
   ```
2. to add property:
   ```
   <DefaultItemExcludes>$(DefaultItemExcludes);**\.DS_Store</DefaultItemExcludes>
   ```
   and use it for ItemGroups:
   ```
   <MauiImage Include="Resources\Images\*" Exclude="$(DefaultItemExcludes)" />
   ```

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets
@@ -33,6 +33,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DefaultItemExcludes>$(DefaultItemExcludes);**/*.*proj</DefaultItemExcludes>
     <DefaultItemExcludes>$(DefaultItemExcludes);**/*.sln</DefaultItemExcludes>
     <DefaultItemExcludes>$(DefaultItemExcludes);**/*.vssscc</DefaultItemExcludes>
+    <DefaultItemExcludes>$(DefaultItemExcludes);**/.DS_Store</DefaultItemExcludes>
 
     <!-- WARNING: This pattern is there to ignore folders such as .git and .vs, but it will also match items included with a
          relative path outside the project folder (for example "..\Shared\Shared.cs").  So be sure only to apply it to items


steveisok (approved):
LGTM!

dsplaisted (approved):
LGTM.  Note that if you want this in .NET 8, it will need to be retargeted (or backported).

This could technically be a breaking change which we'd want to document.  Thoughts on whether this merits a breaking change notice @baronfel?

baronfel:
We shouldn't need a breaking change notice here - _everyone's_ got a 'delete .DS_Store' commit in their history from a time they accidentally checked it into a repo, and there are very few reasons for it to ever be in a shipped product. If we document the default includes/excludes somewhere though we should definitely add this change to that.

mattleibow:
Is it possible to get this into net8 for the next RC so we don't have to do terrible things in the maui workload.

rmarinho:
Yeah we would like to see this backported. Do we need to create a new pr?

baronfel:
Yes. Both to release/8.0.1xx-rc1 and release/8.0.1xx branches to ensure you get into RC1 as well as RC2+.

moljac:
> Yeah we would like to see this backported. Do we need to create a new pr?

Are you gonna do it? Or should I?

moljac:
PR 1  release/8.0.1xxx

https://github.com/dotnet/sdk/pull/34710

PR #34693, opened 2023-08-16
Implement automated cleanup of stage 0

marcpopMSFT:
In discussions, we've seen folks have a lot of issues with the .dotnet folder grow larger and larger and cause issues (like with multiple template folders). Since we use stage0 both to build the SDK but also to create the stage2 SDK, it can cause issues with our tooling as well in the Overlay target.

In this PR, we write out a .version file in the .dotnet directory with the sdk version in global.json (which we just installed). 

In future builds, if this version is different than what's in the global.json, we delete a bunch of folders from the .dotnet repo and then ask the developer to rerun. I don't believe we can rerun as we would risk a circular dependency.

Do folks want this to be in place?

### DIFF
--- a/eng/restore-toolset.ps1
+++ b/eng/restore-toolset.ps1
@@ -154,6 +154,9 @@ function CleanOutStage0ToolsetsAndRuntimes {
       Remove-Item (Join-Path $coreRuntimePath "$majorVersion.*") -Recurse
       Remove-Item (Join-Path $wdRuntimePath "$majorVersion.*") -Recurse
       Remove-Item (Join-Path $sdkPath "$majorVersion.*") -Recurse
+      Remove-Item (Join-Path $dotnetRoot "packs") -Recurse
+      Remove-Item (Join-Path $dotnetRoot "sdk-manifests") -Recurse
+      Remove-Item (Join-Path $dotnetRoot "templates") -Recurse
       throw "Installed a new SDK, deleting existing shared frameworks and sdk folders. Please rerun build"
     }
   }
@@ -166,5 +169,3 @@ function CleanOutStage0ToolsetsAndRuntimes {
 InitializeCustomSDKToolset
 
 CleanOutStage0ToolsetsAndRuntimes
-
-throw "exit early"
--- a/eng/restore-toolset.ps1
+++ b/eng/restore-toolset.ps1
@@ -133,4 +133,38 @@ function InstallDotNetSharedFramework([string]$version) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


marcpopMSFT:
FYI, I would prefer this would be solved in arcade: https://github.com/dotnet/arcade/issues/13964

eng/restore-toolset.ps1
```diff
@@ -133,4 +133,39 @@ function InstallDotNetSharedFramework([string]$version) {
...
+      Remove-Item (Join-Path $dotnetRoot "sdk-manifests") -Recurse
+      Remove-Item (Join-Path $dotnetRoot "templates") -Recurse
+      throw "Installed a new SDK, deleting existing shared frameworks and sdk folders. Please rerun build"
+    }
+  }
+  else
+  {
+    $dotnetSdkVersion | Out-File -FilePath $versionPath -NoNewline
+  }
+}
```
MiYanni, line 167: So, if there is a `$versionPath` but the `$lastInstalledSDK` is the same as the SDK version in the global.json, this function is a no-op, correct? And if either there is no `$versionPath` or there is one and `$lastInstalledSDK` isn't the same version in the global.json, then it will write out the version into `$versionPath`, correct?

eng/restore-toolset.sh
```diff
@@ -68,4 +68,52 @@ export NUGET_PACKAGES=$NUGET_PACKAGES
...
+      ExitWithExitCode 1
+    fi
+  else
+    echo $dotnetSdkVersion > $versionPath
+  fi
+}
+
 InitializeCustomSDKToolset
+
+CleanOutStage0ToolsetsAndRuntimes
```
MiYanni, line 119: General Q: I guess the design of Arcade requires duplicating scripts between 2 different scripting languages? I find that odd since modern PowerShell works cross-plat. I guess there must be some reason we couldn't just require PowerShell on Linux?

MiYanni (approved):

eng/restore-toolset.ps1
marcpopMSFT: correct. The logic above is if the file doesn't exist, write out the version from global.json (this is basically so it doesn't fail the first build run in any new repo). If the file does exist AND it's different than what's in global.json, delete everything, write out the file, and exit.

So it should never write to the file if it matches what's in global.json

eng/restore-toolset.sh
marcpopMSFT: That's a dnceng question. This is how it works across all repos. It might be a limitation of powershell on some linux distros.

PR #34732, opened 2023-08-17
Enable interceptors feature when ConfigurationBindingGenerator is enabled

layomia:
Assuming we don't change our minds about this approach, I'm setting up this PR to merge along side the ConfigBindingGenerator implementation. https://github.com/dotnet/runtime/pull/90340/.

cc @ericstj


eerhardt:
FYI - you probably want to target a `release/8` branch and not `main`, if you want this change included in 8.0.

layomia:
Will open another PR.

PR #34863, opened 2023-08-23
[release/8.0.1xx] Containers: prefer starting the app using 'dotnet <assembly.dll>'.

### LINKED ISSUE #34860: Container image built using source-built SDK may fail to run
By default applications are published using an apphost, the container tooling targets Microsoft base images, and the tooling uses the apphost to start the app.

However, the source-built apphost may be incompatible with the Microsoft base images.

When I run this using the Fedora .NET 8 preview 7 SDK:

```
dotnet new web -o web
cd web
dotnet publish /p:PublishProfile=DefaultContainer
```

The resulting image does not work:
```
$ podman run web
You must install .NET to run this application.

App: /app/web
Architecture: x64
App host version: 8.0.0-preview.7.23375.6
.NET location: Not found

Learn more:
https://aka.ms/dotnet/app-launch-failed

Download the .NET runtime:
https://aka.ms/dotnet-core-applaunch?missing_runtime=true&arch=x64&rid=fedora.37-x64&os=debian.12&apphost_version=8.0.0-preview.7.23375.6
```

The source-built apphost is unable to find the .NET root because in our builds we've changed the sources to where the .NET root to match where we have it on Fedora. I don't think we actually still need to patch this location... but the issue remains:

The source-built apphost can be binary incompatible with the Microsoft base images: it may have a glibc version issue, and on a musl-based distro (like Alpine) the apphost isn't going to run on the glibc Microsoft base images either.

Can we update the container logic for .NET 8 so that it uses the apphost command only for `SelfContained`?

cc @baronfel @richlander @MichaelSimons @omajid

tmds:
Fixes https://github.com/dotnet/sdk/issues/34860.

The default apphost, which comes with the SDK, may be incompatible with the container base image.

For example: the apphost of a musl-based Linux SDK is not compatible with the default (glibc Debian) based Microsoft base images.

And, a source-build glibc based apphost may also be incompatible with the Microsoft glibc based images due to glibc symbol versioning.

@baronfel ptal.


baronfel:
Separately from the issue you raised (which I'm still wrapping my head around) - I think the previous logic would fail at runtime when `SelfContained` is true and `UseAppHost` is false. This isn't something the SDK would ever _infer_ for you but a user could have set UseAppHost explicitly for some reason. So I think your change insulates us from that particular problem as well.

baronfel:
so - if you want this to be in .NET 8 you'll need to retarget to release/8.0.1xx (or backport there after this merges).

tmds:
> if you want this to be in .NET 8 you'll need to retarget to release/8.0.1xx (or backport there after this merges).

Yes, I'd like this to be in .NET 8.

I assume if I would target the release branch with my PR, we'll need to forward port to main?
What's the preferred option?

baronfel:
We do periodic codeflow from release branches to main, so that should just happen automatically.

tmds:
ok, I'll update this to target the 8.0 release branch.

baronfel:
This looks good to me and solves a concrete problem we have today - I think we should review it and take it for RC2.

baronfel (approved):
Thanks again for fixing this gap!

PR #34871, opened 2023-08-23
Validate "this" extension methods and limit visibility

### LINKED ISSUE #34824: [API Compat] Add extension method validation
Add validation that `this` modifier is present in ref of the extension method, to prevent issues like this one: https://github.com/dotnet/runtime/issues/90911

Currently, there's no such validation, so it is possible to miss `this` modifier by accident, if ref is updated manually.

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/34824

cc @CarnaViire


andriipatsula (approved):
Looks good. Please create an issue to cover this scenario with Tests.

ViktorHofer:
I will add tests directly to this PR.

PR #34872, opened 2023-08-23
Validate member types

### LINKED ISSUE #34864: [API Compat] Changing a member return type should be detected as a breaking change
### Describe the bug
Changing the return type of a member, whether its a field, property or method isn't detected by the API Compat tool.

### To Reproduce
The following zip files has two assemblies, `before.dll` and `after.dll`, each with a single class `C` having three members. The member names are the same for the two assemblies, but their return type is different. These are breaking changes, and will result in `MissingFieldException`/`MissingMemberException`/`MissingMethodException` at runtime.

[diff.zip](https://github.com/dotnet/sdk/files/12418773/diff.zip)

Before.dll:
```csharp
public class C
{
    public readonly string Field = "f";
    public int Property => 123;
    public double Method() => 45.6;
}
```

After.dll:
```csharp
public class C
{
    public readonly int Field = 0;
    public string Property => "p";
    public DateTime Method() => DateTime.Now;
}
```

Run
```
apicompat -l=./before.dll -r=./after.dll
```

Expected: APICompat reports three breaking changes.
Actual: "APICompat ran successfully without finding any breaking changes."

### Further technical details
Reproduced with both:
 - Latest release: 7.0.400+8e1f76c1f41ecd37935288c617c53342533db704
 - Latest nightly build: 9.0.100-alpha.1.23417.7+4249701de6f005c670c7f371141e772d02f89156

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/34864

cc @MrJul

### DIFF
--- a/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
+++ b/src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
@@ -531,5 +531,56 @@ namespace CompatTests
                 CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.MemberMustExist, string.Empty, DifferenceType.Removed, "M:CompatTests.First.F(System.String)")
             }, differences);
         }
+
+        [Fact]
+        public void MemberTypesChangeFlagged()
+        {
+            string leftSyntax = @"
+using System;
+
+namespace CompatTests
+{
+  public class First
+  {
+    public string S;
+    public bool Prop { get; set; }
+    public string M() => null;
+    public delegate void SampleEventHandler(object sender, EventArgs e);
+    public event SampleEventHandler E;
+  }
+}
+";
+            string rightSyntax = @"
+using System;
+
+namespace CompatTests
+{
+  public class First
+  {
+    public delegate void SampleEventHandler(object sender, EventArgs e);
+
+    public int S;
+    public string Prop { get; set; }
+    public bool M() => false;
+    public delegate void SampleEventHandler1(object sender, EventArgs e);
+    public event SampleEventHandler1 E;
+  }
+}
+";
+            IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
+            IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
+            ApiComparer differ = new(s_ruleFactory);
+
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(left, right);
+
+            Assert.Equal(new[]
+            {
+                CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.MemberMustExist, string.Empty, DifferenceType.Removed, "F:CompatTests.First.S"),
+                // CompatTests.First.Prop.set isn't reported as the return types match: 'void'.
+                CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.MemberMustExist, string.Empty, DifferenceType.Removed, "M:CompatTests.First.get_Prop"),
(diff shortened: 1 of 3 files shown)


src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
```diff
@@ -531,5 +531,56 @@ public static class First
...
+
+            IEnumerable<CompatDifference> differences = differ.GetDifferences(left, right);
+
+            Assert.Equal(new[]
+            {
+                CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.MemberMustExist, string.Empty, DifferenceType.Removed, "F:CompatTests.First.S"),
+                // CompatTests.First.Prop.set isn't reported as the return types match: 'void'.
+                CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.MemberMustExist, string.Empty, DifferenceType.Removed, "M:CompatTests.First.get_Prop"),
+                CompatDifference.CreateWithDefaultMetadata(DiagnosticIds.MemberMustExist, string.Empty, DifferenceType.Removed, "M:CompatTests.First.M")
+                // CompatTests.First.E_add and CompatTests.First.E_remove aren't reported as the symbol's DisplayString doesn't include the parameter type.
```
ViktorHofer, line 582: Reached out to @CyrusNajmabadi offline to clarify if this behavior is expected.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
```diff
@@ -531,5 +531,56 @@ public static class First
...
+
+namespace CompatTests
+{
+  public class First
+  {
+    public string S;
+    public bool Prop { get; set; }
+    public string M() => null;
+    public delegate void SampleEventHandler(object sender, EventArgs e);
+    public event SampleEventHandler E;
```
carlossanlop, line 549: What about nested member types? Class, struct, enum?

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
carlossanlop, line 582: What is "DisplayString" here?

carlossanlop (approved):
Left two questions, but lgtm.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
ViktorHofer: The `DisplayString` is the string that is returned from the ISymbol's `GetDisplayString()` method. It is configurable but it currently doesn't include the parameter type or the parameter at all for the event method.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
ViktorHofer: This is just about members: fields, properties, events and methods. They behave the same in nested classes/structs.

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/MembersMustExistTests.cs
CyrusNajmabadi: discussed offline.  i feel like this is expected.  note: this space makes me *very* wary.  It's very much *not* an intended use case of .ToDisplayString (which, as the name suggests, is entirely designed around just displaying symbols to a user, not for any sort of comparison/versioning story).

carlossanlop: @CyrusNajmabadi should comparisons be done between DocIds instead?

PR #34899, opened 2023-08-24
Removing stale msbuild feed

JanKrivanek:
### Context

The feed seems to be stale, and based on stats not actually used: https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/724b07b9-93de-4ad8-a0d0-29346fed06b5/nuget/v3/query2?q=Microsoft.Build&prerelease=true

MSBuild main gets pushed to [dotnet-tools](https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json) feed.

So trying to remove - to see if it get's used anywhere (and if yes - then why it is used)

FYI @marcpopMSFT


marcpopMSFT (approved):

marcpopMSFT:
Thanks. The feed was added because we needed to target an unreleased stable version of msbuild (17.6.1) at one point but it's no longer required for the moment at least.

PR #34928, opened 2023-08-25
Use Razor with renamed compiler DLLs

jjonescz:
PoC for https://github.com/dotnet/razor/issues/8400.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -33,6 +33,8 @@
     <add key="vs-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" />
     <!-- Used for Rich Navigation indexing task -->
     <add key="richnav" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-buildservices/nuget/v3/index.json" />
+    <!-- For testing purposes only -->
+    <add key="general-testing" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/general-testing/nuget/v3/index.json" />
   </packageSources>
   <disabledPackageSources>
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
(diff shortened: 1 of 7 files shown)


eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -38,4 +38,9 @@
     <UsagePattern IdentityGlob="System.Reflection.MetadataLoadContext/*7.0.0*" />
     <UsagePattern IdentityGlob="System.Security.Cryptography.ProtectedData/*7.0.0*" />
   </IgnorePatterns>
+  <Usages>
+    <!-- Dependencies of the legacy `rzc.dll` (see `_LegacyRazorTool` in `GenerateLayout.targets`). -->
+    <Usage Id="Microsoft.AspNetCore.Mvc.Razor.Extensions.Tooling.Internal" Version="7.0.0-preview.23330.1" IsDirectDependency="true" />
+    <Usage Id="Microsoft.CodeAnalysis.Razor.Tooling.Internal" Version="7.0.0-preview.23330.1" IsDirectDependency="true" />
```
jjonescz, line 44: @dotnet/source-build-internal - we would like to have a legacy version of `rzc.dll` built here which would depend on a fixed version of Razor DLLs. Can these pre-builts be added as exceptions?

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -38,4 +38,9 @@
     <UsagePattern IdentityGlob="System.Reflection.MetadataLoadContext/*7.0.0*" />
     <UsagePattern IdentityGlob="System.Security.Cryptography.ProtectedData/*7.0.0*" />
   </IgnorePatterns>
+  <Usages>
```
MichaelSimons, line 41: These will break source-build as customers require building offline without prebuilts.  @jaredpar started a thread with the source-build team on this including @mmitche and @mthalman.

jjonescz: Yes, I'm trying a different approach now, thanks.

PR #34946, opened 2023-08-25
Workload history

Forgind:
This is based off of @dsplaisted's changes. I additionally resolved conflicts, modernized it (i.e., replaced references to things that didn't exist, etc.), and addressed most of the explicit todos. I think it should output workload history records now but does not include a mechanism for using them properly yet.

More testing this afternoon.


src/Cli/dotnet/commands/dotnet-workload/history/WorkloadHistoryCommand.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        public override int Execute()
+        {
+            var displayRecords = WorkloadHistoryDisplay.ProcessWorkloadHistoryRecords(_workloadInstaller.GetWorkloadHistoryRecords());
+
+            if (!displayRecords.Any())
+            {
+                //  TODO: Localize
+                Reporter.WriteLine("No workload history found");
```
MiYanni, line 72: Should we handle creating the resources for each of these TODOs before merging?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -68,67 +70,100 @@ private void ValidateWorkloadIdsInput()
         public override int Execute()
         {
             bool usedRollback = !string.IsNullOrWhiteSpace(_fromRollbackDefinition);
-            if (_printDownloadLinkOnly)
+
+            WorkloadHistoryRecorder recorder = new WorkloadHistoryRecorder(_workloadResolver, _workloadInstaller);
+            recorder.HistoryRecord.CommandName = IsRunningRestore ? "restore" : "install";
```
MiYanni, line 75: I see these manual strings being inserted into this CommandName property. Would it make sense to use an enum instead with the different types of commands?

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
```diff
@@ -43,47 +44,61 @@ internal class WorkloadUninstallCommand : WorkloadCommandBase
...
             var sdkFeatureBand = new SdkFeatureBand(_sdkVersion);
-            _workloadInstaller = WorkloadInstallerFactory.GetWorkloadInstaller(Reporter, sdkFeatureBand, creationResult.WorkloadResolver, Verbosity, creationResult.UserProfileDir, VerifySignatures, PackageDownloader, creationResult.DotnetPath);
+            _workloadInstaller = WorkloadInstallerFactory.GetWorkloadInstaller(Reporter, sdkFeatureBand, _workloadResolver, Verbosity, creationResult.UserProfileDir, VerifySignatures, PackageDownloader, creationResult.DotnetPath);
         }
 
         public override int Execute()
         {
+            WorkloadHistoryRecorder recorder = new WorkloadHistoryRecorder(_workloadResolver, _workloadInstaller);
+            recorder.HistoryRecord.CommandName = "uninstall";
+            recorder.HistoryRecord.WorkloadArguments = _workloadIds.Select(id => id.ToString()).ToList();
```
MiYanni, line 59: I noticed a pattern where in each `Execute` method for the workload commands, you're creating a new `WorkloadHistoryRecorder` and then using it. Instead, why not make a `WorkloadHistoryRecorder` a property of the base `WorkloadCommandBase` and then use that property when it is necessary in each command? You'd have to send the necessary information to the constructor of `WorkloadCommandBase` to make a new recorder.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -52,65 +54,88 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
 
         public override int Execute()
         {
-            if (!string.IsNullOrWhiteSpace(_downloadToCacheOption))
+            WorkloadHistoryRecorder recorder = new WorkloadHistoryRecorder(_workloadResolver, _workloadInstaller);
+            recorder.HistoryRecord.CommandName = "update";
+            //recorder.HistoryRecord.WorkloadArguments = _workloadIds.Select(id => id.ToString()).ToList();
```
MiYanni, line 59: nit: Commented code

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -52,65 +54,88 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
...
+                    {
+                        throw new GracefulException(string.Format(LocalizableStrings.WorkloadCacheDownloadFailed, e.Message), e, isUserError: false);
+                    }
+                }
+                else if (_printDownloadLinkOnly)
+                {
+                    var packageUrls = GetUpdatablePackageUrlsAsync(_includePreviews).GetAwaiter().GetResult();
+                    Reporter.WriteLine("==allPackageLinksJsonOutputStart==");
+                    Reporter.WriteLine(JsonSerializer.Serialize(packageUrls, new JsonSerializerOptions() { WriteIndented = true }));
+                    Reporter.WriteLine("==allPackageLinksJsonOutputEnd==");
```
MiYanni, line 92: Isn't this duplicate code from `WorkloadInstallCommand`?

MiYanni (commented):
I'll review again when the compilation errors are resolved.

src/Cli/dotnet/commands/dotnet-workload/WorkloadHistoryDisplay.cs
```diff
@@ -0,0 +1,58 @@
...
+//
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.DotNet.Workloads.Workload;
+
+namespace Microsoft.DotNet.Cli.commands.dotnet_workload
```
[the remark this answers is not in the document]
dsplaisted: Yes, the folder layout we use means that the generated namespaces are not what we want so we have to manually adjust them.

src/Cli/dotnet/commands/dotnet-workload/WorkloadHistoryRecord.cs
```diff
@@ -0,0 +1,88 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Workloads.Workload
```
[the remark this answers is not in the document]
Forgind: This pattern appears to be very common for some reason? It's often followed by an extra name, though, so I'll add History to the end.

src/Cli/dotnet/commands/dotnet-workload/WorkloadHistoryRecord.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    internal class WorkloadHistoryState
+    {
+        public Dictionary<string, string> ManifestVersions { get; set; }
+
+        public List<string> InstalledWorkloads { get; set; }
+
+        //  Possibly we should add this in the future, but it requires changes to the IAL which to some degree break the abstraction
+        //public List<(string id, string version)> InstalledWorkloadPacks { get; set;}
```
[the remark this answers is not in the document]
Forgind: https://github.com/dotnet/sdk/issues/35452

src/Cli/dotnet/commands/dotnet-workload/history/WorkloadHistoryCommand.cs
Forgind: Ah, yeah; forgot about this. I should add it to my todo list 👍

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
Forgind: Maybe...personally, I think I'd vote against this because although there aren't that many possibilities, we don't touch it much (it's basically set once, then used once), and when we ultimately use it, it needs to be a string for printing, so that would add a little more indirection for a little more tidiness...

Certainly a reasonable suggestion; I just prefer a string here, personally.

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
Forgind: One of the arguments to make a WorkloadHistoryRecorder is the _workloadInstaller, and though that's present on WorkloadCommandBase, it's initialized in the constructor for each command separately, so I don't think I could move the WorkloadHistoryRecorder's initialization all the way to the base's constructor. I think I could move it back to the constructor, but I'm not sure I see a lot of added value in doing that.

Did I miss something?

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
Forgind: Yeah; I can probably merge those 🙂

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
MiYanni: Doing a quick search in the PR, it looks like `InstallingWorkloadCommand` uses `WorkloadCommandBase` but doesn't create a `WorkloadHistoryRecorder`. That seems a bit... odd, right? I was looking for uses of `WorkloadCommandBase` in the PR for context.

For this pattern, since it only relies on `_workloadResolver` and `_workloadInstaller` to create a `WorkloadHistoryRecorder` and both of those are created in the constructor, it looks logical to make an class-wide instance of `WorkloadHistoryRecorder`, right? Do you need a new one each call to `Execute`? That's generally how to think about it. What is the lifetime necessary for it? Does it bind to the class lifetime or `Execute`'s lifetime?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
MiYanni: The purpose of using an enum is it consolidates all the possible values to one location. So, for me to find the values possible, I had to jump around to all uses of `CommandName` in your PR and keep track of the values. I saw:
- `install`
- `uninstall`
- `restore`
- `update`

If the only values applicable for `CommandName` is a closed set (meaning, not user-inputable or other external names), then using an enum is preferred. It lets every dev know that it is intended to be a finite set of values.

Here's a quick mockup showing how easy it is. `ToString` on an enum value in C# outputs the name of the value, not the backing integer. https://dotnetfiddle.net/AuISsD

image

Shows this output:
image

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
Forgind: Aren't they the same anyway? When someone calls a command, say, `dotnet workload install <args>`, it initializes the class then calls execute. Once execute is finished, so is the command, and it disposes of everything, right? So I think they're equivalent performance-wise.

MiYanni: > So I think they're equivalent performance-wise.

I wasn't particularly concerned about performance in this situation. It is more about code duplication and maintenance. If you can create something one place and reuse it in multiple places in the code, it is (generally) more maintainable. But I'm trying to rationalize if it makes sense here, as returning to this 3 weeks later is giving my brain a bit of stretching to remember. 😅

>Aren't they the same anyway?

So, I was thinking about it and these classes can *technically* be used for purposes other than running execute. But I don't have any examples of this. Even the `dotnet restore` is a completely separate command than the restore operation logic (called `RestoringCommand`) that is used by things like Build, Pack, Publish, etc.

I think the primary purpose of my original comment here wasn't on it causing any functional differences. I just saw a pattern that involved duplication and wondered if it was necessary to do that duplication. Base classes usually handle that burden of reducing duplication. Consolidating boilerplate helps maintainability, generally.

baronfel:
closing in favor of #36809

PR #34980, opened 2023-08-28
Change containers publish target to require IsPublishable (and explicit SDK Container support)

baronfel:
Closes https://github.com/dotnet/sdk-container-builds/issues/464

This change brings containers in line with other publish outputs in respecting user opt outs for this property.


baronfel:
This should be a no-op for console users because console apps are IsPublishable by default.

rainersigwald (approved):

PR #34983, opened 2023-08-28
Added JavaScript import after

joj:
Esprojs may be referenced by ASP.Net projects (or others). When this happens, the .net project will have to not reference output assembly on the esproj, because there is no output assembly. This targets file provides that behavior.

### DIFF
--- a/src/Layout/redist/MSBuildImports/Current/Microsoft.Common.targets/ImportAfter/Microsoft.JavaScript.targets
+++ b/src/Layout/redist/MSBuildImports/Current/Microsoft.Common.targets/ImportAfter/Microsoft.JavaScript.targets
@@ -1,0 +1,11 @@
+﻿<Project>
+
+  <Target Name="IgnoreJavaScriptOutputAssembly" BeforeTargets="AssignProjectConfiguration">
+      <ItemGroup>
+        <ProjectReference Condition="'%(Extension)' == '.esproj'">
+          <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+        </ProjectReference>
+      </ItemGroup>
+  </Target>
+
+</Project>


baronfel:
cc @rainersigwald @dsplaisted for review/insight here - @joj is from the Esproj team trying to make consuming `.esproj` projects nicer for other SDK-style projects in general. Right now, consumers of `.esproj` ProjectReferences have to markup every one of their references. 

Ideally this is something that could be done in by the esproj targets themselves, but these references are on the _consuming_ side.  We couldn't find another viable extension spot given the current ways we have to extend MSBuild, so this was the best thing that could be found.

dsplaisted:
Seems fine to me.  Does this need to work in non-SDK style consuming projects?  If not, it would probably be easier to put it somewhere in the .NET SDK targets, and you wouldn't have to worry about mirroring it for .NET CLI and full MSBuild.

joj:
I tried the ItemDefinitionGroup and couldn't make it work, but I do believe the target still makes a lot of sense. @dsplaisted I'm all for a better location for this targets file. Where would a good location in .Net SDK Targets be? I wasn't sure, so just looked for other well known common after targets :)

rainersigwald:
> this does in fact need to work for non-SDK Style projects

Then the best home for this is probably in https://github.com/dotnet/msbuild/blob/main/src/Tasks/Microsoft.Common.CurrentVersion.targets.

joj:
You mean modifying the actual common file? I can do that... I thought the bar was very high on that one :)

rainersigwald:
It is pretty high, but it's the only way to get common functionality between .NET SDK and VS. Were you thinking of shipping this ImportAfter to VS in a VSIX as well as here?

joj:
Moved to msbuild to live in currentversion targets

PR #34996, opened 2023-08-29
Create project templates target to 9.0

v-wuzhai:


v-wuzhai:
@MiYanni We still have to update dotnet sdk to 9.0 judging from these test failures.

- [ ] Update the sdk target to net9.0: #35031

MiYanni:
> @MiYanni We still have to update dotnet sdk to 9.0 judging from these test failures.

I'm not sure of the order of updates that need to occur. I'm new to the team and haven't been around for a major version rebranding before. Jacques or Daniel might have a better grasp on it.

v-wuzhai:
@baronfel @marcpopMSFT There are some container test failures like this: 

> Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.

Could you help to take a look?

baronfel:
These specific container tests are in an awkward position: They rely on dotnet/runtime-deps, dotnet/runtime, dotnet/aspnet base images being available with .NET and ASP.NET Runtimes matching the one being targeted for (which don't exist yet), and so they deliberately pin to versions behind the latest. We need a better story here for testing changes, but there aren't currently any SDK Container tooling features that are reliant on .NET 9, so we could probably just keep these tests pinned to .Net 8 (where we do have usable base images) and bump them to 9 at a later time, once we begin creating usable .NET 9 images from dotnet/dotnet-docker.

marcpopMSFT:
FYI, I'm waiting for main codeflow into installer to get unblocked before reviewing and approving.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/CreateNewImageTests.cs
```diff
@@ -133,7 +133,7 @@ public void ParseContainerProperties_EndToEnd()
     /// <summary>
     /// Creates a console app that outputs the environment variable added to the image.
     /// </summary>
-    [DockerAvailableFact]
+    [DockerAvailableFact(Skip = "https://github.com/dotnet/sdk/issues/36160")]
```
marcpopMSFT, line 136: if you'd kept the test targeting CurrentTargetFramework, would it have needed to be skipped?

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/CreateNewImageTests.cs
```diff
@@ -31,7 +31,7 @@ public void CreateNewImage_Baseline()
 
         newProjectDir.Create();
 
-        new DotnetNewCommand(_testOutput, "console", "-f", ToolsetInfo.CurrentTargetFramework)
+        new DotnetNewCommand(_testOutput, "console", "-f", ToolsetInfo.NextTargetFramework)
```
marcpopMSFT, line 34: Do we have an issue filed to track swapping 9.0 to current target framework (probably in Dec or Jan after all the templates have moved forward and 9 has stabilized some)?

src/Tests/dotnet-new.Tests/MSBuildEvaluationTests.cs
```diff
@@ -35,7 +35,8 @@ public void Class_BasicTest()
             cmd = new DotnetNewCommand(Log)
                 .WithCustomHive(tempSettingsDir)
                 .WithWorkingDirectory(projectPath)
-                .Execute("TestAssets.ClassTemplate", "--name", "MyTestClass");
+                // To run the template anyway, use '--force' option, refer to https://aka.ms/templating-exit-codes#100
```
marcpopMSFT, line 38: what was the reason for force being needed here and below?

src/Tests/dotnet-new.Tests/CommonTemplatesTests.cs
```diff
@@ -165,8 +165,8 @@ public void NuGetConfigPermissions()
         public async void AotVariants(string name, string language)
         {
             // templates have not be updated to 9 yet "net8.0";
```
marcpopMSFT, line 167: This comment should be updated since this PR is updating templates to 9. We could remove the comment and hardcoded line or change the comment to say something like "template framework needs to be hardcoded here during the major version transition"

template_feed/Microsoft.DotNet.Common.ProjectTemplates.9.0/content/ClassLibrary-CSharp/.template.config/localize/templatestrings.de.json
```diff
@@ -5,8 +5,8 @@
   "symbols/TargetFrameworkOverride/description": "Überschreibt das Zielframework",
   "symbols/TargetFrameworkOverride/displayName": "Zielframeworküberschreibung",
   "symbols/Framework/description": "Das Zielframework für das Projekt.",
-  "symbols/Framework/choices/net8.0/description": "Ziel net8.0",
-  "symbols/Framework/choices/net8.0/displayName": ".NET 8.0",
+  "symbols/Framework/choices/net9.0/description": "Target net9.0",
```
marcpopMSFT, line 8: See Kasper's cmment above for all of these files. Leave the translated string for Target in place in all of these loc files.

marcpopMSFT:
@v-wuzhai thank you for putting this together. It looks pretty good to me. I left a few comments. The biggest one is around the localization of Target. It looks like when Vlada did the 8 migration, she just changed everything back to english and then let the translators swap it back in a future onelocbuild.

So it sounds like it might be safe to leave as is but it also seems like it would save the translators time if we just left the translation as is. @baronfel @YuliiaKovalova do ya'll know why Vlada chose to swap the localized json files to have english (is there tooling that generates those and so they switched automatically?)

YuliiaKovalova:
> @v-wuzhai thank you for putting this together. It looks pretty good to me. I left a few comments. The biggest one is around the localization of Target. It looks like when Vlada did the 8 migration, she just changed everything back to english and then let the translators swap it back in a future onelocbuild.
> 
> So it sounds like it might be safe to leave as is but it also seems like it would save the translators time if we just left the translation as is. @baronfel @YuliiaKovalova do ya'll know why Vlada chose to swap the localized json files to have english (is there tooling that generates those and so they switched automatically?)

I have no idea here.
When I created item templates last year, I didn't swap anything - just committed the created templates with the localization files (https://github.com/dotnet/sdk/pull/29655/files#diff-cf06339338d8bb65d704739d4f01f726d7676c2102729696a1e01e3a1cfbeb76) and later it was translated by a dedicated tool in a separate PR https://github.com/YuliiaKovalova/sdk/commit/693354e733673334792813acb7c583c805f00dc8

marcpopMSFT:
So if we switched back to the loc templates, that'd be fine and it also might be fine to leave as is (as it'll get updated later)?

v-wuzhai:
> So if we switched back to the loc templates, that'd be fine and it also might be fine to leave as is (as it'll get updated later)?

Looks like there will be an update to these localized files later.
https://github.com/dotnet/sdk/commit/b4867728f2cde31f2be77eee7aa12229ceb5def4

marcpopMSFT:
I think the drawback is that it won't get fixed until we create a 9.0.1xx branch and target localization at that branch (so templates will be unlocalized until August). @baronfel thoughts on whether we should have Jason go fix all the localization now or just let it get fixed in normal flow in August?

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/CreateNewImageTests.cs
v-wuzhai: I tried to change the test to CurrentTargetFramework and got an error message. Like this:

>  StdErr:
> Error: Invalid option(s):
> -f net8.0
> 'net8.0' is not a valid value for -f. The possible values are:
> net9.0   - Target net9.0

YuliiaKovalova:
> So if we switched back to the loc templates, that'd be fine and it also might be fine to leave as is (as it'll get updated later)?

Yes, I think so,

marcpopMSFT (approved):

marcpopMSFT:
@kasperk81 thanks for catching that and the tip on reverting it.  @v-wuzhai thanks for updating the PR to include that. It looks good to me now.

@baronfel @MiYanni @joeloff any concerns before we merge this?

baronfel:
Nope! Let's get this thing rolling!

src/Tests/dotnet-new.Tests/Approvals/DotnetCSharpClassTemplatesTest.record.langVersion=9.0.verified/record/TestItem1.cs
```diff
@@ -1,6 +1,8 @@
 ﻿namespace ClassLib
 {
-    public record TestItem1
+    //Record was added in C# 9 and later, so Class was used instead. 
+    //See more info: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record
+    public class TestItem1
```
MiYanni, line 5: I'm confused here as the name of the test says it is testing records on C# 9:
`DotnetCSharpClassTemplatesTest.record.langVersion=9.0.verified/record/TestItem1.cs`

MiYanni (approved):

src/Tests/dotnet-new.Tests/Approvals/DotnetCSharpClassTemplatesTest.record.langVersion=9.0.verified/record/TestItem1.cs
v-wuzhai: I have pushed a change for this [67d76cc](https://github.com/dotnet/sdk/pull/34996/commits/67d76cc706074a949869952cc8441569d3b83ed7)

PR #35026, opened 2023-08-30
Fix possible NullReferenceException due to &/&& typo

j3parker:
I was looking for usages of `bool operator&(bool, bool)` in a few codebases and noticed a possible bug here. I didn't investigate in-depth (e.g. to see if `value` could ever actually be `null` here.)


j3parker:
👋

marcpopMSFT (approved):

danmoseley:
Nice catch!

Issue #35027, opened 2023-08-30
```dotnet run``` should populate a DOTNET_LAUNCH_PROFILE environment variable


mitchdenny:
We would like to extend ```dotnet run``` so that when it runs a project it populates an environment variable named ```DOTNET_LAUNCH_PROFILE```. We are working on some API surfaces for an upcoming project that will need this information to drive some default behaviors when launching sub-processes.

mitchdenny:
Done. Note that there is a lingering issue that we might want to look at for .NET 9.0:

#35197

PR #35029, opened 2023-08-30
Add DOTNET_LAUNCH_PROFILE environment when invoking dotnet run

mitchdenny:
Addresses #35027 

This PR modifies the ```dotnet run``` command to insert a ```DOTNET_LAUNCH_PROFILE``` environment variable based on the selected launch profile.

Given the following ```launchSettings.json``` file:

```json
{
  "profiles": {
    "First": {
      "commandName": "Project",
    },
    "Second": {
      "commandName": "Project",
    }
  }
}
```

And the following Program.cs file:

```csharp
var value = Environment.GetEnvironmentVariable("DOTNET_LAUNCH_PROFILE");
Console.WriteLine($"DOTNET_LAUNCH_PROFILE={value}");
```

# Launch profile specified and exists

```
$ dotnet run --launch-profile First
DOTNET_LAUNCH_PROFILE=First
```

# Launch profile not specified, first one selected

```
$ dotnet run
DOTNET_LAUNCH_PROFILE=First
```

# Launch profile specified but does not exist

```
$ dotnet run --launch-profile Third
The launch profile "Third" could not be applied.
A launch profile with the name 'Third' doesn't exist.
DOTNET_LAUNCH_PROFILE=
```

# Launch with no profile specified 

```
$ dotnet run --no-launch-profile
DOTNET_LAUNCH_PROFILE=
```

### DIFF
--- a/src/Assets/TestProjects/AppThatOutputsDotnetLaunchProfile/Program.cs
+++ b/src/Assets/TestProjects/AppThatOutputsDotnetLaunchProfile/Program.cs
@@ -1,0 +1,16 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace MSBuildTestApp
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var value = Environment.GetEnvironmentVariable("DOTNET_LAUNCH_PROFILE");
+            Console.WriteLine($"DOTNET_LAUNCH_PROFILE=<<<{value}>>>");
+        }
+    }
+}
--- a/src/BuiltInTools/dotnet-watch/DotNetWatcher.cs
+++ b/src/BuiltInTools/dotnet-watch/DotNetWatcher.cs
@@ -71,6 +71,7 @@ namespace Microsoft.DotNet.Watcher
                 context.RequiresMSBuildRevaluation = false;
 
                 processSpec.EnvironmentVariables["DOTNET_WATCH_ITERATION"] = (context.Iteration + 1).ToString(CultureInfo.InvariantCulture);
+                processSpec.EnvironmentVariables["DOTNET_LAUNCH_PROFILE"] = context.LaunchProfileName ?? string.Empty;
 
                 var fileSet = context.FileSet;
                 if (fileSet == null)
--- a/src/BuiltInTools/dotnet-watch/DotNetWatcher.cs
+++ b/src/BuiltInTools/dotnet-watch/DotNetWatcher.cs
@@ -71,7 +71,7 @@ namespace Microsoft.DotNet.Watcher
                 context.RequiresMSBuildRevaluation = false;
 
                 processSpec.EnvironmentVariables["DOTNET_WATCH_ITERATION"] = (context.Iteration + 1).ToString(CultureInfo.InvariantCulture);
-                processSpec.EnvironmentVariables["DOTNET_LAUNCH_PROFILE"] = context.LaunchProfileName ?? string.Empty;
+                processSpec.EnvironmentVariables["DOTNET_LAUNCH_PROFILE"] = context.LaunchSettingsProfile?.LaunchProfileName ?? string.Empty;
 
                 var fileSet = context.FileSet;
                 if (fileSet == null)
--- a/src/BuiltInTools/dotnet-watch/Program.cs
+++ b/src/BuiltInTools/dotnet-watch/Program.cs
@@ -198,6 +198,7 @@ namespace Microsoft.DotNet.Watcher
 
             var context = new DotNetWatchContext
             {
+                LaunchProfileName = launchProfileName,
                 HotReloadEnabled = enableHotReload,
                 ProcessSpec = new ProcessSpec
                 {
--- a/src/BuiltInTools/dotnet-watch/Program.cs
+++ b/src/BuiltInTools/dotnet-watch/Program.cs
@@ -198,7 +198,6 @@ namespace Microsoft.DotNet.Watcher
 
             var context = new DotNetWatchContext
             {
-                LaunchProfileName = launchProfileName,
                 HotReloadEnabled = enableHotReload,
                 ProcessSpec = new ProcessSpec
                 {
--- a/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ILaunchSettingsProvider.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ILaunchSettingsProvider.cs
@@ -1,13 +1,15 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#nullable enable
+
 using System.Text.Json;
 
 namespace Microsoft.DotNet.Tools.Run.LaunchSettings
 {
     internal interface ILaunchSettingsProvider
     {
-        LaunchSettingsApplyResult TryGetLaunchSettings(string launchProfileName, JsonElement model);
+        LaunchSettingsApplyResult TryGetLaunchSettings(string? launchProfileName, JsonElement model);
     }
 
 }
--- a/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ILaunchSettingsProvider.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ILaunchSettingsProvider.cs
@@ -7,7 +7,7 @@ namespace Microsoft.DotNet.Tools.Run.LaunchSettings
 {
     internal interface ILaunchSettingsProvider
     {
-        LaunchSettingsApplyResult TryGetLaunchSettings(JsonElement model);
+        LaunchSettingsApplyResult TryGetLaunchSettings(string launchProfileName, JsonElement model);
     }
 
 }
--- a/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
+++ b/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
@@ -40,12 +40,13 @@ namespace Microsoft.DotNet.Tools.Run.LaunchSettings
                         return new LaunchSettingsApplyResult(false, LocalizableStrings.LaunchProfilesCollectionIsNotAJsonObject);
                     }
 
+                    var selectedProfileName = profileName;
                     JsonElement profileObject;
                     if (string.IsNullOrEmpty(profileName))
                     {
-                        profileObject = profilesObject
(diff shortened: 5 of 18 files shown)


DamianEdwards:
I think we'll need to update `dotnet watch` too as it has its own processing logic for launch profiles and doesn't dispatch to `dotnet run`.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
```diff
@@ -40,12 +41,13 @@ public static LaunchSettingsApplyResult TryApplyLaunchSettings(string launchSett
                         return new LaunchSettingsApplyResult(false, LocalizableStrings.LaunchProfilesCollectionIsNotAJsonObject);
                     }
 
+                    var selectedProfileName = profileName;
```
nagilson, line 44: nit: if this is nullable, lean towards using nullable enable and marking the type with ?. I don't think profile name does it either, but really, it should...

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ProjectLaunchSettingsProvider.cs
```diff
@@ -11,9 +11,11 @@ internal class ProjectLaunchSettingsProvider : ILaunchSettingsProvider
 
         public string CommandName => CommandNameValue;
 
-        public LaunchSettingsApplyResult TryGetLaunchSettings(JsonElement model)
+        public LaunchSettingsApplyResult TryGetLaunchSettings(string launchProfileName, JsonElement model)
```
nagilson, line 14:
```suggestion
        public LaunchSettingsApplyResult TryGetLaunchSettings(string? launchProfileName, JsonElement model)
```

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ILaunchSettingsProvider.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.DotNet.Tools.Run.LaunchSettings
 {
     internal interface ILaunchSettingsProvider
     {
-        LaunchSettingsApplyResult TryGetLaunchSettings(JsonElement model);
+        LaunchSettingsApplyResult TryGetLaunchSettings(string launchProfileName, JsonElement model);
```
nagilson, line 10:
```suggestion
        LaunchSettingsApplyResult TryGetLaunchSettings(string? launchProfileName, JsonElement model);
```

src/Assets/TestProjects/AppThatOutputsDotnetLaunchProfile/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+
+namespace MSBuildTestApp
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var value = Environment.GetEnvironmentVariable("DOTNET_LAUNCH_PROFILE");
+            Console.WriteLine($"DOTNET_LAUNCH_PROFILE=<<<{value}>>>");
```
nagilson, line 13: What is the <<< and >>> for? Is that to separate the output so it's more clear it is the string coming from the console?

nagilson (approved):
A few nits, but this is overall pretty awesome! Great job on the tests and it's commendable you had a feature you wanted and then just.. added it. Thanks for this addition to the SDK, I quite love it, it's a good idea.

It would probably be wise to create an issue in the dotnet docs repo https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables for this, so it can be documented.

mitchdenny:
> I think we'll need to update `dotnet watch` too as it has its own processing logic for launch profiles and doesn't dispatch to `dotnet run`.

I'll do a follow up PR for this since this one is approved. But will start working on it straight away.

mitchdenny:
> It would probably be wise to create an issue in the dotnet docs repo https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables for this, so it can be documented.

https://github.com/dotnet/docs/issues/36937

src/Assets/TestProjects/AppThatOutputsDotnetLaunchProfile/Program.cs
mitchdenny: Its for the Contains check in the test assertion. I can't just assert on the presence of ```DOTNET_LAUNCH_PROFILE=``` for a null/empty result.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs
mitchdenny: Done.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ILaunchSettingsProvider.cs
mitchdenny: Done.

src/Cli/dotnet/commands/dotnet-run/LaunchSettings/ProjectLaunchSettingsProvider.cs
mitchdenny: Done.

mitchdenny:
Apologies for the noise on this PR I had to rebase my changes onto ```release/8.0.1xx```  but hadn't done the update to my branch before I repointed the PR itself. This PR just contains the changes I'm proposing now.

mitchdenny:
@nagilson @baronfel can I get another look at this PR after feedback from PR. Note that I repointed the PR to merge onto ```release/8.0.1xx``` because we need this to go on for .NET 8.0

/cc @DamianEdwards

src/BuiltInTools/dotnet-watch/Program.cs
```diff
@@ -198,6 +198,7 @@ private async Task<int> RunAsync(CommandLineOptions options, CancellationToken c
 
             var context = new DotNetWatchContext
             {
+                LaunchProfileName = launchProfileName,
```
tmat, line 201: Doesn't seem correct. If neither `--launch-profile` or `--no-launch-profile` are specified, the default (first that has Project command) profile from `launchSettings.json` is selected (if any). In that case `launchProfileName` will be null here but you'll want to set the env variable to the default profile name, no?

DamianEdwards: Right, this should be something like `LaunchProfileName = launchProfile.Name` instead? Which would require adding `Name` to `LaunchSettingsProfile`.

tmat: Yes, but then it would need to be excluded from JSON deserialization because the type is used for parsing the JSON file. 
Alternatively, `ReadLaunchProfile` can return both the `LaunchSettingsProfile` and the name in a tuple or a new `NamedLaunchSettingsProfile` type.

src/BuiltInTools/dotnet-watch/DotNetWatcher.cs
```diff
@@ -71,6 +71,7 @@ public async Task WatchAsync(DotNetWatchContext context, CancellationToken cance
                 context.RequiresMSBuildRevaluation = false;
 
                 processSpec.EnvironmentVariables["DOTNET_WATCH_ITERATION"] = (context.Iteration + 1).ToString(CultureInfo.InvariantCulture);
+                processSpec.EnvironmentVariables["DOTNET_LAUNCH_PROFILE"] = context.LaunchProfileName ?? string.Empty;
```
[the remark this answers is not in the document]
mitchdenny: I think we should "own" the environment variable which means we would set it to be whatever the evaluated value is - including being empty when no launch profile is used.

Allowing the environment variable to be inherited (and they not be the actual launch profile) would be even more confusing IMHO.

src/BuiltInTools/dotnet-watch/Program.cs
mitchdenny: I'll take a look at this. For dotnet run i did put the launch profile name on the launch profile settings. But the implementation of dotnet watch and dotnet run in this regard is quite different.

dotnet run actually had hand crafted parsing vs just using deserialisation.

mitchdenny: > Doesn't seem correct. If neither `--launch-profile` or `--no-launch-profile` are specified, the default (first that has Project command) profile from `launchSettings.json` is selected (if any). In that case `launchProfileName` will be null here but you'll want to set the env variable to the default profile name, no?

Give ```launchSettings.json``` as follows:

```json
{
  "profiles": {
    "First": {
      "commandName": "Project"
    },
    "Second": {
      "commandName": "Project"
    }
  }
}
```

If I invoke ```dotnet watch``` (with no arguments) then the environment variable ```DOTNET_LAUNCH_PROFILE``` will be populated with ```First```. Here is the testcase covering that:

```csharp
        [Fact]
        public async Task RunsWithDotnetLaunchProfileEnvVariableWhenNotExplicitlySpecified()
        {
            Assert.True(string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DOTNET_LAUNCH_PROFILE")), "DOTNET_LAUNCH_PROFILE cannot be set already when this test is running");

            var testAsset = TestAssets.CopyTestAsset(AppName)
                .WithSource()
                .Path;

            await App.StartWatcherAsync(testAsset);
            Assert.Equal("<<<First>>>", await App.AssertOutputLineStartsWith("DOTNET_LAUNCH_PROFILE = "));
        }
```

mitchdenny: Hrm ... I works, but I am not sure why it works. Investigating further.

mitchdenny: OK I reworked this to be a bit more consistent with what I did with ```dotnet run``` where I modified ```LaunchProfileSettings``` to include the launch profile name. I think the previous approach was working/would work but it was more of a side effect of the way that we did command-line parsing rather than an explicit capture of the launch profile name that we actually use.

tmat: > Hrm ... I works, but I am not sure why it works. Investigating further.

Note that there are two different modes `dotnet watch` can run in: 1) Hot Reload 2) legacy (any change rebuilds the app and restarts). They execute in a different way. The later uses `dotnet run` the former does not.

mitchdenny: Yeah. I've added logic to both code paths regardless.

adityamandaleeka:
@tmat @nagilson Can you take another look? The feedback has been addressed.

baronfel (approved):
This LGTM, and the tests are clear :+1:

nagilson (approved):
LGTM :) Would be good to get @tmat's approval though

Issue #34979, opened 2023-08-28
ContainerHttpException contains request instead of response


tmds:
I'm hitting this exception:

https://github.com/dotnet/sdk/blob/0a6f8335c92725de0e65df952f6940d3f7fadaf1/src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultManifestOperations.cs#L42-L45

The exception was meant to include the response from the server as its last argument, but currently the request json is being passed.

Side-note about the request: the JSON has its plus signs escaped. The server should still accept this as it is valid JSON.
```json
{
"schemaVersion": 2,
"mediaType": "application/vnd.docker.distribution.manifest.v2\u002Bjson",
"config": {
"mediaType": "application/vnd.docker.container.image.v1\u002Bjson",
...
```

cc @baronfel

baronfel:
Definitely a bug, would accept a PR that fixes this if I can't get around to sending one myself.

The escaped JSON I'd seen before, but it doesn't seem to have any impact on the registries I've been testing against. If it does end up being a problem we'd need to scour our uses of the STJ json serializer and use a consistent set of options that forbid this encoding.

tmds:
Fixed by https://github.com/dotnet/sdk/pull/35040.

PR #35040, opened 2023-08-30
Containers: DefaultManifestOperations.PutAsync: log failure response and include status code in exception.

tmds:
Fixes https://github.com/dotnet/sdk/issues/34979.

@baronfel ptal.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/ContainerHttpException.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/ContainerHttpException.cs
@@ -6,12 +6,10 @@ namespace Microsoft.NET.Build.Containers;
 internal sealed class ContainerHttpException : Exception
 {
     private const string ErrorPrefix = "Containerize: error CONTAINER004:";
-    string? jsonResponse;
     string? uri;
-    public ContainerHttpException(string message, string? targetUri, string? jsonResp)
-            : base($"{ErrorPrefix} {message}\nURI: {targetUri ?? "Unknown"}\nJson Response: {jsonResp ?? "None."}")
+    public ContainerHttpException(string message, string? targetUri)
+            : base($"{ErrorPrefix} {message}\nURI: {targetUri ?? "Unknown"}")
     {
-        jsonResponse = jsonResp;
         uri = targetUri;
     }
 }
--- a/src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultManifestOperations.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultManifestOperations.cs
@@ -41,7 +41,8 @@ internal class DefaultManifestOperations : IManifestOperations
 
         if (!putResponse.IsSuccessStatusCode)
         {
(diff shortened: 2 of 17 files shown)


src/Containers/Microsoft.NET.Build.Containers/ContainerHttpException.cs
```diff
@@ -6,12 +6,10 @@ namespace Microsoft.NET.Build.Containers;
 internal sealed class ContainerHttpException : Exception
 {
     private const string ErrorPrefix = "Containerize: error CONTAINER004:";
-    string? jsonResponse;
     string? uri;
-    public ContainerHttpException(string message, string? targetUri, string? jsonResp)
-            : base($"{ErrorPrefix} {message}\nURI: {targetUri ?? "Unknown"}\nJson Response: {jsonResp ?? "None."}")
+    public ContainerHttpException(string message, string? targetUri)
```
baronfel, line 10: I don't think we should remove the `jsonResp` here - registries often have json payload describing what went wrong with the upload that can be helpful.

src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultManifestOperations.cs
```diff
@@ -41,7 +41,8 @@ public async Task PutAsync(string repositoryName, string reference, ManifestV2 m
 
         if (!putResponse.IsSuccessStatusCode)
         {
-            throw new ContainerHttpException(Resource.GetString(nameof(Strings.RegistryPushFailed)), putResponse.RequestMessage?.RequestUri?.ToString(), jsonString);
+            await putResponse.LogHttpResponseAsync(_logger, cancellationToken).ConfigureAwait(false);
+            throw new ContainerHttpException(Resource.FormatString(nameof(Strings.RegistryPushFailed), putResponse.StatusCode), putResponse.RequestMessage?.RequestUri?.ToString());
```
baronfel, line 45: instead of passing `jsonString` here we should be passing something like `await putResponse?.Content?.ReadAsStringAsync().ConfigureAwait(false)` so that we can get the payload returned by the server. These often have details that can help diagnose issues.

baronfel (changes requested):

src/Containers/Microsoft.NET.Build.Containers/ContainerHttpException.cs
baronfel: Ah, I see what we started doing here - we started just logging the full details and hiding the raw responses from the user. See the `LogHttpResponseAsync` extension method used in `DefaultBlobUploadOperations` for an example. In that case, please do the following:
* keep your code the same
* in the DefaultManifestOperations call that throws the ContainerHttpException, could you call the `LogHttpResponseAsync` extension to log the data to the binlog?

That would keep the analysis data around, and clean up the UX for users.

tmds: > could you call the LogHttpResponseAsync extension to log the data to the binlog?

It's already there :)

baronfel: Good grief, I am bad at comprehension this morning. Thank you for walking me through that. In that case this PR is 100% perfect, no notes.

baronfel (approved):

PR #35044, opened 2023-08-30
Fix workload info

joeloff:
# Description

When workloads are installed by VS, `dotnet --info` and `dotnet workload --info` fail to display the installed workloads. Unlike list, the commands are only checking if there are workloads installed by the SDK. If so, they will then display the full set of installed workloads, including the ones from Visual Studio.

Below is from an internal build of VS with RC1:

`dotnet --info`

image

`dotnet workload list`

image


nagilson (approved):
LGTM... looked at the blame and it looks like it was me who committed that one. Whoops!

joeloff:
> LGTM... looked at the blame and it looks like it was me who committed that one. Whoops!

And I probably approved it

PR #35088, opened 2023-09-01
Add automatically tool upgrade and downgrade for different versions for dotnet tool install

JL03-Yue:
https://github.com/dotnet/sdk/issues/9500

Update dotnet tool local install so that it support automatic upgrade, downgrade, and error-throwing behaviors.

- If the command specify version 3.0.0 without --allow-downgrade flag and version 4.0.0 is installed, it throws out an error.
- If the command specify version 3.0.0 with--allow-downgrade flag and version 4.0.0 is installed, downgrade to 3.0.0. Exit code = 0


JL03-Yue:
@dsplaisted @baronfel Hi! I found a long hanging pr #31549 that is supposed to automatically upgrade and downgrade a tool given specified version. I thought this feature is quite useful because it allows users more flexibility in terms of handling tool versions and less prone to errors, so I recreated this pr and refine some of the checks. Is it possible if someone can review it? I'd love to resolve follow-ups as well~

Besides, this change only allows automatic upgrade and downgrade to local tools, I wonder if we would like to expand it some time to global tool as well~? Depending on business priorities:D

dsplaisted:
I haven't reviewed the code yet, but the behavior from the PR description sounds good.  I think global and local tool installs should both behave this way.

Right now it looks like this code is failing to compile.

PR #35107, opened 2023-09-01
Add --no-http-cache option and hide --no-cache option for dotnet restore command

Nigusu-Allehu:
This PR will
- Add a new option (--no-http-cache)
- Remove (--no-cache) from the help output
- The NuGet side would be implemented here : https://github.com/NuGet/NuGet.Client/pull/5389


src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
```diff
@@ -147,10 +147,16 @@ private static IEnumerable<CliOption> ImplicitRestoreOptions(bool showHelp, bool
 
             yield return new ForwardedOption<bool>("--no-cache")
             {
-                Description = showHelp ? LocalizableStrings.CmdNoCacheOptionDescription : string.Empty,
-                Hidden = !showHelp
+                Description = string.Empty,
```
kartheekp-ms, line 150: I think we should still show some description. May be update the string to notify customers that this option is in deprecation path and recommend to use the new option (--no-http-cache) instead.

Nigusu-Allehu: I was thinking we should hide --no-cache from the help output so that no new users use it.  However, we do show --no-http-cache in the help output

kartheekp-ms:
I would be great to have some test coverage for this change. There are several references to `--no-cache` option in this repo https://github.com/search?q=repo%3Adotnet%2Fsdk+--no-cache+language%3AC%23&type=code&l=C%23 few references under source code and few of them under test. I guess we should check these references and see if they work with the new option.

kartheekp-ms:
PR title says `add --no-http-cache option` but I think we are also hiding `--no-cache` option.  Please consider updating the title so that it would be helpful for future reference.

src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
kartheekp-ms: I agree with the hiding part, but I was wondering how would customers using `--no-cache` currently know about the new option. I see we added an information message to the restore log if customers use the old option https://github.com/NuGet/NuGet.Client/pull/5389/files#diff-b43b40cbcbf3bef2bbf71983856b74834ffa03eb82215efb5a132c42b75f9d18R152. Maybe we should be okay.

On a side note, I think we should also update the docs https://learn.microsoft.com/dotnet/core/tools/dotnet-restore. Please feel free to create an issue in the [NuGet docs repo](https://github.com/NuGet/docs.microsoft.com-nuget/issues). In the past when an option is deprecated for dotnet run there was a blog post explaining the reason for changing the option name https://learn.microsoft.com/dotnet/core/compatibility/sdk/6.0/deprecate-p-option-dotnet-run.

Nigusu-Allehu: Just created a PR for updating the docs : https://github.com/NuGet/docs.microsoft.com-nuget/pull/3163 .

nkolev92 (approved):

PR #35138, opened 2023-09-04
Container Build: Handle multiple tags wtihin one image

### LINKED ISSUE #379: Change Request: One Image for all tags
As part of #375  I realized that sdk-container-builds creates a new image for every tag. I think it would be better to create only one single tar.gz with all tags set. This would: 

* Reduce the load needed to generate the tar
* Reduce the load on the daemon side as only 1 docker load needs to happen
* Eliminate the problem that one application currently has N output tars. 

The change should be valid from a [Docker Image Specification v1.2.0](https://github.com/moby/moby/blob/master/image/spec/v1.2.md) side.

Danielku15:
Fixes https://github.com/dotnet/sdk-container-builds/issues/379 

Main changes: 

- Split of `ImageReference` into `SourceImageReference` (single tag) and `DestinationImageReference` (multiple tags)
- Extension of remote registry push to push the manifest for each tag
- Extension of local docker registry implementation to contain all tags in one `manifest.json`
- Extended EndToEndTests to use multiple tags to verify functionality.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
@@ -43,11 +43,11 @@ public static class ContainerBuilder
 
         bool isLocalPull = string.IsNullOrEmpty(baseRegistry);
         Registry? sourceRegistry = isLocalPull ? null : new Registry(baseRegistry, logger);
-        ImageReference sourceImageReference = new(sourceRegistry, baseImageName, baseImageTag);
+        SourceImageReference sourceImageReference = new(sourceRegistry, baseImageName, baseImageTag);
 
         bool isLocalPush = string.IsNullOrEmpty(outputRegistry);
         Registry? destinationRegistry = isLocalPush ? null : new Registry(outputRegistry!, logger);
-        IEnumerable<ImageReference> destinationImageReferences = imageTags.Select(t => new ImageReference(destinationRegistry, imageName, t));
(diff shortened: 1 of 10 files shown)


baronfel:
This is great work @Danielku15 - the test failure doesn't look related so I've re-triggered it.

baronfel (approved):
Great use of types to separate intent - happy to see that this was a natural evolution of the codebase :+1:

baronfel:
@Danielku15  currently this is targeting `main`, which is for .NET 9 now that we've 'snapped' for release. If you want this in .NET 8 RC2, please retarget to `release/8.0.1xx`.

Danielku15:
@baronfel Retargeted to 8.0.1xx. Looks like I might have accidentally triggered with the retargeting a review of quite many people, sorry for that.

baronfel:
Code still looks good to me after the retargeting. I'm going to do a run of baronfel/sdk-container-demo with this change as a smoke test, and if that's green merge it. It should be part of RC2 in October when merged.

Danielku15:
@baronfel Great to hear 🥳 . I also prepared already the "tar-gz" file writing on top of this branch. Once this is merged I will likely open the next PR 😉 Then I should have a foundation for the multi-registry push based on [this idea](https://github.com/dotnet/sdk-container-builds/discussions/495#discussioncomment-6905628) to use the tar-gz + further content in it as a foundation to create a new push task devs can call. 
 
Would love to see those features in net8.0

baronfel:
This PR looks great in the linked actions run - and as expected it's making the multi-tag scenarios _much_ faster. Thanks again for this great contribution!

PR #35146, opened 2023-09-05
Update Verify.Xunit package version

v-wuzhai:
1. Force update the xunit.core package to 2.5.0.
2. Add NU1608 and NETSDK1023 to NoWarn to avoid warnings.

### DIFF
--- a/src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
+++ b/src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
@@ -8,10 +8,18 @@
     <IsUnitTestProject>true</IsUnitTestProject>
     <EnableStyleCopAnalyzer>true</EnableStyleCopAnalyzer>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+    <!-- NU1608: Disable dependency constraints. Make packages referenced by the project directly rather than packages from other sources. -->
+    <!-- NETSDK1023: Disable implicit reference so that packages can be referenced in the project. -->
+    <NoWarn>$(NoWarn);NU1608;NETSDK1023</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
     <None Include="TemplateInstallTests/TestTemplates/**/*" CopyToOutputDirectory="Always" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="xunit.core" VersionOverride="2.5.0" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


marcpopMSFT:
@v-wuzhai is this meant as a replacement for the two dependabot PRs?  If so, should we target 8.0.1xx and let this flow to main rather than worrying about 7.0.4xx?

src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
```diff
@@ -8,12 +8,20 @@
...
+    <!-- NETSDK1023: Disable implicit reference so that packages can be referenced in the project. -->
+    <NoWarn>$(NoWarn);NU1608;NETSDK1023</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
     <None Include="TemplateInstallTests/TestTemplates/**/*" CopyToOutputDirectory="Always" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="xunit.core" VersionOverride="2.5.0" />
```
marcpopMSFT, line 22: Could the version go in the directory.packages.props?

src/Tests/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
```diff
@@ -8,12 +8,20 @@
     <IsUnitTestProject>true</IsUnitTestProject>
     <EnableStyleCopAnalyzer>true</EnableStyleCopAnalyzer>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+    <!-- NU1608: Disable dependency constraints. Make packages referenced by the project directly rather than packages from other sources. -->
+    <!-- NETSDK1023: Disable implicit reference so that packages can be referenced in the project. -->
+    <NoWarn>$(NoWarn);NU1608;NETSDK1023</NoWarn>
```
marcpopMSFT, line 14: I assume NU1608 is needed because this is overriding the version that's in arcade, correct?  Why is 1023 needed those? I'm surprised that adding this reference is triggering that since that should only trigger for packages that come with the runtime

v-wuzhai: It looks like the arcade version is covered, maybe we should just have to wait for the arcade update. Aslo, I'm confused why the error information is different between 7.0.4xx and 8.0.1xx. Is it due to CPM?

PR #35190, opened 2023-09-06
Update Asp.Net FrameworkReference in BrowserRefresh

wtgodbe:
This project needs to stay at net6.0, which creates issues for source build, since that means we need to include the 6.0 targeting pack in source build. To minimize churn, we pin at the lowest possible version of the targeting pack, which is 6.0.2 due to some unintended API churn early in 6.0.


mkArtakMSFT (approved):
Thanks @wtgodbe !

wtgodbe:
We just need to flow https://github.com/dotnet/source-build-reference-packages/pull/778 into here, right?

MichaelSimons:
> We just need to flow [dotnet/source-build-reference-packages#778](https://github.com/dotnet/source-build-reference-packages/pull/778) into here, right?

Correct.

src/BuiltInTools/BrowserRefresh/Microsoft.AspNetCore.Watch.BrowserRefresh.csproj
```diff
@@ -14,4 +14,11 @@
     <EmbeddedResource Include="BlazorHotReload.js" />
   </ItemGroup>
 
+  <!-- Reference 6.0.2 Asp.Net targeting pack in Source Build - this was the earliest version after the API surface stabilized -->
+  <ItemGroup Condition="'$(DotNetBuildFromSource)' == 'true'">
+    <FrameworkReference
```
MichaelSimons, line 19: Does this affect the Microsoft.NETCore.App.Ref version as well?  <Usage Id="Microsoft.NETCore.App.Ref" Version="6.0.20" /> is being reported as a prebuilt.  Wondering if a another FwRef is needed to address that?

wtgodbe: It shouldn't - what version is being reported as a prebuilt? We can hard-code to 6.0.0 if we need to

MichaelSimons:
```
    <Usage Id="Microsoft.NETCore.App.Ref" Version="6.0.20" />
```

mthalman: This should be 6.0.2 to match AspNetCore. That's the version I added in SBRP.

wtgodbe: Updated

MichaelSimons (approved):

mthalman (approved):

PR #35204, opened 2023-09-07
Container Build: Add OCI Archive Format writing

Danielku15:
This is a follow up of https://github.com/dotnet/sdk/pull/35151 and partially relating to https://github.com/dotnet/sdk-container-builds/issues/376

This format is a lot closer to what registries also expect on the Docker Registry HTTP API. It contains blobs organized in digests, and an index.json describing the manifests and tags relating to the container.

Important detail regarding the test: I plan to implement pushing of OCI tars to registries via a new MSBuild task in an upcoming PR. With this feature then we can improve the EndToEndTest around the OCI container to push with all tags and run all flavours. A sneak peek of the ongoing implementation for that can be seen here: https://github.com/Danielku15/dotnet-sdk/compare/feature/oci-archive...Danielku15:dotnet-sdk:feature/custom-push-task


Danielku15:
@baronfel Do you have an idea how we could reach from a `docker run` command the `registry` in the test infrastructure? locally with docker desktop we have some special hostnames available to reach the host. 
image

We might only need a temporary solution until the next PR with the custom push task is available.

Danielku15:
Any chance this PR (and likely my next one) make it to net8.0 so we have OCI and manual registry push capabilities of these? If not I need to start searching alternatives for us like maintaining a fork which uses different parameters. 
Unfortunately since sdk-container-builds is part of the .net SDK we have conflicts using a custom build consumed via NuGet as the SDK built-in variant activates. 

Ping @baronfel

marcpopMSFT:
Old PR triage: @dotnet/sdk-container-builds-maintainers this is old and targeting 8.0.1xx. Can you engage and determine if this should be closed or refreshed targeting a newer branch?

baronfel:
Can close for now and refresh/retarget

PR #35231, opened 2023-09-08
Add support for Haiku

trungnt2910:
This PR contains the changes to the `PortableRuntimeIdentifierGraph.json` as suggested in https://github.com/dotnet/runtime/pull/86391#discussion_r1318538026, as well as some small changes to the CLI tools to allow a Haiku-aware SDK.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Utils/RuntimeEnvironment.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Utils/RuntimeEnvironment.cs
@@ -11,7 +11,8 @@ namespace Microsoft.DotNet.Cli.Utils
         Darwin = 3,
         FreeBSD = 4,
         illumos = 5,
-        Solaris = 6
+        Solaris = 6,
+        Haiku = 7
     }
 
     internal static class RuntimeEnvironment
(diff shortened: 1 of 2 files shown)


dsplaisted (approved):
This looks fine, but per [@ViktorHofer's comment](https://github.com/dotnet/runtime/pull/86391#discussion_r1319653325) we may want to move the portable graph to the dotnet/runtime repo instead.

marcpopMSFT:
@trungnt2910 did you want us to merge this or wait for a decision on where hte portable rid graph should live first? Also, this is in main which is a 9.0 branch and would need to be backported if desired in 8.0.100

CC @elinor-fung @ViktorHofer

trungnt2910:
> did you want us to merge this or wait for a decision on where hte portable rid graph should live first?

If this could be merged right now, then the Haiku changes should move around with the portable RID graph, so I would prefer that.

> Also, this is in main which is a 9.0 branch

I am aware of that; that's why the base branch is named `haiku-dotnet9`.

Seems like .NET 8 is already about to be released in a couple of months, so there's not much chance all the other changes required to get the [dotnet/runtime](https://github.com/dotnet/runtime) repo working on Haiku to land on .NET 8. I therefore don't think there's much need to backport, unless the 9.0 branch won't be used by `dotnet/runtime` for quite a while.

baronfel:
That's reasonable to me - if the changes required to support Haiku were only SDK-level then you'd have plenty of opportunity during the 8 cycle to onboard. I don't think that we do OS support changes at a runtime-level mid-cycle though - @richlander do you have any thoughts there?

jkotas:
> I don't think that we do OS support changes at a runtime-level mid-cycle though 

Correct. Haiku support is not something that can be added in .NET 8 runtime servicing.

jkotas (approved):

PR #35246, opened 2023-09-08
[Containers] Validate that user applications will run on the selected base image

baronfel:
Fixes https://github.com/dotnet/sdk-container-builds/issues/499

When we select an image during multi-image inference we do RID negotiation to select the best base image. This acts as a form of validation that the application assets can actually run on the base image chosen. However, we don't do any such validation on the single-image scenario. This means that users who select a single specific image as their base can easily create containers that do not actually run. 

This PR adds a validation check that the user's RuntimeIdentifier is compatible with the final-selected base image. To do this, we needed to add a couple missing properties that are present on the Image Configuration so that we could infer PlatformInformation for the base image. With that information, we could make use of existing functions to do the check.

Handling these error message in a unified way was resulting in a lot of `catch` blocks, so I refactored some of the messages into a new `FormattableException` base class, whose job is to hold the name of the resource string and the format arguments so that we can apply them whenever the message needs to be rendered out. This makes error handling on the CLI app and the MSBuild task more uniform.


src/Containers/Microsoft.NET.Build.Containers/Exceptions/FormattableException.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    internal FormattableException(string resourceName, params object[] formatArgs) {
+        ResourceName = resourceName;
+        FormatArguments = formatArgs;
+    }
+    
+    internal object[] FormatArguments { get; init; }
+
+    internal string ResourceName { get; init; }
+}
```
baronfel, line 18: New base class for messages that are backed by format strings

src/Containers/Microsoft.NET.Build.Containers/Exceptions/UnsupportedPlatformException.cs
```diff
@@ -0,0 +1,13 @@
...
+using Microsoft.NET.Build.Containers.Resources;
+
+namespace Microsoft.NET.Build.Containers;
+
+internal sealed class UnsupportedPlatformException : FormattableException
+{
+    internal UnsupportedPlatformException(string repositoryName, string reference, PlatformInformation platform):base(nameof(Strings.BaseImageUnsupportedPlatform), repositoryName, reference, platform)
+    {
+    }
+}
```
baronfel, line 13: look at how easy it is to make new errors now!

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -66,6 +66,11 @@ public static class ContainerBuilder
                     ridGraphPath,
                     cancellationToken).ConfigureAwait(false);
             }
+            catch (FormattableException e)
+            {
+                logger.LogError(String.Format(e.ResourceName, e.FormatArguments));
+                return 1;
+            }
```
baronfel, line 73: Ooooh, ahhhh - unified behavior/handling of all of these errors.

src/Containers/Microsoft.NET.Build.Containers/ImageConfig.cs
```diff
@@ -53,21 +61,30 @@ internal ImageConfig(JsonNode config)
...
-    // Return values from the base image config.
     internal string? GetUser() => _config["config"]?["User"]?.ToString();
     internal string[]? GetEntrypoint() => _config["config"]?["Entrypoint"]?.AsArray()?.Select(node => node!.GetValue<string>())?.ToArray();
     private string[]? GetCmd() => _config["config"]?["Entrypoint"]?.AsArray()?.Select(node => node!.GetValue<string>())?.ToArray();
     private List<HistoryEntry> GetHistory() => _config["history"]?.AsArray().Select(node => node.Deserialize<HistoryEntry>()!).ToList() ?? new List<HistoryEntry>();
     private string GetOs() => _config["os"]?.ToString() ?? throw new ArgumentException("Base image configuration should contain an 'os' property.");
     private string GetArchitecture() => _config["architecture"]?.ToString() ?? throw new ArgumentException("Base image configuration should contain an 'architecture' property.");
+    private string? GetVariant() => _config["variant"]?.ToString();
+    private string? GetOsVersion() => _config["os.version"]?.ToString();
+    private string[]? GetOsFeatures() => _config["os.features"]?.AsArray()?.Select(node => node!.GetValue<string>())?.ToArray();
```
baronfel, line 87: In order to check platform compatibility, we needed to read the various fields that are part of the 'platform information' from the base image config. We had some of them before, but we needed these to complete the picture.

src/Containers/Microsoft.NET.Build.Containers/ImageConfig.cs
```diff
@@ -37,6 +40,11 @@ internal sealed class ImageConfig
     public ReadOnlyDictionary<string, string> EnvironmentVariables => _environmentVariables.AsReadOnly();
     public HashSet<Port> Ports => _exposedPorts;
 
+    /// <summary>
+    /// Represents the platform information of this image, including architecture, os, variant, os features and os version.
+    /// </summary>
+    public PlatformInformation PlatformInformation => new(_architecture, _os, _variant, _osFeatures ?? [], _osVersion);
```
baronfel, line 46: With all of the new data, we can synthesize our PlatformInformation structure for compat-checking purposes.

src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
```diff
@@ -129,31 +133,48 @@ public async Task<ImageBuilder> GetImageManifestAsync(string repositoryName, str
...
+        if (configRid is null)
+        {
+            throw new UnsupportedPlatformException(repositoryName, reference, imageConfig.PlatformInformation);
+        }
+        var configRids = new[] { configRid };
+        var bestManifestRid = GetBestMatchingRid(ridGraph, applicationRid, configRids);
+        if (bestManifestRid is null)
+        {
+            throw new BaseImageNotFoundException(applicationRid, repositoryName, reference, configRids);
+        }
```
baronfel, line 158: And here we are, checking compatibility to help give the user some guardrails.

rainersigwald: I think this guardrail makes sense for almost everyone, but should there be a way to disable it? I'm worried about breaking someone who "knows what they're doing".

src/Containers/Microsoft.NET.Build.Containers/Resources/Strings.resx
```diff
@@ -26,79 +26,79 @@
...
             <xsd:complexType>
               <xsd:sequence>
-                <xsd:element name="value" type="xsd:string" minOccurs="0" />
+                <xsd:element name="value" type="xsd:string" minOccurs="0"/>
               </xsd:sequence>
-              <xsd:attribute name="name" use="required" type="xsd:string" />
-              <xsd:attribute name="type" type="xsd:string" />
-              <xsd:attribute name="mimetype" type="xsd:string" />
-              <xsd:attribute ref="xml:space" />
+              <xsd:attribute name="name" use="required" type="xsd:string"/>
```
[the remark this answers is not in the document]
baronfel: ugh yeah - there's a VSCode resource editor that is...opinionated.

src/Containers/Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -81,16 +86,17 @@ public static class ContainerBuilder
...
+            {
+                logger.LogError(e, "Unexpected error while fetching image");
+                throw;
+            }
         }
         else
         {
             throw new NotSupportedException(Resource.GetString(nameof(Strings.ImagePullNotSupported)));
         }
-        if (imageBuilder is null)
```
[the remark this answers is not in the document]
baronfel: the compiler said it was ok - the previous case for checking this was that the `try/catch` was not exhaustive. it now is, and so now we can say that if we reach this line, then `imageBuilder` is definitely not null.

src/Containers/Microsoft.NET.Build.Containers/Exceptions/UnsupportedPlatformException.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.NET.Build.Containers.Resources;
+
+namespace Microsoft.NET.Build.Containers;
+
+internal sealed class UnsupportedPlatformException : FormattableException
```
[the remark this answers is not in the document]
baronfel: can do!

src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
baronfel: This is a great point - I should find a way to translate this error into a warning-with-code instead of a hard-stop.

tmds:
Since we've moved to start the apps using `dotnet <assembly.dll>` instead of the apphost, framework dependent deployments are no longer rid dependent, so they should work on 'any' runtime/aspnet base image.

We may want to avoid printing a warning/error for those based on the assumed rid of `linux`:

https://github.com/dotnet/sdk/blob/b575887007e9390607cdff70621cdbeb59896314/src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets#L92-L95

baronfel:
Going to close this as it needs rework both due to conflicts, and to take @tmds's changes regarding entrypoint into account.

Issue #35248, opened 2023-09-08
`dotnet help` dumps stack in GitHub Codespaces


rainersigwald:
### To Reproduce
In a Codespace, run `dotnet help pack`.

### Exceptions (if any)

```sh-session
$ dotnet help pack
Unhandled exception: System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'xdg-open' with working directory '/workspaces/source-build-externals'. No such file or directory
   at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at Microsoft.DotNet.Tools.Help.HelpCommand.Execute()
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at Microsoft.DotNet.Cli.Parser.<>c__DisplayClass17_0.<<UseParseErrorReporting>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.CommandLineBuilderExtensions.<>c__DisplayClass11_0.<<UseHelp>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.CommandLineBuilderExtensions.<>c.<<UseSuggestDirective>b__17_0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.CommandLineBuilderExtensions.<>c__DisplayClass15_0.<<UseParseDirective>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.CommandLineBuilderExtensions.<>c__DisplayClass7_0.<<UseExceptionHandler>b__0>d.MoveNext()
```

### Further technical details
- Include the output of `dotnet --info`
```
.NET SDK:
 Version:   7.0.306
 Commit:    f500069cb7

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  20.04
 OS Platform: Linux
 RID:         ubuntu.20.04-x64
[...]

PR #35255, opened 2023-09-08
Make HelpCommand use the runtime to launch HTML

baronfel:
Fixes https://github.com/dotnet/sdk/issues/35248

This does a few specific things:
* defer to the runtime to launch the help pages, instead of using older lists of what tools to use for this
* if that launch process fails, fallback to emitting the CLI help
* if help is called on a command that isn't a documented command (e.g. `dotnet workload install`) emit the CLI help
* BONUS: we also support using `dotnet help` with subcommands now

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-help/HelpCommand.cs
+++ b/src/Cli/dotnet/commands/dotnet-help/HelpCommand.cs
@@ -23,7 +23,7 @@ namespace Microsoft.DotNet.Tools.Help
 
             result.ShowHelpOrErrorIfAppropriate();
 
-            if (!string.IsNullOrEmpty(result.GetValue(HelpCommandParser.Argument)))
+            if (result.GetValue(HelpCommandParser.Arguments) is { } args)
             {
                 return new HelpCommand(result).Execute();
             }
@@ -46,10 +46,12 @@ namespace Microsoft.DotNet.Tools.Help
 
         public static Process ConfigureProcess(string docUrl)
         {
-            return Process.Start(new ProcessStartInfo()
+            return new Process() { StartInfo = new ProcessStartInfo()
             {
-                FileName = docUrl
-            });
+                FileName = docUrl,
+                UseShellExecute = true
+            }
+            };
         }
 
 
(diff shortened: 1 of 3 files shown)


src/Cli/dotnet/commands/dotnet-help/HelpCommand.cs
```diff
@@ -23,7 +23,7 @@ public static int Run(ParseResult result)
 
             result.ShowHelpOrErrorIfAppropriate();
 
-            if (!string.IsNullOrEmpty(result.GetValue(HelpCommandParser.Argument)))
+            if (result.GetValue(HelpCommandParser.Arguments) is { } args)
```
MiYanni, line 26: I don't think you need the variable name if you don't use it... or using the discard, `_`, as the name.

src/Cli/dotnet/commands/dotnet-help/HelpCommand.cs
```diff
@@ -46,75 +46,73 @@ public static void PrintVersionHeader()
...
+                    }
+                    else
+                    {
+                        return innerParseResult.ShowHelp();
+                    }
+                }
+                else
+                {
+                    return innerParseResult.ShowHelp();
+                }
```
MiYanni, line 103: Code path simplification suggestion. Also, the `catch` should really have some kind of exception type with it. Catching everything is generally not recommended.

```suggestion
                if (commandResult.Command is DocumentedCommand documentedCommand
                    && !string.IsNullOrEmpty(documentedCommand.DocsLink))
                {
                    var process = ConfigureProcess(documentedCommand.DocsLink);
                    try
                    {
                        process.Start();
                        process.WaitForExit();
                        if (process.ExitCode == 0)
                        {
                            return process.ExitCode;
                        }
                    }
                    catch { }
                }

                // if the 'open browser' process fails, we should fallback to
                // calling `--help` for that command and outputting that
                return innerParseResult.ShowHelp();
```

baronfel: Valid point on the `catch` - but two points that led me to not care as much:
* this is deliberately a fall-back - if _anything_ went wrong launching the browser, then we should just emit the CLI help. We're not doing any diagnostics or trying to recover here at all.
* for urls, the most common exception I see happening here is on Linux the various shell-opening utilities may not be installed. That'll be raised as an exception by the runtime, and that's the only exceptional case I'd ever really expect to see here.

baronfel:
@KalleOlaviNiemitalo I like that suggestion - apps like `git` don't need to do that because they're always launching local HTML files, but since the SDK doesn't ship those files we might need some kind of fallback.

marcpopMSFT:
Old PR triage: @baronfel this is old and targeting 8.0.1xx. Should this be refreshed and target a newer branch or closed?

baronfel:
Can close - I can revive if I get time.

PR #35263, opened 2023-09-09
RunCommand prints "Building..." by default if it needs to build

epidemicz:
When ```dotnet run``` needs to build the project it will now print the ```Building...``` message by default unless dotnetRunMessages is set to false.

- added test project that has dotnetRunMessages set to false
- added a test to ensure it is printed when dotnetRunMessage is true
- added a test to ensure it isn't printed when dotnetRunMessage is false
- changed test to ensure it is printed by default/no launch settings

I found it confusing that certain project templates would print the ```Building...``` message and others wouldn't.

It seems more consistent to have ```dotnet run``` print ```Building...``` by default.  This also mirrors the behavior of ```dotnet watch```

### DIFF
--- a/src/Assets/TestProjects/TestAppThatWaits/Properties/launchSettings.json
+++ b/src/Assets/TestProjects/TestAppThatWaits/Properties/launchSettings.json
@@ -1,0 +1,12 @@
+{
+  "profiles": {
+    "TestAppThatWaits": {
+      "commandName": "Project",
+      "dotnetRunMessages": "false",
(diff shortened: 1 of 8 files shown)


epidemicz:
@dotnet-policy-service agree

baronfel:
Dotnet run should not show any build messages by default - specifically because users use this command as a shorthand for actually invoking their applications. This differs greatly from watch in intent.

epidemicz:
You're right, I don't want to break anybody.  Well a little, but it's probably not for the best.  Thanks for the feedback, I think I will look into a PR  to add some information about ```dotnetRunMessages```, I think it could use it.  Thanks again.

PR #35273, opened 2023-09-11
Use MVC extensions from Razor transport

jjonescz:
Prerequisite for https://github.com/dotnet/razor/issues/8400.
Razor counterpart: https://github.com/dotnet/razor/pull/9266

MVC Razor compiler parts that previously came from NuGet now come from Razor transport package, i.e., directly from SDK when deployed.

### DIFF
--- a/src/RazorSdk/Tool/DiscoverCommand.cs
+++ b/src/RazorSdk/Tool/DiscoverCommand.cs
@@ -89,7 +89,7 @@ namespace Microsoft.NET.Sdk.Razor.Tool
                 }
             }
 
-            PatchExtensions(ExtensionNames, ExtensionFilePaths);
+            PatchExtensions(ExtensionNames, ExtensionFilePaths, Error);
 
             return true;
         }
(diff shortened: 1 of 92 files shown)


src/RazorSdk/Tool/DiscoverCommand.cs
```diff
@@ -89,14 +89,48 @@ protected override bool ValidateArguments()
...
+            for (int i = 0; i < extensionNames.Values.Count; i++)
+            {
+                var replacementFileName = extensionNames.Values[i] switch
+                {
+                    "MVC-1.0" or "MVC-1.1" => "Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.dll",
+                    "MVC-2.0" or "MVC-2.1" => "Microsoft.AspNetCore.Mvc.Razor.Extensions.Version2_X.dll",
+                    _ => null,
+                };
+
+                if (replacementFileName != null && HasExpectedFileName(extensionFilePaths.Values[i]))
```
jaredpar, line 116: Should we write a warning if the `replacmentFileName != null` but it does not have the expected name? Can't think of a scenario where that would make sense.

jaredpar (approved):

PR #35281, opened 2023-09-11
Msbuild.Test.Utilites build fix for VS

MiYanni:
Related: https://github.com/dotnet/sdk/pull/34659

## Summary
This is a problem that was in `main` but I fixed it in the PR above when I ran `dotnet format`. The fix for this problem doesn't exist in `release/8.0.1xx`, though. Adding just this fix so building in VS on this branch works properly. Without this fix, this project doesn't build in VS.

## Details
This project uses a shared `Program.cs` file that depends on `Microsoft.NET.TestFramework` to compile. Adding the project reference and utilizing the global using fixes the build problem in VS. This is what VS shows for the situation without these changes:
image


JL03-Yue (approved):

JL03-Yue:
Looks straight forward to me~

dsplaisted (approved):

Issue #34026, opened 2023-07-14
`-a` sets app to framework-dependent when `SelfContained` is `true`


richlander:
I'm working on some samples and following the excellent guidance at https://devblogs.microsoft.com/dotnet/improving-multiplatform-container-support/. The guidance doesn't work as intended when I target SCD.

My project file.

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <PublishSingleFile>true</PublishSingleFile>
    <PublishTrimmed>true</PublishTrimmed>
    <PublishReadyToRun>true</PublishReadyToRun>
    <PublishSelfContained>true</PublishSelfContained>
  </PropertyGroup>

</Project>
```

In Dockerfile:

```Dockerfile
# copy csproj and restore as distinct layers
COPY *.csproj .
RUN dotnet restore -a $TARGETARCH

# copy and publish app and libraries
COPY . .
RUN dotnet publish -a $TARGETARCH -o /app
```

This fails.

```bash
5.453 /root/.nuget/packages/microsoft.net.illink.tasks/8.0.0-preview.6.23329.7/build/Microsoft.NET.ILLink.targets(193,5): error NETSDK1102: Optimizing assemblies for size is not supported for the selected publish configuration. Please ensure that you are publishing a self-contained app. [/source/releases.csproj]
```

Works if I do:

```Dockerfile
RUN dotnet publish -a $TARGETARCH -o /app --self-contained
```

We should fix `-a` to be less strong. It should not override `SelfContained` or `PublishSelfContained` if set.

nagilson:
Interesting, yeah it looks like the RID Short-Hand Resolution is called by the Architecture Flag, and if you haven't specified SelfContained then it will inject this: 

`properties = properties.Append("-property:SelfContained=false").ToArray();
`, and since it's injected as a global property it would override your selfcontained or publishselfcontained value.

Not great, good catch! We should fix this.

richlander:
`-property:SelfContained=false")` ... That shouldn't be needed anymore, right?

nagilson:
Ya, that really shouldn't be there anymore. It is definitely going to break people when we fix it (yay) but it should be fixed.

nagilson:
[...]

PR #35286, opened 2023-09-11
Remove SelfContained set to false for -a

MiYanni:
Fixes: https://github.com/dotnet/sdk/issues/34026

> **Note**
> PR https://github.com/dotnet/sdk/pull/35281 is contained within this. Once merged, the changes to `Msbuild.Tests.Utilities.csproj` will not be present.

## Summary
When using `-a`, we were forcing `SelfContained` to be set to `false`. This is considered a breaking change, but the current functionality isn't desired. So, we change that here to no longer update the `SelfContained` property.

## Details
This change is thematically the same as https://github.com/dotnet/sdk/pull/30038. [`-a`](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-publish#options) sets the RID and we changed RID handling to no longer update the `SelfContained` property to `true`. Here in this situation, it was doing the same but changing it to `false` instead. This was only specific to the `-a` option unlike the other set of changes.

### DIFF
--- a/src/Cli/dotnet/CommonOptions.cs
+++ b/src/Cli/dotnet/CommonOptions.cs
@@ -201,8 +201,7 @@ namespace Microsoft.DotNet.Cli
                 return Array.Empty<string>();
             }
 
-            var selfContainedSpecified = (parseResult.GetResult(SelfContainedOption) ?? parseResult.GetResult(NoSelfContainedOption)) is not null;
-            return ResolveRidShorthandOptions(null, arg, selfContainedSpecified);
+            return ResolveRidShorthandOptions(null, arg);
         }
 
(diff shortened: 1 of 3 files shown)


baronfel (approved):
Thanks!

baronfel:
@MiYanni the CLA check is stuck - good to go whenever you can get that check green.

nagilson (commented):
LGTM, but make sure to update the breaking change documents as well. Thank you for helping make this change! :)

nagilson:
Ah jk it's already here https://github.com/dotnet/docs/issues/37058

MiYanni:
> Ah jk it's already here [dotnet/docs#37058](https://github.com/dotnet/docs/issues/37058)

Thanks @nagilson and @baronfel for the breaking change issue!

PR #35339, opened 2023-09-13
Update NuGet certificate thumbprints

joeloff:
## Description

Microsoft authored packages on NuGet are being signed using an [updated certificate](https://devblogs.microsoft.com/nuget/microsoft-author-signing-certificate-update-2023/). The CLI verifies the certificate thumbprint when acquiring packages related to workloads. This apply to both file-based and admin installs on Windows. The new certificate  thumbprint is unknown and the CLI needs to be updated to include thumbprints for both the leaf and intermediate nodes - the latter is a fallback mechanism in the CLI.

## Testing

Unit tests have been updated to verify a package signed with the new certificate.

## Workaround

Users can turn of sign verification when installing workloads through the CLI by passing the `--skip-sign-check` option, but this is not ideal and won't work if the policy key was specified on Windows to block bypassing signature validation for workload packages.


MiYanni (approved):
:shipit:

dtivel (approved):

PR #35343, opened 2023-09-13
Add `.mono` to `.gitingore`

PascalSenn:
I recently started using https://github.com/dotnet/vscode-csharp and it adds a folder `.mono` to the repository. 

I guess this should also be part of the template


WeihanLi:
The test case expected gitignore file should be updated also

https://github.com/dotnet/sdk/blob/main/src/Tests/dotnet-new.Tests/Approvals/AllCommonItemsCreate.-o%23dotnet-gitignore-file%23-n%23item.verified/dotnet-gitignore-file/.gitignore#L19

PR #35366, opened 2023-09-13
Add escape hatch for PublishAot without RuntimeIdentifier. Fixes #33414.

### LINKED ISSUE #33414: Overeager check for RuntimeIdentifier when PublishAot=true
There's a check to verify that RuntimeIdentifier is set when PublishAot=true:

https://github.com/dotnet/sdk/blob/f90e558092ac61038ffa1017fe3a5aca1af5ae7e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets#L224-L226

Unfortunately this check is not valid when building universal apps for macOS and Mac Catalyst (the `netX-macos` and `netX-maccatalyst` target frameworks).

When building a universal app, the project file will set RuntimeIdentifiers (plural):

```xml
<TargetFramework>net8.0-macos</TargetFramework>
<RuntimeIdentifiers>osx-x64;osx-arm</RuntimeIdentifiers>
```

and then during the build, the macOS SDK will run two inner builds, with RuntimeIdentifiers unset, and RuntimeIdentifier set to each of the rids.

The problem is that the outer build, where RuntimeIdentifiers is set, but RuntimeIdentifier isn't, triggers the above-mentioned verification, and that fails the build.

I attempted to unset PublishAot for the outer build, but that doesn't work for two reasons:

1. It can't be unset if customers pass /p:PublishAot=true on the command line.
2. The package restore won't restore the NativeAOT packages nor include them in the build, and thus effectively NativeAOT is turned off for the inner builds as well, even if we set PublishAot=true for them.

This workaround worked around the second point:

```xml
<Target Name="EnablePublishAot" BeforeTargets="ProcessFrameworkReferences">
    <PropertyGroup>
        <_OriginalPublishAot>$(PublishAot)</_OriginalPublishAot>
        <PublishAot>true</PublishAot>
    </PropertyGroup>
</Target>

<Target Name="RestorePublishAot" AfterTargets="ProcessFrameworkReferences">
    <PropertyGroup>
        <PublishAot>$(_OriginalPublishAot)</PublishAot>
    </PropertyGroup>
</Target>
```

but it's on the uglier side of ugly, and not a complete solution anyways (point 1 is still a problem).

I see two potential fixes:

a. Add a property to skip this check, and set this property in the macOS SDK:

```xml
    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true' and '$(AllowPublishAotWithoutRuntimeIdentifier)' != 'true'"
                FormatArguments="PublishAot"/>
```

b. Check for RuntimeIdentifiers (plural) as well:

```xml
    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(RuntimeIdentifiers)' != 'true' and '$(_IsPublishing)' == 'true'"
                FormatArguments="PublishAot"/>
```

rolfbjarne:
This adds an escape hatch for a sanity check when trying to use PublishAot without a RuntimeIdentifier, because the check is not valid when building universal apps for macOS and Mac Catalyst (the netX-macos and netX-maccatalyst target frameworks).

When building such an app, the project file will set RuntimeIdentifiers (plural):

    <TargetFramework>net8.0-macos</TargetFramework>
    <RuntimeIdentifiers>osx-x64;osx-arm</RuntimeIdentifiers>

and then during the build, the macOS SDK will run two inner builds, with RuntimeIdentifiers unset, and RuntimeIdentifier set to each of the rids (and at the end merge the result into a single app).

The problem is that the outer build, where RuntimeIdentifiers is set, but RuntimeIdentifier isn't, triggers the sanity check, and that fails the build.

I attempted to unset PublishAot for the outer build, but that doesn't work for two reasons:

* It can't be unset if customers pass /p:PublishAot=true on the command line.
* The package restore won't restore the NativeAOT packages nor include them in the build, and thus effectively NativeAOT is turned off for the inner builds as well, even if we set PublishAot=true for them.

So instead add an escape hatch for the sanity check.

Fixes https://github.com/dotnet/sdk/issues/33414.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets
@@ -221,7 +221,7 @@ Copyright (c) .NET Foundation. All rights reserved.
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishSingleFile"/>
 
-    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true'"
+    <NETSdkError Condition="'$(PublishAot)' == 'true' and '$(RuntimeIdentifier)' == '' and '$(_IsPublishing)' == 'true' and '$(AllowPublishAotWithoutRuntimeIdentifier)' != 'true'"
                 ResourceName="ImplicitRuntimeIdentifierResolutionForPublishPropertyFailed"
                 FormatArguments="PublishAot"/>
 
--- a/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/RuntimeIdentifiersTests.cs
@@ -380,5 +380,48 @@ namespace Microsoft.NET.Publish.Tests
                 .Should()
                 .Pass();
         }
+
+        [Fact]
+        public void PublishFailsWithPublishAotAndNoRuntimeIdentifier()
+        {
+            var targetFramework = ToolsetInfo.CurrentTargetFramework;
+            var testProject = new TestProject()
+            {
+                IsExe = true,
+                TargetFrameworks = targetFramework
+            };
+
+            testProject.AdditionalProperties["PublishAot"] = "true";
+            var testAsset = _testAssetsManager.CreateTestProject(testProject);
+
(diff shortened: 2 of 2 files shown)


dalexsoto (approved):

nagilson (approved):
LGTM, thank you for making a PR with a fix. I'm sorry I did not get to the issue in a timely manner. This is a smart way to go about it.

nagilson:
Keep in mind this will only go into .NET 9, you would need to target the 8.0-rc branches (new one is being created tonight/tomorrow morning) to get it in sooner.

PR #35409, opened 2023-09-14
Find dotnet.exe from VS's MSBuild

Forgind:
Fixes [AB#1872193](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1872193)

/cc: @jjonescz

@dsplaisted, can you review this please? I'm not 100% confident of this change or if there's a better way to do this.

Sorry for pinging on two PRs; I mistargeted the first one.

### DIFF
--- a/src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
+++ b/src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
@@ -274,6 +274,40 @@ namespace Microsoft.NET.Sdk.Razor.Tool
             }
         }
 
+        private static string FindDotNetExecutable()
+        {
+            var expectedPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
+            if (!string.IsNullOrEmpty(expectedPath))
+            {
+                return expectedPath;
+            }
+
+#if NET
+            expectedPath = System.Environment.ProcessPath;
+#else
+            expectedPath = Process.GetCurrentProcess().MainModule.FileName;
+#endif
+
+            if ("dotnet".Equals(Path.GetFileNameWithoutExtension(expectedPath), StringComparison.Ordinal))
+            {
+                return expectedPath;
+            }
+
+            // We were probably running from Visual Studio or Build Tools and found MSBuild instead of dotnet. Use the PATH...
+            var paths = Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator);
+            var exeName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dotnet.exe" : "dotnet";
+            foreach (string path in paths)
+            {
+                var dotnetPath = Path.Combine(path, exeName);
+                if (File.Exists(dotnetPath))
+                {
+                    return dotnetPath;
+                }
+            }
+
+            return exeName;
+        }
+
         // Internal for testing.
         internal static bool TryCreateServerCore(string clientDir, string pipeName, out int? processId, bool debug = false)
         {
@@ -282,15 +316,7 @@ namespace Microsoft.NET.Sdk.Razor.Tool
             // The server should be in the same directory as the client
             var expectedCompilerPath = Path.Combine(clientDir, ServerName);
 
-            var expectedPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
-            if (string.IsNullOrEmpty(expectedPath))
-            {
-#if NET
-                expectedPath = System.Environment.ProcessPath;
-#else
-                expectedPath = Process.GetCurrentProcess().MainModule.FileName;
-#endif
-            }
+            var expectedPath = FindDotNetExecutable();
 
             var argumentList = new string[]
             {


src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
```diff
@@ -274,6 +274,37 @@ public static string GetTempPath(string workingDir)
...
+                {
+                    var dotnetPath = Path.Combine(path, "dotnet.exe");
+                    if (File.Exists(dotnetPath))
+                    {
+                        return dotnetPath;
+                    }
+                }
+            }
+
+            return File.Exists(expectedPath) && Path.GetFileNameWithoutExtension(expectedPath).Equals("dotnet") ? expectedPath : "dotnet";
```
dsplaisted, line 305: I think there's a redundant clause here given the previous if statement.

```suggestion
            return File.Exists(expectedPath) ? expectedPath : "dotnet";
```

dsplaisted (approved):

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
Forgind: That clause is for if you are running from VS and find MSBuild rather than dotnet, and dotnet is not on your PATH for some reason. If you have a custom assembly resolve hook, I could theoretically imagine this being important...though I'll admit that I'm thinking < 1% of cases.

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
```diff
@@ -274,6 +274,38 @@ public static string GetTempPath(string workingDir)
...
+                return expectedPath;
+            }
+
+#if NET
+            expectedPath = System.Environment.ProcessPath;
+#else
+            expectedPath = Process.GetCurrentProcess().MainModule.FileName;
+#endif
+
+            if (!Path.GetFileNameWithoutExtension(expectedPath).Equals("dotnet"))
```
jjonescz, line 291: Nit: technically safer if `Path.GetFileNameWithoutExtension` returned `null`.

```suggestion
            if (Path.GetFileNameWithoutExtension(expectedPath) != "dotnet")
```

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
jjonescz: > if you are running from VS and find MSBuild rather than dotnet, and dotnet is not on your PATH

Why not simply do this?

```cs
if (Path.GetFileNameWithoutExtension(expectedPath) != "dotnet" || !File.Exists(expectedPath))
{
    // Find dotnet on PATH
    foreach { ... }

    // Fallback
    return "dotnet";
}

return expectedPath;
```

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
```diff
@@ -274,6 +274,38 @@ public static string GetTempPath(string workingDir)
             }
         }
 
+        private static string FindDotNetExecutable()
```
jjonescz, line 277: @jaredpar for thoughts (you've changed similar code in Roslyn IIRC)

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
Forgind: I don't think that's a risk here since the currently executing process should have a path, but I can make this a yoda condition anyway.

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
Forgind: On that note, the File.Exists check is unnecessary, too, since the current process's path should exist. I refactored this a bit, and it's cleaner now in my opinion. Thanks for the suggestions 🙂

jjonescz (approved):

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
```diff
@@ -274,6 +274,40 @@ public static string GetTempPath(string workingDir)
...
+            foreach (string path in paths)
+            {
+                var dotnetPath = Path.Combine(path, exeName);
+                if (File.Exists(dotnetPath))
+                {
+                    return dotnetPath;
+                }
+            }
+
+            return "dotnet";
```
jaredpar, line 308:
```suggestion
            return exeName;
```

jaredpar (changes requested):

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
Forgind: Can you explain why you think it's necessary to use exeName rather than "dotnet"? I think either should work. From a functional perspective, the difference lies in whether the extension is included. I chose "dotnet" because that's [how it was before my original change](https://github.com/dotnet/sdk/blame/4ef9128d136cfc9e55c0374d01b117a9790c78a6/src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs#L291), and reverting to previous behavior seems safest to me.

jaredpar: It's less precise on Windows. It would for example allow invocation of `dotnet.cmd` which seems wrong. My mind starts asking questions like is that spoofable?

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
```diff
@@ -274,6 +274,40 @@ public static string GetTempPath(string workingDir)
...
+                return expectedPath;
+            }
+
+#if NET
+            expectedPath = System.Environment.ProcessPath;
+#else
+            expectedPath = Process.GetCurrentProcess().MainModule.FileName;
+#endif
+
+            if ("dotnet".Equals(Path.GetFileNameWithoutExtension(expectedPath)))
```
danmoseley, line 291: Generally we recommend to always explicitly pass comparer to string comparisons so if nothing else it's clear to the reader.

src/RazorSdk/Tool/ServerProtocol/ServerConnection.cs
Forgind: If we've reached this code path, we've already given up on being truly safe. That said, some users may be very cautious about which executables they download without noticing dotnet.cmd, so I could see this making the user just slightly safer. I can make the change, thanks.

jaredpar (approved):

PR #35411, opened 2023-09-14
Write the full path of the dotnet tools directory in /etc/paths.d/dotnet-cli-tools

### LINKED ISSUE #23165: macOS dotnet CLI path for global tools is bad - Installer issue
### Describe the bug
On macOS when a .Net tool is installed globally, it will not be found in the PATH.

The dotnet installer does attempt to make the directory for globally installed tools part of the PATH, but the mechanism the installer is relying on doesn't work as the installer apparently expects.

### To Reproduce
1. Install dotnet on macOS using the Microsoft Installer from https://dotnet.microsoft.com/en-us/download.
2. Install a tool globally, e.g. `dotnet tool install -g Microsoft.dotnet-httprepl`
3. Assuming zsh which is the default shell on macOS (Catalina and newer), run the command `which httprepl`
4. The output of the which command will be "httprepl not found" indicating that the httprepl tool can’t be found in the PATH. 

#### Why does this happen?
macOS has a mechanism for adding paths to the PATH environment variable. The files in /etc/paths.d are used by the path_helper tool to build up the PATH environment variable.

The installer for dotnet is adding two files for paths:
- /etc/paths.d/dotnet
- /etc/paths.d/dotnet-cli-tools

The _dotnet_ file contains
```
/usr/local/share/dotnet
```

The _dotnet-cli-tools_ file contains
```
~/.dotnet/tools
```

However, paths in these files should be literal. The tilde (~) is not expanded to be the current user home directory. Environment variable references are also not expanded.

##### bash
The default standard shell on macOS Catalina (released in 2019) and newer is zsh. Prior to Catalina the default standard shell was bash. The bash shell appears to expand the tilde at a later point for an interactive shell so this issue may not manifest for bash users. However, the files created in /etc/paths.d by the dotnet installer need to support all shells.

#### Can't Specify the Current User Home Directory
The tilde can't be expected to be expanded to be the current user home directory. ${HOME} also can’t be used. There is not a way to specify the current user home directory in a file in /etc/paths.d.

I don’t know if not supporting the current user home directory is by design or is an omission. 

Given the current functionality, the dotnet installer should not create the /etc/paths.d/dotnet-cli-tools file at all.

#### Workaround for Individual Users
In the .profile file, the user can either add the following line to add the needed path:
```
# dotnet-cli-tools
export PATH=$PATH:~/.dotnet/tools
```
or, assuming the bad path exists in the PATH, add the following ‘fix-up’ to replace any tildes in the PATH (equivalent to bash):
```
export PATH=${PATH//\~/$HOME}
```

The .zprofile may be set to source the .profile file with the following line:
```
[[ -e ~/.profile ]] && emulate sh -c 'source ~/.profile'
```

0xced:
Fixes #23165 and #9415

Also remove the instructions to manually tweak the PATH when using zsh since they are no longer necessary.


jrdodds:
The `/etc/paths.d/` directory and the `dotnet-cli-tools` file are global for all users. Writing the full path of a specific user in this file should not be done.

b4ux1t3:
> The `/etc/paths.d/` directory and the `dotnet-cli-tools` file are global for all users. Writing the full path of a specific user in this file should not be done.

I'm trying to understand the issue here.

On the one hand, it could expose some of a user's home folder structure.

However, it's generally (and by that I mean I would be surprised if you found even one counter-example) going to be the very well-documented dotnet tools path, and as such isn't likely to cause any raised eyebrows.

For most users, even if there is another user's home path in that global path, there won't be any issues, as they likely won't have permissions to that other user's path and, as such, path resolution will just skip that path as "empty".

There are no reasonable security concerns here. There are no performance concerns that aren't inherent with any path lookup. There are, however, very real concerns of users not being able to use global dotnet tools at all.

jrdodds:
@0xced Is this PR changing the installer?

cicorias:
> The `/etc/paths.d/` directory and the `dotnet-cli-tools` file are global for all users. Writing the full path of a specific user in this file should not be done.

This is such a rare situation that a dotnet sdk developer will have a device that other users are also using. Even if it did, the comment that @b4ux1t3 indicates -- it won't affect other users -- is true as well -- lack of perms.

cicorias:
> The `/etc/paths.d/` directory and the `dotnet-cli-tools` file are global for all users. Writing the full path of a specific user in this file should not be done.

this is such a rare situation that a dotnet sdk developer will have a device that other users are also using. Even if it did, the comment that @b4ux1t3 indicates -- it won't affect other users -- is true as well -- lack of perms.

jrdodds:
The root issue is that "global" tools are not installed globally. They are installed only for a specific user.

If the global tools were installed in a folder under `/usr/local/share/dotnet`, that tools path could be shared in `/etc/paths.d/`.

Or 'global' actually means user scoped and the tools path in the user directory needs to be added to the `PATH` in a user scoped file (e.g. `.zprofile`) and not via `/etc/paths.d/`.

Also note that if UserA installs a global tool and then UserB installs the same tool globally, one of these users will be unable to use the tool. It appears it will be UserA because it appears that the `dotnet-cli-tools` file is overwritten. But even if both user tool directories were in the `PATH`, the user whose directory is second in the `PATH` will be unable to use the tool because the other user's copy will be found.

b4ux1t3:
@jrdodds So, you're saying if they allowed for a `user` scope for dotnet tools, adding further complexity, that would be better than "fixing it for a vast majority of use cases" without adding any additional complexity?

jrdodds:
> @jrdodds So, you're saying if they allowed for a `user` scope for dotnet tools, adding further complexity, that would be better than fixing it for a vast majority of use cases without adding any additional complexity?

No. You have what I'm saying wrong. I don't think another 'scope' should be added.

jrdodds:
For `dotnet tool`, the "global" scope is the scope of a specific user. It does not mean global to all users. It means global to all projects for a specific user.

On Windows the install folder for a global tool is `%USERPROFILE%\.dotnet\tools`. This is added to the `PATH` in the user environment variables.

On macOS the install folder for a global tool is `$HOME/.dotnet/tools`. It follows that the `PATH` should be updated in a manner roughly equivalent to the Windows user environment variables. That would be `$HOME/.zprofile` or another file in the user home directory.

But instead the "All Users" `/etc/paths.d/` mechanism is used. This is a defect.

| OS | macOS | Windows |
| --- | ------- | --------- |
| All Users | `/etc/paths.d/`  | System Env Var `PATH` |
| Specific User | `$HOME/.zprofile`  | User Env Var `PATH`  |

The message about a one time user action to add the tools directory to the `~/.zprofile` is correctly addressing the defect.

An improvement over the current message would be to extend the `dotnet` tool to have an equivalent to Homebrew's [`brew shellenv`](https://docs.brew.sh/Manpage#shellenv-bashcshfishpwshshtcshzsh).

(Note that [shell command completion for `dotnet`](https://learn.microsoft.com/en-us/dotnet/core/tools/enable-tab-autocomplete) is already implemented with a `dotnet complete` command.)

b4ux1t3:
There is a marked difference between the concept of environment variables on Windows and *nix (Including, for this case, Mac).

I do not allow third party tools to edit my shell configuration files, such as `.zprofile` or `.zshrc`. I use my dotfiles on multiple systems with multiple operating systems (including sharing it between Linux and Mac).

This is not an uncommon use of those files. They are not _machine-specific_ configurations, they are _user-specific_ configurations, which must maintain usability wherever that user uses them, _even on other machines_.

Conversely, the `/etc/paths.d` is a machine-specific configuration, just like the Windows environment variable setup.

How Windows separates the User Environment from the System Environment is a Windows-specific detail, and should not be applied to other platforms. The User Environment is a configuration for that User _on that machine_. Yes, it can be synchronized with a federated windows login. That isn't how it is generally used.

When you're working on different platforms, you have to take the conventions of the platforms into account. For many *nix developers, both on Linux and Mac, their shell configuration is _their_ shell configuration. It isn't a dumping ground for whatever nonsense other applications want to put there.

That's what `paths.d` is for. That, in this case, it happens to include paths that aren't accessible to all users isn't a defect, it's a side effect, and it doesn't hurt anything.

b4ux1t3:
@KalleOlaviNiemitalo You're absolutely right. Those are concerns. I just don't believe that they're reasonable concerns (I don't know if you're arguing they are, just clarifying my position).

jrdodds:
I very strongly agree with:

> When you're working on different platforms, you have to take the conventions of the platforms into account. For many *nix developers, both on Linux and Mac, their shell configuration is their shell configuration. It isn't a dumping ground for whatever nonsense other applications want to put there.

I disagree with:

>  That's what paths.d is for. That, in this case, it happens to include paths that aren't accessible to all users isn't a defect, it's a side effect, and it doesn't hurt anything.

When you dig into it, paths.d is not for user specific configuration. (That paths.d 'works' with the bash shell is a side effect of how bash handles the tilde.)

There are tools that respect both that paths.d is not for user specific configuration and that dotfiles should not be edited by a third party.

There are 'replacement' versions of path_helper written by others that extend the feature to support per user paths by adding support for a user level paths.d folder, e.g. `$HOME/Library/Paths/paths.d`.

We agree that the conventions and rules of a platform have to be taken into account. It's not okay to break a rule because the rule is inconvenient. We disagree about the rules for paths.d.

cicorias:
well, first I'd opt for to stop mucking with `/etc/path.d` altogether and just have the installer emit a clear message -- or the dotnet tool itself provide a warning of some time that "someting is misssing"..


it's tiresome to keep telling macOS folks to change that file on every SDK update.

jrdodds:
> it's tiresome to keep telling macOS folks to change that file on every SDK update.

@cicorias Are you telling people to change the `/etc/paths.d/dotnet-cli-tools` file?

For zsh, the following could be added to `.zprofile`:

```zsh
# replace ~ with $HOME
export PATH=${PATH//\~/$HOME}
```

cicorias:
> > it's tiresome to keep telling macOS folks to change that file on every SDK update.
> 
> @cicorias Are you telling people to change the `/etc/paths.d/dotnet-cli-tools` file?
> 
> For zsh, the following could be added to `.zprofile`:
> 
> ```shell
> # replace ~ with $HOME
> export PATH=${PATH//\~/$HOME}
> ```

I was -- however, I'll try this -- tx.

cicorias:
> > > it's tiresome to keep telling macOS folks to change that file on every SDK update.
> > 
> > 
> > @cicorias Are you telling people to change the `/etc/paths.d/dotnet-cli-tools` file?
> > For zsh, the following could be added to `.zprofile`:
> > ```shell
> > # replace ~ with $HOME
> > export PATH=${PATH//\~/$HOME}
> > ```
> 
> I was -- however, I'll try this -- tx.

@jrdodds  thanks so much. This works well.

PR #35418, opened 2023-09-14
Metrics switch support in the SDK

tarekgh:
Within the Runtime, we've introduced a configuration switch called `System.Diagnostics.Metrics.Meter.IsSupported` to manage the activation and deactivation of metrics functionality. This adjustment has been made to facilitate compatibility with the msbuild property named `MetricsSupport`.

https://github.com/dotnet/runtime/issues/89880
https://github.com/dotnet/runtime/pull/92019

I have conducted comprehensive testing for the end-to-end scenario by configuring the `MetricsSupport` in a standalone application using the latest SDK and runtime. I've ensured that when `MetricsSupport` is absent or set to `true` in the csproj, the metrics functionality remains enabled. Conversely, when it's set to `false`, the metrics functionality is disabled. You can find further details and context regarding the need for this switch in .NET 8.0 in the pull request located here: https://github.com/dotnet/runtime/pull/92019, where an explanatory template outlines the rationale behind its implementation.


tarekgh:
@eerhardt @dsplaisted could you please help reviewing this change? 

@artl93 @marcpopMSFT @ericstj for the merger approval to this rc2 branch. 

CC @noahfalk @grendello

marcpopMSFT:
We'll be snapping for RC2 so merge before then for inclusion.

src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAProjectWithAllFeatures.cs
```diff
@@ -61,6 +61,7 @@ public void It_publishes_the_project_correctly(string targetFramework, string []
             ""System.ComponentModel.TypeConverter.EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization"": false,
             ""System.Diagnostics.Debugger.IsSupported"": true,
             ""System.Diagnostics.Tracing.EventSource.IsSupported"": false,
+            ""System.Diagnostics.Metrics.Meter.IsSupported"": false,
```
eerhardt, line 64:
```suggestion
            ""System.Diagnostics.Metrics.Meter.IsSupported"": false,
            ""System.Diagnostics.Tracing.EventSource.IsSupported"": false,
```

(nit) the comment above says "keep this list sorted".

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -504,6 +504,10 @@ Copyright (c) .NET Foundation. All rights reserved.
                                     Value="$(DebuggerSupport)"
                                     Trim="true" />
 
+    <RuntimeHostConfigurationOption Include="System.Diagnostics.Metrics.Meter.IsSupported"
+                                    Condition="'$(MetricsSupport)' != ''"
+                                    Value="$(MetricsSupport)" />
```
eerhardt, line 509:
```suggestion
                                    Value="$(MetricsSupport)"
                                    Trim="true" />
```

You'll want to pass `Trim=true` here, so it gets flown into the trimmer and AOT compiler.

ericstj:
Approve this scenario.  Thank you for testing end-to-end.  Please consider @eerhardt's feedback and do any follow up e2e testing needed to confirm that change is needed (and why earlier tests didn't notice it).

src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantToBuildALibrary.cs
```diff
@@ -896,6 +896,50 @@ public void It_can_implicitly_define_predefined_Cultures_only(string targetFrame
...
+                TargetFrameworks = ToolsetInfo.CurrentTargetFramework,
+                IsExe = true,
+            };
+
+            if (value is not null)
+            {
+                testProj.AdditionalProperties["MetricsSupport"] = value;
+            }
+
+            var testAsset = _testAssetsManager.CreateTestProject(testProj, identifier: $"{ToolsetInfo.CurrentTargetFramework}{value}");
```
dsplaisted, line 917: The identifier only needs to differentiate between different Theory parameter combinations, so if the target framework isn't changing between theory instances, it doesn't need to be included.

```suggestion
            var testAsset = _testAssetsManager.CreateTestProject(testProj, identifier: value);
```

eerhardt (approved):
LGTM. Thanks!

dsplaisted (approved):
Looks good.

What happens if you do something like [`<MetricsSupport>FileNotFound</MetricsSupport>`](https://thedailywtf.com/articles/What_Is_Truth_0x3f_)? 
 It looks like for other similar properties we don't do any validation that the value is valid, so it's probably fine as is.

tarekgh:
> What happens if you do something like [<MetricsSupport>FileNotFound</MetricsSupport>](https://thedailywtf.com/articles/What_Is_Truth_0x3f_)?
> It looks like for other similar properties we don't do any validation that the value is valid, so it's probably fine as is.

In the code we do the following to read the switch
```C#
            AppContext.TryGetSwitch("System.Diagnostics.Metrics.Meter.IsSupported", out bool isSupported) ? isSupported : true;

```

So, if the value is defined with something that is not `true` or `false`, we will still have the switch value as `true` which is the correct behavior we need.  I'll try to add a test for this case too.

Issue #35431, opened 2023-09-15
SDK shouldn't upload manifest separately when uploading tags.


tmds:
I'm investigating an issue where the SDK tooling fails to upload an image to the internal OpenShift container image registry.

The internal image registry is integrated with OpenShift and for images pushed to it, matching Kubernetes resources are created.

The push fails when the image is trying to upload the manifest under its digest here:

https://github.com/dotnet/sdk/blob/f8a9e2866134d9d04805da55fcb3984c8b94e008/src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs#L428

That digest has no known tag, and the OpenShift registry doesn't like it.

CLI tools (like `podman`) work well with the registry.
They are behaving slightly different from the SDK.
This is the log from `podman` when pushing an image to quay.io:
```
Writing manifest to image destination
time="2023-09-15T10:07:48+02:00" level=debug msg="PUT https://quay.io/v2/tmds/my-image/manifests/my-tag"
time="2023-09-15T10:07:49+02:00" level=debug msg="Called push.PersistentPostRunE(podman push --log-level trace quay.io/tmds/my-image:my-tag)"
time="2023-09-15T10:07:49+02:00" level=debug msg="Shutting down engines"
```

As seen in the log, `podman` doesn't upload the manifest under its digest, and it only uploads it as the tag.

Uploading the manifest digests is only needed when tags refer to a manifest list.
The manifest list for the tags can then refer to the image manifests using that digest.

To fix the issue with the OpenShift, we can make the SDK behave like the CLI tools and only upload the manifest separately when needed for a manifest list upload.

cc @baronfel

baronfel:
Closing as we did this in #35437 - should be part of RC2.

tmds:
I imagine these get auto closed when those changes are merged to the main branch? Or doesn't that auto close happen?

PR #35437, opened 2023-09-15
Containers: don't push manifest digest when pushing manifest tags.

tmds:
This behavior is consistent with CLI tools and fixes an issue with the OpenShift internal image repository.

Fixes https://github.com/dotnet/sdk/issues/35431.

@baronfel ptal.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
@@ -361,7 +361,10 @@ internal sealed class Registry
 
     }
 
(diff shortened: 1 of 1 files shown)


tmds:
@baronfel can you please check this change with your tests that run against multiple registries?

I expect there to be no issue because this matches the behavior of `podman`, which has been battle tested against many registry implementations.

It may be interesting to check what `docker` does exactly, if there there is a simple way to get the logs. It may be doing the exact same thing as `podman` and what we're changing to here.

baronfel:
Corresponding Actions run for the manifest upload tests: https://github.com/baronfel/sdk-container-demo/actions/runs/6199744366

So far seems to be good!

baronfel (approved):
This passes tests and unblocks a partner scenario. Thanks for the contribution, @tmds!

tmds:
Thank you, @baronfel!

PR #35450, opened 2023-09-15
Trigger enhanced page nav on hot reload

captainsafia:
Depends on https://github.com/dotnet/aspnetcore/pull/50437.

Addresses https://github.com/dotnet/aspnetcore/issues/49144.

| Scenario | Validated |
|---|---|
| Statically server rendered page | ✅  |
| Interactive server rendered page |  ✅ |
| Page with stream rendering enabled |  ✅  |
| Interactive client rendered page |  ❌ (https://github.com/dotnet/aspnetcore/issues/50765)  |
| Statically server rendered page with interactive components (all render modes) |  ✅  |

### DIFF
--- a/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
+++ b/src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
@@ -1,4 +1,5 @@
 setTimeout(async function () {
+  const hotReloadActiveKey = '_dotnet_watch_hot_reload_active';
   // Ensure we only try to connect once, even if the script is both injected and manually inserted
   const scriptInjectedSentinel = '_dotnet_watch_ws_injected';
   if (window.hasOwnProperty(scriptInjectedSentinel)) {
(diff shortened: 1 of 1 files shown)


src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
```diff
@@ -181,18 +182,33 @@ setTimeout(async function () {
...
     if (window.Blazor) {
-      // If this page has any Blazor, don't refresh the browser.
-      notifyHotReloadApplied();
+      window[hotReloadActiveKey] = true;
+      // hotReloadApplied triggers an enhanced navigation to
+      // refresh pages that have been statically rendered with
+      // Blazor SSR.
+      if (window.Blazor?._internal?.hotReloadApplied)
+      {
+        Blazor._internal.hotReloadApplied();
```
SteveSandersonMS, line 206: From reading `Boot.Web.ts` in dotnet/aspnetcore, `hotReloadApplied` is defined as:

```ts
  Blazor._internal.hotReloadApplied = () => {
    if (hasProgrammaticEnhancedNavigationHandler()) {
      performProgrammaticEnhancedNavigation(location.href, true);
    }
  };
```

It looks like that will do nothing if `options.ssr.disableDomPreservation` is set to true. Do we also need something else to make hot reload work in that case? For example we could add an `else` block that forces a full-page load.

I appreciate you can't do that in this PR since that's in a different repo. Just check if that's something we need to follow up on today!

SteveSandersonMS (approved):
Looks good! Added a comment about one thing that might need followup but I'm unsure since maybe it's being handled some other way.

src/BuiltInTools/BrowserRefresh/WebSocketScriptInjection.js
captainsafia: Opened https://github.com/dotnet/aspnetcore/issues/50790 to follow up.

Issue #29859, opened 2022-11-17
Upgraded VS2022 to latest supporting .Net 7 - Now dotnet publish fails. 


Relki:

pimbrouwers:
I can confirm/repro this using the 7.0.102 SDK. 

I've been working on a project for the past several months targeting `net7.0` publishing to `win-x64` without any issues whatsoever. After coming back from the holiday's I now get this error as well during restore. I was able to repro it with a blank console app, adding the `win-x64` runtime identifier to the project file.

FsConsole.fsproj

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Program.fs" />
  </ItemGroup>

</Project>
```

Program.fs

``` 
printfn "Hello world"
```

Console output

```
PS C:\Users\pim\Downloads\FsConsole> dotnet clean
MSBuild version 17.4.1+9a89d02ff for .NET
Build started 2023-01-15 11:31:20 AM.
     1>Project "C:\Users\pim\Downloads\FsConsole\FsConsole.fsproj" on node 1 (Clean target(s)).
     1>C:\Program Files\dotnet\sdk\7.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1047: 
       Assets file 'C:\Users\pim\Downloads\FsConsole\obj\project.assets.json' doesn't have a target for 'net7.0/win-x64'. Ensure that restore has 
       run and that you have included 'net7.0' in the TargetFrameworks for your project. You may also need to include 'win-x64' in your project's 
       RuntimeIdentifiers. [C:\Users\pim\Downloads\FsConsole\FsConsole.fsproj]
     1>Done Building Project "C:\Users\pim\Downloads\FsConsole\FsConsole.fsproj" (Clean target(s)) -- FAILED.

Build FAILED.

       "C:\Users\pim\Downloads\FsConsole\FsConsole.fsproj" (Clean target) (1) ->
       (ResolvePackageAssets target) -> 
         C:\Program Files\dotnet\sdk\7.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1047
[...]

PR #35491, opened 2023-09-18
Meaningful error on runtime identifier mismatch in project.assets.json

### LINKED ISSUE #30199: `dotnet clean` (`--all`?) can always fail before a build is run for apps with a `RuntimeIdentifier`
As reported by @pimbrouwers in: https://github.com/dotnet/sdk/issues/29859 
If you do something like `dotnet new console -lang "F#"` and add a `RuntimeIdentifier` to the project and try `dotnet clean`, it will fail every time until you run `dotnet build`, in which case it will work.  

This is because clean only cleans the folder for the exact RID, Config, TFM, etc, and there is no `runtime config assets.json` as that is created by build but never cleaned up, clean only cleans the output directory, and relies on this file to do so. We would maybe like to have an error instead of crash, though we could also design a `dotnet clean --all` which would clean all output... though we are concerned about deleting people's files.

BogdanYarotsky:
@nagilson
Resolves #30199

Since the error appeared in the `ResolvePackageAssets` task of `Microsoft.PackageDependencyResolution.targets` I have added RID validation before it. `CheckRuntimeIdentifier` task handles the situations when there is a mismatch between the runtime identifier of the project and project.assets.json targets. More descriptive build error is now shown when `dotnet clean` is invoked.

Here is a screenshot of the updated error message (and current workflow):
image

### DIFF
--- a/src/Tasks/Common/Resources/Strings.resx
+++ b/src/Tasks/Common/Resources/Strings.resx
@@ -948,5 +948,9 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1213: Targeting .NET 8.0 or higher in Visual Studio 2022 17.7 is not supported.</value>
     <comment>{StrBegin="NETSDK1213: "}</comment>
   </data>
-  <!-- The latest message added is Net8NotCompatibleWithDev177. Please update this value with each PR to catch parallel PRs both adding a new message -->
+  <data name="AssetsFileRuntimeIdentifierMismatch" xml:space="preserve">
+    <value>NETSDK1214: Runtime Identifier {0} is not present in project.assets.json. Please ensure the Runtime Identifier is specified in your project file (e.g., &lt;RuntimeIdentifier&gt;{0}&lt;/RuntimeIdentifier&gt;) and the project was restored before invoking the command.</value>
+    <comment>{StrBegin="NETSDK1214: "}</comment>
+  </data>
+  <!-- The latest message added is AssetsFileRuntimeIdentifierMismatch. Please update this value with each PR to catch parallel PRs both adding a new message -->
 </root>
--- a/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckRuntimeIdentifierTask.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckRuntimeIdentifierTask.cs
@@ -1,0 +1,252 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using FluentAssertions;
+using Microsoft.NET.TestFramework;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.NET.Build.Tasks.UnitTests;
+
+public class GivenACheckRuntimeIdentifierTask : SdkTest
+{
+    private const string AssetsFileName = "project.assets.json";
+    private const string TargetFramework = "net8.0";
+    private const string UbuntuX64RuntimeIdentifier = "ubuntu-x64";
+    private const string WindowsX64RuntimeIdentifier = "win-x64";
+    private const string ExpectedErrorCode = "NETSDK1214";
+
+    private readonly CheckRuntimeIdentifier _task;
+    private readonly string _assetsFilePath;
+
+    public GivenACheckRuntimeIdentifierTask(ITestOutputHelper log) : base(log)
+    {
+        var testDirectory = _testAssetsManager.CreateTestDirectory();
+        _assetsFilePath = Path.Combine(testDirectory.Path, AssetsFileName);
+        _task = new CheckRuntimeIdentifier
+        {
+            ProjectAssetsFile = _assetsFilePath,
+            TargetFramework = TargetFramework,
+            BuildEngine = new MockBuildEngine()
+        };
+    }
+
+    [Fact]
+    public void ItShouldFailWithABuildErrorIfAssetsFileIsMissingTargetRuntimeIdentifier()
+    {
+        WriteToAssetsFile(AssetsJsonWithoutAddedRuntimeIdentifiers);
+        _task.RuntimeIdentifier = WindowsX64RuntimeIdentifier;
+        _task.Execute();
+        AssertThatTaskThrewExpectedBuildError();
+    }
+
+    [Fact]
+    public void ItShouldExecuteWithoutErrorsIfRuntimeIdentifierIsPresentInAssetsFile()
+    {
+        WriteToAssetsFile(AssetsJsonWithWithAddedWinX64RuntimeIdentifier);
+        _task.RuntimeIdentifier = WindowsX64RuntimeIdentifier;
+        _task.Execute();
+        AssertThatTaskThrewNoErrors();
+    }
+
+    [Fact]
+    public void ItShouldFailWithABuildErrorIfTargetIdentifierDiffersFromCurrentOne()
+    {
+        WriteToAssetsFile(AssetsJsonWithWithAddedWinX64RuntimeIdentifier);
+        _task.RuntimeIdentifier = UbuntuX64RuntimeIdentifier;
+        _task.Execute();
+        AssertThatTaskThrewExpectedBuildError();
+    }
+
+    [Fact]
+    public void ItShouldNotFailIfRuntimeIdentifierIsMissing()
+    {
+        WriteToAssetsFile(AssetsJsonWithWithAddedWinX64RuntimeIdentifier);
+        _task.RuntimeIdentifier = null;
+        _task.Execute();
+        AssertThatTaskThrewNoErrors();
+    }
+
+    private void WriteToAssetsFile(string content) => File.WriteAllText(_assetsFilePath, content);
+
+    private void AssertThatTaskThrewExpectedBuildError()
+    {
+        var engine = (MockBuildEngine)_task.BuildEngine;
+        engine.Errors.Should().HaveCount(1);
+        var error = engine.Errors.Single();
+        error.Code.Should().Be(ExpectedErrorCode);
+    }
+
+    private void AssertThatTaskThrewNoErrors()
+    {
+        var engine = (MockBuildEngine)_task.BuildEngine;
+        engine.Errors.Should().HaveCount(0);
+    }
+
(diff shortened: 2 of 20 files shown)


nagilson:
Thank you so much for your contribution, I will take a look at this within the next few business days :) I saw the test failures and was going to ask you to what you just did first, but looks like you're on top of it. I'm excited to take a look and we all appreciate your work 🔥

BogdanYarotsky:
thanks @nagilson!
I hope that I took the steps in the right direction :)

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.PackageDependencyResolution.targets
```diff
@@ -227,6 +229,19 @@ Copyright (c) .NET Foundation. All rights reserved.
     
   </Target>
 
+  <!-- Throws an error when currently used runtime identifier is not present in project.assets.json. -->
+  <Target Name="CheckRuntimeIdentifier"
+          Condition="'$(DesignTimeBuild)' != 'true' And Exists('$(ProjectAssetsFile)')"
```
nagilson, line 234: MSBuild lets you query environment variables as if they were variables in MSBuild. Generally when we add a new warning, we add a flag enviroment variable to disable the warning/error in case it breaks customers unexpectedly. I would add one here such as `DOTNET_CLI_DISABLE_RUNTIME_ASSETS_RESTORE_CHECK`.

src/Tasks/Common/Resources/Strings.resx
```diff
@@ -948,5 +948,9 @@ You may need to build the project on another operating system or architecture, o
     <value>NETSDK1213: Targeting .NET 8.0 or higher in Visual Studio 2022 17.7 is not supported.</value>
     <comment>{StrBegin="NETSDK1213: "}</comment>
   </data>
-  <!-- The latest message added is Net8NotCompatibleWithDev177. Please update this value with each PR to catch parallel PRs both adding a new message -->
+  <data name="AssetsFileRuntimeIdentifierMismatch" xml:space="preserve">
+    <value>NETSDK1214: Runtime Identifier {0} is not present in project.assets.json. Please ensure the Runtime Identifier is specified in your project file (e.g., &lt;RuntimeIdentifier&gt;{0}&lt;/RuntimeIdentifier&gt;) and the project was restored before invoking the command.</value>
```
nagilson, line 952: This error is very clean, concise, and provides an actionable item. I love it, it's very well done 😄

nagilson (commented):
A few suggestions are made below. One more test also still needs investigation. Thank you for your hard work and following up. Please let me know if there's any questions :) 

Once this is responded to, I can take another dive into it.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckRuntimeIdentifierTask.cs
```diff
@@ -0,0 +1,252 @@
...
+            ""runtimeIdentifierGraphPath"": ""C:\\Program Files\\dotnet\\sdk\\8.0.100-preview.3.23178.7\\RuntimeIdentifierGraph.json""
+          }
+        }
+      }
+    }";
+
+
+    private const string AssetsJsonWithWithAddedWinX64RuntimeIdentifier = @"
+    {
+      ""version"": 3,
```
nagilson, line 164: We typically put assets like these in the src/test/TestAssets/ folder. I would load them there instead of using a hard-coded string, this way it is easier to reproduce having an actual file on disk ;)

src/Tasks/Common/Resources/Strings.resx
BogdanYarotsky: Thank you @nagilson for a thorough review! It's a pleasure developing with you 🙂
I will adress the comments this week.

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckRuntimeIdentifierTask.cs
BogdanYarotsky: Hi @nagilson,

I initially used the GivenAResolvePackageAssetsTask test file as a reference since it contained some hard-coded asset JSON within the test class itself.

I attempted to move these JSONs to src/Assets and use _testAssetsManager.CopyTestAsset(). However, I encountered an issue where the obj and bin folders weren't copied over for some reason; they appear to be omitted by _testAssetsManager.

From my understanding, the alternative approach would be to set up a clean project within src/Assets and modify it as needed in each individual test case (e.g., restore, add runtime identifier, run task).

One challenge I'm facing is to find a place for this integration test. Microsoft.NET.Build.Tasks.UnitTests doesn't seem like the most suitable project as it focuses mainly on tasks, making it too narrow for this use case. On the other hand, incorporating this directly into the dotnet CLI tests feels too broad, and I'm uncertain about how to validate error codes in that context.

Do you have any suggestions on where this test might best fit? I've retained the hard-coded JSON for now until I find a better spot for it.

Thank you for your time and guidance 😃

src/Tasks/Common/Resources/Strings.resx
nagilson: It's a pleasure developing with you too, thank you for your contributions 😄

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckRuntimeIdentifierTask.cs
nagilson: I see, thanks for your thorough explanation. You're right in that it'd probably be easiest to make one test project and modify it there. `\src\Tests\Microsoft.NET.Build.Tests\GlobalPropertyFlowTests.cs` has a great example of our library code we have to edit project files to have a `RuntimeIdentifier` or whatever else. 

For example, you could use this code:
```cs

            _testProject = new TestProject()
            {
                TargetFrameworks = ToolsetInfo.CurrentTargetFramework,
            };
            
            var testAsset = _testAssetsManager.CreateTestProject(_testProject);
```

Then run the `dotnetBuild` command, then set the property of the project afterward and try to run restore? I think that should work.


Such as 
```cs
          var properties = _testProject.GetPropertyValues(testAsset.TestRoot, targetFramework: ToolsetInfo.CurrentTargetFramework);

              properties["RuntimeIdentifier"] = "foo";
          
```

I think you have found the best fit place for this test :) Since that's where the task is, it makes sense that its tests would be held in a mirror location in the test folder structure. I wouldn't put the test project in the root folder called `test`, I would put it in the `TestProjects` folder in the `Assets` folder if you still need it, but likely you don't. Another thing that we'd avoid is including unnecessary information in the runtime assets json file and include only the things necessary for the test (if the data is needed to make it run, then void it out to empty string etc.)

nagilson (commented):
Overall the change looks great, thank you for sticking through! :smile: Just some nits about the tests to match our existing codebase a bit better.

@ocalles May you also see if you agree with this change, since you seem to know this area very well? Thanks. :) If not, that's cool too.

nagilson:
@dotnet/dotnet-cli Heya folks, I've reviewed this PR pretty thoroughly from the community contributor @BogdanYarotsky, but it's always good to have a 2nd look if anyone has time. If not, that's cool too; I think it looks pretty good.

dsplaisted:
I'm not clear on how this changes the user experience.  It seems like it's moving an error out of `ResolvePackageAssets` into a new `CheckRuntimeIdentifier` task and target.  But isn't the error and error message pretty similar in either case?

On a technical level, this bypasses the caching that `ResolvePackageAssets` does.  The assets file can get big for large projects, so there can be some overhead with reading it.  The `ResolvePackageAssets` caches its output in an optimized format to avoid reading the assets file if its not necessary.  If the `CheckRuntimeIdentifier` task has to load the assets file anyway, we lose the benefit of that optimization.

BogdanYarotsky:
@dsplaisted thank you for the feedback. I agree that it doesn't change much but rather moves error handling from one place another. As I understood from the open issue the current error message didn't give the clear instructions to the user who reported this: https://github.com/dotnet/sdk/issues/29859#issuecomment-1383193991

Since I'm not that familiar with the codebase I thought to extend it instead of modyifing existing functionality. But since ResolvePackageAssets is built with performance in mind - I think that changing the build error when target RID not found - would be a simpler/better approach. (It's also tested which will reduce total amount of changes). What do you think about this solution?

dsplaisted:
> What do you think about this solution?

That sounds fine.  The current error message is generated in `LockFileExtensions.GetTargetAndThrowIfNotFound`, which is mostly used by `ResolvePackageAssets` but also in a few other places.  I think that would be the place to make the change, either by changing the existing error messages, or adding some logic to add a new error message (for example, if the TargetFramework was found in the assets file but the RuntimeIdentifier wasn't).

Thanks!

nagilson:
> adding some logic to add a new error message (for example, if the TargetFramework was found in the assets file but the RuntimeIdentifier wasn't).

I agree with this suggestion, I'm not sure if the error had changed since writing the issue, but it does seem to be fairly good. It could be improved, however, with Danie's suggestion to be more specific (either if the TFM is missing or the RID.) I'm sorry to have you spend time and have to change the approach, that is on me for not writing up the issue clearly.

BogdanYarotsky:
@nagilson don't worry, it's all good :) Thank you and @dsplaisted for coming back with the feedback!
I'll give it another shot a bit later with this info in mind.

BogdanYarotsky:
Hi @nagilson, sorry for long time without a response. I have re-examined all the information from the original issue and comments in this PR. After spending some time with it I think that there is no technical problem here. The existing error message clearly says that restore needs to run before invoking `dotnet clean` and runtime identifier must be set in csproj. Doing so fixes the issue. Even the original poster of #29859 said that adding RID helped. And this was exactly what an error NETSDK1112 suggested to him:
`The runtime pack for Microsoft.NETCore.App.Runtime.win-x64 was not downloaded. Try running a NuGet restore with the RuntimeIdentifier 'win-x64'. [C:\sc\ws1\YOV Services\Account\Account.EF\Account.EF.csproj]`

The only problem that I see here is that adding a runtime identifier means that `dotnet clean` will never clean the artifacts from builds before a runtime identifier was added. So your proposal to add `dotnet clean --all` could be used for this. But it's a whole different thing to tackle.

In any case, I'll close this PR as it's way too behind the current main.
Thank you for your time!

PR #35594, opened 2023-09-20
We don't need this 🙃

MiYanni:
## Summary
When looking over some CLI code, I kept noticing `this`. Since `this` doesn't do anything, I removed `this`. In some codebases, `this` can help differentiate between 2 different members. Here, I did not find a situation that requires `this`. Since `this` appears grey in VS, let's just remove `this`.


smitpatel (approved):

Forgind (approved):
Nice name 😁

Issue #34347, opened 2023-07-31
`dotnet tool restore` throws `ArgumentOutOfRangeException`


AristurtleDev:
### Describe the bug
When executing `dotnet tool restore`, it throws an exception stating `ArgumentOutOfRangeException`.  This seems to only occur when the localization of the PC is set to any language that includes non-Latin characters, such as Thai.  

This seems to be related to the same issue that happened with `dotnet run` in Issue #23004 

The full exception is included below in the exception section.

### To Reproduce
Set the localization language of your PC to any language that includes non-Latin characters. In this case, to reproduce the issue as we saw it, set the localization language to Thai.

After doing this, attempt to install any dotnet tool, for instance you can try `dotnet tool install -g dotnet-ef` and the error will occur.

As for the specific instance we noticed this issue, it was while helping a fellow community member diagnose the issue when creating a MonoGame project which includes several dotnet tools that needed to be restored.  A basic sample of this project can be found at https://github.com/AristurtleDev/MonoGameSample

Before attempting use the sample project, ensure you have set your PC localization to anything that includes non-Latin characters such as Thai in our specific scenario. 

### Exceptions (if any)
```sh
 at System.CommandLine.CommandLineBuilderExtensions.<>c__DisplayClass7_0.<<UseExceptionHandler>b__0>d.MoveNext()
System.ArgumentOutOfRangeException: startIndex cannot be larger than length of string. (Parameter 'startIndex')
   at System.String.ThrowSubstringArgumentOutOfRange(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Microsoft.DotNet.Cli.RestoreCommandParser.RestoreRuntimeArgFunc(IEnumerable`1 rids)
   at Microsoft.DotNet.Cli.ForwardedOption`1.<>c__DisplayClass6_0.<SetForwardingFunction>b__0(T o)
   at System.Linq.Enumerable.SelectManySingleSelectorIteratSystem.ArgumentOutOfRangeException: startIndex cannot be larger than length of string. (Parameter 'startIndex')
   at System.String.ThrowSubstringArgumentOutOfRange(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Microsoft.DotNet.Cli.RestoreCommandParser.RestoreRuntimeArgFunc(IEnumerable`1 rids)
[...]

PR #35637, opened 2023-09-21
Update RID parsing's `LastIndexOf` call to use invariant culture

JL03-Yue:
Per #34347, update RID parsing's `LastIndexOf` call to use invariant culture


dsplaisted (approved):
Looks good.  Can you retarget this to release/8.0.2xx (and make sure to get rid of any additional changes that brings in)?  Then we should talk about what other branches we want to backport this to.

MiYanni (commented):
General question: Is there a reason to use static methods for parsing the RID instead of just having a `RID` class that does a `Split` once and saves the parts as properties? My thought is that having static methods encourages reparsing (which we wouldn't want) whereas a `RID` class encourages passing around an instance of itself for use where the info is needed.
CC @dsplaisted

Issue #22882, opened 2021-12-02
Always provide existing version information in `dotnet workload list --machine-readable`, even if no update is available


sandyarmstrong:
### Is your feature request related to a problem? Please describe.
From a CI perspective, we need to be able to deterministically check whether a specific version of a given workload is installed. Currently we do this by locating the appropriate `WorkloadManifest.json` and verifying the version/etc from there. This is not ideal because it depends on implementation details that could change on us down the road.

### Describe the solution you'd like
`dotnet workload list --machine-readable` provides JSON output that could solve this problem, but right now it is focused on updater scenarios. It only shows `existingManifestVersion` for a workload if there is an update available for it:

```json
{
    "installed": [
        "macos",
        "ios"
    ],
    "updateAvailable": [
        {
            "existingManifestVersion": "12.0.101-preview.10.249",
            "availableUpdateManifestVersion": "12.0.101-preview.10.251",
            "description": ".NET SDK Workload for building macOS applications.",
            "workloadId": "macos"
        },
        {
            "existingManifestVersion": "15.0.101-preview.9.31",
            "availableUpdateManifestVersion": "15.0.101-preview.10.251",
            "description": ".NET SDK Workload for building iOS applications.",
            "workloadId": "ios"
        }
    ]
}
```

 If no update is available, you only see something like this (no version information for `macos` workload):

```json
{
    "installed": [
        "macos",
        "ios"
    ],
    "updateAvailable": [
        {
            "existingManifestVersion": "15.0.101-preview.9.31",
            "availableUpdateManifestVersion": "15.0.101-preview.10.251",
            "description": ".NET SDK Workload for building iOS applications.",
            "workloadId": "ios"
        }
    ]
}
```

Ideally, `existingManifestVersion` would have been in the `installed` array, regardless of any updates. To avoid changing behavior other tools might depend on, it's probably necessary to add a new top-level property. Maybe something like this:

```json
{
    "installed": [
        "macos",
        "ios"
    ],
    "updateAvailable": [
        {
            "existingManifestVersion": "15.0.101-preview.9.31",
            "availableUpdateManifestVersion": "15.0.101-preview.10.251",
            "description": ".NET SDK Workload for building iOS applications.",
            "workloadId": "ios"
        }
    ],
    "installedDetails": {
        "macos": {
            "manifestVersion": "12.0.101-preview.10.251"
        },
        "ios": {
            "manifestVersion": "15.0.101-preview.9.31"
        }
    }
}
```

[...]

PR #35640, opened 2023-09-22
Workload commands: JSON only output adjustments

### LINKED ISSUE #22887: [workload] Machine readable outputs should only output JSON

Commands like `dotnet workload update --print-rollback` and `dotnet workload list --machine-readable` are very helpful, but they should print only JSON to stdout.

Right now I run `dotnet workload list --machine-readable` and this is what gets printed to stdout:

```
Failed to update the advertising manifest microsoft.net.sdk.tvos: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.sdk.android: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.sdk.maui: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.workload.emscripten: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.sdk.macos: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.sdk.maccatalyst: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.sdk.ios: Unable to load the service index for source https://REDACTED/index.json..
Failed to update the advertising manifest microsoft.net.workload.mono.toolchain: Unable to load the service index for source https://REDACTED/index.json..
==workloadListJsonOutputStart==
{"installed":["macos","ios"],"updateAvailable":[{"existingManifestVersion":"12.0.101-preview.10.249","availableUpdateManifestVersion":"12.0.101-preview.10.251","description":".NET SDK Workload for building macOS applications.","workloadId":"macos"},{"existingManifestVersion":"15.0.101-preview.9.31","availableUpdateManifestVersion":"15.0.101-preview.10.251","description":".NET SDK Workload for building iOS applications.","workloadId":"ios"}]}
==workloadListJsonOutputEnd==
```

I have to process this and it complicates my tooling.

Commands like `rg`, `lsblk`, `losetup`, etc that offer JSON output modes _only_ print JSON to stdout. Everything else goes to stderr. `dotnet workload` commands that are intended to offer machine-readable output would be most useful if they functioned similarly.

This opens up much easier shell scripting workflows. In powershell you could pipe to `ConvertFrom-Json` without any other ceremony. In other shells you could pipe to `jq`.

### Additional context

I would imagine this would break compatibility with tools used to the current output, so unfortunately it would require another option, like `--really-machine-readable`.

MiYanni:
Fixes: https://github.com/dotnet/sdk/issues/22887

> [!NOTE]
> This PR is a breaking change. 
> I do plan to also work on [this issue](https://github.com/dotnet/sdk/issues/22882) which would be good to lump together in the same breaking change doc.

## Summary
There were a couple of primary changes:
- Removed the Start/End boundary lines from the JSON only workload commands
- Only have the JSON being written (no logging messages) for the JSON only workload commands

Commands adjusted:
- `dotnet workload list --machine-readable`
- `dotnet workload install --print-download-link-only`
- `dotnet workload update --print-download-link-only`
- `dotnet workload update --print-rollback`

## Details
The first change (boundary lines) was very simple and just involved removing a couple of `WriteLine` outputs for each command. Then, adjusting tests accordingly.

The second change was a bit more involved. We use a `Reporter` to write out to the console. Think of it as just an output stream. Right now, these commands are designed to:
- Pass an instance of `Reporter` down through the workload command hierarchy.
- If one is not provided, the default `Reporter.Output` will be used.
- The `Reporter` property can only be set via this constructor hierarchy.

For the commands in question, we need to:
- Use the configured `Reporter` to write the JSON appropriately
- Every other usage of `Reporter` beyond writing the JSON is replaced with `NullReporter.Instance`

To do this was a bit tricky. I had to add the ability to pass a `Reporter` to several methods that utilized it. But there was something else that also needed adjusted. The `PackageDownloader` is also created deep in the hierarchy using the `Reporter` from the constructor. Since I actually want to utilize the `Reporter` being passed in to write the JSON, I can't just force the `Reporter` property to be `NullReporter`.

So, I ended up creating a new instance of `NuGetPackageDownloader` for this situation. It also allowed me to use a `NullLogger` which applies to the NuGet messages. For unit testing, I had to add `IsPackageDownloaderProvided` to indicate that a `PackageDownloader` was passed through the constructor hierarchy. Then, I use that `PackageDownloader` instead of creating a new one.

### DIFF
--- a/src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
+++ b/src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
@@ -82,12 +82,14 @@ namespace Microsoft.DotNet.Workloads.Workload
             _workloadManifestUpdaterFromConstructor = workloadManifestUpdater;
         }
 
-        protected async Task<List<WorkloadDownload>> GetDownloads(IEnumerable<WorkloadId> workloadIds, bool skipManifestUpdate, bool includePreview, string downloadFolder = null)
+        protected async Task<List<WorkloadDownload>> GetDownloads(IEnumerable<WorkloadId> workloadIds, bool skipManifestUpdate, bool includePreview, string downloadFolder = null,
+            IReporter reporter = null, INuGetPackageDownloader packageDownloader = null)
         {
+            reporter ??= Reporter;
+            packageDownloader ??= PackageDownloader;
+
             List<WorkloadDownload> ret = new();
-
             DirectoryPath? tempPath = null;
-
             try
             {
                 if (!skipManifestUpdate)
@@ -109,7 +111,7 @@ namespace Microsoft.DotNet.Workloads.Workload
 
                     if (!manifestDownloads.Any())
                     {
-                        Reporter.WriteLine(Strings.SkippingManifestUpdate);
+                        reporter.WriteLine(Strings.SkippingManifestUpdate);
                     }
 
                     foreach (var download in manifestDownloads)
@@ -117,8 +119,8 @@ namespace Microsoft.DotNet.Workloads.Workload
                         //  Add package to the list of downloads
                         ret.Add(download);
 
-                        //  Download package                        
-                        var downloadedPackagePath = await PackageDownloader.DownloadPackageAsync(new PackageId(download.NuGetPackageId), new NuGetVersion(download.NuGetPackageVersion),
+                        //  Download package
+                        var downloadedPackagePath = await packageDownloader.DownloadPackageAsync(new PackageId(download.NuGetPackageId), new NuGetVersion(download.NuGetPackageVersion),
                             _packageSourceLocation, downloadFolder: folderForManifestDownloads);
 
                         //  Extract manifest from package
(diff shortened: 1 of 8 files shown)


Forgind:
Can you explain why you had to pass in the null reporter as opposed to just removing more instances of `Reporter.WriteLine` or equivalent? I'm thinking that if we decide the same NuGetPackageDownloader should print out something, we'd have to undo this change and fix it another way, then add on that change. At minimum, perhaps we could use a boolean to say "write stuff" versus not?

MiYanni:
@Forgind 
> Can you explain why you had to pass in the null reporter as opposed to just removing more instances of `Reporter.WriteLine` or equivalent?

The lines written by those other classes are used by all the other commands that don't feature the flags that print out JSON. So, the commands (`list`, `install`, `update`) that don't have the flags (`--machine-readable`, `--print-download-link-only`, `--print-rollback`) respectively will still use the normal `Reporter` and print these lines as expected. The purpose of the PR is to make these flags only allow JSON to be printed so tooling can pipe the output if these commands directly into a JSON parser.

> At minimum, perhaps we could use a boolean to say "write stuff" versus not?

The main issue is just a result of the class hierarchy and construction. Also, the combination of the "good path" constructor (path the code normally takes when running) with the "unit test" constructor makes it a bit convoluted in how to solve this without breaking tests. Most designs have these as 2 separate constructors.

Adding a Boolean doesn't buy us much, other than adding boilerplate code around every `Reporter.WriteLine` callsite. It is easier to have a `Reporter` that just does nothing when called. A potentially better design would be to just have a global `Reporter` we just switch the operating mode when needed. That design is your Boolean idea but it is a global "lightswitch" for reporting as opposed to a Boolean that only works for these specific commands and classes. However, that feels like a bigger endeavor than is needed to solve this particular scenario. My heart *yearns* for a rewrite, but it isn't a priority at the moment.

Forgind:
Thanks for the explanation! So I guess the only case in which we'd have to roll back part of this change would be if we decided to add text to one of those subcommands...which seems unlikely.

On the global boolean idea, it might be interesting to have an optional argument on WriteLine that, if false (default: true), would only write the line if the global boolean is true. I do think that would be a nice, clean structure and very forward-compatible, but I think it's fine to skip it for now since you said it isn't a priority 🙂

Forgind (approved):
Looks good to me! I was thinking about proposing either unifying the new NuGetPackageDownloader calls in one place, but I don't think that's really worth it. You can't just do it in place of the boolean unless you want it for commands that don't need it, which would be wasteful.

MiYanni:
Breaking change issue for dotnet/docs created here: https://github.com/dotnet/docs/issues/37278

PR #35661, opened 2023-09-22
branding update for 8.0.2xx

Forgind:


nagilson:
I think you need to change `MirrorBranch: release/8.0.1xx` to 2xx in `eng/build.yml`

Forgind:
If I'm understanding the example set by https://github.com/dotnet/sdk/pull/29942 properly, I think that change should really go into the final branding PR for 8.0.1xx? It'll flow to 8.0.2xx later, but it should be there first. Am I wrong about that?

nagilson:
> If I'm understanding the example set by #29942 properly, I think that change should really go into the final branding PR for 8.0.1xx? It'll flow to 8.0.2xx later, but it should be there first. Am I wrong about that?

Hrm, good point... though that is the final branding for 7.0.2xx, its changing the mirror to 7.0.3xx. Based on that, I think it would be appropriate to say it was 7.0.2xx from the beginning of the branch, which is what this change would be? 

Anyway, I'm not confident, especially if it wasn't in the list of steps to do, and I don't remember having to do that manually. @marcpopMSFT  Is this supposed to be changed?

marcpopMSFT:
That controls which branch we localize. It can be done in 8.0.1xx later and flowed up. I've often done it at the same time as branching just so I don't have to remember later but either way is ok.

marcpopMSFT (approved):

Issue #34810, opened 2023-08-22
[Repo] Add automatic style/formatting mechanism


MiYanni:
Discussion based on: https://github.com/dotnet/sdk/pull/34659#pullrequestreview-1581582025

## High-Level Goals
- Applies formatting when building via `build.cmd`
- Applies formatting when using Visual Studio and/or VSCode (on-build or otherwise)
- **Does not** cause build errors related to style/formatting
- *(Nice to have)* Support format on save behavior

## Quick Details
The list above is a compilation of the discussion from the [PR on running `dotnet format`](https://github.com/dotnet/sdk/pull/34659). The main formatting that is applied in that PR is:
- Whitespace cleanup
- Using statement reordering

There are likely other aspects we could have but those are the basics. Other agreed-upon conventions can be enabled in our `.editorconfig`.

PR #35675, opened 2023-09-22
IDE0002 "Name can be simplified" cleanup

MiYanni:
## Summary
I've been noticing a lot of random greyed out code in the codebase. So, I'm going through and using tooling to apply fixes for what the analyzers are detecting. Here, I ran the fix for [Simplify member access (IDE0002)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0002). These types of automated changes are non-functional changes. Just purely code cleanup.


src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.cs
```diff
@@ -511,7 +511,7 @@ private static void WriteListCommandExample(InstantiateCommandArgs instantiateAr
                   Example
                       .For<NewCommand>(instantiateArgs.ParseResult)
                       .WithSubcommand<ListCommand>()
-                      .WithArgument(ListCommand.NameArgument, instantiateArgs.ShortName));
+                      .WithArgument(BaseListCommand.NameArgument, instantiateArgs.ShortName));
```
Forgind, line 514: Was this done automatically? Why is it ok?

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -145,7 +145,7 @@ public int ExecuteInProc(string[] arguments)
                 try
                 {
                     // Execute MSBuild in the current process by calling its Main method.
-                    return Microsoft.Build.CommandLine.MSBuildApp.Main(arguments);
```
Forgind, line 148: Specifically on this one, "Microsoft.Build" feels like a unit to me, so I'd rather keep the greyed-out part.

Forgind (approved):
Outside of the MSBuild case and the other thing I commented on, I like this change.

nagilson (approved):

src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.cs
nagilson: I agree with Nathan, I prefer the 'ListCommand' over 'BaseListCommand'... same for 'BaseUninstallCommand' vs 'UninstallCommand', etc... all of the other stuff looks awesome though :)

nagilson:
Thanks Yanni for cleaning stuff up, I like little changes like these :)

src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.cs
MiYanni: So, the way the rule works is that it tries to apply the most *precise* way of accessing members. Here, `NameArgument` doesn't exist as a member on `ListCommand` directly. `NameArgument` is a static member in `BaseListCommand`. However, `ListCommand` derives of `BaseListCommand` and static members can be accessed transitively through it.

Honestly, asking about this situation convinced me *why* this rule should be done, as opposed to changing it back. The reason is that this shows the design choices being made and brings to light how things could be architected differently. The question that could be asked here instead is, "Why is InstantiateCommand accessing an argument from the BaseListCommand?" Other places in these CLI commands use a `SharedOptions` class for this purpose of keeping static options. Maybe creating a `SharedArguments` would be a cleaner design? That's kinda how I think about things. 🙃

Forgind: I can see your argument. I'm guessing the initial reason for why it accessed an argument from BaseListCommand is that it was there, and someone needed it, so they used it—more pragmatic than architectural.

One concern I'd have with switching to the BaseListCommand is that if the person who wrote this actually was thinking architecturally, perhaps there's a concept of a "NameArgument" on ListCommand that differs from the concept of a "NameArgument" on BaseListCommand in some critical way. I can't think of a reason here, but say there were an "UpdateCounter" method on both, but the ListCommand has an extra counter to update. ListCommand would then override that base method. We'd want ListCommand's version regardless of whether it had actually overridden it yet...so if it isn't overridden yet, VS might suggest a change like this, but then we wouldn't automatically pick up the override from ListCommand if it comes.

That was a little contorted; did it make sense?

MiYanni: @Forgind 
> We'd want ListCommand's version regardless of whether it had actually overridden it yet...so if it isn't overridden yet, VS might suggest a change like this

It doesn't. The analyzer has contextual awareness of how the classes are constructed. There could be situations like those that exist in the codebase right now, but this PR wouldn't have changed those. From my basic understanding of the IL, they are essentially 2 different callsites. The fact they have the same name only applies to the user (aka us, the devs) as we'd need to add a `new` to override the name conflict. Roslyn is used to drive these analyzers so it has those smarts. Trust the tooling 😉 (until we make it **too** smart and it becomes Skynet, and now we fear for our mortal lives... 💥💀)

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
MiYanni: @Forgind I looked at this file and I have a couple of questions. If you have any answers, I'd appreciate it.
- Is there a reason this was fully-qualified to begin with? Usually, this is only done to avoid naming clashes. I didn't see anything else named `MSBuildApp`.
- This file is wrapped in a `#if NET`. What is the purpose of that? It seems to mess with some of the automatic tooling, at least when trying to "Clean Up Usings" which I attempted to do.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.cs
Forgind: Yeah; I agree this won't break anything. For the example I provided with UpdateCounter, the idea was that at the outset (i.e., now) ListCommand doesn't override BaseListCommand's implementation of UpdateCounter, but if it does so in the future, which should we use? I'd imagine we should use ListCommand's version, and I think this would make us use BaseListCommand's.

src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
Forgind: > Is there a reason this was fully-qualified to begin with? Usually, this is only done to avoid naming clashes. I didn't see anything else named MSBuildApp.

I didn't write that 😆

The function it's invoking isn't in the SDK at all. It's importing it from MSBuild. VS needs to know where to get the type, and Microsoft.Build is in context, so it can go from there to find MSBuildApp. Alternatively, you can just add a using at the top:
`using Microsoft.Build.CommandLine;`

And then you can skip the qualification and just call MSBuildApp.Main.

The previous version seems to have used reflection, so this was probably chosen as the more similar way that doesn't use reflection. Personally, I like the full qualification in this case so I know it doesn't come from the SDK, but some would argue that I should be able to figure that out by mousing over it or looking at the using. ¯\_(ツ)_/¯

> This file is wrapped in a #if NET. What is the purpose of that? It seems to mess with some of the automatic tooling, at least when trying to "Clean Up Usings" which I attempted to do.

It appears that this one is compiled for both framework and core but only used on framework. It has some things that only exist on core, including this MSBuildApp.Main call (the framework version has no arguments...and also doesn't show up? But I think that's just a using-type issue), and it isn't used on framework, so it may as well only exist on core. The cleaner solution is to just not compile the file at all, but I'm guessing someone added the folder with it and didn't bother adding an Exclude. If you Exclude this file and remove the #if parts, I'll sign off.

src/Cli/Microsoft.TemplateEngine.Cli/Commands/create/InstantiateCommand.cs
MiYanni: In this context, there is no intrinsic relationship between the usage in this code and the base/non-base version of `ListCommand`. This code is the `dotnet new create` command (aka, `dotnet new <TEMPLATE>`) and the `Base/ListCommand` is from `dotnet new list`. Picking either one could make sense depending on the goal of the developer. Personally, I don't think there is any native bias between one or the other.

If there is a `new`ed version of `NameArgument`, the developer would need to be aware to investigate *all* callsites of the old one either way. One cannot assume the new version would work in every situation.

v-wuzhai:
@MiYanni Can these PRs changes be included in a manual scheduled maintenance schedule, or can we have robots do it on a regular basis?

MiYanni:
@v-wuzhai 
> Can these PRs changes be included in a manual scheduled maintenance schedule, or can we have robots do it on a regular basis?

I previously added an issue to address this request [here](https://github.com/dotnet/sdk/issues/34810). I'll be digging into the 'regular maintenance' side of things later. For now, I'm just trying to establish a baseline of a somewhat 'clean' repo.

PR #35678, opened 2023-09-22
IDE0200 "Lambda expression can be removed" cleanup

MiYanni:
## Summary
I've been noticing a lot of random greyed out code in the codebase. So, I'm going through and using tooling to apply fixes for what the analyzers are detecting. Here, I ran the fix for [Remove unnecessary lambda expression (IDE0200)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0200). These types of automated changes are non-functional changes. Just purely code cleanup.


Forgind:
My recollection on this one is that yes, that's correct, but it's actually less performant because it recreates the function repeatedly instead of storing it in a static. I think @ladipro taught me that; am I correct that this change would hurt performance?

baronfel:
C# 10 or 11 learned how to cache delegates that are passed as lambda parameters, so now the correct thing doesn't result in any performance hit.

MiYanni:
@KalleOlaviNiemitalo 
> If the called method throws, this change can make the stack trace of the exception harder to understand.

So, I made a little test code to see how this works:

image

From the `Lambda` section, you get a callstack like this every time it stops at the WriteLine call;

image

From the `Method` section, you get a callstack like this every time it stops at the WriteLine call;

image

The difference is, it has an extra call in the callstack for the lambda version, shown as `AnonymousMethod`. In the IDE, it highlights the call to `Write(s)`. For the method version, there is no extra entry in the callstack. However, both will show were the start of the "loop" occurs. You can always walk the stack to find this callsite. Personally, I don't think this buys you much since this only will happen in this scenario. If you have multiple statements, you still need the lambda expression. This only is for optimizing the callstack and callsite to reduce boilerplate for using these methods.

I've often adapted the shape of my methods to fit the syntax necessary for these situations, so I can avoid introducing these extraneous lambdas, which I think is *generally* a good practice. Keep in mind, if the method has any additional parameters, this situation also won't occur and a lambda will be needed. It is strictly for accepting the type of the collection and processing it in some way.

If we don't want this pattern, I can turn off the suggestion in the `.editorconfig` for it.

nagilson:
I agree with Kalle's concerns, while I appreciate the thorough investigation into the call stacks, I think I would rather turn off this suggestion. I also dont like leaving things to be implicit in code if they can be explicit, given they aren't redundant/obvious

MiYanni:
Closing in favor of https://github.com/dotnet/sdk/pull/35797

PR #35679, opened 2023-09-22
IDE0017 "Object initialization can be simplified" cleanup

MiYanni:
## Summary
I've been noticing a lot of random squiggles in the codebase. So, I'm going through and using tooling to apply fixes for what the analyzers are detecting. Here, I ran the fix for [Use object initializers (IDE0017)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0017). These types of automated changes are non-functional changes. Just purely code cleanup.

### DIFF
--- a/src/BuiltInTools/dotnet-watch/Internal/FileWatcher/PollingFileWatcher.cs
+++ b/src/BuiltInTools/dotnet-watch/Internal/FileWatcher/PollingFileWatcher.cs
@@ -31,9 +31,11 @@ namespace Microsoft.DotNet.Watcher.Internal
             _watchedDirectory = new DirectoryInfo(watchedDirectory);
             BasePath = _watchedDirectory.FullName;
 
-            _pollingThread = new Thread(new ThreadStart(PollingLoop));
-            _pollingThread.IsBackground = true;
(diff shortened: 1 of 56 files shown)


Forgind:
There are some cases of this that I support like with the pattern:
```C#
myObject = new()
{
thing = thing1
}
myObject.thing2 = thing3
```

getting changed. But there are others where I don't really like it. In general, if there are more lines of code than there were before, I probably think it's a mildly negative change. What do others think?

nagilson:
> There are some cases of this that I support like with the pattern:
> 
> ```cs
> myObject = new()
> {
> thing = thing1
> }
> myObject.thing2 = thing3
> ```
> 
> getting changed. But there are others where I don't really like it. In general, if there are more lines of code than there were before, I probably think it's a mildly negative change. What do others think?

IMO, I like this change. I think it 'chunks' the code into more readable units with the brackets, instead of having many lines where it _could_ be object initialization, or it could be something totally unrelated. But I see where you're coming from and I'm not super strong one way or the other

nagilson (approved):

PR #35680, opened 2023-09-23
IDE0049 "Name can be simplified" cleanup

MiYanni:
## Summary
I've been noticing a lot of random greyed out code in the codebase. So, I'm going through and using tooling to apply fixes for what the analyzers are detecting. Here, I ran the fix for [Use language keywords instead of framework type names for type references (IDE0049)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0049). These types of automated changes are non-functional changes. Just purely code cleanup.


nagilson (approved):
This looks good, thank you for cleaning stuff up! :)

PR #35697, opened 2023-09-25
Make CheckForDuplicateItems case insensitive

### LINKED ISSUE #35693: CheckForDuplicateItems case sensitive
The task CheckForDuplicateItems is case sensitive, potentially missing some duplicates that differ in case.

The fix would be on this line to pass a comparer:
https://github.com/dotnet/sdk/blob/c382c93f8dcad09417409799595a88da407e7e6d/src/Tasks/Microsoft.NET.Build.Tasks/CheckForDuplicateItems.cs#L38

MSBuild's [RemoveDuplicates](https://learn.microsoft.com/en-us/visualstudio/msbuild/removeduplicates-task?view=vs-2022) task is case insensitive, so this would align with MSBuild standard processing

stan-sz:
Fixes #35693

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckForDuplicateItems.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckForDuplicateItems.cs
@@ -49,7 +49,7 @@ namespace Microsoft.NET.Build.Tasks.UnitTests
             // execute task
             var task = new CheckForDuplicateItems()
             {
-                engine = engine,
+                BuildEngine = engine,
                 Items = compile,
                 ItemName = "Compile",
                 PropertyNameToDisableDefaultItems = "PropertyNameToDisableDefaultItems",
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenACheckForDuplicateItems.cs
```diff
@@ -0,0 +1,68 @@
...
+                ItemName = "Compile",
+                PropertyNameToDisableDefaultItems = "PropertyNameToDisableDefaultItems",
+                MoreInformationLink = "MoreInformationLink",
+                DefaultItemsEnabled = true,
+                DefaultItemsOfThisTypeEnabled = true
+            };
+            task.Execute().Should().BeFalse();
+
+            engine.Errors.Count.Should().Be(1);
+            engine.Errors[0].Message.Should().Be("Duplicate 'Compile' items were included. The .NET SDK includes 'Compile' items from your project directory by default. You can either remove these items from your project file, or set the 'PropertyNameToDisableDefaultItems' property to 'false' if you want to explicitly include them in your project file. For more information, see MoreInformationLink. The duplicate items were: 'foo.cs'");
```
dsplaisted, line 63: Generally we would avoid checking exact error strings in a test.  Usually we would usually just check for the error code (`NETSDK1022` in this case), or we would use the same source string (`Strings.DuplicateItemsError`) to compare with so that if we update the error text, we don't have to duplicate the change in the tests.

dsplaisted (approved):
Thanks for your contribution!  This looks like a good change.

PR #35710, opened 2023-09-25
Add ProjectCapability for Native AOT properties.

melytc:
The `NativeAOT` capability is being used in project system to control the visibility of the Native AOT properties introduced recently to .NET 8. The idea is that, in general, we want to set this capability to true so that we show these properties for projects targeting .NET 8 or higher. 

The exceptions at the moment would be class libraries, WPF and WinForms projects, which are filtered using the property condition. Other project types, like MAUI and Web projects, can disable this capability by removing the `ProjectCapability` in their props file: `<ProjectCapability Remove="NativeAOT"/>`. 

This PR was created in lieu of https://github.com/dotnet/sdk/pull/34485 to be considered for the release branch 8.0.1xx.

Fixes [AB#1831643](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1831643)

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1278,6 +1278,20 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="CrossPlatformExecutable" />
   </ItemGroup>
 
+  <!-- Native AOT compilation -->
+  <ItemGroup>
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
+  </ItemGroup>
+
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
+  <PropertyGroup>
+    <ShowNativeAOTProperties Condition="'$(UseWPF)' != 'true'
+                             and '$(UseWindowsForms)' != 'true'
+                             and '$(ShowNativeAOTProperties)' == ''
+                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                             and $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '8.0'))">true</ShowNativeAOTProperties>
+  </PropertyGroup>
+
   <!-- Reference Manager capabilities -->
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
     <ProjectCapability Remove="ReferenceManagerAssemblies" />
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1279,18 +1279,18 @@ Copyright (c) .NET Foundation. All rights reserved.
   </ItemGroup>
 
   <!-- Native AOT compilation -->
-  <ItemGroup>
-    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
-  </ItemGroup>
-
   <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
   <PropertyGroup>
-    <ShowNativeAOTProperties Condition="'$(UseWPF)' != 'true'
+    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
+                             and '$(UseWPF)' != 'true'
                              and '$(UseWindowsForms)' != 'true'
-                             and '$(ShowNativeAOTProperties)' == ''
                              and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
                              and $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '8.0'))">true</ShowNativeAOTProperties>
   </PropertyGroup>
+
+  <ItemGroup Condition="'$(ShowNativeAOTProperties)' == 'true'">
+    <ProjectCapability Include="NativeAOTProperties"/>
+  </ItemGroup>
 
   <!-- Reference Manager capabilities -->
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
@@ -1280,16 +1280,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <!-- Native AOT compilation -->
   <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
-  <PropertyGroup>
-    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
-                             and '$(UseWPF)' != 'true'
-                             and '$(UseWindowsForms)' != 'true'
-                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
-                             and $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '8.0'))">true</ShowNativeAOTProperties>
-  </PropertyGroup>
-
-  <ItemGroup Condition="'$(ShowNativeAOTProperties)' == 'true'">
-    <ProjectCapability Include="NativeAOTProperties"/>
+  <ItemGroup Condition="'$(UseWPF)' != 'true'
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1278,6 +1278,20 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="CrossPlatformExecutable" />
   </ItemGroup>
 
+  <!-- Native AOT compilation -->
+  <ItemGroup>
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
+  </ItemGroup>
+
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
```
eerhardt, line 1286: (nit) It reads easier if this property group is above the ItemGroup that uses it.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1278,6 +1278,20 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="CrossPlatformExecutable" />
   </ItemGroup>
 
+  <!-- Native AOT compilation -->
+  <ItemGroup>
+    <ProjectCapability Include="NativeAOTProperties" Condition="'$(ShowNativeAOTProperties)' == 'true'"/>
```
eerhardt, line 1283: (nit) to be consistent with the ProjectCapability entries around it:

```suggestion
  <ItemGroup Condition="'$(ShowNativeAOTProperties)' == 'true'">
    <ProjectCapability Include="NativeAOTProperties" />
```

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1278,6 +1278,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  </ItemGroup>
+
+  <!-- Native AOT properties should be shown by default for projects targeting .NET 8 or higher, except for WPF and WinForms projects -->
+  <PropertyGroup>
+    <ShowNativeAOTProperties Condition="'$(UseWPF)' != 'true'
+                             and '$(UseWindowsForms)' != 'true'
+                             and '$(ShowNativeAOTProperties)' == ''
+                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                             and $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '8.0'))">true</ShowNativeAOTProperties>
```
eerhardt, line 1292:
```suggestion
    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
                             and '$(UseWPF)' != 'true'
                             and '$(UseWindowsForms)' != 'true'
                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
                             and $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '8.0'))">true</ShowNativeAOTProperties>
```

It's convention, and good practice, to check if the property is already set first. Makes it easier for readers to see that it is checked.

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.props
```diff
@@ -20,6 +20,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ServerGarbageCollection Condition="'$(ServerGarbageCollection)' == ''">true</ServerGarbageCollection>
     <IsPackable Condition="'$(IsPackable)' == ''">false</IsPackable>
     <WarnOnPackingNonPackableProject Condition="'$(WarnOnPackingNonPackableProject)' == '' and '$(IsPackable)' == 'false'">true</WarnOnPackingNonPackableProject>
+    <ShowNativeAOTProperties>false</ShowNativeAOTProperties>
```
eerhardt, line 23: Why is this set to `false`?

melytc: At the moment, we don't want to show these properties in web projects like ASP.NET. Is this the right place to set it for those types of projects?

melytc: Good question for @sayedihashimi. You can find more context on his spec on https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1831643/?triage=true

Forgind (approved):

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.props
melytc: If I remember correctly, ASP.NET already has a couple of properties, so there might be a plan to remove them after this is done.

eerhardt: I think you just want to turn this property off for "Web Assembly" / "Blazor WASM" projects. For "normal" ASP.NET Core projects, we definitely want to show Native AOT properties in the property page. We did a bunch of work in .NET 8 to make Native AOT work with ASP.NET Minimal APIs.

eerhardt (dismissed):
We should be showing Native AOT properties for ASP.NET Core server projects.

src/WebSdk/ProjectSystem/Targets/Microsoft.NET.Sdk.Web.ProjectSystem.props
melytc: I'll leave this out of this PR, thanks for the pointer.

eerhardt (approved):
Thanks!

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets
```diff
@@ -1278,6 +1278,20 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <PropertyGroup>
+    <ShowNativeAOTProperties Condition="'$(ShowNativeAOTProperties)' == ''
+                             and '$(UseWPF)' != 'true'
+                             and '$(UseWindowsForms)' != 'true'
+                             and '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                             and $([MSBuild]::VersionGreaterThanOrEquals('$(_TargetFrameworkVersionWithoutV)', '8.0'))">true</ShowNativeAOTProperties>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(ShowNativeAOTProperties)' == 'true'">
+    <ProjectCapability Include="NativeAOTProperties"/>
```
tmeschter, line 1292: The name `NativeAOTProperties` implies that the capability is only about showing certain properties in the UI. Would it instead make sense to have a more generic name such as "SupportsNativeAOT" and show the properties based on that? Or do we need to make a distinction between projects that support Native AOT and those that should show the properties?

eerhardt: IMO, "SupportsNativeAOT" could be interpreted so many different ways. I'm not sure it would be a good name for the ProjectCapability. For example, it could mean that the project successfully publishes to Native AOT and works correctly, which is not true for many (most?) apps.

"Available" or "Capable" might be better qualifiers.

Note if we rename the ProjectCapability, we should also rename the property `ShowNativeAOTProperties` to match.

melytc: I think we can make the distinction here between showing the properties and actually supporting AOT compilation (which I think it's covered by other properties like `PublishAot`, `IsAotCompatible`, `EnableAotAnalyzer`, etc.).

tmeschter: I think the "Capable" part is implied by the `ProjectCapability` item type. 😀 My preference would actually be to just call the capability `NativeAOT`, but I'd be fine with `NativeAOTAvailable` or `NativeAOTCapable`.

tmeschter: > Note if we rename the ProjectCapability, we should also rename the property ShowNativeAOTProperties to match.

Or maybe just get rid of the property?

eerhardt: > Or maybe just get rid of the property?

See the top comment:

> Other project types, like MAUI and Web projects, can disable this capability by setting the ShowNativeAOTProperties property to false in their props file. 

Would you rather them `<ProjectCapability Remove="NativeAOTProperties"/>`?

melytc: That makes sense, let me get rid of the property and rename the `ProjectCapability`.

tmeschter (approved):

PR #35712, opened 2023-09-25
Clean up polling file watcher

tmat:
Fix a couple of issues:
1) `DirectoryInfo.EnumerateFileSystemInfos` doesn't guarantee that the returned entries will be unique (https://github.com/dotnet/aspnetcore/issues/38719)
2) `FileSystemInfo.LastWriteTime` is initialized lazily. Previous implementation captured `FileSystemInfo` and then later compared its `LastWriteTime` with the new `FileSystemInfo` entry. If the file is updated in between these checks, while the former info is not initialized yet, the comparison would be comparing the same time stamps and miss the change.


tmat:
@tlmii ptal

MiYanni:
@tmat It looks like this PR is very outdated with `main`.

PR #35716, opened 2023-09-25
Add spelling exclusions file

MiYanni:
## Summary
In my investigation of VS squiggles for the repo, there were a number of them for spelling issues. I've added a `.dic` file that includes a list of words that can be treated as spelling exclusions. Details on the exclusions file can be [found here](https://learn.microsoft.com/en-us/visualstudio/ide/text-spell-checker?view=vs-2022#how-to-customize-the-spell-checker).

I couldn't figure out how to investigate **all** spelling issues for the SDK. Instead, I just flipped through a bunch of files and found random spelling mistakes that I could either fix or add to the exclusions list.

I encourage others to add more words to the `exclusion.dic` as they run into them.


danmoseley:
@MiYanni  how is the spell checker hooked up - is it the default thing I get in VS?

danmoseley:
I wonder whether there's a way we can share a suppression dictionary across our repos through Arcade?

Is there any way to run it on the command line?

Forgind:
> I wonder whether there's a way we can share a suppression dictionary across our repos through Arcade?

It looks like this was added through the editorconfig, so each repo would have to add it to their own editorconfig, but then Arcade could provide a exclusion dictionary for all repos.

> Is there any way to run it on the command line?

Is there a way to make the default spell checker run on the command line? I thought it wouldn't catch any spelling mistakes outside VS, and we'd need that before we can worry about an exclusion list.

Forgind (approved):
Love it!

MiYanni:
@danmoseley 
> how is the spell checker hooked up - is it the default thing I get in VS?

[Here](https://learn.microsoft.com/en-us/visualstudio/ide/text-spell-checker?view=vs-2022) is the article on the spell checker. It was added in VS in 17.5. AFAIK, it is on by default. [Here](https://learn.microsoft.com/en-us/visualstudio/ide/text-spell-checker?view=vs-2022#how-to-customize-the-spell-checker) is the section of the article on the spell checker for this exclusion setting.

> I wonder whether there's a way we can share a suppression dictionary across our repos through Arcade?

Since the exclusion setting is just a path, you could set it to `.\eng\common\exclusions.dic` for example. However, the setting it put in the `.editorconfig` which is normally at the root of the repo. So yeah, just mirroring what Nathan said about it.

> Is there any way to run it on the command line?

Based on [this](https://learn.microsoft.com/en-us/visualstudio/ide/text-spell-checker?view=vs-2022#how-does-it-work), I don't think so:
> [!Note]
> Feedback from early users of this feature informed us that developers only want to know about errors in the document they are currently working with. In response to that feedback, the spell checker only scans the document(s) that are active in the editor.

My intuition would lead me to think this would be a feature that would need to be added to `dotnet format` if we wanted a commandline version.

### Toward your Arcade proposal
If there was a way to run it for an entire solution, compiling a list of common exclusions would be easily possible. The best we could do right now is to do a GitHub search for `spelling_exclusion_path` (which is quite unique) and then aggregate many of the `.dic` files and compile a list manually. That's what I'd do if I was making a shared, multi-repo dictionary file (like one Arcade would distribute).

exclusion.dic
```diff
@@ -0,0 +1,14 @@
+awaiter
+crossgen
+cshtml
+deserializer
+nuget
+nupkg
+nuspec
+sdk
+sdks
```
nagilson, line 9:
```suggestion
sdks
msbuild
```

nagilson (approved):
This looks awesome, thanks @MiYanni  :)

exclusion.dic
MiYanni: I'll add it in alphabetical order. One sec.

src/ApiCompat/Microsoft.DotNet.ApiCompat.Task/SuppressibleMSBuildLog.cs
```diff
@@ -7,16 +7,16 @@
 namespace Microsoft.DotNet.ApiCompat.Task
 {
     /// <summary>
-    /// Class that can log Suppressions in an MSBuild task, by implementing MSBuildLog and ISuppressableLog.
+    /// Class that can log Suppressions in an MSBuild task, by implementing MSBuildLog and ISuppressibleLog.
     /// </summary>
-    internal sealed class SuppressableMSBuildLog : MSBuildLog, ISuppressableLog
+    internal sealed class SuppressibleMSBuildLog : MSBuildLog, ISuppressableLog
```
danmoseley, line 12:
```suggestion
    internal sealed class SuppressibleMSBuildLog : MSBuildLog, ISuppressibleLog
```
?

MiYanni: Oh golly. Okay. Let me get all those too.

MiYanni: Fixed all the uses of 'suppressable' in the repo.

ViktorHofer (approved):
Approving the APICompat changes.

MiYanni:
Note: This is currently bugged in some recent versions of VS. It isn't working for me currently on my desktop machine, and I wanted to know why. So, if you experience the same thing, we'll just have to wait for whatever VS version has this fix: https://developercommunity.visualstudio.com/t/VS2022-1752-Spell-Checker-Exclusion-Di/10328416

Issue #35753, opened 2023-09-27
Container tooling creates many HTTP connections


tmds:
I've been testing the SDK container tooling against various registries over the past two weeks.
As part of my test registries, I have 3 OpenShift test clusters available.

On two of the clusters I haven't seen issues. One the other one however, the tooling is failing more than it is succeeding.

I looked at the cluster configuration with a registry expert, and the conclusion is: this registry is configured with two backing pods and the the way they shared storage can cause consistency issues.

So the root cause for this issue isn't with the .NET implementation, but it is the registry storage configuration.

I have used this cluster before with the built-in tooling and in contrast to the .NET tooling, I have never seen it fail.
So the .NET tooling is somehow causing the consistency issues to appear more than the built-in container tooling.

After investigating further, I learned that the main difference is that the .NET client is using much more HTTP connections than the built-in client. Because it uses more connection, it's more likely to be using both backend pods while pushing an image.

The reason it's using many connection is that, connections are configured to be used no longer than 10ms:

https://github.com/dotnet/sdk/blob/2b89a6a1e70bb07f8c437e018de3056cbaf9d0f1/src/Containers/Microsoft.NET.Build.Containers/Registry/DefaultRegistryAPI.cs#L35

I don't think there's a reason to set the connection lifetime.
Sometimes this is set to avoid stale DNS information, but that is not expected for the duration of the registry operations.
Also, the value used here is very low and it will expire multiple times during a push.

Another (unexpected) source of an additional connection is in the auth handler:

https://github.com/dotnet/sdk/blob/2b89a6a1e70bb07f8c437e018de3056cbaf9d0f1/src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs#L323-L329

When the request that caused the Unauthorized response wasn't fully read, that keeps the connection "in-use".
Then when the new request is made for authentication, it's forced to create a new HTTP connection.

Every new connection comes at the cost of some round-trips for the tls handshake, and it can rapidly account for tens of milliseconds (per connection made).

cc @baronfel

baronfel:
Closed because the linked PR fixed this

PR #35754, opened 2023-09-27
Containers: reuse HTTP connections more.

tmds:
## Description

Closes https://github.com/dotnet/sdk/issues/35753

This removes the artificially-short Pool Lifetime used by the HttpClientHandler that the SDK Container publishing tooling uses to communicate with remote container registries.
This short lifetime was something we put in very early on in the lifetime of the code, and likely was part of a mitigation around our incorrect authentication handling in earlier versions.

## Customer Impact

This short lifetime lead to many connections getting used instead of fewer connections being re-used. Some registries may have configurations that rely on some connection-local state (like sticky sessions) and using more connections that required can impact those flows. In addition, reusing connections improves performance, because negotiating new connections takes non-trivial amounts of time.

## Regression

**No** - we've had this code since our early versions.

## Risk

**Low** - while this does change the configuration of the HttpClientHandler used for all registry communication, testing against our set of supported live registries shows no regression in behaviors, and manual testing confirms this.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
@@ -212,7 +212,7 @@ internal sealed partial class AuthHandshakeMessageHandler : DelegatingHandler
             Content = new FormUrlEncodedContent(parameters)
         };
 
-        HttpResponseMessage postResponse = await base.SendAsync(postMessage, cancellationToken).ConfigureAwait(false);
+        using HttpResponseMessage postResponse = await base.SendAsync(postMessage, cancellationToken).ConfigureAwait(false);
         if (!postResponse.IsSuccessStatusCode)
         {
             await postResponse.LogHttpResponseAsync(_logger, cancellationToken).ConfigureAwait(false);
@@ -260,7 +260,7 @@ internal sealed partial class AuthHandshakeMessageHandler : DelegatingHandler
             var message = new HttpRequestMessage(HttpMethod.Get, builder.ToString());
             message.Headers.Authorization = header;
 
-            var tokenResponse = await base.SendAsync(message, cancellationToken).ConfigureAwait(false);
+            using var tokenResponse = await base.SendAsync(message, cancellationToken).ConfigureAwait(false);
             if (!tokenResponse.IsSuccessStatusCode)
             {
                 throw new UnableToAccessRepositoryException(_registryName);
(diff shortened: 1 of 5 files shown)


tmds:
@baronfel I've targeted the 8.0 branch for now. I can change to the main branch is that is preferable.

@antonfirsov if you want, can you take a look at these HttpClient related changes and approve the PR if they look good to you?

baronfel:
This one almost certainly will not make RC2. We could consider for GA, though - I'll try to kick off a run of sdk-container-demo with this change to help serve as evidence that things still behave.

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -324,6 +324,10 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 }
                 else if (response is { StatusCode: HttpStatusCode.Unauthorized } && TryParseAuthenticationInfo(response, out string? scheme, out AuthInfo? authInfo))
                 {
+                    // Load the reply so the HTTP connection becomes available to send the authentication request.
+                    // Ideally we'd call LoadIntoBufferAsync, but it has no overload that accepts a CancellationToken so we call ReadAsByteArrayAsync instead.
+                    _ = await response.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
tmds: I think the difference is that here we're in a `DelegatingHandler` so we don't get the `HttpCompletionOption` behavior.

The connection becomes available when the response content gets read entirely.
It doesn't require the response object to be disposed.

I'm reading the content so the object can still be returned from the method.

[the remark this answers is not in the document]
baronfel: Ok, that fills in the gap for me. Thank you!

baronfel (approved):

baronfel:
Here's an Actions run using this payload: https://github.com/baronfel/sdk-container-demo/actions/runs/6382668229/job/17321761464

It's looking green as expected. Manually checking requests using a local HTTP inpector looks fine as well.

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
antonfirsov, line 329: Nit: is it really important to have the buffering cancellable? How much data is drained typically? Cancellation will close and drop the connection. Is this something you want here?

https://github.com/dotnet/runtime/blob/82fee2692b3954ba8903fa4764f1f4e36a26341a/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L473-L475

baronfel:
Approved over email - @tmds can you check out @antonfirsov's review comment above and respond before this is merged?

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
tmds: > Is it really important to have the buffering cancellable

It's important we drain the response while respecting the cancellation token.

> How much data is drained typically?

Not much. None to a few hundred of bytes at most.

> Cancellation will close and drop the connection. Is this something you want here?

Yes.

antonfirsov (approved):

src/Containers/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -324,6 +324,10 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 }
                 else if (response is { StatusCode: HttpStatusCode.Unauthorized } && TryParseAuthenticationInfo(response, out string? scheme, out AuthInfo? authInfo))
                 {
+                    // Load the reply so the HTTP connection becomes available to send the authentication request.
+                    // Ideally we'd call LoadIntoBufferAsync, but it has no overload that accepts a CancellationToken so we call ReadAsByteArrayAsync instead.
```
danmoseley, line 328: > LoadIntoBufferAsync, but it has no overload that accepts a CancellationToken

worth an API proposal?

PR #35760, opened 2023-09-27
Declare dependencies for two Extensions.Logging packages

mthalman:
There can be circumstances for source-build where the referenced version of the runtime by the sdk is a different version than is actually built. This can happen when the same commit has multiple versions associated with it.

This can cause the build of the sdk in the VMR to get an error similar to the following:

```
   /vmr/src/sdk/artifacts/source-build/self/src/src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj : error NU1102: Unable to find package Microsoft.Extensions.Logging with version (>= 8.0.0-rc.2.23476.7) [/vmr/src/sdk/artifacts/source-build/self/src/source-build.slnf]
    /vmr/src/sdk/artifacts/source-build/self/src/src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj : error NU1102:   - Found 4 version(s) in previously-source-built [ Nearest version: 8.0.0-rc.1.23419.4 ] [/vmr/src/sdk/artifacts/source-build/self/src/source-build.slnf]
    /vmr/src/sdk/artifacts/source-build/self/src/src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj : error NU1102:   - Found 3 version(s) in reference-packages [ Nearest version: 7.0.0 ] [/vmr/src/sdk/artifacts/source-build/self/src/source-build.slnf]
    /vmr/src/sdk/artifacts/source-build/self/src/src/Containers/Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj : error NU1102:   - Found 1 version(s) in source-built [ Nearest version: 8.0.0-rc.2.23475.17 ] [/vmr/src/sdk/artifacts/source-build/self/src/source-build.slnf]
```

This is because there is no declared dependency for `Microsoft.Extensions.Logging` and `Microsoft.Extensions.Logging.Abstractions`. Version.props defines those versions to be derived from the version of `Microsoft.Extensions.Logging.Console`. But those properties get defined before source-build's override of `MicrosoftExtensionsLoggingConsoleVersion`.

To fix this, I've declared those two packages as explicit dependencies and updated the associated properties to have the actual version values instead of deriving it. Also removed the `Package` portion of the property name which is not needed and keeps things consistent with the existing `MicrosoftExtensionsLoggingConsoleVersion` property.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -34,8 +34,8 @@
     <PackageVersion Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
     <PackageVersion Include="Microsoft.Extensions.FileProviders.Abstractions" Version="$(MicrosoftNETCoreAppRefPackageVersion)" />
     <PackageVersion Include="Microsoft.Extensions.FileSystemGlobbing" Version="$(MicrosoftExtensionsFileSystemGlobbingPackageVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingPackageVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingConsoleVersion)" />
     <PackageVersion Include="Microsoft.Extensions.ObjectPool" Version="$(MicrosoftNETCoreAppRefPackageVersion)"/>
     <PackageVersion Include="Microsoft.FSharp.Compiler" Version="$(MicrosoftFSharpCompilerPackageVersion)" />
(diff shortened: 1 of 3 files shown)


Directory.Packages.props
```diff
@@ -34,8 +34,8 @@
     <PackageVersion Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
     <PackageVersion Include="Microsoft.Extensions.FileProviders.Abstractions" Version="$(MicrosoftNETCoreAppRefPackageVersion)" />
     <PackageVersion Include="Microsoft.Extensions.FileSystemGlobbing" Version="$(MicrosoftExtensionsFileSystemGlobbingPackageVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingPackageVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsVersion)" />
```
baronfel, line 38: nit: naming wise, most of the other properties in this file are `XYZPackageVersion` - should we unify?

baronfel (approved):
Thanks! I'm pretty sure Containers introduced this and didn't do the correct thing here - I appreciate you getting us fixed up.

Directory.Packages.props
MichaelSimons: PackageVersion is the older of the two patterns.  There are some references without `Package`.  I have no objection to making the pattern consistent but IMO that should be done separately and in the main branch.

MichaelSimons (approved):

MichaelSimons:
@mthalman - friendly reminder to monitor the flow to main.  Mentioning this because most version.details.xml changes get stripped out from the flow PRs.

PR #35767, opened 2023-09-27
Detect correctly multitargeted apps for trim/AOT/single-file warnings

sbomer:
The warnings introduced by https://github.com/dotnet/sdk/pull/34077 and described in https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/trimming-unsupported-targetframework are unnecessary noise for projects that multitarget to include a TargetFramework that is compatible with trimming/AOT/single-file, and is a low enough version to ensure that it will be picked over any other TFM's assets when the library is consumed in a trimmed/AOT'd/single-file app.

This fixes the issue by detecting correctly multi-targeted libraries and suppressing the warnings in these cases.

Note that prior to the .NET 8 SDK, netstandard libraries would still get the `IsTrimmable` attribute embedded in the assembly when setting `IsTrimmable`, and could use trim analysis despite incomplete warnings due to unannotated ref assemblies. With the .NET 8 SDK this is no longer the case. Even for correctly multi-targeted libraries, the netstandard output assembly will no longer contain the `IsTrimmable` attribute. We debated whether it was worth warning in this case, but decided that the negative impact of the warning on library developers following the "golden path" was too large to justify keeping this as a warning.

Fixes https://github.com/dotnet/sdk/issues/35528


src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishASingleFileApp.cs
```diff
@@ -633,24 +633,33 @@ public void ILLink_analyzer_warnings_are_produced_using_EnableSingleFileAnalyzer
...
+        [InlineData("netstandard2.0;net6.0", false)] // Net6.0 is the min TFM supported for single-file and targeting.
+        [InlineData("netstandard2.0;net8.0", true)] // Net8.0 is supported for single-file, but leaves a "gap" for the supported net6./net7.0 TFMs.
+        public void EnableSingleFile_warns_when_expected_for_not_correctly_multitargeted_libraries(string targetFrameworks, bool shouldWarn)
         {
+            
             var testProject = new TestProject()
             {
                 Name = "ClassLibTest",
-                TargetFrameworks = targetFramework
+                TargetFrameworks = targetFrameworks
```
eerhardt, line 646: Do we have a test that explicitly sets `<TargetFramework>netstandard2.0</TargetFramework>`? Note that is using the singular TargetFramework and not the plural TargetFrameworks.

dsplaisted: The `TestProject` class being used here conflates the two, ie it only offers the plural `TargetFrameworks` property, and then if there's no semicolon in it, it will write the value to the `TargetFramework` property in the project instead.

dsplaisted:
Unfortunately, we want to avoid parsing other TargetFrameworks values for multi-targeted projects.  The reason is we want to support [TargetFramework aliases](https://github.com/NuGet/Home/issues/5154) (see also [here](https://github.com/NuGet/Home/pull/12124)), and this kind of check would not correctly handle the aliases.

You could disable the warning when the project is multi-targeted at all without trying to parse the other TargetFramework values.  I think we've done something similar in other situations.

vitek-karas:
Personally I'd be fine if this didn't work with the aliases. This is not a correctness feature, it's a convenience feature. So if it works in majority of scenarios, it's good enough. If there's a typical "golden path" way which works, I think that would be good enough for this.

dsplaisted:
> Interesting, thanks for the pointers. Still trying to understand this - doesn't the pattern of using `IsTargetFrameworkCompatible` with `$(TargetFramework)` already have the problem, since it (I assume) wouldn't handle `TargetFramework` values that are just aliases?

Yes, we shouldn't be using `IsTargetFrameworkCompatible` in generic code.  We should be using the `TargetFrameworkIdentifier` and `TargetFrameworkVersion` properties instead, for example: `'$(TargetFrameworkIdentifier)' == '.NETCoreApp' and $([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 5.0))`.  `IsTargetFrameworkCompatible` is OK to use in project files where you're in control of the list of TargetFrameworks and you know you're not using aliases.

> Personally I'd be fine if this didn't work with the aliases. This is not a correctness feature, it's a convenience feature. So if it works in majority of scenarios, it's good enough. If there's a typical "golden path" way which works, I think that would be good enough for this.

In projects where it doesn't work due to aliasing, wouldn't you always get the warnings?

dsplaisted:
> I believe this change doesn't make the behavior any worse for projects with aliases. For unsupported TFMs, you'd get the normal warning (the new logic won't detect aliases and so won't silence the warnings). If the TFMs are are all supported (as detected through `TargetFrameworkIdentifier`/`TargetFrameworkVersion`, there will be no warnings to silence in the first place.

OK, that makes sense.

vitek-karas (approved):
Looks good to me.

dsplaisted (approved):
I'm OK with this in general, but see my comment about the complexity of understanding it and consider whether it's worth simplifying.

Issue #35647, opened 2023-09-22
Container tooling doesn't understand the rtm channel


tmds:
The .NET 8 CI builds have moved to use `rtm` instead of `rc` in their version.

The code that determines the base image tag doesn't understand `rtm`:

https://github.com/dotnet/sdk/blob/bdfa7a19282caf2673dd1c59758543e6f286f006/src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs#L97-L119

This causes publishing container images to fail with CI built SDKs: 

```
error : CONTAINER2018: Invalid SDK prerelease version 'rtm' - only 'rc' and 'preview' are supported.
```

It's not possible to work around the failure by explicitly setting `ContainerBaseImage` because the base image tag is always calculated.

@baronfel the fix is to add `rtm`. I don't know what branch to target.

baronfel:
I'd expect that to need to target release/8.0.1xx - I believe that branch has moved to release branding.

baronfel:
Versioning label logic is [here](https://github.com/dotnet/sdk/blob/main/eng/Versions.props#L20C1-L22) - we may need to add both `rtm` and `servicing` so that we don't have unexpected failures in the future.

baronfel:
Closing this - we got this fixed for GA!

PR #35775, opened 2023-09-28
ComputeDotnetBaseImageTag: handle rtm and servicing channels.

tmds:
## Description

The SDK Version is used as part of computing what the proper tag for the MS Dotnet Docker images (dotnet/runtime-deps, dotnet/runtime, dotnet/aspnet) should be used for a given project. The SDK version is SemVer format, but has special meaning for the first 'prerelease segment', but our original algorithm didn't take into account the meaning of two stages of SDK versioning:

* `rtm`
* `servicing`

that aren't readily visible to external users.  This PR updates our tag calculates to recognize the special segments and treat them as stable versions for purposes of detecting the tag.

## Customer Impact

Users that test or create containers using rtm or servicing-versioned SDK versions would get an error about an unknown SDK version. There is a workaround - the user can eject entirely from 'base image inference' and explicitly set a base image. This defeats a lot of the value of the SDK Container Publish, though.

## Regression

**No** - this gap has been in the base image selection code since it was added, the reason we didn't see this for .NET 7 is because the container publish wasn't integrated into the SDK until 7.0.300 - well after SDK versioning was stablized.

## Risk

**Low** - this is a very simple change that includes automated tests.

Fixes https://github.com/dotnet/sdk/issues/35647

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
@@ -85,37 +85,25 @@ public sealed class ComputeDotnetBaseImageTag : Microsoft.Build.Utilities.Task
     }
 
      private string? DetermineLabelBasedOnChannel(int major, int minor, string[] releaseLabels) {
-      // this would be a switch, but we have to support net47x where Range and Index aren't available
-        if (releaseLabels.Length == 0)
+        var channel = releaseLabels.Length > 0 ? releaseLabels[0] : null;
+        switch (channel)
         {
-            return $"{major}.{minor}";
-        }
-        else
-        {
-            var channel = releaseLabels[0];
-            if (channel == "rc" || channel == "preview")
-            {
+            case null or "rtm" or "servicing":
+                return $"{major}.{minor}";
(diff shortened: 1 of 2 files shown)


src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
```diff
@@ -85,37 +85,25 @@ public override bool Execute()
...
+        switch (channel)
         {
-            return $"{major}.{minor}";
-        }
-        else
-        {
-            var channel = releaseLabels[0];
-            if (channel == "rc" || channel == "preview")
-            {
+            case null or "rtm" or "servicing":
```
baronfel, line 91: can you add cases for this to [CanComputeTagsForSupportedSDKVersions](https://github.com/dotnet/sdk/blob/79aaa542200480f0dc26d3ee7bd9d2b0364ae424/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/TargetsTests.cs#L170] please?

tmds: #done

baronfel (approved):

PR #35794, opened 2023-09-29
[release/8.0.1xx] Allow Worker SDK projects to invoke Publish Profile-derived targets

baronfel:
## Description

Fixes https://github.com/dotnet/sdk-container-builds/issues/274.

This change enables Worker SDK projects to participate in the Publish SDK, which includes all of the Publish Profiles. There are two parts to this change:

* The first part is simple - the Worker SDK Imports the props and targets from the Publish SDK, in a way similar to what the Web SDK does to enable the same behavior.
* The second is more complex - even with the Publish SDK included, container publishing doesn't work because of a gap in the target dependency order.  We bridge this gap in a defensive way.

## Customer Impact

Customers that create Worker SDK projects will be able to publish their applications as containers by default, with no changes to their project file.

## Regression

No - this is the first time we're enabling this behavior

## Risk

**Low** 

We are automating something that Visual Studio does already (implicitly adding the Publish SDK to a project), and we are limiting this to a project SDK type that is known to already work with the features after a PackageReference is added.

We have added an escape hatch MSBuild property so that if there is any unintended user impact the change can be reverted with a simple property addition.

We have added automated test coverage to ensure the behavior works and does not regress, and have performed manual test runs to verify the behavior as well.

## Technical Details

The primary target dependency chain here is 

```xml
<_DotNetCLIPostPublishDependsOn>
      _InitProjectCapabilityProperties;
      _AspNetCoreProjectSystemPostPublish;
      AfterPublish;
</_DotNetCLIPostPublishDependsOn>
```

Of most interest is the `_AspNetCoreProjectSystemPostPublish` target, which has a DependsOn chain of:

```xml
<_AspNetCoreProjectSystemPostPublishDependsOn>
      _InitPublishIntermediateOutputPath;
      $(_DotNetPublishTransformFiles);
      _PublishFiles;
</_AspNetCoreProjectSystemPostPublishDependsOn>
```

The `_PublishFiles` target is the one that PublishProfiles hook into. The problem is that `_AspNetCoreProjectSystemPostPublish` has a condition such that it only fires for ASP.NET projects. We need a way for non-ASP.NET projects to invoke similar targets.

To this end, we've added a new target to the `_DotNetCLIPostPublishDependsOn` dependency chain: `_OtherNetProjectPostPublish`. This target has the following dependency chain:

```xml
<_OtherNetProjectPostPublishDependsOn>
      _InitPublishIntermediateOutputPath;
      _PublishFiles;
</_OtherNetProjectPostPublishDependsOn>
```
The two targets added are the non-ASP.NET-specific publish targets. This new target is also conditioned (for now!) to only allow Worker SDK projects to use it. This limits the blast radius of this change so that Console apps don't accidentally get included.

### DIFF
--- a/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ContainerCli.cs
+++ b/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ContainerCli.cs
@@ -34,7 +34,10 @@ static class ContainerCli
     public static RunExeCommand LoadCommand(ITestOutputHelper log, params string[] args)
       => CreateCommand(log, "load", args);
 
-    private static RunExeCommand CreateCommand(ITestOutputHelper log, string command, string[] args)
+    public static RunExeCommand PortCommand(ITestOutputHelper log, string containerName, int port)
+      => CreateCommand(log, "port", containerName, port.ToString());
+
+    private static RunExeCommand CreateCommand(ITestOutputHelper log, string command, params string[] args)
     {
         string commandPath = IsPodman ? "podman" : "docker";
 
--- a/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
+++ b/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
@@ -272,7 +272,7 @@ public class EndToEndTests : IDisposable
         }
 
         string imageName = NewImageName();
-        string imageTag = $"1.0-{projectType}";
+        string imageTag = $"1.0-{projectType}-{addPackageReference}";
 
         // Build & publish the project
         CommandResult commandResult = new DotnetCommand(
@@ -304,7 +304,7 @@ public class EndToEndTests : IDisposable
             .Execute()
             .Should().Pass();
 
-        var containerName = "test-container-1";
+        var containerName = $"test-container-1-{projectType}-{addPackageReference}";
         CommandResult processResult = ContainerCli.RunCommand(
             _testOutput,
             "--rm",
--- a/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
+++ b/src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
@@ -209,12 +209,15 @@ public class EndToEndTests : IDisposable
         return publishDirectory;
     }
 
-    [DockerAvailableTheory(Skip = "https://github.com/dotnet/sdk/issues/33858")]
-    [InlineData(false)]
-    [InlineData(true)]
-    public async Task EndToEnd_NoAPI_Web(bool addPackageReference)
+
+    [DockerAvailableTheory()]
+    [InlineData("webapi", false)]
+    [InlineData("webapi", true)]
+    [InlineData("worker", false)]
+    [InlineData("worker", true)]
+    public async Task EndToEnd_NoAPI_ProjectType(string projectType, bool addPackageReference)
     {
-        DirectoryInfo newProjectDir = new DirectoryInfo(Path.Combine(TestSettings.TestArtifactsDirectory, $"CreateNewImageTest_{addPackageReference}"));
+        DirectoryInfo newProjectDir = new DirectoryInfo(Path.Combine(TestSettings.TestArtifactsDirectory, $"CreateNewImageTest_{projectType}_{addPackageReference}"));
         DirectoryInfo privateNuGetAssets = new DirectoryInfo(Path.Combine(TestSettings.TestArtifactsDirectory, "ContainerNuGet"));
 
         if (newProjectDir.Exists)
@@ -230,7 +233,7 @@ public class EndToEndTests : IDisposable
         newProjectDir.Create();
         privateNuGetAssets.Create();
 
-        new DotnetNewCommand(_testOutput, "webapi", "-f", ToolsetInfo.CurrentTargetFramework)
+        new DotnetNewCommand(_testOutput, projectType, "-f", ToolsetInfo.CurrentTargetFramework)
             .WithVirtualHive()
             .WithWorkingDirectory(newProjectDir.FullName)
             // do not pollute the primary/global NuGet package store with the private package(s)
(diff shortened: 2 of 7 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -14,7 +14,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup>
     <DefaultCopyToPublishDirectoryMetadata Condition="'$(DefaultCopyToPublishDirectoryMetadata)' == ''">true</DefaultCopyToPublishDirectoryMetadata>
     <_GetChildProjectCopyToPublishDirectoryItems Condition="'$(_GetChildProjectCopyToPublishDirectoryItems)' == ''">true</_GetChildProjectCopyToPublishDirectoryItems>
-    <IsPublishable Condition="'$(IsPublishable)' == ''">true</IsPublishable>
+    <IsPublishable Condition="'$(IsPublishable)' == '' and '$(OutputType)' != 'Library'">true</IsPublishable>
```
baronfel, line 17: Without this change, Library projects participate in Publish. previously this value was assumed to be true because different project types deliberately Imported the Publish SDK. This is no longer the case, so we have to be a bit more proscriptive.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/ContainerCli.cs
```diff
@@ -34,7 +34,10 @@ public static RunExeCommand InspectCommand(ITestOutputHelper log, params string[
     public static RunExeCommand LoadCommand(ITestOutputHelper log, params string[] args)
       => CreateCommand(log, "load", args);
 
-    private static RunExeCommand CreateCommand(ITestOutputHelper log, string command, string[] args)
+    public static RunExeCommand PortCommand(ITestOutputHelper log, string containerName, int port) =>
```
baronfel, line 37: The test that was failing was often failing due to colliding port numbers - I've fixed this by letting Docker assign host ports, then parsing those ports using this command. This makes us much less likely to fail.

src/Tests/Microsoft.NET.Build.Containers.IntegrationTests/EndToEndTests.cs
```diff
@@ -209,12 +209,17 @@ private string BuildLocalApp([CallerMemberName] string testName = "TestName", st
...
-    [InlineData(true)]
-    public async Task EndToEnd_NoAPI_Web(bool addPackageReference)
+
+    [DockerAvailableTheory()]
+    [InlineData("webapi", false)]
+    [InlineData("webapi", true)]
+    [InlineData("worker", false)]
+    [InlineData("worker", true)]
+    [InlineData("console", false)]
+    [InlineData("console", true)]
```
baronfel, line 219: These are our positive test cases for container publish. We should have negative test cases for test projects and library projects, but perhaps in the Publish SDK tests directly? @vijayrkn any good ideas where those should live?

src/WebSdk/Worker/Sdk/Sdk.props
```diff
@@ -23,6 +24,7 @@ Copyright (c) .NET Foundation. All rights reserved.
   </PropertyGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
+  <Import Condition="'$(WorkerSdkImportPublishSdk)' == 'true'" Sdk="Microsoft.NET.Sdk.Publish" Project="Sdk.props" />
```
vijayrkn, line 27: Today this is getting added in Visual Studio through the publish process. Once this change goes in, we should be able to remove that: 
`<Project Sdk="Microsoft.NET.Sdk.Worker;Microsoft.NET.Sdk.Publish">`

src/WebSdk/Publish/Targets/DotNetCLIToolTargets/Microsoft.NET.Sdk.DotNetCLITool.targets
```diff
@@ -84,6 +85,26 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+         ASP.NET-specific transforms stuff that's part of $(_DotNetPublishTransformFiles).
+         _InitPublishIntermediateOutputPath sets up the 'absolute' publish directory,
+         _PublishFiles is the extensibility point for Publish Profiles -->
+    <_OtherNetProjectPostPublishDependsOn>
+      _InitPublishIntermediateOutputPath;
+      _PublishFiles;
+    </_OtherNetProjectPostPublishDependsOn>
+  </PropertyGroup>
+
+  <!-- For safety, we're limiting this target to Worker SDK projects only. After 8.0.100 GA we should consider
```
vijayrkn, line 99: Thank you for adding these comments very helpful.

src/WebSdk/Worker/Sdk/Sdk.props
baronfel: Thank you - I completely agree. Think kind of specific, location feedback is exactly what I was hoping you'd bring, and as always you delivered :)

src/WebSdk/Worker/Sdk/Sdk.targets
```diff
@@ -12,6 +12,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
+  <Import Condition="'$(WorkerSdkImportPublishSdk)' == 'true'" Sdk="Microsoft.NET.Sdk.Publish" Project="Sdk.targets" />
```
vijayrkn, line 15: Same as the props. It would be best to move this right after the project system sdk import.

vijayrkn (changes requested):
Thank you for making this change.

src/WebSdk/Worker/Sdk/Sdk.targets
baronfel: done, thank you!

src/WebSdk/Worker/Sdk/Sdk.props
baronfel: done, thanks!

vijayrkn (approved):

nagilson (approved):
It looks good to me as well, though may you provide a bit more context onto what this does?
```
$(_DotNetPublishTransformFiles); 
```
Since its not part of the depends on chain for your similar target, but the ASP one seems to rely on this transformation? I'm guessing it is unnecessary, but since I am not familiar with this space, I figured it was worth checking.

baronfel:
Sure! `_DotNetPublishTransformFiles` is a list of targets that all relate to generating or transforming secondary artifacts for a given environment. These are almost all Web-project based (Web.config files for old asp.net, AppSettings, EF SQL generation) - these are heavily ASP.NET focused and would not have any meaning or impact for Worker or other project types. That's why it wasn't included in the set of targets to run as part of PostPublish for Worker projects. 

You can see them in the SDK repo at https://github.com/dotnet/sdk/blob/c53c8ac81e1bcb8118c31bc4f5622396c72e9959/src/WebSdk/Publish/Targets/TransformTargets/Microsoft.NET.Sdk.Publish.TransformFiles.targets.

nagilson:
I see, makes sense to me! Thank you :)

dsplaisted (approved):

MiYanni (approved):

baronfel:
Approved over Tactics email.

PR #35797, opened 2023-09-29
Disable IDE0200 "Lambda expression can be removed" warning

MiYanni:
## Summary
After a discussion in the [PR for cleaning up IDE0200](https://github.com/dotnet/sdk/pull/35678), the conclusion was to, instead, simply disable the warning. Now, VS will no longer show greyed out bits for lambda simplification.


nagilson (approved):
LGTM, thank you :)

PR #35805, opened 2023-10-02
SDK resolver perf: Don't check current process for dotnet.exe on .NET Framework

ladipro:
When evaluating projects in Visual Studio, more than half of the CPU spent in SDK resolution is in `EnvironmentProvider.GetCurrentProcessPath()`. Making this call in the .NET Framework version of the resolver does not make sense as there is no Framework `dotnet.exe`. This PRs reorders the checks so that `EnvironmentProvider.GetCurrentProcessPath()` does not run except for failure cases, which speeds up evaluation of the OrchardCore solution by ~18%.

The Framework `Process.MainModule` getter is slow. In this trace the total CPU spent in evaluation was 9459 ms.

image


ladipro:
> Is _getCurrentProcessPath() used only for checking whether the current process is dotnet.exe and returning its path in that case? If so, I imagine it could be skipped entirely on .NET Framework, as dotnet.exe won't be running that.

Indeed, I've come to the same realization. `Environment.ProcessPath` internally caches the path in a static field and we don't need this in the Framework build at all. A better fix is to just #ifdef it out on Framework.

ladipro:
I have updated the PR to not check for `dotnet.exe` on Framework. The perf impact is the same.

Forgind (approved):
LGTM!

PR #35807, opened 2023-10-02
SDK resolver perf: Optimize EnvironmentProvider.SearchPaths

ladipro:
The current code has multiple issues.
- It calls `Path.GetInvalidPathChars()` in a loop, which means unnecessary allocations.
- It uses an inherently quadratic `.Any(c => p.Contains(c)` instead of `IndexOfAny`.
- It caches the paths in `_searchPaths` but still performs some filtering every time `GetCommandPath` is called.

All this costs us 120 ms of CPU when evaluating the OrchardCore solution (~10% of SDK resolution and ~1% of evaluation overall).

image

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
+++ b/src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
@@ -33,17 +33,12 @@ namespace Microsoft.DotNet.NativeWrapper
         {
             get
             {
-                if (_searchPaths == null)
-                {
-                    var searchPaths =
-                        _getEnvironmentVariable(Constants.PATH)
-                        .Split(new char[] { Path.PathSeparator }, options: StringSplitOptions.RemoveEmptyEntries)
-                        .Select(p => p.Trim('"'))
-                        .Where(p => p.IndexOfAny(s_invalidPathChars) == -1)
-                        .ToList();
-
-                    _searchPaths = searchPaths;
-                }
+                _searchPaths ??=
+                    _getEnvironmentVariable(Constants.PATH)
+                    .Split(new char[] { Path.PathSeparator }, options: StringSplitOptions.RemoveEmptyEntries)
+                    .Select(p => p.Trim('"'))
+                    .Where(p => p.IndexOfAny(s_invalidPathChars) == -1)
+                    .ToList();
 
                 return _searchPaths;
             }
@@ -53,7 +48,8 @@ namespace Microsoft.DotNet.NativeWrapper
         {
             var commandNameWithExtension = commandName + Constants.ExeSuffix;
             var commandPath = SearchPaths
-                .FirstOrDefault(p => File.Exists(Path.Combine(p, commandNameWithExtension)));
+                .Select(p => Path.Combine(p, commandNameWithExtension))
+                .FirstOrDefault(File.Exists);
 
             return commandPath;
         }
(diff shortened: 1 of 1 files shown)


src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
```diff
@@ -33,12 +35,12 @@ private IEnumerable<string> SearchPaths
             {
                 if (_searchPaths == null)
                 {
-                    var searchPaths = new List<string>();
-
-                    searchPaths.AddRange(
+                    var searchPaths =
```
Forgind, line 38: nit:
Maybe replace this whole block with:
_searchPaths ??=
...

?

src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
```diff
@@ -51,9 +53,7 @@ public string GetCommandPath(string commandName)
         {
             var commandNameWithExtension = commandName + Constants.ExeSuffix;
             var commandPath = SearchPaths
-                .Where(p => !Path.GetInvalidPathChars().Any(c => p.Contains(c)))
-                .Select(p => Path.Combine(p, commandNameWithExtension))
-                .FirstOrDefault(File.Exists);
+                .FirstOrDefault(p => File.Exists(Path.Combine(p, commandNameWithExtension)));
```
Forgind, line 56: This now returns the search path without the commandNameWithExtension, whereas it previously included the commandNameWithExtension, right? Should we add one more Path.Combine wrapping this commandPath before returning?

ladipro: Great catch! Fixed, thank you.

src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
ladipro: Done!

Forgind (approved):
Thanks!

src/Resolvers/Microsoft.DotNet.NativeWrapper/EnvironmentProvider.cs
```diff
@@ -51,7 +48,6 @@ public string GetCommandPath(string commandName)
         {
             var commandNameWithExtension = commandName + Constants.ExeSuffix;
             var commandPath = SearchPaths
-                .Where(p => !Path.GetInvalidPathChars().Any(c => p.Contains(c)))
```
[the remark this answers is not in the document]
ladipro: Oh, so your concern is that `Path.Combine` may run unnecessarily on more than just the first item that satisfies the `File.Exists` filter? I don't believe that's true. This is lazy enumeration and the `Select` callback runs only on the minimum items needed to get the result. This sample should prove it: https://dotnetfiddle.net/nSXd1P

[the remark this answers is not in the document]
Forgind: Cool! The more you know 🙂

This looks good to me!

baronfel:
This is currently slated for .net 9 because that's what main is. If we want this sooner we may want to consider a backport to release/8.0.2xx so that folks can get this benefit starting in that release.

ladipro:
> This is currently slated for .net 9 because that's what main is. If we want this sooner we may want to consider a backport to release/8.0.2xx so that folks can get this benefit starting in that release.

#35805 is also impactful and in the same area, so maybe they can be backported together. I'll get it started when they're both merged. Thank you!

PR #35835, opened 2023-10-03
Including RuntimePackAsset in DesignerRuntimeImplementationProjectOutputGroup

Shyam-Gupta:
`RuntimePackAsset` paths are not available in SDK and they are required by WinForms .NET designer.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DesignerSupport.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DesignerSupport.targets
@@ -127,20 +127,19 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ItemGroup>
       <_DesignerShadowCopy Include="@(ReferenceCopyLocalPaths)" />
 
-      <!--
-        For .NET Core, we do not include NuGet package assets,
-        irrespective of self-contained / copy-local settings. Designer
-        will load these from the NuGet cache or shared framework always.
-      -->
+      <!-- For .NET Core, we do not include NuGet package assets, as the designer will load these from the NuGet cache. -->
       <_DesignerShadowCopy
         Remove="@(_ResolvedCopyLocalBuildAssets)"
         Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'"
         />
 
-      <!-- Include runtimepack assets -->
+      <!-- For .NET Core runtime pack assets, we do not include them regardless of whether the app is
+              self-contained, as they will be loaded from the shared framework.  However, for runtime pack
+              assets where RuntimePackAlwaysCopyLocal is true, there is no shared framework, so we still
+              include those.  -->
       <_DesignerShadowCopy
-        Include="@(RuntimePackAsset->WithMetadataValue('RuntimePackAlwaysCopyLocal', 'true'))"
-        Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'"
+        Remove="@(RuntimePackAsset)"
+        Condition="'%(RuntimePackAsset.RuntimePackAlwaysCopyLocal)' != 'true'"
         />
 
       <DesignerRuntimeImplementationProjectOutputGroupOutput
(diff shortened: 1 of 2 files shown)


dsplaisted (approved):
`RuntimePackAsset` items are generally in a Shared Framework and can be loaded by the designers without shadow copying.  The exception to this is the Microsoft.Windows.SDK.NET.Ref runtime pack.  Those items can be identified by checking the `RuntimePackAlwaysCopyLocal` metadata on them.  If that is set to true, then they should be included in the set of files to shadow copy.  If it's not set to true, then they should continue to be excluded.

It would be nice to have test coverage of this change.  There is one test that covers the shadow copying: `GivenThatWeWantDesignerSupport.It_provides_runtime_configuration_and_shadow_copy_files_via_outputgroup`.  That test could be expanded or a similar one added to cover the Windows SDK reference.

Also, this PR is currently targeting main, which will be .NET 9.  You probably want to target it to 8.0.2xx (or 8.0.1xx if you want to take the fix through servicing).

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DesignerSupport.targets
```diff
@@ -128,12 +128,18 @@ Copyright (c) .NET Foundation. All rights reserved.
...
       <!--
-        For .NET Core, we do not include NuGet package assets or runtime pack
-        assets, irrespective of self-contained / copy-local settings. Designer
+        For .NET Core, we do not include NuGet package assets,
+        irrespective of self-contained / copy-local settings. Designer
         will load these from the NuGet cache or shared framework always.
       -->
       <_DesignerShadowCopy
-        Remove="@(_ResolvedCopyLocalBuildAssets);@(RuntimePackAsset)"
+        Remove="@(_ResolvedCopyLocalBuildAssets)"
```
dsplaisted, line 136: At this point the `_DesignerShadowCopy` list includes all of the runtime pack assets and the files from NuGet packages, and we want to remove those (except the runtime packs in the case of RuntimePackAlwaysCopyLocal).  I think what you want is something like this (assuming I got the syntax right):

```xml
      <!-- For .NET Core, we do not include NuGet package assets, as the designer will load these from the NuGet cache. -->
      <_DesignerShadowCopy
        Remove="@(_ResolvedCopyLocalBuildAssets)"
        Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'"
        />

      <!-- For .NET Core runtime pack assets, we do not include them regardless of whether the app is
              self-contained, as they will be loaded from the shared framework.  However, for runtime pack
              assets where RuntimePackAlwaysCopyLocal is true, there is no shared framework, so we still
              include those.  -->
      <_DesignerShadowCopy
        Remove="@(RuntimePackAsset)"
        Condition="'%(RuntimePackAsset.RuntimePackAlwaysCopyLocal)' != 'true'"
        />

```

dsplaisted (commented):
Since we're modifying logic that interacts with self-contained, it might be good to update the test to have some cases where the app tested is self-contained.

src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DesignerSupport.targets
Shyam-Gupta: Thanks, I have updated this file as per your suggestion and confirmed that its working fine. I have also updated the test to run for `SelfContained` apps.
I remember testing `_DesignerShadowCopy` at that point earlier but found that it was not containing `RuntimePackAsset` paths. But I think I made some mistake in my testing.

Let me know if you have more feedback. Thanks.

src/Tests/Microsoft.NET.Build.Tests/GiventThatWeWantDesignerSupport.cs
```diff
@@ -14,12 +14,18 @@ public GivenThatWeWantDesignerSupport(ITestOutputHelper log) : base(log)
         }
 
         [Theory]
-        [InlineData("net46")]
-        [InlineData("netcoreapp3.0")]
-        [InlineData("net5.0-windows")]
-        public void It_provides_runtime_configuration_and_shadow_copy_files_via_outputgroup(string targetFramework)
+        [InlineData("net46", "true")]
+        [InlineData("net46", "false")]
```
dsplaisted, line 18: Self-contained is something that only applies to .NET Core, so we can remove a case here.

```suggestion
        [InlineData("net46", "false")]
```

dsplaisted (approved):
Looks good.  Feel free to merge when ready.

PR #35837, opened 2023-10-04
Treat old TFMs as unsupported by ILLink analyzers

sbomer:
Up to .NET 7, the analyzers versioned with the SDK (so you would always
get the latest analyzers). This allowed using the analyzers (and related
options which enable the analyzers) even on older TFMs.

https://github.com/dotnet/sdk/pull/32045 changed to versioning the
analyzers with the target framework, but still allowed use of the analyzers
on target frameworks where the .NET 7 ILLink pack was made available
for backwards compatibility.

This change removes such backwards compat by treating the analyzers
as unsupported for target frameworks before the framework was annotated
with the relevant attributes. This increases the scope of the warning added in
https://github.com/dotnet/sdk/pull/29441 and https://github.com/dotnet/sdk/pull/34077 to also warn for these unsupported
TFMs.

The first commit adds testcases to show the old behavior. The second commit
implements the new behavior and shows the changes to these testcases.

PublishTrimmed/PublishSingleFile/PublishAot used to enable the corresponding
analyzers. With this change, we now need to avoid failing
PublishTrimmed/PublishSingleFile for older TFMs that supported these publish
options, but don't support the analyzers. This change handles that by only
defaulting the `Enable*Analyzer` settings based on the publish settings for
TFMs where the analyzers are known to be supported.


src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets
```diff
@@ -22,6 +22,14 @@ Copyright (c) .NET Foundation. All rights reserved.
...
     <PublishTrimmed Condition="'$(PublishTrimmed)' == '' And '$(PublishAot)' == 'true'">true</PublishTrimmed>
     <IsTrimmable Condition="'$(IsTrimmable)' == '' and '$(IsAotCompatible)' == 'true'">true</IsTrimmable>
+
+    <_FirstTargetFrameworkToSupportTrimming>net6.0</_FirstTargetFrameworkToSupportTrimming>
+    <_FirstTargetFrameworkToSupportAot>net7.0</_FirstTargetFrameworkToSupportAot>
+    <_FirstTargetFrameworkToSupportSingleFile>net6.0</_FirstTargetFrameworkToSupportSingleFile>
+
+    <_FirstTargetFrameworkVersionToSupportTrimAnalyzer>$([MSBuild]::GetTargetFrameworkVersion('$(_FirstTargetFrameworkToSupportTrimming)'))</_FirstTargetFrameworkVersionToSupportTrimAnalyzer>
+    <_FirstTargetFrameworkVersionToSupportAotAnalyzer>$([MSBuild]::GetTargetFrameworkVersion('$(_FirstTargetFrameworkToSupportAot)'))</_FirstTargetFrameworkVersionToSupportAotAnalyzer>
+    <_FirstTargetFrameworkVersionToSupportSingleFileAnalyzer>$([MSBuild]::GetTargetFrameworkVersion('$(_FirstTargetFrameworkToSupportSingleFile)'))</_FirstTargetFrameworkVersionToSupportSingleFileAnalyzer>
```
dsplaisted, line 32: It doesn't look like the first set of properties are used anywhere else.  Can we specify the version numbers directly instead of specifying the target frameworks and then parsing out the versions?

```suggestion
    <_FirstTargetFrameworkVersionToSupportTrimAnalyzer>6.0</_FirstTargetFrameworkVersionToSupportTrimAnalyzer>
    <_FirstTargetFrameworkVersionToSupportAotAnalyzer>7.0</_FirstTargetFrameworkVersionToSupportAotAnalyzer>
    <_FirstTargetFrameworkVersionToSupportSingleFileAnalyzer>6.0</_FirstTargetFrameworkVersionToSupportSingleFileAnalyzer>
```

dsplaisted (approved):

vitek-karas (approved):

Issue #35455, opened 2023-09-15
Support for new Workload Set Versions


marcpopMSFT:
New list based on today's conversation that needs to be merged into the list further below:

- [x] publish workload sets from the workload set repo
- [x] figure out branching
- [x] `dotnet workload update --workload-state global` state saved into programdata on whether to use workload sets or not
- [x] dotnet workload install, restore, update should work with the state
  - [x] file based install
  - [x] admin based install
- [x] how does rollback work if workload-state is on set (should override)
- [x] explore how we store the workload install state (set versus not) in workload history
- [x] #39991 
- [x] #41102
- [x] #40548
- [x] Workload set garbage collection
- [x] #41420 
- [x] Document workload sets and new switches
- [x] Review the process for disabling workload sets or changing out of it
- ~[ ] Baseline workload set in installer~
- [ ] #41607
  - [ ] Workload set drop creation
  - [ ] workload set insertion
  - [ ] Do we want to create drops/insertions for workloads themselves
  - [ ] figure out automated codeflow
  - [ ] figure out auomated publishing (publish to VS eventually?)
- [x] #40297
- [x] #45936
- [x] Figure out the process of publishing the workload sets to nuget.org

This is the initial work in arcade and the resolver to support workload sets
https://github.com/dotnet/designs/pull/294

### Arcade
- [x] #34282

### Workload resolver (overlap with workload set version)
  - [x] sync on serialization effort within global.json code 
  - [ ] Error if trying to pin to a baseline workload set version (as those packs likely aren't available)
  - [ ] TBD do we resolve the latest manifests available or the latest with packs installed
  - [x] #34561
  - [x] #34562
  - [ ] ensure that resolver can get out of workload sets (i.e. detect Workload Install State and use that to determine loose manifests vs Workload Sets)
    - [ ] potential workaround: use a rollback file explicitly and specify that

### Workload CLI
  - [x] enable support for workload set versions for --info/list - https://github.com/dotnet/sdk/issues/33912  
  - [ ] Workload updates with workload sets
    - [ ] Update `TempDirectoryWorkloadManifestProvider` if needed
  - [ ] Figure out what to do with baseline manifests and workload set versions during garbage collection
  - [ ] Support for specifying a workload set version during install/uninstall commands
  - [ ] baseline workload set version in installer
  - [ ] file based install of workload set version
  - [ ] Enable support for workload set versions for repair
  - [ ] enable support for workload set versions for history
  - [ ] enable support for workload set versions for clean
[...]

PR #35857, opened 2023-10-04
Add workload garbage collection tests for side-by-side manifests

dsplaisted:


src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            ManifestRecord("testmanifest", "2.0.0", "6.0.300", "6.0.300").Should().NotExist();
+            ManifestRecord("testmanifest", "3.0.0", "6.0.300", "6.0.300").Should().Exist();
+
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.100", "testmanifest", "1.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "2.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "3.0.0", "WorkloadManifest.json")).Should().Exist();
+        }
+
+        [Fact]
+        public void GarbageCollectManifestsWithInstallState()
```
nagilson, line 173: This test is rather very similar to `GarbageCollectManifests`, it's very likely better to write it as a Theory with a boolean for `installState`.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+                Directory.CreateDirectory(Path.GetDirectoryName(pack.Path));
+                using var _ = File.Create(pack.Path);
+            }
+            else
+            {
+                Directory.CreateDirectory(pack.Path);
+            }
+        }
+
+        private void CreateInstallState(string featureBand, string installStateContents)
```
nagilson, line 318: Some of these functions have been duplicated a lot across tests, such as this one. There is a function `CreateMockInstallState` in `SdkDirectoryWorkloadManifestProviderTests` which is nearly identical, however, it does not use the featureBand parameter correctly. I do understand the simplicity of using these functions per test file, though I think it would be best for maintenance's sake to try to coalesce them into one `WorkloadTestUtilities.cs` file

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+                CreatePackInfo("Xamarin.Android.Runtime", "8.4.7.4", WorkloadPackKind.Library)
+            };
+            var sdkVersions = new [] { "6.0.100", "6.0.300" };
+
+            // Write packs
+            foreach (var sdkVersion in sdkVersions)
+            {
+                foreach (var pack in packsToKeep.Concat(packsToCollect))
+                {
+                    CreateInstalledPack(pack, sdkVersion);
```
nagilson, line 55: I see you have cleaned up the tests quite a bit, I quite like this `CreateInstalledPack` function to de-duplicate the logic. :smile:

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            ManifestRecord("testmanifest", "1.0.0", "6.0.100", "6.0.300").Should().NotExist();
+            ManifestRecord("testmanifest", "2.0.0", "6.0.300", "6.0.300").Should().Exist();
+            ManifestRecord("testmanifest", "3.0.0", "6.0.300", "6.0.300").Should().NotExist();
+
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.100", "testmanifest", "1.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "2.0.0", "WorkloadManifest.json")).Should().Exist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "3.0.0", "WorkloadManifest.json")).Should().NotExist();
+        }
+
+        //  Additional scenarios to add tests for once workload sets are added:
```
nagilson, line 239:
```suggestion
        //  TODO: Additional scenarios to add tests for once workload sets are added:
```

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.100", "testmanifest", "1.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "2.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "3.0.0", "WorkloadManifest.json")).Should().Exist();
+        }
+
+        [Fact]
+        public void GarbageCollectManifestsWithInstallState()
+        {
+            CreateMockManifest("testmanifest", "1.0.0", "6.0.100", sourceManifestName: @"Sample2.json");
+            CreateMockManifest("testmanifest", "2.0.0", "6.0.300", sourceManifestName: @"Sample2_v2.json");
```
nagilson, line 176:
```suggestion
            manifestToKeepVersion = "2.0.0"
            CreateMockManifest("testmanifest", manifestToKeepVersion, "6.0.300", sourceManifestName: @"Sample2_v2.json");
```

Although not important, I would prefer to have version numbers these as variables, so if we ever want to change the number, we only need to change it in once place. That being said, I can see how that might make it a bit harder to read, but it also makes it more clear what each item is testing.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+                PackRecord(pack, "6.0.300").Should().Exist();
+            }
+
+            ManifestRecord("testmanifest", "1.0.0", "6.0.100", "6.0.300").Should().NotExist();
+            ManifestRecord("testmanifest", "2.0.0", "6.0.300", "6.0.300").Should().Exist();
+            ManifestRecord("testmanifest", "3.0.0", "6.0.300", "6.0.300").Should().NotExist();
+
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.100", "testmanifest", "1.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "2.0.0", "WorkloadManifest.json")).Should().Exist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "3.0.0", "WorkloadManifest.json")).Should().NotExist();
```
nagilson, line 236: Should we also check that the sdk-manifests folder for 6.0.100 no longer exists?

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.100", "testmanifest", "1.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "2.0.0", "WorkloadManifest.json")).Should().NotExist();
+            new FileInfo(Path.Combine(_dotnetRoot, "sdk-manifests", "6.0.300", "testmanifest", "3.0.0", "WorkloadManifest.json")).Should().Exist();
+        }
+
+        [Fact]
+        public void GarbageCollectManifestsWithInstallState()
+        {
+            CreateMockManifest("testmanifest", "1.0.0", "6.0.100", sourceManifestName: @"Sample2.json");
```
nagilson, line 175:
```suggestion
            const string manifestWithNoMatchingSDKVersion = "1.0.0";
            const string manifestWithNoInstallRecordVersion = "3.0.0";
            CreateMockManifest("testmanifest", "1.0.0", "6.0.100", sourceManifestName: @"Sample2.json");
```

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
nagilson, line 239: Sorry for not being in the loop here, but the workload set garbage collection code has been merged, but is that not taking affect yet?  I don't see a work item to make these tests here, either: https://github.com/dotnet/sdk/issues/35455

nagilson (commented):
Overall, the test changes look good to me, thanks for keeping up on them :) 
I have a few nits that I've added below, however.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+        {
+            if (resolvedPackageId == null)
+            {
+                resolvedPackageId = id;
+            }
+
+            string path;
+            if (kind == WorkloadPackKind.Library)
+            {
+                path = Path.Combine(_dotnetRoot, "library-packs", $"resolvedPackageId.{version}.nupkg");
```
joeloff, line 280: Should this be `$"{resolvedPackageId}` instead (curly braces seems to be missing)

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
```diff
@@ -0,0 +1,393 @@
...
+            }
+            else
+            {
+                path = Path.Combine(_dotnetRoot, "packs", id, version);
+            }
+
+            return new PackInfo(new WorkloadPackId(id), version, kind, path, resolvedPackageId);
+        }
+
+        FileInfo PackRecord(PackInfo pack, string sdkFeatureBand)
```
joeloff, line 290: Should we parameterize the v1 and default it to `"v1"` in case we ever need v2?

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: That's what I started out trying to do when writing this test, but I realized that the `packsToKeep` and `packsToCollect` would have to differ, as well as the assertions for which manifests and manifest records should exist.  With so many conditionals in the Theory, it seemed like it might be better to write it as a separate test, so I ended up going with that.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: Yes, I copied this from `SdkDirectoryWorkloadManifestProviderTests`, and noticed the bug so I fixed it in this PR.

I think most of the helper methods use private fields of the class so that they're less verbose to call.  So it's not so simple to factor them out into a shared file, I think you'd need a base class for workload tests.  I'd be happy to see someone try that, but I didn't want to tackle that here.

Overall, I don't like duplicated code, but I think I'm a lot more OK with it in tests.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: The garbage collection for side-by-side manifests has been merged.  Garbage collection for workload sets themselves is only partially implemented, and we don't yet create or install workload sets.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: Good catch, thanks.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: I don't think so.  The `v1` in the path is a future-proofing mechanism that we may never need.  If we do need it, then we would probably be changing more about how the records are stored, so the test code would have to update anyway with more than just a change from `v1` to `v2`.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: I tried this out but we don't actually walk up the sdk-manifests tree and delete empty folders like we do with the installation records.  Normally I would probably fix that but at this point I want to keep the PR restricted to test changes.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
dsplaisted: There are so many different versions related in different ways that I don't think the test would be likely to still work if we were to change one of them.  I think it would also end up hurting the readability to have so many different variable names.

So instead, to improve the readability of the new tests, I've added a bunch of comments to them and rearranged them a bit.

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
nagilson: I made a remark about this here: https://github.com/dotnet/sdk/issues/35455. It would definitely be easy to write an unsuspecting bug that expects the folders to be deleted, so eventually it should be fixed, though I agree for now it's OK

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
nagilson: Ah I see, the comments help a lot. I actually misunderstood the test a bit. Thank you!

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
nagilson: That's fair!

src/Tests/dotnet-workload-install.Tests/WorkloadGarbageCollectionTests.cs
nagilson: Ah, I did not even notice that you had fixed `CreateMockInstallState`. I would also be happy to have some of this compressed down together, though it doesn't have to be in this PR. I will give it a 👍, though I would be more pleased once these get cleaned up in the future ;)

nagilson (approved):
LGTM now, thanks for refactoring a bit :) Ideally, some of the utility-function duplication gets removed down the road in another PR. Anywho, I am happy to have these new tests to improve our code coverage for garbage collection!

MiYanni (approved):

PR #35859, opened 2023-10-04
Workload manifest updater cleanup

MiYanni:
Related: https://github.com/dotnet/sdk/issues/22882

## Summary
I was planning on working on the linked issue above, but ended up going down a rabbit hole. A majority of the changes take place in `WorkloadManifestUpdater.cs`. A lot of the cleanup is around the uses of many different ValueTuple types. I created records or aliases to attempt to simplify the situation. No functional changes should be taking place in this PR.

### Notable Changes
- Created the `ManifestUpdateWithWorkloads` and `ManifestVersionWithBand` records to help with some of the tuple situation.
- Removed `_tempDirPath` as it was unused. The value is currently part the `WorkloadInstallerFactory.GetWorkloadInstaller()`, so the provided `IWorkloadManifestInstaller` will contain and use this temp directory.
  - Similarly, `_tempDirManifestPath` was not used in the mocked version for testing, so this was removed too.
- Created a `WorkloadCollection` alias since a dictionary of `WorkloadId` to `WorkloadDefinition` is used in multiple places in the code.
- Used deconstruction in many places to simplify readability.
- Refactored `GetManifestPackageDownloadsAsync` into a loop since it was a series of if-statements that did the same logic 3 times in a row.
- Ad-hoc formatting cleanup and some style simplification.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Extensions.EnvironmentAbstractions;
 using Microsoft.NET.Sdk.WorkloadManifestReader;
+using WorkloadCollection = System.Collections.Generic.Dictionary<Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadId, Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadDefinition>;
 
 namespace Microsoft.DotNet.Workloads.Workload.Install
 {
@@ -12,13 +13,9 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
         Task BackgroundUpdateAdvertisingManifestsWhenRequiredAsync();
 
-        IEnumerable<(
-            ManifestVersionUpdate manifestUpdate,
-            Dictionary<WorkloadId, WorkloadDefinition> Workloads
-            )> CalculateManifestUpdates();
+        IEnumerable<ManifestUpdateWithWorkloads> CalculateManifestUpdates();
 
-        IEnumerable<ManifestVersionUpdate>
-            CalculateManifestRollbacks(string rollbackDefinitionFilePath);
+        IEnumerable<ManifestVersionUpdate> CalculateManifestRollbacks(string rollbackDefinitionFilePath);
 
         Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsync(bool includePreviews, SdkFeatureBand providedSdkFeatureBand, SdkFeatureBand installedSdkFeatureBand);
 
@@ -26,4 +23,6 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
         void DeleteUpdatableWorkloadsFile();
     }
+
+    internal record ManifestUpdateWithWorkloads(ManifestVersionUpdate ManifestUpdate, WorkloadCollection Workloads);
 }
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -433,50 +433,27 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 
         private static string GetAdvertisingWorkloadsFilePath(string userProfileDir, SdkFeatureBand featureBand) => Path.Combine(userProfileDir, $".workloadAdvertisingUpdates{featureBand}");
 
-        private async Task<String> GetOnlinePackagePath(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, bool includePreviews)
-        {
-            string packagePath = await _nugetPackageDownloader.DownloadPackageAsync(
-                _workloadManifestInstaller.GetManifestPackageId(manifestId, sdkFeatureBand),
-                packageSourceLocation: _packageSourceLocation,
-                includePreview: includePreviews);
-
-            return packagePath;
-        }
-
-        private string GetOfflinePackagePath(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, DirectoryPath? offlineCache = null)
-        {
-            string packagePath = Directory.GetFiles(offlineCache.Value.Value)
-                .Where(path =>
-                {
-                    if (!path.EndsWith(".nupkg"))
-                    {
-                        return false;
-                    }
-                    var manifestPackageId = _workloadManifestInstaller.GetManifestPackageId(manifestId, sdkFeatureBand).ToString();
-                    return Path.GetFileName(path).StartsWith(manifestPackageId, StringComparison.OrdinalIgnoreCase);
-                })
-                .Max();
-
-            return packagePath;
-        }
-
         private async Task<(bool Success, string PackagePath)> GetManifestPackageUpdate(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, bool includePreviews, DirectoryPath? offlineCache = null)
         {
-            if (offlineCache == null || !offlineCache.HasValue)
+            var manifestPackageId = _workloadManifestInstaller.GetManifestPackageId(manifestId, sdkFeatureBand);
+            if (offlineCache != null)
             {
-                try
-                {
-                    string onlinePath = await GetOnlinePackagePath(sdkFeatureBand, manifestId, includePreviews);
-                    return (Success: true, PackagePath: onlinePath);
-                }
-                catch (NuGetPackageNotFoundException)
-                {
-                    return (Success: false, PackagePath: null);
-                }
+                string offlinePackagePath = Directory.GetFiles(offlineCache.Value.Value)
+                    .Where(path => path.EndsWith(".nupkg") && Path.GetFileName(path).StartsWith(manifestPackageId.ToString(), StringComparison.OrdinalIgnoreCase))
+                    .Max();
+                return (Success: offlinePackagePath != null, PackagePath: offlinePackagePath);
             }
 
-            string offlinePath = GetOfflinePackagePath(sdkFeatureBand, manifestId, offlineCache);
-            return (Success: offlinePath != null, PackagePath: offlinePath);
+            try
+            {
+                string onlinePackagePath = await _nugetPackageDownloader.DownloadPackageAsync(
+                    manifestPackageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreviews);
+                return (Success: true, PackagePath: onlinePackagePath);
+            }
+            catch (NuGetPackageNotFoundException)
+            {
+                return (Success: false, PackagePath: null);
+            }
         }
 
         private string GetAdvertisingManifestPath(SdkFeatureBand featureBand, ManifestId manifestId) => Path.Combine(_userProfileDir, "sdk-advertising", featureBand.ToString(), manifestId.ToString());
--- a/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Text.Json;
 using Microsoft.DotNet.Cli;
 using Microsoft.DotNet.Cli.NuGetPackageDownloader;
@@ -12,6 +13,7 @@ using Microsoft.Extensions.EnvironmentAbstractions;
 using Microsoft.NET.Sdk.WorkloadManifestReader;
 using NuGet.Common;
 using NuGet.Versioning;
+using WorkloadCollection = System.Collections.Generic.Dictionary<Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadId, Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadDefinition>;
 
 namespace Microsoft.DotNet.Workloads.Workload.Install
 {
@@ -22,7 +24,6 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
         private readonly INuGetPackageDownloader _nugetPackageDownloader;
         private readonly SdkFeatureBand _sdkFeatureBand;
         private readonly string _userProfileDir;
-        private readonly string _tempDirPath;
         private readonly PackageSourceLocation _packageSourceLocation;
         Func<string, string> _getEnvironmentVariable;
         private readonly IWorkloadInstallationRecordRepository _workloadRecordRepo;
@@ -33,7 +34,6 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             IWorkloadResolver workloadResolver,
             INuGetPackageDownloader nugetPackageDownloader,
             string userProfileDir,
-            string tempDirPath,
             IWorkloadInstallationRecordRepository workloadRecordRepo,
             IWorkloadManifestInstaller workloadManifestInstaller,
             PackageSourceLocation packageSourceLocation = null,
@@ -44,7 +44,6 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             _reporter = reporter;
             _workloadResolver = workloadResolver;
             _userProfileDir = userProfileDir;
-            _tempDirPath = tempDirPath;
             _nugetPackageDownloader = nugetPackageDownloader;
             _sdkFeatureBand = sdkFeatureBand ?? new SdkFeatureBand(_workloadResolver.GetSdkFeatureBand());
             _packageSourceLocation = packageSourceLocation;
@@ -72,15 +71,14 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
                 workloadResolver, VerbosityOptions.normal, userProfileDir, verifySignatures: false);
             var workloadRecordRepo = installer.GetWorkloadInstallationRecordRepository();
 
-            return new WorkloadManifestUpdater(reporter, workloadResolver, nugetPackageDownloader, userProfileDir, tempPackagesDir.Value, workloadRecordRepo, installer);
+            return new WorkloadManifestUpdater(reporter, workloadResolver, nugetPackageDownloader, userProfileDir, workloadRecordRepo, installer);
         }
 
         public async Task UpdateAdvertisingManifestsAsync(bool includePreviews, DirectoryPath? offlineCache = null)
         {
             // this updates all the manifests 
             var manifests = _workloadResolver.GetInstalledManifests();
-            await Task.WhenAll(manifests.Select(manifest => UpdateAdvertisingManifestAsync(manifest, includePreviews, offlineCache)))
-                .ConfigureAwait(false);
+            await Task.WhenAll(manifests.Select(manifest => UpdateAdvertisingManifestAsync(manifest, includePreviews, offlineCache))).ConfigureAwait(false);
             WriteUpdatableWorkloadsFile();
         }
 
@@ -161,37 +159,26 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             }
         }
 
-        public IEnumerable<(
-            ManifestVersionUpdate manifestUpdate,
-            Dictionary<WorkloadId, WorkloadDefinition> Workloads
-            )>
-            CalculateManifestUpdates()
+        public IEnumerable<ManifestUpdateWithWorkloads> CalculateManifestUpdates()
         {
-            var manifestUpdates =
-                new List<(ManifestVersionUpdate manifestUpdate,
-                    Dictionary<WorkloadId, WorkloadDefinition> Workloads)>();
             var currentManifestIds = GetInstalledManifestIds();
             foreach (var manifestId in currentManifestIds)
             {
-                var currentManifestVersion = GetInstalledManifestVersion(manifestId);
-                var advertisingManifestVersionAndWorkloads = GetAdvertisingManifestVersionAndWorkloads(manifestId);
-                if (advertisingManifestVersionAndWorkloads == null)
+                var (installedVersion, installedBand) = GetInstalledManifestVersion(manifestId);
+                var advertisingInfo = GetAdvertisingManifestVersionAndWorkloads(manifestId);
+                if (advertisingInfo == null)
                 {
                     continue;
                 }
+                var ((adVersion, adBand), adWorkloads) = advertisingInfo.Value;
 
-                if (advertisingManifestVersionAndWorkloads != null &&
-                    ((advertisingManifestVersionAndWorkloads.Value.ManifestVersion.CompareTo(currentManifestVersion.manifestVersion) > 0
-                        && advertisingManifestVersionAndWorkloads.Value.ManifestFeatureBand.Equals(currentManifestVersion.sdkFeatureBand)) ||
-                    advertisingManifestVersionAndWorkloads.Value.ManifestFeatureBand.CompareTo(currentManifestVersion.sdkFeatureBand) > 0))
+                if ((adVersion.CompareTo(installedVersion) > 0 && adBand.Equals(installedBand)) ||
+                    adBand.CompareTo(installedBand) > 0)
                 {
-                    manifestUpdates.Add((new ManifestVersionUpdate(manifestId, currentManifestVersion.manifestVersion, currentManifestVersion.sdkFeatureBand.ToString(),
-                        advertisingManifestVersionAndWorkloads.Value.ManifestVersion, advertisingManifestVersionAndWorkloads.Value.ManifestFeatureBand.ToString()),
-                        advertisingManifestVersionAndWorkloads.Value.Workloads));
+                    var update = new ManifestVersionUpdate(manifestId, installedVersion, installedBand.ToString(), adVersion, adBand.ToString());
+                    yield return new(update, adWorkloads);
                 }
             }
-
-            return manifestUpdates;
         }
 
         public IEnumerable<WorkloadId> GetUpdatableWorkloadsToAdvertise(IEnumerable<WorkloadId> installedWorkloads)
@@ -213,20 +200,19 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             var currentManifestIds = GetInstalledManifestIds();
             var manifestRollbacks = ParseRollbackDefinitionFile(rollbackDefinitionFilePath);
 
-            var unrecognizedManifestIds = manifestRollbacks.Where(rollbackManifest => !currentManifestIds.Contains(rollbackManifest.Item1));
+            var unrecognizedManifestIds = manifestRollbacks.Where(rollbackManifest => !currentManifestIds.Contains(rollbackManifest.Id));
             if (unrecognizedManifestIds.Any())
             {
                 _reporter.WriteLine(string.Format(LocalizableStrings.RollbackDefinitionContainsExtraneousManifestIds, rollbackDefinitionFilePath, string.Join(" ", unrecognizedManifestIds)).Yellow());
-                manifestRollbacks = manifestRollbacks.Where(rollbackManifest => currentManifestIds.Contains(rollbackManifest.Item1));
+                manifestRollbacks = manifestRollbacks.Where(rollbackManifest => currentManifestIds.Contains(rollbackManifest.Id));
             }
 
-            var manifestUpdates = manifestRollbacks
-                .Select(manifest =>
-                {
-                    var installedManifestInfo = GetInstalledManifestVersion(manifest.id);
-                    return new ManifestVersionUpdate(manifest.id, installedManifestInfo.manifestVersion, installedManifestInfo.sdkFeatureBand.ToString(),
-                        manifest.version, manifest.featureBand.ToString());
-                });
+            var manifestUpdates = manifestRollbacks.Select(manifest =>
+            {
+                var (id, (version, band)) = manifest;
+                var (installedVersion, installedBand) = GetInstalledManifestVersion(id);
+                return new ManifestVersionUpdate(id, installedVersion, installedBand.ToString(), version, band.ToString());
+            });
 
             return manifestUpdates;
         }
@@ -285,10 +271,7 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             return downloads;
         }
 
-        private IEnumerable<ManifestId> GetInstalledManifestIds()
-        {
-            return _workloadResolver.GetInstalledManifests().Select(manifest => new ManifestId(manifest.Id));
-        }
+        private IEnumerable<ManifestId> GetInstalledManifestIds() => _workloadResolver.GetInstalledManifests().Select(manifest => new ManifestId(manifest.Id));
 
         private async Task UpdateAdvertisingManifestAsync(WorkloadManifestInfo manifest, bool includePreviews, DirectoryPath? offlineCache = null)
         {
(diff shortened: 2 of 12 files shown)


src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -105,29 +106,23 @@ public override int Execute()
...
-            List<UpdateAvailableEntry> updateList = new();
-            foreach ((ManifestVersionUpdate manifestUpdate, Dictionary<WorkloadId, WorkloadDefinition> workloads) in manifestsToUpdate)
+            foreach ((ManifestVersionUpdate manifestUpdate, WorkloadCollection workloads) in manifestsToUpdate)
             {
-                foreach ((WorkloadId WorkloadId, WorkloadDefinition workloadDefinition) in
-                    workloads)
+                foreach ((WorkloadId workloadId, WorkloadDefinition workloadDefinition) in workloads)
                 {
-                    if (installedWorkloads.Contains(new WorkloadId(WorkloadId.ToString())))
+                    if (installedList.Contains(workloadId))
```
MiYanni, line 118: This *could* be a functional change but I couldn't figure out why it was creating a new `WorkloadId` from a `WorkloadId`. Looked suspicious...

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -161,37 +158,26 @@ public static void AdvertiseWorkloadUpdates()
...
-                var currentManifestVersion = GetInstalledManifestVersion(manifestId);
-                var advertisingManifestVersionAndWorkloads = GetAdvertisingManifestVersionAndWorkloads(manifestId);
-                if (advertisingManifestVersionAndWorkloads == null)
+                var (installedVersion, installedBand) = GetInstalledManifestVersion(manifestId);
+                var advertisingInfo = GetAdvertisingManifestVersionAndWorkloads(manifestId);
+                if (advertisingInfo == null)
                 {
                     continue;
                 }
+                var ((adVersion, adBand), adWorkloads) = advertisingInfo.Value;
```
MiYanni, line 172: Great example of how deconstruction can make the code below much more readable.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -213,68 +199,57 @@ public IEnumerable<ManifestVersionUpdate> CalculateManifestRollbacks(string roll
...
+                var (id, (version, band)) = manifest;
+                var (installedVersion, installedBand) = GetInstalledManifestVersion(id);
+                return new ManifestVersionUpdate(id, installedVersion, installedBand.ToString(), version, band.ToString());
+            });
 
             return manifestUpdates;
         }
 
-
         public async Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsync(bool includePreviews, SdkFeatureBand providedSdkFeatureBand, SdkFeatureBand installedSdkFeatureBand)
```
MiYanni, line 219: This method should be checked side-by-side. But the intent is not to cause any functional changes that would affect the operation of the method.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -483,9 +449,12 @@ private async Task<string> GetOnlinePackagePath(SdkFeatureBand sdkFeatureBand, M
         private string GetOfflinePackagePath(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, DirectoryPath? offlineCache = null)
         {
             string packagePath = Directory.GetFiles(offlineCache.Value.Value)
-                .Where(path => path.EndsWith(".nupkg"))
```
MiYanni, line 486: We shouldn't chain `Where` statements. I collapsed the condition into the other `Where`.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -494,43 +463,27 @@ private string GetOfflinePackagePath(SdkFeatureBand sdkFeatureBand, ManifestId m
             return packagePath;
         }
 
-        private async Task<(bool, string)> GetManifestPackageUpdate(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, bool includePreviews, DirectoryPath? offlineCache = null)
+        private async Task<(bool Success, string PackagePath)> GetManifestPackageUpdate(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, bool includePreviews, DirectoryPath? offlineCache = null)
```
MiYanni, line 466: Always name the elements in a `ValueTuple`. Otherwise, no one knows what is their purpose without digging through the references.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -213,68 +199,54 @@ public IEnumerable<ManifestVersionUpdate> CalculateManifestRollbacks(string roll
...
-                    {
-                        downloads.Add(new WorkloadDownload(manifest.Id, packageId.ToString(), latestVersion.ToString()));
-                    }
-
-                    if (!success && !installedSdkFeatureBand.Equals(providedSdkFeatureBand))
+                    PackageId? providedPackageId = null;
+                    var fallbackFeatureBand = new SdkFeatureBand(manifest.ManifestFeatureBand);
+                    SdkFeatureBand[] bands = [providedSdkFeatureBand, installedSdkFeatureBand, fallbackFeatureBand];
+                    var success = false;
+                    foreach (var band in bands.Distinct())
```
MiYanni, line 230: Instead of checking if the next band equals the previous one, we just use `Distinct()` so that we're only processing actual unique bands.

dsplaisted:
I haven't looked at the code changes yet, but this should probably target 8.0.2xx instead of main, since we're going to continue working on workloads in 8.0.2xx, and if we put a refactoring like this in main it will be harder to merge later.

src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
```diff
@@ -3,6 +3,7 @@
 
 using Microsoft.Extensions.EnvironmentAbstractions;
 using Microsoft.NET.Sdk.WorkloadManifestReader;
+using WorkloadCollection = System.Collections.Generic.Dictionary<Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadId, Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadDefinition>;
```
nagilson, line 6:
```suggestion
using WorkloadCollection = Dictionary<WorkloadId, WorkloadDefinition>;
```

nagilson (commented):
Some suggestions for this change, overall I am quite happy with these improvements, so thank you :)

src/Tests/dotnet-workload-list.Tests/GivenWorkloadInstallerAndWorkloadsInstalled.cs
```diff
@@ -56,28 +58,28 @@ private void Setup(string identifier)
                             new WorkloadId("other"), false, "other description",
                             WorkloadDefinitionKind.Dev, null, null, null)
                     }),
-                (
+                new(
```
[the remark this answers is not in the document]
MiYanni: `new` calls the constructor for the record. These were previously tuples which are only instantiated with parentheses. The properties/constructor of the record are in the same order as the tuple, so only `new` is needed. This feature of omitting the type from the `new` expression was added in C# 9: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features

src/Tests/dotnet-workload-list.Tests/GivenWorkloadInstallerAndWorkloadsInstalled.cs
```diff
@@ -56,28 +58,28 @@ private void Setup(string identifier)
...
-                    new Dictionary<WorkloadId, WorkloadDefinition>
+                    new WorkloadCollection
                     {
                         [new WorkloadId("other-manifest-workload")] = new(
                             new WorkloadId("other-manifest-workload"), false,
                             "other-manifest-workload description",
                             WorkloadDefinitionKind.Dev, null, null, null)
                     }),
-                (
+                new(
```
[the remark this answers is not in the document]
MiYanni: Same answer as https://github.com/dotnet/sdk/pull/35859#discussion_r1347941689

src/Tests/dotnet-workload-list.Tests/GivenWorkloadInstallerAndWorkloadsInstalled.cs
```diff
@@ -40,14 +42,14 @@ private void Setup(string identifier)
 
             _mockManifestUpdates = new()
             {
-                (
+                new(
```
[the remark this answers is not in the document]
MiYanni: Same answer as https://github.com/dotnet/sdk/pull/35859#discussion_r1347941689

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.NET.Sdk.WorkloadManifestReader;
 using Microsoft.TemplateEngine.Cli.Commands;
 using InformationStrings = Microsoft.DotNet.Workloads.Workload.LocalizableStrings;
+using WorkloadCollection = System.Collections.Generic.Dictionary<Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadId, Microsoft.NET.Sdk.WorkloadManifestReader.WorkloadDefinition>;
```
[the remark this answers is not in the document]
MiYanni: It isn't a type "definition", unfortunately. It is just an alias. And aliases are file-scoped only. There's a bunch of different ways to fix this, like creating a new type that has a dictionary as a property, or making a type that derives of dictionary, or just implements the right interfaces. Using an alias doesn't incur additional allocations of intermediary types. I'd likely restructure things if I was going to do a bigger change than that, but this was only for readability's sake.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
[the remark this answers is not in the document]
MiYanni: I'd be suspicious of that. The information already exists as a list and there is processing required to create a HashSet from that. Also, a lot of those perf measurements, especially for `Contains`, relies on the number of elements. In a single feature band, there aren't even 100's of workloads that exist. The set is quite small. I highly doubt there is a practical perf change here. I'm more concerned with why it was creating a `WorkloadId` out of a `WorkloadId`... I could dig into the history to try to find that out.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: I could add a comment (or multiple comments) that explains what it does, if you think that'll aid in readability. And this code doesn't result in a perf gain. It just takes 3 duplicate blocks of code and puts them into a loop that runs (potentially) 3 times.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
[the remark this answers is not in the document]
nagilson: I see, I had no idea c# had nothing like a 'typedef' in C++, that's unfortunate. This change I do quite like, then :D

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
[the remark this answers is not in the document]
nagilson: Hm, that makes sense. But yeah the workloadId thing, no clue... the constructor doesn't appear to do anything. May very well be a silly mistake!

MiYanni:
@dsplaisted 
> but this should probably target 8.0.2xx instead of main

🤔 The situation is:

1. I did [this change](https://github.com/dotnet/sdk/pull/35640) previously which is breaking change to the workload commands that output JSON. This was done in `main`.
2. This current PR is a precursor for [this issue](https://github.com/dotnet/sdk/issues/22882) which is a follow-up for the PR I just mentioned. So, the changes *for that issue* should definitely be `main` as it is modifying the JSON output (even though the proposed changes are additive).

So, is it a problem if we merge this into 8.0.2xx and then (using /backport) also put it into main? Or should I just wait until the codeflow happens for 8.0.2xx -> main? I'd like this in main *soon* to do my next set of changes.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -66,9 +67,9 @@ public override int Execute()
             {
                 _workloadListHelper.CheckTargetSdkVersionIsValid();
 
-                UpdateAvailableEntry[] updateAvailable = GetUpdateAvailable(installedList);
-                ListOutput listOutput = new(installedList.Select(id => id.ToString()).ToArray(),
-                    updateAvailable);
+                var updateAvailable = GetUpdateAvailable(installedList);
+                var installed = installedList.Select(id => id.ToString()).ToArray();
```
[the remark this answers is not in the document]
MiYanni: So, it was named after what the property name is in the record.
image

So, at the time I wrote that, I wasn't sure myself either what it is within the code. But what it is is... literally `installed`:
image

But what it actually is is:
image

So it is the collection of installed workload ids. I'll be changing this code in the PR that is actually implementing the issue enhancements, which is not this PR. This spot in the code is what spawned doing the rest of the cleanup, but it itself isn't cleaned up since I'll be changing it anyway.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -161,37 +158,26 @@ public static void AdvertiseWorkloadUpdates()
...
-            var manifestUpdates =
-                new List<(ManifestVersionUpdate manifestUpdate,
-                    Dictionary<WorkloadId, WorkloadDefinition> Workloads)>();
             var currentManifestIds = GetInstalledManifestIds();
             foreach (var manifestId in currentManifestIds)
             {
-                var currentManifestVersion = GetInstalledManifestVersion(manifestId);
-                var advertisingManifestVersionAndWorkloads = GetAdvertisingManifestVersionAndWorkloads(manifestId);
-                if (advertisingManifestVersionAndWorkloads == null)
+                var (installedVersion, installedBand) = GetInstalledManifestVersion(manifestId);
```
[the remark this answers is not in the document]
MiYanni: Ah, good find. I did this change like a few hours before I did the other deconstruction below. So, I didn't relate the new variable names to each other. Let me do that.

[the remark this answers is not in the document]
MiYanni: Ah wait. I ended up looking at the method holistically. 

image

The purpose of the method is to compare manifest versions (that's what `CalculateManifestUpdates` seems to mean). You can see the name matching I did between the underlined variables. So, I think keeping the name as since makes sense since they're all operating on manifests. Plus, if you hover over `installedVersion`, it says the type is `ManifestVersion`. Further clarification shouldn't be necessary when actually interacting with the code, but it is definitely harder to see in a code review.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -354,51 +323,47 @@ private async Task UpdateAdvertisingManifestAsync(WorkloadManifestInfo manifest,
...
-            var manifestPath = Path.Combine(GetAdvertisingManifestPath(_sdkFeatureBand, manifestId),
-                "WorkloadManifest.json");
+            var manifestPath = Path.Combine(GetAdvertisingManifestPath(_sdkFeatureBand, manifestId), "WorkloadManifest.json");
             if (!File.Exists(manifestPath))
             {
                 return null;
             }
 
-            using (FileStream fsSource = new(manifestPath, FileMode.Open, FileAccess.Read))
-            {
```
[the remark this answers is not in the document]
MiYanni: It provides no functional purpose as the `using` variable calls `Dispose` at the end of the method in both situations. Both end up with the same call stack. If it needed more encapsulation, another method/function would help, but it doesn't add anything since this logic is the entire point of the method itself.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -483,9 +446,12 @@ private async Task<string> GetOnlinePackagePath(SdkFeatureBand sdkFeatureBand, M
         private string GetOfflinePackagePath(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, DirectoryPath? offlineCache = null)
         {
             string packagePath = Directory.GetFiles(offlineCache.Value.Value)
-                .Where(path => path.EndsWith(".nupkg"))
                 .Where(path =>
                 {
+                    if (!path.EndsWith(".nupkg"))
+                    {
+                        return false;
```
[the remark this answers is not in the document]
MiYanni: The code in this lambda expression already called `return` 3 lines below this. `return` in a lambda returns a result, like in a function or method. The LINQ method takes a delegate that returns a boolean. This logic is identical to the original `.Where(path => path.EndsWith(".nupkg"))` except it won't allocate all the LINQ expression boilerplate to combine the `Where`s together. It is generally bad practice to use multiple `Where` expressions in a row since you can just make 1 lambda handle all the boolean logic at the same time.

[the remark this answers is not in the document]
MiYanni: I've removed this method and merged the 1 statement into the only method calling it. I also merged the boolean logic into 1 expression with an `&&`, which is the much better way to do this.

src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
MiYanni: That's invalid. Using statements don't have the context of other using statements or namespaces unless they are declared inside a namespace.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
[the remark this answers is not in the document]
MiYanni: After seeing the screenshot above when I sat down at my desk, I realized I could move the variable creation together. Now, `installedVersion`/`installedBand` are created right before `adVersion`/`adBand`.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -213,68 +199,54 @@ public IEnumerable<ManifestVersionUpdate> CalculateManifestRollbacks(string roll
...
-                    (success, var latestVersion) = await GetPackageVersion(packageId, packageSourceLocation: _packageSourceLocation, includePreview: includePreviews);
-                    if (success)
-                    {
-                        downloads.Add(new WorkloadDownload(manifest.Id, packageId.ToString(), latestVersion.ToString()));
-                    }
-
-                    if (!success && !installedSdkFeatureBand.Equals(providedSdkFeatureBand))
+                    PackageId? providedPackageId = null;
+                    var fallbackFeatureBand = new SdkFeatureBand(manifest.ManifestFeatureBand);
+                    SdkFeatureBand[] bands = [providedSdkFeatureBand, installedSdkFeatureBand, fallbackFeatureBand];
```
Forgind, line 228: I wasn't actually familiar with this notation. Nice.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
Forgind: As I understand it, Distinct does _not_ guarantee that the order will remain the same. I seem to recall it puts everything into a Dictionary then just randomly pulls things out. In practice, I think it's often in the same order, but I wouldn't rely on that being true here if it's avoidable.

Since there are only three things in the list, we could potentially have a "check if we already processed this one (by checking the previous one(s)) and 'continue' if so"?

Forgind: For the record, I was more confused by the original code than MiYanni's version. If you have three things you want to check, you should do it in a loop, not in nested if statements.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -354,51 +323,47 @@ private async Task UpdateAdvertisingManifestAsync(WorkloadManifestInfo manifest,
...
-                .FirstOrDefault(manifest => manifest.Id.ToLowerInvariant().Equals(manifestId.ToString()));
+            var manifest = _workloadResolver.GetInstalledManifests().FirstOrDefault(manifest => manifest.Id.ToLowerInvariant().Equals(manifestId.ToString()));
             if (manifest == null)
             {
                 throw new Exception(string.Format(LocalizableStrings.ManifestDoesNotExist, manifestId.ToString()));
             }
-            return (new ManifestVersion(manifest.Version), new SdkFeatureBand(manifest.ManifestFeatureBand));
+            var version = new ManifestVersion(manifest.Version);
+            var band = new SdkFeatureBand(manifest.ManifestFeatureBand);
+            return new(version, band);
```
Forgind, line 360: I kinda prefer the collapsed form (just with a 'new' in front)

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -457,80 +421,43 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
...
-
-            return packagePath;
-        }
-
-        private async Task<(bool, string)> GetManifestPackageUpdate(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, bool includePreviews, DirectoryPath? offlineCache = null)
+        private async Task<(bool Success, string PackagePath)> GetManifestPackageUpdate(SdkFeatureBand sdkFeatureBand, ManifestId manifestId, bool includePreviews, DirectoryPath? offlineCache = null)
         {
-            if (offlineCache == null || !offlineCache.HasValue)
+            var manifestPackageId = _workloadManifestInstaller.GetManifestPackageId(manifestId, sdkFeatureBand);
+            if (offlineCache != null)
```
Forgind, line 439: Do you still need the HasValue check?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
```diff
@@ -457,80 +421,43 @@ private IEnumerable<(ManifestId id, ManifestVersion version, SdkFeatureBand feat
...
-                    return (true, packagePath);
-                }
-                catch (NuGetPackageNotFoundException)
-                {
-                    return (false, null);
-                }
+                string offlinePackagePath = Directory.GetFiles(offlineCache.Value.Value)
+                    .Where(path => path.EndsWith(".nupkg") && Path.GetFileName(path).StartsWith(manifestPackageId.ToString(), StringComparison.OrdinalIgnoreCase))
+                    .Max();
+                return (Success: offlinePackagePath != null, PackagePath: offlinePackagePath);
```
Forgind, line 444: I'm wondering if `offlinePackagePath != null` should be `!string.IsNullOrEmpty(offlinePackagePath)`

That is a slight change in behavior, but it feels a little safer to me.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
```diff
@@ -105,29 +106,23 @@ public override int Execute()
             return 0;
         }
 
-        internal UpdateAvailableEntry[] GetUpdateAvailable(IEnumerable<WorkloadId> installedList)
+        internal IEnumerable<UpdateAvailableEntry> GetUpdateAvailable(IEnumerable<WorkloadId> installedList)
         {
-            HashSet<WorkloadId> installedWorkloads = installedList.ToHashSet();
```
Forgind, line 110: How many ids are in this list? Making this change could be a perf regression if the number is large, though I'd be a bit surprised by that.

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -165,10 +165,10 @@ public void GivenNoWorkloadsInstalledInfoOptionRemarksOnThat()
             // However, we can test a setup where no workloads are installed and --info is provided. 
 
             _reporter.Clear();
-            var mockWorkloadIds = new WorkloadId[] { new WorkloadId("xamarin-android"), new WorkloadId("xamarin-android-build") };
+            _ = new WorkloadId[] { new WorkloadId("xamarin-android"), new WorkloadId("xamarin-android-build") };
```
Forgind, line 168: Why do you need to make these at all if they're just throwaways?

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
```diff
@@ -449,7 +446,7 @@ public void GivenWorkloadInstallItTreatsPreviewsAsSeparateFeatureBands()
             };
 
             var nugetDownloader = new MockNuGetPackageDownloader(dotnetRoot);
-            var manifestUpdater = new MockWorkloadManifestUpdater(manifestUpdates, tempDirManifestPath);
```
Forgind, line 452: I've been quietly facepalming every time I see this change. Thank you for deleting it.

Forgind (approved):
A few minor things, but overall, this looks great! Thanks for making this change!

exclusion.dic
```diff
@@ -2,6 +2,7 @@ awaiter
 crossgen
 cshtml
 deserializer
+infos
```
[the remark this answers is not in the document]
MiYanni: If something is named eg. `argumentInfo`, we'd call a collection of them `argumentInfos`. There are other things that don't abide by English rules but make sense in programmer country. 🧑‍🌾

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: Collection expressions are a new thing. But they do come in handy and are really clean compared to traditional collection initializers.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: > As I understand it, Distinct does not guarantee that the order will remain the same.

This is true... to an extent. I had a long discussion with Smit as our resident expert on this kind of stuff. `Distinct` is *declared* to not preserve order but in operation on something like an array or list, it does based on how it is implemented. Smit wrote on the whiteboard right behind my desk a LINQ expression that would solve this were you make sure to order the elements by index in a `Select` that would do the `Distinct` call. But that's a lot of extra work and not worth the effort. The likelihood of the underlying operation for `Distinct` on an array/list changing is very, very low, as it would be a massive breaking change in the runtime.

> "check if we already processed this one (by checking the previous one(s)) and 'continue' if so"?

You've missed the multiple evolutions of this method, but I did do that previously. The problem with that is it took me *quite a while* to realize that was the purpose of the `!fallbackFeatureBand.Equals(installedSdkFeatureBand)` checks in the original code. It wasn't blatantly obvious that was the point of those checks is to not duplicate work on already processed bands. I feel `Distinct` does that job very concisely. However, I was planning on also adding some comments here based on Noah's feedback.

After some discussion with Smit, he gave me a suggestion on another way to do this that should make the clarity of `success` and preserve order. It doesn't use a loop, so I'm going to give that a try now.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: So it would be:
```C#
return new(new ManifestVersion(manifest.Version), new SdkFeatureBand(manifest.ManifestFeatureBand));
```

Or I could also do:
```C#
ManifestVersion version = new(manifest.Version);
SdkFeatureBand band = new(manifest.ManifestFeatureBand);
return new(version, band);
```

Preference? I'm actually not against the second one now that I see it. 🤔

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: I actually tested this before changing it because I questioned how `Max` works.
https://dotnetfiddle.net/ZjvAt1

The `Where` can only return non-empty strings of elements or `null`. It isn't possible for it to return empty string based on the condition set in the `Where`.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
MiYanni: Noah said something similar. I responded to that concern here: https://github.com/dotnet/sdk/pull/35859/files#r1347950133

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
MiYanni: That is certainly a question... 🤔 Let me dig in the history. If they're remnants from the past, I'll just remove them.

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
MiYanni: It was just an artifact left over from the creation of `IWorkloadManifestInstaller`, which now has this value built into it. When seeing unused properties like this, you generally should look at the code history to see if it was, in fact, no longer used or is actually some kind of glaring bug that was missed.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: Oh man. This version of it is so clean... but is has one large caveat. It has no way to deduplicate the bands if they're identical bands! 💢 It would add a bunch of goofy boilerplate to make it do that. This would definitely have a perf penalty but I like it so much! 😭

```C#
        public async Task<IEnumerable<WorkloadDownload>> GetManifestPackageDownloadsAsync(bool includePreviews, SdkFeatureBand providedSdkFeatureBand, SdkFeatureBand installedSdkFeatureBand)
        {
            var downloads = new List<WorkloadDownload>();
            PackageId? providedPackageId = null;
            foreach (var manifest in _workloadResolver.GetInstalledManifests())
            {
                try
                {
                    var fallbackSdkFeatureBand = new SdkFeatureBand(manifest.ManifestFeatureBand);
                    // Try each feature band for acquiring the workload download info.
                    WorkloadDownload workloadDownload =
                        await CreateWorkloadDownload(manifest.Id, providedSdkFeatureBand) ??
                        await CreateWorkloadDownload(manifest.Id, installedSdkFeatureBand) ??
                        await CreateWorkloadDownload(manifest.Id, fallbackSdkFeatureBand);

                    // If we couldn't create the workload download, report the error.
                    if (workloadDownload == null)
                    {
                        _reporter.WriteLine(LocalizableStrings.ManifestPackageUrlNotResolved, providedPackageId);
                    }
                    else
                    {
                        downloads.Add(workloadDownload);
                    }
                }
                catch
                {
                    _reporter.WriteLine(LocalizableStrings.ManifestPackageUrlNotResolved, manifest.Id);
                }
            }

            return downloads;

            async Task<WorkloadDownload> CreateWorkloadDownload(string manifestId, SdkFeatureBand band)
            {
                var packageId = _workloadManifestInstaller.GetManifestPackageId(new ManifestId(manifestId), band);
                providedPackageId ??= packageId;

                try
                {
                    var latestVersion = await _nugetPackageDownloader.GetLatestPackageVersion(packageId, _packageSourceLocation, includePreviews);
                    return new WorkloadDownload(manifestId, packageId.ToString(), latestVersion.ToString());
                }
                catch (NuGetPackageNotFoundException)
                {
                }

                return null;
            }
        }
```

src/Tests/dotnet-workload-install.Tests/GivenDotnetWorkloadInstall.cs
MiYanni: After some sleuthing, it was @nagilson in this commit! Dun dun daaaaaaaaa. 🙈
https://github.com/dotnet/sdk/commit/011e1e8b86cc4f2772db191a6373b4c8bd5d225d#diff-c8bc89094b206b506f35f0cb9ac6e55591ae2468130efe0a19dab1810c0f0c1cL185-R191

Looks like an artifact of the previous implementation of the test. I'll remove these lines.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: For something using `Nullable<T>`, checking if it is `null` or checking if it has a value is the same thing. The "nullness" of the class represents if it has a value of T. https://stackoverflow.com/a/676089/294804

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
Forgind: I prefer the first. If constructing one or both parts is complicated (like if there have to be nested constructors under it), I think it's good to separate them out...but these are quite simple.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: After further investigation and trial/error, I'm gonna keep this current implementation with the `foreach` and `Distinct` for a couple of reasons:
- `Distinct` is [currently implemented](https://github.com/dotnet/runtime/blob/f20b2b50f159e64ec45d65e0dc0e7c644ee75b37/src/libraries/System.Linq/src/System/Linq/Distinct.cs#L122-L130) to preserve order. It uses a backing `HashSet` to determine if a value is unique and only returns an element if it was able to be added to the `HashSet`. It uses the iterator of the collection passed to it, so if that collection iterates in order (here, the `Array` iterates in order), then it will always return the elements in order. [Here](https://stackoverflow.com/a/69527769/294804) is a simple explanation of that.
- This actually avoids a perf issue with the previous implementation. The previous implementation worked this way:

| Variable | Value
| --- | ---
| A | 1
| B | 2
| C | 1

Using the little chart here, we know there were 3 bands being checked; I'm using A, B, C, to represent those. It would run `A` and if `A` failed, it would check if `B == A`. If it didn't, it would run `B`. If `B` failed, it would check if `C == B`. If it didn't, then run `C`. The issue here is that it never checks if `C == A`. So it ends up running the **same band twice**. If we use `Distinct`, this scenario can't happen.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: I've changed this to the first (single statement return).

exclusion.dic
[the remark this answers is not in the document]
MiYanni: This file doesn't exist in this branch, so this is no longer being added.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
MiYanni: I've added some code comments to this method that should hopefully aid in understanding. I've also updated `UpdateAdvertisingManifestAsync` to follow this same pattern since it was using the previous pattern except with 2 bands instead of 3.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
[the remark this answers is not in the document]
nagilson: That is fair, I tend to lean towards using {} because of this infamous Apple bug, https://medium.com/swlh/apples-most-notorious-code-bug-6478ebaea44f but it wouldn't change the behavior here, so im cool with this 😄

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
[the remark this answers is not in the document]
nagilson: I see, that makes sense -- thank you for the explanation :)

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadManifestUpdater.cs
nagilson: Thank you, looking at it again I much prefer the new version, plus it avoids the A == C problem is quite good, I do believe its a problem because the fallback feature band is the manifest feature band, which could be the same as the current sdk feature band

nagilson: Also the temptation for the clean version is real but I think what you settled with is for the best 😄

nagilson (approved):
LGTM now, this is a large improvement, thank you for making this :)

dsplaisted:
This looks good, thanks!

Could / should we use a global using for `WorkloadCollection` so it doesn't need to be repeated in different files?

MiYanni:
@dsplaisted 

> Could / should we use a global using for WorkloadCollection so it doesn't need to be repeated in different files?

Oh, that's a great idea! I forgot you could put aliases in those. I'll do that in a separate PR real quick.

src/Cli/dotnet/commands/dotnet-workload/list/WorkloadListCommand.cs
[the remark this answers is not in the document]
MiYanni: Daniel just pointed out we can put the alias in the *global usings* we have. I completely forgot aliases can go in there as global usings is a newer feature. I'm making a separate PR for it.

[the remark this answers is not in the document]
MiYanni: PR here: https://github.com/dotnet/sdk/pull/36092

PR #35884, opened 2023-10-05
Fix concurrent tool restore and version range error

JL03-Yue:
**Customer Impact**

The change, however, affect the behavior in tool restore. For one, the change in tool install affect the parallel tool restore as it might read multiple too info into the same config file. For another, the change of getting tool's version tries to read from a property `.originalString` that might not exist. 

The intended way to fix this is to make the parallel tool restore sequential, and to get rid of reading the `.originalString` property. To make the tool restore sequential, instead of using `AsParallel()` in the code, the updated code uses `AsEnumerate()` to reduce the chance of error. For version, this pull request adds a condition to check if the `versionRange` passed in is null and set it to a default value. 
Both changes are validated by tests.

**Testing**
Existing unit tests passed without changes. New tested are added to validate tool restore returns the correct version, and tool restore is able to restore multiple tools. The new testing scenarios pass with changes in the pull request. 

**Risk**
Low. This change resolves some errors caused by breaking change by add conditions to set the `versionRange` and make parallel tool restore sequential, without adding complexity to the behavior. 

Current unit tests passed without changes, and new tests are added to validate the change. Even though making parallel tool restore sequential might reduce the speed of tool restore, this is a validate approach because it reduces the risk of issues from concurrent restoration.

**Original description**
Referred to [Add known issue for .NET tool restore in .NET 8 RC2 ](https://github.com/dotnet/core/pull/8799/commits/86917f474fcadfe909443f29a87e4d3284e7a7f5 )

### DIFF
--- a/src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
+++ b/src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
@@ -26,6 +26,7 @@ using NuGet.Repositories;
 using NuGet.RuntimeModel;
 using NuGet.Versioning;
 using NuGet.Configuration;
+using Microsoft.TemplateEngine.Utils;
 
 namespace Microsoft.DotNet.Cli.ToolPackage
 {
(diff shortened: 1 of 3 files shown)


dsplaisted (approved):
Looks great!

wtgodbe:
Approved over email

JL03-Yue:
fixes #35999

Issue #33819, opened 2023-07-06
build.cmd the dotnet/sdk Something went wrong


tangyanzhi:
I built dotnet/sdk via build.cmd, but something went wrong, the error is as follows:

```
Build FAILED.

G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\GivenThatWeWantToPublishAFrameworkDependentApp.cs(125,103): error
 CS1056: Unexpected character '\' [G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\Microsoft.NET.Publish.Tests.cs
proj]
G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\GivenThatWeWantToPublishAFrameworkDependentApp.cs(125,104): error
 CS1003: Syntax error, ',' expected [G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\Microsoft.NET.Publish.Tests.
csproj]
G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\GivenThatWeWantToPublishAFrameworkDependentApp.cs(125,121): error
 CS1056: Unexpected character '\' [G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\Microsoft.NET.Publish.Tests.cs
proj]
G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\GivenThatWeWantToPublishAFrameworkDependentApp.cs(125,122): error
 CS1003: Syntax error, ',' expected [G:\sdk-main\sdk\src\Tests\Microsoft.NET.Publish.Tests\Microsoft.NET.Publish.Tests.
csproj]
    0 Warning(s)
    4 Error(s)

Time Elapsed 00:00:56.76
Build failed with exit code 1. Check errors above.
```

How can I fix this error?

joeloff:
Have you been able to build successfully? This could have been a bad merge and transient issue.

PR #35891, opened 2023-10-06
Update GivenThatWeWantToPublishAFrameworkDependentApp.cs to fix chars…

zcxsythenew:
…et error.

This pull request addresses Issue #33819 .

When I run `build.cmd` on Windows 11 in Chinese environment, I reproduced the above issue. Then I tried to open the source file in Visual Studio, and there was a popup window which is shown below.

图片

And I found that there are two special characters in Line 115, which causes the issue. (The special characters are replaced with `?` below.)

```text
.HaveStdOutContaining(Strings.FrameworkDependentAppHostRequiresVersion21.Replace("?", "\"").Replace("?", "\""));
```

Then I opened the binary editor and found the code of the two special characters, one of which is `93` and another is `94`.

图片

I think we had better use unicode escape sequences instead of placing the special characters in souce code. Therefore, I raised this pull request. This pull request just replaces the two special characters with `\u0093` and `\u0094`.

### DIFF
--- a/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAFrameworkDependentApp.cs
+++ b/src/Tests/Microsoft.NET.Publish.Tests/GivenThatWeWantToPublishAFrameworkDependentApp.cs
@@ -112,7 +112,7 @@ namespace Microsoft.NET.Publish.Tests
                 .Should()
                 .Fail()
                 .And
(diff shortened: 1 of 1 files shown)


Forgind (approved):
That sounds like a good idea to me, thanks!

Markliniubility:
I experience #33819 as well, and this PR fixes it. Thanks!

PR #35956, opened 2023-10-10
IDE0036 "Modifiers are not ordered" cleanup

MiYanni:
## Summary
I ran the fix for [Order modifiers (IDE0036)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036). These types of automated changes are non-functional changes. Just purely code cleanup.

### DIFF
--- a/src/Cli/Microsoft.DotNet.Cli.Sln.Internal/FileManipulation/FileUtil.cs
+++ b/src/Cli/Microsoft.DotNet.Cli.Sln.Internal/FileManipulation/FileUtil.cs
@@ -28,7 +28,7 @@
 
 namespace Microsoft.DotNet.Cli.Sln.Internal.FileManipulation
 {
-    static internal class FileUtil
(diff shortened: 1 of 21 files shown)


danmoseley:
Is there a way to change the rule to cause errors now (or is that a non goal)?

marcpopMSFT (approved):

MiYanni:
@danmoseley 

> Is there a way to change the rule to cause errors now (or is that a non goal)?

Non-goal. We decided vanity changes should not block builds. I created [this issue](https://github.com/dotnet/sdk/issues/34810) to track our thoughts. *Personally*, I still think the IDE should support auto-formatting on-save for certain types of rules/analyzers, like a more advanced linter. This IDE0036 is a perfect example of that situation, "I'm just gonna type all these access modifiers. I don't care what the order is. You handle it for me, please." It really should be the industry standard at this point.

PR #35964, opened 2023-10-10
Move compatibility tools into common folder

ViktorHofer:
Three commits:
1. Move compatibility tools (APICompat & GenAPI) together into a common "Compatibility" folder. This allows to define common properties more easily and avoids references across the repository.
2. Update the paths in build files and solution files because of the move.
3. Small TFM fixes and updates


ericstj (approved):
LGTM.  This will make it easier to work on these tools.

Issue #35566, opened 2023-09-20
[8.0.100-rc.2.23469.4] some deprecated global tools get NuGetPackageNotFoundException when installing


Junjun-zhao:
**Issue Description**:
When installing these global tools with "dotnet tool install" command, NuGetPackaeNotFoundException will be thrown on dotnet-sdk-8.0.100-rc.2.23469.4. But they are installed successful with dotnet-sdk-8.0.100-rc.2.23466.1.

**Application Name**: Elemental.SysInfoTool, dotnet-giio, dotnet-sqldb
**OS**： Windows 10 21H2
**CPU**: X64
**.NET Build Number**: dotnet-sdk-8.0.100-rc.2.23469.4

**Verify Scenarios**:
1. Windows10 21h2 x64 + dotnet-sdk-8.0.100-rc.2.23469.4: Fail
2. Windows10 21h2 x64 + dotnet-sdk-8.0.100-rc.2.23466.1: Pass

**Repro steps**:
The machine only has 8.0.100-rc.2.23469.4 installed and set DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX environment variable as 2.
1. Open cmd.exe form %windir%\System32 
2. Input ‘dotnet tool install -g Elemental.SysInfoTool’ 
3. Press keyboard ‘Enter’ 

**Expected Result**:
Elemental.SysInfoTool install successful.

**Actual Result**:
Install failed with following exception:
```
[...]

PR #36021, opened 2023-10-11
Enable users to install unlisted version of tool if specify the exact match of the version of the tool

JL03-Yue:
**Customer Impact**

The change, however, disable users from installing unlisted tools. Previously, users were able to install unlisted tools per fix in #28951. However, since the mechanism of tool installation are changed, the unlisted tools' downloading are disabled by default. For users, installing unlisted tools are a useful method because it provides clarity and simplicity for those who write tools independently or do not share the tools to the public. 

The intended way to fix this is to provide a feature to allow users to install the unlisted version but only if the user specify the exact match, i.e. `[version]`. (referred to [NuGet package version reference](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning)). To add this feature, an unlisted parameter is passed to the NuGet API if a exact version is detected. Changes are validated by tests.

**Testing**
Existing unit tests passed without changes. New tested are added to validate an unlisted tool can be installed with exact version match, and failed to be installed without the version match. The new testing scenarios pass with changes in the pull request. 

**Risk**
Low. This change adds a feature that was missed from breaking change without adding complexity to the behavior. 

**Original description**
Referred to [some deprecated global tools get NuGetPackageNotFoundException when installing](https://github.com/dotnet/sdk/issues/35566)

### DIFF
--- a/src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
+++ b/src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
@@ -93,6 +93,12 @@ namespace Microsoft.DotNet.Cli.ToolPackage
                     var nugetPackageDownloader = new NuGetPackageDownloader.NuGetPackageDownloader(toolDownloadDir, verboseLogger: nugetLogger, isNuGetTool: true);
 
                     var packageSourceLocation = new PackageSourceLocation(packageLocation.NugetConfig, packageLocation.RootConfigDirectory, null, packageLocation.AdditionalFeeds);
+
+                    bool givenSpecificVersion = false;
+                    if (!string.IsNullOrEmpty(versionRange.OriginalString))
+                    {
+                        givenSpecificVersion = true;
+                    }
                     NuGetVersion packageVersion = nugetPackageDownloader.GetBestPackageVersionAsync(packageId, versionRange, packageSourceLocation).GetAwaiter().GetResult();
 
                     rollbackDirectory = isGlobalTool ? toolDownloadDir.Value: Path.Combine(toolDownloadDir.Value, packageId.ToString(), packageVersion.ToString());
@@ -116,7 +122,7 @@ namespace Microsoft.DotNet.Cli.ToolPackage
 
                     if (package == null)
                     {
-                        DownloadAndExtractPackage(packageLocation, packageId, nugetPackageDownloader, toolDownloadDir.Value, _toolPackageStore, packageVersion, packageSourceLocation).GetAwaiter().GetResult();
+                        DownloadAndExtractPackage(packageLocation, packageId, nugetPackageDownloader, toolDownloadDir.Value, _toolPackageStore, packageVersion, packageSourceLocation, givenSpecificVersion).GetAwaiter().GetResult();
                     }
                     else if(isGlobalTool)
                     {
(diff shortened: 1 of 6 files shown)


src/Tests/dotnet.Tests/CommandTests/ToolInstallGlobalOrToolPathCommandTests.cs
```diff
@@ -352,11 +352,11 @@ public void WhenRunWithValidVersionRangeItShouldSucceed()
                     PackageVersion).Green());
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/dotnet/sdk/pull/36021")]
```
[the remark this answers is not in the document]
JL03-Yue: Yes

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -116,7 +122,7 @@ internal class ToolPackageDownloader : IToolPackageDownloader
 
                     if (package == null)
                     {
-                        DownloadAndExtractPackage(packageLocation, packageId, nugetPackageDownloader, toolDownloadDir.Value, _toolPackageStore, packageVersion, packageSourceLocation).GetAwaiter().GetResult();
+                        DownloadAndExtractPackage(packageLocation, packageId, nugetPackageDownloader, toolDownloadDir.Value, _toolPackageStore, packageVersion, packageSourceLocation, givenSpecificVersion).GetAwaiter().GetResult();
```
dsplaisted, line 125: Including the parameter name here should make this a bit easier to read/understand.

```suggestion
                        DownloadAndExtractPackage(packageLocation, packageId, nugetPackageDownloader, toolDownloadDir.Value, _toolPackageStore, packageVersion, packageSourceLocation, includeUnlisted: givenSpecificVersion).GetAwaiter().GetResult();
```

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -93,6 +93,12 @@ internal class ToolPackageDownloader : IToolPackageDownloader
                     var nugetPackageDownloader = new NuGetPackageDownloader.NuGetPackageDownloader(toolDownloadDir, verboseLogger: nugetLogger, isNuGetTool: true);
 
                     var packageSourceLocation = new PackageSourceLocation(packageLocation.NugetConfig, packageLocation.RootConfigDirectory, null, packageLocation.AdditionalFeeds);
+
+                    bool givenSpecificVersion = false;
+                    if (!string.IsNullOrEmpty(versionRange.OriginalString))
```
dsplaisted, line 98: I don't think the fact that it has an OriginalString necessarily means it is for a specific version.  The original string could be "*" or something.

I think it would be better to check that the min and max version are the same, similar to what you did in `GetBestPackageVersionAsync`:

```c#
if(versionRange.MinVersion != null && versionRange.MaxVersion != null && versionRange.MinVersion == versionRange.MaxVersion)
```

src/Tests/dotnet.Tests/CommandTests/ToolInstallGlobalOrToolPathCommandTests.cs
```diff
@@ -351,6 +352,31 @@ public void WhenRunWithValidVersionRangeItShouldSucceed()
                     PackageVersion).Green());
         }
 
+        [Fact(Skip = "https://github.com/dotnet/sdk/pull/36021")]
+        public void WhenRunWithValidUnlistedVersionRangeItShouldSucceed()
```
dsplaisted, line 356: I think this test should be enabled in CI if it can work there.

Can you also add a test that checks that an unlisted package won't be installed if you don't specify a version?

dsplaisted (approved):
Looks good!

I have a few comments but I'm marking this as approved since I'll be out tomorrow.

src/Tests/dotnet.Tests/CommandTests/ToolInstallGlobalOrToolPathCommandTests.cs
JL03-Yue: Thank you! A new test is added for unlisted package without version and the previous test was updated to use `DotnetCommand`

PR #36067, opened 2023-10-12
Use async test for async method

lewing:


ViktorHofer:
This got already fixed in main with https://github.com/dotnet/sdk/pull/36061

PR #36090, opened 2023-10-12
Reduce test timeout to speed up timeouts.

marcpopMSFT:
This will affect ~.2% of test work items

From some queries, in the past 30d we ran 1031819 test work items in Helix. Of those, 1695 took longer than 45 minutes. Most of those were 45-60 minutes but it is probably almost as fast to timeout and retry than to let it keep going in many of those cases.

### DIFF
--- a/src/Tests/UnitTests.proj
+++ b/src/Tests/UnitTests.proj
@@ -8,7 +8,7 @@
 
     <EnableAzurePipelinesReporter Condition="'$(BUILD_BUILDNUMBER)' != ''">true</EnableAzurePipelinesReporter>
 
-    <XUnitWorkItemTimeout>2:00:00</XUnitWorkItemTimeout>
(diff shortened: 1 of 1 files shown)


MiYanni (approved):

src/Tests/UnitTests.proj
```diff
@@ -8,7 +8,7 @@
 
     <EnableAzurePipelinesReporter Condition="'$(BUILD_BUILDNUMBER)' != ''">true</EnableAzurePipelinesReporter>
 
-    <XUnitWorkItemTimeout>2:00:00</XUnitWorkItemTimeout>
+    <XUnitWorkItemTimeout>45:00</XUnitWorkItemTimeout>
```
baronfel, line 11:
```suggestion
    <XUnitWorkItemTimeout>00:45:00</XUnitWorkItemTimeout>
```

Looks like XUnit has a required format.

Issue #35128, opened 2023-09-04
[NETE2ESDK] With .NET 8 RC2 SDK pkg/deb/rpm installed, "An issue was encountered verifying workloads. For more information, run "dotnet workload update"." displayed when running the first dotnet command.


WenJunJi01:
**Repro steps:**
1. Prepare clean Mac OS or Linux machine.
2. Install .NET 8 RC2 SDK(8.0.100-rc.2.23451.1(runtime-8.0.0-rc.2.23431.9)) from https://github.com/dotnet/installer 
(pkg on Mac OS, deb/rmp on Linux)
4. Open a terminal and run the first dotnet command.
  
**Expected:**
It should execute successfully with no errors or warnings.

**Actual result:**
The following error displayed, "An issue was encountered verifying workloads. For more information, run "dotnet workload update"." 
Mac OS:
![image](https://github.com/dotnet/sdk/assets/87680983/48729732-b165-4bee-be56-258e44674ca0)
Linux:
![image](https://github.com/dotnet/sdk/assets/87680983/ca375624-f4a9-4361-8808-1807603283f1)

**Note:**
1. Not repro with exe installed on Windows and tar.gz installed on Mac OS/Linux.
2. Only appears the first time the dotnet command is run.

**dotnet --info:**
.NET SDK:
 Version:   8.0.100-rc.2.23451.1
 Commit:    dabc29073d

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  14.0
 OS Platform: Darwin
 RID:         osx-x64
 Base Path:   /usr/local/share/dotnet/sdk/8.0.100-rc.2.23451.1/

.NET workloads installed:
 [wasm-tools]
   Installation Source: SDK 8.0.100-rc.2
   Manifest Version:    8.0.0-rc.2.23431.9/8.0.100-rc.2
   Manifest Path:       /usr/local/share/dotnet/sdk-manifests/8.0.100-rc.2/microsoft.net.workload.mono.toolchain.current/8.0.0-rc.2.23431.9/WorkloadManifest.json
   Install Type:        FileBased

Host:
  Version:      8.0.0-rc.2.23431.9
  Architecture: x64
  Commit:       3c48925a6c
  RID:          osx-x64

.NET SDKs installed:
  8.0.100-rc.2.23451.1 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
[...]

PR #36123, opened 2023-10-14
Skip workload integrity check on Linux and non-sudo

MiYanni:
Fixes: https://github.com/dotnet/sdk/issues/35128

## Summary
For Linux, it seems that `sudo` is required to update workload manifests. Therefore, the workload integrity check will fail if the first-run command is not `sudo` on Linux. I've added a condition when setting the default `skipWorkloadIntegrityCheck` to account for this scenario so the check will be skipped appropriately. Also, minor cleanup in the `IsRunningUnderSudo` method (for simplification purposes).

### DIFF
--- a/src/Cli/dotnet/Program.cs
+++ b/src/Cli/dotnet/Program.cs
@@ -166,9 +166,12 @@ namespace Microsoft.DotNet.Cli
                     bool telemetryOptout = environmentProvider.GetEnvironmentVariableAsBool(EnvironmentVariableNames.TELEMETRY_OPTOUT, defaultValue: CompileOptions.TelemetryOptOutDefault);
                     bool addGlobalToolsToPath = environmentProvider.GetEnvironmentVariableAsBool(EnvironmentVariableNames.DOTNET_ADD_GLOBAL_TOOLS_TO_PATH, defaultValue: true);
                     bool nologo = environmentProvider.GetEnvironmentVariableAsBool(EnvironmentVariableNames.DOTNET_NOLOGO, defaultValue: false);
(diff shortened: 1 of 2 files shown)


MiYanni:
@KalleOlaviNiemitalo 

> Do you intend to fix it for macOS in a different way?

Ah right. I'll flip the logic to skip it for non-Windows instead of specifically Linux. 👍

> Is the workload integrity check a security feature?

It is not. It only checks if the workloads on-disk are out-of-date or incorrect for the given SDK feature band.

dsplaisted:
Mac / Linux don't always require sudo to update workloads, it depends on how they are installed.  If they are installed with a package manager then you will usually need sudo to update workloads, but if the SDK was just unzipped somewhere, it probably won't.

Would it be possible on these OS's to check whether the current process has write permission to the `dotnet` or `sdk-manifests` directory, and if not then skip the workload integrity check?

blowdart:
> It only checks if the workloads on-disk are out-of-date or incorrect for the given SDK feature band.

Can we expand on "incorrect" here?

I'd also argue out of date is a security check if newer versions of a workload contain security fixes.

joeloff:
> Mac / Linux don't always require sudo to update workloads, it depends on how the .NET SDK is installed. If they are installed with a package manager then you will usually need sudo to update workloads, but if the SDK was just unzipped somewhere, it probably won't.
> 
> Would it be possible on these OS's to check whether the current process has write permission to the `dotnet` or `sdk-manifests` directory, and if not then skip the workload integrity check?

You should be able to query permissions using the stat API in Linux. Not sure whether we have wrappers for this in .NET already.

MiYanni:
I'm closing this for the time being. I'll revisit when I have time to produce a proper solution.

PR #36156, opened 2023-10-17
try reverting to 17.7 to unblock the full framework tests

marcpopMSFT:


marcpopMSFT:
full framework leg still broke in the same way so this isn't an option

PR #36161, opened 2023-10-17
Use VS 17.7 for FullFramework tests

jjonescz:


marcpopMSFT:
I also tried this and it didn't work

PR #36195, opened 2023-10-17
[release/8.0.1xx] Define a default set of hot reload capabilities for Blazor WebAssembly

MackinnonBuck:
## [release/8.0.1xx] Define a default set of hot reload capabilities for Blazor WebAssembly

Fixes an issue preventing `dotnet watch` from working when a Blazor Web app is configured to use WebAssembly interactivity.

## Description

If a `dotnet watch` hot reload session attempts to initialize and read WebAssembly hot reload capabilities before the WebAssembly .NET runtime starts, hot reload breaks for the entirety of the session. This bug is almost inevitable in Blazor Web scenarios, since Blazor WebAssembly initialization gets deferred until the first interactive WebAssembly component appears on the page.

This PR addresses the issue by defining a set of default WASM hot reload capabilities to use if the WebAssembly runtime is not available by the time the hot reload session initializes.

Fixes https://github.com/dotnet/aspnetcore/issues/50765

## Customer Impact

Most customers developing a "Full stack Blazor"-style app with WebAssembly interactivity enabled will have a completely broken hot reload experience without this fix.

## Regression?

- [ ] Yes
- [X] No

This bug only appears in "Full stack Blazor" apps, which are new in .NET 8.

## Risk

- [ ] High
- [ ] Medium
- [X] Low

The fix is fairly straightforward and is isolated to the development experience (app functionality is not affected).

## Verification

- [X] Manual (required)
- [ ] Automated

- [ ] Yes
- [ ] No
- [X] N/A

### DIFF
--- a/src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
@@ -13,8 +13,13 @@ namespace Microsoft.DotNet.Watcher.Tools
 {
     internal sealed class BlazorWebAssemblyDeltaApplier : SingleProcessDeltaApplier
     {
+        private const string DefaultCapabilities60 = "Baseline";
+        private const string DefaultCapabilities70 = "Baseline AddMethodToExistingType AddStaticFieldToExistingType NewTypeDefinition ChangeCustomAttributes";
+        private const string DefaultCapabilities80 = "Baseline AddMethodToExistingType AddStaticFieldToExistingType NewTypeDefinition ChangeCustomAttributes AddInstanceFieldToExistingType GenericAddMethodToExistingType GenericUpdateMethod UpdateParameters GenericAddFieldToExistingType";
+
         private static Task<ImmutableArray<string>>? s_cachedCapabilties;
         private readonly IReporter _reporter;
+        private Version? _targetFrameworkVersion;
         private int _sequenceId;
 
         public BlazorWebAssemblyDeltaApplier(IReporter reporter)
@@ -30,6 +35,8 @@ namespace Microsoft.DotNet.Watcher.Tools
 
             // Configure the app for EnC
             context.ProcessSpec.EnvironmentVariables["DOTNET_MODIFIABLE_ASSEMBLIES"] = "debug";
+
+            _targetFrameworkVersion = context.FileSet?.Project?.TargetFrameworkVersion;
         }
 
         public override Task<ImmutableArray<string>> GetApplyUpdateCapabilitiesAsync(DotNetWatchContext context, CancellationToken cancellationToken)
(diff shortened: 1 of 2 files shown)


src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
```diff
@@ -59,12 +66,24 @@ async Task<ImmutableArray<string>> GetApplyUpdateCapabilitiesCoreAsync()
...
 
                     var capabilities = Encoding.UTF8.GetString(buffer.AsSpan(0, response.Value.Count));
+                    var shouldFallBackToDefaultCapabilities = false;
 
                     // error while fetching capabilities from WASM:
                     if (capabilities.StartsWith("!"))
                     {
-                        _reporter.Error($"Exception while reading WASM runtime capabilities: {capabilities[1..]}");
-                        return ImmutableArray<string>.Empty;
+                        _reporter.Warn($"Exception while reading WASM runtime capabilities: {capabilities[1..]}");
```
MackinnonBuck, line 74: I downgraded this to a warning since it no longer indicates a fatal error.

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
```diff
@@ -76,6 +95,15 @@ async Task<ImmutableArray<string>> GetApplyUpdateCapabilitiesCoreAsync()
                     ArrayPool<byte>.Shared.Return(buffer);
                 }
             }
+
+            static string GetDefaultCapabilities(Version? targetFrameworkVersion)
+                => targetFrameworkVersion?.Major switch
```
tmat, line 100: Do we need to make this TFM-dependent? It shouldn't matter what the application targets. We know we are running on .NET 8 runtime, no? Or is it possible for .NET 8 runtime to load older version of WASM?

MackinnonBuck: I did it this way to handle the case where a .NET 8 SDK is used with a project targeting an earlier target framework version. I didn't think the dotnet watch tool would be tied to the runtime version of the app - is it?

tmat: The runtime is not determined by the TFM, but rather by global.json if present, right?
If a project builds against net50 it would still run on .NET 8 if it is launched from .NET 8 SDK.

MackinnonBuck: Hm, I thought the TFM determined the runtime, not the SDK. But I could be wrong. I believe [this](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#:~:text=The%20global.json%20file%20allows%20you%20to%20define%20which%20.NET%20SDK%20version%20is%20used%20when%20you%20run%20.NET%20CLI%20commands.%20Selecting%20the%20.NET%20SDK%20version%20is%20independent%20from%20specifying%20the%20runtime%20version%20a%20project%20targets) suggests that it's the TFM that controls the runtime version?

tmat: Never mind, you are right.

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
```diff
@@ -30,6 +35,8 @@ public override void Initialize(DotNetWatchContext context, CancellationToken ca
 
             // Configure the app for EnC
             context.ProcessSpec.EnvironmentVariables["DOTNET_MODIFIABLE_ASSEMBLIES"] = "debug";
+
+            _targetFrameworkVersion = context.FileSet?.Project?.TargetFrameworkVersion;
```
tmat, line 39: `context.TargetFramework` should work

tmat (approved):

mkArtakMSFT:
Added the `NO MERGE` label because I think @MackinnonBuck was trying to get @javiercn approval on this.

javiercn (approved):

src/BuiltInTools/dotnet-watch/HotReload/BlazorWebAssemblyDeltaApplier.cs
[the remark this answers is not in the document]
MackinnonBuck: Would that be preferable, even if we can recover successfully and have hot reload continue to work? This 'error' code path is what gets taken in https://github.com/dotnet/aspnetcore/issues/50765 and https://github.com/dotnet/aspnetcore/issues/51208, but hot reload can continue to work in both cases with this fallback logic.

MackinnonBuck: @tmat Should we remove the `console.warn(error)` as well? I'm thinking that might be good since it will be a common warning and customers might get annoyed that they can't get rid of it.

Issue #33388, opened 2023-06-16
SBRP System.Configuration.ConfigurationManager 7.0.0 missing public types


JanKrivanek:
### Describe the Bug
[system.configuration.configurationmanager.7.0.0](https://github.com/dotnet/source-build-reference-packages/tree/main/src/referencePackages/src/system.configuration.configurationmanager/7.0.0) in https://github.com/dotnet/source-build-reference-packages is missing some public types available in the nuget package.
One example being `ConfigurationPropertyAttribute` - this is as a result causing source build build breaks when we attempt to oboard newest SBRP in msbuild: https://github.com/dotnet/msbuild/pull/8893

The missing type is not added even when I try to regenerate package with `generate.cmd -p system.configuration.configurationmanager,7.0.0`

### Steps to Reproduce
1) https://github.com/dotnet/source-build-reference-packages/blob/main/src/referencePackages/src/system.configuration.configurationmanager/7.0.0/lib/net7.0/System.Configuration.ConfigurationManager.cs
2) Search for `ConfigurationPropertyAttribute` - missing
3) https://www.fuget.org/packages/System.Configuration.ConfigurationManager/7.0.0/lib/net7.0/System.Configuration.ConfigurationManager.dll/System.Configuration/ConfigurationPropertyAttribute - type is available (and public)

MichaelSimons:
> @MichaelSimons this is happening because SBRP explicitly excludes that type when generating the reference package source: https://github.com/dotnet/source-build-reference-packages/blob/9cee0e5f54df0b2f2f5161aec2102b82ad6043df/src/packageSourceGenerator/DefaultGenApiDocIds.txt#LL9C1-L9C54

Thanks for pointing that out.

> That said, someone could argue that the types in that list should only be excluded when appearing as attributes and not excluding the types themselves.

Agree, that appears to be what is needed for msbuild.

ViktorHofer:
Taking a look and will try to send a PR until tomorrow so that the msbuild PR gets unblocked.

JanKrivanek:
@ViktorHofer I very appreciate the efforts! Be aware this is not a blocker for us currently - so no need to work under pressure!

MichaelSimons:
While I think this issue should be fixed, I don't think the msbuild issue requires it.  

[...]

PR #36272, opened 2023-10-19
Fix: Validate attribute type defines when excluded

ViktorHofer:
Fixes half of https://github.com/dotnet/sdk/issues/33388
Regressed as part of .NET 8 Preview 2: https://github.com/dotnet/sdk/commit/137a4ac17a10b10b661e956fc1e56e280c4fd5c8

When attributes are configured to be excluded from APICompat compatibility validation, the attribute definitions themselves were also incorrectly excluded, not just their usage on members.

This fixes that by storing the attribute exclusion filter in a separate field in the rule settings.

Added a test that excercises that behavior.

### DIFF
--- a/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
+++ b/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
@@ -75,15 +75,9 @@ namespace Microsoft.DotNet.ApiCompatibility.Rules
             IList<CompatDifference> differences)
         {
             // Filter out attributes that should be excluded based on the settings.
-            if (_settings.AttributeDataSymbolFilter is not null)
-            {
-                left = left.Where(attributeData => attributeData.AttributeClass is not null && _settings.AttributeDataSymbolFilter.Include(attributeData.AttributeClass)).ToImmutableArray();
-                right = right.Where(attributeData => attributeData.AttributeClass is not null && _settings.AttributeDataSymbolFilter.Include(attributeData.AttributeClass)).ToImmutableArray();
-            }
-
-            // See discussion in https://github.com/dotnet/sdk/pull/27774. ApiCompat intentionally considers non excluded attribute arguments.
-            left = left.ExcludeNonVisibleOutsideOfAssembly(_settings.SymbolFilter, excludeWithTypeArgumentsNotVisibleOutsideOfAssembly: false);
-            right = right.ExcludeNonVisibleOutsideOfAssembly(_settings.SymbolFilter, excludeWithTypeArgumentsNotVisibleOutsideOfAssembly: false);
+            // ApiCompat intentionally considers non excluded attribute arguments. See discussion in https://github.com/dotnet/sdk/pull/27774.
+            left = left.ExcludeNonVisibleOutsideOfAssembly(_settings.AttributeDataSymbolFilter, excludeWithTypeArgumentsNotVisibleOutsideOfAssembly: false);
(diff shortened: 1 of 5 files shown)


src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/AttributesMustMatch.cs
```diff
@@ -74,6 +74,13 @@ public AttributesMustMatch(IRuleSettings settings, IRuleRegistrationContext cont
             ImmutableArray<AttributeData> right,
             IList<CompatDifference> differences)
         {
+            // Filter out attributes that should be excluded based on the settings.
+            if (_settings.AttributeDataSymbolFilter is not null)
+            {
+                left = left.Where(attributeData => attributeData.AttributeClass is not null && _settings.AttributeDataSymbolFilter.Include(attributeData.AttributeClass)).ToImmutableArray();
```
ericstj, line 80: Is there a need to realize this as an ImmutableArray?  I see later on we create another ImmutableArray in ExcludeNonVisibleOutsideOfAssembly.  One way we could avoid this double-pass would be to create an aggregate filter for attributes.  So make the attribute filter a CompositeSymbolFilter of the DocIDFilter and the regular SymbolFilter.

ericstj: In other words, it's good to have a filter dedicated to filtering attributes.  Instead of making that filter *just* for exclusions, make it for both visibility and exclusions.  That saves you from doing multiple filtering passes.

src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompatibility/Rules/RuleSettings.cs
```diff
@@ -21,6 +21,11 @@ public interface IRuleSettings
         /// </summary>
         IEqualityComparer<ISymbol> SymbolEqualityComparer { get; }
 
+        /// <summary>
+        /// The attribute data symbol filter.
```
ericstj, line 25: Nit
```suggestion
        /// The attribute data symbol filter.  This filter is responsible for all filtering of the application of attributes, including visibility.
```

src/Tests/Microsoft.DotNet.ApiCompatibility.Tests/Rules/AttributesMustMatchTests.cs
```diff
@@ -1327,11 +1326,10 @@ public void EnsureDiagnosticIsReported(string leftSyntax, string rightSyntax, Co
             using TempDirectory root = new();
             string filePath = Path.Combine(root.DirPath, "exclusions.txt");
             File.Create(filePath).Dispose();
-            TestRuleFactory s_ruleFactory = new((settings, context) => new AttributesMustMatch(settings, context));
             IAssemblySymbol left = SymbolFactory.GetAssemblyFromSyntax(leftSyntax);
             IAssemblySymbol right = SymbolFactory.GetAssemblyFromSyntax(rightSyntax);
             ApiComparer differ = new(s_ruleFactory);
-            differ.Settings.SymbolFilter = GetAccessibilityAndAttributeSymbolFiltersAsComposite(filePath);
+            differ.Settings.AttributeDataSymbolFilter = new DocIdSymbolFilter(new string[] { filePath });
```
ericstj, line 1332: nit: for these tests you could actually keep this as
```c#
differ.Settings.AttributeDataSymbolFilter = GetAccessibilityAndAttributeSymbolFiltersAsComposite(filePath);
```
Which might better match the way this is used in the task/app.

ericstj (approved):
Product change looks good, couple of nits on docs/test

PR #36279, opened 2023-10-19
Fix: Emit attribute type defines when excluded

ViktorHofer:
Fixes half of https://github.com/dotnet/sdk/issues/33388 Regressed as part of .NET 8 Preview 2: https://github.com/dotnet/sdk/commit/137a4ac17a10b10b661e956fc1e56e280c4fd5c8

When attributes are configured to be excluded from GenAPI, the attribute definitions themselves were also incorrectly excluded, not just their usage on members.

Added a test to verify this behavior. I tested the changes locally and @MichaelSimons is verifying this in SBRP as well.

### DIFF
--- a/src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
+++ b/src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
@@ -277,7 +277,7 @@ namespace Microsoft.DotNet.GenAPI
 
         private SyntaxNode GenerateForwardedTypeAssemblyAttributes(IAssemblySymbol assembly, SyntaxNode compilationUnit)
         {
-            foreach (INamedTypeSymbol symbol in assembly.GetForwardedTypes().Where(_attributeDataSymbolFilter.Include))
+            foreach (INamedTypeSymbol symbol in assembly.GetForwardedTypes().Where(_symbolFilter.Include))
             {
                 if (symbol.TypeKind != TypeKind.Error)
                 {
--- a/src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
+++ b/src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
@@ -27,7 +27,7 @@ namespace Microsoft.DotNet.GenAPI
         private readonly ILog _logger;
         private readonly TextWriter _textWriter;
         private readonly ISymbolFilter _symbolFilter;
-        private readonly ISymbolFilter? _attributeDataSymbolFilter;
+        private readonly ISymbolFilter _attributeDataSymbolFilter;
         private readonly string? _exceptionMessage;
         private readonly bool _includeAssemblyAttributes;
         private readonly AdhocWorkspace _adhocWorkspace;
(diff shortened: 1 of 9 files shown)


src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/SyntaxNodeExtensions.cs
```diff
@@ -15,10 +16,15 @@ internal static class SyntaxNodeExtensions
...
         public static SyntaxNode AddMemberAttributes(this SyntaxNode node,
             SyntaxGenerator syntaxGenerator,
+            ISymbol member,
             ISymbolFilter symbolFilter,
-            ISymbol member)
+            ISymbolFilter? attributeDataSymbolFilter)
         {
-            foreach (AttributeData attribute in member.GetAttributes().ExcludeNonVisibleOutsideOfAssembly(symbolFilter))
+            ImmutableArray<AttributeData> memberAttributes = member.GetAttributes()
+                .ExcludeWithFilter(attributeDataSymbolFilter)
```
ericstj, line 24: Same feedback as the other PR - we can combine filters here to save double filtering.

src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
```diff
@@ -272,7 +279,15 @@ private SyntaxNode GenerateAssemblyAttributes(IAssemblySymbol assembly, SyntaxNo
 
         private SyntaxNode GenerateForwardedTypeAssemblyAttributes(IAssemblySymbol assembly, SyntaxNode compilationUnit)
         {
-            foreach (INamedTypeSymbol symbol in assembly.GetForwardedTypes().Where(_symbolFilter.Include))
+            ImmutableArray<INamedTypeSymbol> forwardedTypesNamespaceSymbols = assembly.GetForwardedTypes();
+
+            // Support filtering out type forward attributes.
+            if (_attributeDataSymbolFilter is not null)
+            {
+                forwardedTypesNamespaceSymbols = forwardedTypesNamespaceSymbols.Where(_attributeDataSymbolFilter.Include).ToImmutableArray();
```
ericstj, line 287: Why are we filtering type-forwards for with the attribute filter?   That filter should only apply to the **application** of attributes.

src/Compatibility/Microsoft.DotNet.ApiSymbolExtensions/AttributeDataExtensions.cs
```diff
@@ -33,13 +21,36 @@ public static class AttributeDataExtensions
             attributes.Where(attribute => attribute.IsVisibleOutsideOfAssembly(symbolFilter, excludeWithTypeArgumentsNotVisibleOutsideOfAssembly)).ToImmutableArray();
 
         /// <summary>
-        /// Checks if an <see cref="AttributeData"/> has <see cref="INamedTypeSymbol"/> arguments that point to a <see cref="Type"/> that isn't visible outside of the containing assembly.
+        /// Excludes <see cref="AttributeData"/> based on a passed in filter.
         /// </summary>
-        public static bool HasTypeArgumentsNotVisibleOutsideOfAssembly(this AttributeData attributeData, ISymbolFilter symbolFilter) =>
+        /// <returns></returns>
```
ericstj, line 26: I think all this diff can go away if we capture the attribute filter as a composite filter.

src/Compatibility/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
ericstj: If anything, I'd expect typeforwards to be filtered with the Exclude API list, not the Exclude Attribute list.

ericstj (approved):
Thank you for addressing the feedback

PR #36284, opened 2023-10-19
[release/8.0.1xx] GenAPI - Emit attribute type defines when excluded

ViktorHofer:
Manual backport of https://github.com/dotnet/sdk/pull/36279
Regression from .NET 8 Preview 2: https://github.com/dotnet/sdk/commit/137a4ac17a10b10b661e956fc1e56e280c4fd5c8

## Customer Impact
_GenAPI is a global tool / MSBuild task that scans an assembly's public API and generates a C# file out of that information_

This roslyn based GenAPI tooling is currently only used by our source build partners and **isn't publicly shipping**. Our partners let us know that when using the opt-in "attribute exclusion" feature, type definitions of those attributes are missing from the generated compiled source file. This fixes the regression introduced during the .NET 8 development lifecycle.

## Testing
Unit test added and compatibility checks performed locally in various repositories (dotnet/source-build-reference-packages).

## Risk
Low - The fix is isolated, well tested, the feature is opt-in and the overall tool isn't publicly shipping (yet).

### DIFF
--- a/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
+++ b/src/GenAPI/Microsoft.DotNet.GenAPI/CSharpFileBuilder.cs
@@ -27,6 +27,7 @@ namespace Microsoft.DotNet.GenAPI
         private readonly ILog _logger;
         private readonly TextWriter _textWriter;
         private readonly ISymbolFilter _symbolFilter;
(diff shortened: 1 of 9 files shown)


ericstj (approved):
@artl93 this is the GenAPI half of the attribute filtering issue we have mentioned is coming.

artl93 (approved):
M2 approved.

PR #36329, opened 2023-10-20
Log SDK Culture mismatch message at Low importance to unblock SDK migration

baronfel:
When previewing Substrate's upgrade to .NET 8 SDK @rainersigwald found that even Messages that _look_ like MSBuild Warnings or Errors would get flagged by Substrate's build as Errors. Unfortunately, since Messages are not controllable/suppressible with NoWarn in the way that Warnings and Errors are, this led to an unavoidable build break on update. While we are investigating a proper fix for this build tooling's behavior, we can also be a bit defensive in the SDK and emit the message as a Low importance, which causes this tooling to not flag it. Roslyn does similar for some IDe diagnostics.

## Customer Impact

Customers that are inspecting Message lines for text that looks like/behaves like an MSBuild diagnostic will detect and raise an un-suppressible error.

## Testing

Automated tests updated and passing

## Risk

**Low** - this is just changing the Importance of an emitted message, not the logic of when that message is fired.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
+++ b/src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
@@ -161,7 +161,7 @@ namespace Microsoft.NET.Build.Tasks.UnitTests
             var invalidContextWarnings = engine.Warnings.Where(msg => msg.Code == "NETSDK1188");
             invalidContextWarnings.Should().HaveCount(shouldHaveWarnings ? 1 : 0);
 
-            var invalidContextMessages = engine.Messages.Where(msg => msg.Code == "NETSDK1188");
+            var invalidContextMessages = engine.Messages;
             invalidContextMessages.Should().HaveCount(shouldHaveWarnings ? 0 : 1);
 
         }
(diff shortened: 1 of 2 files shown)


baronfel:
We figured out that it was the Importance in the end, thanks to @rainersigwald and @jaredpar.

rainersigwald (approved):

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
```diff
@@ -161,7 +161,7 @@ public void It_warns_on_invalid_culture_codes_of_resources(string tfm, bool shou
             var invalidContextWarnings = engine.Warnings.Where(msg => msg.Code == "NETSDK1188");
             invalidContextWarnings.Should().HaveCount(shouldHaveWarnings ? 1 : 0);
 
-            var invalidContextMessages = engine.Messages.Where(msg => msg.Code == "NETSDK1188");
+            var invalidContextMessages = engine.Messages.Where(msg => msg.Code == "NETSDK1187" && msg.Importance == MessageImportance.Low);
```
dsplaisted, line 164: The error code here changed.  Was the test broken before?

dsplaisted (approved):
What's the impact of reducing the importance of the message logged to low in normal scenarios?

src/Tasks/Microsoft.NET.Build.Tasks.UnitTests/GivenAResolvePackageAssetsTask.cs
baronfel: This was a bad copy/paste on my part - I'll fix that now.

jaredpar:
> We figured out that it was the Importance in the end, thanks to @rainersigwald and @jaredpar.

That really was the important detail

baronfel:
Approved via email - good to merge once green.

PR #36340, opened 2023-10-21
[release/8.0] Fix emsdk dep for stable version

lewing:
Updated variant of https://github.com/dotnet/sdk/pull/35569 with the proper darc feed, the dependency is incoherent until https://github.com/dotnet/sdk/pull/33126 lands but correct


lewing:
@dotnet/source-build-internal  please take a look at the failure here

lewing:
same issue as https://github.com/dotnet/runtime/pull/93801

darc needs stable package names, sdk wants the workload package names to change for each version band and sourcebuild breaks on any change

MichaelSimons:
If source-build isn't flagging this as a prebuilt, I think it is correct to drop the source-build adornment from the dependency.  If it were reported as a prebuilt then adding a dependency on the intermediate as done in https://github.com/dotnet/runtime/pull/93801/files/be0554c70b9444795ac78c5f617bcddce0b922e6..8c1faf561b8a305661cda047cf0298250a0c919a would be the correct was to resolve this.

MichaelSimons (approved):

PR #36357, opened 2023-10-23
Fixes O(2^N) problem in HotReloadProfileReader #36356

### LINKED ISSUE #36356: HotReloadProfileReader has O(2^N) performance issue
### Describe the bug
We have a large solution with many dependent projects and have been unable to use `dotnet watch run` because of how long it takes to startup on some of our projects. When looking into the dotnet watch source further and adding some debugging output, I noticed this is because the InferHotReloadProfile() method does a BFS search on the project hierarchy and repeatedly enqueues each referenced project (and all descendant references) which cause the queue grow exponentially. We have over 100 projects in our solution, and one of our projects (which has about 45 projects referenced) causes the queue to grow to a max size of over 10M entries and iterates over a total of 59M items. This takes over 30 seconds just in the call to InferHotReloadProfile.

### To Reproduce

https://github.com/coolkev/DotnetWatchFail/blob/main/DotnetWatchFail/Program.cs
I created this project to demonstrate the problem. It just creates 26 projects (A-Z.csproj) with each one referencing the next. Obviously this is not how our solution is setup, but it accurately demonstrates the problem because the InferHotReloadProfile method takes about 25s to run on my machine for this very simple solution with no real code.

Running the project the first time will generate the 26 projects and run a variant of the original InferHotReloadProfile method vs an optimized version. You can also run using the `dotnet watch run --project A\A.csproj` command to see how slow it is.

image

coolkev:
Fixes #36356

### DIFF
--- a/src/BuiltInTools/dotnet-watch/HotReload/HotReloadProfileReader.cs
+++ b/src/BuiltInTools/dotnet-watch/HotReload/HotReloadProfileReader.cs
@@ -16,6 +16,8 @@ namespace Microsoft.DotNet.Watcher.Tools
 
             ProjectInstance? aspnetCoreProject = null;
 
+            var visited = new HashSet<ProjectGraphNode>();
+
             while (queue.Count > 0)
             {
                 var currentNode = queue.Dequeue();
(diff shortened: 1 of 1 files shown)


coolkev:
@dotnet/domestic-cat this is ready for review.

baronfel:
Hot Reload is under @tmat - can you take a look when you have time?

akoeplinger:
@tmat ping

tmat (approved):

tmat:
@coolkev Thanks!

PR #36387, opened 2023-10-23
add packagesprops template

### LINKED ISSUE #36367: dotnet new template for  `Directory.Packages.props` file

Add item template for `Directory.Packages.props` file

```sh
dotnet new pkgprops

dotnet new packageprops
```

WeihanLi:
fixes #36367 

add `packagesprops` template

- [ ] The naming may need to be confirmed
- [ ] The tags and description drafted, also need to be confirmed
- [ ] Should we support `inherit`? does the `inherit` template look good? I assume the root `Directory.Packages.props` should not `inherit` and should have the `<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>` property defined


template_feed/Microsoft.DotNet.Common.ItemTemplates/content/PackagesProps/Directory.Packages.props
```diff
@@ -0,0 +1,11 @@
+<Project>
+<!--#if inherit -->
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Packages.props', '$(MSBuildThisFileDirectory)../'))"
```
jeffkl, line 3: I think the root `Directory.Packages.props` should not try to inherit from any parents.  Would it be possible to add an option to the template so by default it creates the `Directory.Packages.props` with just the PropertyGroup and then with some command-line argument it makes a `Directory.Packages.props` with this inheritance?

baronfel: @jeffkl that's all already wired up with the template symbol `inherit` defined in the [template.json](https://github.com/dotnet/sdk/pull/36387/files#diff-b5221c4a707625fac869d1c5ba3fa00b20f62de0e6657ffdaf4497bdcd157434R21-R26) - currently the default invocation of this template (much like the Directory.Build.props/targets templates we also have in the SDK) implicitly passes `inherit=false`, but a user can pass `--inherit true` on the CLI and then the Import will be added.

jeffkl: Perfect, thanks for clarifying.

jeffkl (approved):

baronfel:
@jeffkl this is currently targeting 9.0 (because of main branch) - would NuGet want this to ship earlier? If so we can backport to release/8.0.2xx after merge.

jeffkl:
> @jeffkl this is currently targeting 9.0 (because of main branch) - would NuGet want this to ship earlier? If so we can backport to release/8.0.2xx after merge.

If its really easy, we might as well have it for 8.x but I would not say its necessary.  This change (https://github.com/NuGet/NuGet.Client/pull/5456) makes it so you don't even need `<ManagePackageVersionsCentrally>` anymore so creating a `Directory.Packages.props` is that much easier.  But that change won't ship in 8.x either.  So I'll leave it up to you if you want to port this template to 8.x.

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/PackagesProps/Directory.Packages.props
```diff
@@ -0,0 +1,11 @@
...
+          Condition="$([MSBuild]::GetPathOfFileAbove('Directory.Packages.props', '$(MSBuildThisFileDirectory)../')) != ''" />
+<!--#else-->
+  <PropertyGroup>
+    <!-- Enable central package management, https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management -->
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+<!--#endif-->
```
baronfel, line 10: One suggestion I have: consider putting an empty `ItemGroup` in this generated file - `PackageVersion` items will have to go into one anyway, and so we know a user is going to end up needing one.

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/PackagesProps/.template.config/template.json
```diff
@@ -0,0 +1,28 @@
...
+  "author": "Microsoft",
+  "classifications": [
+    "MSBuild",
+    "packages",
+    "props",
+    "CPM"
+  ],
+  "name": "MSBuild Directory.Packages.props file",
+  "generatorVersions": "[1.0.0.0-*)",
+  "description": "An empty Directory.Packages.props file which can be used to specify package version centrally",
```
baronfel, line 12:
```suggestion
  "description": "An empty Directory.Packages.props file which can be used to specify package versions centrally",
```

baronfel:
@WeihanLi two minor comments, but other than that I'm quite happy with this. Thank you for contributing it! I can't access the builds anymore to check the tests failures, so I'll re-start that now.

NatElkins:
Just want to chime in and say I would really appreciate if this were backported to 8.x!!

baronfel:
> > @jeffkl this is currently targeting 9.0 (because of main branch) - would NuGet want this to ship earlier? If so we can backport to release/8.0.2xx after merge.
> 
> If its really easy, we might as well have it for 8.x but I would not say its necessary. This change ([NuGet/NuGet.Client#5456](https://github.com/NuGet/NuGet.Client/pull/5456)) makes it so you don't even need `<ManagePackageVersionsCentrally>` anymore so creating a `Directory.Packages.props` is that much easier. But that change won't ship in 8.x either. So I'll leave it up to you if you want to port this template to 8.x.

@jeffkl what's the mechanism that NuGet's planning to use to prevent that change from shipping in 8.x at all? I'm not familiar with your branching strategy - are you like MSBuild where your `main` goes into the 'next' SDK regardless of if it's a feature band like 8.0.200, or are you like...well, the SDK itself where `main` goes into 9.x at this point? Or something else?

baronfel:
@WeihanLi it looks like the two failing tests are just missing the new template name being added to the expected suggestion lists - that should be a pretty straightforward change and then we can merge this!

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/PackagesProps/Directory.Packages.props
WeihanLi: Do we want to mention the `PackageVersion` in a commented code likes follows code?
and should we added it for non `inherited` template only? Added for both for now

```xml
  <ItemGroup>
    <!-- PackageVersion should be defined here -->
  </ItemGroup>
```

NatElkins:
@baronfel Would love to see this get in (if possible)!

baronfel:
Thanks for fixing those tests @WeihanLi - happy to merge this now :)

WeihanLi:
My pleasure, and thanks very much for the help

PR #36437, opened 2023-10-24
upgrade stage 0 and unwind a few steps that were done for the 9.0 upg…

marcpopMSFT:
…rade


marcpopMSFT:
@dotnet/source-build-internal I'm not sure why source build still sees 8.0 prebuilts. Since this is main, should I add back the prebuilt exclusion for now and track that for removal later?

mthalman:
> @dotnet/source-build-internal I'm not sure why source build still sees 8.0 prebuilts. Since this is main, should I add back the prebuilt exclusion for now and track that for removal later?

There are still projects that are referencing `net8.0`:
* Microsoft.DotNet.ApiCompat.Task
* Microsoft.DotNet.PackageValidation
* Microsoft.DotNet.ApiSymbolExtensions

Those are the ones bringing in these ref packs.

marcpopMSFT:
@ericstj I believe your team owns those. Did you want to upgrade those to target 9 or should I add back the prebuilt exclusion for now?

ericstj:
So the reason those projects target an older framework is so that the `Microsoft.DotNet.ApiCompat.Task` package can be used in older SDKs.  
Should we condition the TargetFramework used by these so that during Source Build it's set to latest but outside sourcebuild is `net8.0`?

ViktorHofer:
> So the reason those projects target an older framework is so that the Microsoft.DotNet.ApiCompat.Task package can be used in older SDKs.

Yes, we should not upgrade those TFMs unless when building from source. net8.0 is the minimum supported TFM as the .NET 6 and .NET 7 SDKs are out-of-support when .NET 9 ships.

> Should we condition the TargetFramework used by these so that during Source Build it's set to latest but outside source build is net8.0? @ViktorHofer

Either that or we could introduce `net9.0` alongside `net8.0` and make the build targets aware of that. See https://github.com/dotnet/sourcelink/pull/1119/files#diff-9b615ac9f5c46b031cbc4346e1ee7927cbc5f7633c7c2999d919fa85a45e0e2fR5-R6 as an example.

mthalman:
@ViktorHofer - Now the problem is that `NetToolCurrent` is still set to `net8.0` in Arcade. I'm thinking the best thing to do is allow these as prebuilts for now. The flow to installer should still pass as we have these two ref packs already included in the prebuilts tarball. Once it flows to the VMR, we can work to remove them from the prebuilts tarball. That will require a patch to set `NetToolCurrent` to `net9.0` in sdk. Then, once Arcade updates `NetToolCurrent` to `net9.0` and flows to sdk and then sdk flows to VMR, we can remove the patch.

@MichaelSimons - Let me know your thoughts.

MichaelSimons:
> @MichaelSimons - Let me know your thoughts.

I am fine with allowing these prebuilts for now.  Instead of adding a patch, I would suggest fixing arcade right away.  Adding a patch doesn't feel like it is adding much value.

mthalman:
> I am fine with allowing these prebuilts for now. Instead of adding a patch, I would suggest fixing arcade right away. Adding a patch doesn't feel like it is adding much value.

There's already a PR for arcade: https://github.com/dotnet/arcade/pull/14108. But that's a bit of a process. Need to get the sdk flowed to the other repos first, then we can update arcade. Don't know the status of the sdk flow.

MichaelSimons:
> There's already a PR for arcade: [dotnet/arcade#14108](https://github.com/dotnet/arcade/pull/14108). But that's a bit of a process. Need to get the sdk flowed to the other repos first, then we can update arcade. Don't know the status of the sdk flow.

Sure, but if we are already caring the prebuilt in the prebuilts tarball in source-build adding the patch is trading the prebuilts tarball for a patch.  Instead of doing more work when the build is already working, I am proposing to just let the real fix flow in.  Are we gaining substantive value by adding the patch now?

ViktorHofer:
I think we should just update NetToolCurrent under sourcebuild to mean net9.0 until Arcade changes its meaning. Let me quickly try that.

ViktorHofer (approved):

eng/restore-toolset.ps1
```diff
@@ -26,8 +26,7 @@ function InitializeCustomSDKToolset {
   InstallDotNetSharedFramework "5.0.0"
   InstallDotNetSharedFramework "6.0.0"
   InstallDotNetSharedFramework "7.0.0"
-  InstallDotNetSharedFramework "8.0.0-preview.7.23375.6"
-  InstallDotNetSharedFramework "8.0.0-rc.2.23457.7"
+  InstallDotNetSharedFramework "8.0.0-rc.2.23479.6"
```
ViktorHofer, line 29: Do we bring the 8.0.0 shared framework in for testing?

marcpopMSFT: Correct, we install each of the downlevel runtimes as that includes apphosts and lets us run apps in testing. We may not need them all at this point but we'd definitely need the 8.0 one for some set of tests.

ViktorHofer: Cool. It would be good to follow-up and discuss the ones that are already out-of-support: anything lower than 6.0.0. I think it sends a weird signal to our customers to still test against out-of-support runtimes.

PR #36443, opened 2023-10-24
Silence workload install logging on `dotnet --info` and related commands.

### LINKED ISSUE #36441: `dotnet --info` should not cause a worklog log to be emitted

Right now when a user runs `dotnet --info` a new workload log is emitted to $TEMP. This clutters the overall list of workload logs and makes finding useful data more difficult.

```console
$ (ls Microsoft.NET.Workload*.log).Count; dotnet --info > $null; (ls Microsoft.NET.Workload*.log).Count
146
147
```

Only commands that modify workload status should emit logs - update, repair, install, uninstall, and related commands.

### Additional context
This probably started when we added workload output to `--info` to aid debugging.

baronfel:
Fixes #36441

This creates a NullLogger which can be used in place of TimestampedFileLogger and wires it up when the codepath triggered by `dotnet --info` is used. This should result in fewer scattered files in the temp directory for workload logs.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/WorkloadInfoHelper.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/WorkloadInfoHelper.cs
@@ -56,7 +56,8 @@ namespace Microsoft.DotNet.Workloads.Workload.List
                 userProfileDir,
                 verifySignatures ?? !SignCheck.IsDotNetSigned(),
                 restoreActionConfig: restoreConfig,
-                elevationRequired: false);
+                elevationRequired: false,
+                shouldLog: false);
 
             WorkloadRecordRepo = workloadRecordRepo ?? Installer.GetWorkloadInstallationRecordRepository();
         }
--- a/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
@@ -1019,9 +1019,12 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
             PackageSourceLocation packageSourceLocation = null,
             IReporter reporter = null,
             string tempDirPath = null,
-            RestoreActionConfig restoreActionConfig = null)
+            RestoreActionConfig restoreActionConfig = null,
+            bool shouldLog = true)
         {
-            TimestampedFileLogger logger = new(Path.Combine(Path.GetTempPath(), $"Microsoft.NET.Workload_{Environment.ProcessId}_{DateTime.Now:yyyyMMdd_HHmmss_fff}.log"));
+            ISynchronizingLogger logger =
+                shouldLog ? new TimestampedFileLogger(Path.Combine(Path.GetTempPath(), $"Microsoft.NET.Workload_{Environment.ProcessId}_{DateTime.Now:yyyyMMdd_HHmmss_fff}.log"))
+                          : new NullInstallerLogger();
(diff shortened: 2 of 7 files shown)


src/Cli/dotnet/commands/dotnet-workload/WorkloadInfoHelper.cs
```diff
@@ -56,7 +56,8 @@ internal class WorkloadInfoHelper : IWorkloadInfoHelper
                 userProfileDir,
                 verifySignatures ?? !SignCheck.IsDotNetSigned(),
                 restoreActionConfig: restoreConfig,
-                elevationRequired: false);
+                elevationRequired: false,
+                shouldLog: false);
```
baronfel, line 60: This is the main way we get the benefit of the NullLogger - `dotnet --info` and `dotnet workload list` use this to get installed workload state and so we can silence both paths in one go.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -1019,9 +1019,12 @@ private void VerifyPackage(MsiPayload msiPayload)
...
             IReporter reporter = null,
             string tempDirPath = null,
-            RestoreActionConfig restoreActionConfig = null)
+            RestoreActionConfig restoreActionConfig = null,
+            bool shouldLog = true)
         {
-            TimestampedFileLogger logger = new(Path.Combine(Path.GetTempPath(), $"Microsoft.NET.Workload_{Environment.ProcessId}_{DateTime.Now:yyyyMMdd_HHmmss_fff}.log"));
+            ISynchronizingLogger logger =
+                shouldLog ? new TimestampedFileLogger(Path.Combine(Path.GetTempPath(), $"Microsoft.NET.Workload_{Environment.ProcessId}_{DateTime.Now:yyyyMMdd_HHmmss_fff}.log"))
+                          : new NullInstallerLogger();
```
baronfel, line 1027: ShoudLog makes it way to this point, where the actual decision point is.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
```diff
@@ -34,9 +35,9 @@ internal class WorkloadInstallerFactory
                 {
                     throw new InvalidOperationException(LocalizableStrings.OSDoesNotSupportMsi);
                 }
-
+                // TODO: should restoreActionConfig be flowed through to the client here as well like it is for the FileBasedInstaller below?
```
baronfel, line 38: this is unrelated to this PR, but @dsplaisted I saw a mismatch between the use of RestoreActionConfig in this codepath vs the codepath directly below - should those be synchronized? `NetSdkMsiInstallerClient.Create` has an optional `restoreActionConfig` parameter that we're not flowing at the moment, and I know RAC has been a hot topic recently.

joeloff (approved):
Let's chat about the RAC issue, otherwise, happy with the change.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallerFactory.cs
dsplaisted: Yes, there are cases for tool installation where we're not flowing command line options to a RestoreActionConfig.  If we support those command line options for workload commands it seems like we should flow them through via a RestoreActionConfig.

dsplaisted (approved):

PR #36499, opened 2023-10-26
Skip a templating test when run internally

marcpopMSFT:


joeloff (approved):
hacky, but okay :)

marcpopMSFT:
> hacky, but okay :)

Know of a better way to do this? It was failing internal codeflow with a 404 error about the feed. Not sure of a better way to detect an internal build and not sure of another way of fixing the test.

joeloff:
> > hacky, but okay :)
> 
> Know of a better way to do this? It was failing internal codeflow with a 404 error about the feed. Not sure of a better way to detect an internal build and not sure of another way of fixing the test.

Possibly pass a preprocessor definition from the YAML in the pipeline, then wrap the test inside a `#ifdef`

joeloff:
> > > hacky, but okay :)
> > 
> > 
> > Know of a better way to do this? It was failing internal codeflow with a 404 error about the feed. Not sure of a better way to detect an internal build and not sure of another way of fixing the test.
> 
> Possibly pass a preprocessor definition from the YAML in the pipeline, then wrap the test inside a `#ifdef` and set some default for when running from the commandline or IDE

PR #36621, opened 2023-11-02
Update stage0

v-wuzhai:
Update stage0 to make the dotnet9.0 option effective in the template creation command.
For example: `dotnet new console -f net9.0 -o test`


v-wuzhai:
@baronfel @marcpopMSFT Not sure if the container test failed because there is no available dotnet9 image?

v-wuzhai:
@baronfel Could you take a look at the failures here? Thanks

baronfel:
@v-wuzhai I think the containers tests do need to stay on .NET 8 TFMs until we have .NET 9 container images (and those are made in the dotnet-docker repo after SDK flows to installer). So we will have some awkwardness there for a while.

v-wuzhai:
The 9.0 template has been enabled.  When the framework target to 8.0, it shows options are invalid as the following error message, so we skip this test and wait for the 9.0 image to be available before enabling it.
```
StdErr:
Error: Invalid option(s):
-f net8.0
'net8.0' is not a valid value for -f. The possible values are:
net9.0   - Target net9.0
```

marcpopMSFT (approved):

Issue #36557, opened 2023-10-30
`dotnet tool update` gives unfriendly error and stack on not found


timheuer:
### Describe the bug
When using `dotnet tool update` (or likely other methods) and the ID is not found for the tool, a full stack unhandled exception is shown instead of a user-friendly error.

### To Reproduce
`dotnet tool update -g pwsh`

### Exceptions (if any)
Unhandled exception: Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageNotFoundException: pwsh::[*, ) is not found in NuGet feeds C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\, https://api.nuget.org/v3/index.json
   at Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageDownloader.GetMatchingVersionInternalAsync(String packageIdentifier, IEnumerable`1 packageSources, VersionRange versionRange, CancellationToken cancellationToken)
   at Microsoft.DotNet.Cli.NuGetPackageDownloader.NuGetPackageDownloader.GetBestPackageVersionAsync(PackageId packageId, VersionRange versionRange, PackageSourceLocation packageSourceLocation)
   at Microsoft.DotNet.Cli.ToolPackage.ToolPackageDownloader.<>c__DisplayClass8_0.<InstallPackage>b__0()
   at Microsoft.DotNet.Cli.TransactionalAction.Run[T](Func`1 action, Action commit, Action rollback)
   at Microsoft.DotNet.Tools.Tool.Update.ToolUpdateGlobalOrToolPathCommand.<>c__DisplayClass14_0.<Execute>b__1()
   at Microsoft.DotNet.Tools.Tool.Update.ToolUpdateGlobalOrToolPathCommand.RunWithHandlingInstallError(Action installAction)
   at Microsoft.DotNet.Tools.Tool.Update.ToolUpdateGlobalOrToolPathCommand.Execute()
   at System.CommandLine.Invocation.InvocationPipeline.Invoke(ParseResult parseResult)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, TimeSpan startupTime, ITelemetry telemetryClient)

JL03-Yue:
Thanks for reporting it

JL03-Yue:
Thanks for reporting it. Fixed in #36641

PR #36641, opened 2023-11-03
Change the format of tool exceptions and NuGet API exceptions to be graceful

JL03-Yue:
#36557 

Change the format of tool exceptions and NuGet API exceptions to be graceful without default stack dump

### DIFF
--- a/src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
+++ b/src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
@@ -1,11 +1,13 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Text.RegularExpressions;
 using System.Threading;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.ToolPackage;
 using Microsoft.DotNet.Tools;
 using Microsoft.Extensions.EnvironmentAbstractions;
+using Microsoft.TemplateEngine.Abstractions;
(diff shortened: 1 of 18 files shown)


src/Cli/dotnet/NugetPackageDownloader/NuGetPackageDownloader.cs
```diff
@@ -450,10 +450,9 @@ private IEnumerable<PackageSource> LoadNuGetSources(PackageId packageId, Package
                 throw new NuGetPackageNotFoundException(
                     string.Format(
                         LocalizableStrings.IsNotFoundInNuGetFeeds,
-                        $"{packageIdentifier}::{versionRange}",
+                        $"{packageIdentifier} of version range {versionRange}",
```
JL03-Yue, line 453: I wonder if this expression make it clearer for customers

baronfel: I might suggest changing the formatting a bit based on the kind of range:
* if the range is exact match, say `Version {versionRange} of package {packageIdentifier}`
* if the range is 'greater than or equal to', say `A version between {min} and {max} of package {packageIdentifier}`
  * this works for 'less than or equal to' as well
* if the range is 'at least', say `A version higher than {minBound} of package {packageIdentifier}`
* if the range is 'at most', say `A version less than than {maxBound} of package {packageIdentifier}`

baronfel (commented):
The format of the different ranges looks good! I think we need:

* some screenshots of the different error cases
* tests in-repo for the 'not found' cases and such so we don't regress

baronfel (approved):

Issue #36651, opened 2023-11-03
ContainerArchiveOutputPath is not effecive on new .NET 8 container support while running on Framework


wasabii:
### Describe the bug
When adding ContainerArchiveOutputPath to a .NET project, and using PublishContainer, running on a Framework version of MSBuild, no container output is generated.

### To Reproduce
Make a .NET project with ContainerArchiveOutputPath, and invoke the Publish;PublishContainer targets. Do it from 'dotnet msbuild'. Works fine. Do it from 'msbuild'. Does not work.

As such, it does not work from inside of Visual Studio.

https://github.com/dotnet/sdk/blob/67103475ad623dd9d98e81951f9d0b6d200d8074/src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImageToolTask.cs#L72

CreateNewImageToolTask, the Framework version of this, never passes --archiveoutputpath to containerize.dll. It looks like containerize.dll supports --archiveoutputpath. But it never receives it.

.NET 8.0 RC 2

wasabii:
@baronfel

baronfel:
Thanks for the spot - we did indeed miss this parameter from the full-framework build.

wasabii:
@baronfel  Assuming this did not make 8.0 release?

baronfel:
Not GA, no :(  I'm trying to get it into a servicing release as well.

wasabii:
For those looking for a 'quick' fix, I threw together this targets file which overwrites the PublishContainer target for running on NET Framework. Probably some bugs in here too, but it works for me.

```
<Project>

    <Target Name="PublishContainer" DependsOnTargets="$(PublishContainerDependsOn)" Condition=" '$(IsPublishable)' == 'true' ">
        <PropertyGroup Condition="'$(DOTNET_HOST_PATH)' == ''">
            <DotNetHostDirectory>$(NetCoreRoot)</DotNetHostDirectory>
            <DotNetHostFileName>dotnet</DotNetHostFileName>
            <DotNetHostFileName Condition="'$(OS)' == 'Windows_NT'">dotnet.exe</DotNetHostFileName>
        </PropertyGroup>

        <ItemGroup>
            <_Args Include="&quot;$(ContainerizeFolder)containerize.dll&quot;" />
            <_Args Include="&quot;$(PublishDir.TrimEnd('\'))&quot;" />
[...]

PR #36652, opened 2023-11-03
add missing --archiveoutputpath option to .NET Framework container tool task

baronfel:
Fixes https://github.com/dotnet/sdk/issues/36651

I'd love to have a better idea of how we could enforce remembering these properties in both task variants.


wasabii:
Auto generate both sides of the picture call using some DTO interfacey thing which is shared between the projects.

Or unit tests.

MichalPavlik (approved):

PR #36669, opened 2023-11-06
Add option to ignore baseline TFMs

### LINKED ISSUE #22901: [Package Validation] We should support excluding tfms when running the validation

ViktorHofer:
Fixes https://github.com/dotnet/sdk/issues/22901

Add frontend option to allow ignoring TFMs when performing baseline package validation.

MSBuild item: `@(PackageValidationBaselineFrameworkToIgnore)`
CLI option: `--baseline-package-frameworks-to-ignore`

See https://github.com/dotnet/runtime/pull/94432 which reacts to this change in runtime.

### DIFF
--- a/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Validators/BaselinePackageValidator.cs
+++ b/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Validators/BaselinePackageValidator.cs
@@ -28,9 +28,15 @@ namespace Microsoft.DotNet.PackageValidation.Validators
 
             ApiCompatRunnerOptions apiCompatOptions = new(options.EnableStrictMode, isBaselineComparison: true);
 
-            // Iterate over all target frameworks in the package.
             foreach (NuGetFramework baselineTargetFramework in options.BaselinePackage.FrameworksInPackage)
             {
+                // Skip target frameworks excluded from the baseline package.
+                if (options.BaselinePackageFrameworksToIgnore is not null &&
+                   options.BaselinePackageFrameworksToIgnore.Contains(baselineTargetFramework.GetShortFolderName()))
+                {
+                    continue;
+                }
+
                 // Retrieve the compile time assets from the baseline package
                 IReadOnlyList<ContentItem>? baselineCompileAssets = options.BaselinePackage.FindBestCompileAssetForFramework(baselineTargetFramework);
                 if (baselineCompileAssets != null)
--- a/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
+++ b/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompat.Shared/ValidatePackage.cs
@@ -30,7 +30,8 @@ namespace Microsoft.DotNet.ApiCompat
             string? baselinePackagePath,
             string? runtimeGraph,
             IReadOnlyDictionary<NuGetFramework, IEnumerable<string>>? packageAssemblyReferences,
-            IReadOnlyDictionary<NuGetFramework, IEnumerable<string>>? baselinePackageAssemblyReferences)
+            IReadOnlyDictionary<NuGetFramework, IEnumerable<string>>? baselinePackageAssemblyReferences,
+            string[]? baselinePackageFrameworksToIgnore)
         {
             // Initialize the service provider
             ApiCompatServiceProvider serviceProvider = new(logFactory,
@@ -70,7 +71,8 @@ namespace Microsoft.DotNet.ApiCompat
                     enableStrictMode: enableStrictModeForBaselineValidation,
                     enqueueApiCompatWorkItems: runApiCompat,
                     executeApiCompatWorkItems: false,
-                    baselinePackage: Package.Create(baselinePackagePath, baselinePackageAssemblyReferences)));
+                    Package.Create(baselinePackagePath, baselinePackageAssemblyReferences),
+                    baselinePackageFrameworksToIgnore));
             }
 
             if (runApiCompat)
--- a/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
+++ b/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompat.Task/ValidatePackageTask.cs
@@ -118,6 +118,11 @@ namespace Microsoft.DotNet.ApiCompat.Task
         /// </summary>
         public ITaskItem[]? BaselinePackageAssemblyReferences { get; set; }
 
+        /// <summary>
+        /// A set of target frameworks to ignore from the baseline package.
+        /// </summary>
+        public string[]? BaselinePackageFrameworksToIgnore { get; set; }
+
         public override bool Execute()
         {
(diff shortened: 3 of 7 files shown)


src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Validators/BaselinePackageValidator.cs
```diff
@@ -28,9 +28,15 @@ public void Validate(PackageValidatorOption options)
 
             ApiCompatRunnerOptions apiCompatOptions = new(options.EnableStrictMode, isBaselineComparison: true);
 
-            // Iterate over all target frameworks in the package.
             foreach (NuGetFramework baselineTargetFramework in options.BaselinePackage.FrameworksInPackage)
             {
+                // Skip target frameworks excluded from the baseline package.
+                if (options.BaselinePackageFrameworksToIgnore is not null &&
+                   options.BaselinePackageFrameworksToIgnore.Contains(baselineTargetFramework.GetShortFolderName()))
```
ericstj, line 35: Should we handle casing and version format differences?  I can imagine folks doing "net462" vs "net4.6.2".  Perhaps if you stored these as parsed `NuGetFramework` values in the hashset it would take care of that difference?

Additionally - should we warn the user if they specified a `BaselinePackageFrameworksToIgnore` that wasn't in the package?

carlossanlop: > "net462" vs "net4.6.2"

Would regex be useful here?

ViktorHofer: > Should we handle casing and version format differences? I can imagine folks doing "net462" vs "net4.6.2". Perhaps if you stored these as parsed NuGetFramework values in the hashset it would take care of that difference?

I considered parsing the string up-front but decided against it for different reasons. I think it's reasonable to expect the exact string that is defined in the `TargetFramework` / `TargetFrameworks` property.

> Additionally - should we warn the user if they specified a BaselinePackageFrameworksToIgnore that wasn't in the package?

This is intended to be a global setting, defined repository-wide. I wouldn't warn for TFMs that aren't part of a package.

ericstj: > I considered parsing the string up-front but decided against it for different reasons

Why was that?  I see us dealing with the same type in other components at this level, for example https://github.com/dotnet/sdk/blob/f9e6f3190aaa9bd6ceaf44ddc3df54cc4bd847ab/src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Package.cs#L81

I was imagining to fix this we'd simply make `BaselinePackageFrameworksToIgnore` a `HashSet<NuGetFramework>` and parse them on construction.  Then you can use equality check built into `NuGetFramework` to handle the subtle differences in casing and version strings that NuGet users expect to be handled.  Feels to me like requiring an exact match of the string is unnecessary fragility.

ViktorHofer: > Why was that? I see us dealing with the same type in other components at this level, for example

We don't parse an arbitrary string into a `NuGetFramework` object anywhere in PackageValidation today. The code that you linked to is a `NuGetFramework` object that we get when inspecting the package assets. The reason why I didn't want to parse a raw string was the following feedback that I got from @zivkan:

> `NuGetFramework.Parse` only works when you know the string is a parseable string. But the VS Extensibility's MSBuild SDK style VS SDK will use something like `<TargetFramework>vs17</TargetFramework>`, so it really won't work if that gets off the ground

For TFM aliases to work (which afaik isn't yet thought through in terms of how they would be represented in packages) you would either need to call `NuGetFramework.ParseComponents(TFM, TPM)` or do a direct string match.

zivkan: There's no immediate plan to have packages support TFM aliases (see the "pack changes" and  "extending pack" section in [the feature spec](https://github.com/NuGet/Home/pull/12124)). Personally, I struggle to imagine how it could be technically feasible. NuGet's current design is that these package subdirectories are parsable names, so that'd be a big change that would need backwards compatibility for packages that don't use TFM alias features. I also don't think it's unreasonable to tell VS extension developers that they need to create multiple packages, and maybe that's what the pack implementation of projects using TFM aliases will end up being, if it ever gets supported.

So, I definitely think that avoiding parsing project TargetFramework values is a good idea ([another datapoint](https://github.com/dotnet/project-system/blob/293fd6050cc3d6694ef6f95d4331153fed2f5b2b/docs/repo/coding-conventions.md?plain=1#L53)). However, assuming that NuGet package conventions will end up changing TFM directories to something non-parsable is making a bet on something that may or may not ever happen.

ericstj (approved):
Per discussion - ok to leave this as a string comparison but make it clear to user that it's a literal string comparison in the user-facing description of the option.

src/Compatibility/ApiCompat/Microsoft.DotNet.PackageValidation/Validators/BaselinePackageValidator.cs
ericstj: Here's where MSBuild parses NuGetFramework : https://github.com/dotnet/msbuild/blob/3b59113aa9333f52efc4ab6bf96d713d05d604d1/src/Build/Utilities/NuGetFrameworkWrapper.cs#L47
And here are the places where the SDK does it: https://github.com/search?q=repo%3Adotnet%2Fsdk+%22NuGetFramework.Parse%22&type=code - including many cases where they're directly parsing user input.

Given we're _asking the user_ to provide a string that _does_ correlate to a `NuGetFramework` in the package - and not actually plumbing any connection with a `TargetFramework` value in the project or any other potentially aliased value-- I still believe that it's better to parse this user provided string and compare it with the rules that NuGet has for frameworks rather than literal string matching - but it's not the end of the world if we don't.  Just makes the new option a little less usable without feedback when folks get it wrong.

PR #35939, opened 2023-10-09
[#17005] Make Mistyping Errors Gentler

Markliniubility:
#17005
Result from this PR:
pic

Changed a part of the message from error output into normal output as error output is always red.

Issue #17005, opened 2021-04-19
Make mistyping errors gentler


KathleenDollard:
In the following, make everything except the error (the first line) light grey

![image](https://user-images.githubusercontent.com/5844318/115151354-5d067000-a021-11eb-824e-669e99751bfb.png)

Markliniubility:
This looks interesting. I would like to work on it.

Forgind:
Resolved by #35939 and #36670

PR #36670, opened 2023-11-06
[#17005] Make Mistyping Errors Gentler (#35939)

Forgind:
This is a manual backport of #35939.

(There was a merge conflict, so automatic backporting failed.)

/cc: @Markliniubility (original author)


Markliniubility:
Thanks for the note. I will base my changes on the "release/8.0.1xx" branch in the future!

marcpopMSFT (approved):

Forgind:
> Thanks for the note. I will base my changes on the "release/8.0.1xx" branch in the future!

Yeah; I think the SDK's branching is a bit weird...main ships in like November, but since we have to build first, current main is the version that will ship next year. Our release/*xx branches ship quarterly, but since those also change, it takes a little extra work to figure out which version is actually the right next one to ship. Always happy to help with retargeting as necessary 🙂

PR #36671, opened 2023-11-06
accept bare version as NuGet exact version for .NET tools

JL03-Yue:
**Customer Impact**

Per user feedback in #35566, he change, however, disable users from installing unlisted tools. This feature was added in #36021 if specifying exact match. However, since the exact match refers to input `[version]`. (referred to [NuGet package version reference](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning)) come with bracket, this caused surprise for some users using bare version as exact version match.

The intended way to fix this is to allow bare version of .NET tool to interpret as specific version, i.e.  version `5.0.0` refers to the exact version match of `5.0.0` instead of `[5.0.0, )`. To add this feature, a semvar detection is added when getting and parsing the version range in `GetVersionRange()`. Changes are validated by tests.

**Testing**
Existing unit tests passed without changes. New tested are added to validate an unlisted bar version tool can be installed with exact version match withtout bracket specified. The new testing scenarios pass with changes in the pull request. 

**Risk**
Low. This change adds a notation that was missed from breaking change without adding complexity to the behavior. 

**Original description**
Per conversation #35566, specifically [discussions on comment](https://github.com/dotnet/sdk/issues/35566#issuecomment-1788852933), this change is made to accept bare versions in .NET tools as NuGet exact version. For example, e.g. `--version 1.2.3` would be interpreted by the NuGet package downloader as `[1.2.3]`

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
@@ -1,20 +1,10 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-using System.Collections.Generic;
 using System.CommandLine;
-using System.CommandLine.Parsing;
-using System.Linq;
-using System.Text.RegularExpressions;
 using Microsoft.DotNet.Cli;
 using Microsoft.DotNet.Cli.Utils;
-using Microsoft.DotNet.NugetSearch;
-using Microsoft.DotNet.ToolPackage;
-using Microsoft.DotNet.Tools.Tool.Search;
-using Newtonsoft.Json.Linq;
 using NuGet.Versioning;
-using static System.Runtime.InteropServices.JavaScript.JSType;
 
 namespace Microsoft.DotNet.Tools.Tool.Install
 {
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
@@ -42,8 +42,7 @@ namespace Microsoft.DotNet.Tools.Tool.Install
             VersionRange versionRange = null;
 
             // accept 'bare' versions and interpret 'bare' versions as NuGet exact versions
-            string text = packageVersion?.Trim();
-            if (!string.IsNullOrEmpty(packageVersion) && IsBarePackageVersion(text))
+            if (!string.IsNullOrEmpty(packageVersion) && SemanticVersion.TryParse(packageVersion, out SemanticVersion version2))
             {
                 packageVersion = "[" + packageVersion + "]";
             }
@@ -56,15 +55,6 @@ namespace Microsoft.DotNet.Tools.Tool.Install
                         packageVersion));
             }
             return versionRange;
-        }
-
-        private static bool IsBarePackageVersion (string packageVersion)
-        {
-            // Define a regular expression pattern to match the "a.b.c" format
-            string pattern = @"^\d+\.\d+\.\d+(-[\w\d]+(\.[\w\d]+)*)?(\+[\w\d]+(\.[\w\d]+)*)?$";
-
-            // Use Regex.IsMatch to check if the input matches the pattern
-            return Regex.IsMatch(packageVersion, pattern);
         }
     }
 }
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
@@ -61,7 +61,7 @@ namespace Microsoft.DotNet.Tools.Tool.Install
         private static bool IsBarePackageVersion (string packageVersion)
         {
             // Define a regular expression pattern to match the "a.b.c" format
-            string pattern = @"^\d+\.\d+\.\d+$";
+            string pattern = @"^\d+\.\d+\.\d+(-[\w\d]+(\.[\w\d]+)*)?(\+[\w\d]+(\.[\w\d]+)*)?$";
 
             // Use Regex.IsMatch to check if the input matches the pattern
             return Regex.IsMatch(packageVersion, pattern);
--- a/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
@@ -34,7 +34,7 @@ namespace Microsoft.DotNet.Tools.Tool.Install
             // accept 'bare' versions and interpret 'bare' versions as NuGet exact versions
             if (!string.IsNullOrEmpty(packageVersion) && NuGetVersion.TryParse(packageVersion, out SemanticVersion version2))
             {
-                packageVersion = "[" + packageVersion + "]";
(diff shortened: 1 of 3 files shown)


src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
```diff
@@ -45,5 +57,14 @@ public static VersionRange GetVersionRange(this ParseResult parseResult)
             }
             return versionRange;
         }
+
+        private static bool IsBarePackageVersion (string packageVersion)
+        {
+            // Define a regular expression pattern to match the "a.b.c" format
+            string pattern = @"^\d+\.\d+\.\d+$";
```
baronfel, line 64: I think we need to also accept semver prerelease segments here - the `--prerelease` flag just _allows_ those versions

JL03-Yue: Thank you for spoting it!

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
```diff
@@ -61,7 +61,7 @@ public static VersionRange GetVersionRange(this ParseResult parseResult)
         private static bool IsBarePackageVersion (string packageVersion)
         {
             // Define a regular expression pattern to match the "a.b.c" format
-            string pattern = @"^\d+\.\d+\.\d+$";
+            string pattern = @"^\d+\.\d+\.\d+(-[\w\d]+(\.[\w\d]+)*)?(\+[\w\d]+(\.[\w\d]+)*)?$";
```
baronfel, line 64: I had perhaps a more straightforward idea - we have access to NuGet's [SemanticVersion](https://github.com/NuGet/NuGet.Client/blob/8bc8c9a2d55b13d033b14482c2beb4722f636c5b/src/NuGet.Core/NuGet.Versioning/SemanticVersion.cs#L16) class - could we use [`SemanticVersionFactory.TryParse`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Versioning/SemanticVersionFactory.cs#L32C31-L32C32) to handle the parsing instead of binding to a Regex?

JL03-Yue: Thank you! It's updated

baronfel (approved):
This looks great! I think another team member should take a look, but I'm happy with it :)

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
```diff
@@ -1,16 +1,20 @@
...
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.DotNet.Cli;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.NugetSearch;
 using Microsoft.DotNet.ToolPackage;
 using Microsoft.DotNet.Tools.Tool.Search;
+using Newtonsoft.Json.Linq;
 using NuGet.Versioning;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
baronfel, line 17: nit - I'm not sure this static using is required.

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
```diff
@@ -36,6 +30,13 @@ public static VersionRange GetVersionRange(this ParseResult parseResult)
             }
 
             VersionRange versionRange = null;
+
+            // accept 'bare' versions and interpret 'bare' versions as NuGet exact versions
+            if (!string.IsNullOrEmpty(packageVersion) && SemanticVersion.TryParse(packageVersion, out SemanticVersion version2))
+            {
+                packageVersion = "[" + packageVersion + "]";
```
dsplaisted, line 37: This works, but I think I'd prefer if instead of adding the brackets here, we change the logic in ToolPackageDownloader and NuGetPackageDownloader to detect whether a specific version was specified differently.  It would look something like this I believe:

```c#
if (versionRange.MinVersion != null && !versionRange.IsFloating &&
    (versionRange.MaxVersion == null || versionRange.MinVersion == versionRange.MaxVersion))
```

@nkolev92 @zivkan Does that logic look like it makes sense?

dsplaisted (approved):
Looking into this, I realized that it's possible to specify floating versions, for example `--version 1.0.*`.  Do we have any tests covering that scenario?  If not it might be a good idea to add some.

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
zivkan: You'll lose the ability to differentiate between someone who enters `--version 1.2.3` and `--version "[1.2.3, )"`, and so will "incorrectly" do an exact match for someone who very explicitly appears to be asking for "the lowest version greater than or equal to 1.2.3".

Also, if you don't check `IsMinInclusive`, then you'll incorrectly detect "version greater than, but not equal to X" `(1.2.3, )`: https://github.com/NuGet/NuGet.Client/blob/c5abfc6291526cf83365f86bdf03319ee3123180/test/NuGet.Core.Tests/NuGet.Versioning.Test/VersionRangeTests.cs#L480-L491. But the likelyhood of anyone wanting this is so low as to probably be not relevant.

I don't think that range `[1.2.3, ]` makes sense. If `VersionRange.Parse` doesn't throw, then technically this isn't the same thing as `1.2.3`, so technically `IsMaxInclusive` should be checked. But now this is getting unhelpfully pedantic, so it probably doesn't matter if this get treated like an exact match

Finally, if you keep this version parsing here, you've already done most of the expensive CPU work, so you could `return new VersionRange(minVersion: packageVersion, isMinInclusive: true, maxVersion: packageVersion, isMaxInclusive: true, originalString = "[" + packageVersion + "]");`

zivkan (approved):

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
JL03-Yue: Thank you very much for suggestions @dsplaisted and @zivkan

src/Cli/dotnet/commands/dotnet-tool/install/ParseResultExtension.cs
```diff
@@ -36,6 +30,13 @@ public static VersionRange GetVersionRange(this ParseResult parseResult)
             }
 
             VersionRange versionRange = null;
+
+            // accept 'bare' versions and interpret 'bare' versions as NuGet exact versions
+            if (!string.IsNullOrEmpty(packageVersion) && NuGetVersion.TryParse(packageVersion, out SemanticVersion version2))
+            {
+                return new VersionRange(minVersion: new NuGetVersion(packageVersion), includeMinVersion: true, maxVersion: new NuGetVersion(packageVersion), includeMaxVersion: true, originalString: "[" + packageVersion + "]");
```
zivkan, line 37: You only need to parse the version string once, not 3 times

```suggestion
            if (!string.IsNullOrEmpty(packageVersion) && NuGetVersion.TryParse(packageVersion, out NuGetVersion version2))
            {
                return new VersionRange(minVersion: version2, includeMinVersion: true, maxVersion: version2, includeMaxVersion: true, originalString: "[" + packageVersion + "]");
```

PR #36676, opened 2023-11-07
Ensure workload install state file has proper ACLs

joeloff:
# Description

Workload commands are performing arbitrary file operations when creating and removing the workload install state file. The state files were added during RC1. While this is acceptable for file-based installations, global installs on Windows require a specific pattern to ensure that the command is properly elevated and the install server perform all the file operations including setting access control entries.

# Impact

Install state files are limited to using rollback files with workload commands. The current implementation may fail if the state file is created/removed because the install client won't have sufficient permissions. If users decide to run from an elevated prompt they can still encounter errors since the security descriptor is incorrect.

# Changes

* Install state file operations were moved to the respective workload installers
* MSI dispatcher was updated to include operations for writing and removing install state files
* Validation logic was added to avoid abusing the new server operations to write to arbitrary locations.
* Manual tests were updated.

# Risk
Low

# Testing

We have manual tests for security descriptors. The tests were updated to include the new install state scenario. Below is the sample output from running the manual tests from a non-elevated prompt and going through UAC before verifying the directory/file descriptors created by the elevated process. Group and owner are set to built-in administrators, full access is granted to built-in administrators and SYSTEM with read/execute/synchronize permissions for built-in users and Everyone. File ACEs should all be inherited from the parent directory.

```
Verifying directory expectations for C:\ProgramData\SDDLTest\workloads\8.0.100\InstallState
Verifying descriptor: O:BAG:BAD:(A;OICIID;0x1200a9;;;WD)(A;OICIID;FA;;;SY)(A;OICIID;FA;;;BA)(A;OICIID;0x1200a9;;;BU)
Verifying file expectations for C:\ProgramData\SDDLTest\workloads\8.0.100\InstallState\default.json
Verifying descriptor: O:BAG:BAD:(A;ID;0x1200a9;;;WD)(A;ID;FA;;;SY)(A;ID;FA;;;BA)(A;ID;0x1200a9;;;BU)
```

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
@@ -550,11 +550,11 @@ namespace Microsoft.DotNet.Installer.Windows
             if (IsElevated)
             {
                 // Create the parent folder for the state file and set up all required ACLs
-                MsiPackageCache.CreateSecureDirectory(Path.GetDirectoryName(path));
+                SecurityUtils.CreateSecureDirectory(Path.GetDirectoryName(path));
 
                 File.WriteAllLines(path, jsonLines);
 
-                MsiPackageCache.SecureFile(path);
+                SecurityUtils.SecureFile(path);
             }
             else if (IsClient)
             {
--- a/src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
@@ -27,6 +27,11 @@ namespace Microsoft.DotNet.Installer.Windows
         private string _dotNetHome;
 
         /// <summary>
+        /// Full path to the root directory for storing workload data.
+        /// </summary>
+        public static readonly string WorkloadDataRoot = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "dotnet", "workloads");
+
+        /// <summary>
         /// Default reinstall mode (equivalent to VOMUS).
         /// </summary>
         public const ReinstallMode DefaultReinstallMode = ReinstallMode.FILEOLDERVERSION | ReinstallMode.FILEVERIFY |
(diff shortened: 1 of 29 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -487,5 +492,75 @@ protected void UpdateDependent(InstallRequestType requestType, string providerKe
...
+            Elevate();
+
+            if (IsElevated)
+            {
+                // Create the parent folder for the state file and set up all required ACLs
+                MsiPackageCache.CreateSecureDirectory(Path.GetDirectoryName(path));
+
+                File.WriteAllLines(path, jsonLines);
+
+                MsiPackageCache.SecureFile(path);
```
dsplaisted, line 557: In this case `CreateSecureDirectory` and `SecureFile` don't seem to be related to the MSI Package Cache.  Should we rename the `MsiPackageCache` class or refactor it somehow?

dsplaisted (approved):

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
joeloff: Good point, we can factor out the generic methods and leave just the cache specific pieces in this class.

joeloff:
Holding until 8.0.2 unless there are changes for 8.0.1

src/Tests/SDDLTests/Program.cs
```diff
@@ -230,6 +246,25 @@ private static void VerifyDescriptors()
...
+        private static void VerifyInstallStateDescriptors()
+        {
+            // Dump the descriptor of ProgramData since it's useful for analyzing.
+            DirectorySecurity ds = new DirectorySecurity(s_programData, s_accessControlSections);
+            string descriptor = ds.GetSecurityDescriptorSddlForm(s_accessControlSections);
+            Console.WriteLine($" Directory: {s_programData}");
+            Console.WriteLine($"Descriptor: {descriptor}");
+
+            VerifyDirectorySecurityDescriptor(s_installStateDirectory, "BA", "BA", 4, "A;OICIID;0x1200a9;;;WD", "A;OICIID;FA;;;SY", "A;OICIID;FA;;;BA", "A;OICIID;0x1200a9;;;BU");
+            VerifyFileSecurityDescriptor(s_installStateFileAssetPath, "BA", "BA", 4, "A;ID;0x1200a9;;;WD", "A;ID;FA;;;SY", "A;ID;FA;;;BA", "A;ID;0x1200a9;;;BU");
```
MiYanni, line 265: I know I've seen this security descriptor notation before, but might be useful to have a comment with a link to the documentation on it. They kinda look like "magic strings" without that knowledge.

MiYanni (approved):

src/Tests/SDDLTests/Program.cs
joeloff: There are links in the code (SecurityUtils and MsiPackageCache) - I just didn't duplicate the links in the manual test code. But this is the old PR. The ported on for 8.0.200 has more changes and likely the one we'll take

MiYanni:
No changes for 8.0.1. Only keeping the backport to 8.0.2xx. Closing.

PR #36707, opened 2023-11-07
Bump dockercredsprovider to pick up a bugfix

baronfel:
Version 2.2.1 includes a fix for Docker credential keys that contain scheme information, not just domain names. This is the only change from 2.2.0.


baronfel:
This bumps the version of an artifacts managed by https://github.com/dotnet/source-build-externals - need to send a bump to that repo. Working on that now.

MiYanni (approved):

Issue #32233, opened 2023-05-03
Why is the launch profile not being passed into `dotnet watch run` correctly?


mrakgr:
### Describe the bug

I have an `launchSettings.json` file with two profiles `Server Learn` and `Server Play`.

```
PS E:\Webdev\Fable\CFR-In-Fable\src\Server> dotnet run --launch-profile "Server Learn"
Name: Leduc Server; PID: 18560
[|"--mode"; "learn"|]
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5001
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: E:\Webdev\Fable\CFR-In-Fable\src\Server
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...
PS E:\Webdev\Fable\CFR-In-Fable\src\Server> dotnet watch run --launch-profile "Server Learn"
dotnet watch � Started
Name: Leduc Server; PID: 16720
[|"--mode"; "play"|]
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5000
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: E:\Webdev\Fable\CFR-In-Fable\src\Server
```

You can tell how it is not passing in the `Server Learn` profile into watch run by the different command line arguments being printed out as well as the port being 5000 instead of 5001.

I'll provide a minimal repro upon request if needed.

I just want to confirm whether I am doing something wrong in how I am running `dotnet watch` first?

### Further technical details

```
PS E:\Webdev\Fable\CFR-In-Fable\src\Server> dotnet --info
.NET SDK:
 Version:   7.0.203
 Commit:    5b005c19f5

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.203\

Host:
  Version:      8.0.0-preview.1.23110.8
  Architecture: x64
  Commit:       7deac7d6da
```

tmat:
[...]

PR #36744, opened 2023-11-08
Allow trailing commas in launch settings JSON for dotnet watch

martincostello:
Make `dotnet watch` resilient to trailing commas and comments in `launchSettings.json` files like `dotnet run` is:

https://github.com/dotnet/sdk/blob/f52240f11ad291e6ee3cff86e83c0f7a21b60370/src/Cli/dotnet/commands/dotnet-run/LaunchSettings/LaunchSettingsManager.cs#L31-L35

Relates to #32233 and dotnet/aspnetcore#51921.

### DIFF
--- a/src/Tests/dotnet-watch.Tests/LaunchSettingsProfileTest.cs
+++ b/src/Tests/dotnet-watch.Tests/LaunchSettingsProfileTest.cs
@@ -1,0 +1,147 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.DotNet.Watcher.Internal;
+using Microsoft.Extensions.Tools.Internal;
+
+namespace Microsoft.DotNet.Watcher.Tools;
+
+public class LaunchSettingsProfileTest
+{
+    private readonly IReporter _reporter;
+    private readonly TestAssetsManager _testAssets;
+
+    public LaunchSettingsProfileTest(ITestOutputHelper output)
+    {
+        _reporter = new TestReporter(output);
+        _testAssets = new TestAssetsManager(output);
+    }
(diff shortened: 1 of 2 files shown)


src/Tests/dotnet-watch.Tests/LaunchSettingsProfileTest.cs
```diff
@@ -0,0 +1,147 @@
...
+    private readonly TestAssetsManager _testAssets;
+
+    public LaunchSettingsProfileTest(ITestOutputHelper output)
+    {
+        _reporter = new TestReporter(output);
+        _testAssets = new TestAssetsManager(output);
+    }
+
+    [Fact]
+    public void LoadsLaunchProfile()
```
tmat, line 21: Wouldn't a single test that has both a comment and comma be sufficient?

martincostello: I can combine them if you like - I just assumed more granular tests would be preferred.

baronfel:
Thanks for looking into this - I would love to see these two commands share more infrastructure overall, but until we get to that point sharing behavior is also good :)

src/Tests/dotnet-watch.Tests/LaunchSettingsProfileTest.cs
tmat: I think it would be better as a single test.

tmat (approved):

tmat:
Is https://github.com/dotnet/sdk/issues/32233 fixed by this change?

martincostello:
It's inconclusive to me as there's no launch profile provided in that issue, but it fixes the issue described by  dotnet/aspnetcore#51921.

PR #36777, opened 2023-11-09
Add custom props imports around DB.props import Fixes #36749

### LINKED ISSUE #36749: Microsoft.NET.Sdk imports CustomBeforeDirectoryBuildProps after DirectoryBuildPropsPath
`Microsoft.Common.props` (from dotnet/msbuild) has these imports surrounding `DirectoryBuildPropsPath`:

```xml
  <Import Project="$(CustomBeforeDirectoryBuildProps)" Condition="'$(CustomBeforeDirectoryBuildProps)' != ''" />

  <Import Project="$(DirectoryBuildPropsPath)" Condition="'$(ImportDirectoryBuildProps)' == 'true' and exists('$(DirectoryBuildPropsPath)')"/>

  <Import Project="$(CustomAfterDirectoryBuildProps)" Condition="'$(CustomAfterDirectoryBuildProps)' != ''" />
```

`Sdk.props` handles `DirectoryBuildPropsPath` early:

https://github.com/dotnet/sdk/blob/82444e07cf1727fa9bf2785e7c866cedc6b84058/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props#L54-L58

This doesn't consider `CustomBeforeDirectoryBuildProps`, which causes `DirectoryBuildPropsPath` to get imported before `CustomBeforeDirectoryBuildProps`.

cc @dsplaisted

Forgind:
Fixes #36749

Sdk.props imports Directory.Build.props. There's new custom logic to import something immediately before or after that import in MSBuild, but it was missed here.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
@@ -51,7 +51,11 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DirectoryBuildPropsPath Condition="'$(_DirectoryBuildPropsBasePath)' != '' and '$(_DirectoryBuildPropsFile)' != ''">$([System.IO.Path]::Combine('$(_DirectoryBuildPropsBasePath)', '$(_DirectoryBuildPropsFile)'))</DirectoryBuildPropsPath>
   </PropertyGroup>
 
+  <Import Project="$(CustomBeforeDirectoryBuildProps)" Condition="'$(CustomBeforeDirectoryBuildProps)' != '' and exists('$(CustomBeforeDirectoryBuildProps)')" />
+
   <Import Project="$(DirectoryBuildPropsPath)" Condition="'$(ImportDirectoryBuildProps)' == 'true' and exists('$(DirectoryBuildPropsPath)')"/>
+
+  <Import Project="$(CustomAfterDirectoryBuildProps)" Condition="'$(CustomAfterDirectoryBuildProps)' != '' and exists('$(CustomAfterDirectoryBuildProps)')" />
 
   <PropertyGroup>
     <ImportDirectoryBuildProps>false</ImportDirectoryBuildProps>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
@@ -30,58 +30,8 @@ Copyright (c) .NET Foundation. All rights reserved.
       Similar to the property above, it must be set here.
     -->
     <UsingNETSdkDefaults>true</UsingNETSdkDefaults>
-  </PropertyGroup>
 
-
-  
-  <!-- We need to put the UseArtifactsOutput logic after the import of Directory.Build.props, but before the MSBuild Project Extensions .props import.
-       However, both of these things happen in Microsoft.Common.props with no opportunity to insert logic in between them.
-
-       So what we do here is duplicate the Directory.Build.props import logic from Microsoft.Common.props, and then set ImportDirectoryBuildProps to
-       false so that it doesn't get imported twice.
-
-       Alternatively, we could add a hook in MSBuild to define a file to import at the right location, to avoid duplicating the logic here.
-       -->
-  <PropertyGroup>
-    <ImportDirectoryBuildProps Condition="'$(ImportDirectoryBuildProps)' == ''">true</ImportDirectoryBuildProps>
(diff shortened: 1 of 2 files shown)


src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -51,8 +51,12 @@ Copyright (c) .NET Foundation. All rights reserved.
     <DirectoryBuildPropsPath Condition="'$(_DirectoryBuildPropsBasePath)' != '' and '$(_DirectoryBuildPropsFile)' != ''">$([System.IO.Path]::Combine('$(_DirectoryBuildPropsBasePath)', '$(_DirectoryBuildPropsFile)'))</DirectoryBuildPropsPath>
   </PropertyGroup>
 
+  <Import Project="$(CustomBeforeDirectoryBuildProps)" Condition="'$(CustomBeforeDirectoryBuildProps)' != '' and exists('$(CustomBeforeDirectoryBuildProps)')" />
```
tmds, line 54: Based on https://github.com/dotnet/sdk/issues/36749#issuecomment-1802667057, I think @dsplaisted meant for the `DirectoryBuildPropsPath` Import to be removed here, and the `CustomBeforeDirectoryBuildProps`/`CustomAfterDirectoryBuildProps` to get set to implement this logic through callbacks from `Microsoft.Common.props`.

dsplaisted: Correct.  We should remove this duplicated import logic instead of adding more to it.

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
```diff
@@ -30,54 +30,8 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-  <PropertyGroup Condition="'$(UseArtifactsOutput)' == 'true' And '$(BaseIntermediateOutputPath)' == '' And '$(UseArtifactsIntermediateOutput)' == 'true'">
-    <BaseIntermediateOutputPath Condition="'$(IncludeProjectNameInArtifactsPaths)' == 'true'">$(ArtifactsPath)\obj\$(ArtifactsProjectName)\</BaseIntermediateOutputPath>
-    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)' == ''">$(ArtifactsPath)\obj\</BaseIntermediateOutputPath>
-  </PropertyGroup>
 
-  <!-- Record whether ArtifactsPath / UseArtifactsOutput was set at this point in evaluation.  We will generate an error if these properties are set
-       after this point (ie in the project file). -->
-  <PropertyGroup Condition="'$(UseArtifactsOutput)' == 'true'">
-    <_ArtifactsPathSetEarly>true</_ArtifactsPathSetEarly>
+    <CustomAfterDirectoryBuildProps>$(MSBuildThisFileDirectory)UseArtifactsOutputPath.props</CustomAfterDirectoryBuildProps>
```
baronfel, line 34: Do we need to defensively check if `CustomAfterDirectoryBuildProps` is empty here? Since this is used in an `Import` we can't allow multiple semicolon-delimited values, but we also can't just accept any arbitrary user input here. cc @rainersigwald / @dsplaisted - this means that no one else can ever use this property for SDK projects, right?

Forgind: I happily retract the statement I made in triage: you can't Import from an Item, but it appears you can import from a semicolon-delimited Property:
https://github.com/dotnet/msbuild/pull/2014

I'll change this to just add it to the end of the list.

Forgind, line 34:
```suggestion
    <CustomAfterDirectoryBuildProps>$(CustomAfterDirectoryBuildProps);$(MSBuildThisFileDirectory)UseArtifactsOutputPath.props</CustomAfterDirectoryBuildProps>
```

dsplaisted (approved):

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
tmds: > <Import Project="import1.props;import2.props" Condition="Exists('import1.props;import2.props')" />

I assume you should be prudent to not add any paths here that don't exist or because the import will then probably get skipped for all paths.

Forgind: Definitely agree

PR #36787, opened 2023-11-10
Simplify and improve signature verification for workloads on Windows

joeloff:
# Description

Improve code signing verification on Windows for workloads (MSI based installs).

* Add revocation checks when calling `WinVerifyTrust`. Revocation checks default to online, except if the `AllowOnlineRevocationChecks` policy key is set to 0. The policy setting can be used to perform revocation checks when network access is restricted.
* Validate that the certificate chain terminates in a trusted Microsoft root (certificate pinning). 

This change also introduces the use of `Microsoft.Windows.CsWin32` to generate all the native structs and API calls. As a result, we can safely remove a chunk of code and rely on the generated source code instead.


joeloff:
@dtivel would you mind looking at the WinVerifyTrust and X509 extension methods?

I'm still working on fixing up Source Build

src/Cli/dotnet/NativeMethods.txt
MiYanni, line 1: What is this file for?

MiYanni (approved):

src/Cli/dotnet/NativeMethods.txt
joeloff: cswin32 is a source generator. This file contains APIs, structs, enums, constants for which C# should be generated.

src/Cli/dotnet/Installer/Windows/Security/AuthentiCode.cs
```diff
@@ -1,190 +1,61 @@
...
+
+                Guid policyGuid = WINTRUST_ACTION_GENERIC_VERIFY_V2;
+
+                trustData[0].cbStruct = (uint)sizeof(WINTRUST_DATA);
+                trustData[0].pPolicyCallbackData = null;
+                trustData[0].pSIPClientData = null;
+                trustData[0].dwUIChoice = WINTRUST_DATA_UICHOICE.WTD_UI_NONE;
+                trustData[0].fdwRevocationChecks = WINTRUST_DATA_REVOCATION_CHECKS.WTD_REVOKE_WHOLECHAIN;
+                trustData[0].dwUnionChoice = WINTRUST_DATA_UNION_CHOICE.WTD_CHOICE_FILE;
+                trustData[0].dwStateAction = WINTRUST_DATA_STATE_ACTION.WTD_STATEACTION_VERIFY;
```
dtivel, line 45: [Docs](https://learn.microsoft.com/windows/win32/api/wintrust/ns-wintrust-wintrust_data) say:

Value | Meaning
-- | --
WTD_STATEACTION_IGNORE 0x00000000 | Ignore the hWVTStateData member.
WTD_STATEACTION_VERIFY 0x00000001 | Verify the trust of the object (typically a file) that is specified by the dwUnionChoice member. The hWVTStateData member will receive a handle to the state data. This handle must be freed by specifying the WTD_STATEACTION_CLOSE action in a subsequent call.
WTD_STATEACTION_CLOSE 0x00000002 | Free the hWVTStateData member previously allocated with the WTD_STATEACTION_VERIFY action. This action must be specified for every use of the WTD_STATEACTION_VERIFY action.

[This sample](https://learn.microsoft.com/windows/win32/seccrypto/example-c-program--verifying-the-signature-of-a-pe-file) shows proper use of WTD_STATEACTION_VERIFY + WTD_STATEACTION_CLOSE.

I don't see a call with WTD_STATEACTION_CLOSE anywhere.  Should you use WTD_STATEACTION_IGNORE instead?

src/Cli/dotnet/Installer/Windows/Security/AuthentiCode.cs
```diff
@@ -1,190 +1,61 @@
...
-        /// </summary>
-        /// <param name="path">The path of the signed file.</param>
-        /// <param name="organizations">A set of trusted organizations.</param>
-        /// <returns><see langword="true"/> if organization described in the certificate subject matches any of the specified 
-        /// organizations; <see langword="false"/> otherwise.</returns>
-        internal static bool IsSignedByTrustedOrganization(string path, params string[] organizations)
+        /// <param name="path">The path of the file to verify.</param>
+        /// <param name="cacheOnlyRevocationChecks">Prevents revocation checks over the network when set to <see langword="true"/>.</param>
+        /// <returns>0 if successful.</returns>
+        public static unsafe int IsSigned(string path, bool cacheOnlyRevocationChecks = false)
```
dtivel, line 25: `cacheOnlyRevocationChecks` is poorly named, as it implies that only revocation checks will be cached when `true`.  (Windows may cache other things, like intermediate certificate authority certificates.)  How about `allowOnlineRevocationChecks`?

src/Tests/dotnet.Tests/WindowsInstallerTests.cs
```diff
@@ -131,58 +131,32 @@ public void InstallMessageDispatcherProcessesMessages()
...
-            Assert.Equal("sha256RSA", certificates[2].SignatureAlgorithm.FriendlyName);
-        }
-
-        [WindowsOnlyFact]
-        public void GetCertificatesRetrievesNothingForUnsignedFiles()
-        {
-            var certificates = AuthentiCode.GetCertificates(Assembly.GetExecutingAssembly().Location);
-
-            Assert.Empty(certificates);
+            X509Certificate certificate = X509Certificate.CreateFromSignedFile(Path.Combine(s_testDataPath, file));
```
dtivel, line 158: Wrap with `using` statement.

src/Cli/dotnet/Installer/Windows/Security/X509CertificateExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Determines whether the certificate is intended for code signing. 
+        /// </summary>
+        /// <returns><see langword="true"/> if the certificate is intended to be used for code signing; otherwise, <see langword="false"/>.</returns>
+        public static bool IsIntendedForCodeSigning(this X509Certificate certificate)
+        {
+            using X509Chain chain = new();
+            chain.ChainPolicy.ApplicationPolicy.Add(ekuCodeSigning);
+            return chain.Build(new X509Certificate2(certificate));
```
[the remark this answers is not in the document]
dtivel: For expired certificates, you have 4 options:

1. Assume you'll never verify an expired certificate.  If you wind up verifying one unexpectedly, `IsIntendedForCodeSigning(...)` will return `false`.  (To me, that's unexpected.)
2. Call `X509Chain.Build(...)` and then filter out `X509ChainStatusFlags.NotTimeValid` in `X509Chain.ChainStatus`.
3. Before calling `X509Chain.Build(...)`, set `X509Chain.ChainPolicy.VerificationTime` with the time from a trusted timestamp.
4. Before calling `X509Chain.Build(...)`, set `X509Chain.ChainPolicy.VerificationFlags` with [`X509VerificationFlags.IgnoreNotTimeValid`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.x509verificationflags).

```C#
using (X509Chain chain = new())
{
    chain.ChainPolicy.VerificationFlags = X509VerificationFlags.IgnoreNotTimeValid;

    bool result = chain.Build(expiredCertificate);

    Console.WriteLine(result);
}
```

Note that `NotTimeValid` applies to both expired certificates and not-yet-valid certificates.  In time, the former is common but the latter is not.  You may want to consider a sanity check on `X509Certificate.NotBefore`.  You should never see a certificate where `(now) < X509Certificate.NotBefore`.

src/Cli/dotnet/Installer/Windows/Security/X509CertificateExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <returns><see langword="true"/> if the certificate has a trusted Microsoft root; otherwise, <see langword="false"/>.</returns>
+        public static bool HasMicrosoftTrustedRoot(this X509Certificate certificate)
+        {
+            unsafe
+            {
+                CERT_CHAIN_POLICY_PARA policyCriteria = default;
+                CERT_CHAIN_POLICY_STATUS policyStatus = default;
+
+                using X509Chain chain = new();
+                bool buildResult = chain.Build(new X509Certificate2(certificate));
```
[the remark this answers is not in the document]
joeloff: My thinking was that at this point I would have verified everything I wanted. I was mainly relying on this to create the context pointer for me so I could call into `CertVerify...` API. But given the issues you mentioned I'll have to restructure this a bit differently, otherwise I need to keep building the same chain and same checks in multiple places.

src/Cli/dotnet/Installer/Windows/Security/X509CertificateExtensions.cs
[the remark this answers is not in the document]
joeloff: Thanks, I verified the `NotTimeValid` using a file that was signed back in 2008 with a SHA1, and it fails to build the chain for the Code Signing EKU. I'll get that fixed.

I'll need to chat with you about some of the status codes.

I looked at the cloning issue you mentioned. When I was reading the documentation for `X509Certificate2(string)`, it seemed that it was intended to be used with a .cer file: *Initializes a new instance of the [X509Certificate2](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2?view=net-7.0) class using a certificate file name.* It works fine with any signed file so I can remove the cloning.

src/Cli/dotnet/Installer/Windows/Security/AuthentiCode.cs
joeloff: No, I think I need to just make the release call to free up the handle.

src/Cli/dotnet/Installer/Windows/Security/AuthentiCode.cs
joeloff: Agree, the suggested name is more clear about the intent.

src/Cli/dotnet/Installer/Windows/Security/X509CertificateExtensions.cs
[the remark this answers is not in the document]
joeloff: Resolving per our conversations. The `X509Chain` dependencies have been removed.

src/Cli/dotnet/Installer/Windows/Security/AuthentiCode.cs
joeloff: Renamed the policy key to `AllowOnlineRevocationChecks`

joeloff:
I'm still waiting on updated runtime-asset packages for the tests. Locally they pass, but until the new package is available in the feed, they'll fail in the PR build.

src/Tests/dotnet.Tests/WindowsInstallerTests.cs
joeloff: Most X509 pieces are removed in the new iteration, but I did add usings for the few that remain

src/Cli/dotnet/Installer/Windows/Security/X509CertificateExtensions.cs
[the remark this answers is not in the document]
joeloff: As per the new iteration, we'll build the chain using the certificate handle and avoid using the `X509Chain.Build()` so we don't recheck policies covered by `WinTrust`

src/Cli/dotnet/Installer/Windows/Security/Signature.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// the first simple chain. 
+        /// </summary>
+        /// <param name="path">The path of the file to verify.</param>
+        /// <returns>0 if the policy could be checked.<see cref="Marshal.GetPInvokeErrorMessage(int)"/> can be called to obtain more detail about the failure.</returns>
+        /// <exception cref="CryptographicException"/>
+        /// <remarks>This method does not perform any other chain validation like revocation checks, timestamping, etc.</remarks>
+        internal static unsafe int HasMicrosoftTrustedRoot(string path)
+        {
+            // Create an X509Certificate2 instance so we can access the certificate context and create a chain context.
+            using X509Certificate2 certificate = new(path);
```
bartonjs, line 33: I feel like WinVerifyTrust already exposes back the certificate, and this has to replicate that work.  I advise checking to see if you can already get this from `trustData->pCert`.  If so, then between the action call to WinVerifyTrust and the cleanup call, do `PCCERT_CONTEXT pCertContext = CertDuplicateCertificateContext(trustData->pCert->psCertContext);`, and pass the PCCERT_CONTEXT (or `CERT_CONTEXT*`, I suppose) into this method instead of `string path`.

It's both more performant, and would avoid a theoretical ToC/ToU problem if the file isn't locked during your call.

richlander:
What will this feature do in airgapped environments? Are we assuming that `AllowOnlineRevocationChecks` policy key is set to 0.

I'm assuming that this feature will create a gap with Linux. What's the thinking on that?

joeloff:
> What will this feature do in airgapped environments? Are we assuming that `AllowOnlineRevocationChecks` policy key is set to 0.
> 
> I'm assuming that this feature will create a gap with Linux. What's the thinking on that?

This is specific to Windows only as it's only used for verifying the signatures for the MSI packages. The policy flag will force the checks to use the cached certificate revocation list in the machine trust store.

baronfel:
@richlander for some context, on non-Windows platforms we use the so-called 'file based' workload installation, which essentially just downloads a NuGet package and extracts its contents to a specific folder structure. In these scenarios we should be making use of NuGet package signing when that is enabled on each platform. This MSI signature verification is something we can do for MSI-based workload installations, which are Windows-only. When we implement package-manager-based workloads for e.g. Linux platforms, we could potentially implement similar signature checks.

src/Cli/dotnet/Installer/Windows/Security/Signature.cs
joeloff: I don't think I have access to it. Since I'm verifying a file, the union struct is a `WINTRUT_FILE_INFO*` and not a `WINTRUST_CERT_INFO*`. In this case, the WinVerifyTrust call creates the cert context from the file, but won't expose that back to me.

I suppose I could do it the other way around, extract the cert context from the file, then let `WinVerifyTrust` verify the cert context, but I'll need to check if the OS does anything else when asked to verify a file.

bartonjs: Ah, I thought that some of the other data got populated by calling into WinVerifyTrust; and had glossed over it being a union type.

So, disregard.

src/Cli/dotnet/Installer/Windows/Security/Signature.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            if (!CertVerifyCertificateChainPolicy(CERT_CHAIN_POLICY_MICROSOFT_ROOT, pChainContext, &policyCriteria, &policyStatus))
+            {
+                CertFreeCertificateChain(pChainContext);
+                // CertVerifyCertificateChainPolicy always returns true, even if the policy check fails. It only returns
+                // false if the policy could not be checked.
+                throw new CryptographicException(string.Format(LocalizableStrings.UnableToCheckCertificateChainPolicy, nameof(CERT_CHAIN_POLICY_MICROSOFT_ROOT)));
+            }
+
+            CertFreeCertificateChain(pChainContext);
```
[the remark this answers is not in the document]
joeloff: Fixed in the latest commit

bartonjs (approved):

Issue #35750, opened 2023-09-27
Redistribute the portable RID graph from the Microsoft.NETCore.Platforms transport package


ViktorHofer:
Runtime now publishes the portable and non portable RID graphs into the MIcrosoft.NETCore.Platforms transport package. The sdk repository should delete their copy of the portable RID graph and pick it up from the transport package instead.

In addition, the build task to manipulate the RID graph was moved over to runtime. Presumably, that task can now be deleted from the sdk repo.

@dsplaisted @elinor-fung @tmds

marcpopMSFT:
Per all of the linked PRs, I believe this was completed.

PR #36807, opened 2023-11-10
Let runtime supply the portable RID graph

ViktorHofer:
Contributes to dotnet/sdk#35750
Related: https://github.com/dotnet/installer/pull/17760

That deleted build task got moved into runtime: https://github.com/dotnet/runtime/blob/faf65649cc5948884e40825889016808b5e36629/src/libraries/Microsoft.NETCore.Platforms/src/Microsoft.NETCore.Platforms.csproj#L52-L60

cc @tmds


dsplaisted:
It looks like tests are failing because the portable RuntimeIdentifier graph can't be found.  Probably the code that creates stage 2 needs to be updated to include that file.

ViktorHofer:
> It looks like tests are failing because the portable RuntimeIdentifier graph can't be found. Probably the code that creates stage 2 needs to be updated to include that file.

Any idea how to do that? The portable and non portable RID graphs are directly flowing from runtime -> installer and bundled into the SDK there.

ViktorHofer:
> these two needed?

Those properties in the sdk are needed, yes.

akoeplinger (approved):

xen0n:
Hi, apparently this change broke [this doc page](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#rid-graph) by accident? As the [`PortableRuntimeIdentifierGraph.json`](https://github.com/dotnet/sdk/blob/main/src/Layout/redist/PortableRuntimeIdentifierGraph.json) was removed here.

akoeplinger:
Thanks, we need to fix the doc page.

ViktorHofer:
Submitted https://github.com/dotnet/docs/pull/41729

PR #36809, opened 2023-11-10
Workload history 3

Forgind:
Update of #34946

This stores before-and-after states of various workload commands (install, update) in workload history records. Those records can be accessed with the `dotnet workload history` command with output like:
```
>dotnet workload history

Id      Date                              Command      Workloads
-------------------------------------------------------------------
1       1/29/2024 4:15:20 PM -08:00       update
2       1/30/2024 12:42:51 PM -08:00      update
3       2/1/2024 1:57:49 PM -08:00        update
4       2/1/2024 1:59:38 PM -08:00        install      maui-android
```

This also supports updating to a particular workload history record via `dotnet workload update --from-history 3`, which updates workloads and manifests to the state immediately after operation 3 was executed.


baronfel:
TODO: @Forgind mentioned in review today that the documentation/spec might need to be updated to account for a couple small changes.

src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
```diff
@@ -182,6 +214,13 @@ protected async Task<List<WorkloadDownload>> GetDownloads(IEnumerable<WorkloadId
             return ret;
         }
 
+        protected void PrintDownloadLink(IEnumerable<string> packageUrls)
+        {
+            Reporter.WriteLine("==allPackageLinksJsonOutputStart==");
+            Reporter.WriteLine(JsonSerializer.Serialize(packageUrls, new JsonSerializerOptions() { WriteIndented = true }));
+            Reporter.WriteLine("==allPackageLinksJsonOutputEnd==");
```
MiYanni, line 221: So, I've removed this kind of start/end tags from the other JSON outputting commands in `main` here: https://github.com/dotnet/sdk/pull/35640
@baronfel Would we want to follow that same paradigm here in 8.0.2xx?

baronfel: Yes, 100%!

baronfel: Yes, 100%!

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -64,67 +66,97 @@ private void ValidateWorkloadIdsInput()
...
-                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadCacheDownloadFailed, e.Message), e, isUserError: false);
+                    try
+                    {
+                        //  Take the union of the currently installed workloads and the ones that are being requested.  This is so that if there are updates to the manifests
+                        //  which require new packs for currently installed workloads, those packs will be downloaded.
+                        //  If the packs are already installed, they won't be included in the results
+                        var existingWorkloads = GetInstalledWorkloads(false);
+                        var workloadsToDownload = existingWorkloads.Union(_workloadIds.Select(id => new WorkloadId(id))).ToList();
+
+                        recorder.Run(() => DownloadToOfflineCacheAsync(workloadsToDownload, new DirectoryPath(_downloadToCacheOption), _skipManifestUpdate, _includePreviews).Wait());
```
dsplaisted, line 112: I'm pretty sure `DownloadToOfflineCacheAsync` doesn't modify the installed workload versions or workloads, so we probably don't need to write a workload history record in this case.

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -64,67 +66,97 @@ private void ValidateWorkloadIdsInput()
...
             return _workloadInstaller.ExitCode;
         }
 
-        public void InstallWorkloads(IEnumerable<WorkloadId> workloadIds, bool skipManifestUpdate = false, bool includePreviews = false, DirectoryPath? offlineCache = null)
+        public void InstallWorkloads(
+            IEnumerable<WorkloadId> workloadIds,
+            bool skipManifestUpdate = false,
+            bool includePreviews = false,
+            DirectoryPath? offlineCache = null,
+            IEnumerable<(ManifestId, ManifestVersion, SdkFeatureBand)> rollbackContents = null)
```
dsplaisted, line 159: Is this parameter used at all?  It looks like it's not.

If it is, similar to other changes I've suggested, I think the type of this parameter can be `WorkloadRollbackInfo`.

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
```diff
@@ -50,38 +51,51 @@ internal class WorkloadUninstallCommand : WorkloadCommandBase
...
+
+                        foreach (var workloadId in _workloadIds)
+                        {
+                            Reporter.WriteLine(string.Format(LocalizableStrings.RemovingWorkloadInstallationRecord, workloadId));
+                            _workloadInstaller.GetWorkloadInstallationRecordRepository()
+                                .DeleteWorkloadInstallationRecord(workloadId, featureBand);
+                        }
 
-                _workloadInstaller.GarbageCollect(workloadSetVersion => _workloadResolverFactory.CreateForWorkloadSet(_dotnetPath, _sdkVersion.ToString(), _userProfileDir, workloadSetVersion));
+                        WorkloadInstallCommand.TryRunGarbageCollection(_workloadInstaller, Reporter, Verbosity, workloadSetVersion => _workloadResolverFactory.CreateForWorkloadSet(_dotnetPath, _sdkVersion.ToString(), _userProfileDir, workloadSetVersion));
```
dsplaisted, line 81: What's the reason for this change?  It looks like the `TryRunGarbageCollection` method will catch any exception thrown, so it won't escape to this method, so if there is a failure the workload history record won't record that correctly.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -46,71 +49,128 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
...
-                    DownloadToOfflineCacheAsync(new DirectoryPath(_downloadToCacheOption), _includePreviews).Wait();
+                    return _workloadHistoryState;
                 }
-                catch (Exception e)
+
+                if (!string.IsNullOrWhiteSpace(_fromHistorySpecified))
                 {
-                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadCacheDownloadFailed, e.Message), e, isUserError: false);
+                    var workloadHistoryRecords = _workloadInstaller.GetWorkloadHistoryRecords().OrderBy(r => r.TimeStarted).ToList();
+                    if (!int.TryParse(_fromHistorySpecified, out int index) || index < 1 || index > workloadHistoryRecords.Count)
```
dsplaisted, line 64: If the index is too large, the current error message text isn't really accurate.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -46,71 +49,128 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
...
             {
-                try
+                if (!string.IsNullOrWhiteSpace(_downloadToCacheOption))
                 {
-                    UpdateWorkloads(_includePreviews, string.IsNullOrWhiteSpace(_fromCacheOption) ? null : new DirectoryPath(_fromCacheOption));
+                    try
+                    {
+                        var workloadIds = GetUpdatableWorkloads();
+                        recorder.HistoryRecord.WorkloadArguments = workloadIds.Select(id => id.ToString()).ToList();
+                        recorder.Run(() => DownloadToOfflineCacheAsync(new DirectoryPath(_downloadToCacheOption), _includePreviews, workloadIds).Wait());
```
dsplaisted, line 90: Downloading to history shouldn't change the current versions or workloads that are installed, so we shouldn't need to create a workload history for it.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -46,71 +49,128 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
...
-                    // Don't show entire stack trace
-                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadUpdateFailed, e.Message), e, isUserError: false);
+                    var packageUrls = GetUpdatablePackageUrlsAsync(_includePreviews).GetAwaiter().GetResult();
+                    PrintDownloadLink(packageUrls);
+                }
+                else if (_adManifestOnlyOption)
+                {
+                    var workloadManifestInfo = _workloadResolver.GetInstalledManifests();
+                    recorder.HistoryRecord.WorkloadArguments = workloadManifestInfo.Select(m => m.Id).ToList();
+                    recorder.Run(() => _workloadManifestUpdater.UpdateAdvertisingManifestsAsync(_includePreviews, string.IsNullOrWhiteSpace(_fromCacheOption) ? null : new DirectoryPath(_fromCacheOption), workloadManifestInfo).Wait());
```
dsplaisted, line 106: I think this option also doesn't need a workload history record.  It does update the advertising manifests, but doesn't otherwise update or install workloads.  The advertising manifests are just there to let you know that there are newer workload versions available to update to.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -46,71 +49,128 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
...
+                                recorder.HistoryRecord.WorkloadArguments = new List<string>();
+                                foreach (var (id, version, featureBand) in rollbackFileContents)
+                                {
+                                    var idString = id.ToString();
+                                    recorder.HistoryRecord.WorkloadArguments.Add(idString);
+                                    rollbackContents[idString] = $"{version}/{featureBand}";
+                                }
+
+                                recorder.HistoryRecord.RollbackFileContents = rollbackContents;
+                            }
```
dsplaisted, line 137: It seems like this code may be redundant, as the same thing happens within `_workloadManifestUpdater.CalculateManifestRollbacks`.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -121,10 +181,46 @@ public void UpdateWorkloads(bool includePreviews = false, DirectoryPath? offline
...
+                    var featureBandAndVersion = m.Value.Split('/');
+                    var currentVersionInformation = _workloadManifestUpdater.GetInstalledManifestVersion(manifestId, throwIfNotFound: false);
+                    var band = currentVersionInformation?.Band;
+                    versionUpdates.Add(new ManifestVersionUpdate(
+                        manifestId,
+                        currentVersionInformation?.Version,
+                        band?.ToString(),
+                        new ManifestVersion(featureBandAndVersion[0]),
+                        featureBandAndVersion[1]));
+                }
```
dsplaisted, line 206: This seems similar to what the manifest updater is doing in `CalculateManifestRollbacks` and `CalculateManifestUpdates`.  So perhaps the logic should be moved there, and possibly refactored to reduce duplication.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -46,71 +49,128 @@ internal class WorkloadUpdateCommand : InstallingWorkloadCommand
...
-            var useRollback = !string.IsNullOrWhiteSpace(_fromRollbackDefinition);
+            var useRollbackOrHistory = !string.IsNullOrWhiteSpace(_fromRollbackDefinition) || !string.IsNullOrWhiteSpace(_fromHistorySpecified);
 
-            var manifestsToUpdate = useRollback ?
-                _workloadManifestUpdater.CalculateManifestRollbacks(_fromRollbackDefinition) :
-                _workloadManifestUpdater.CalculateManifestUpdates().Select(m => m.ManifestUpdate);
+            var manifestsToUpdate = CalculateManifestUpdates(recorder);
 
-            UpdateWorkloadsWithInstallRecord(_sdkFeatureBand, manifestsToUpdate, useRollback, offlineCache);
+            UpdateWorkloadsWithInstallRecord(_sdkFeatureBand, manifestsToUpdate, useRollbackOrHistory, offlineCache);
```
dsplaisted, line 173: Nit: This should make it a bit clearer how the parameter is being used.

```suggestion
            UpdateWorkloadsWithInstallRecord(_sdkFeatureBand, manifestsToUpdate, shouldUpdateInstallState: useRollbackOrHistory, offlineCache);
```

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -143,26 +239,33 @@ public void UpdateWorkloads(bool includePreviews = false, DirectoryPath? offline
             transaction.Run(
                 action: context =>
                 {
+                    UpdateInstalledWorkloadsFromHistory(sdkFeatureBand, context, offlineCache);
```
dsplaisted, line 242: If we do this before updating the manifests, then it will potentially install the old versions of a workload and then have to install the new version after updating the manifests.  I think this should be after the manifest update.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -143,26 +239,33 @@ public void UpdateWorkloads(bool includePreviews = false, DirectoryPath? offline
...
+                        var workloads = GetUpdatableWorkloads();
+                        _workloadInstaller.InstallWorkloads(workloads, sdkFeatureBand, context, offlineCache);
                     }
-                    else
+
+                    UpdateInstallState(shouldUpdateInstallState, manifestsToUpdate, _workloadInstaller.GetFailingWorkload());
+                    if (!string.IsNullOrWhiteSpace(_fromHistorySpecified))
                     {
-                        _workloadInstaller.DeleteInstallState(sdkFeatureBand);
+                        _workloadInstaller.GarbageCollect(workloadSetVersion => _workloadResolverFactory.CreateForWorkloadSet(_dotnetPath, _sdkVersion.ToString(), _userProfileDir, workloadSetVersion, useInstallStateOnly: true), offlineCache);
```
dsplaisted, line 268: This is probably related to the `useInstallStateOnly` parameter, but why is there an extra garbage collection here?  This probably needs an explanation in comments.

dsplaisted (commented):
Thanks for all this work.  Sorry it took so long to review, I was able to review most of the implementation.

src/Cli/dotnet/commands/dotnet-workload/WorkloadHistoryRecord.cs
```diff
@@ -0,0 +1,78 @@
...
+        public string ErrorMessage { get; set; }
+
+        public WorkloadHistoryState StateBeforeCommand { get; set; }
+
+        public WorkloadHistoryState StateAfterCommand { get; set; }
+    }
+
+    internal class WorkloadHistoryState
+    {
+        public Dictionary<string, string> ManifestVersions { get; set; }
```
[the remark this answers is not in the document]
Forgind: I agree, and I can make a couple tweaks here, but I don't think I should do too much until we have changes out to properly implement install and update via workload sets. Without that, I could easily imagine breaking some of the critical pieces. We definitely do need to make sure to circle back to this once we have that, though 👍

src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
```diff
@@ -88,10 +93,37 @@ internal abstract class InstallingWorkloadCommand : WorkloadCommandBase
...
+                foreach (ManifestVersionUpdate update in manifestVersionUpdates)
+                {
+                    if (update.ManifestId.ToString().Equals(failingWorkload))
+                    {
+                        throw new Exception($"Failing workload: {failingWorkload}");
+                    }
+                }
+            }
+
+            var defaultJsonPath = Path.Combine(WorkloadInstallType.GetInstallStateFolder(_sdkFeatureBand, _dotnetPath), "default.json");
```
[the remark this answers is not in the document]
Forgind: I rather dramatically altered a lot of this logic in https://github.com/dotnet/sdk/pull/37346. joeloff changed how some of this was done from my original design beyond what was necessary, and I partially reverted that here, but now that I understand both sides of that, I think a compromise is best, so I'd like to merge 37346 first then pull that change into this and adjust if that's ok.

src/Cli/dotnet/commands/InstallingWorkloadCommand.cs
```diff
@@ -88,10 +93,37 @@ internal abstract class InstallingWorkloadCommand : WorkloadCommandBase
             _workloadManifestUpdaterFromConstructor = workloadManifestUpdater;
         }
 
-        protected IEnumerable<string> GetInstallStateContents(IEnumerable<ManifestVersionUpdate> manifestVersionUpdates) =>
-            ToJsonEnumerable(WorkloadSet.FromManifests(
-                    manifestVersionUpdates.Select(update => new WorkloadManifestInfo(update.ManifestId.ToString(), update.NewVersion.ToString(), /* We don't actually use the directory here */ string.Empty, update.NewFeatureBand))
-                    ).ToDictionaryForJson());
+        protected internal void UpdateInstallState(bool createDefaultJson, IEnumerable<ManifestVersionUpdate> manifestVersionUpdates, string failingWorkload)
+        {
+            if (failingWorkload is not null)
```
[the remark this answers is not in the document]
Forgind: Ah, yeah; sorry about that. Looking at what's available at this site, I think I can even just use the _workloadInstaller in InstallingWorkloadCommand, which should take that out of the API. Hopefully that works for you 🙂

src/Cli/dotnet/commands/dotnet-workload/update/LocalizableStrings.resx
```diff
@@ -153,4 +156,10 @@
   <data name="FromRollbackDefinitionOptionDescription" xml:space="preserve">
     <value>Update workloads based on specified rollback definition file.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="FromHistoryOptionDescription" xml:space="preserve">
+    <value>Update workloads to a previous version specified by the argument. Use the 'dotnet workload history' to see available workload history records.</value>
```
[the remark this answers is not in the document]
Forgind: After, but that's only relevant if you think of it as an operation and choosing to be before or after that operation as opposed to a state. My recollection is that we were moving towards the second interpretation; is that correct @baronfel? Or should I specify that it's _after_ the operation performed to create state x?

src/Cli/dotnet/commands/dotnet-workload/install/IInstaller.cs
```diff
@@ -33,19 +43,6 @@ internal interface IInstaller : IWorkloadManifestInstaller
...
-        /// </summary>
-        /// <param name="sdkFeatureBand">The SDK feature band of the install state file.</param>
-        void DeleteInstallState(SdkFeatureBand sdkFeatureBand);
-
-        /// <summary>
-        /// Writes the specified JSON contents to the install state file.
-        /// </summary>
-        /// <param name="sdkFeatureBand">The SDK feature band of the install state file.</param>
-        /// <param name="jsonLines">The JSON contents describing the install state.</param>
-        void WriteInstallState(SdkFeatureBand sdkFeatureBand, IEnumerable<string> jsonLines);
```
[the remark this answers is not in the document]
Forgind: Same as https://github.com/dotnet/sdk/pull/36809#discussion_r1421109980

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -540,6 +554,46 @@ void RollBackMsiInstall(WorkloadDownload msiToRollback, DirectoryPath? offlineCa
             }
         }
 
+        string GetWorkloadHistoryDirectory()
+        {
+            return Path.Combine(Environment.GetEnvironmentVariable("PROGRAMDATA"), "dotnet", "workloads", _sdkFeatureBand.ToString(), "history");
```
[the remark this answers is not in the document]
joeloff: This should use the environment special folder properties, not the environment variable. - look at the MsiPackageCache

Creating the directory and the initial file will require the same implementation patterns we have for install state. The initial directory and file creation need to be properly ACL'd. Once the file exist, the elevate process should be able to just read from or write to it.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -540,6 +554,46 @@ void RollBackMsiInstall(WorkloadDownload msiToRollback, DirectoryPath? offlineCa
...
+        }
+
+        public void WriteWorkloadHistoryRecord(WorkloadHistoryRecord workloadHistoryRecord)
+        {
+            var historyDirectory = GetWorkloadHistoryDirectory();
+            Directory.CreateDirectory(historyDirectory);
+            string logFile = Path.Combine(historyDirectory, $"{workloadHistoryRecord.TimeStarted:yyyy'-'MM'-'dd'T'HHmmss}_{workloadHistoryRecord.CommandName}.json");
+            File.WriteAllText(logFile, JsonSerializer.Serialize(workloadHistoryRecord, new JsonSerializerOptions() { WriteIndented = true }));
```
joeloff, line 567: This needs to follow the proper patterns for writing to secure locations:
- Custom messages for the server process
- Use the SecurityUtils module to create and assign proper permissions.

src/Cli/dotnet/commands/dotnet-workload/repair/WorkloadRepairCommand.cs
```diff
@@ -58,32 +58,41 @@ internal class WorkloadRepairCommand : WorkloadCommandBase
 
         public override int Execute()
         {
+            WorkloadHistoryRecorder recorder = new WorkloadHistoryRecorder(_workloadResolver, _workloadInstaller);
+            recorder.HistoryRecord.CommandName = "uninstall";
```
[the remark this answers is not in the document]
joeloff: Perhaps we should introduce an abstract proeprty in the workload base command

```csharp
protected abstract string CommandName
{ 
    get;
}
```

The derived classes are forced to implement it and we can just use the property everywhere

src/Cli/dotnet/commands/dotnet-workload/install/IWorkloadManifestUpdater.cs
```diff
@@ -8,20 +8,24 @@ namespace Microsoft.DotNet.Workloads.Workload.Install
 {
     internal interface IWorkloadManifestUpdater
     {
-        Task UpdateAdvertisingManifestsAsync(bool includePreviews, DirectoryPath? offlineCache = null);
+        Task UpdateAdvertisingManifestsAsync(bool includePreviews, DirectoryPath? offlineCache = null, IEnumerable<WorkloadManifestInfo> manifests = null);
```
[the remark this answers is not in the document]
Forgind: I wanted to store it in a workload history record, but the UpdateAdvertisingManifestsAsync method call happens without having calculated that particular argument, so I could either pass in the WorkloadHistoryRecorder and update its argument within the UpdateAdvertisingManifestsAsync method, or I could calculate the installed manifests early. I chose the latter approach, but once it's calculated, there's no reason not to reuse it rather than calculating it again, so that's why I added an argument.

src/Cli/dotnet/commands/dotnet-workload/repair/WorkloadRepairCommand.cs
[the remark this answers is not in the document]
Forgind: You'll laugh at me, but I actually noticed that when I did 'repair', it claimed I did uninstall in the history, and I thought that just meant repair was implemented as a partial uninstall then a partial install with the former taking precedence...oops 😅

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
Forgind: I restructured how this works to pass the WorkloadHistoryRecorder forward so we don't have to calculate the rollback info early and lug it around. It looks a lot cleaner this way (to me, at least).

src/Cli/dotnet/commands/dotnet-workload/uninstall/WorkloadUninstallCommand.cs
Forgind: I don't actually 100% remember why I did this. I think it was either matching something I'd done elsewhere or confusion from a merge conflict. Regardless, shouldn't be important, so I switched back to calling _workloadInstaller.GarbageCollect directly.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
Forgind: They have some pieces in common, most notably the structure of the final output and the general goal...looking at it now, though, I think they have too many subtle differences for code sharing to be practical. I can certainly move it over, though.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
Forgind: It is! I'll add this in comment, but the quick idea is that we already checked if you asked for any manifests we didn't have installed and tried to install those, but we haven't checked if there were manifests you _didn't_ want that were installed. The way we agreed to do that was to just write the list of manifests we want to keep (i.e., those the customer asked for) to the install state file, then garbage collect. That does mean we need a garbage collection right at the end, however.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
Forgind: It occurred to me when you said that that it probably should also only execute if !_historyManifestOnlyOption...

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -378,6 +381,42 @@ protected uint UninstallMsi(string productCode, string logFile, bool ignoreDepen
             throw new InvalidOperationException($"Invalid configuration: elevated: {IsElevated}, client: {IsClient}");
         }
 
+        internal protected string GetWorkloadHistoryDirectory(string sdkFeatureBand)
+        {
+            return Path.Combine(Environment.GetEnvironmentVariable("PROGRAMDATA"), "dotnet", "workloads", sdkFeatureBand.ToString(), "history");
```
joeloff, line 386: Please use the get special folder methods to find this directory instead of the environment variable

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -323,6 +328,16 @@ private void DeleteAllWorkloadInstallationRecords()
 
         public IWorkloadInstallationRecordRepository GetWorkloadInstallationRecordRepository() => RecordRepository;
 
+        public void RemoveWorkloadManifest(string featureBand, string manifestId, string version, DirectoryPath? offlineCache)
```
joeloff, line 331: This may not be this simple. Since you're dealing with MSIs, you will need to verify that the dependencies allow you to remove the installation, or decide that if the current SDK band is the final dependent, that you can remove the dependent first and then uninstall the MSI.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -182,6 +185,8 @@ public void GarbageCollect(Func<string, IWorkloadResolver> getResolverForWorkloa
             }
         }
 
+        public string GetFailingWorkloadFromTest() => null;
```
joeloff, line 188: Is this only intended for testing? If so, consider making it internal and add the InternalsVisibleToAttribute if it's not already set on the project to which this file belongs

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
Forgind: Great point! And fortunately, this is also dead code—I talked with dsplaisted about that a while ago, and we agreed to modify the install state then run garbage collection instead, which should resolve that issue, but then I forgot to remove this. Will fix 🙂

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
Forgind: Yes, only for testing. Note that this whole class is internal, so this being marked 'internal' instead of 'public' doesn't actually affect anything, and since it's implementing the (also internal) IInstaller interface, it has to be public.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
joeloff: As in the install state file? That doesn't change the dependents against the installer?

Forgind: What does that mean?

joeloff: The same MSIs can have multiple parents holding references to them. For workloads, that's usually the SDK. The dependents are stored in registry keys under `HKLM\SOFTWARE\Classes\Installer\Dependencies`. When an SDK releases its reference, the key is simply deleted. When there are no more dependents registered, the CLI can then remove the installation - this happens as part of the workload garbage collection.

Forgind: To be clear, you're saying that the garbage collector will not update that reference properly, so we should (manually?) delete references to the manifests we don't want in the registry?

dsplaisted:
Could you update the design / documentation in `workload-history.md` so it reflects the behavior with this PR, such as the output and table formatting for the history command and the syntax to revert to a previous history state.  It looks like you responded to most of the code feedback, now I think we need to settle on the design.

documentation/general/workloads/workload-history.md
```diff
@@ -34,11 +34,11 @@ The ID will be an identifier for the record which will start at 1 for the oldest
...
+A `--from-history` option will be added to the `dotnet workload update` command, which will update to the state after a workload history record was written. By default the command will update the manifests to the versions specified, as well as install or uninstall workloads to match the workloads that were installed at the specified point. A `--manifests-only` option will be supported which will update the manifests, but not change the current workloads that are installed (this may result in installing or uninstalling packs to match the new manifest versions).
```
marcpopMSFT, line 37: Thanks for updating this. To confirm then, the output of the history will look like what's above in this .md file? Are there any sample outputs in the discussion that I could look at? Additionally, sounds like now we limit the behavior to only after a specific update is listed. I assume that's because we're really only recording the action and end state we got to after a command was run, correct?

Forgind: > Thanks for updating this. To confirm then, the output of the history will look like what's above in this .md file? Are there any sample outputs in the discussion that I could look at?

Yes; it should look like that. I don't think I attached any screenshots previously, but here's what a sample output I just made:
```
>dotnet workload history

Id      Date                              Command      Workloads
----------------------------------------------------------------
1       1/29/2024 4:15:20 PM -08:00       update
2       1/30/2024 12:42:51 PM -08:00      update
3       2/1/2024 1:57:49 PM -08:00        update
```

After I took that, I called `dotnet workload install maui-android`, and it changed to this:
```
>dotnet workload history

Id      Date                              Command      Workloads
-------------------------------------------------------------------
1       1/29/2024 4:15:20 PM -08:00       update
2       1/30/2024 12:42:51 PM -08:00      update
3       2/1/2024 1:57:49 PM -08:00        update
4       2/1/2024 1:59:38 PM -08:00        install      maui-android
```

Additionally, sounds like now we limit the behavior to only after a specific update is listed. I assume that's because we're really only recording the action and end state we got to after a command was run, correct?

Not exactly. We still record both what the state was before and what the state was after—which is important if we want to be able to insert 'unknown' markers. The change here was for clarity on the CLI.

nagilson: This format looks great, I really like this and I think it will be rather useful. :)
That said, I think it'd also be good to contain a bit more information from the history record on here, namely what other arguments were used. For example if they did workload clean, did they use the --all flag? Was this decided against being valuable enough to put here?

Forgind: I think the argument (hehe) against including the full command line is that it can get long in some cases, and it would look awkward if it gets split onto multiple lines. The nice thing about Id/Date/Command/Workloads is that they're all (well, mostly) guaranteed to be pretty short.

I do think it could be useful to have some extra arguments, though. Perhaps it could display full command line except the first three words (i.e., 'dotnet workload update') but also cut it off if it's longer than...20? characters and replace the last three with '...' What do you think of that design? (And I obviously just threw out 20, so if you think that number should be different, feel free to opine.)

baronfel: @Forgind I think logging the whole command line in the history is a good move, but for display purposes I think if you're using the 'Table' abstraction we have it should auto-truncate the display.  One thing I was also thinking about was we may want a `--format <json|table>` argument the way that tool list and a few other commands have, for tooling purposes. That would make it easier to justify truncating the display in the Table format.

Forgind:
Check whether dateTime format is properly outputted for different cultures

documentation/general/workloads/workload-history.md
```diff
@@ -0,0 +1,44 @@
...
+## `dotnet workload history`
+
+The `dotnet workload history` command will show the history of workload installation commands.  The output will be similar to the following format:
+
+|ID|Date|Command|Workloads|
+|--|----|-------|---------|
+|3|2023-02-01|Update|android, maui, wasm-tools|
+| |2023-02-01|Unknown|Unknown|
+|2|2023-01-05|Install|maui, wasm-tools|
+|1|2023-01-01|Install|android|
```
dsplaisted, line 31: Based on our discussion yesterday, I think we're converging on a design.  Here's the updates I would suggest:

- Reverse the order of the workload history so that the latest changes are listed last (it looks like the implementation is already this way, so just this doc needs to be updated).
- Add an "initial state" row with ID 1 that represents the state before the first history record.  This means that `dotnet workload update --from-history 1` will generally roll back to the baseline manifests.
- Add an ID to the "Unknown" rows.  This will allow rolling back from workload commands even if the previous action was "unknown"
- Try to reword "Unknown" to be more understandable.  Perhaps "Unlogged changes"?  Also, if there are any of these rows, add a note at the bottom explaining something like "Rows with Unlogged changes represent changes from actions other than .NET CLI workload commands.  Usually this represents an update to the .NET SDK or to Visual Studio."
- Once we add support for workload sets, we should add a workload version column to the table.  That doesn't need to be in the initial PR we merge for workload history though, we can add it later.

@baronfel What do you think of these proposals?  Any suggestions for better wording?

dsplaisted (commented):
I've added my feedback on the design.  Let's get agreement on the design, then you can implement the design and I can take a second pass at a detailed implementation review.  That will help me avoid reviewing code that's going to change anyway and reviewing the same code multiple times.

documentation/general/workloads/workload-history.md
Forgind: Those all sound reasonable to me. I hadn't even noticed that the ordering was opposite to what was in the design. Might've just had that wrong in my head after reading the original spec.

KieranFoot:
This would close my issue #38681 , since it would allow me to revert to a working state. Providing that it doesn't error out on unsigned packages like every other command does except `dotnet workload install --skip-sign-check`.

Forgind:
> This would close my issue #38681 , since it would allow me to revert to a working state. Providing that it doesn't error out on unsigned packages like every other command does except `dotnet workload install --skip-sign-check`.

Well, as currently implemented, it has roughly the same semantics as `dotnet workload update --from-rollback-file`, so if that one fails, I have a feeling this unfortunately won't help you either.

documentation/general/workloads/workload-history.md
baronfel: Definitely agree on all points, and I can help brainstorm a name for that concept

Forgind:
Things I tested:
1. Update from rollback still works
2. Workload update records workload set version if applicable
3. Workload set version is loaded and used in populating workload history
4. Workload set version is loaded and used if relevant in updating --from-history
5. Update --from-history with a global.json throws an error
6. Update via global.json updates to the relevant version
7. Update can decrease the version of an individual workload even as it updates to a higher workload set version

All testing was done with 8.0.400 and 8.0.401 workload sets. Previews were not tested. MSI-based installs were not tested.

PR #36850, opened 2023-11-14
Create dummy AfterSdkPublish target Fixes #26710

### LINKED ISSUE #26710: "AfterPublish" doesn't invoked
### Describe the bug
image

ASP.Net Core project

image

### To Reproduce
1. download [the Avalonia Gallery](https://github.com/AvaloniaUI/xamlcontrolsgallery)
2. create a task that depends on "AfterPublish" task
3. `dotnet publish`
4. "AfterPublish" and custom task doesn't not invoked
5. if it is a ASP.Net Core project, "AfterPublish" does be invoked.

### Exceptions (if any)
both of the two tasks should be invoked

### Further technical details
sdk: `6.0.300`

Forgind:
Fixes #26710

After discussions with @baronfel and @vijayrkn, it sounds like the cleanest solution for customers who want an "AfterPublish" target in non-web projects is to introduce a new dummy "AfterSdkPublish" target in the base SDK. There appear to be no references to such a target on GitHub, through an online search, or in the DevDiv repo on AzDO, so this is likely safe and hopefully helpful.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props
@@ -12,6 +12,13 @@ Copyright (c) .NET Foundation. All rights reserved.
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
+    <_AfterSdkPublishDependsOn Condition="'$(_IsAspNetCoreProject)' == 'true'">AfterPublish</_AfterSdkPublishDependsOn>
+    <_AfterSdkPublishDependsOn Condition="'$(_IsAspNetCoreProject)' != 'true'">Publish</_AfterSdkPublishDependsOn>
+  </PropertyGroup>
+
+  <Target Name="AfterSdkPublish" AfterTargets="$(_AfterSdkPublishDependsOn)"></Target>
+
+  <PropertyGroup>
     <!--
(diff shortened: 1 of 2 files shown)


dsplaisted (approved):
It would be good to have some tests for this.

oskarb:
I have a strong suspicion that this change does not actually work for all cases.

The unit test seems to only cover the *non-web* case (when `_IsAspNetCoreProject` would be false or unset).


For the *web* case... it looks to me like the earliest time `_IsAspNetCoreProject` is set to true is *inside the `_InitProjectCapabilityProperties`  target*: https://github.com/dotnet/sdk/blob/856d3483ba4d2dab84fcf8e7db66fa8555c4c0ed/src/WebSdk/Publish/Targets/DotNetCLIToolTargets/Microsoft.NET.Sdk.DotNetCLITool.targets#L17-L21

In other words, isn't `_IsAspNetCoreProject` *always* unset at the point where this PR evaluates it first thing in Sdk.props? Thereby making `_AfterSdkPublishDependsOn` always contain `Publish`, and thus making this PR fail to resolve its goal (i.e. to provide a single hook target that works for both web and non-web projects.)

oskarb:
Should it be based on `UsingMicrosoftNETSdkWeb` instead of `_IsAspNetCoreProject`?

https://github.com/dotnet/sdk/blob/856d3483ba4d2dab84fcf8e7db66fa8555c4c0ed/src/WebSdk/Web/Targets/Sdk.Server.props#L17

PR #36853, opened 2023-11-14
Change the default for 'EnforceCodeStyleInBuild' to true with AnalysisLevel 9.0 or greater

mavasani:
Back when we introduced CodeStyle NuGet package and build-time enforcement of IDE CodeStyle analyzers, it was an experimental feature, and hence the MSBuild property `EnforceCodeStyleInBuild` guarding this feature was defaulted to `false`. Now that it has been few release cycles with this package and feature, we will default this property to `true`. The new default value will only apply for AnalysisLevel/TFM 9.0 or greater to avoid any breaking changes in existing projects that configure IDExxxx to warning/error severity in their editorconfig, but do not set this property and hence see warnings/errors in the IDE live analysis, but not on build.

NOTE: This change has been approved in the offline email discussion with Roslyn IDE and compiler teams.

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets
@@ -95,7 +95,10 @@ Copyright (c) .NET Foundation. All rights reserved.
                                       '$(IsAotCompatible)' == 'true'
                                   )">true</EnableAotAnalyzer>
 
-    <!-- EnforceCodeStyleInBuild Allows code style analyzers to be disabled in bulk via msbuild if the user wants to -->
+    <!-- EnforceCodeStyleInBuild allows code style analyzers that are configured to warning/error severity to be enabled on build via msbuild if the user wants to.
+         It is defaulted to 'true' for AnalysisLevel/TFM greater than or equals '9.0', and 'false' otherwise.
+    -->
+    <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == '' And $([MSBuild]::VersionGreaterThanOrEquals($(EffectiveAnalysisLevel), '9.0'))">true</EnforceCodeStyleInBuild>
     <EnforceCodeStyleInBuild Condition="'$(EnforceCodeStyleInBuild)' == ''">false</EnforceCodeStyleInBuild>
   </PropertyGroup>
 


mavasani:
This PR led to a bunch of test failures in tests that were not even verifying/enabling code style build enforcement. The failures were due to the following new warnings showing up:

```
CSC : warning CS9057: The analyzer assembly 'C:\h\w\B1F70958\p\d\sdk\9.0.100-ci\Sdks\Microsoft.NET.Sdk\codestyle\cs\Microsoft.CodeAnalysis.CodeStyle.dll' references version '4.9.0.0' of the compiler, which is newer than the currently running version '4.8.0.0'. [C:\h\w\B1F70958\t\dotnetSdkTests\lh0httyn.1qv\NativeAot_hw_---788461B1_2\HellowWorldNativeAotApp\HellowWorldNativeAotApp.csproj]

CSC : warning CS9057: The analyzer assembly 'C:\h\w\B1F70958\p\d\sdk\9.0.100-ci\Sdks\Microsoft.NET.Sdk\codestyle\cs\Microsoft.CodeAnalysis.CSharp.CodeStyle.dll' references version '4.9.0.0' of the compiler, which is newer than the currently running version '4.8.0.0'. [C:\h\w\B1F70958\t\dotnetSdkTests\lh0httyn.1qv\NativeAot_hw_---788461B1_2\HellowWorldNativeAotApp\HellowWorldNativeAotApp.csproj]
```

Upon further analysis, it seems the test is running with the compiler from C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\Roslyn\csc.exe instead of the one in the SDK, which is version 4.9.0.0.

Given that our CodeStyle analyzer packages are coupled to the compiler version that they were built with (which should be the compiler version that ships in the SDK), enabling 'EnforceCodeStyleInBuild' by default requires that all the builds would need to be with a compiler version greater than or equals one matching the one in the SDK. This seems like a very strict requirement to enforce by default, especially when large number of .NET projects do not care about code style enforcement on build. Having the users explicitly opt into 'EnforceCodeStyleInBuild' ensures that they are signing up to abide by this requirement as they care about always using the latest CodeStyle analyzers + compiler from the SDK for all their build configurations, versus referencing a static version of CodeStyle NuGet package.

In short, I think enabling 'EnforceCodeStyleInBuild' by default will likely lead to many complaints about such CS9057 breaking change warnings on just updating to newer SDK, and that is definitely not desirable. I am going to close out this PR.

PR #36869, opened 2023-11-14
Fix loc mirror branch

marcpopMSFT:


Forgind (approved):
I'm not sure if I just missed that, or if the PR I was looking at had also missed it, but thanks for finding that!

PR #36974, opened 2023-11-16
Don't overwrite user-provided ContainerUser values

baronfel:
Fix https://github.com/dotnet/sdk-container-builds/issues/520

In .NET 8, the .NET Container base images allow for opting to run in a rootless mode by setting a particular user id as the default user for the container. The SDK Container feature opts into this mode by default, inspecting the base images for this value and setting it in our generated images on behalf of the user.

Because this is an impactful change, we also created a mechanism to allow users to opt out of this behavior. Users set `<ContainerUser>root</ContainerUser>` in their project file or their build command and we take that and flow it through the container creation process accordingly.

Or at least we used to. At some point we messed up the sequencing and since we never had an end-to-end test for the way that the implicit user id for the base image interacted with the feature we released in a broken state.

## Customer Impact

Users building containers with the SDK always build containers in rootless mode, and the MSBuild switch to opt out of this has no effect. If the user code does things that require admin permissions, for example creating files in the application's working directory), the application may crash. There are two workarounds:

* As part of container execution (so long after the build), have your container runtime (docker, kubernetes, etc) explicitly run the SDK-generated container image as a user with root permissions. For example, here is how you might run a container as the `root` user: `docker run --user root my-webapp:latest`. This requires users to find and alter their deployment steps, which may not be located with their code.
* Update the application to work in a rootless mode of execution. This can vary wildly in terms of effort. While this is the long-term best option from a security perspective, most users will likely want to schedule and plan for such changes.

## Regression

**Yes** - this probably broke as a result of the changes introduced in https://github.com/dotnet/sdk/pull/34154 prior to RC1.

## Risk

**Low** - tests cover the changes and this path is only hit when a user is explicitly asking for a change.

## Verification/Testing

Automated tests now verify the expected changes to image manifest configurations and enforce the expected ordering of implicit user id selection vs explicit user-provided actions.

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/ImageBuilder.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/ImageBuilder.cs
@@ -110,7 +110,7 @@ internal sealed class ImageBuilder
     /// <summary>
     /// Sets the USER for the image.
     /// </summary>
-    internal void SetUser(string user) => _baseImageConfig.SetUser(user);
+    internal void SetUser(string user, bool isExplicitUserInteraction = true) => _baseImageConfig.SetUser(user, isExplicitUserInteraction);
 
     internal static (string[] entrypoint, string[] cmd) DetermineEntrypointAndCmd(
         string[] entrypoint,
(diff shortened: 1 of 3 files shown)


rainersigwald (approved):

baronfel:
Approved by tactics over email.

PR #37013, opened 2023-11-18
Use .NET 9.0 versions for task dependencies

ellahathaway:
Addresses https://github.com/dotnet/source-build/issues/3665

The following 9.0 prebuilts are eliminated in this PR:

      "Microsoft.Bcl.AsyncInterfaces": {
          "8.0.0-rc.1.23419.4": [
              "src/sdk/artifacts/buildObj/source-build/self/src/artifacts/obj/Release/Sdks/Microsoft.NET.Sdk/tools/project.assets.json"
          ]
      },
      "Microsoft.Extensions.DependencyModel": {
          "8.0.0-rc.1.23419.4": [
              "src/sdk/artifacts/buildObj/source-build/self/src/artifacts/obj/Release/Sdks/Microsoft.NET.Sdk/tools/project.assets.json"
          ]
      },
      "Microsoft.NET.HostModel": {
          "8.0.0-rc.1.23419.4": [
              "src/sdk/artifacts/buildObj/source-build/self/src/artifacts/obj/Release/Sdks/Microsoft.NET.Sdk/tools/project.assets.json"
          ]
      },
      "System.Text.Json": {
          "8.0.0-rc.1.23419.4": [
              "src/sdk/artifacts/buildObj/source-build/self/src/artifacts/obj/Release/Sdks/Microsoft.NET.Sdk/tools/project.assets.json"
          ]
      }

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -32,10 +32,5 @@
     <UsagePattern IdentityGlob="System.Diagnostics.EventLog/*7.0.0*" />
     <UsagePattern IdentityGlob="System.Reflection.MetadataLoadContext/*7.0.0*" />
     <UsagePattern IdentityGlob="System.Security.Cryptography.ProtectedData/*7.0.0*" />
-    
-    <!-- msbuild is still on 8.0 versions of these so we have to pin in SDK main for now or tasks will fail -->
-    <UsagePattern IdentityGlob="Microsoft.NET.HostModel/*8.0.0*" />
-    <UsagePattern IdentityGlob="Microsoft.Extensions.DependencyModel/*8.0.0*" />
-    <UsagePattern IdentityGlob="Microsoft.Bcl.AsyncInterfaces/*8.0.0*" />
   </IgnorePatterns>
 </UsageData>
--- a/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj
@@ -52,10 +52,10 @@
     <PackageReference Include="Microsoft.Build" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="Runtime" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" ExcludeAssets="Runtime" />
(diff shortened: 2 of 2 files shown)


MichaelSimons (approved):

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -32,10 +32,5 @@
     <UsagePattern IdentityGlob="System.Diagnostics.EventLog/*7.0.0*" />
     <UsagePattern IdentityGlob="System.Reflection.MetadataLoadContext/*7.0.0*" />
     <UsagePattern IdentityGlob="System.Security.Cryptography.ProtectedData/*7.0.0*" />
-    
-    <!-- msbuild is still on 8.0 versions of these so we have to pin in SDK main for now or tasks will fail -->
```
dsplaisted, line 36: Does this comment no longer apply? Did MSBuild switch to version 9.0 of these assemblies?

dsplaisted (approved):

eng/SourceBuildPrebuiltBaseline.xml
ellahathaway: I'm a bit confused by that comment.

I don't understand how these assemblies are related to msbuild since System.Text.Json, the ref that was related to these other dependencies per https://github.com/dotnet/sdk/pull/35048/files#r1330753409, is a [7.0 version in msbuild](https://github.com/dotnet/msbuild/blob/5fcddc790f4eeaf953a3d283e39751dd0e1f2992/eng/Version.Details.xml#L55). Second, these prebuilts were added as a workaround to an issue with the sdk tasks as part of https://github.com/dotnet/sdk/pull/35048. https://github.com/dotnet/sdk/pull/35048/files#r1330753409 suggests that the issue was related to the fact that System.Text.Json was on 8.0, but msbuild is on 7.0?

This PR is getting a clean build and my local copy of the VMR with these changes is also building successfully. Based on this, I'm assuming that this comment no longer applies, but as I pointed out above, I'm a bit confused as to how the comment was relevant in the first place.

@marcpopMSFT - What are your thoughts?

dsplaisted: If the full framework CI leg is passing then it's probably OK, for now at least.  When running under Full Framework MSBuild, there can be assembly loading failures if there isn't a binding redirect for the assembly and the assembly versions don't match exactly.  (We just encountered this when building 8.0.1, where we were trying to use version 8.0.1 of System.Text.Json, but Microsoft.Extensions.DependencyModel depended on version 8.0.0.  The resolution was to not update to the new version of System.Text.Json for the full Framework task builds.)

PR #37054, opened 2023-11-20
Improve Container support for AOT publishing and targeting musl runtimes.

baronfel:
Fixes https://github.com/dotnet/sdk-container-builds/issues/524
Fixes https://github.com/dotnet/sdk-container-builds/issues/301

This PR moves the inference logic that used to be in the targets directly into a task because the logic was getting too gnarly. We now check a number of signals about the project to determine the
* repository
* version tag portion (i.e. 8.0)
* variant tag portion (i.e. alpine-aot)
required to infer the base image the project should use.

The decision tree looks very similar to before:
* explicit user `ContainerBaseImage` skips this process entirely
* otherwise inference occurs:
  * the TFM and SDK version are used to determine which version tag should be selected (this logic didn't change at all)
  * the kind of project is investigated to choose between the `aspnet`, `runtime`, `runtime-deps`, and `nightly/runtime-deps` (for aot/extras scenarios) repositories
    * [x] TODO: add tests that verify for .NET 7 and below none of the nightly/aot/extra images are selected
  * the target RID is investigated to see if we should be using `-alpine` variants of these images
  * the AOT capabilities and globalization states are investigated to determine if the `-aot` or `-extras` variants of the image should be used

The first commit is a refactoring of our tests that used a local registry run inside a Docker container. Recent versions of Docker for Windows have introduced validation that prevent pushing a manifest list tag (i.e. the 6.0 base image tag that contains N image variants for that tag) unless the entire manifest's tags are present in that container. Since we do not test or cover that use case, I slightly refactored the tests to make it easy to just push up the single variants we use.


baronfel:
I added a [bunch of tests](https://github.com/dotnet/sdk/pull/37054/files#diff-4be74b4623e1ff907195394520ed5605a52c846c47c77a07eccf9387e050124bR255-R358) that should clarify the expected behaviors here with regards to image and tag selection:

* MuslRidsGetAlpineContainers
* AOTAppsGetAOTImages
* AOTAppsWithCulturesGetExtraImages
* AOTAppsLessThan8DoNotGetAOTImages
* AOTAppsLessThan8WithCulturesDoNotGetExtraImages

src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
```diff
@@ -7,9 +7,49 @@
...
+    public PlatformSpecificManifest? PickBestManifestForRid(IReadOnlyDictionary<string, PlatformSpecificManifest> ridManifestDict, string runtimeIdentifier)
+    {
+        var bestManifestRid = GetBestMatchingRid(_runtimeGraph, runtimeIdentifier, ridManifestDict.Keys);
+        if (bestManifestRid is null) {
+            return null;
+        }
+        return ridManifestDict[bestManifestRid];
+    }
+
+    private static string? GetBestMatchingRid(RuntimeGraph runtimeGraph, string runtimeIdentifier, IEnumerable<string> availableRuntimeIdentifiers)
```
MichalPavlik, line 35: nit: Why is this method `static`? :) It's invoked form instance method and first parametr represents instance data.

src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageAndTag.cs
```diff
@@ -0,0 +1,245 @@
...
+    public string ContainerFamily { get; set; }
+
+    /// <summary>
+    ///  The final base image computed from the inputs (or explicitly set by the user if IsUsingMicrosoftDefaultImages is true)
+    /// </summary>
+    [Output]
+    public string? ComputedContainerBaseImage { get; private set; }
+
+    private bool IsAspNetCoreProject => FrameworkReferences.Length > 0 && FrameworkReferences.Any(x => x.ItemSpec.Equals("Microsoft.AspnetCore.App", StringComparison.Ordinal));
+    private bool IsMuslRID => TargetRuntimeIdentifier.StartsWith("linux-musl", StringComparison.Ordinal);
```
MichalPavlik, line 78: nit: I think this should be named `IsMuslRid`

MichalPavlik (approved):

src/Containers/Microsoft.NET.Build.Containers/Registry/Registry.cs
baronfel: In general I prefer static functions - that way a reader knows that no other state/inputs other than the function parameters is required to understand the function. The fancy term is 'locality of reference' and it's a concept that all of us functional programming types are prone to doing :)

src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageAndTag.cs
baronfel: great point, renaming now!

Issue #36593, opened 2023-11-01
TaskCanceledException: A task was canceled. hitting ctrl-c during container create


danmoseley:
Installed today's 8.0.100 SDK on Windows. 
Started a `dotnet publish /t:PublishContainer -r linux-x64` then quickly hit Ctrl-C

The `Microsoft.NET.Build.Containers.DockerCli.LoadAsync` or above appears to be missing a catch block for `TaskCanceledException`. I would have expected no message, or a one line saying it was canceled. Then the task should return immediately from the `Microsoft.NET.Build.Containers.Tasks.CreateNewImage.Execute()` method.
```
C:\Program Files\dotnet\sdk\8.0.100\Containers\build\Microsoft.NET.Build.Containers.targets(202,5): error MSB4018: The "CreateNewImage" task failed unexpectedly. [C:\git\eshop2\src\WebApp\WebApp.csproj
]
C:\Program Files\dotnet\sdk\8.0.100\Containers\build\Microsoft.NET.Build.Containers.targets(202,5): error MSB4018: System.Threading.Tasks.TaskCanceledException: A task was canceled. [C:\git\eshop2\src\
WebApp\WebApp.csproj]
C:\Program Files\dotnet\sdk\8.0.100\Containers\build\Microsoft.NET.Build.Containers.targets(202,5): error MSB4018:    at System.Diagnostics.Process.WaitForExitAsync(CancellationToken cancellationToken)
 [C:\git\eshop2\src\WebApp\WebApp.csproj]
C:\Program Files\dotnet\sdk\8.0.100\Containers\build\Microsoft.NET.Build.Containers.targets(202,5): error MSB4018:    at Microsoft.NET.Build.Containers.DockerCli.LoadAsync(BuiltImage image, SourceImage
Reference sourceReference, DestinationImageReference destinationReference, CancellationToken cancellationToken) in /_/src/Containers/Microsoft.NET.Build.Containers/LocalDaemons/DockerCli.cs:line 101 [C
:\git\eshop2\src\WebApp\WebApp.csproj]
C:\Program Files\dotnet\sdk\8.0.100\Containers\build\Microsoft.NET.Build.Containers.targets(202,5): error MSB4018:    at Microsoft.NET.Build.Containers.Tasks.CreateNewImage.PushToLocalRegistryAsync(Bui
[...]

PR #37055, opened 2023-11-20
wrap and handle cancellation exceptions in .NET Task

baronfel:
Fixes https://github.com/dotnet/sdk/issues/36593

Instead of letting the Task cancellation bubble up, we internalize those two errors specifically. This provides a very clean error experience:

image

### DIFF
--- a/src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
+++ b/src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
@@ -31,7 +31,19 @@ public sealed partial class CreateNewImage : Microsoft.Build.Utilities.Task, ICa
 
     public override bool Execute()
     {
-        return Task.Run(() => ExecuteAsync(_cancellationTokenSource.Token)).GetAwaiter().GetResult();
+        try
+        {
+            Task.Run(() => ExecuteAsync(_cancellationTokenSource.Token)).ContinueWith(t => t).GetAwaiter().GetResult();
+        }
+        catch (TaskCanceledException ex)
+        {
+            Log.LogWarningFromException(ex);
+        }
+        catch (OperationCanceledException ex)
+        {
+            Log.LogWarningFromException(ex);
+        }
+        return !Log.HasLoggedErrors;
     }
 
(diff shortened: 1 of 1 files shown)


src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -31,7 +31,19 @@ public sealed partial class CreateNewImage : Microsoft.Build.Utilities.Task, ICa
...
-        return Task.Run(() => ExecuteAsync(_cancellationTokenSource.Token)).GetAwaiter().GetResult();
+        try
+        {
+            Task.Run(() => ExecuteAsync(_cancellationTokenSource.Token)).ContinueWith(t => t).GetAwaiter().GetResult();
+        }
+        catch (TaskCanceledException ex)
+        {
+            Log.LogWarningFromException(ex);
+        }
+        catch (OperationCanceledException ex)
```
WeihanLi, line 42: Seemed we could just catch `OperationCanceledException`, and maybe we could add a `when(_cancellationTokenSource.IsCancellationRequested)` exception filter to catch exception only when `_cancellationTokenSource` cancelled

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
```diff
@@ -31,7 +31,19 @@ public sealed partial class CreateNewImage : Microsoft.Build.Utilities.Task, ICa
 
     public override bool Execute()
     {
-        return Task.Run(() => ExecuteAsync(_cancellationTokenSource.Token)).GetAwaiter().GetResult();
+        try
+        {
+            Task.Run(() => ExecuteAsync(_cancellationTokenSource.Token)).ContinueWith(t => t).GetAwaiter().GetResult();
```
MichalPavlik, line 36: Please remove the task continuation if it's not necessary

MichalPavlik (approved):

src/Containers/Microsoft.NET.Build.Containers/Tasks/CreateNewImage.cs
baronfel: good catch, removed it now.

PR #37072, opened 2023-11-21
Containers support was moved higher to enable creating image of CLI apps

### LINKED ISSUE #513: Console applications should be able to be packaged by default
The Container targets aren't imported by console applications in the base .NET SDK, and so the `PublishContainer` target isn't usable for them. We should push the Import of these props/targets down into the Base SDK so that multi-project orchestration scenarios and console publishing works.

MichalPavlik:
Fixes https://github.com/dotnet/sdk-container-builds/issues/513

### DIFF
--- a/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.targets
+++ b/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.targets
@@ -52,4 +52,17 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Import Project="$(NuGetBuildTasksPackTargets)"
           Condition="Exists('$(NuGetBuildTasksPackTargets)') AND '$(ImportNuGetBuildTasksPackTargetsFromSdk)' == 'true'"/>
+
+  <!-- Import targets from Microdoft.NET.Build.Container.targets -->
+  <PropertyGroup>
+    <_ContainersTargetsDir Condition=" '$(_ContainersTargetsDir)'=='' ">$(MSBuildThisFileDirectory)..\..\..\Containers\build\</_ContainersTargetsDir>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props"
+          Condition="'$(EnableSdkContainerSupport)' == 'true' AND Exists('$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props')" />
+
+  <Import Project="$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets"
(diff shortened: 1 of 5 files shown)


MichalPavlik:
@baronfel, we have CheckContainersPackage target checking Microsoft.NET.Build.Containers package presence. Should I move it too? This check is for users using nuget package which supported only web apps.

src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.targets
```diff
@@ -52,4 +52,17 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    <_ContainersTargetsDir Condition=" '$(_ContainersTargetsDir)'=='' ">$(MSBuildThisFileDirectory)..\..\..\Containers\build\</_ContainersTargetsDir>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props"
+          Condition="'$(EnableSdkContainerSupport)' == 'true' AND Exists('$(MSBuildThisFileDirectory)..\..\..\Containers\build\Microsoft.NET.Build.Containers.props')" />
+
+  <Import Project="$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets"
+    Condition="'$(EnableSdkContainerSupport)' == 'true'
+             AND Exists('$(_ContainersTargetsDir)Microsoft.NET.Build.Containers.targets')
+             AND '$(TargetFramework)' != ''" />
```
baronfel, line 67: I think we need to remove `EnableSdkContainerSupport` from these conditions, and instead put that condition on the `PublishContainer` target directly - this will enable someone to call `-t:PublishContainer` at the solution level and have the target be available (but not enabled) for all projects.

MichalPavlik: Okay, I wanted to avoid importing these files when publishing is not required.

baronfel: Here's an example of a scenario that should become possible after we move the Container targets into the base SDK:

* myapp.sln
  * myapp.lib.csproj
  * myapp.web.csproj (references myapp.lib)
  * myapp.cli.csproj (references myapp.lib, has manually set `EnableSdkContainerSupport` to `true`)
  * myapp.tests.csproj (references myapp.lib)

We should be able to run `dotnet publish -t:PublishContainer` on `myapp.sln` and get containers generated for `myapp.web` and `myapp.cli` as a result.

baronfel:
This looks correct and great to me - I'd like some level of scenario test to show folks that it does work, and then after merge I can look at updating docs and other resources to talk about the property.

baronfel (approved):
With the addition of the test showing how it all works, I'm very happy to approve this PR! Thanks for plumbing it through @MichalPavlik!

Issue #29839, opened 2023-01-09
dotnet.exe steals -d option from dotnet-* tools starting in .NET 6


bricelam:
### Describe the bug
After investigating https://github.com/dotnet/efcore/issues/26687, it appears that, starting in .NET 6, `dotnet` is stealing the `-d` option preventing it from being passed to the `dotnet-ef` tool.

### To Reproduce

```cmd
dotnet new console
dotnet add package Microsoft.EntityFrameworkCore.Design

dotnet tool install -g dotnet-ef
dotnet ef dbcontext scaffold X X -d --verbose
```

The verbose output shows that the option has been stripped when invoking the "inner" tool:

> dotnet exec ...ef.dll dbcontext scaffold X X ...

If I invoke the tool using the stub instead, it works:

```cmd
dotnet-ef dbcontext scaffold X X -d --verbose
```

> dotnet exec ...ef.dll dbcontext scaffold X X -d ...

### Further technical details
- Repros using both the .NET 6 and 7 SDKs (via global.json)
- Works using the .NET 5 SDK
- I suspect it was broken by 938a4048e534fb5abad2ba0b901c1884c202b917

baronfel:
I think you're right - specifically [this line](https://github.com/dotnet/sdk/blob/be8529b23ca7d3748343df9f2e699b401abcd9ea/src/Cli/dotnet/Program.cs#L238) I think has removed all aliases of the Diagnostics option.

Normally we want this for other SDK-bundled commands, but for tools we should have an alternative mechanism that doesn't mutate the args array.

Forgind:
This was fixed, though it didn't meet the bar for servicing, so it's in 8.0.2xx

PR #37091, opened 2023-11-21
Parse -d command differently based on position

Forgind:
Fixes https://github.com/dotnet/sdk/issues/29839

Description
If -d is at the beginning of the list of subarguments, that is, before the subcommand, it should be associated with 'dotnet' and therefore ignored when getting at the list of subarguments. If it's after the subcommand, however, it should be parsed as part of the subcommand.

This differentiates between the two positionings.

Impact
Some users specify -d, intending to get diagnostic-verbosity output. Other specify it as a subcommand for dotnet-ef to indicate that it should use attributes to configure the model. There is currently no way to disambiguate between the two, and dotnet (as the initial command) takes preference, automatically stripping any -d anywhere in the input, preventing users from using it as an option to their subcommands (in this case dotnet-ef).

Changes
Alter -d parsing to parse the -d as diagnostic if it comes before the subcommand but as an option to the subcommand if it comes after it.

Risk
Some people might append -d to the end of their command, intending to get diagnostic output, and that would no longer work as intended. I don't know how common that is.

Testing
I changed a unit test to ensure -d is parsed as intended.

### DIFF
--- a/src/Cli/dotnet/ParseResultExtensions.cs
+++ b/src/Cli/dotnet/ParseResultExtensions.cs
@@ -122,16 +122,38 @@ namespace Microsoft.DotNet.Cli
             var subargs = args.ToList();
 
             // Don't remove any arguments that are being passed to the app in dotnet run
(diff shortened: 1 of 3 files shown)


baronfel (approved):
Thanks for forward-porting this and getting us to make a decision on severity :D

PR #37160, opened 2023-11-27
Treat rtm as stable Fixes #36500

### LINKED ISSUE #36500: For workload feature bands, treat `-rtm` preview versions the same as the stabilized, non-preview versions
When calculating SDK feature bands for workloads, we ignore the prerelease specifier if it's `-dev` or `-ci`.  We should probably update this to also ignore `-rtm` versions.

This is compounded by the fact that we don't parse `-rtm` versions correctly.  Since we normally expect a preview band of something like `-preview.7` or `-rc.1`, when it comes to an SDK version such as `8.0.100-rtm.23526.1`, we end up with a feature band of `8.0.100-rtm.23526`, which is wrong.

Forgind:
Fixes #36500


src/Containers/Microsoft.NET.Build.Containers/Tasks/ComputeDotnetBaseImageTag.cs
```diff
@@ -99,7 +99,7 @@ public override bool Execute()
                 }
                 Log.LogError(Resources.Strings.InvalidSdkPrereleaseVersion, channel);
                 return null;
-            case "alpha" or "dev" or "ci":
+            case "alpha" or "dev" or "ci" or "rtm":
```
baronfel, line 102: this change doesn't do anything - see line 93 above.

Forgind: Oh 🙂 You're right; I can remove it

baronfel (approved):

MiYanni (approved):

PR #37165, opened 2023-11-27
parse auth scheme case-insensitively

baronfel:
This is a potential fix for https://github.com/dotnet/sdk-container-builds/issues/527

The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#auth-scheme) for the www-authenticate header's `auth-scheme` parameter state that the value should be treated case-insensitively, but we were doing case-sensitive comparisons here. This means that compliant-but-unexpected schemes like `BASIC` (which was returned from Sonatype Nexus 3) wouldn't ever authenticate.


schmitch (approved):
I mostly tested it, if the changes worked, and everything worked

rainersigwald (approved):

baronfel:
Going to merge this as a spec-compliance fix.

PR #37221, opened 2023-11-29
don't use cookies to unblock CSRF workflows

baronfel:
Fixes https://github.com/dotnet/sdk-container-builds/issues/418 (possibly)

Since Harbor enforces CSRF tokens we have to reject all uses of cookies to unblock our support of it. This seems to be what [other registry clients](https://github.com/moby/moby/blob/master/registry/search_session.go#L167C1-L171C18) do.


baronfel:
This has been tested and validated by @djheap [here](https://github.com/dotnet/sdk-container-builds/issues/418#issuecomment-1832552645), so marking this up for review.

MichalPavlik (approved):

PR #37223, opened 2023-11-29
Fix install state file for MSI based installs

joeloff:
In addressing the final PR feedback, a number of mistakes slipped in that won't be caught by the unit tests because this is running elevated and we don't have a testable installation until code flows to installer.

- Fix the `File.Exists` check
- Add the `default.json` filename to the actual path. Originally the full path was send through the dispatcher, but in subsequent iteration it was decided to just send the feature band, so the message handlers need to set up the full path.


dsplaisted (approved):
Can you link the previous PR that this is fixing up?  Did that only go into 8.0.200, or is it also in 8.0.100?

joeloff:
We merged the original 8.0.1xx PR (which is marked as do not merge) through a backport with additional fixes/cleanup, so the changes are currently only in 8.0.2xx: https://github.com/dotnet/sdk/pull/36748

I'm almost inclined to close the 8.0.1xx PR and then we can decide if we want this for 8.0.1xx

Issue #26824, opened 2022-07-27
runtime-options (--roll-forward) for local tools


KalleOlaviNiemitalo:
### Is your feature request related to a problem? Please describe.
If a .NET tool does not allow [major-version runtime roll forward](https://docs.microsoft.com/dotnet/core/whats-new/dotnet-core-3-0#major-version-runtime-roll-forward) and is run without any of the runtimes that it targets, it will fail. I would like a `dotnet` command-line option that overrides this setting.

### Describe the solution you'd like
Allow <var>runtime-options</var> in `dotnet` commands that run local tools. For example, `dotnet --fx-version 6.0.7 sarif` or `dotnet tool run --roll-forward Major sarif`.

A similar effect can already be achieved by setting the [`DOTNET_ROLL_FORWARD` environment variable](https://docs.microsoft.com/dotnet/core/tools/dotnet-environment-variables#dotnet_roll_forward), which works with both global and local tools, as well as other .NET applications. However, that would be inherited by child processes too, unlike the command-line option. Inheritance is a good thing if the child process runs an executable that belongs to the same tool, but perhaps not if the child process runs an executable whose path the user specifies as an argument of the tool.

### Additional context
`sarif` in the previous examples refers to [Sarif.Multitool 2.4.16](https://www.nuget.org/packages/Sarif.Multitool/2.4.16), which does not allow major-version runtime roll forward.  I filed <https://github.com/microsoft/sarif-sdk/issues/2511> about fixing it.  <https://github.com/dotnet/sdk/issues/26417#issuecomment-1172672371> suggests a build warning for tools like that.

If a future version of .NET SDK allows an unmanaged executable to be packaged and run as a .NET tool (i.e. without `Command/@Runner="dotnet"` in `DotnetToolSettings.xml`), then it will not be possible to support <var>runtime-options</var> for that.

This is starting to look like a bad idea overall.

baronfel:
System.CommandLine also hit this with their dotnet-suggest tool, as logged in https://github.com/dotnet/command-line-api/issues/2014

marcpopMSFT:
Rather than default to rollforward like #30336 says, we plan on adding an option so it allows opting in at install time.

JL03-Yue:
Close as merged in #37231

PR #37231, opened 2023-11-29
dotnet tool run and install roll-forward option

JL03-Yue:
> ! This pr does not contains the detection of compatible runtimes as it is still being discussed

### dotnet tool run `--allow-roll-forward` option

For `dotnet tool run` command, a `--allow-roll-forward` Option to parse the `--allow-roll-forward` option anywhere in the command using `System.CommandLine` is included, and it feed the value of this option (if any) into the actual `Process.Start` calls to launch tools (e.g. in the `ToolRunCommand`. Note that the `--allow-roll-forward` parsed is always inserted into the 'final' dotnet command first for positional argument purposes. 

**Use case**
When the user has the local tool `sample` installed, and they invoke is using `dotnet tool run sample --allow-roll-forward LatestMajor`. This code rewrites that into `dotnet --roll-forward Major <path to sampleEntryPoint.dll in the tool storage location>`. 

### dotnet tool install `--allow-roll-forward` option for global tools

For `dotnet tool install` command for global tools, a `--allow-roll-forward` option to parse the `--allow-roll-forward` option anywhere in the dotnet tool install command for global tools is included. It updates the runtime config file for the tool to include a `rollForward` property set to `Major` according to the documentation of [roll forward selection](https://learn.microsoft.com/en-us/dotnet/core/versions/selection). 

**Use case**
When the user is installing the global tool sample, and they install it using `dotnet tool install sample -g --allow-roll-forward`. This code will rewrite the runtimeConfig the for tool and set `rollForward` property to `Major`.

### dotnet tool install `--allow-roll-forward` option for local tools

For `dotnet tool install` command for local tools, a `--allow-roll-forward` option to parse the `--allow-roll-forward` option anywhere in the dotnet tool install command for local tools is included. It stores the rollForward state in the local tool manifest file using the `Add` function in `ToolManifestEditor`, and read from the `ToolManifestPackage` using `ToolManifestFinder`. The `rollForward` is set to `Major` according to the documentation of [roll forward selection](https://learn.microsoft.com/en-us/dotnet/core/versions/selection). 

Note that for tools in the same directory with same tool command, it returns the tool that has been added first. For tools installed then later installed with the `--allow-roll-forward` option. It updates the manifest to change the rollForward property to True. Note that to run the tool with `dotnet [toolname]`, the rollForward does not work. 

**Use case**
When the user is installing the local tool sample, and they install it using `dotnet tool install sample --allow-roll-forward`. This code will store the state in the manifest file and read from manifest file when running `dotnet tool run [sampleToolCommand]`


JL03-Yue:
> Seems related to #26824 but that one is for all runtime-options, not just --roll-forward.

Yes, it is a similar solution. The other issue discussion is at #30336.

KalleOlaviNiemitalo (commented):
I'd want to use `Major` roll-forward rather than `LatestMajor`, to minimize the risk that installing a later version of .NET Runtime breaks some tool that previously worked via roll-forward.  [RollForwardOptionDescription in install/LocalizableStrings.resx](https://github.com/dotnet/sdk/blob/c0b0f817871d598de607d5168c9a1db441bb160a/src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx#L238-L240) looks like `--roll-forward=Major` should work; but [ToolInstallCommandParser.RollForwardOption](https://github.com/dotnet/sdk/blob/c0b0f817871d598de607d5168c9a1db441bb160a/src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallCommandParser.cs#L59-L62) is defined as a `CliOption<bool>`, and [ToolPackageDownloader.UpdateRuntimeConfigFile](https://github.com/dotnet/sdk/blob/c0b0f817871d598de607d5168c9a1db441bb160a/src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs#L260) can only set "LatestMajor".

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -193,6 +203,66 @@ internal class ToolPackageDownloader : IToolPackageDownloader
...
+            if (commandElement != null)
+            {
+                XAttribute entryPointAttribute = commandElement.Attribute("EntryPoint");
+                if (entryPointAttribute != null)
+                {
+                    runtimeConfigFileFolder = entryPointAttribute.Value;
+                }
+            }
+
+            var runtimeConfigFilePath = Path.Combine(Path.GetDirectoryName(settingFilePath), $"{Path.GetFileNameWithoutExtension(runtimeConfigFileFolder)}{".runtimeconfig.json"}");
```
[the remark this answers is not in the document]
JL03-Yue: Thanks for the suggestion.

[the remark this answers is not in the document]
JL03-Yue: Resolved.

KalleOlaviNiemitalo (changes requested):

src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
```diff
@@ -235,4 +235,7 @@ If you would like to create a manifest, use `dotnet new tool-manifest`, usually
   <data name="CreateManifestIfNeededOptionDescription" xml:space="preserve">
     <value>Create a tool manifest if one isn't found during tool installation. For information on how manifests are located, see https://aka.ms/dotnet/tools/create-manifest-if-needed</value>
   </data>
+  <data name="RollForwardOptionDescription" xml:space="preserve">
+    <value>Roll forward to framework version  (LatestPatch, Minor, LatestMinor, Major, LatestMajor, Disable).</value>
```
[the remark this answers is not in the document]
JL03-Yue: resolved

dsplaisted:
Some feedback on the design before I look at the code:

From the description, it sounds like in some context `--roll-forward` needs a parameter to specify the roll forward policy, and in other contexts it doesn't.  I think instead of adding `--roll-forward`, we should instead add an `--allow-roll-forward` option which does not take a parameter, and always means the roll-forward policy will be `Major`.  In the context of installing a tool authored by someone else, I don't think the additional complexity of the different roll-forward options is worth exposing.

For `dotnet tool run sample --allow-roll-forward`, it could be ambiguous whether `--allow-roll-forward` is an option for the .NET CLI or whether it's a parameter that should be passed to the tool itself.  Because of this, should we recommend or require this parameter to come before the tool name, ie `dotnet tool run --allow-roll-forward sample`?

@baronfel @richlander @dotnet/dotnet-cli Any thoughts?

> Note that to run the tool with `dotnet [toolname]`, the rollForward does not work.

Are you saying `dotnet tool run [toolname]` would work but `dotnet [toolname]` won't?  What's the reason for this, and can we fix it?

ViktorHofer:
I think this issue closes https://github.com/dotnet/sdk/issues/10375?

src/Tests/dotnet.Tests/CommandTests/ToolRunCommandTests.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        private readonly IFileSystem _fileSystem;
+        private const string ManifestFilename = "dotnet-tools.json";
+        private readonly string _testDirectoryRoot;
+        private DirectoryPath _nugetGlobalPackagesFolder;
+        private readonly LocalToolsResolverCache _localToolsResolverCache;
+
+        public ToolRunCommandTests() 
+        {
+            _fileSystem = new FileSystemMockBuilder().UseCurrentSystemTemporaryDirectory().Build();
```
dsplaisted, line 28: When writing new tests, I think we should prefer deriving from `SdkTest` and using `_testAssetsManager.CreateTestDirectory` instead of using the temporary directory.  This means we won't be able to set up as much in the constructor (we can use a helper method instead), but makes it easier to debug / investigate tests, since they use a predictable directory that's easier to find.

src/Tests/dotnet.Tests/CommandTests/ToolRunCommandTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        }
+
+        [Fact]
+        public void WhenRunWithRollForwardOptionItShouldIncludeRollForwardInNativeHost()
+        {
+            var parseResult = Parser.Instance.Parse($"dotnet tool run $TOOLCOMMAND$ --roll-forward Major");
+
+            var toolRunCommand = new ToolRunCommand(
+            parseResult);
```
dsplaisted, line 44: Nit: This doesn't seem like it needs to be two lines.  For statements that do go over one line, usually the lines after the first one should be indented for better readability.

dsplaisted (commented):
Great work so far.  See also my comments on the design, where I think we should use `--allow-roll-forward` instead of `--roll-forward` for the options we're adding.

There are probably also more test scenarios we should cover if it's not too hard.  We can discuss those sometime.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -193,6 +203,66 @@ internal class ToolPackageDownloader : IToolPackageDownloader
             lockFileLib.ToolsAssemblies.AddRange(toolsGroup);
         }
 
+        private static void UpdateRuntimeConfig(
+            PackageId packageId,
+            NuGetVersion packageVersion,
+            DirectoryPath toolDownloadDir
+            )
+        {
+            var toolsDirectory = Path.Combine(toolDownloadDir.ToString().Trim('"'), packageId.ToString(), packageVersion.ToString(), "tools");
```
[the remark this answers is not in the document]
JL03-Yue: The path generated will have `"` quotation marks at sides. When combining paths, the quotation marks will be together combined into the final path if not trimmed. For example, the final path will be `""first path"/second_path..."` if not trimmed

src/Cli/dotnet/ToolManifest/ToolManifestEditor.cs
```diff
@@ -379,8 +395,8 @@ public string ToJson()
                         {
                             writer.WriteStringValue(toolCommandName);
                         }
-
                         writer.WriteEndArray();
+                        writer.WriteString(JsonPropertyRollForward, tool.RollForward.ToString());
```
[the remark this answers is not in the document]
JL03-Yue: Thank you for the suggestion!

richlander:
In many cases, people will figure out that they need roll-forward after they install a tool. It would be very useful to present an error that a given tool is unlikely to run since (A) it requires a .NET version that isn't installed, and (B) no `RollForward` value is set. Such an error could provide the user with instructions on how either `--force` or `--roll-forward` the installation.

marcpopMSFT:
> In many cases, people will figure out that they need roll-forward after they install a tool. It would be very useful to present an error that a given tool is unlikely to run since (A) it requires a .NET version that isn't installed, and (B) no `RollForward` value is set. Such an error could provide the user with instructions on how either `--force` or `--roll-forward` the installation.

I believe the plan was to add this option first and then follow up with a PR to add an error/warning message at install/update time that let's the customer know that this option now exists. That could be rolled into this PR but it's already pretty big already.

richlander:
Cool. Just wanted to ensure that this aspect of the DX wasn't forgotten.

src/Tests/dotnet.Tests/CommandTests/ToolRunCommandTests.cs
JL03-Yue: Thanks for the suggestion

src/Cli/dotnet/CommandFactory/CommandResolution/MuxerCommandSpecMaker.cs
```diff
@@ -20,12 +20,22 @@ internal static class MuxerCommandSpecMaker
             {
                 throw new Exception(LocalizableStrings.UnableToLocateDotnetMultiplexer);
             }
+            IEnumerable<string> modifiedArguments = commandArguments;
+
+            // Add --roll-forward argument first if exists
+            if (commandArguments != null && commandArguments.Any(arg => arg.Equals("--roll-forward", StringComparison.OrdinalIgnoreCase)))
```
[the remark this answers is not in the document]
JL03-Yue: Thank you so much for the comment. Could you elaborate more on the System.CommandLine APIs?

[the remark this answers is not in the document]
baronfel: @dsplaisted I think in these muxers for tool invocation we are outside of S.CL entirely - this is something I'd like to change longer term but is probably outside the scope of this change currently.

src/Cli/dotnet/CommandFactory/CommandResolution/MuxerCommandSpecMaker.cs
```diff
@@ -20,12 +20,22 @@ internal static class MuxerCommandSpecMaker
...
+                modifiedArguments = commandArguments.Where((element, i) => i != index && i != index + 1);
+            }
 
             arguments.Add(commandPath);
 
-            if (commandArguments != null)
+            if (modifiedArguments != null)
             {
-                arguments.AddRange(commandArguments);
+                arguments.AddRange(modifiedArguments);
```
MiYanni, line 38: GitHub doesn't let me comment on the entire method, but this is how I'd write it. It isn't necessary to do all the manual reshuffling of stuff. The only thing you need to know is if the current or previous argument is `--roll-forward`. So, I just grouped them by that predicate and then create the result set with the values in the grouped subsets.

```C#
        internal static CommandSpec CreatePackageCommandSpecUsingMuxer(
            string commandPath,
            IEnumerable<string> commandArguments)
        {
            var muxer = new Muxer();
            var host = muxer.MuxerPath;
            if (host == null)
            {
                throw new Exception(LocalizableStrings.UnableToLocateDotnetMultiplexer);
            }
            string previousArg = string.Empty;
            // Group the arguments by if the previous argument or the current argument is --roll-forward.
            // The use of bitwise OR is necessary here as we don't want to short-circuit the boolean calculation since previousArg always needs to be reassigned.
            var argGroups = (commandArguments ?? [])
                .GroupBy(a => previousArg.Equals("--roll-forward", StringComparison.OrdinalIgnoreCase) | (previousArg = a).Equals("--roll-forward", StringComparison.OrdinalIgnoreCase))
                .ToArray();

            string[] arguments = [.. argGroups.Where(g => g.Key).SelectMany(g => g), commandPath, .. argGroups.Where(g => !g.Key).SelectMany(g => g)];
            return CreateCommandSpec(host, arguments);
        }
```

Here's the test console app I wrote for it:
```C#
string commandPath = "whatever";
IEnumerable<string> commandArguments = new List<string> { "a", "b", "--roll-forward", "value", "c", "d", "e" };

string previousArg = string.Empty;
// Group the arguments by if the previous argument or the current argument is --roll-forward.
// The use of bitwise OR is necessary here as we don't want to short-circuit the boolean calculation since previousArg always needs to be reassigned.
var argGroups = (commandArguments ?? [])
    .GroupBy(a => previousArg.Equals("--roll-forward", StringComparison.OrdinalIgnoreCase) | (previousArg = a).Equals("--roll-forward", StringComparison.OrdinalIgnoreCase))
    .ToArray();

List<string> result = [..argGroups.Where(g => g.Key).SelectMany(g => g), commandPath, ..argGroups.Where(g => !g.Key).SelectMany(g => g)];

result.ForEach(Console.WriteLine);
```

The output of the console app looks like this:
```
--roll-forward
value
whatever
a
b
c
d
e
```

JL03-Yue: Thank you so much for the comment!

src/Cli/dotnet/commands/dotnet-tool/run/LocalizableStrings.resx
```diff
@@ -129,6 +129,9 @@
   <data name="CommandNameArgumentName" xml:space="preserve">
     <value>COMMAND_NAME</value>
   </data>
+  <data name="RollForwardDescription" xml:space="preserve">
+    <value>Roll forward to framework version  (LatestPatch, Minor, LatestMinor, Major, LatestMajor, Disable).</value>
+  </data>
```
dsplaisted, line 134: I think we can delete this string (which doesn't take into account the change to `--allow-roll-forward`) and use the same `RollForwardOptionDescription` string from the install command.  You should be able to qualify it with the right namespace when you use it so that you don't have to duplicate the string in the resource files.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -193,6 +205,26 @@ internal class ToolPackageDownloader : IToolPackageDownloader
...
 
+        private static void UpdateRuntimeConfig(
+            ToolPackageInstance toolPackageInstance
+            )
+        {
+            var executableFilePath = toolPackageInstance.Commands[0].Executable;
+            var runtimeConfigFilePath = Path.ChangeExtension(executableFilePath.ToString(), ".runtimeconfig.json");
+
+            // Update the runtimeconfig.json file
+            string existingJson = File.ReadAllText(runtimeConfigFilePath);
```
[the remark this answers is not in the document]
JL03-Yue: Thank you for the suggestion!

src/Cli/dotnet/ToolManifest/ToolManifestPackage.cs
```diff
@@ -43,7 +46,8 @@ public bool Equals(ToolManifestPackage other)
                    EqualityComparer<NuGetVersion>.Default.Equals(Version, other.Version) &&
                    CommandNamesEqual(other.CommandNames) &&
                    FirstEffectDirectory.Value.TrimEnd('/', '\\')
-                     .Equals(other.FirstEffectDirectory.Value.TrimEnd('/', '\\'), StringComparison.Ordinal);
+                     .Equals(other.FirstEffectDirectory.Value.TrimEnd('/', '\\'), StringComparison.Ordinal) &&
+                     RollForward.Equals(other.RollForward);
```
JL03-Yue, line 50: Potential tests might be needed for `Equal()` and `GetHashCode()` functions

src/Cli/dotnet/CommandFactory/CommandResolution/LocalToolsCommandResolver.cs
```diff
@@ -73,13 +74,20 @@ public CommandSpec Resolve(CommandResolverArguments arguments)
         }
 
         private CommandSpec GetPackageCommandSpecUsingMuxer(CommandResolverArguments arguments,
-            ToolCommandName toolCommandName)
+            ToolCommandName toolCommandName, bool allowRollForward = false)
         {
             if (!_toolManifest.TryFind(toolCommandName, out var toolManifestPackage))
```
JL03-Yue, line 79: Tests should be added if allowRollForward is false in `LocalToolsCommandResolver`

JL03-Yue (commented):
Some more tests to be covered

src/Cli/dotnet/CommandFactory/CommandResolution/LocalToolsCommandResolver.cs
```diff
@@ -73,13 +74,20 @@ public CommandSpec Resolve(CommandResolverArguments arguments)
...
             {
                 return null;
             }
 
+            if( arguments.CommandArguments != null)
+            {
+                var commandArgument = (allowRollForward != false ? new List<string> { "--roll-forward", "Major" } : Enumerable.Empty<string>()).Concat(arguments.CommandArguments);
+                arguments.CommandArguments = commandArgument;
+            }
```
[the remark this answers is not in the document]
JL03-Yue: Thank you for the suggestion. Indeed, more tests should be covered

baronfel (approved):
Thank you for the detailed description and coverage of the various scenarios! I'm excited for this first step along the road of making tools  easier to use.

src/Tests/dotnet.Tests/CommandTests/ToolRunCommandTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            var toolRunCommand = new ToolRunCommand(parseResult);
+
+            (FilePath fakeExecutable, LocalToolsCommandResolver localToolsCommandResolver) = DefaultSetup("dotnet-a");
+            IEnumerable<string> testForwardArgument = Enumerable.Empty<string>();
+
+            var result = localToolsCommandResolver.Resolve(new CommandResolverArguments()
+            {
+                CommandName = "dotnet-a",
+                CommandArguments = (toolRunCommand._allowRollForward != false ? new List<string> { "--roll-forward", "Major" } : Enumerable.Empty<string>()).Concat(testForwardArgument)
```
dsplaisted, line 41: I'm not sure but it seems like you shouldn't need to conditionally be adding `--roll-forward Major` to the CommandArguments here.  I think the `LocalToolsCommandResolver` should be doing that based on what was parsed from the command.  It looks like the parseResult and the toolRunCommand are not actually used though, so I'm not sure what's going on here.

src/Cli/dotnet/ToolPackage/ToolPackageDownloader.cs
```diff
@@ -193,6 +205,32 @@ internal class ToolPackageDownloader : IToolPackageDownloader
             lockFileLib.ToolsAssemblies.AddRange(toolsGroup);
         }
 
+        private static void UpdateRuntimeConfig(
+            ToolPackageInstance toolPackageInstance
+            )
+        {
+            for (int i = 0; i < toolPackageInstance.Commands.Count; i++)
+            {
+                var command = toolPackageInstance.Commands[i];
```
dsplaisted, line 214: Nit: Looks like this can be simplified with a `foreach` loop.

```suggestion
            foreach (var command in toolPackageInstance.Commands)
            {
```

src/Tests/dotnet.Tests/CommandTests/ToolRunCommandTests.cs
JL03-Yue: Thank you so much. I updated these two tests in `ToolRunCommandTests` to use `ResolveStrict` method in the `LocalToolsCommandResolver`, which is the method to be used in `ToolRunCommand`. The state for roll forward is determined when constructing the `toolRunCommand`, get passed in the `GetPackageCommandSpecUsingMuxer` function in the `LocalToolsCommandResolver`, and the `--roll-forward Major` in the native host is added in the `MuxerCommandSpecMaker`. Sorry for being unclear in the test and hopefully the updated tests are clearer!

src/Cli/dotnet/CommandFactory/CommandResolution/MuxerCommandSpecMaker.cs
```diff
@@ -11,23 +11,23 @@ internal static class MuxerCommandSpecMaker
...
-            {
-                arguments.AddRange(commandArguments);
-            }
+            // Group the arguments by if the previous argument or the current argument is --roll-forward.
+            var argGroups = (commandArguments ?? [])
+                .GroupBy(a => previousArg.Equals("--roll-forward", StringComparison.OrdinalIgnoreCase)
+                | (previousArg = a).Equals("--roll-forward", StringComparison.OrdinalIgnoreCase))
+                .ToArray();
 
+            string[] arguments = [..argGroups.Where(g => g.Key).SelectMany(g => g), commandPath, .. argGroups.Where(g => !g.Key).SelectMany(g => g)];
```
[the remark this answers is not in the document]
JL03-Yue: Thank you so much! The logic is updated based on your comment.

Previously, the `--roll-forward Major` arguments are added in the `GetPackageCommandSpecUsingMuxer` function in the `LocalToolsCommandResolver`. In the `MuxerCommandSpecMaker`, the `--roll-forward Major` get pulled out from the list of the arguements and get placed in the front of all arguments, as positioning arguments for running native code requires. The updated code adds `--allow-roll-forward` in the `GetPackageCommandSpecUsingMuxer` function, and "translated" to `--roll-forward Major` in `MuxerCommandSpecMaker` as your code suggests. Thank you for the comment and this simplifies the logic and clarity while maintaining the feature!

Issue #34949, opened 2023-08-25
dotnet watch test doesn't run tests in dotnet 8


yori-s:
### Describe the bug
**dotnet test** works fine but **dotnet watch test** doesn't execute any tests.

### To Reproduce
Create a failing test (using docker below)
```
docker run --rm -it --name dotnet8-watch mcr.microsoft.com/dotnet/sdk:8.0.100-preview.7-jammy bash

mkdir /app && cd $_
dotnet new xunit
cat > UnitTest1.cs <<- EOF
namespace app;

public class UnitTest1
{
  [Fact] public void Fail() => Assert.True(false);
}
EOF
```

`dotnet watch test` just sits with something like the following:
```
dotnet watch : Polling file watcher is enabled
dotnet watch : Hot reload enabled. For a list of supported edits, see https://aka.ms/dotnet/hot-reload.
   Press "Ctrl + R" to restart.
dotnet watch : Building...
  Determining projects to restore...
  All projects are up-to-date for restore.
/usr/share/dotnet/sdk/8.0.100-preview.7.23376.3/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.RuntimeIdentifierInference.targets(314,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy [/app/app.csproj]
  app -> /app/bin/Debug/net8.0/app.dll
dotnet watch : Started
dotnet watch : Exited
dotnet watch : Waiting for a file to change before restarting dotnet...
```

Works fine in dotnet 7 (`mcr.microsoft.com/dotnet/sdk:7.0-jammy`)

### Further technical details
- Include the output of `dotnet --info`
   ```
   .NET SDK:
    Version:   8.0.100-preview.7.23376.3
    Commit:    daebeea8ea
 
   Runtime Environment:
    OS Name:     ubuntu
    OS Version:  22.04
    OS Platform: Linux
    RID:         ubuntu.22.04-x64
    Base Path:   /usr/share/dotnet/sdk/8.0.100-preview.7.23376.3/
 
   .NET workloads installed:
   There are no installed workloads to display.
 
   Host:
     Version:      8.0.0-preview.7.23375.6
     Architecture: x64
     Commit:       65b696cf5e
     RID:          linux-x64
 
   .NET SDKs installed:
     8.0.100-preview.7.23376.3 [/usr/share/dotnet/sdk]
 
   .NET runtimes installed:
[...]

PR #31054, opened 2023-03-08
Refactor command line options parsing, add/improve imp of --launch-profile and --no-launch-profile options.

tmat:
Adds support for `--framework` and `--property` command line options.

Fixes https://github.com/dotnet/aspnetcore/issues/45629
Fixes https://github.com/dotnet/aspnetcore/issues/38408
Fixes https://github.com/dotnet/aspnetcore/issues/39443
Fixes https://github.com/dotnet/aspnetcore/issues/38408
Fixes https://github.com/dotnet/aspnetcore/issues/40472
Fixes https://github.com/dotnet/aspnetcore/issues/33726

Might fix https://github.com/dotnet/aspnetcore/issues/37504 -- needs validation.

PR #37233, opened 2023-11-30
Add support for test and build commands

tmat:
Fixes https://github.com/dotnet/sdk/issues/34949

Handle `test` and `build` as valid commands that may be specified on `dotnet watch` command line.

The original implementation of command line parsing allowed `dotnet watch test` and `dotnet watch build` to work, although without any tests. PR https://github.com/dotnet/sdk/pull/31054 broke this when fixing other command line parsing issues.

## Customer Impact

Customers unable to use `dotnet watch test` and `dotnet watch build` with .NET SDK 8.

## Regression?

- [x] Yes
- [ ] No

7.0

## Risk

- [ ] High
- [x] Medium
- [] Low

Some risk of breaking command line parsing for other scenarios that we don't have test coverage for.

## Verification

- [x] Manual (required)
- [x] Automated

- [ ] Yes
- [x] No
- [ ] N/A

### DIFF
--- a/src/Assets/TestProjects/WatchNoDepsApp/WatchNoDepsApp.csproj
+++ b/src/Assets/TestProjects/WatchNoDepsApp/WatchNoDepsApp.csproj
@@ -15,4 +15,8 @@
     <WarningsNotAsErrors>CS9057</WarningsNotAsErrors>
   </PropertyGroup>
 
+  <Target Name="TestTarget">
+    <Warning Text="The value of property is '$(TestProperty)'" />
+  </Target>
+
 </Project>
--- a/src/BuiltInTools/dotnet-watch/CommandLineOptions.cs
+++ b/src/BuiltInTools/dotnet-watch/CommandLineOptions.cs
@@ -8,13 +8,6 @@ using Microsoft.Extensions.Tools.Internal;
 
 namespace Microsoft.DotNet.Watcher;
 
-internal sealed class RunCommandLineOptions
-{
-    public required bool NoLaunchProfile { get; init; }
-    public required string? LaunchProfileName { get; init; }
-    public required IReadOnlyList<string> RemainingArguments { get; init; }
-}
-
 internal sealed class CommandLineOptions
 {
     private const string Description = @"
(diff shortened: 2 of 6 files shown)


src/BuiltInTools/dotnet-watch/WatchCommandLineOptions.cs
```diff
@@ -136,69 +138,74 @@ console output.
         };
 
         var remainingWatchArgs = new CliArgument<string[]>("forwardedArgs") { Description = "Arguments to pass to the child dotnet process." };
-        var remainingRunArgs = new CliArgument<string[]>("remainingRunArgs");
+        var remainingCommandArgs = new CliArgument<string[]>("remainingRunArgs");
 
         var runCommand = new CliCommand("run") { Hidden = true };
+        var testCommand = new CliCommand("test") { Hidden = true };
+        var buildCommand = new CliCommand("build") { Hidden = true };
```
baronfel, line 145: this implies that this is the only set of supported commands - why not just have an `string[]` catch-all that is whatever command the user wants to re-trigger on change?

baronfel: Ok, is there a catch-all for all other commands that allows for executing arbitrary commands as well? That's IMO what users expect out of a `watch` command.

tmat: Another issue is that `dotnet watch x y z` runs the app in Hot Reload mode with arguments `x`, `y` and `z`.
We can't run `build` and `test` commands in Hot Reload mode. So these commands need to be special-cased.

This is an unfortunate design. It'd be better to have a separate verb for Hot Reload instead of using `watch`. 
E.g. `dotnet hotreload`.

yori-s: 👍 to `dotnet hotreload` - then wouldn't have had this issue in the first place

MarkusRodler: I think it would also be clearer for the users to have separate commands

davidwengier (approved):

yori-s:
🍾

dgilleland:
I note that the *Risk* is "Medium", but the inconvenience is "High"...

laurentpayot:
@dgilleland if NodeJs is available I guess you can use [Chokidar CLI](https://github.com/open-cli-tools/chokidar-cli) as a workaround.

yori-s:
mostly echoing @laurentpayot - you can also get the watchlist from `dotnet watch --list` (I think I passed it to nodemon, which uses chokidar)

justinmchase:
While we're at it can we request that the build and test commands just get a --watch arg instead of requiring a watch command?

PR #37310, opened 2023-12-04
Minimize and update MSBuild in dotnet-watch

rainersigwald:
Dotnet-watch was distributing copies of MSBuild assemblies, including StringTools, from the older (VS compatibility) MSBuild version
referenced repo-wide instead of the "current" MSBuild distributed with the SDK itself.

Added central versions for the MSBuild packages that weren't there already, referenced them in the redist and tool_msbuild projects, and updated dotnet-watch to stop distributing the other MSBuild assemblies.

This should fix the test failures in #37242 once it merges around.


Forgind (approved):

rainersigwald:
Container build stuff and F# redist MSBuild and we should ideally fix that.

PR #37311, opened 2023-12-04
Add automatically tool upgrade and downgrade for different versions for dotnet tool install

JL03-Yue:
https://github.com/dotnet/sdk/issues/9500

## Update dotnet tool install for global or local tools so that it support automatic upgrade, downgrade, and error-throwing behaviors.

**behaviors**

- If the command specify version 3.0.0 without --allow-downgrade flag and version 4.0.0 is installed, it throws out an error.
- If the command specify version 3.0.0 with--allow-downgrade flag and version 4.0.0 is installed, downgrade to 3.0.0. Exit code = 0

**User cases**

- Tool install
```
$ dotnet__ tool install (-g) Cake.tool --version 3.0.0 --verbosity minimal
You can invoke the tool using the following command: dotnet-cake
Tool 'cake.tool' (version '3.0.0') was successfully installed.
$ dotnet tool install (-g) Cake.tool --version 3.0.0 --verbosity minimal
Tool 'cake.tool' was reinstalled with the stable version (version '3.0.0').
$ dotnet tool install (-g) Cake.tool --verbosity minimal
Tool 'cake.tool' was successfully updated from version '3.0.0' to version '4.0.0'.
$ dotnet tool install (-g) Cake.tool --version 2.0.0 --verbosity minimal
The requested version 2.0.0 is lower than existing version 4.0.0.
$ dotnet tool install (-g) Cake.tool --version 2.0.0 --allow-downgrade --verbosity minimal
Tool 'cake.tool' was successfully updated from version '4.0.0' to version '2.0.0'.
```

- Tool update
```
$ dotnet tool install (-g) Cake.tool --version 3.0.0 --verbosity minimal
You can invoke the tool using the following command: dotnet-cake
Tool 'cake.tool' (version '3.0.0') was successfully installed.
$ dotnet tool update (-g) Cake.tool --verbosity minimal
Tool 'cake.tool' was successfully updated from version '3.0.0' to version '4.0.0'.
$ dotnet tool update (-g) Cake.tool --version 4.0.0 --verbosity minimal
Tool 'cake.tool' was reinstalled with the stable version (version '4.0.0').
$ dotnet tool update (-g) Cake.tool --version 2.0.0 --verbosity minimal
The requested version 2.0.0 is lower than existing version 4.0.0.
```


src/Tests/dotnet.Tests/CommandTests/ToolUpdateGlobalOrToolPathCommandTests.cs
```diff
@@ -152,6 +152,21 @@ public void GivenAnExistedLowerversionInstallationWhenCallItCanPrintSuccessMessa
                 _packageId, LowerPackageVersion, HigherPackageVersion));
         }
 
+        [Fact]
+        public void GivenAnExistedHigherversionInstallationWhenUpdateToLowerVersionItErrors()
```
dsplaisted, line 156: We should probably have a test for this same scenario succeeding when `--allow-downgrade` is specified.  Likewise, we should have the same two scenarios for local tools.

dsplaisted (approved):
Looks great!

We should probably have test cases for each of the different behaviors in the PR description, for global and local tools.  Probably some of those cases are already covered.

src/Cli/dotnet/commands/dotnet-tool/update/ToolUpdateGlobalOrToolPathCommand.cs
```diff
@@ -2,272 +2,52 @@
...
-                    if (!_verbosity.IsQuiet())
-                    {
-                        PrintSuccessMessage(oldPackageNullable, newInstalledPackage);
-                    }
-                });
-
-                scope.Complete();
-            }
-
+            _toolInstallGlobalOrToolPathCommand.Value.Execute();
```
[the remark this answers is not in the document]
JL03-Yue: Yes, the `dotnet tool install` and `dotnet tool update` are of similar function. The options are different.

src/Tests/dotnet.Tests/CommandTests/ToolUpdateGlobalOrToolPathCommandTests.cs
JL03-Yue: Do you think the `--allow-downgrade` option should be added to tool update (apart from install) as well? Given the inference of the word "update"

src/Cli/dotnet/commands/dotnet-tool/update/ToolUpdateGlobalOrToolPathCommand.cs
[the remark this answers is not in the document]
JL03-Yue: Thank you. The options are consolidated.

src/Tests/dotnet.Tests/CommandTests/ToolUpdateGlobalOrToolPathCommandTests.cs
JL03-Yue: The `--allow-downgrade` option is added to tool update for this pr

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
```diff
@@ -80,65 +80,70 @@ internal class ToolInstallGlobalOrToolPathCommand : CommandBase
...
+
+            using (var scope = new TransactionScope(
+                TransactionScopeOption.Required,
+                TimeSpan.Zero))
             {
-                IToolPackage package = null;
-                using (var scope = new TransactionScope(
-                    TransactionScopeOption.Required,
-                    TimeSpan.Zero))
+                if (oldPackageNullable != null)
```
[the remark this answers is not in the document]
JL03-Yue: This exception will be caught by `ToolUninstallCommandLowLevelErrorConverter.ShouldConvertToUserFacingError(ex)`, if I tested correctly.

src/Tests/dotnet.Tests/CommandTests/ToolUpdateLocalCommandTests.cs
```diff
@@ -311,6 +311,29 @@ public void GivenFeedVersionIsLowerRunPackageIdItShouldThrow()
                 _manifestFilePath));
         }
 
+        [Fact]
+        public void GivenFeedVersionIsLowerWithDowngradeFlagRunPackageIdItShouldSucceeds()
```
dsplaisted, line 315: In this case, will it downgrade the package?  I think it probably shouldn't.

It looks like the scenario is when the latest version of the package on the feed is lower than the version currently installed.  In that case, I think the update command should succeed but not actually change the version installed.  This should be the case whether or not the `--allow-downgrade` option is specified.

src/Cli/dotnet/commands/dotnet-tool/install/ToolInstallGlobalOrToolPathCommand.cs
```diff
@@ -114,66 +115,76 @@ public override int Execute()
...
-            }
+                    scope.Complete();
+                }
+                catch (Exception)
+                {
+                    // Log the exception or perform other error handling if needed.
+                    // Rollback the transaction to ensure consistency.
+                    scope.Dispose();
+                    throw;
+                }
```
dsplaisted, line 186: Is this actually needed?  The `using` block should ensure that the scope is disposed when it exits that block, whether or not it happens via an exception.

src/Cli/dotnet/commands/dotnet-tool/update/LocalizableStrings.resx
```diff
@@ -218,10 +218,10 @@ and the corresponding package Ids for installed tools using the command
   <data name="UpdateLocalToolSucceeded" xml:space="preserve">
     <value>Tool '{0}' was successfully updated from version '{1}' to version '{2}' (manifest file {3}).</value>
   </data>
-  <data name="UpdateLocaToolToLowerVersion" xml:space="preserve">
-    <value>The requested version {0} is lower than existing version {1} (manifest file {2}).</value>
+  <data name="UpdateLocalToolToLowerVersion" xml:space="preserve">
+    <value>The requested version {0} is lower than existing version {1} (manifest file {2}). Use the --allow-downgrade option to allow this updated.</value>
```
dsplaisted, line 222: I think there was a typo in my previous suggestion.

```suggestion
    <value>The requested version {0} is lower than existing version {1} (manifest file {2}). Use the --allow-downgrade option to allow this update.</value>
```

dsplaisted (approved):

src/Tests/dotnet.Tests/CommandTests/ToolUpdateLocalCommandTests.cs
JL03-Yue: Thank you! I added a `--version` flag into the test and it should refer to the package with the specified lowered version.

dsplaisted: Great.  Are we sure the behavior where you don't specify the version flag is that it won't downgrade?

PR #37312, opened 2023-12-04
Enable parallel copy in PublishDotnetWatchToRedist

rainersigwald:
The performance improvement here is negligible (~200ms->70ms with N=1 on my machine), but SDK should be an example of good MSBuild practices, and I happened to see this, so I'd like to tighten it up.


baronfel (approved):
Thank you!

PR #37346, opened 2023-12-06
Permit setting whether to use workload sets or loose manifests Fixes #34561

### LINKED ISSUE #34561: Add flag for --workload-state when installing/update workloads for the old install behavior
`dotnet workload update --workload-state global` state saved into programdata on whether to use workload sets or not

Forgind:
Fixes #34561

Adds a new subcommand (dotnet workload update --mode) taking one of three arguments: workloadset, loosemanifest, or auto. This will set the relevant bool in the install state according to that value.

That bool is not yet used.

### DIFF
--- a/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
+++ b/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
@@ -33,7 +33,7 @@ namespace Microsoft.DotNet.Installer.Windows
         /// The contents of the install state file. Each element corresponds to a single line of
         /// the JSON file to be written.
         /// </summary>
-        public string[] InstallStateContents
+        public Dictionary<string, string> InstallStateContents
         {
             get;
             set;
--- a/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
+++ b/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
@@ -12,6 +12,15 @@ namespace Microsoft.DotNet.Installer.Windows
     internal class InstallRequestMessage : InstallMessageBase
     {
         /// <summary>
+        /// Path to the directory with the dotnet executable
+        /// </summary>
+        public string DotnetDir
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
         /// The dependent value to add or remove when updating reference counts.
         /// </summary>
         public string Dependent
--- a/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
+++ b/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
@@ -121,6 +121,14 @@ namespace Microsoft.DotNet.Installer.Windows
         }
 
         /// <summary>
+        /// The new mode to use: workloadset or loosemanifests
+        /// </summary>
+        public bool WorkloadMode
+        {
+            get; set;
+        }
+
+        /// <summary>
         /// Converts a deserialized array of bytes into an <see cref="InstallRequestMessage"/>.
         /// </summary>
         /// <param name="bytes">The array of bytes to convert.</param>
--- a/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
+++ b/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
@@ -12,15 +12,6 @@ namespace Microsoft.DotNet.Installer.Windows
     internal class InstallRequestMessage : InstallMessageBase
     {
         /// <summary>
-        /// Path to the directory with the dotnet executable
-        /// </summary>
-        public string DotnetDir
-        {
-            get;
-            set;
-        }
-
-        /// <summary>
         /// The dependent value to add or remove when updating reference counts.
         /// </summary>
         public string Dependent
--- a/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
+++ b/src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
@@ -33,7 +33,7 @@ namespace Microsoft.DotNet.Installer.Windows
         /// The contents of the install state file. Each element corresponds to a single line of
         /// the JSON file to be written.
         /// </summary>
-        public Dictionary<string, string> InstallStateContents
+        public Dictionary<string, string> InstallStateManifestVersions
         {
             get;
             set;
@@ -123,7 +123,7 @@ namespace Microsoft.DotNet.Installer.Windows
         /// <summary>
         /// The new mode to use: workloadset or loosemanifests
         /// </summary>
-        public bool WorkloadMode
+        public bool UseWorkloadSets
         {
             get; set;
         }
--- a/src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
@@ -170,6 +170,9 @@ Tool '{1}' (version '{2}') was successfully installed. Entry is added to the man
   <data name="LocalOptionDescription" xml:space="preserve">
     <value>Install the tool and add to the local tool manifest (default).</value>
   </data>
+  <data name="WorkloadSetMode" xml:space="preserve">
+    <value>Determine whether dotnet should use workload sets or loose manifests when installing or updating workloads.</value>
+  </data>
   <data name="InstallFullCommandNameLocalized" xml:space="preserve">
     <value>.NET Install Command</value>
   </data>
--- a/src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
@@ -170,9 +170,6 @@ Tool '{1}' (version '{2}') was successfully installed. Entry is added to the man
   <data name="LocalOptionDescription" xml:space="preserve">
     <value>Install the tool and add to the local tool manifest (default).</value>
   </data>
-  <data name="WorkloadSetMode" xml:space="preserve">
-    <value>Determine whether dotnet should use workload sets or loose manifests when installing or updating workloads.</value>
-  </data>
   <data name="InstallFullCommandNameLocalized" xml:space="preserve">
     <value>.NET Install Command</value>
   </data>
--- a/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
@@ -6,10 +6,10 @@ namespace Microsoft.DotNet.Workloads.Workload
     internal class InstallStateContents
     {
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-        public bool? useWorkloadSets { get; set; }
+        public bool? UseWorkloadSets { get; set; }
 
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-        public Dictionary<string, string> manifests { get; set; }
+        public Dictionary<string, string> Manifests { get; set; }
 
         public static InstallStateContents FromString(string contents)
         {
--- a/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
@@ -6,10 +6,10 @@ namespace Microsoft.DotNet.Workloads.Workload
     internal class InstallStateContents
     {
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-        public bool? UseWorkloadSets { get; set; }
+        public bool? useWorkloadSets { get; set; }
 
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-        public Dictionary<string, string> Manifests { get; set; }
+        public Dictionary<string, string> manifests { get; set; }
 
         public static InstallStateContents FromString(string contents)
         {
--- a/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
@@ -11,18 +11,20 @@ namespace Microsoft.DotNet.Workloads.Workload
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
         public Dictionary<string, string> Manifests { get; set; }
 
+        private static readonly JsonSerializerOptions s_options = new()
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            WriteIndented = true,
+        };
+
         public static InstallStateContents FromString(string contents)
         {
-            return JsonSerializer.Deserialize<InstallStateContents>(contents);
+            return JsonSerializer.Deserialize<InstallStateContents>(contents, s_options);
         }
 
         public override string ToString()
         {
-            return JsonSerializer.Serialize<InstallStateContents>(this, new JsonSerializerOptions()
-            {
-                PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-                WriteIndented = true,
-            });
+            return JsonSerializer.Serialize<InstallStateContents>(this, s_options);
         }
     }
 }
--- a/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
+++ b/src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
@@ -1,0 +1,24 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.DotNet.Workloads.Workload
+{
+    internal class InstallStateContents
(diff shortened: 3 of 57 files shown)


src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -497,9 +534,9 @@ protected void UpdateDependent(InstallRequestType requestType, string providerKe
         /// Deletes install state file for the specified feature band.
         /// </summary>
         /// <param name="sdkFeatureBand">The feature band of the install state file.</param>
-        protected void RemoveInstallStateFile(SdkFeatureBand sdkFeatureBand)
+        protected void RemoveInstallStateFile(string dotnetDir, SdkFeatureBand sdkFeatureBand)
```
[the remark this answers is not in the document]
Forgind: Great question!

How do we know if we're an MSI-based install or a file-based install? The current check goes to _dotnetDir_ to look for the msi sentinel file. That was the part that was actually throwing previously; as you said, we don't need to know dotnetDir if we're just writing to ProgramData, but we first check if we're an mis-based install, and that check threw an exception.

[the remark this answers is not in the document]
dsplaisted: This is in the MsiInstallerBase class, so we know we're in an MSI-based install already.  We do need to pass the `dotnetDir` to `WorkloadInstallType.GetInstallStateFolder`, but rather than passing it through these methods, I think we can use the `DotNetHome` property, and if not we should be able to figure it out some other way.

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -525,23 +562,23 @@ protected void RemoveInstallStateFile(SdkFeatureBand sdkFeatureBand)
         /// </summary>
         /// <param name="sdkFeatureBand">The path of the isntall state file to write.</param>
         /// <param name="jsonLines">The contents of the JSON file, formatted as a single line.</param>
-        protected void WriteInstallStateFile(SdkFeatureBand sdkFeatureBand, IEnumerable<string> jsonLines)
+        protected void WriteInstallStateFile(string dotnetDir, SdkFeatureBand sdkFeatureBand, Dictionary<string, string> jsonLines)
```
dsplaisted, line 565: Naming: Since this method is no longer being passed all of the data that goes into the install state file, I think it might be clearer if it (and other similar methods) were renamed to something like `SaveInstallStateManifestVersions`.  The `jsonLines` parameter could be renamed to `manifestVersions`.

src/Cli/dotnet/commands/dotnet-workload/install/IInstaller.cs
```diff
@@ -38,14 +38,16 @@ internal interface IInstaller : IWorkloadManifestInstaller
...
 
         /// <summary>
         /// Writes the specified JSON contents to the install state file.
         /// </summary>
         /// <param name="sdkFeatureBand">The SDK feature band of the install state file.</param>
         /// <param name="jsonLines">The JSON contents describing the install state.</param>
-        void WriteInstallState(SdkFeatureBand sdkFeatureBand, IEnumerable<string> jsonLines);
+        void WriteInstallState(string dotnetDir, SdkFeatureBand sdkFeatureBand, Dictionary<string, string> jsonLines);
+
+        void AdjustInstallMode(string dotnetDir, SdkFeatureBand sdkFeatureBand, string newMode);
```
dsplaisted, line 50: Nit: I think `UpdateInstallMode` would be more idiomatic than `AdjustInstallMode`.

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerServer.cs
```diff
@@ -94,15 +94,21 @@ public void Run()
...
 
                         case InstallRequestType.RemoveInstallStateFile:
-                            RemoveInstallStateFile(new SdkFeatureBand(request.SdkFeatureBand));
+                            RemoveInstallStateFile(request.DotnetDir, new SdkFeatureBand(request.SdkFeatureBand));
                             Dispatcher.ReplySuccess($"Deleted install state file for {request.SdkFeatureBand}.");
                             break;
 
+                        case InstallRequestType.AdjustWorkloadMode:
+                            AdjustInstallMode(request.DotnetDir, new SdkFeatureBand(request.SdkFeatureBand), request.ProductCode);
+                            string newMode = request.ProductCode.Equals("true", StringComparison.OrdinalIgnoreCase) ? "workload sets" : "loose manifests";
```
dsplaisted, line 108: The PR description mentions 3 possible values: workloadset, loosemanifest, and auto.  It looks like this only handles the first two, and represents the value as a string.

Could we represent the value as an Enum, or possibly a nullable Boolean?

src/Cli/dotnet/commands/dotnet-workload/install/WorkloadInstallCommand.cs
```diff
@@ -229,7 +229,11 @@ internal static void TryRunGarbageCollection(IInstaller workloadInstaller, IRepo
 
                     if (usingRollback)
                     {
-                        installer.WriteInstallState(sdkFeatureBand, GetInstallStateContents(manifestsToUpdate));
+                        installer.WriteInstallState(_dotnetPath, sdkFeatureBand, GetInstallStateContents(manifestsToUpdate));
+                    }
+                    else
+                    {
+                        installer.DeleteInstallState(_dotnetPath, sdkFeatureBand);
```
dsplaisted, line 236: Now that the install state file can include both the installation mode and the manifest versions, I don't think we want to delete the whole file here.  That would change the installation mode, which we don't want.  Rather, I think we want to remove the manifests from the install state file.

src/Cli/dotnet/commands/dotnet-workload/update/WorkloadUpdateCommand.cs
```diff
@@ -80,6 +82,22 @@ public override int Execute()
...
+                    _workloadInstaller.AdjustInstallMode(_dotnetPath, _sdkFeatureBand, "true");
+                }
+                else if (_workloadSetMode.Equals("loosemanifest", StringComparison.OrdinalIgnoreCase) ||
+                         _workloadSetMode.Equals("auto", StringComparison.OrdinalIgnoreCase))
+                {
+                    _workloadInstaller.AdjustInstallMode(_dotnetPath, _sdkFeatureBand, "false");
+                }
+                else
+                {
+                    throw new GracefulException(string.Format(LocalizableStrings.WorkloadSetModeTakesWorkloadSetLooseManifestOrAuto, _workloadSetMode), isUserError: true);
```
dsplaisted, line 98: I think if we use an Enum to represent the workload set mode (and use that enum in the option, ie `CliOption<WorkloadSetInstallMode>`), then System.CommandLine will handle this error checking for us.

Is this correct @baronfel?

dsplaisted (commented):
Looking good so far.

The mode isn't used yet, but we should probably have some tests that it is save and updated correctly.

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -272,6 +275,40 @@ protected uint RepairMsi(string productCode, string logFile)
...
+                SecurityUtils.CreateSecureDirectory(Path.GetDirectoryName(path));
+
+                File.WriteAllText(path, InstallingWorkloadCommand.AdjustInstallState("useWorkloadSets", File.Exists(path) ? File.ReadAllText(path) : null, singleValue: newMode));
+
+                SecurityUtils.SecureFile(path);
+            }
+            else if (IsClient)
+            {
+                InstallResponseMessage response = Dispatcher.SendMsiRequest(InstallRequestType.AdjustWorkloadMode,
+                    null, null, newMode, dotnetDir, sdkFeatureBand.ToString());
```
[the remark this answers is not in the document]
Forgind: I was actually thinking about going in the opposite direction and renaming this parameter something more generic like "argument"...every extra parameter is a small loss of performance, and I imagine we'll keep adding more things here until the requests are bulky and hard to use. I'd rather keep it slim and a bit generic. If you feel strongly, I can move towards new parameters instead.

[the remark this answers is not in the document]
dsplaisted: I don't feel very strongly either way.  Thanks!

src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerServer.cs
Forgind: This doesn't handle 'auto' because auto is already handled in the WorkloadUpdateCommand where it is the same as loosemanifest (for now) but will later be changed to being the same as workloadset. Since there are only two possible values at this stage, I'm not sure I see the value of changing it to an enum.

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
```diff
@@ -120,6 +120,14 @@ public string WorkloadId
             set;
         }
 
+        /// <summary>
+        /// The new mode to use: workloadset or loosemanifests
+        /// </summary>
+        public bool WorkloadMode
```
dsplaisted, line 126: Naming nit: Calling a boolean property `WorkloadMode` doesn't tell you what true or false values mean.  I think it would be better as `UseWorkloadSets`, `WorkloadSetsEnabled`, or similar.  This also applies to method parameters where this value is passed, etc.

```suggestion
        public bool UseWorkloadSets
```

src/Cli/dotnet/Installer/Windows/InstallRequestMessage.cs
```diff
@@ -33,7 +33,7 @@ public string ManifestPath
         /// The contents of the install state file. Each element corresponds to a single line of
         /// the JSON file to be written.
         /// </summary>
-        public string[] InstallStateContents
+        public Dictionary<string, string> InstallStateContents
```
dsplaisted, line 36: Naming nit: I think `InstallStateManifestVersions` would be a bit clearer here.

```suggestion
        public Dictionary<string, string> InstallStateManifestVersions
```

src/Cli/dotnet/commands/dotnet-workload/InstallStateContents.cs
```diff
@@ -0,0 +1,24 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.DotNet.Workloads.Workload
+{
+    internal class InstallStateContents
+    {
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        public bool? useWorkloadSets { get; set; }
```
dsplaisted, line 9: You can set up the Json serialization options to use camelCase, and then you can leave the C# properties PascalCase: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/customize-properties?pivots=dotnet-7-0#use-a-built-in-naming-policy

src/Cli/dotnet/commands/dotnet-workload/install/FileBasedInstaller.cs
```diff
@@ -452,21 +452,34 @@ public void GarbageCollect(Func<string, IWorkloadResolver> getResolverForWorkloa
...
 
-        public void DeleteInstallState(SdkFeatureBand sdkFeatureBand)
+        public void RemoveManifestsFromInstallState(SdkFeatureBand sdkFeatureBand)
         {
             string path = Path.Combine(WorkloadInstallType.GetInstallStateFolder(_sdkFeatureBand, _dotnetDir), "default.json");
             
             if (File.Exists(path))
             {
-                File.Delete(path);
+                var installStateContents = InstallStateContents.FromString(File.Exists(path) ? File.ReadAllText(path) : "{}");
```
dsplaisted, line 461: If the file doesn't exist, I think it's better to just create a new `InstallStateContents` object directly rather than parsing an empty JSON string.  See also other cases of this later in the file.

```suggestion
                var installStateContents = File.Exists(path) ? InstallStateContents.FromString(File.ReadAllText(path)) : new InstallStateContents();
```

src/Cli/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
```diff
@@ -170,6 +170,9 @@ Tool '{1}' (version '{2}') was successfully installed. Entry is added to the man
   <data name="LocalOptionDescription" xml:space="preserve">
     <value>Install the tool and add to the local tool manifest (default).</value>
   </data>
+  <data name="WorkloadSetMode" xml:space="preserve">
+    <value>Determine whether dotnet should use workload sets or loose manifests when installing or updating workloads.</value>
+  </data>
```
dsplaisted, line 175: I think this string is in the wrong .resx file (this one is for tool installation, not workload installation).  It looks the string is also in the right .resx, so I think this can just be deleted.

src/Cli/dotnet/commands/dotnet-workload/install/MsiInstallerBase.cs
```diff
@@ -272,6 +275,40 @@ protected uint RepairMsi(string productCode, string logFile)
...
+        protected void UpdateInstallMode(SdkFeatureBand sdkFeatureBand, bool newMode)
+        {
+            Elevate();
+
+            if (IsElevated)
+            {
+                string path = Path.Combine(WorkloadInstallType.GetInstallStateFolder(sdkFeatureBand, DotNetHome), "default.json");
+                // Create the parent folder for the state file and set up all required ACLs
+                SecurityUtils.CreateSecureDirectory(Path.GetDirectoryName(path));
+                var installStateContents = InstallStateContents.FromString(File.Exists(path) ? File.ReadAllText(path) : "{}");
```
dsplaisted, line 295: Same as the file based installer, I'd suggest creating an `InstallStateContents` object directly instead of parsing JSON if the file doesn't exist.

src/Tests/dotnet-workload-install.Tests/GivenFileBasedWorkloadInstall.cs
```diff
@@ -26,6 +27,46 @@ public GivenFileBasedWorkloadInstall(ITestOutputHelper log) : base(log)
...
+            (string dotnetRoot, FileBasedInstaller installer, _, _) = GetTestInstaller();
+            var stringFeatureBand = "6.0.300"; // This is hard-coded in the test installer, so if that changes, update this, too.
+            var sdkFeatureBand = new SdkFeatureBand(stringFeatureBand);
+            var path = Path.Combine(dotnetRoot, "metadata", "workloads", stringFeatureBand, "InstallState", "default.json");
+
+            // The install state shouldn't exist, but delete it in case it wasn't cleaned up properly from the last run.
+            if (File.Exists(path))
+            {
+                File.Delete(path);
+            }
```
dsplaisted, line 42: This shouldn't be necessary.  The test infrastructure automatically makes sure that when it creates a test directory, that it's empty.  If the directory already exists, it will first delete it.  If that doesn't work, the test will fail.

src/Tests/dotnet-workload-install.Tests/GivenFileBasedWorkloadInstall.cs
```diff
@@ -26,6 +27,46 @@ public GivenFileBasedWorkloadInstall(ITestOutputHelper log) : base(log)
...
+                { "third", "fourth" },
+            });
+            installState = InstallStateContents.FromString(File.ReadAllText(path));
+            installState.manifests.Count.Should().Be(2);
+            installState.manifests["first"].Should().Be("second");
+            installState.manifests["third"].Should().Be("fourth");
+            installState.useWorkloadSets.Should().BeTrue();
+
+            installer.UpdateInstallMode(sdkFeatureBand, false);
+            installState = InstallStateContents.FromString(File.ReadAllText(path));
```
dsplaisted, line 61: I would add a check here to be sure it still preserved the manifest versions.

```suggestion
            installState = InstallStateContents.FromString(File.ReadAllText(path));
            installState.manifests.Count.Should().Be(2);
```

src/Tests/dotnet-workload-install.Tests/MockPackWorkloadInstaller.cs
```diff
@@ -159,20 +165,24 @@ public void ReplaceWorkloadResolver(IWorkloadResolver workloadResolver)
...
         }
 
-        public void DeleteInstallState(SdkFeatureBand sdkFeatureBand)
+        public void RemoveManifestsFromInstallState(SdkFeatureBand sdkFeatureBand)
         {
             string path = Path.Combine(WorkloadInstallType.GetInstallStateFolder(sdkFeatureBand, _dotnetDir), "default.json");
             if (File.Exists(path))
             {
-                File.Delete(path);
+                var installStateContents = InstallStateContents.FromString(File.Exists(path) ? File.ReadAllText(path) : "{}");
```
dsplaisted, line 173: More instances where we can do `new InstallStateContents()` instead of parsing `{}`.

dsplaisted (approved):
Looks good.

See my comments, but feel free to merge when ready.

PR #37381, opened 2023-12-07
Check dependencies of Microsoft.DotNet.MSBuildSdkResolver against baseline

ladipro:
As pointed out in https://github.com/dotnet/msbuild/pull/9439#discussion_r1409619157, we have to be careful when changing the set of dependencies of the SDK resolver assembly as it may impact binding redirects in MSBuild's app.config.

This PR adds a simple post-build step which compares the dependencies of the just-built assembly with a baseline.

### DIFF
--- a/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Microsoft.DotNet.MSBuildSdkResolver.csproj
+++ b/src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Microsoft.DotNet.MSBuildSdkResolver.csproj
@@ -157,7 +157,7 @@
 
     <Error Text="$(AssemblyName) is expected to depend on %(ExpectedDependencies.Identity). $(DependencyMismatchErrorText)"
            Condition="!($([System.String]::Copy('$(ResolvedDependenciesList)').Contains('%(ExpectedDependencies.Identity)')))" />
-    <Error Text="$(AssemblyName) is not expected to depend on %(ResolvedDependencies.FusionName). (DependencyMismatchErrorText)"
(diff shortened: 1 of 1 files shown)


src/Resolvers/Microsoft.DotNet.MSBuildSdkResolver/Microsoft.DotNet.MSBuildSdkResolver.csproj
```diff
@@ -132,4 +132,33 @@
...
+    <!-- Check that the dependencies of the output assembly match our expectations -->
+    <PropertyGroup>
+      <ResolvedDependenciesList>@(ResolvedDependencies->'%(FusionName)')</ResolvedDependenciesList>
+      <ExpectedDependenciesList>@(ExpectedDependencies->'%(Identity)')</ExpectedDependenciesList>
+      <DependencyMismatchErrorText>This may have consequences for MSBuild.exe binding redirects, please get signoff from the MSBuild team.</DependencyMismatchErrorText>
+    </PropertyGroup>
+
+    <Error Text="$(AssemblyName) is expected to depend on %(ExpectedDependencies.Identity). $(DependencyMismatchErrorText)"
+           Condition="!($([System.String]::Copy('$(ResolvedDependenciesList)').Contains('%(ExpectedDependencies.Identity)')))" />
+    <Error Text="$(AssemblyName) is not expected to depend on %(ResolvedDependencies.FusionName). (DependencyMismatchErrorText)"
```
rainersigwald, line 160:
```suggestion
    <Error Text="$(AssemblyName) is not expected to depend on %(ResolvedDependencies.FusionName). $(DependencyMismatchErrorText)"
```

rainersigwald (approved):

PR #37394, opened 2023-12-08
add tool list format option

### LINKED ISSUE #37239: add output format support for dotnet tool list command
> I drafted a dotnet-tool `dotnet-update-all-tools` according to the dotnet tool list output

The exact format of the `dotnet tool list` output is not documented at <https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-list>.  "The command lists the package name, version installed, and the tool command", but this does not say anything about column headers, or about the "Manifest" column when listing local tools.  I suspect a future version of .NET SDK may change the format and break any software that parses it.  `dotnet tool list` doesn't yet support a `--format json` option like in `dotnet list package`.

_Originally posted by @KalleOlaviNiemitalo in https://github.com/dotnet/sdk/issues/19650#issuecomment-1833280431_

It would help us have an expected output consistently and also benefit the developers who want to use the output.

Currently, I drafted a dotnet tool to update all the dotnet tools installed according to the console output, while when the console output format changed, it would break, it would be better to support a consistent output such as json format, and keep using console as the default format

Usage:

``` sh
# global tools
dotnet tool list --global --format json

# local tools
dotnet tool list --local --format json
```

possible `--format` option values:

- `console`, default format as now
- `json`, output json text
- `csv`

possible JSON output details

```json
{
    "packageId": "dotnet-counters",
    "version": "8.0.0",
    "commands": "dotnet-counters"
}
```

(intended for better readability?)

possible CSV output details

```csv
packageId,version,commands
dotnet-counters,8.0.0,dotnet-counters
```

WeihanLi:
fixes #37239 

Introduce `--format` option for `dotnet tool list` command, support `json` for json output

for global tool list `dotnet tool list --global --format json`

```json
{ 
  "version": 1,
  "data": [
    {
      "packageId": "dotnet-execute",
      "version": "0.15.0",
      "commands": [
        "dotnet-exec"
      ]
    }
  ]
}
```

for local tool list `dotnet tool list --local--format json`

```json
{
  "version": 1,
  "data": [
    {
      "packageId": "dotnet-execute",
      "version": "0.15.0",
      "commands": [
        "dotnet-exec"
      ],
      "manifest": "/some/full/path/to/dotnet-tool.json"
    }
  ]
}
```


WeihanLi:
@dotnet/domestic-cat @baronfel could you please help have a look on this, thanks very much

src/Cli/dotnet/commands/dotnet-tool/list/ToolListCommandParser.cs
```diff
@@ -23,6 +23,11 @@ internal static class ToolListCommandParser
 
         public static readonly CliOption<string> ToolPathOption = ToolAppliedOption.ToolPathOption;
 
+        public static readonly CliOption<string> ToolListFormatOption = new("--format")
```
baronfel, line 26: Instead of a string, we should make a proper enum type here with `Table` and `Json` values - System.CommandLine can bind to these, and we get a nicer help text + validation for that. It also makes the 'switch' statement in the implementation more natural.

WeihanLi:
seemed failed test not related

baronfel:
I pushed a minor change to the enum casing and description after testing it locally. Thanks for doing this! I'll approve, but I'd like someone else on the team to take a look as well.

baronfel (approved):

nagilson:
I am happy to take a look once I wrap up a few other things 😄

src/Cli/dotnet/commands/dotnet-tool/list/ToolListJsonHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.DotNet.Tools.Tool.List;
+
+internal sealed class VersionedDataContract<TContract>
+{
+    [JsonPropertyName("version")]
```
nagilson, line 11:
```suggestion
        /// <summary>
        /// The version of the JSON format for dotnet tool list.
        /// </summary>
    [JsonPropertyName("version")]
```

This version is meant to be a version for this JSON data contract, right? Since there are two things just named 'version' I think it is best to specify that with a docstring.

nagilson (approved):
Thank you for your contribution and patience! Very cool that you just went ahead and implemented what you were looking for. I definitely think we need to get this feature documented for other people to use. I appreciate that you added tests and it is all a very clean change!

PR #37412, opened 2023-12-09
add devcontainer support

### LINKED ISSUE #21299: Support for codespaces

When I try to build the repository with codespace, I got out of space error, by default the image would install some unnecessary packages, it would be great if we could have better support for codespace, thus people could make their contributions much more easily.

Add specific codespace config for the repo, likes what had been done in asp.net core repo <https://github.com/dotnet/aspnetcore/blob/main/.devcontainer/devcontainer.json>

WeihanLi:
fixes #21299


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v8.0/containers/dotnet
+{
+  "name": "dotnet",
+  "build": {
+    "dockerfile": "Dockerfile",
```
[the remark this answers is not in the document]
WeihanLi: Thanks for the feedback.

The main purpose of using `Dockerfile` is to keep some customization easy, nothing for now, so we could update to use `Dockerfile` when needed instead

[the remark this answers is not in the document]
baronfel: What customization are you anticipating? This version of the SDK installed isn't even really mandated by the base image, because the build.cmd script in the repository is going to redownload all required SDKs and Runtimes.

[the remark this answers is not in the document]
baronfel: (to be clear, this is great and I can't wait to merge it in)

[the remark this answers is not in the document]
WeihanLi: Maybe some system requirements like some libgdi+ for some test case, just some wondering, maybe not a real valid scene

.devcontainer/scripts/post-creation.sh
```diff
@@ -0,0 +1,4 @@
+# Install SDK and tool dependencies before container starts
+# Also run the full restore on the repo so that go-to definition
+# and other language features will be available in C# files
+./restore.sh
```
baronfel, line 4: For this command, I think you may also need to run eng/sdk-build-env.sh as well to ensure everything is set up correctly for VSCode.

baronfel (approved):
This is great! Thanks for adding it, it'll be great to have a ready to go dev environment for this repo.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
WeihanLi: > This version of the SDK installed isn't even really mandated by the base image, because the build.cmd script in the repository is going to redownload all required SDKs and Runtimes.

Updated base image to `mcr.microsoft.com/devcontainers/base:debian` since we do not need to have dotnet installed. Use `debian` tag since we're using `debian` for the default dotnet docker image

.devcontainer/scripts/post-creation.sh
WeihanLi: Thanks, will have a check and test

WeihanLi: @baronfel , I did not find the `eng/sdk-build-env.sh` file, is there a typo?

image

baronfel: It's created during the build.sh/restore.sh calls:

https://github.com/dotnet/sdk/blob/main/eng%2Frestore-toolset.sh

It sets up the environment variables for the .NET SDK to force it to use the just-downloaded SDK and Runtime(s).

WeihanLi: Seemed file not found from my test
image

baronfel: ah, that's because only the `build` command in the build.sh script calls the generation. I'd suggest that instead of calling `restore.sh` in your post-creation.sh script, you should call `./build.sh`. This will be the most likely way to get a usable installation, in my opinion.

WeihanLi: Ah, I found it in `artifacts/sdk-build-env.sh`, the `restore.sh`/`build.sh` would produce the script both.

From the script content, seemed there's duplicate `/` in the dotnet path, does this matter?
image

It may come from 
https://github.com/dotnet/sdk/blob/a5caad8ad13f3564574f6ac459f3ed43320cb97f/eng/common/tools.sh#L144-L146

baronfel: it _shouldn't_ matter, but it's worth fixing. good spot!

WeihanLi: from my test, the `sdk-build-env.sh` would not keep effective when execting, any special flag when executing?

image

think maybe it's not a need to execute, or should I configure the environment variables in the devcontainer config like `PATH` do https://github.com/dotnet/aspnetcore/blob/f26b79eb34d13648bedc5ec98e2151f8c55db978/.devcontainer/devcontainer.json#L33

baronfel: That's bizarre. If the script isn't working, I think setting the environment variables explicitly like aspnetcore is a fine way to go. Thanks for taking the time to dig into this.

WeihanLi:
@baronfel @nagilson could you please help try to rerun the CI build, seemed the failed test case is not related
image

baronfel:
That test in particular has been failing very often today.

WeihanLi:
> That test in particular has been failing very often today.

failed again 😢

baronfel:
@dotnet/razor-compiler is there a known issue with Razor? This test has been failing on multiple SDK CI flows today.

jjonescz:
tracking the razor issue here as it affects many (all?) PRs: https://github.com/dotnet/sdk/issues/38942

nagilson:
New failure across the repo, it is not related to your change. Hope we fixed it, but we will see

PR #37429, opened 2023-12-11
Add `--arch` and `--os` options to commands that already support `--runtime` option

### LINKED ISSUE #37245: New `--os` argument for `dotnet restore`
## Summary
Currently, the `dotnet restore` command lacks an option to specify the target operating system directly, similar to the existing options in `dotnet build` and `dotnet publish`. I propose the addition of a new argument: `--os <OS>`.

## Motivation
The motivation behind this request is to align the `dotnet restore` command with the capabilities of `dotnet build` and `dotnet publish`, which both support the `--os` argument. While `dotnet restore` already has the `-r|--runtime` option to specify the target runtime identifier, having a dedicated `--os` argument would provide consistency.

## Usage
```bash
dotnet restore --os linux
```

## Additional Information
N/A

ghost:
Fixes #37245.

I'll create a PR for updating the documentation once this PR has been approved.


src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
```diff
@@ -60,14 +60,15 @@ private static CliCommand ConstructCommand()
             var command = new DocumentedCommand("restore", DocsLink, LocalizableStrings.AppFullName);
 
             command.Arguments.Add(SlnOrProjectArgument);
+            command.Options.Add(CommonOptions.ArchitectureOption);
```
joegoldman2, line 63: Currently `--arch` option is added even if the runtime option is not available. Should we move both architecture and OS options line 183 (`if (includeRuntimeOption)`) to make them available only if the runtime option is available as well?

joegoldman2:
@baronfel I added several test cases.

Currently, the restore command already supports `--arch` option. Passing both `--arch` and several `--runtime` options, on a Windows machine, like this for example: `dotnet restore --runtime linux-x64 --runtime osx-x64 --arch arm64` will produce the following MSBuild command:
`msbuild.exe -maxcpucount -verbosity:m -nologo -target:Restore -property:RuntimeIdentifier=win-arm64 -property:RuntimeIdentifiers=linux-x64%3Bosx-x64`
So I considered that having both `RuntimeIdentifier` and `RuntimeIdentifiers` in the final command is correct (current behaviour).

With this PR, passing both `--os` and several `--runtime` will produce the same kind of result with both `RuntimeIdentifier` and `RuntimeIdentifiers` in the final MSBuild command (a test case is covering this case).

src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
joegoldman2, line 63: Currently `--arch` option is already supported and added even if the runtime option is not available. Should we move both architecture and OS options line 183 (`if (includeRuntimeOption)`) to make them available only if the runtime option is available as well?

joegoldman2:
Is there anything you are waiting on before you can begin the review? @baronfel

joegoldman2:
@baronfel sorry for asking, is there anything missing for you or someone else to start reviewing this PR? Thank you!

PR #37459, opened 2023-12-12
Add `--framework` option to restore command

### LINKED ISSUE #35408: `dotnet restore` does not support `-f TFM`
### Describe the bug

I was trying to restore and build a specific TFM, and I found that `dotnet restore` does not like the `-f` flag.

### To Reproduce

```
dotnet new maui
dotnet restore -f net8.0-windows10.0.19041.0
```

### Exceptions (if any)

```
MSBUILD : error MSB1008: Only one project can be specified.
    Full command line: 'C:\Program Files\dotnet\sdk\8.0.100-rc.1.23463.5\MSBuild.dll -maxcpucount -verbosity:m -nologo -target:Restore -property:RestoreForce=true --property:PublishReadyToRun=false -property:RuntimeIdentifiers=win10-x64 CrashTestWin.csproj net8.0-windows10.0.19041.0 -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,C:\Program Files\dotnet\sdk\8.0.100-rc.1.23463.5\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,C:\Program Files\dotnet\sdk\8.0.100-rc.1.23463.5\dotnet.dll'
  Switches appended by response files:
Switch: net8.0-windows10.0.19041.0

For switch syntax, type "MSBuild -help"
```

### Further technical details
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and its version

ghost:
Fixes #35408.

- I didn't reuse `CommonOption.FrameworkOption` option because it has an `-f` alias that is already used as an alias for the `--force` option in the restore command.
- Only one `--framework` argument is allowed when using this argument.
- I didn't translate the new resx entries.
- I'll create a PR for updating the documentation if this PR is accepted.

### DIFF
--- a/src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
+++ b/src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
@@ -25,6 +25,13 @@ namespace Microsoft.DotNet.Cli
         }.ForwardAsSingle(o => $"-property:RestoreSources={string.Join("%3B", o)}")
         .AllowSingleArgPerToken();
 
+        public static readonly CliOption<string> FrameworkOption = new ForwardedOption<string>("--framework")
+        {
+            Description = LocalizableStrings.CmdFrameworkOptionDescription,
+            HelpName = LocalizableStrings.CmdFrameworkArgument
+        }.ForwardAsSingle(o => $"-property:TargetFramework={o}")
+        .AddCompletions(Complete.TargetFrameworksFromProjectFile);
+
         private static IEnumerable<CliOption> FullRestoreOptions() =>
             ImplicitRestoreOptions(true, true, true, true).Concat(
                 new CliOption[] {
@@ -60,6 +67,7 @@ namespace Microsoft.DotNet.Cli
             var command = new DocumentedCommand("restore", DocsLink, LocalizableStrings.AppFullName);
 
             command.Arguments.Add(SlnOrProjectArgument);
+            command.Options.Add(FrameworkOption);
             command.Options.Add(CommonOptions.DisableBuildServersOption);
 
             foreach (var option in FullRestoreOptions())
--- a/src/Cli/dotnet/commands/dotnet-restore/LocalizableStrings.resx
+++ b/src/Cli/dotnet/commands/dotnet-restore/LocalizableStrings.resx
@@ -135,6 +135,12 @@
   <data name="CmdSourceOptionDescription" xml:space="preserve">
     <value>The NuGet package source to use for the restore.</value>
   </data>
+  <data name="CmdFrameworkArgument" xml:space="preserve">
+    <value>FRAMEWORK</value>
+  </data>
+  <data name="CmdFrameworkOptionDescription" xml:space="preserve">
+    <value>The target framework to restore for. The target framework must also be specified in the project file.</value>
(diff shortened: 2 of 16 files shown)


src/Cli/dotnet/commands/dotnet-restore/RestoreCommandParser.cs
```diff
@@ -25,6 +25,13 @@ internal static class RestoreCommandParser
         }.ForwardAsSingle(o => $"-property:RestoreSources={string.Join("%3B", o)}")
         .AllowSingleArgPerToken();
 
+        public static readonly CliOption<string> FrameworkOption = new ForwardedOption<string>("--framework")
+        {
+            Description = LocalizableStrings.CmdFrameworkOptionDescription,
+            HelpName = LocalizableStrings.CmdFrameworkArgument
+        }.ForwardAsSingle(o => $"-property:TargetFramework={o}")
+        .AddCompletions(Complete.TargetFrameworksFromProjectFile);
+
```
joegoldman2, line 34:
```suggestion
        public static readonly CliOption<IEnumerable<string>> FrameworkOption = new ForwardedOption<IEnumerable<string>>("--framework")
        {
            Description = LocalizableStrings.CmdFrameworkOptionDescription,
            HelpName = LocalizableStrings.CmdFrameworkArgument
        }.ForwardAsSingle(o => $"-property:TargetFrameworks={string.Join(";", o)}")
        .AllowSingleArgPerToken()
        .AddCompletions(Complete.TargetFrameworksFromProjectFile);
```

We can probably change the option this way in case we would like to allow multiple `--framework` option to be specified.

joegoldman2, line 34:
```cs
public static readonly CliOption<IEnumerable<string>> FrameworkOption = new ForwardedOption<IEnumerable<string>>("--framework")
{
    Description = LocalizableStrings.CmdFrameworkOptionDescription,
    HelpName = LocalizableStrings.CmdFrameworkArgument
}.ForwardAsSingle(o => $"-property:TargetFrameworks={string.Join(";", o)}")
.AllowSingleArgPerToken()
.AddCompletions(Complete.TargetFrameworksFromProjectFile);
```

We can probably change the option this way in case we would like to allow multiple `--framework` option to be specified.

marcpopMSFT:
We're not sure we want to take this as it wouldn't make sense for solutions or project references where the TF value doesn't flow. The nuget targets join the TF and TFs values together when restore so there'd be that to consider as well.

I think we'd need to understand more from @mattleibow what the scenario was he was trying to do (was it just a quick test of something or a more enduring behavior expected for MAUI apps).

CC @nkolev92 as well for a nuget opinion.

nkolev92:
I'd say restore not accepting a framework argument is preferable. 
Other commands can forward those arguments, but for restore itself, it's a project operation, there's only 1 assets file, so it'd just lead to stomping as the TFM arguments change, even though they're subsets.

joegoldman2:
If it's ok for everyone, we can close this PR and the associated issue.

marcpopMSFT:
Thanks for the PR. This is a complex area with a lot of history.

PR #37492, opened 2023-12-13
Enable TerminalLogger by default for .NET 9 SDKs

baronfel:
Now that the MSBuild team has cleared out the blocking bugs for Terminal Logger discovered as part of the .NET 8 GA release, we'd like to enable it by default for .NET 9. This performs that enablement, and should help us get broad feedback on the feature as soon as possible.

As before, Terminal Logger can be disabled on a per-command basis with `-tl:off`.


rainersigwald (approved):

MiYanni (approved):
I guess this means this repo has no tests that use build output for validation?

baronfel:
> I guess this means this repo has no tests that use build output for validation?

Either that, or this change doesn't actually so what is expected :)  In any case we need some tests to verify the behavior.

baronfel:
> > I guess this means this repo has no tests that use build output for validation?
> 
> Either that, or this change doesn't actually so what is expected :) In any case we need some tests to verify the behavior.

It was actually a bug :D I've verified locally now that this change _does something_ so now let's see what breaks.

MiYanni (approved):
Thanks for fixing the tests.

ViktorHofer:
Please hold this change off for one day. We need to get a new SDK out and I would prefer to get this change in afterwards.

baronfel:
@ViktorHofer I added the 'do not merge' label as a warning for us.

baronfel:
We are probably going to hold this until after the holidays in general because it may require action by folks that are out on holiday.

baronfel:
Breaking change doc issue has been written, and folks are mostly back from the holidays. I'm going to merge this and then send an email to the breaking change alias to give folks a heads-up.

PR #37516, opened 2023-12-14
Fix sln add for duplicate projects under different paths with same root

marcpopMSFT:
Change the logic for how we determine collisions in the SLN file. The item blocked by VS and msbuild is when one solution folder has two items parented by it that have the same name (whether they are projects or solution folders).

Instead of trying to figure out the solution folder and matching it to existing ones, I moved to checking the exact parent at the time of adding it which should be more precise.


baronfel (approved):
Thanks for fixing this! I took it for a spin locally and was able to add projects of matching names (but non-matching paths) to the solution as expected.

PR #37625, opened 2023-12-20
Base version

Forgind:
Otherwise, this causes a StackOverflow Exception

(I'd tested in a prior version where the method names were distinct)


src/Cli/dotnet/commands/dotnet-workload/install/NetSdkMsiInstallerClient.cs
```diff
@@ -186,7 +186,7 @@ public void GarbageCollect(Func<string, IWorkloadResolver> getResolverForWorkloa
             RemoveManifestsFromInstallStateFile(sdkFeatureBand);
 
         public new void SaveInstallStateManifestVersions(SdkFeatureBand sdkFeatureBand, Dictionary<string, string> manifestContents) =>
-            SaveInstallStateManifestVersions(sdkFeatureBand, manifestContents);
+            base.SaveInstallStateManifestVersions(sdkFeatureBand, manifestContents);
```
dsplaisted, line 189: Would it be simpler to make the base method public and delete this derived version?

dsplaisted (approved):

PR #37659, opened 2023-12-26
Handle HTTP 403 Forbidden in Containers

ltrzesniewski:
I got the following error when uploading an image to AWS ECR:

> Containerize: error CONTAINER004: CONTAINER1014: Manifest pull failed.

This is a very generic error message, and I had to look in the binlog to see it was caused by a 403 Forbidden HTTP response.

This PR changes this behavior so that 403 Forbidden gets you the same error as 401 Unauthorized. The error message currently provided fits both kinds of errors:

> Unable to access the repository '{0}' in the registry '{1}'. Please confirm your credentials are correct and that you have access to this repository and registry.


baronfel (approved):
Great idea, thanks for adding this!

baronfel:
@ltrzesniewski if it's ok with you I'd like to backport this to the 8.0.200 release after this PR merges. I think it's a low-risk fix that makes the user experience much better in this situation.

ltrzesniewski:
@baronfel of course, please go ahead 🙂

PR #37662, opened 2023-12-27
Add `CannotReOrderStructProperties` rule to API-Compat

ShreyasJejurkar:
1. This PR adds a new rule to API-Compat tooling where it will issue a diagnosis if the struct properties are re-ordered. 

Fixes #26460


ShreyasJejurkar:
> "Compat" rather than "Compact"

Oops, sorry. Changed it. There is one issue am facing please see attached issue after 10 mins.

ShreyasJejurkar:
> The order of _fields_ matters in [sequential-layout types](https://learn.microsoft.com/dotnet/api/system.type.islayoutsequential?view=netstandard-2.0) (which can be classes too, not just structs) but this PR currently checks the order of _properties_.

Ohh ok. Issue states for struct, so am targeting that now and already facing issues. 😅. But I will have look into this as well.

ericstj:
@ShreyasJejurkar did you have plans to return to this PR?

